PR #17, opened 2015-07-10
Import Configuration Service

jacquemier:
The use cases are available on redmine:
https://forge.in2p3.fr/projects/pipelines-use-cases/wiki

It is base on Python library Argparse, more specifically ArgumentParser class.
Use of Configution class is exactly the same  than use of ArgumentParser
1/ Define which command line option you expect.
2/ Call the parse_arg method.
3/ Command line option are accessible through Configuration instance data member

In addition, Configuration service allows to write and read configuration option (section key/value comment) to/from  file.
3 implementations exist, one for simple text file, one is using Fits header and one using Fits data table.
The  use case  "choosing what modules run in the pipeline is not implemented for the time being.


ctapipe/configuration/configuration.py
```diff
@@ -0,0 +1,312 @@
+'''
```
cdeil: Having `configuration/configuration.py` can lead to confusion.
What Astropy does is instead call it `configuration/core.py` ... OK to rename?

ctapipe/configuration/configuration.py
```diff
@@ -0,0 +1,312 @@
...
+https://docs.python.org/2/library/argparse.html
```
cdeil: Remove this reference to argparse.

ctapipe/configuration/configuration.py
```diff
@@ -0,0 +1,312 @@
...
+class Configuration(ArgumentParser):
+    DEFAULT       = "DEFAULT"
+    FITS          = "FITS"
+    INI           = "INI"
+    DATAIMPL      = "DATAIMPL"
+    HEADERIMPL    = "HEADERIMPL"
+    VALUE_INDEX   = 0
+    COMMENT_INDEX = 1
+
+    #--------------------------------------------------------------------------#
```
cdeil: Remove these comment lines (let's follow the common Python / Numpy / Scipy / Astropy coding style)

ctapipe/configuration/configuration.py
```diff
@@ -0,0 +1,312 @@
...
+'''
+
+
+from argparse import ArgumentParser
+from configparser import RawConfigParser
+from astropy.io import fits
+
+import numpy as np
+import sys
+
```
cdeil: Add `__all__` list with the classes / functions that are part of the "public API" and not implementation details.
Only if you do this, they will show up in the Sphinx docs.

ctapipe/configuration/__init__.py
```diff
@@ -0,0 +1,5 @@
+# Licensed under a 3-clause BSD style license - see LICENSE.rst
+"""
+Configuration
+"""
+
```
cdeil: Add this after renaming `configuration.py` to `core.py`:
```
from .core import *
```

ctapipe/configuration/test/TestConfiguration.py
```diff
@@ -0,0 +1,200 @@
+#/bin/env python3
```
cdeil: Use pytest (not unittest) and follow these testing guidelines:
http://astropy.readthedocs.org/en/latest/development/testguide.html

It means re have to re-write this quite a bit:
- File name should be `test_core.py`
- `#!` line and docstring at the top and `__main__` part at the bottom can be removed
- Test asserts have to be rewritten to just use asserts like is common in pytest (check astropy tests for examples)

ctapipe/configuration/test/__init__.py
```diff
@@ -0,0 +1,5 @@
+# Licensed under a 3-clause BSD style license - see LICENSE.rst
```
cdeil: Please rename the `test` folder to `tests`.
The `__init__.py` file in `tests` can be completely empty ... no need for this license line and docstring.

ctapipe/configuration/configuration.py
```diff
@@ -0,0 +1,312 @@
...
+        
+
+        hduList.writeto(filename,clobber=True)
+        
+                
+        #table_0.append(filename, data=data, header=header, checksum=False, verify=True)
+
+    
+    #--------------------------------------------------------------------------#
+    def read(self, filenames, type=FITS, implementation=DATAIMPL, encoding=None):
```
cdeil: `type` is a Python built-in keyword and re-setting it to something else can cause big confusion.
Maybe use `format` instead?

cdeil:
@jacquemier  – Thanks for the PR!

I've left a bunch of superficial inline comments about file names / coding style / how to write tests.
Basically I think we should mainly follow what's done in Astropy:
http://astropy.readthedocs.org/en/latest/#developer-documentation
It's pretty good and it means we don't have to write a ton of developer documentation / guidelines.
Of course some things will be different / better here, e.g. because we don't have to support Python 2.

One more thing you should do with every PR is check the continuous integration travis-ci build logs for errors and fix those:
https://travis-ci.org/cta-observatory/ctapipe/builds/70367162

I can do more coding style / Python implementation code review, but first @kosack should review this and comment if the functionality is what he wants.

ctapipe/configuration/configuration.py
jacquemier: I agree, format is better than type.

ctapipe/configuration/configuration.py
jacquemier: Yes, I agree

PR #21, opened 2015-08-31
Add pipeline example for configuration use case implementation

jacquemier:
This is a very simple  pipeline.
Only one stage can be connected to one Producer and one Consumer.
It allows to test Configuration service implementation.


jacquemier:
It seems to be to early for pipeline implementation

kosack:
Yes probably, but it would be nice maybe to put it in the
examples/brainstorm directory. We can collect there idea for pipelining.

PR #30, opened 2015-11-03
call self.axes._sci instead of plt.sci, fix #28

### LINKED ISSUE #28: CameraDisplay colorbar breaks when multiple displays per figure
CameraDisplay now uses `plt.sci(self.pixels)` to set the *current image* to update the color bar if it exists. 

This fails with a ValueError if more than one plot is on a figure (see e.g. `examples/camera_display_multi.py`).  

For now, I catch the ValueError and ignore it, but that is not a good solution as the color bar does not get updated in that case. 

@MaxNoe do you know how to fix that?

MaxNoe:


kosack:
Thanks! It's strange they don't give a public object-oriented api function for that (other than the high-level one), but at least this should work unless they change the private `_sci` method.

cdeil:
Accessing private parts of matplotlib (or any package really) is always means there's a good chance your package will break and some poor guy has to fix this later.
If you have the time, it would be good to ask on the matplotlib mailing list or stack overflow if there's a "proper" way to do this and, if not, file a feature request.

PR #45, opened 2015-11-18
hillas_parameters_2 is inside ctapipe.reco.hillas, not ctapipe.reco

migueln:
Small fix in camera_display.ipynb notebook.


kosack:
in fact, you should just use hillas_parameters, not hillas_parameters_2 (the first is an alias for the second). I cleaned that up recently, so all notebooks should just drop the _2

Issue #48, opened 2015-12-02
Improve instrumental data classes


kosack:
__Make  `CameraGeometry` and `ArrayLayout` subclasses of `astropy.table.Table`?__

Right now we have 2 instrumental data classes: `CameraGeometry` and `ArrayLayout` that will eventually be merged into a comprehensive instrument model.  However, their implementation leaves a lot to be desired. 

Currently, `CameraGeometry` is simple a class with members like `pix_x`, `pix_y` that should contain an array of pixel positions.  There is separate support to read them from various formats. 

There is a common use case (working with  cleaned images) that is made difficult by this implementation. After an image is cleaned, one has a boolean _mask_ that can be applied to the image data.  To apply that also to the pixel positions requires masking each member of `CameraGeometry` manually:
```python
geom = CameraGeometry(...)
image = ...
mask = tailcuts_clean(geom, image)

cleaned_image = image[mask]
cleaned_x = CameraGeometry.pix_x[mask]
cleaned_y = CameraGeometry.pix_y[mask]
```
etc.

It would be nicer to allow slicing on the whole `CameraGeometry` object, but currently it contains both vector quantities and scalar quantities like `pix_type`.  

```python
cleaned_image = image[mask]
cleaned_geom = geom[mask]
```

One suggestion: use `astropy.table.Table` as the base class for all instrument classes.  

* Advantages
  * get automatic "slicing" of all data (as above)
  * get automatic read/write to all table formats (so data model is directly connected to the class)
  * get nice print-outs (pretty-printed columns, html in notebook mode)
  * can "browse" the data structures in a web interface
  * could extend to add more reader/writer routines like the current classes
* Disadvantages:
  * can access members by "dot" syntax (would need `geom['pix_x'] rather than `geom.pix_x`)
  * scalar quantities would need to be in the metadata (which is probably correct anyhow), e.g. `geom.meta.pix_type = 'hexagonal'`

If implemented this way, you can do:
```python
geom = CameraGeometry.from_file("test.simtel.gz",5)
geom.write("CT5.fits")
```

cdeil:
@kosack - Sounds like it's worth a try.

A word of warning: my impression is that `Table` is a pretty complex (but well written) class, so it can be complex to sub-class and get the exact behaviour you want (e.g. when it comes to the constructor, `read`, `write` and `info` methods.

[...]

PR #49, opened 2015-12-03
Changes to Instrument Module

justuszorn:
I created a first version of the Instrument Module.
+ able to read (some) hessio, fits, and ASCII sim_telarray files
+ a unit test is included
+ an example is included

All functions of io/camera.py have been merged into the instrument module. 
If you agree with that or at least with the main idea, I would change all files where currently the CameraGeometry class is called and replace the corresponding parts by the new Instrument Module.

Next week, I will add a possibility of using "fake data" so that some functions can be used for testing even without giving a file, e.g. the function "make_rectangular_camera_geometry" which currently is in io/camera.py, but also already in instrument/telescope/camera/CameraDescription.py.


kosack:
Here are a few comments:

1. the directory structure is too complex (it looks like a Java program instead of python): it really necessary to have such a deep hierarchy of modules? Couldn't we just put them all as separate files in the instrument module? (rather than instrument.telescope.optics.TelescopeDescription, just have instrument.TelescopeDescription).  The only reason I can see would be if there are many classes in each subdirectory, but there are only one so far. All of that should be hidden from the user, who just wants to do something like:
```python
from instrument import OpticsDescription

od = OpticsDescription.from_file("test.eventio.gz")
od = OpticsDescription.from_metadata(obsid=123456, telid=56)
```
etc.

More comments:
* I find the API is a bit confusing. Perhaps if instead of the user having to construct a separate file object and then pass it, that should be hidden (it can happen automatically inside)
* see #48 . It may be useful to leverage `astropy.table` and it's many i/o backends to do the reading/writing.  That would simplify a lot of code by removing the need to have separate initilizers and things like that.
* run `autopep8` on all the files to fix the formatting (they do not follow the standards)
* don't use camelCase in member variables (or at least we should agree on a standard)
* there are a lot of functions (and even classes!)  called "initialize"  that do many different things, and that is very confusing. Use more descriptive names. 
    * e.g. rename `InstrumentDescription.initilize()` to just be `InstrumentDescription.from_file("filename")`  
    * The `Initilize` classes are also confusing. Why are they a class? why is there also a function (non-member) called initialize?  I think I understand why you use them: they provide a generic way to load data with multiple backends. That begs for a class hierarchy. E.g. DescriptionReader -> TelescopeDescriptionReader, etc...  Again, those are generally hidden from the user, but it should also not be confusing to developers either

Finally (not really comments, but just musings)
*  Preferably we should hide the detail of where the data are read from from the user (in a future version).  So when the user tries to get a CameraDescription for a run, it would either look in a local directory of files, or in a database (which method would be configurable globally) . That way if we launch a job on a cluster that has access to the CTA database, the database could be used, but on a user system, a directory with FITS files containing the same data could be used).  That probably already would work with your model, just by implementing a database backend. 

* This does start to satisfy the requirement that we abstract I/O however, and maybe we shuold think about generalizing it to other parts of the code (I have to think about that) FOr reference the relevant requirement was: 
_The software framework will provide an I/O layer abstraction that hides the complexities of data access from the user, and allows access to all the various file formats and databases used in CTA through a common interface. The I/O layer is responsible for reading or writing data from/to a serialized format > into the in-memory data model (e.g. data structures used by the algorithm components). Minimally FITS and the CTA Raw data format will be supported._

justuszorn:
Thank you for the comments. Some questions/comments from my side:
- changed the directory structure and hierarchy of the modules:

kosack:
Thanks! I'll take a look at the new structure shortly.

> Did I understand you correctly that you mean that by using astropy.table I would be able to skip the   various reading methods ('from_file_*') for the different file extensions? I don"t understand how that could be possible. E.g. how can I read the hessio files by using astropy.table?

My idea was to use astropy.table as a base, but extend it to also read from other formats, so by default it would work with any format supported by astropy.table (fits, hdf, ascii, etc), but we'd have to add io modules for hessio and database tables (since in the final software, I suspect a lot of instrument config stuff will be stored in a database, not a file).  But the structure of `astropy.table.Table` is somewhat similar to what we need: it is a dict of columns that contain values with units and names, along with metadata for "scalar" values and other info.  The advantage is that a lot of work has been done.

Another option would be to have our own base class and internally use a `Table` instance, ranther than sub-classing it. That is more work, but maybe more flexible if they make changes to astropy.table.

justuszorn:
I made some changes to the instrument module. First I have to stress that this version is not ready yet for merging to the main branch! However, I would be very happy of getting feedback and comments.
What is new in this version:
The user reads a file with instrument data (currently either a fits or a hessio file). All data is written into astropy.table-Tables. This is very easy when the original file is a fits file because astropy already has a fits I/O to read fits file. However, for reading a hessio file, I just wrote a function which reads all the data from the hessio file and writes it step by step into the table (see ctapipe.instrument.util_functions.load_hessio(filename) ).
Now, at this point we could stop, i.e. the return object of the reading function could just be the astropy table. The user would have to type sth. like "instr_table['FOV']" and the column with the FOV of all telescope would be printed.
The other way (and this is how it is currently implemented) is that the way of accessing the data is the same as before, i.e. the user has to do sth. like: "cam.fov", where "cam" is a class object. This maybe is a little easier because then the user can - already when reading the file - specify which data of which telescope he wants to get and does not have to look through the whole table to get the data of the telescope he wants.
Before I continue, I would like to discuss the following things:
- Adding a general reading function for hessio files to the astropy I/O would be nice, but is - from my understanding - somehow time-consuming and maybe (??) not really necessary.
- It is not yet clear to me, how I should add the data I read from the simtelarray-config-files to the table and even more important: Is it really necessary to have a reading function for these files? Will the data flow not just be in the other direction in the final version, i.e. the simtelarray-config-files will be created out of the telescope data which will be stored in a database or sth similar. This mean, we would need a function in the instrument module which creates these config-files and not a function which reads them.

justuszorn:
Again some changes to the instrument module, not finished but quite a number of things changed. I would be happy if you have a look at it in case you find some time. Especially the example file examples/read_instrument_description.py could be helpful in understanding the things.

kosack:
Thanks! I will look at it shortly.  For the question of reading from HESSIO files, I think that is not a critical feature, since hessio will not be the final format that we use (so even a hacked solution is fine for now, or even a converter that just dumps the info to FITS files).  In the future when we have a true central archive system, the instrument config will be read/written from/to some kind of database, e.g. mysql or perhaps a non-relational db, but exactly what solution has not yet been determined.  So a future feature would be to implement the ability to do that, which should be fairly straightforward if the internal Table schema is identical or close to the database table schema. That will be much easier than mapping to a custom format like hessio.

justuszorn:
To be able to use the example file "examples/read_instrument_description.py" you have to add some datasets to "ctapipe-extra/datasets" which you can find here: [https://github.com/justuszorn/ctapipe-extra/tree/master/datasets](url).

justuszorn:
I don't understand why the Travis CI build fails... It has something to do with importing the pyhessio module. Could maybe someone explain that to me? And of course you can give me some feedback as soon as someone has the time for that ;)

jacquemier:
I simplified pyhessio installation.
Now, adding these 3 lines to .travis.yml is enough to build it and install it.

    # PYHESSIO DEPENDENCY
    - git clone https://github.com/cta-observatory/pyhessio
    - cd pyhessio
    - python setup.py install

I have just test it on travis(branch  pyhessio_travis), pyhessio installation is correct but travis has another problem when trying to download setuptools

As soon as travis will come back to normal, I will commit a new .travis.yml

jacquemier:
Justus, pyhessio installation step for travis is valided and commit on master branch.

jacquemier:
travis SETUP_CMD='build_sphinx -w' still produce error because of pyhessio import module failed.
So I think your module will also fail for same reason.
I still working on this issue.

https://travis-ci.org/cta-observatory/ctapipe/jobs/111758307

jacquemier:
I finally manage to install pyhessio module during travis test.
Your pull request should not failed anymore.
Could you please merge last  master branch commits on your branch and send a PULL request again ?

justuszorn:
It still fails...

PR #51, opened 2015-12-17
Implementations, documentation and changes related with the calibration

rdelosreyes:
- Contains a new structure of the calib module in agreement with the different calibration modules in the WBS. Incluides README files for the code developers.
- Implementation code to calibrate MC simtel array files. Python implementation of the current functions in reconstruction.c (hessioxxx) and corresponding test functions.
- Example macro of how to calibrate the MC files (tested only in Prod2).
- First draft on the documentation about the different calibration modules.

Note: Test functions does not seems to work well due to some issues with the camera geometry.


ctapipe/calib/camera/mc.py
```diff
@@ -0,0 +1,578 @@
...
+    if integrated_charge is None:
+        return None
+    pe_pix_tel = []
+    for ipix in range(0, get_num_pixels(telid)):
+        pe_pix = 0
+        int_pix_hg = integrated_charge[get_num_channel(telid)-1][ipix]
+        # If the integral charge is between -300,2000 ADC ts, we choose the HG
+        # Otherwise the LG channel
+        # If there is only one gain, it is the HG (default)
+        if (int_pix_hg > -1000 and int_pix_hg < 10000 or
```
kosack: this fails for me with an error "ValueError: The truth value of an array with more than one element is ambiguous. Use a.any() or a.all()."
Maybe you need  `(int_pix_hg >-1000).any()`?

ctapipe/calib/camera/mc.py
```diff
@@ -0,0 +1,578 @@
...
+
+In general the integration functions corresponds one to one in name and
+functionality with those in hessioxxx package.
+The same the interpolation of the pulse shape and the adc2pe conversion.
+
+"""
+
+
+def qpol(x, np, yval):
+    ix = round(x)
```
kosack: This function fails for me because `round()` doesn't exist. I think you need to do:
```python
from numpy import round
```
at the top

MaxNoe:
That's strange, round is a python builtin:
https://docs.python.org/3.5/library/functions.html#round

PR #53, opened 2016-02-26
zfits module prototype for ctapipe

migueln:
Zfits module prototype for ctapipe (very preliminary)
Add pixel index list to RawCameraData container (in case not all pixels are kept)

### DIFF
--- a/ctapipe/io/containers.py
+++ b/ctapipe/io/containers.py
@@ -57,6 +57,10 @@ class RawCameraData(Container):
     Parameters
     ----------
 
+    pixel_sums : dict by channel
+        arrays of all integrated ADC pixels (n_pixels)
+    pixel_samples : dict by channel
+        arrays of non-integrated ADC sample pixels (n_pixels)
     adc_sums : dict by channel
         arrays of all integrated ADC data (n_pixels)
     adc_samples : dict by channel
@@ -67,6 +71,8 @@ class RawCameraData(Container):
     """
     def __init__(self, tel_id):
         super(RawCameraData, self).__init__("CT{:03d}".format(tel_id))
+        self.add_item('pixel_sums', dict()) ### To be determined. Maybe not necessary.
+        self.add_item('pixel_samples', dict())
         self.add_item('adc_sums', dict())
         self.add_item('adc_samples', dict())
         self.add_item('num_channels')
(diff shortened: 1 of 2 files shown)


migueln:
pixel_ refers to the pixel indices of those that have registered data about the event (might or might not be the whole camera, that's why I guess we need to keep track of them). I think it makes sense to keep both arrays into the same container as the data itself, or at least in a container at the same level. 

The reason why there are two of them (and not a common) is that as far as I know there can be 2 different sets of pixels (few pixels with samples, another group of pixels with integrated signal). Am I wrong?.

MaxNoe:
Where can I get this protozfitsreader python library? Google came up with nothing.

migueln:
It's in the CamerasToACTL svn at svn.in2p3.fr/cta

MaxNoe:
Any plans to make this open source?  This would be nice for us,  as FACT is also using zits and having a python reader might help from time to time.

migueln:
I discussed with Etienne about this. The problem is that FACT doesn't make use of protobuffers, which is what we are currently using to send the messages from the reader to the python wrapper, so the proposed solution is not that useful as it is. If FACT people wants to use this code, we would need to have a separate implementation.

MaxNoe:
I do not really understand that. How does the way the files are written affect the reading?

Yes we are currently not unsing protobuf, but I have it installed and could use it for reading zfits files.

kosack:
Can you change the pixel indices (`pixel_sums`) etc to be a boolean mask? That makes more sense for in-memory data (using a index list is good for storage, but not necessarily for fast algorithms), and then can be used with masked arrays.    I'd also suggest renaming them to `pixel_sums_mask` and `pixel_samples_mask` or something like that.

kosack:
> Can you change the pixel indices (pixel_sums) etc to be a boolean mask? 
An alternative would be to just make `adc_samples` and `adc_sums` to be `np.masked_array`s, which allow for having empty entries. Then from a code standpoint you can ignore the fact that some pixels are missing samples:

e.g.:
```python
    analyze_traces( adc_samples ) # the mask is included in the array automatically
```
instead of explicitly doing:
```python
    analyze_traces( adc_samples[pix_with_samples_mask] )
```

migueln:
I changed adc_samples and adc_sums to be numpy masked arrays.

kosack:
Thanks! Looks good. I'll merge it now

migueln:
Is it merged? I don't see the zfits.py file where it is supposed to be.

kosack:
It is now! I must have clicked close instead of merge!

PR #74, opened 2016-03-18
change in reading camera geometries from simtealarray files

moralejo:
Note: this breaks some of the examples, since class CameraGeometry has now an additional argument (camera rotation angle).  Only read_hessio_single_tel.py  works properly with these changes. Note however the calibration that is currently applied is faulty (unrelated to these changes) and images will sometimes look weird.

The changes are the following:

Now focal length is used to distinguish cameras with the same number of pixels (instead of pixel separation)

Got rid of assumption that pixels 0 and 1 are contiguous in the calculation of pixel spacing

Introduced an overall camera rotation to bring all cameras to a common system of coordinates - actually, though the cameras are no longer "rotated", images are not yet consistent among telescopes in the same location

### DIFF
--- a/ctapipe/io/camera.py
+++ b/ctapipe/io/camera.py
@@ -18,17 +18,17 @@ __all__ = ['CameraGeometry',
 
 # dictionary to convert number of pixels to camera type for use in
 # guess_camera_geometry.
-# Key = (npix, pix_seperation_m)
-# Value = (type, subtype, pixtype)
+# Key = (npix, pix_separation_m)
+# Value = (type, subtype, pixtype, pixrotation, camrotation)
 _npix_to_type = {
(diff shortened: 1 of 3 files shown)


kosack:
the problem with the calibration is in pyhessio, so that should be fixed soon (the  data are interleaved incorrectly). 

For the examples, we may just decide to remove the read_hessio example (the one with all telescopes), since that is not really a supported mode. Or we can fix it.

Note that this should also be added to the instrument module, since CameraGeometry will be deprecated soon and replaced by the equivalent in instrument.  

Let me know when this is ready to be merged.

jacquemier:
Could you confirm that calibration error  in pyhessio is result of get_calibration function ?

I have just checked and I did not find where low gain data and high gain data  are interlaced.

PR #80, opened 2016-03-23
make calibration an option in read_hessio_single_tel

kosack:
- needs update to pyehssio to fix reading the calib coeffs

### DIFF
--- a/examples/read_hessio_single_tel.py
+++ b/examples/read_hessio_single_tel.py
@@ -56,6 +56,8 @@ if __name__ == '__main__':
                         help='write images to files')
     parser.add_argument('-s', '--show-samples', action='store_true',
                         help='show time-variablity, one frame at a time')
+    parser.add_argument('', '--calibrate', action='store_true',
+                        help='apply calibration coeffs from MC')
     args = parser.parse_args()
 
     source = hessio_event_source(args.filename,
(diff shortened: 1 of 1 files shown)


jacquemier:
Could you confirm that calibration error in pyhessio is result of get_calibration function ?

I have just checked and I did not find where low gain data and high gain data are interlaced.

kosack:
it's a good question - is the correct way to read the pedestals and gains for channel 0 the following:

```python
peds = pyhessio.get_pedestal(tel_id)[0]
gains = pyhessio.get_calibration(tel_id)[0]
```

If so, there seems to be an issue.  If you plot the peds or gains that you get in that case, you end up with values that are in 2 distributions:  `plot(peds,gains)`:

pedsgains

The two gain values and two pedestal values look like what you would get if you were looking at a mixture of the low and high-gain channels.

kosack:
The effect on an image is that every other pixel looks like it is bright or dark, which implies some mixing. It looks like an indexing error (e.g. a 972x2 array is being read as a 2x972 array)

kosack:
It seems only to be an issue for prod3. In Prod2, I get the following for a single camera, which looks correct (there is a single distribution, not two):
pedsgains_prod2

So it could be that the data layout changed between the prods?

kosack:
I'll open an issue...

jacquemier:
Did you use same pyhessio version for PROD2 and PROD 3 ?
What's I know for sur, iis that there was indeed a bug in pyhessio.
I have just corrected/commited and merged.

PR #88, opened 2016-04-07
Fix CameraGeometry.rotate and test_guess_camera

MaxNoe:
* fix wrong handling of angle in CameraGeomtry.rotate

* adapt test of `CameraGeomtry.guess` to new api

PR #91, opened 2016-04-07
Add optical_foclen to geom = CameraGeometry.guess() in test_guess_cam…

jacquemier:
Warning: geom.rotate('10.0d') should work. I do not understand why is not currently working


MaxNoe:
You have to convert it to an `astropy.coordinate.Angle` before subtracting it from the `pixel_rotation`. You cannot subtract a str from a number.

See https://github.com/cta-observatory/ctapipe/pull/88.

PR #95, opened 2016-04-12
Coordinates

ParsonsRD:
Added first pass at coordinate transformations using the astropy frames code, + a simple script showing how the transformations are made for most frames. 

Currently it is rather slow and needs more documentation. Also, we need a good system in place for providing the current pointing direction (currently hard coded) of a given telescope as well as the array. Some small changes were also made to the Hillas parameter maker to perform parameterisation in the nominal system.

Not sure if this is ready for the master yet, but comments on how to get this into shape and discussion of how to pass on pointing information would be helpful.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -8,9 +8,8 @@ env:
         # The following versions are the 'default' for tests, unless
         # overidden underneath. They are defined here in order to save having
         # to repeat them for all configurations.
-        - NUMPY_VERSION=1.9
         - ASTROPY_VERSION=stable
-        - CONDA_INSTALL='conda install -c astropy-ci-extras --yes'
+        - CONDA_INSTALL='conda install --yes'
         - PIP_INSTALL='pip install'
         - INSTALL_OPTIONAL=true
     matrix:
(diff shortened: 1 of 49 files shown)


kosack:
I don't like the idea of doing any transforms inside the `hillas_parameters` routine - it should be completely generic. Hillas parameters can be computed regardless of the system.  Also, since transforms are slow if done event-by-event, we should avoid doing them in any function that is called event-wise.  

The correct way to do it would be then to read a set of events, transform them to the proper system (all at once), and then do the hillas parameterization (or else do the transform after the parameterization, which is maybe even faster - e.g. transform the parameters)

That is one of the negative "lessons learned" from HESS: coordinate transforms are done and undone many times, and they dominate the CPU when processing. That's very inefficient. We should avoid doing any transforms when possible, and only do them once they are really needed.

What should can do now is keep the `hillas_parameters` as before, and make wrapper class that does the transform and calls the hillas routine.

ParsonsRD:
Hi, agreed keeping this code within the hillas code doesn't seem like a great idea. I'm a little reluctant about transforming the Hillas parameters themselves as for some telescopes (maybe the SC based ones) we may need to perform a non linear transformation to get from the camera frame to the nominal frame, due to distortions in the FoV which will not really work with Hillas params.

Already the the code will process transformations of an N dimensional numpy array. Does this need a wrapper class? or will ultimately the "chunked" data be processed directly.

ParsonsRD:
I have made some changes to the code to remove the modification to the Hillas parameter maker and perform the Hillas reconstruction elsewhere.

PR #101, opened 2016-04-28
True charge array

watsonjj:
Commits to update compatibility with cta-observatory/pyhessio@63ec30b67e805ade7f9e0b073a24bfc2fe12506f, and is therefore stored directly into event.mc.tel[#].photo_electrons as a numpy array.

@tino-michael I understand this changes the implementation you created for photo_electrons. I figured it might make more sense to have it accessed in a similar way to event.dl0.tel[#].adc_samples. Let me know if you have issue with this.

### DIFF
--- a/ctapipe/io/containers.py
+++ b/ctapipe/io/containers.py
@@ -69,7 +69,7 @@ class MCEvent(MCShowerData):
     def __str__(self):
         return_string = super().__str__()+"\n"
         npix = np.sum([np.sum(t.photo_electrons > 0) for t in self.tel.values()])
-        return_string += "hit pixels: {}".format( npix )
+        return_string += "total photo_electrons: {}".format( npix )
         return return_string
 
 
--- a/ctapipe/io/containers.py
+++ b/ctapipe/io/containers.py
@@ -2,6 +2,7 @@
 """
 
 from ctapipe.core import Container
+import numpy as np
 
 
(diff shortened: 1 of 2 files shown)


ctapipe/io/containers.py
```diff
@@ -52,14 +52,24 @@ def __str__(self):
...
         super().__init__(name)
-        self.add_item('photo_electrons',dict())
+        self.add_item('tel',dict())
     def __str__(self):
         return_string = super().__str__()+"\n"
         NPix=0
-        for tel in self.photo_electrons.values():
-            for pix in tel.values():
+        for t in self.tel.values():
+            for pix in t.photo_electrons:
```
kosack: please follow pep8 naming conventions (don't use capitals for variables), so change NPix to `n_pix` or `npix`.

Also, avoid for-loops over pixels, which you don't need since they are in numpy arrays.
you can write this better as:
``` npix = np.sum(t.photo_electrons > 0) ```

or even get rid of the first loop too like this:
``` npix = np.sum([ np.sum(t.photo_electrons > 0) for t in tel.values]) ```

kosack:
Otherwise this looks ok, and I'll merge it once you make the small change mentioned aboce.  Of course, we need to have a review of the data container structures soon - now that we process a single telescope at a time, it may not longer be necessary to have the two-level structures where we have dicts of telescopes. That would clean this up a lot.

tino-michael:
> Also, avoid for-loops over pixels

my bad... I just quickly added this to get a feeling for the event. If there is no general interest in the total number of PE, this figure can also go completely...

PR #102, opened 2016-05-03
Add two converters:

jacquemier:
- From traitlets.config.loader.Config to FITS (in header part)
- From json file discribing traitlets configuration to FITS
Add the two corresponding unit tests.

-Do we also need a converter from FITS to JSON ?
-Do we also need a configuration system (based on traitslets) that uses FITS file header containing configuration instaled on json file ?

### DIFF
--- a/ctapipe/configuration/__init__.py
+++ b/ctapipe/configuration/__init__.py
@@ -5,3 +5,4 @@ ctapipe project.
 """
 
 from .core import *
+from .json2fits import *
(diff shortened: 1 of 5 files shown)


kosack:
I guess we should write some use cases for the configuration, to see exactly what we need to support. It's not so easy to guess at the beginning.

jacquemier:
2 use cases for the configuration already exist in redmine: 
[https://forge.in2p3.fr/projects/pipelines-use-cases/wiki].

PR #129, opened 2016-06-13
Event Reconstruction Algorithm 6 

ParsonsRD:
This is a simple reconstruction code for Werner's paper:
http://arxiv.org/pdf/astro-ph/9904234v1.pdf

I implemented this as a basis for future minimisation based reconstruction methods.

Again this is not a finalised version of the code, it had not been thoroughly tested (but does not give crazy results). Could people take a quick look at the formulation of this class to make sure this makes sense. Particularly with regards to the input parameters for the reconstruction function. Ideally for speed I would like to give a 2D numpy array containing for example the pixel x positions for all cameras, however in this case the arrays would be different lengths, this introduces a look in the code.

Is there a way round this? Could we use an array the size of the largest camera?

### DIFF
--- a/ctapipe/reco/weighted_axis_minimisation.py
+++ b/ctapipe/reco/weighted_axis_minimisation.py
@@ -13,7 +13,7 @@ __all__ = [
 ]
 
 
-def reconstruct_event(hillas_parameters,telescope_positions,shower_seed=None):
+def reconstruct_event(hillas_parameters,telescope_positions,pixel_list,shower_seed=None):
     """
     Perform event reconstruction
 
@@ -23,6 +23,8 @@ def reconstruct_event(hillas_parameters,telescope_positions,shower_seed=None):
         Hillas parameter objects
     telescope_positions: list
         XY positions of telescopes (ground system)
+    shower_seed: shower object?
+        Seed position to begin shower minimisation
 
     Returns
     -------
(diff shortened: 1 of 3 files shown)


ctapipe/reco/weighted_axis_minimisation.py
```diff
@@ -0,0 +1,154 @@
...
+"""
+import numpy as np
+import astropy.units as u
+from iminuit import Minuit
+__all__ = [
+    'weighted_axis_minimisation'
+]
+
+
+class weighted_axis_minimisation:
```
kosack: use CapWords for class names (so `WeightedAxisMinimization`)following the PEP8 standards.

Also, coming soon we will have a nice Configurable base class that can be a parent, and allows for user-parameters to be  added to such a class that automatically can become command-line or conf-file parameters, so this is a good start.

ParsonsRD:
Argh, my IDE even told me this did not conform to PEP8, fixed now.

ctapipe/reco/weighted_axis_minimisation.py
```diff
@@ -0,0 +1,296 @@
...
+            Test core position in tilted system
+        y_grd: float
+            Test core position in tilted system
+
+        Returns
+        -------
+            Sum of weighted pixel distances from predicted axis
+        """
+
+        sum = 0
```
MaxNoe: You are overwriting the python built-in function `sum` here.You shouldn't do this. You could use `sum_` or `weighted_average`

ctapipe/reco/weighted_axis_minimisation.py
```diff
@@ -0,0 +1,296 @@
...
+        for pos_x,pos_y,tel_x,tel_y,weight in zip(self.pixel_pos_x,self.pixel_pos_y,self.tel_pos_x,self.tel_pos_y,self.pixel_weight):
+            phi = np.arctan2((tel_y-y_grd),(tel_x-x_grd)) * u.rad
+            sum += self.get_dist_from_axis(pos_x,pos_y,x_src,y_src,phi,weight)
```
MaxNoe: As far as I see it, this complete loop could be replaced by a single `np.sum` if `pixel_pos_*` were a 2d array and not a list.

PR #132, opened 2016-06-15
Shower container class

ParsonsRD:
Created simple shower container class for holding reconstructed shower information. This object will probably need to by used by several reconstruction techniques so is created in a separate file.


tino-michael:
There is already a MCShowerData class in io/containers.py ...
I would rather suggest that the reco shower inherits from the mc shower (or a more general shower class) and adds needed members -- like uncertainties, xmax etc. -- to keep the interface and member access consistent...

ParsonsRD:
Yes, you're right it would make much more sense to inherit from this class (ultimately the 2 classes may want to inherit from a single common "shower" class though). I'll try preparing something...

Issue #89, opened 2016-04-07
Circular import around ctapipe.core.Container


MaxNoe:
### `ctapipe.io.__init__`:
```
from .containers import *
```
### `ctapipe.io.containers`:
```
from ctapipe.core import Container
```

### `ctapipe.core.__init__`

```
from ctapipe.io.files import get_file_type
```

kosack:
Is there a suggestion how to resolve this? Originally `io` only depended on `core` and not the other way around, but now there is a loop.

The problem seems only to be that in `core.container` the `write()` method uses `io.get_file_type()`. Either we need to separate the writing of containers (delegating any writing to some back-end written in io), or move `get_file_type()` to `core`...

PR #144, opened 2016-07-08
parallelization system

jacquemier:
online documentation
http://jacquemier.github.io/ctapipe/pipeline/index.html


kosack:
There doesn't seem to be a python3 version of pyside in anaconda. At least I get:
```
> conda search pyside
pyside                       1.1.1                    py27_0  defaults        
                             1.1.2                    py27_0  defaults        
                             1.1.2                    py27_1  defaults      
```
So I cannot install it.  However, I think it has the same interface/API as `pyqt`, which is there.  So you might do something like:

```python
try:
    from pyside import  X,Y,Z
except ImportError:
    from pyqt import X,Y,Z
```

ctapipe/pipeline/algorithms/hessio_reader.py
```diff
@@ -0,0 +1,35 @@
+from ctapipe.utils.datasets import get_path
+from ctapipe.io.hessio import hessio_event_source
+from ctapipe.configuration.core import Configuration, ConfigurationException
+import threading
+from ctapipe.core import Component
+from traitlets import Unicode
+
+
+class HessioReader(Component):
```
kosack: I'd rename this `SimTelArrayReader` (we should start to rename things to avoid the HESS name)

ctapipe/pipeline/algorithms/list_teldata.py
```diff
@@ -0,0 +1,17 @@
+from time import sleep
+import threading
+from ctapipe.core import Component
+
+
+class ListTelda(Component):
```
kosack: rename this (doesn't follow naming conventions of using full words and capitalize each word, CapWordsStyle).  Maybe `ListTelsWithData`  or something like that.

ctapipe/pipeline/algorithms/prun2ptabhillas.py
```diff
@@ -0,0 +1,37 @@
+from time import sleep
+import threading
+import subprocess
+import os
+from ctapipe.core import Component
+from traitlets import Unicode
+from ctapipe.pipeline.algorithms.base_process import build_command
+
+
+class Prun2PtabHillas(Component):
```
kosack: rename this class (no idea what it does from the name): 
* don't use abbreviations
*  use consistent capitalization
* Add a `description` field.

ctapipe/pipeline/algorithms/reco.py
```diff
@@ -0,0 +1,37 @@
+from time import sleep
+import threading
+import subprocess
+import os
+from ctapipe.core import Component
+from traitlets import Unicode
+from ctapipe.pipeline.algorithms.base_process import build_command
+
+
+class Reco(Component):
```
kosack: rename to `Reconstruct`

ctapipe/pipeline/zmqpipe/consumer_zmq.py
```diff
@@ -0,0 +1,112 @@
+# Licensed under a 3-clause BSD style license - see LICENSE.rst
+# coding: utf8
+
+
+import zmq
+from threading import Thread, Lock
+import pickle
+
+
+class ConsumerZmq(Thread):
```
kosack: rename: `ConsumerZMQ`  (acronyms in all caps)

ctapipe/pipeline/zmqpipe/pipeline_zmq.py
```diff
@@ -0,0 +1,615 @@
...
+        for index, step in enumerate(self.stagers_conf):
+            if step['name'] == name:
+                return index
+        return -1
+
+def main():
+    tool = Pipeline()
+    tool.run()
+
+if __name__ == 'main':
```
kosack: should be `if __name__ == '__main__'`

kosack:
I'm still unable to install pyside, even with pipe (it does not support python 3.5, which is what anaconda currently provides by default).  I'd prefer not to add a dependency that doesn't have a maintained conda package, if possible, but I do think the pyside license is better than the Qt one (I'm not sure why anaconda dropped it).  Anyhow, perhaps we can see if we can get it to  work with either one, since they have only minor differences in the API.

MaxNoe:
Why PySide and not PyQt4? Anaconda went for PyQt4. I do not remember the reasons, but the pyqt4 package is actively maintained by anaconda.

jacquemier:
Change PySide to PyQt4 (same API)
Rename file, class and method to follow naming conventions

ctapipe/pipeline/algorithms/simtelarray_reader.py
```diff
@@ -0,0 +1,37 @@
...
+    filename = Unicode('gamma_test.simtel.gz', help='simtel MC input file').tag(
+        config=True, allow_none=True)
+    source = None
+
+    def init(self):
+        self.log.info("--- HessioReader init ---")
+        try:
+            in_file = get_path(self.filename)
+            self.source = hessio_event_source(in_file, max_events=10)
+        except():
```
MaxNoe: except is not a function.

jacquemier:
autopep8 and change "except" following max recommendation

kosack:
Thanks, I will merge this in now. I'm still having some problems running parts of it, but that seems to be due to  issue #89  introduced in a previous pull-request, so is independent of this.

PR #154, opened 2016-08-04
Improvements to calibration/charge integration methods, and introduction of input file handling and basic plotting

watsonjj:
As well as improvements to calibration/charge integration methods, this pull request adds two main features:

InputFile - A class inside io/files.py that handles the reading of any data file, using a switch to find the correct io generator to fill the event container. Also delivers common features like extracting a particular event by either event_index or event_id.

plotting - A new module to hold common plots to avoid duplication of code, and ensure consistency. Currently contains the module camera.py, which contains methods to plot camera and waveform images, and also annotate them.

I have also created a new example: display_integrator. This can be used to show how an integrator/calibration performs, and is also a demonstration of how to use the above two features.

### DIFF
--- a/ctapipe/calib/camera/__init__.py
+++ b/ctapipe/calib/camera/__init__.py
@@ -2,8 +2,42 @@
 """
 Camera calibration module.
 
-To calibrate an event look at calibrators.py.
+--------------
+calibrators.py
+--------------
+This module selects the correct calibrator based on the
+origin of the file. For example, it will use the calibrator in mc.py if the
+file’s origin is hessio. It then calibrates every telescope in that event,
+and stores the pe_charge in the new dl1 container of the event (along with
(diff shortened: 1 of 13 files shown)


kosack:
Looks good! It even satisfies one of the goals I had for the file handling (to have a generic factory class that opens any stream type, not just files).   Therefore, it may be better to rename `InputFile` to something even more generic and have it support toy-monte-carlo sources like the mock generator as well, since then it works for all test cases, etc.  Perhaps `EventSource` is a better name?  Anyhow, I'll accept it for now and take a look on how to make it more general.

kosack:
One question I had: is there a reason that you had to create a new module (`plotting`) rather than put things in `visualization`?  I would rather not add to many new modules, with confusingly similar names.

PR #161, opened 2016-09-14
Before pull request

jacquemier:
A flow-based framework.
In flow-based programming, applications are defined as networks of black-box components that exchange data across predefined connections. These components can be reconnected to form different applications.
ctapipe-flow executes ctapipe processing modules in a sequantial or multiprocessus environment.

Documentation here: http://jacquemier.github.io/ctapipe/flow/index.html

2 examples provided: 
 - complex  flow-based with fork : examples/flow/switch.json
 - calibration (based on examples/calibration_pipeline.py  ) examples/flow/calibration.json
Warning: calibration create PDF wtih matplotlib figure. multiprocessus mode only work on Linux. 
OsX isssue : "__the_process_has_forked_and_you_cannot_use_this_corefoundation_functionality___you_must_exec__()"


kosack:
Is the forking problem something internal to the `multiprocessing` module, or do you call `fork()` explicitly?  

It probably won't make a difference, but `multiprocessing` is slowly being replaced by the `concurrent.futures` module, which provides a more common API for multi-process/threading/etc.  I would imagine the implementation is pretty much the same though.

jacquemier:
The forking problem is  internal to the multiprocessing. I do not call fork() explicitly,

kosack:
It seems there is a pull request in matplotlib to fix this, so it may not be an issue soon:

https://github.com/matplotlib/matplotlib/pull/6637

kosack:
in any case, since this doesn't conflict with other modules, I will merge it in so people can play with it at the next workshop.

jacquemier:
I currently study python Dask framework. http://dask.pydata.org
I will try to prepare a comparative with my own development version for the next workshop.

PR #172, opened 2016-09-30
Pixel highlighting

MaxNoe:
* Method for the `CameraDisplay` to highlight pixels, which can either be indices or a boolean mask.

* Example script

camera

### DIFF
--- a/ctapipe/visualization/mpl.py
+++ b/ctapipe/visualization/mpl.py
@@ -130,6 +130,11 @@ class CameraDisplay:
         self.pixels = PatchCollection(patches, cmap=cmap, linewidth=0)
         self.axes.add_collection(self.pixels)
 
+        self.pixel_highlighting = copy.copy(self.pixels)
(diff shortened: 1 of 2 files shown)


kosack:
Looks good, and definitely useful extension of the original single-highlighted pixel functionality.  

One question: What happens when the image changes? (via `disp.image = newimage`) Is there a method to clear the highlighted pixels? What is the procedure a user should follow to update it?

It occurs to me that we should also make a camera display (subclass or something) to support the old Whipple-style camera display, where instead of colors, the size of the filled area inside a pixel is used to represent intensity. The advantage of that is that multiple colors can be used to mark pixels (e.g. showing intensity + selected pixels, broken pixels, etc, all on the same plot without it being over-complicated).  That can be a future enhancement though.

MaxNoe:
You can call `disp.highlight_pixels` with an empty list to remove the highlighting.

PR #176, opened 2016-10-05
Calibration arguments

watsonjj:
Update to the calibration argument parsing, making it more flexible so individual cameras can specify their own arguments they require for calibration.

### DIFF
--- a/ctapipe/calib/camera/calibrators.py
+++ b/ctapipe/calib/camera/calibrators.py
@@ -169,15 +169,15 @@ def calibrate_event(event, params, geom_dict=None):
                           event.meta.optical_foclen[telid])
         # Check if geom is even needed for integrator
         if inverted[params['integrator']] in integrators_requiring_geom():
-            if geom_dict is not None and cam_dimensions in geom_dict:
-                geom = geom_dict[cam_dimensions]
+            if geom_dict is not None and telid in geom_dict:
+                geom = geom_dict[telid]
             else:
                 log.debug("[calib] Guessing camera geometry")
                 geom = CameraGeometry.guess(*event.meta.pixel_pos[telid],
                                             event.meta.optical_foclen[telid])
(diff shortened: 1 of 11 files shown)


kosack:
For options and arguments, please use the `ctapipe.core.Container` and `ctapipe.core.Tool` functionality and not `argparse` or a custom parser.  That way all modules will work the same and there is a single configuration format.

kosack:
Another major problem is that the configuration parameters should *not* be passed around as dicts, since those cannot be validated or checked, and that can lead to unexpected behavior.  This is confusing to users and hard to debug.

HEre is an example of how it should work: rather than having a master function like calibrate event that takes all parameters, break it down in to specific tasks. For instance trace integration should have a class hierarchy with all intergrator types (`Integrator`, `SimpleWindowIntegrator`, etc), adn those should handle integration parameters like the window size, etc.

Any options to a function must be explicit named parameters with defaults if they are known:
```python
def calibrate_event(event, integrator=DefaultIntegrator(window=5)):
    ...
```
Use a `Component` as the base class for any algorthms that require several functions that share parameters:

```python
from traitlets import Integer
class WindowIntegrator(Component):
     window_size = Integer(default_value=6, help='integration window size in samples')\
                                          .tag(config=True)

    def integrate(samples):
         ...
```

See the `ctapipe.core.Container` docs for more examples (for some reason the example codes doesn't appear in the sphinx output, but it's in the code itself - I'll fix that soon)

kosack:
I've created an issue #179 to track this, since it's  a major change, but I'll probably just accept this pull request for now, and leave the refactoring for later.

watsonjj:
Hi Karl. Thanks, I was hoping you would be okay with that.

I mostly follow what you have said, and I would like to implement it. This is the first time I've looked into the application of `ctapipe.core.Container` and `ctapipe.core.Tool`, and it will take a little time to understand how to develop the calibration to that level.

PR #181, opened 2016-10-08
YAML configuration

mireianievas:
- Small script to test the ASCII -> YAML conversion (or a possible one).
- Begin to implement a YAML reader in InstrumentDescription (still needs some work, see the attachment for some example).

[array_config_proto_v00.yaml.txt](https://github.com/cta-observatory/ctapipe/files/517809/array_config_proto_v00.yaml.txt)


ctapipe/instrument/InstrumentDescription.py
```diff
@@ -905,6 +907,222 @@ def load_config(filename):
...
+    tel_num = 10
+    tel_id = [random.randrange(1,124,1) for x in range(0,tel_num)]
+    
+    telescope = {}
+    camera    = {}
+    optics    = {}
+    auxiliary = {}
+    
+    with open(filename, 'r') as fin:
+        yamlobject = yaml.load(fin)
```
MaxNoe: One always should use `yaml.safe_load` if the addtional parsing features of `yaml.load` are not needed.
See https://youtu.be/kjZHjvrAS74?t=6m50s.

ctapipe/instrument/InstrumentDescription.py
```diff
@@ -905,6 +907,222 @@ def load_config(filename):
...
+
+    for item in yamlobject:
+        if '_' not in item:
+            itemname = item
+        else:
+            itemname = '_'.join(item.split("_")[1:])) # Telescope_Type_1 becomes Type_1
+    
+
+        # If organized in Header/Data ~ FITS drop-replacement.
+        if ('TelescopeType' or 'TelType') in item:
```
MaxNoe: This does not work like I think you intendet.

`'TelescopeType' or 'TelType'` will eveluate to `'TelescopeType'`, thus only '`TelescopType'` will be checked. I think you meant `'TelescopType' in item or 'TelType' in item`

ctapipe/instrument/InstrumentDescription.py
```diff
@@ -905,6 +907,222 @@ def load_config(filename):
...
+    tel_table_prime['MirA'].unit = u.m**2
+    mirror_num = [random.randrange(1,124,1) for x in range(tel_num)]
+    tel_table_prime['MirN'] = mirror_num
+    foclen = [random.uniform(1,100) for x in range(tel_num)]
+    tel_table_prime['FL'] = foclen
+    tel_table_prime['FL'].unit = u.m
+    
+    telescope['TelescopeTable_Version%s' % version] = tel_table_prime    
+    
+    for t in range(len(tel_id)):
```
MaxNoe: This is a python anti-pattern. Better is `for tel_id in tel_ids`. If you also need the index, use `for t, tel_id in enumerate(tel_ids)`

MaxNoe:
@mireianievas  I think having yaml is a good idea, as it's an easy to write and widely used file format :+1: 
I made some comments, you might want to check out.

ctapipe/instrument/InstrumentDescription.py
```diff
@@ -905,6 +907,222 @@ def load_config(filename):
...
+    try:
+        entry=item[entry]
+    except:
+        entry = None
+    
+
+    # Case A) no variable, no item, override the whole table
+
+    if newvalue==None:
+        return(table)
```
mireianievas: Return became a function in Python 3, the same applies to python. Somehow became also possible to use them as functions in recent version of the Python 2.X branch and I adopted it to try to make the code as compliant as possible with 3.X. But it's fine to go back to the old 2.X behavior, I will change it.

ctapipe/instrument/InstrumentDescription.py
mireianievas: Ooops, you're right here.

ctapipe/instrument/InstrumentDescription.py
MaxNoe: `xrange` became `range` in python 3. If you want the old python 2 range, the python 3 equivalent would be `list(range(start, stop, stepsize))`

MaxNoe:
`xrange` is python2 and does not exist in python 3

kosack:
You seem to have included an update/downgrade to the astropy_helpers submodule (probably by accident). Can you remove that from your commit?

Issue #183, opened 2016-10-10
Astropy dependence in setup.py


mackaiver:
Hi,

I'm confused about the need for all the astropy things in the setup.py.
Is this supposed to be like an official astropy plugin or something similar? 
Maybe we can simplify the setup.py. Might also help with travis builds. I don't know.

PR #184, opened 2016-10-10
Cleanup setup.py, remove ah_bootstrap and astropy helpers

MaxNoe:
This pull request is intedend as a start of a possible discussion about why this setup.py got so complicated.

The test fail, which is connected to the version, which has no githash and release tags anymore.

Also see #183


kosack:
It got complicated only because it started that way (with the astropy affiliated package template).

We use a few nice features of the astropy_helpers package (like the auto version naming, and better documentation plugins), but we could easily separate out those parts to simplify things.

things like:

```
% ctapipe-info --version

*** ctapipe version info ***

version: 0.1.dev1038
release: False
githash: 5b7d1e8c57c3320c24b29948c77896e7082e66f4

```

or 

```
python setup.py bdist-conda
```

both use `astropy_helpers` functionality

mackaiver:
I think this is covered by #294 and #295

PR #186, opened 2016-10-10
reverted to previous behavior of hessio_event_source

kosack:
- reverted to old behavior where the return value is a pointer to a
Container, and each subsequent access only  updates data in the
container, without creating a new one (this was intentionally for speed
and memory use)
- the "new event" behavior can be achieved by simply wrapping with a
deepcopy version of the generator `uniquesource = (deepcopy(event) for event in source)`
- Created a top-level EventContainer that sets up all items (no more adding items in the event source code, which is not easy to enforce)


MaxNoe:
Did you actually test how much performance we gain through this and if it is worth the side effects (e.g. havin dl1 data from the previous event in a container?)

PR #193, opened 2016-10-12
higher order moments in hillas parameter version 2

wrijupan:
hillas_parameters_1 and hillas_parameters_2 are now compatible in units


MaxNoe:
the hillas parameters got moved. You need to 
```
git fetch upstream
git rebase upstream/master
```

To resolve the conflicts.

I think this is needed quite urgend, because the missing units is one of the last reasons the docs fail to build (issue warnings).

ctapipe-extra
```diff
@@ -1 +1 @@
-Subproject commit f7fe2209646dc8f99a28ec9ab4df8257687e1896
+Subproject commit 536490d27e34c5f3b0c361f828de208f7e17b60f
```
MaxNoe: You downgraded ctapipe-extra here. I think unintentionally.

kosack: yeah, that's a common problem. It will be better when ctapipe-extra is it's own package (we just need to migrate the code in ctapipe that uses it to call the correct function)

ctapipe/image/hillas.py
```diff
@@ -146,11 +127,11 @@ def hillas_parameters_1(pix_x, pix_y, image):
     a = d2 / d1
     # Angle between ellipse major ax. and x-axis of camera. Will be used for
     # disp
-    delta = np.pi / 2.0 + np.arctan(a)
+    delta = ((np.pi / 2.0) + np.arctan(a)) * u.rad
```
wrijupan: The motivation to introduce units to delta is because it is returned in degrees as psi. But it is true that this just slows down the computation. To make more sense, I have just now made a new commit where the entire code uses the angle values in radians and at the return line deals with all the units and unit conversions. 

And also, I realized the name 'delta' might create confusion. Hence I renamed it to psi, since it is returned as psi in the output.

ctapipe-extra
wrijupan: git pull upstream master does not seem to update ctapipe-extra. What is the correct way to do it?

MaxNoe: You should also avoid doing 
```
git add .
```

It is less error-prone to just add the files you want to commit explicitly.
Also helpful is `git status --short`, which only shows an overview over the files.

MaxNoe:
Now, something ist seriously broken. I do not know how. But you added the submodules themselves to the repository.

MaxNoe:
The problem is the second to last commit: 87396e8

I think it is best do remove it completely from the history. 

What you could do (but which is kind of dangerous and you should back up your local copy of the repo before) is:

```
git reset --hard 8e547f8  # go back to the last commit before the strange stuff happend
git push -f
```

wrijupan:
I removed the problematic commits! Now I do the following -
`git pull upstream master # upstream is alias to the cta-observatory repo
 git submodule update
 git push -u origin master  #master is alias to my github repo`

But ctapipe-extra still seems to be the old version.

MaxNoe:
Thats because you commited the change in submodule and there was no conflict with the master, so git kept your change.

You have to update back to the version it was before.

So:
```
cd ctapipe-extra
git checkout f7fe2209646dc8f99a28ec9ab4df8257687e1896
cd ..
git add ctapipe-extra
git commit -m 'Revert ctapipe-extra downgrade'
git submodule update
```
should do the trick

MaxNoe:
You again pushed the complete history of the submodules, I do not know why this happens. Can you describe what you are doing and in which directory?

wrijupan:
The directory structure is cta_dir/ctapipe. Inside this I do `git init`. There are two remotes - origin (my github repo) , master (cta-observatory repo). cta_dir/ctapipe contains many directories including ctapipe, ctapipe-extra, astropy-helpers. In cta_dir/ctapipe I do - `cd ctapipe-extra

git checkout f7fe2209646dc8f99a28ec9ab4df8257687e1896
cd ..
git add ctapipe-extra
git commit -m 'Revert ctapipe-extra downgrade'
git submodule update`

After this from my local copy cta_dir/ctapipe/ctapipe everything gets deleted with just an empty 'image' folder.
And inside cta_dir/ctapipe two new directories ctapipe-extra and datasets gets created!

MaxNoe:
Mmmh. This sounds strange.

Ok, let's fix this strange commit first.

For this, go to the `cta_dir/ctapipe` directory. The one that contains the `setup.py` and the  `ctapipe-extra` and `astropy_helpers` directories.

Here we remove the last two commits by going back to the commit with the hash 8e547f864c861e7c259abcd11aad10fa34ad83fc:

```
git reset --hard 8e547f864c861e7c259abcd11aad10fa34ad83fc
```

After that, you need to do
```
git push -f
```
The `-f` is required because you delete commits from the online repository.

wrijupan:
The problematic commit removed. Now I just start from the scratch. Now if I do a clone of my git hub repo, the directories ctapipe-extra and astropy_helpers are empty in my local copy! 
Although I can see them in the online github repo.

MaxNoe:
Yes. Now you can do 
```
git submodule init
git submodule update
```

The repositories are filled then.

wrijupan:
Yes! Now `git push` ?

MaxNoe:
There should be nothing you can push right now. We only fully cloned the repo with all submodules.

Now we need to update the ctapipe-extra version to the one in master.

ah, and i think i found the mistake.

Could you do

```
cd ctapipe-extra
git checkout f7fe2209646dc8f99a28ec9ab4df8257687e1896
cd ..
```
And tell me the output of `git status`?

wrijupan:
On branch master
Your branch is up-to-date with 'origin/master'.

Changes not staged for commit:
  (use "git add <file>..." to update what will be committed)
  (use "git checkout -- <file>..." to discard changes in working directory)

	modified:   ctapipe-extra (new commits)

no changes added to commit (use "git add" and/or "git commit -a")

wrijupan:
Now ctapipe-extra looks updated in the local copy.

MaxNoe:
And `git diff`?

wrijupan:
diff --git a/ctapipe-extra b/ctapipe-extra
index 536490d..f7fe220 160000
--- a/ctapipe-extra
+++ b/ctapipe-extra
@@ -1 +1 @@
-Subproject commit 536490d27e34c5f3b0c361f828de208f7e17b60f
+Subproject commit f7fe2209646dc8f99a28ec9ab4df8257687e1896

MaxNoe:
OK, that looks good.

If you now do `git add ctapipe-extra` what does `git diff --staged` say?

wrijupan:
After `git add ctapipe-extra`, `git diff --staged` says -

diff --git a/ctapipe-extra b/ctapipe-extra
index 536490d..f7fe220 160000
--- a/ctapipe-extra
+++ b/ctapipe-extra
@@ -1 +1 @@
-Subproject commit 536490d27e34c5f3b0c361f828de208f7e17b60f
+Subproject commit f7fe2209646dc8f99a28ec9ab4df8257687e1896

MaxNoe:
Great, I think we got it. 

`git commit` and `git push`

wrijupan:
Yes!!! Phew! Thank you very much.

MaxNoe:
Your changes to the `hillas_parameter` function broke the camera_display example and probably also a few others. This is why the documentations fails to build. Can you adapt the examples?

wrijupan:
Now the hillas_parameters code returns two things - MomentParameters and HighOrderMomentParameters (two namedtuples). But the camera_display example now works with only one return value and hence cannot recognize two returns. 

How to resolve this? Should a single namedtuple be used instead of two?

The interactive_test_hillas.py code in ctapipe/image/tests handles the two return values correctly for now.

MaxNoe:
I think both versions of the `hillas_parameters` should return a single `Container`, probably a new `HillasContainer`, if I understood the current return value concept correctly.

@kosack

wrijupan:
I just added a preliminary version of `HillasContainer` for demo of storing the `hillas_parameters` along with an interactive test script. A single Container is used to store the `MomentParameters` and `HighOrderMomentParameters`.

wrijupan:
Okay! Now I have committed changes to use a single `HillasContainer` holding all the shower moments as items. The camera_display example should work now.

MaxNoe:
Container looks good. However, you downgraded the astropy helpers. Just never add changes in astropy_helpers or ctapipe_extra if it's not your intention do upgrade or downgrade them. Ypu should revert 	80eaa20:

```
git revert 	80eaa20
git push
```

MaxNoe:
There are some markup errors in the doc strings. I would also remove the change log in the file. This is what we have git for.

wrijupan:
I fixed some of the doc string errors that I could identify manually, but `make doc` did not show any doc string warnings/errors for `hillas.py`.

kosack:
> I think both versions of the hillas_parameters should return a single Container, probably a new  HillasContainer, if I understood the current return value concept correctly.

The idea of the Containers is that they are just table definitions for output files, so when we do a hillas analysis, they should include anything that we want to be written to the output table (so all Hillas parameters in this case).  I'd say the higher-order moments are always useful, so there is no reason to separate them into two.

ctapipe/image/hillas.py
```diff
@@ -186,27 +143,20 @@ def hillas_parameters_1(pix_x, pix_y, image):
...
+    HillasContainer.add_item("length", length * unit)
+    HillasContainer.add_item("width", width * unit)
+    HillasContainer.add_item("dist", r * unit)
+    HillasContainer.add_item("phi", (phi * u.rad).to(u.deg))
+    HillasContainer.add_item("psi", (psi * u.rad).to(u.deg))
+    HillasContainer.add_item("miss", miss * unit)
+    HillasContainer.add_item("Skewness", skewness)
+    HillasContainer.add_item("Kurtosis", kurtosis)
+    HillasContainer.add_item("Asymmetry", asym)
+    return HillasContainer
```
kosack: This is not how containers work - a HillasContainer class should be defined in io/containers.py  (also you will need to pull in the latest changes, since the Container definition no longer uses `add_item`. Otherwise it looks fine.

kosack:
I think this is all super-ceded by other commits. All hillas algorithms (4 of them!) seem to be in agreement and have common high-order outputs.

Issue #188, opened 2016-10-11
Configuration module uses status codes


MaxNoe:
The configuration modules return status codes. 

They should raise exceptions instead.

jacquemier:
Which configuration module ?
ctapipe.util.json2fits ?

MaxNoe:
Yes.

MaxNoe:
They do stuff like

```{python}
try:
    stuff()
except IOError:
    return False
return True
```

I do not think this is a good idea. Also for logging exceptions, there is `logging.exception` which includes the traceback.

```{python}
try:
    stuff()
except:
    logging.exception('Could not do stuff')
    raise
```
Would log the exception including traceback and the raise the same exception. So only logging is added.

jacquemier:
https://github.com/cta-observatory/ctapipe/pull/196

PR #196, opened 2016-10-13
remove return status code and replace it by exception raising

jacquemier:
refer to  issue https://github.com/cta-observatory/ctapipe/issues/188

### DIFF
--- a/ctapipe/utils/tests/test_json_2_fits.py
+++ b/ctapipe/utils/tests/test_json_2_fits.py
@@ -123,10 +123,10 @@ def test_traitlets_config_to_fits():
     app = MyApp()
     app.initialize()
     app.start()
+    
     app.traitlets_config_to_fits()
-
     sys.argv = backup
-    s
+
 def test_jsonToFits():
     backup = sys.argv
     full_config_name = get_path('config.json')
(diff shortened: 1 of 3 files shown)


ctapipe/utils/tests/test_json_2_fits.py
```diff
@@ -123,7 +123,10 @@ def test_traitlets_config_to_fits():
     app = MyApp()
     app.initialize()
     app.start()
-    assert(app.traitlets_config_to_fits() == True)
+    try:
+        app.traitlets_config_to_fits()
+    except:
+        assert()
```
MaxNoe: You do not need this try except blog. If it raises, it fails.

ctapipe/utils/tests/test_json_2_fits.py
```diff
@@ -133,7 +136,10 @@ def test_jsonToFits():
     app = MyApp()
     app.initialize()
     app.start()
-    assert(app.jsonToFits() == True)
+    try:
+        app.jsonToFits()
+    except:
+        assert()
```
MaxNoe: Same here.

ctapipe/utils/tests/test_json_2_fits.py
MaxNoe: `assert()` means the same as `assert tuple()` Assert is not a function. You test if the empty tuple evaluates to True.

MaxNoe:
I think this should be merged.

Issue #47, opened 2015-12-02
restructure the modules to reflect the code better


kosack:
The module names are currently quite confusing and should be restructured. One suggestion:

* move `CameraGeometry` from `io` to `instrument` (it's really a instrumental data model, and `io` should be purely for data access). 
* create an `image` module that handles all image-level routines that are currently in `reco` 
  * hillas parameterization , image cleaning, feature extraction, etc, should all be in `image`
  * shower reconstruction methods should be in `reco`

PR #200, opened 2016-10-13
Moved all hillas and image cleaning algos to `ctapipe.image`

kosack:
the `reco` module is from now on only for shower reconstruction, and all Cherenkov image (datacube) processing is in the `image` module


kosack:
fixes #47

PR #205, opened 2016-10-13
added unit test for ShowerMaxEstimator

tino-michael:


ctapipe/reco/shower_max.py
```diff
@@ -31,9 +33,9 @@ def __init__(self, filename, col_altitude=0, col_thickness=2):
         
         self.atmosphere = Histogram(axisNames=["altitude"])
         self.atmosphere.hist = thickness*u.g * u.cm**-2
-        self.atmosphere.bin_lower_edges = [np.array(altitude)*u.km]
+        self.atmosphere._binLowerEdges = [np.array(altitude)*u.km]
```
MaxNoe: here, you add the underscore

ctapipe/reco/shower_max.py
```diff
@@ -77,16 +79,18 @@ def find_shower_max_height(self,energy,h_first_int,gamma_alt):
...
         t_shower_max = t_first_int + c
         
         # now find the height with the wanted thickness
         for ii, thick1 in enumerate(self.atmosphere.hist):
             if t_shower_max > thick1:
+                
                 height1 = self.atmosphere.bin_lower_edges[0][ii]
                 height2 = self.atmosphere.bin_lower_edges[0][ii-1]
-                thick2  = self.atmosphere.get_value([height2.to(self.atmosphere._binLowerEdges[0].unit).value])[0]
+                thick2  = self.atmosphere.get_value([height2.to(self.atmosphere.bin_lower_edges[0].unit).value])[0]
```
MaxNoe: but here, you remove the underscore.

ctapipe/reco/shower_max.py
MaxNoe: also camel case seems to be mixed up.

ctapipe/reco/shower_max.py
tino-michael: _binLowerEdges is the data member of the histogram class (to set up the edges, which is actually not meant to be supported like this, hence the underscore)
bin_lower_edges is a @property and the one that should be used as soon as the binning is properly defined.

MaxNoe: Ah, ok. But than it probably should be `_bin_lower_edges` (No camel case).

tino-michael: This is the way it's implemented in the histogram class... I'm just using it here.
I needs to be changed there as well, @kosack ?

tino-michael:
requested changes implemented

kosack:
Still have an error in Travis: 

```
FileNotFoundError: [Errno 2] No such file or directory: '/home/travis/build/cta-observatory/ctapipe/ctapipe-extra/datasets/atmprof_paranal.dat'
```

Which I guess means the ctapipe-extra pointer needs to be updated as well to a new version of that repo?

tino-michael:
@MaxNoe ?

PR #208, opened 2016-10-13
Shower reco

tino-michael:


ctapipe/reco/tests/test_FitGammaHillas.py
```diff
@@ -0,0 +1,62 @@
...
+            mask = tailcuts_clean(tel_geom[tel_id], pmt_signal, 1,picture_thresh=10.,boundary_thresh=5.)
+            if True in mask:
+                pmt_signal[mask==False] = 0
+                        
+            try:
+                moments, moms2 = hillas_parameters(fit.cameras(tel_id)['PixX'],
+                                                   fit.cameras(tel_id)['PixY'],
+                                                   pmt_signal)
+                Hillas_Dict[tel_id] = moments
+            except Exception as e :
```
MaxNoe: Should we really catch every exception in a test?

tino-michael: my idea was that an exception in the hillas parametrisation should show up in the respective test and not in the test for the direction fit...

TarekHC:
Hi @tino-michael ,

I think we should do the pull request between shower_reco and master soon. I updated the shower_reco branch with master, and there were some changes, mainly the cleaning and hillas parameterization were moved to the "image" folder.

Could you update your repo with these changes? As soon as you do we will merge it with the cta-obs/shower_reco branch and create the new pull request (and there we may continue discussion). But if we do not do it now, it will get complicated later...

Thanks!

tino-michael:
@TarekHC I changed the path for the hillas parametrisation and the tailcut cleaning in the test function. The reconstruction class is independent of both anyway.

kosack:
make sure you pull in changes from the upstream/master branch as well (the image module now contains the image cleaning and Hillas parameterization, which might require some modification on your part)

For the exceptions in tests, you should define exceptions for each possible failure mode (or class of them). See for example `image.hillas.HillasParameterizationError`, and only catch the ones you are expecting. 

You can test that they are thrown correctly too using for example the following in your tests 

 ```python

from pytest import raises
...

def mytest():
    with raises(ZeroDivisionError):
        1/0
```
That will cause the test to fail if the exception *is not* raised

TarekHC:
Hi @tino-michael , @kosack , @MaxNoe 

I will open now the new pull request, between `cta-observatory:shower_reco` -> `cta-observatory:master`. Discussion and development should continue there (following the scheme that we discussed during the f2f meeting).

PR #212, opened 2016-10-13
First version of serializer

elehcim:
For now an astropy Table is used as only backend.
It can't save to table hierarchical containers and container with dictionaries inside.
It can be used also with `with ... as` (context management) syntax.

To test and try:
```bash
python example/serialization.py
```


examples/serialization.py
```diff
@@ -0,0 +1,56 @@
...
+
+from ctapipe.io.serializer import Serializer
+from astropy import log
+import pickle
+import gzip
+
+log.setLevel('DEBUG')
+
+def write_dl0_example(filename, data):
+    S = Serializer(filename, overwrite=True)
```
jacquemier: variable in function should be lowercase

examples/serialization.py
```diff
@@ -0,0 +1,56 @@
...
+    # Save table to disk
+    S.save()
+    return S
+
+
+def write_dl1_tel_example(filename, data):
+
+    t38 = data[0].dl1.tel[38]
+
+    S_cal = Serializer(filename, overwrite=True)
```
jacquemier: variable in function should be lowercase

examples/serialization.py
```diff
@@ -0,0 +1,56 @@
...
+
+
+if __name__ == "__main__":
+
+    with gzip.open('example_container.pickle.gz', 'rb') as f:
+        data = pickle.load(f)
+
+    S = write_dl0_example("output.fits", data)
+    S_cal = write_dl1_tel_example("cal_output.fits", data)
+    S_context = context_manager_example("output_context.fits", data)
```
jacquemier: variable in function should be lowercase

examples/serialization.py
```diff
@@ -0,0 +1,56 @@
...
+
+
+def write_dl1_tel_example(filename, data):
+
+    t38 = data[0].dl1.tel[38]
+
+    S_cal = Serializer(filename, overwrite=True)
+    S_cal.write(t38)
+
+    print(S_cal._writer.table)
```
jacquemier: Convert  print statements to logging

examples/serialization.py
```diff
@@ -0,0 +1,56 @@
...
+    # S_cal.write(t11_1) # This will not work because shape of data is different from tel to tel.
+
+    S_cal.save()
+    return S_cal
+
+
+def context_manager_example(filename, data):
+    with Serializer(filename, overwrite=True) as writer:
+        for container in data:
+            print(container.dl0)
```
jacquemier: Convert print statements to logging

examples/serialization.py
```diff
@@ -0,0 +1,56 @@
...
+            writer.write(container.dl0)
+            print(writer._writer.table)
+
+
+if __name__ == "__main__":
+
+    with gzip.open('example_container.pickle.gz', 'rb') as f:
```
jacquemier: Move example_container.pickle.gz to ctapipe-extra/datasets

jacquemier:
Could you please explain why It can't save to table hierarchical containers and container with dictionaries inside ?
Could we use a Fits Binary tables to store serialized data with Pickle ?

MaxNoe:
> Could we use a Fits Binary tables to store serialized data with Pickle ?

Please, no. What is the point of using a standardized file format to store data only readable by a certain program? We could then just pickle the complete object.

jacquemier:
Close and replace by
PR #241

PR #214, opened 2016-10-13
Fix tests

MaxNoe:
This removes two left over status codes (we missed two returns in the last PR) and truncates keys with more than eight characters.

A warning is raised if a truncation happens.

**Question:** Do we want that an error is raised if a key is to long? Or is it ok to truncate the key?

### DIFF
--- a/ctapipe/utils/json2fits.py
+++ b/ctapipe/utils/json2fits.py
@@ -36,10 +36,17 @@ def traitlets_config_to_fits(config, fits_filename, clobber=True):
     for section, entry in config.items():
         header = fits.Header()
         for key, value in entry.items():
-            if len(key) > 8:
-                warnings.warn('Key "{}" will be truncated to {}'.format(key, key[:8]))
 
-            header[key[:8]] = value
+            # CONTINUE and HIERARCH are incompatible, so we have to decide
+            # to either truncate key or value. I went for the key. @maxnoe
+            if isinstance(value, str):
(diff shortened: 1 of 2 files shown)


jacquemier:
Adding the HIERARCH keyword in front keyword allows keyword names to be longer then 8 characters.
By this way we do not lose information.

MaxNoe:
Then I do not understand why it failed in the first place.

From the astropy docs:

> Note
> A final point to keep in mind about the Header class is that much of its design is intended to abstract away quirks about the FITS format. This is why, for example, it will automatically created CONTINUE and HIERARCH cards. The Header is just a data structure, and as user you shouldn’t have to worry about how it ultimately gets serialized to a header in a FITS file.

I tried explicitly prepending `HIERARCH` but the test fail as well.

MaxNoe:
The Problem is: both key and value are too long. If the key is too long, we need `HIERARCH`, if the value is too long, we need `CONTINUE`.

If only one is too long, astropy handles it automatically. I will adjust the if statements to check for 
```{python}
if len(key) > 8 and len(value) > 80:
```

As I think a not truncated value is more important, right?

MaxNoe:
Ok, I think this is working now. Keys are only truncated if the value is a string and both key and value are longer than fits allows.

MaxNoe:
Ah, yes. Thats a good idea.

MaxNoe:
I think this can be merged now.

mackaiver:
I agree. Looks fine so far. :+1:

PR #215, opened 2016-10-13
Fix docs

MaxNoe:
* The `sphinx_rtd_theme` was missing on travis, thus building the docs failed.
* Some fixes to the docs


mackaiver:
Seems like it still fails on travis. It says

```
The build_sphinx travis build FAILED because sphinx issued documentation warnings (scroll up to see the warnings).
```

There are quite alot of warnings. some seem to be in astropy related stuff. Not sure if we can fix those. Should we ignore those warnings for now?

MaxNoe:
There are a few warnings because of astropy and some because an example is failing, see #216 .

But alot got fixed (e.g. missing dependencies)

MaxNoe:
This works now! Tests and building the docs without warning.

PR #219, opened 2016-10-14
Camera calibration compenents

watsonjj:
WIP of update to camera calibration to use Components and Tools


ctapipe/calib/camera/charge_extraction.py
```diff
@@ -0,0 +1,281 @@
...
+
+from ctapipe.core import Component
+
+
+def all_subclasses(cls):
+    return cls.__subclasses__() + [g for s in cls.__subclasses__()
+                                   for g in all_subclasses(s)]
+
+
+class ChargeExtractor(Component):
```
MaxNoe: You need to pass `metaclass=ABCMeta` to the class definition to be able to use @abstractmethod.

ctapipe/calib/camera/charge_extraction.py
```diff
@@ -0,0 +1,281 @@
...
+        self.__window_width = np.zeros((self.nchan, self.npix), dtype=np.intp)
+        self.__window_start = np.zeros((self.nchan, self.npix), dtype=np.intp)
+
+        self.integration_window = np.zeros_like(self.waveforms, dtype=bool)
+        self.windowed_waveforms = None
+        self.integrated_waveform = None
+        self.peakpos = None
+
+    @property
+    def window_width(self):
```
MaxNoe: Using double underscores will make it impossible to inherit these values. Is it intentional? 

See https://docs.python.org/2/tutorial/classes.html#private-variables-and-class-local-references

MaxNoe: Also, why is it a property? A properties should only be needed if they do more than value assignments.

ctapipe/calib/camera/charge_extraction.py
```diff
@@ -0,0 +1,281 @@
...
+        self.peakpos = [None, None]
+
+    def get_window_width(self):
+        self.window_width[:] = self.nsamples
+
+    def get_window_start(self):
+        self.window_start[:] = 0
+
+
+class WindowIntegrator(Integrator):
```
MaxNoe: `metaclass=ABCMeta`

ctapipe/calib/camera/charge_extraction.py
```diff
@@ -0,0 +1,281 @@
...
+
+    def __init__(self, waveforms, parent, **kwargs):
+        super().__init__(waveforms, parent=parent, **kwargs)
+        self.peakpos = [None, None]
+
+    def get_window_start(self):
+        self.window_start[:] = self.window_start_arg
+
+
+class PeakFindingIntegrator(WindowIntegrator):
```
MaxNoe: `metaclass=ABCMeta`

ctapipe/calib/camera/charge_extraction.py
watsonjj: Adding metaclass=ABCMeta gives me an error I do not understand how to solve.

```
/Users/Jason/Software/oxpytools/jason_testbench/calibration_tool.py in <module>()
      5 import numpy as np
      6
----> 7 from ctapipe.calib.camera.charge_extraction import *
      8
      9

/Users/Jason/Software/ctapipe/ctapipe/calib/camera/charge_extraction.py in <module>()
     11
     12
---> 13 class ChargeExtractor(Component, metaclass=ABCMeta):
     14     name = 'ChargeExtractor'
     15

TypeError: metaclass conflict: the metaclass of a derived class must be a (non-strict) subclass of the metaclasses of all its bases
```

Do you know how to solve this?

MaxNoe: Yes. I opened a PR #229

watsonjj:
@kosack 
The latest version contains two new temporary files: 
calib.camera.factory_proposal.py - A proposal for a base factory object that would be exist in core.factory.py
calib.camera.new_tool_proposal.py - A proposal for replacing Tool, with a version that handles factories that are specified.

Full explanations can be found in the docstring.

I have included the example examples/calibration_tool.py to demonstrate how they work.

This is just my suggestion for how factory classes should be handled, and has functionality that I would hope exists in the final factory class implementation...

(like only parsing the traitlets for the class obtained via the factory, not any of the traitlets that belong to sibling classes - I believe this is important so the user isn't confused why some arguments seems to have no effect due to them being intended for a different class that could be the result of the factory given a different discriminator.)

I understand you might have a different approach for factory classes in mind...

watsonjj:
This commit contains a new factory proposal. Compared to my last factory proposal, this does not change anything about how Tools operate. However it does require that all the traits that exist in the different possible products from the factory, must also exist in the factory class. The factory class will copy the correct traits to the product class when the product is created from the factory.

Once again, I have included the example examples/calibration_tool.py to demonstrate how this proposal works.

watsonjj:
This PR requires #273 to be merged before merging.

PR #220, opened 2016-10-14
Added test to  test_muon_ring_finder, minor modif to muon_ring finder

mdpunch:
- In muon_ring_finder, changed def to follow PEP (spaces after comma)
- In test_muon_ring_finder, added test of horizontal line of pixels
 This gives NaN as ring radius, which is okay (Inf would be better)
 Should probably add a fit status to the all the muon ring finders

### DIFF
--- a/ctapipe/image/muon/tests/test_muon_ring_finder.py
+++ b/ctapipe/image/muon/tests/test_muon_ring_finder.py
@@ -22,11 +22,26 @@ def test_ChaudhuriKunduRingFitter():
 
     ZZ[(r>min_r) & (r<max_r)] = 1 * u.deg
     output = fitter.fit(XX,YY,ZZ)
-#    print(output)
+    #print(output)
 
     lim_p = 0.05 * u.deg
     lim_r = 1 * u.deg
     rad_a = 0.5*(max_r+min_r)
 
     assert(abs(output.ring_center_x-c_x)<lim_p and (abs(output.ring_center_y-c_y)<lim_p) and (abs(output.ring_radius-rad_a)<lim_r))
+
+
+def test_ChaudhuriKunduRingFitterHline():
+
+    fitter = muon_ring_finder.ChaudhuriKunduRingFitter()
+
+    x = np.linspace(20 *u.deg, 30 *u.deg, 10)   #Make linear array in x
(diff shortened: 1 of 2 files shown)


ctapipe/image/muon/tests/test_muon_ring_finder.py
```diff
@@ -22,11 +22,26 @@ def test_ChaudhuriKunduRingFitter():
...
+
+    x = np.linspace(20 *u.deg, 30 *u.deg, 10)   #Make linear array in x
+    y = np.full_like(x, 15 *u.deg)              #Fill y array of same size with y
+    z = np.full_like(x, 3.1415)                 #Fill intensity array with value
+
+    output = fitter.fit(x,y,z)
+    #print(output)
+
+    #TODO in muon_ring_fitter decide what to do if unreconstructable ... add Status Flag?
+    assert(output.ring_radius is not np.NaN)
```
MaxNoe: `assert` is not a function but a statement.

You should probably use

`assert not np.isnan(output.ring_radius)`

ctapipe/image/muon/tests/test_muon_ring_finder.py
```diff
@@ -22,11 +22,26 @@ def test_ChaudhuriKunduRingFitter():
...
     ZZ[(r>min_r) & (r<max_r)] = 1 * u.deg
     output = fitter.fit(XX,YY,ZZ)
-#    print(output)
+    #print(output)
 
     lim_p = 0.05 * u.deg
     lim_r = 1 * u.deg
     rad_a = 0.5*(max_r+min_r)
 
     assert(abs(output.ring_center_x-c_x)<lim_p and (abs(output.ring_center_y-c_y)<lim_p) and (abs(output.ring_radius-rad_a)<lim_r))
```
MaxNoe: assert is a statement, you should not use parentheses.

MaxNoe: Probably splitting this assert to multple asserts would improve readability:
```python
assert abs(output.ring_center_x - c_x) < lim_p
assert abs(output.ring_center_y - c_y) < lim_p
assert abs(output.ring_radius - rad_a) < lim_r
```

PR #223, opened 2016-10-16
Added small timing test to test_integrators.py, just to compare execu…

moralejo:
Just a small check, not shown by default, of the execution time of "local peak integration" vs. "next-neighbor peak integration" (the latter is currently very slow). The test just uses time.time(), I could not manage to use timeit to test just the integration part, without all the needed preparations like subtracting pedestals or reading in the camera geometry.

### DIFF
--- a/ctapipe/calib/camera/tests/test_integrators.py
+++ b/ctapipe/calib/camera/tests/test_integrators.py
@@ -1,7 +1,6 @@
 from ctapipe.io.hessio import hessio_event_source
 from ctapipe.utils.datasets import get_path
 from ctapipe.io import CameraGeometry
-import time
 import numpy as np
 
 from ..integrators import integrator_switch, full_integration, \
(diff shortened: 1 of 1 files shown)


astropy_helpers
```diff
@@ -1 +1 @@
-Subproject commit 3e044eb26395c8569e3e15bf821c6ae5eeea324a
+Subproject commit 160b48587680e11f12b66a51641a2601b869d1fa
```
MaxNoe: I think you accidentally downgraded astropy_helpers here

MaxNoe: Yes. We updated the astropy helpers upstream. You might have commited after that. Or you need to explictly update. You could do a `git merge upstream/master`

moralejo: But shouldn't  "git pull upstream master" have done the merge already?

MaxNoe: I don't now when you did this, we updated upstream yesterday evening

moralejo: This morning, and there were (I think) other upstream updates which did not give any similar trouble.

moralejo:
Apparently the test functions  cannot be run with python (instead of py.test), it complains about relative imports.  
Also: I think it would be good to have some speed tests of specific functions included in the standard, automatic tests - is this possible?  I mean, instead of just running the profiler  manually.

MaxNoe:
There is a relative import in line 7, which should be replace by `from ctapipe.calib.camera import ...`

There is a py.test benchmark extension:
https://pypi.python.org/pypi/pytest-benchmark

Which we could used, but I never tried it.

moralejo:
python -m cProfile -o output.txt script.py
Ok, it works once the relative import is replaced, but the "-o" output is not plain text. If instead I redirect the stdout to a file, then I can read it, but it is a 4700 lines file, in which there is no reference to the functions I am interested in. Hence I think this does not replace the simple test I wanted to have - its goal being to allow me to quickly check the effect of modifications of the next-neighbor-peak search algorithm on its speed.

MaxNoe:
Ah, sorry. Yes. The output is binary. You can look at it using the pstats module.

E. g.
```python
import pstats

stats = pstats.Stats('<filename>')
stats.sort_stats('cumulative').print_stats()
```

moralejo:
The issue with astropy_helpers should be gone (I reverted the commit, though I do not understand how could it happen in the first place). About pstats, it gives the same output that goes to stdout if the -o switch is not used, and as I said, it does not seem to be of much help for the simple comparison of the execution speed of two different pulse integrators.

moralejo:
About the astropy_helpers problem:

it seems after the change you did yesterday in astropy_helpers,  I should have done a "make develop" to get it updated (git pull did not do it because it is a Submodule I think). Instead, I did a "git add" which added astropy_helpers to the subsequent commit.

Now I no longer get the message about "new commits" in astropy_helpers when I do git_status.

MaxNoe:
Ah, yes. You have to do `git submodule update` (which is also done by `make init` and thus by `make develop`)

To be safe, do not do `git add .`, add the single files. And do not add the git submodules as long as you do  not wish to change the version of the submodules.

watsonjj:
@moralejo I am working to improve the speed of the neighbour_peak_integration scheme at the moment. I have already developed a faster `neighbour_peak_integration()` in `calib.camera.charge_extractor.NeighbourPeakIntegration(container)` #219 (which I am still developing). 

I also intend to only obtain the neighbours instead of the whole camera geometry in the new version, which should greatly improve the speed.

PR #229, opened 2016-10-17
Make Component abstract

MaxNoe:
This makes `ctapipe.core.Component` abstract (as i think it was meant to be a baseclass only).

### DIFF
--- a/ctapipe/core/__init__.py
+++ b/ctapipe/core/__init__.py
@@ -4,4 +4,4 @@ from .component import Component
 from .container import Container
 from .tool import Tool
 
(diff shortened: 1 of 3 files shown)


MaxNoe:
Yes, this is also what make we wondering. But it seems that a class needs at least one `@abstractmethod` to raise an error on creation.

MaxNoe:
I fixed the type and made parent optional. I found no solution for the Problem that on can still directly instantiate `Component`. But now it is possible to define components as base classes (for example a MuonRingFitter Component) to enforce a certain API.

PR #230, opened 2016-10-17
Revert "adding astropy_helpers"

MaxNoe:
This reverts commit 709f319e323e9dabced7db2162e33412472e50f0.


AMWMitchell:
OK, seems to be in agreement with the master, but I don't know which is correct.  @kosack can you comment?

MaxNoe:
Let me explain:

The version of the `astropy_helpers` was update in the `master`.

When you merged the master into the `muonmaster` branch, you local copy of the `astropy_helpers` was still at the old commit. Thus, git showed it as a change because the commit the submodule was at did not match the version in the now updated git submodules.

This could have been resolved by doing `git submodule update` (as done by `make init`). 

You added and commited the change, which basically reverted the update done in master.

Long story short:
After changing to a different branch or updating a local copy, do `git submodule update` or `make init`

AMWMitchell:
OK, thanks - was told essentially the same thing by someone else.

PR #231, opened 2016-10-17
Coordinates

ParsonsRD:
This is a relatively major refactoring of the code, lots of variable names have been changed such that then are more understandable and the frames have been split into 2 file to differentiate between ground frames and angular systems as this two frames type never interact with one another.

Also a new coordinate representation have been added for systems that can be described as a 2d plane. Finally the Horizon system has been added, which is functionally equivalent to the AltAz system but allows us to convert to it directly from the camera frame.

PR #233, opened 2016-10-18
Shower reco

TarekHC:
Hi @kosack, @MaxNoe, @tino-michael, please continue discussion (#208) here.

Pull request containing:
- @tino-michael's reconstruction algorithms
- New containers with "standardised" outputs
- New base class for reco algorithms (needs to be improved)


ctapipe/reco/reco_algorithms.py
```diff
@@ -0,0 +1,23 @@
+from ctapipe.core import Component
+from ctapipe.io.containers import RecoShowerGeom
+
+
+class RecoShowerGeomAlgorithm(Component):
+    """This is the base class from which all direction reconstruction algorithms should inherit from"""
+
+    def __init__(self, model = None):
```
kosack: don't override `__init__`  here, the current way Components work will break if you do that.  Probably we need to think about refactoring them though.     

What is the `model` parameter used for? Can it be removed or made a trait property of the class?

ctapipe/reco/reco_algorithms.py
```diff
@@ -0,0 +1,23 @@
...
+        Parameters:
+        -----------
+        tels_dict : dict
+            general dictionary containing all triggered telescopes data
+
+        Returns:
+        --------
+        Standard  `RecoShowerGeom` container
+        """
+        return RecoShowerGeom()
```
kosack: need a newline at end of file

ctapipe/reco/FitGammaHillas.py
```diff
@@ -0,0 +1,308 @@
...
+        pix_dirs.append(pix_dir*u.dimless)
+        
+    return pix_dirs
+
+
+class FitGammaHillas(RecoShowerGeomAlgorithm):
+    
+    def __init__(self):
+        self.tel_geom = {}
+        self.circles = {}
```
kosack: this doesn't match the `Component` `__init__`.  If you want  to define an __init__ you must call `super().__init__(...)`  to make sure the superclass is initialized correctly.  So far you aren't really using the `Component` functionality however, so you could just remove it as the base class, but if you ever want any user-configurable parameters, it is useful to have. If there are no parameters, or the paraemters can be added by wrapping the whole thing in a `Container`, then it's fine to leave the base class  of `RecoShowerGeomAlgorithm` empty.

ctapipe/reco/tests/test_FitGammaHillas.py
```diff
@@ -0,0 +1,62 @@
...
+    fit = FitGammaHillas()
+    fit.setup_geometry( *load_hessio(filename),phi=180*u.deg,theta=20*u.deg )
+    
+    tel_geom = {}
+
+    source = hessio_event_source(filename)
+
+    for event in source:
+
+        Hillas_Dict = {}
```
kosack: use lower-case for variables unless there's a really good reason not to (best to be consistent).  Classes are upper-case + CamelCase, variables are lower-case with underscores.

ctapipe/reco/tests/test_FitGammaHillas.py
```diff
@@ -0,0 +1,62 @@
...
+            except Exception as e :
+                print(e)
+                continue
+        
+        if len(Hillas_Dict) < 2: continue
+    
+        fit_result = fit.predict(Hillas_Dict)
+        
+        print(fit_result)
+        assert fit_result
```
kosack: perhaps should raise an exception of the fit result is not there, rather than stopping the whole program (there could be events that don't fit, and that's ok).  

Use `self.log.debug()` or related functions for any printing

kosack: oops, this is a test function, ignore previous comment (assert is ok!)

tino-michael:
#234 pull request to cta-observatory:shower_reco applying most of the requested changes

ctapipe/reco/FitGammaHillas.py
```diff
@@ -0,0 +1,308 @@
...
+    
+    def setup_geometry(self, telescopes, cameras, optics, phi=180.*u.deg, theta=20.*u.deg):
```
MaxNoe: Missing docs

ctapipe/reco/FitGammaHillas.py
```diff
@@ -0,0 +1,308 @@
...
+        self.OptVer = 'OpticsTable_Version{}_TelID'.format(self.Ver)
+        
+        self.telescopes = telescopes[self.TelVer]
+        self.cameras    = lambda tel_id : cameras[self.CamVer+str(tel_id)]
+        self.optics     = lambda tel_id : optics [self.OptVer+str(tel_id)]
+    
+        self.tel_phi   = phi
+        self.tel_theta = theta
+
+    def predict(self, Hillas_dict, seed_pos=[0,0] ):
```
MaxNoe: Missing docs

ctapipe/reco/FitGammaHillas.py
```diff
@@ -0,0 +1,308 @@
...
+        result.is_valid = True
+        
+        result.average_size    = np.mean([ h.size for h in Hillas_dict.values() ])
+        
+        result.alt_uncert      = -1.
+        result.az_uncert       = -1.
+        result.core_uncert     = -1.
+        result.h_max           = -1.
+        result.h_max_uncert    = -1.
+        result.goodness_of_fit = -1.
```
MaxNoe: Is this a nan-marker? Than it should be `np.nan`.

ctapipe/reco/FitGammaHillas.py
MaxNoe: Is `Hillas_dict` supposed to be a MomentContainer?

ctapipe/reco/FitGammaHillas.py
```diff
@@ -0,0 +1,308 @@
...
+                defaults to _n_angle_sum if none is given
+                _n_angle_sum seemingly superior to _MEst
+            
+            Returns:
+            --------
+            direction : length-3 numpy array as dimensionless quantity
+                best fit for the origin of the gamma from the minimisation process
+        """
+        
+        if test_function == None: test_function = self._n_angle_sum
```
MaxNoe: comparison should be `is None`.

MaxNoe: Why not set the default of test_function to `self._n_angle_sum`?

ctapipe/reco/FitGammaHillas.py
```diff
@@ -0,0 +1,308 @@
...
+from ctapipe.io.containers import RecoShowerGeom
+
+
+
+__all__ = ["FitGammaHillas"]
+
+
+
+
+def guessPixDirection(pix_x, pix_y, tel_phi, tel_theta, tel_foclen, camera_rotation=-100.893 * u.degree):
```
kosack: should be `guess_pix_direction` (following style guide)

ctapipe/reco/FitGammaHillas.py
```diff
@@ -0,0 +1,308 @@
...
+            
+            camera_rotation = -90.*u.deg
+            #if tel_id in TelDict["LST"]          : camera_rotation = -110.893*u.deg
+
+            
+            circle = GreatCircle(guessPixDirection( np.array([ moments.cen_x, (moments.cen_x + moments.length * np.cos( moments.psi + np.pi/2 ))] ) * u.m,
+                                                    np.array([ moments.cen_y, (moments.cen_y + moments.length * np.sin( moments.psi + np.pi/2 ))] ) * u.m,
+                                                    self.tel_phi, self.tel_theta, self.telescopes['FL'][tel_id-1] * u.m, camera_rotation=camera_rotation
+                                                  )
+                                )
```
kosack: still some code that goes well over 80 columns. Can you either reformat this or break it into 2 steps (e.g. define the 2 `np.arrays` and then pass them to `guess_pix_direction`?  The 80 column rule makes it much easier later to review and merge changes by looking at two pieces of code side-by-side, and forces peoepl to not use too complex lines of code.  In Python, you also don't need to use a continuation line separator if code is between parentheses, for example:

```python
guess_pix_direction(self.tel_phi,
                    self.tel_theta,
                    self.telescopes..)
```
is valid code

ctapipe/reco/FitGammaHillas.py
```diff
@@ -0,0 +1,308 @@
...
+                best fit for the origin of the gamma from the minimisation process
+        """
+        
+        if test_function == None: test_function = self._n_angle_sum
+        '''
+            using the sum of the cosines of each direction with every other direction
+            don't use the product -- with many steep angles, the product will 
+            become too small and the weight (and the whole fit) useless
+        '''
+        weights = [ np.sum( [ linalg.length( np.cross(A.norm,B.norm) ) for A in self.circles.values() ] ) *B.weight
```
kosack: reformat

kosack: in general , run `pylint` on the code if your editor doesn't do auto PEP8 compliance (or run `autopep8`, which you can install via `pip install autopep8` or `conda install -c conda-forge autopep8`

ctapipe/io/containers.py
```diff
@@ -180,3 +181,139 @@ def __init__(self, tel_id):
...
+    """
+    Standard output of gamma/hadron classification algorithms
+
+    Parameters
+    ----------
+
+    prediction : float
+        prediction of the classifier, defined between [0,1], where values
+        close to 0 are more gamma-like, and values close to 1 more
+        hadron-like (TODO: Do people agree on this? This is very MAGIC-like)
```
TarekHC: This will probably be changed in the future... But this is MAGIC `hadronness` definition. It is not clear to me how the classification will be done within ctapipe (as we don't have only 2 event classes...) but for now, this was the most straightforward thing to do.

Should we find now agreement on how to define the prediction of a binary classificator?

ctapipe/reco/FitGammaHillas.py
MaxNoe: It's taste I guess, but I usually preferer

```python
guess_pix_direction(
    self.tel_phi,
    self.tel_theta,
    self.telescopes
)
```
as it still looks well for long function names or in deeper nested structures

ctapipe/io/containers.py
```diff
@@ -180,3 +181,139 @@ def __init__(self, tel_id):
...
+    is_valid : bool
+        classificator validity flag. True if the predition was successful
+        within the algorithm validity range
+    tel_ids : uint array
+        array containing the telescope ids used in the reconstruction
+        of the shower
+    goodness_of_fit : float
+        goodness of the algorithm fit (TODO: agree on a common meaning?)
```
TarekHC: We should define a common meaning, but what I had in mind is something like this: how the different telescopes calculating each individual `hadronness` agree between each other. If 5 telescopes agree on the `hadronness=0.7` is not the same as if the 5 telescopes disagree completely, and the weighted average is 0.7. 

Does it make sense?

ctapipe/reco/FitGammaHillas.py
```diff
@@ -0,0 +1,308 @@
...
+    # alpha is the polar angle between the y-axis and the pixel
+    # to find the direction the pixel is looking at, 
+    #  - the pixel direction is set to the telescope direction
+    #  - offset by beta towards up
+    #  - rotated around the telescope direction by the angle alpha
+    
+    pix_alpha = np.arctan2(pix_x, pix_y)
+    pix_beta  = (pix_x**2 + pix_y**2)**.5
+
+    pix_beta  = pix_beta / tel_foclen * u.rad
```
tino-michael: also concerning all other comments to this function:
This is just a temporary solution @ParsonsRD is working on a proper implementation of coordinate transformations. This function is only used here and not intended for further spread. It will be removed as soon as dan's transformation from the camera frame to alt_az is available.

TarekHC:
I applied `autopep8` and corrected the bugs spotted by @kosack and @MaxNoe to the io/containers.py and reco/reco_algorithms.py.

@tino-michael, I don't want to apply the `autopep8` to the code that you are probably editing. If you want me to do it let me know.

ctapipe/reco/FitGammaHillas.py
tino-michael: because "self" is not known in the function header.

tino-michael:
@TarekHC I'm on it

ctapipe/io/containers.py
MaxNoe: No, i don't think this discussion is appropriate here.

MaxNoe:
One more comment, which affects testability:
The fit class uses a lot of global (in terms of the class) state. This makes it hard to test the different methods.

For example the `n_angle_sum` would work perfectly fine as a class method or even a simple function if it would just get the circles as argument and not look them up in `self`. 

Decoupling this will make the single methods much more easy to test.

ctapipe/reco/FitGammaHillas.py
MaxNoe: Ah, of course. Sorry.

ctapipe/io/containers.py
TarekHC: What we currently do in the CTA MC analysis using (MARS chain) is to train a RF classifier (regressor) for estimating the `hadronness` (`energy`) of each event. One RF is trained per telescope type. Then, all telescopes triggered by each event are used in each case, calculating the final `hadronness` (`energy`) by performing a weighted average.

Fancier methods will be implemented in the future, but it made sense to add at least the possibility to check for such a GoF.

ctapipe/reco/FitGammaHillas.py
kosack: perhaps mark it as deprecated so there is no confusion.  So far we have no specific decorator for that, but you can use the astropy one for now:
```python
from astropy.utils.decorators import deprecated

@deprecated(0.1, "will be replaced with real coord transform")
def func():
    ...
```

ctapipe/io/containers.py
TarekHC: We did update the energy reconstruction algorithm during the f2f meeting:

(Sorry for the private link): https://forge.in2p3.fr/issues/12794

We should probably do something similar for the case of the gamma/hadron separation. I may do it during this week if I find the time.

ctapipe/reco/FitGammaHillas.py
```diff
@@ -0,0 +1,308 @@
...
+        # the direction of this cross section is the cross-product of the normal vectors of the circle and the horizontal plane
+        # here we only care about the direction; not the orientation...
+        for circle in self.circles.values():
+            circle.trace = linalg.normalise( np.cross( circle.norm, zdir) )
+        
+        
+        # minimising the test function
+        self.fit_result_core = minimize( test_function, seed,
+                                         method='BFGS', options={'disp': False}
+                                       )
```
tino-michael: Even if the fit fails, the result is not worse than the seed, so I will leave it like this and simply return the result. For people interested in the success status, I set the result as a instance member any way.

ctapipe/reco/FitGammaHillas.py
tino-michael: it's a dictionary with the telescope IDs as keys and MomentParameters as values

tino-michael:
added documentation and reformatted lots of too long lines (80 characters seems quite few on a modern wide-screen) in #235

ctapipe/reco/FitGammaHillas.py
ParsonsRD: Hi, the code for transformation to the altaz (or Horizon) frame is now implemented and included in the pull request #231

tino-michael: nice. when the update is merged with master, I suppose we also should switch the RecoShower container's alt, az, core_x, core_y members to proper coordinates as well.

MaxNoe:
Nowadays most project use 90 characters

TarekHC:
Let me get back to this PR, as it seems stalled.

I updated this branch with the latest master.

@MaxNoe @kosack Is there anything else to be fixed? I believe it is PEP8 compliant now... Anything else missing?

kosack:
I don't see any major issues (it will evolve anyhow), so I think it's ready to include. Also, I'm im the process of completely changing the Container classes, so it would be nice to have this merged soon so I can include it in the update.

TarekHC:
One question regarding our gitHub workflow.

Should we leave this branch open for "any" shower reconstruction development? (like the muon branch? not sure). Or should we delete and re-open it with future algorithms? 

I would go for option 2. If you prefer option 1, I should not be the one "in charge", as others are way more active than me. :)

kosack:
I'd delete it once it's merged, and create a new branch if there is future development. Otherwise, it becomes confusing if the branch is under development or not.

kosack:
@MaxNoe  are you ok with the changes they have made? Your request is blocking the merge for now.

PR #236, opened 2016-10-19
Add Processor, a Component-like function

MaxNoe:
I observed how much functionality now is put into differently named methods of classes inheriting from `Component` which normally would just be simple functions.

Thinking about the pipeline, I think we should have a common interface for function-like `Component`. 
For that I created the `Processor` class which has the abstract method `__call__`. 

So for example something like the `FitGammaHillas` from #233 would have to override `__call__` instead of having a method `predict`. Or the RingFitter would also override `__call__` instead of having a method `fit`.

### DIFF
--- a/ctapipe/core/__init__.py
+++ b/ctapipe/core/__init__.py
@@ -1,7 +1,8 @@
 # Licensed under a 3-clause BSD style license - see LICENSE.rst
 
 from .component import Component
+from .processor import Processor
 from .container import Container
 from .tool import Tool
 
-__all__ = ['Component', 'Container', 'Tool']
+__all__ = ['Component', 'Container', 'Tool', 'Processor']
--- a/ctapipe/core/processor.py
+++ b/ctapipe/core/processor.py
@@ -1,0 +1,19 @@
+from abc import abstractmethod
+from .component import Component
+
+
+class Processor(Component):
+    '''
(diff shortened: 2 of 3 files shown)


kosack:
This is also closely related to the `flow` notion of components, which are already higher-level wrappers for Components. We want to avoid re-implementing a flow-based framework, and rather wrap algorithms in one. Of course, one could just change the `run(input)` method to call `self(input)`.

The one question I have: is it always clearer to override `__call__` rather than implement a method that describes exactly what is happening for a given algorithm? (e.g. having all algorithms that make parameters use `parameterize()`, or all that do a fit call it `fit()`, are clear to what they are doing). 

E.g. which is simpler:

```python
cleaner = MyCleaner(threshold=3)
parameterizer = MyParameterizer(option=12, option2='a')

# which is more readable:

cleaned = cleaner.clean(image)
params = parameterizer.parameterize(cleaned)

# or:

cleaned = cleaner(image)
params = parameterizer(cleaned)

```

The second is simpler, but could be confusing if the classes do more than one thing or are named poorly.

kosack:
FOr now, I think this is premature, since we don't even have a settled upon structure for algorithms. I'd rather wait until we have things implemented and then see how to make them have a common interface.

Issue #71, opened 2016-03-17
Should we process data by Chunk (set of N events) or by Event (single event)?


kosack:
Should we chunk the data or not? Here we are not talking at the file or i/o level (which is chunked by default in some optimal way), but rather for the algorithms.  Therefore it implies a major difference to the internal data structure.
#70 is an example implementation of a chunked data structure that could replace the current `RawDataContainer` structure in `ctapipe`
- _event-wise_:  
  - simple to implement algorithms
  - low memory footprint
- _chunk-wise_:
  - can make use of CPU and GPU features, since have many events in a block of memory
  - the camera trace data is a 3D array (nevnt, npix, ntimeslice)
  - simplifies some algorithms that need to loop multiple times over a set of events

However, it was pointed out that if we use a fancier stream processing system, that system can buffer events into a chunk, and provide efficient event-wise access.

mackaiver:
I agree that a chunk-wise operation can be really useful for some algorithms that need multiple events.

I think many of these algorithms need many (in the order of thousands) events to accumulate  reasonable statistics for whatever they are calculating. In that case we need huge chunks or have running statistics over many chunks. 
For the latter we could use something like  [python_runstats](https://github.com/grantjenks/python_runstats) (I haven't tried it but it looks nice.)

> However, it was pointed out that if we use a fancier stream processing system, that system can buffer events into a chunk, and provide efficient event-wise access.

I'm personally not to worried about that. As long as the chunksize is reasonably small this will work just fine. The RDD messages used in Apache Spark for example are basically "chunks".

However I think  it's important to completely hide the implementation details from the user.

kosack:
seems the answer so far is by event

PR #241, opened 2016-10-20
Pickle serializer

jacquemier:
Implement a Container serialization with Pickle and gzip
This can be used to exange data of different level while waiting a Fits implementation

Refer to PR #212


ctapipe/io/serializer.py
```diff
@@ -0,0 +1,356 @@
...
+from astropy import log
+from pickle import load
+from pickle import dump
+from pickle import PickleError
+from traitlets import Unicode
+import numpy as np
+import gzip
+
+__all__ = ['FitsSerializer', 'PickleSerializer']
+log.setLevel('INFO')
```
MaxNoe: Library code should not set a log level

MaxNoe:
I think we should not keep all containers in Memory until a final write. A serializer should be able to serialize streamed if the output format permits it (most do, even fits).

ctapipe/io/serializer.py
```diff
@@ -0,0 +1,356 @@
...
+
+    @abstractmethod
+    def write(self):
+        """
+        Write all containers to outfile
+        """
+        pass
+
+    @abstractmethod
+    def __iter__(self):
```
MaxNoe: I don't think a Serializer should be iterable. A Serializer could take an iterable.

ctapipe/io/serializer.py
```diff
@@ -0,0 +1,356 @@
...
+        """
+        self.file = file
+        self.overwrite = overwrite
+        super().__init__()
+
+    def __enter__(self):
+        log.debug("Serializing on {0}".format(self.file))
+        return self
+
+    def __exit__(self, *args):
```
MaxNoe: I think if a serializer keeps all objects in Memory, then `__exit__` should call write.

Also, the call signature for exit should be `__exit__(self, exc_type, exc_value, traceback)`, see https://docs.python.org/3/reference/datamodel.html#object.__exit__

ctapipe/io/serializer.py
```diff
@@ -0,0 +1,356 @@
...
+        try:
+            file_split = self.file.split('.')
+            if file_split[-1] != 'gzip' or file_split[-2] != 'pickle':
+                output_filename = self.file+'.pickle.gzip'
+            else:
+                output_filename = self.file
+            with gzip.open(output_filename, 'wb') as f_out:
+                dump(self.containers, f_out)
+            return output_filename
+        except (PickleError, PermissionError, FileNotFoundError):
```
MaxNoe: What's the point of catching the Exception, if you directly reraise it? Did you maybe forget to log the exception?

elehcim:
Thanks @jacquemier. Integrating pickle serializer has a big practical value expecially for current DL0 and DL1 data. I can continue on the FITS serializer after the CM in Bologna, but please feel free to close this if you want. Actually I'd like to test on the recent `Reco*` (cf. #233) and `Muon*` containers which are very suitable to be serialized on FITS, when the relative PR will be merged.

ctapipe/io/serializer.py
jacquemier: I did not catch the difference ?

elehcim:
> I think we should not keep all containers in Memory until a final write. A serializer should be able to serialize streamed if the output format permits it (most do, even fits).

Yes, maybe this is an advanced feature, in my opinion not with high priority now. In case of astropy.Table back-end, there are a [StreamingHDU](http://docs.astropy.org/en/stable/io/fits/api/hdus.html#streaminghdu) class and -when the format is HDF5- an ["append"](http://docs.astropy.org/en/stable/io/unified.html#hdf5) keywork argument which can be used to add single rows (of sub-tables, aka chunks) to an existing table. We can add an option to save chunks of data. Of course the chunk size should be tunable, even 1 event at a time (see #71).

ctapipe/io/serializer.py
MaxNoe: the `__iter__` dunder method is called when you do `iter(serializer_instance)` or if you do stuff like `for stuff in serializer_instance`.

MaxNoe: Ah, sorry. Now I get It. You mixed serialization and deserialisation. I think we should keep that seperate. Having writers and readers. We sources in `io`, correct? 

So maybe a PickleSource would be a better idea.

Especially: what happens if I mix iteration with `add_container`? I think this should not be possible and an easy solution would be to have seperate classes for readers and writers.

elehcim:
Concerning the class architecture, after a brief talk with @kosack in Paris, for the future we were actually oriented to an object relationship where a Serializer object _has-a_ Writer object rather than _is-a_ Writer object. This may allow more naturally for multiple back-ends which can be chosen by the user, just like astropy.Table does in case of different "format" keyword.
For example:

    s = Serializer(writer=TableWriter("output.fits"))
    s = Serializer(writer=ImageWriter("output.img"))
    # output.img could be for example an hdf5 file (with data with different sizes not writable on FITS Binary Table)

or with some mapping-like tricks, more user-friendly

    s = Serializer(writer="fits", file="output.fits")

But now this is quite bike-shedding maybe, let's start with something usable. We can modify it later, I think.

ctapipe/io/serializer.py
jacquemier: I mixed serialization and deserialization in order to be close to Pickle implementation.
There is no sources in io (at least in mater branch)
For your last point, add_container and iteration could not be excecuted on same object because file must be open with a read OR a write flag. An opening with write flag do not allows to read.

_In [4]: file_object = open('test.pickle', 'wb')
In [5]: print(load(file_object))
UnsupportedOperation                      Traceback (most recent call last)
<ipython-input-5-c9947b36a037> in <module>()
----> 1 print(load(file_object))

UnsupportedOperation: read
_

ctapipe/io/serializer.py
```diff
@@ -4,90 +4,255 @@
...
+    def write(self):
+        pass
+
+
+class Reader(ABC):
+    def __init__(self, infile):
+        self.infile = infile
+
+    @abstractmethod
+    def get_next_container(self):
```
MaxNoe: python objects can have a __next__ method, which is called for iterators and if you do next(object). I think this would fit quite well here.

ctapipe/io/serializer.py
```diff
@@ -0,0 +1,426 @@
...
+        if self.format == 'fits':
+            self._writer = TableWriter(outfile=filename, mode=mode,
+                                       format=format)
+        elif self.format == 'pickle':
+            if mode == 'r':
+                self._reader = GZipPickleReader(infile=filename)
+            else:
+                self._writer = GZipPickleWriter(outfile=filename,
+                                                mode=mode)
+        elif self.format == 'img':
```
MaxNoe: I think this should be a general `else`, like this a completely unknown format string would not result in an error

jacquemier:
Refractor Serializer  according to suggestions issued by Max and Michele.
- Concerning class architecture, now Serializer object has-a Writer object 
- Serializer class is able to Read and Write Container. But I am not against separate them if there are good reasons.
- The 'format' argument shadows built-in name. I let it because it corresponds to astropy.Table.write argument

elehcim:
I've produced a refactored (again) Serializer code with a separate Reader (or deserialization) part. My work is here: https://github.com/elehcim/ctapipe/tree/mm/serializer.

I've opened another PR #246.

If needed please open a branch with push access for both so that we can collaborate easily and avoid opening several similar PR.

ctapipe/io/serializer.py
```diff
@@ -4,90 +4,255 @@
...
+        """
+        Parameters
+        ----------
+        infile: str
+             full path input file name
+        """
+        super().__init__(infile)
+        self.file_object = gzip_open(infile, 'rb')
+
+    def get_next_container(self):
```
jacquemier: Done in PR https://github.com/cta-observatory/ctapipe/pull/246

ctapipe/io/serializer.py
jacquemier: Done in https://github.com/cta-observatory/ctapipe/pull/246

PR #245, opened 2016-10-26
added cutflow class and unit test

tino-michael:


ctapipe/utils/tests/test_CutFlow.py
```diff
@@ -0,0 +1,46 @@
...
+                                  UndefinedCutException, \
+                                  PureCountingCutException
+import numpy as np
+from pytest import raises
+
+
+def test_CutFlow():
+    flow = CutFlow("TestFlow")
+    flow.set_cut("smaller5", lambda x: x < 5)
+    flow.add_cut("smaller3", lambda x: x < 3)
```
kosack: what's the difference between `add_cut` and `set_cut`?

kosack:
This is useful for cases when we want to cut events one-event-at-a-time, but that is not the normal mode of operation.  

The more correct way will be to generate all events in an event list, and then apply cuts to the list all at once (which is more speed efficient and flexible).  However, this could still be useful for some small cases during reconstruction where things are cut out before making it into the event list, so I'm happy to include it as long as we warn users it is not intended for things like event selection.

ctapipe/utils/tests/test_CutFlow.py
tino-michael: there is none, one is alias to the other.
we talked about it where I named it `set_cut` and you mentioned maybe `add_cut` would be more suitable. but because it overwrites any existing cuts with the same name, I left `set_cut` and simply added the alias.

tino-michael:
> This is useful for cases when we want to cut events one-event-at-a-time, but that is not the normal mode of operation.

Of course, if you already have all events reconstructed and classified in a table, an event selection by masking the rows of the table is more efficient.
I use this to keep track of events and images during calibration → cleaning → reconstruction → classification, not during the creation of the IRFs...

PR #246, opened 2016-10-27
Separate serialization and deserialization

elehcim:
Here a second version of the serializer including @jacquemier code on Pickle serializer. I separated the Serializer and deserializer code. I think this clarifies the interface.
It includes:

- [Iterator Protocol](https://docs.python.org/3/library/stdtypes.html#iterator-types) for readers
- [Context management](https://docs.python.org/3/library/stdtypes.html#context-manager-types) for serializer
- FITS and Pickle serializer (thank you Jean)
- Pickle Reader

TODO:
- [ ] Test on Muon/Shower containers
- [ ] Change `ctapipe.core.Container` underlying class from dict to OrderedDict
- [ ] Implement FITS->Container reader (for now data can be read using astropy.io.fits), but a common reader is needed

low priority:
- [ ] Implement append mode (see [StreamingHDU](http://docs.astropy.org/en/stable/io/fits/api/hdus.html#streaminghdu) and ["append" keyword](http://docs.astropy.org/en/stable/io/unified.html#hdf5) )


kosack:
see also PR #249 , which will have an impact on this (new container class implementation)

PR #249, opened 2016-11-04
Improved container classes and event hierarchy

kosack:
This pull request includes a large change to the way `Container` classes work, their implementation, and the definitions in the `io` module for an event.

All tests and examples have been updated, but it may break some code that is not checked into ctapipe (though it is not difficult to update)

Major changes:

* `core.Container` base class:
  * are defined in a more *pythonic* way (no more c-style `add_item()`), with metadata for each item stored in the class
  * now closer to being column-definitions for an output file
  * supports telescope-wise dictionaries via `Map`
  * can be converted to `dict`s both recursively and with a *flatten* option (where sub-containers append their name to the keys creating columns for each sub-item)
  * default values can be re-applied with the `reset()` method, so they do not need to be set manually.
* `io.containers` module:
  * restructured some containers to have a more clear hierarchy, and renamed to *Container
  * moved some container attributes to be metadata rather than data (things that do not change per-event)
* `io` sources:
  * updated `hessio_event_source` and `mock_source` to work with new scheme. Others pending (could not test e.g. the compressed fits source)
* other modules and examples:
  * updated to use new scheme
  * provided examples and documentation

containers are now defined like this:

```python
class SubContainer(Container):
    z = Item(default=100, "z value")

class MyContainer(Container):
    x = Item(default=0, description="x value", unit=u.m)
    y = Item(-1, "y value", unit=u.m)
    tel = Item(Map(SubContainer), "tel_id to sub-container map")
    sub = Item(SubContainer(), "a container hierarchy")

cont = MyContainer()
print(cont.y)  # returns -1, since default is set automatically
cont.z = 100 # throws AttributeError, since z not defined
cont.sub.z = 100 # ok

print(cont.tel[100].z)  # prints 100, and tel[100] gets allocated on-the-fly

print(cont)  # prints all container items and sub-containers

cont.x = 10
cont.reset()  # sets x back to default of 0

```


MaxNoe:
I think this is a big step forward.

I would also recommend to get rid of acronyms like `inst` or `optical_foclen`. Just write it out.

kosack:
Yes, I agree. I want to do a second pass and rename a lot of the variables. So far, I wasn't sure what a lot of them meant, particularly in the CalibratedCameraContainer which I think needs more cleaning up (so will have to as the calib people for help).   

For the `InstrumentContainer`, it is only intended as a stop-gap until the instrument module works in a nice way, so will eventually disappear. 

For the initialization, you're right it should be possible to have an init function that just maps keys to the items. Is it ever really needed though?

kosack:
I've merged in and re-named the Reco containers.  @TarekHC  you might want to take a look to make sure I haven't missed something. THis will replace what is in the master soon.

Now there is a dl2 container (a ReconstructedContainer) in the top-level EventContainer that works as follows:
```python
   event.dl2.shower['algo']  # shower params for algorithm "algo"
   event.dl2.energy['algo']  # energy from algorithm 'algo'
   event.dl2.classification['algo'] # classification from algorith 'algo'
```

it works like this:
```python
from ctapipe.io.containers import EventContainer
ev = EventContainer()
ev
```

```
out[]:
ctapipe.io.containers.EventContainer:
                         dl0.*: Raw Data
                        trig.*: central trigger information
                          mc.*: Monte-Carlo data
                         count: number of events processed
                         dl1.*:
                        inst.*: instrumental information (deprecated
                         dl2.*: Reconstructed Shower Information

```


```python
in[]: 
ev.dl2

out[]:
ctapipe.io.containers.ReconstructedContainer:
                     energy[*]: Map of algorithm name to energy info
                     shower[*]: Map of algorithm name to shower info
             classification[*]: Map of algorithm name to classification info

```

kosack:
Does anyone have a problem if I make this merge? It's a large change, but all unit tests pass currently, and it's better to do it soon before more code is written that needs modification.

TarekHC:
I'm totally ok with it also. In 2 weeks (after MAGIC general meeting) I will find the time to play with the new containers.

kosack:
Ok great! I'll merge it shortly (just waiting for some better descriptions of a few items from the calibration people)

MaxNoe:
Do we have any advantages with the code used here to forbid creating members compared to pythons standard method using `__slots__`?
```
In [1]: class Point:
    ...:     __slots__ = ('x', 'y')
    ...:     
    ...:     def __init__(self, x, y):
    ...:         self.x = x
    ...:         self.y = y
    ...:         

In [2]: p = Point(0, 0)

In [3]: p.x
Out[3]: 0

In [4]: p.z = 3
--------------------------------------------------------------------------
AttributeError                            Traceback (most recent call last)
<ipython-input-14-a288bf462759> in <module>()
----> 1 p.z = 3

AttributeError: 'Point' object has no attribute 'z'
```

kosack:
There's no real difference, I think, except I couldn't figure out a way to nicely set `__slots__` from the class definition at the same time as assigning the metadata for each Item.  Perhaps it can be done with a metaclass, though. In any case, the mechanism is hidden from the user, so we could use `__slots__` instead of modifying `__setitem___` if it seems better in the future.

kosack:
The big issue for me is now how to proceed with the data structures. I like this new mechanism, so I may just merge it as is, and start a new PR for bigger changes.  There is the second issue of re-structuring the data, which i started to a small extent here, but didn't make any large changes other than better naming of variables.  Should we explicitly make it work like the *CTA Data Model Definitions* document? That would make the meaning of each data item very clear, but more complex (e.g. `event.dl0.trig.tels_with_trigger` becomes `dl0.sub.event.trig.tels_with_trigger`).  That may be better left to the future (though it could be a lot of work to change).

The (somewhat official but not yet accepted) hierarchy in the document is:

image

However, I'm the main author, so it could easily be changed if we don't like it.

TarekHC:
Hi @kosack ,

I finally have time to play with the "dl2" containers you created, and start implementing some preliminary figures of merits. But some questions pop-up all the time. I'm afraid they might be a bit naive, but to non python experts they are not...

- Are these containers designed to be used to contain the information of one event? Or bunches?
I understand that these containers may contain any type of variable, such as numpy arrays, and that algorithms should be designed to used numpy operations, etc... Is this correct?

- Are they still reasonable if using very large arrays? 
For example: if I want to import many events from Prod-3, contain all their reconstructed quantities, and do operations with them. Is it reasonable to create one "dl2" container and fill it with huge (size ~10⁶) arrays? Or should we bunch those events and do operations "per bunch"?

As a side note, maybe adding simple examples within a "test_containers.py" would help. I could take care of that, as long as I am sure about the correct method to fill them.

Thanks!

Tarek

Issue #244, opened 2016-10-26
Remove code using instrument module to read camera def


kosack:
The issue with segfaults in #226  is likely due to the opening of 2 simultaneous simtel files (which the underlying libhessio.so library does not support). 

Any code that needs to read the camera defs should not use the  `instrument` module until it is redesigned. 

`io.CameraGeometry` should be sufficient, and it should only be constructed from the data in the `EventContainer` generated by `hessio_event_source`, not via `pyhessio` functions directly. 

This is causing very unstable code, that fails randomly on macOS machines in particular, and occasionally on linux (due to different memory handling likely)

kosack:
This particularly affects the test code in the calib module which must be re-written

tino-michael:
are we sure that the dictionary in `_npix_to_type` is correct/complete?
particularly for ASTRI cameras I found a rotation of −90° and not the given 0°
the attached plots show the "error" on the tilt angle of the hillas parametrisation for my rotation (top) and the no rotation (bottom)

![alpha_vs_signal_-90degree](https://cloud.githubusercontent.com/assets/18286015/20184990/b8832bd0-a76a-11e6-83a9-0e226bb895c3.png)
![alpha_vs_signal_0degree](https://cloud.githubusercontent.com/assets/18286015/20184994/bb0ff1a8-a76a-11e6-8d7d-b1ec49b2160b.png)

kosack:
No, I think the `npix_to_type`  array was simply reverse-engineered duing the Prod-2 production, and almost certainly needs updating.

PR #251, opened 2016-11-14
update for ShowerReco

tino-michael:
Moved from InstrumentDescription to event.meta for instrument information
as has been pointed out by Karl in #244

additionally moved the score functions for the direction and position minimizer
outside of the fit class to improve modularity and testability (as proposed by
Max in #233#issuecomment-254495070)

also fixed incorrect rotation for ASTRI telescopes in io.camera._npix_to_type

added unit test specifically for the core fit


tino-michael:
@kosack this passes the tests now and can be merged if there are no further issues.

PR #266, opened 2016-12-01
2 d camera geometry

tino-michael:
This module converts hexagonal camera grids stored in 1D arrays into a square grid stored in 2D arrays through slanting and stretching. Back-conversion is provided as well.

It supports summed pixel arrays -- e.g. `event.dl0.tel[tel_id].adc_sums[0]` -- and time-sliced arrays -- e.g. `event.dl0.tel[tel_id].adc_samples[0]`.

The functions create a `CameraGeometry` object which can be used to draw the camera image with `visualization.CameraDisplay` as usual.

This is intended for algorithms that need rectangular images/pixels like the wavelet cleaning that we are using. Preliminary tests show that this wavelet cleaning works also on hexagonal images skewed to look rectangular. Attached are two example images to demonstrate the forth and back conversion.

geometry_conversion_test_1
geometry_conversion_test_2

### DIFF
--- a/ctapipe/image/geometry_converter.py
+++ b/ctapipe/image/geometry_converter.py
@@ -4,7 +4,7 @@ from astropy import units as u
 
 from numba import jit
 
-from .camera import CameraGeometry
+from ctapipe.io import CameraGeometry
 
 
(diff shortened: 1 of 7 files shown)


tino-michael:
You are right of course. I'm not sure how I ended up putting this in io...

PR #268, opened 2016-12-02
Colored log output

watsonjj:
Created new logging.Formatter for Tool logger so that levels have colored format

### DIFF
--- a/ctapipe/core/tool.py
+++ b/ctapipe/core/tool.py
@@ -1,8 +1,43 @@
 from traitlets import Unicode
 from traitlets.config import Application
 from abc import abstractmethod
+import logging
 
 from ctapipe import version
+
+
+class ColoredFormatter(logging.Formatter):
+    """
+    Custom logging.Formatter that adds colors in addition to the original
+    Application logger functionality from LevelFormatter (in application.py)
+    """
+    highlevel_limit = logging.WARN
+    highlevel_format = " %(levelname)s |"
+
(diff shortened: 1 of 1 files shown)


MaxNoe:
There is: https://pypi.python.org/pypi/colorlog, maybe it's interesting for the task.

kosack:
isn't that already the case if we use the astropy logging?  It comes out in color for me.

kosack:
though I'm happy to have a local version, since it would be nice to abandon the dependency on astropy-helpers

watsonjj:
Yes that is true if you use the astropy logger. However the Tool logger is instead from the traitlets Application module. These are two individual loggers configured differently.

kosack:
also, we need this to be able to be disabled if the terminal doesn't support color, or if the log is redirected to a file, so you don't get non-printable characters.  I think that is possible since I've seen other code that does it (and I've done similar in C++).   Alternately what we can do is always write the log to the screen and a file (using the logger's ability to have two streams), and have only the screen one formatted in color.

kosack:
aha. one way to do it is to use the  "colorama" package (in anaconda) instead of doing the ANSI escape-sequences manually. It adds yet another dependency, but it's small, and it automatically turns off if the output is redirected to a file. 

https://github.com/tartley/colorama

watsonjj:
Okay, I will look into changing this pull request to use colorama in the Tool logger later

PR #271, opened 2016-12-08
Impact interpolator

ParsonsRD:
This is just a simple first implementation of the ImPACT code 
https://arxiv.org/abs/1403.2993

This is probably jumping the gun a little as we don't have other reconstruction methods available yet, but this branch is here just to allow Jason and I to work on it.


ctapipe/reco/shower_max.py
```diff
@@ -36,11 +36,11 @@ def __init__(self, filename, col_altitude=0, col_thickness=2):
 
         self.atmosphere = Histogram(axisNames=["altitude"])
         self.atmosphere.hist = thickness * u.g * u.cm ** -2
-        self.atmosphere.bin_lower_edges = [np.array(altitude) * u.km]
+        self.atmosphere._binLowerEdges = [np.array(altitude) * u.km]
```
tino-michael: I have another pull request pending that nobody wants to merge dealing with the renaming of this variable.
It has been suggested to not use `_binLowerEdges` but `_bin_lower_edges` which needs to be changed in the histogram class...

kosack: which pull request was that? I'll merge it but can't find it

tino-michael: #205 I guess you found it ;)

ParsonsRD: Ah cool, if this request has been approved I will update this branch today.

ParsonsRD:
The Travis build seems to be failing based on the documentation build. But I can't figure out the problem, could someone take a look and let me know if they figure it out.

kosack:
yeah, I tried to fix the doc warnings, but couldn't figure it out. I did find some issues with old code that wasn't a problem before, so maybe it's due to a new version of sphinx or something.

ParsonsRD:
Would it be OK to merge this to the feature branch, then we can fix any problems before the merge to the master branch?

PR #272, opened 2016-12-08
In hessio.py read instrument data only once

watsonjj:
Read instrument data only once.

This is possible with cta-observatory/pyhessio#42 as num_channels is now obtained from hsdata->camera_org[itel].num_gains instead of hsdata->event.teldata[itel].raw, allowing it to be extracted regardless if data exists for that telescope in the current event.


kosack:
seems to still have a few tests failing... I'll wait to merge

watsonjj:
Unfortunately there doesn't seem to be a place where you can extract the number of samples for each telescope if the current run does not have any data...

Therefore either num_samples should be completely removed from event.inst, or we fill it like in the following commit:

watsonjj:
I suggest it is instead moved from event.inst.num_samples[tel_id] to event.dl0.tel[tel_id].num_samples, as a convenience so users can avoid event.dl0.tel[tel_id].adc_samples[0].shape[1]

ctapipe/io/containers.py
```diff
@@ -78,6 +77,7 @@ class RawCameraContainer(Container):
                             "integrated ADC data (n_pixels)"))
     adc_samples = Item(Map(), ("map of channel to arrays of "
                                "(n_pixels, n_samples)"))
+    num_samples = Item(int, "number of time samples")
```
kosack: remember that anything in this container is something that changes *per event*, so `num_samples` does not belong here. It could however go in the container's metadata.  I would still call it an instrumental feature, though, so it should be in the `inst` container.

kosack: think of Containers as column definitions of a table.  Would you want to repeat num_samples for every row in the table? If it never changes, then it doesn't go there, but in a "header" container like inst, or in the metadata (in which case it becomes header keywords in the file)

watsonjj: Understood...but with the way num_samples is returned from pyhessio, it does change per event. If a telescope has data in the current event, this could equal 96, but for the next event if the telescope does not have data, this would then equal 0 (but the key in event.dl0 would not exist of course).

watsonjj: I very much dislike the way it is right now. But there does not seem to be an alternate method in hessio to obtain number of samples directly. Perhaps the best option would be to not store this value, forcing the user to obtain the number of samples from adc_samples.shape.

kosack: it's strange that it changes per event in pyhessio, but that must have to do with how the data is read. That is still header info, not event info.  Why is it a problem reading it from the inst container?

Of course remember that in the end reading any camera or layout info from the hessio containers will disappear, so this is all just a temporary hack. Eventually this will be an interface to a database or local data description file handled by the instrument module. You'll get a structure somewhat like `CameraGeometry` that contains a num_channels field. 

And of course even further down the line, there will be no channels visible to the calibration code, since we will start with DL0 data and they are dropped by the camera in R1.

watsonjj: The reason it changes per event is because its obtained from `hsdata->event.teldata[itel].raw->num_samples`. This used to be where num_channels was obtained from before my pyhessio PR, but now i obtain it from `hsdata->camera_org[itel].num_gains`, which allows you to obtain the num channels for every telescope, even if the telescope has no data in the current event.

Sure num_channels will be dropped, but its num_samples that we have a problem about in this discussion.

Would you be happier with solution I used in 0a9e02d33187741be71b49ca41dace75333de2af, where i fill the num_samples[tel_id] as soon as we reach an event where the tel_id has data. This is similar to the method used in master, however it is only num_samples that updates at each new event, instead of the whole inst container.

PR #273, opened 2016-12-08
Num channels extra dimension

watsonjj:
Change containers such as dl0.tel[].adc_samples, dl0.tel[].adc_sums, and mc.tel[].reference_pulse_shape such that they are no longer dicts, and instead the channel/gain dimension is represented as the first dimension (matching other arrays such as mc.tel[].pedestals.

Requires pyhessio pull request: cta-observatory/pyhessio#42


kosack:
Thanks for doing this, it is a very nice idea and simplifies the code quite a bit (and probably allows some speed as well, since we can apply algorithms to both channels at once).

kosack:
We can merge it in once the pyhessio PR is accepted

kosack:
by the way, it will probably require updating some code in `examples/` and other places. Basically anwhere where a camera is displayed or used, but that shouldn't be too tough.

watsonjj:
This PR requires #272 to be merged before merging.

PR #283, opened 2016-12-15
EventFileReader

watsonjj:
Created a new component called EventFileReader. This is a base component, from which different FileReaders can inherit from and define their own methods for reading different files that contain waveforms.

The first example of a child of EventFileReader is HessioFileReader. This handles checking if the file exists, and calling hessio_event_source with the correct arguments depending on your requirements (obtain the source generator with read(), or obtain a specific event with get_event(), or even find the number of events in the file in the fastest way possible by accessing the num_events attribute.

This provides a nice way to read files in the Tool/Component framework.

Additionally, EventFileReaderFactory can be used to obtain the correct EventFileReader depending on the file path supplied, allowing full flexibility of running scripts from any file source.

This is an improvement on the old InputFile class (or the FileReader component).

This PR first needs #219 to be reviewed and merged before merging.


kosack:
can't be merged until you resolve conflicts

PR #284, opened 2016-12-15
R0 Container

watsonjj:
A big revamp of the event container to more closely match the proposed CTA data flow.

The dl0 container was renamed to r0, as the adc_samples extracted from hessio files resembles the raw camera data that would exist as R0.

Then created an r1 container for the calibration of r0 adc_samples into pe_samples, this calibration is typically intended to be done inside the Camera Functional Unit, but for the MC has to be done as part of ctapipe. There are also instances where camera teams may want to use ctapipe as a base for their prototype software, so an r0 and r1 container also makes sense for that.

Then re-created a dl0 container, but this time containing data volume reduced pe_samples.

Alongside these new containers, the calib.camera module has been changed to better represent the different data levels: 

- r1.py handles the calibration from r0 adc_samples to r1 pe_samples. 

- dl0.py handles the application of data reduction techniques. It accepts a reductor (from reductor.py) as an argument to use for the data volume reduction method in an identical way the dl1 calibration accepts a charge extractor from charge_extraction.py. If no reductor is passed to it, no data-reduction is applied and r1.tel[].pe_samples == dl0.tel[].pe_samples. This layout provides a nice test-bench for different data reduction techniques.

- dl1.py handles the calibration from dl0 pe_samples to a dl1 image, where the charge has been extracted using a charge extractor from charge_extractor.py, and you are left with a value per pixel.

With this new calibration Component format, scripts can be made to be flexible with many data-sources. 

- MC data read through hessio is put into the r0 container, and is the calibrated through the whole chain into dl1. 

- Camera data that is already in the dl0 format is read into the dl0 container. This dl0 is unchanged when applying the methods in r1.py and dl0.py, as the r0 and r1 containers are empty. Then the data is in a identical format as the MC data is at the dl0 stage, and can continue the rest of the script in an identical fashion (calibration into dl1, plotting, etc.)

A script that uses this format is examples/calibration_pipeline.py.

This PR first needs #283 to be reviewed and merged before merging.


ParsonsRD:
Hi @watsonjj , from the travis build it looks like the calibration tests need to be updated for the new data structures.

watsonjj:
@ParsonsRD Yes, there was an update to pyhessio, but the version used by ctapipe (and travis) still points to the old version. This needs updating to use the new one, then the tests should pass.

ParsonsRD:
Ah OK, thanks!

jacquemier:
Concerning the update of pyhessio.
API changes are not backward-compatible (get_num_samples does not exist anymore).
So we have to increment the major version: from 1.0.1 to 2.0.0
Versions in setup.py and meta.yaml must be updated to this number.
A new tag and a new release must be created on pyhessio github repo (https://github.com/cta-observatory/pyhessio/releases)
pyhessio version must also be updated in ctapipe .travis.yml file

One remarks about API changes that are not backward-compatible:
Once commited, developers must update their pyhessio installation, which is quite tedious, so, from my point of view, in the future,  will be better to avoid to change the API to often.

kosack:
This is looking good! All the tests fail though. Does travis.yml need updating? Does it work on your local machine?

watsonjj:
Yes I think it does need updating. All tests pass on my machine.

kosack:
great. Can you check that the version of pyhessio you use locally is the same one downloaded and installed in travis.yml?  Probably just updating that will fix it, since the tests that fail are ones that use pyhessio.

watsonjj:
I think a new release still needs to be made of pyhessio. V1.0.1 (the latest release and the one used in travis.yml) does not have my latest pull requests.

kosack:
I merged one of your earlier PRs and now I cannot merge this due to conflicts, can you merge them in?

ParsonsRD:
Hi, can I just give a bump to this request. Is there are there any serious issues holding this one back? I guess this is a rather major revision to the code structure and I'm a bit reluctant to change much while this is still open.

kosack:
NO serious issues, other than it requires everyone to update all other branches.  Now that there is a stable release of ctapipe-0.3.4, it should be ok to make this change. We just have to warn the other large feature branches that they will need to make a lot of changes to ensure they still work.

ParsonsRD:
Ah OK, I think both Alison and I are doing rather major refactoring of the code at the moment anyway, so is no big problem for us.

kosack:
I made a number of small changes to fix bugs introduced by this PR, and I guess that means I need to open a new PR to both merge this in and include the changes.  I will merge this first and then pull them.  Please wait until the second PR is merged

PR #287, opened 2017-01-04
Added 2 Hillas routines, based on old Whipple ForTran code, + corrections

mdpunch:
…, other corrections

Removed returning of asymmetry, just left skewness...
    either MP/Whipple or MAGIC asymmetry can be calculated then in function

Added two Boolean variables to the call: higherMoments (default True) and ReCalcPix (default True)
    If higherMoments is False, then it doesn't calculate them (if True, calculated for those routines have them)
    If reCalcPix is False, then the multiples of pixel positions are calculated always (if True, only if not existing)

Hillas_parameter_1 doesn't work for the case of horizontal image
    It raises an exception wrongly; no reason to raise exception for S_xy = 0

In hillas_parameter_1, it says delta/psi is angle with x-axis, but adds pi/2 (as does hillas_parameter_2).
    cos_delta and sin_delta are correct, though.
     => changed this in hillas_parameter_1, and in modified the display visualization (in visualization/mpl.py)... and "mock.py" in reco.

Added higher moments to hillas_parameters_2, and added a sub-function to hold static variables
    of pixel multiplications, if pixel positions don't change

Added Pythonified version of hillas_parameters_3, called hillal_parameters_4, also using
    a sub-function's attributes to hold "static variables" of pixel multiplications, if pixel positions don't change

Added the historic 1993 Whipple calculation (translated Fortran code via c), as hillas_parameters_3
    Note, MAGIC's "skewness" is the same as the Whipple/MP "asymmetry^3", why ???
    ... and also, Whipple/MP "asymmetry" * "length" = MAGIC "asymmetry"
    ... so, MAGIC "asymmetry" = MAGIC "skewness"^(1/3) * "length"
    I don't know what MAGIC's "asymmetry" is supposed to be.

MAGIC (hillas_parameters_1) seems to be using the Whipple 1989 Weekes et al. formula calculation!
  The Whipple 1993 Reynolds et al formula calculation is better (using non central or raw moments, not central moments).


ctapipe/reco/hillas.py
```diff
@@ -181,73 +306,595 @@ def hillas_parameters_2(pix_x, pix_y, image):
...
+    tanpsi_denom = (2.0 * vxy * ym) - (dd - zz) * xm
+
+    psi = (np.arctan2(tanpsi_numer, tanpsi_denom)) * u.rad
+
+    if not higherMoments:
+        return MomentParameters(size=size, cen_x=xm*unit, cen_y=ym*unit,
+                            length=length*unit, width=width, r=rr*unit, phi=phi, psi=psi.to(u.deg), miss=miss*unit)
+
+    # -- Asymmetry and other higher moments
+    if length != 0.0:
```
kosack: probably better to compare to a small epsilon value instead of 0.0, e.g. with `np.isclose(length, 0.0)`

kosack:
The `recalc_pix` mechanism to avoid recalculation is not necessary. One could just break it into two funcitons and add an `@lru_cache` decorator to the one that does the static pixel calculation, and it will automatically be cached, even for multiple camera geometries.

ctapipe/reco/hillas.py
mdpunch: Maybe, but really the question is to avoid a division by exactly zero, which might give a NaN.
If the pixel values are such that the denomator value is very small, I don't think it will be a problem (since the numerator value will be small too).
But, if it's a generally accepted way to go, sure.

mdpunch:
@kosack 
For using @jit for speed, I tried this but with very strange results.  In some cases, the routines give errors when I run, complaining about undefined variables.  It seems that I can't have @jit before more than one routine at a time?

mdpunch:
@kosack 
Concerning your comment (why can't I reply above?):
> The recalc_pix mechanism to avoid recalculation is not necessary. One could just break it into two funcitons and add an @lru_cache decorator to the one that does the static pixel calculation, and it will automatically be cached, even for multiple camera geometries.
Sounds good!  But what happens when there is a loop through an event, and the camera switches from say LST to GCT?  How can the @lru_cache mechanism keep separate caches per geometry?

kosack:
> Sounds good! But what happens when there is a loop through an event, and the camera switches from say LST to GCT? How can the @lru_cache mechanism keep separate caches per geometry?

That should work automatically since the input (pix_x, pix_y) will no longer be the same so it will cache a new version. If you switch back to an LST it will still have the LST version cached as well (the default is up to 256 inputs cached).  So if you have something like:

```python
@lru_cache
def precalculate(px, py):
    # do stuff
    return precalculated_stuff

def hillas_parameters(px,py,signal):
    precalcualted_stuff = precalculate(px,py)
    ...
```

precalculate will only be called once the first time hillas_parameters is called with a new set of px,py.

kosack:
By the way, when testing you can also generate a "jit" and non-jit verson of a functoin to compare speed. Instead of adding the `@jit` decorator, just call it as a function like this:

```python
hillas_3_jit = jit(hillas_parameters_3)

%timeit hilllas_parameters_3(px,py,s)
%timeit hillas_3_jit(px,py,s)
```

mdpunch:
- Wow, the `@lru_cache` does sound pretty intelligent.  So, seems ideal for this, and so we could drop the `reCalcPix` boolean option.  Still not sure that we want to get rid of the option to not calculate higher moments, though.
- I'll look at the jit timings with new functions, it'll be easier than adding `@jit`. (for the debugging, I discovered the ```
%load_ext autoreload
%autoreload 2
``` magic, so that's good.
- I still can't reply directly to your messages above in a threaded way... is this usual?

PR #294, opened 2017-01-05
Get rid of astropy_helpers

mackaiver:
I've removed all references to the astropy helpers packages as well as the entire submodule. The setup.py is now much simpler than before. The version string will still be automatically generated from the current git commit and the latest git tag. It looks like this:

```
In [1]: import ctapipe

In [2]: ctapipe.__version__
Out[2]: 'v0.2.0.post47+git9028887'
```

This will hopefully make installation a bit easier. I think  the next step would be to remove the ctapipe-extras submodule.


kosack:
On that last point, the astropy doc standards that we are following (http://docs.astropy.org/en/stable/development/docrules.html) use several pre-defined sections: *Parameters*, *Returns*, and *Raises*, *Other Parameters*, and *See Also* that are parsed and displayed in a particular way. This seems to be broken now (tons of `SEVERE: Unexpected section title` errors during sphinx build) .

kosack:
strangely that only happens when running "python setup.py build_sphinx", and there are fewer warnings with `cd docs; make html`. SO maybe we should just get rid of the setup.py way of doing it.

kosack:
Also the sphinx math extension seems to be missing. 

```
 WARNING: using "math" markup without a Sphinx math extension active, please use one of the math extensions described at http://sphinx-doc.org/ext/math.html
```

mackaiver:
Made some more changes. When running `make html` from the docs folder there are now more warnings and the output looks similar to the current docs.

MaxNoe:
I think the big step, is that astropy-helpers is not required anymore for the setup.py itself. This means we can use astropy-helpers everywhere else we need it with

```
setup(
    ...,
    setup_requires=['astropy-helpers==<version>'],
    docs_require=...,
)
```

PR #306, opened 2017-01-19
reworked documentation to render more nicely

tino-michael:


ctapipe/reco/FitGammaHillas.py
```diff
@@ -302,10 +316,11 @@ def fit_origin_crosses(self):
         return linalg.normalise(np.sum(crossings, axis=0)) * u.dimless, crossings
 
     def fit_origin_minimise(self, seed=(0, 0, 1), test_function=neg_angle_sum):
-        '''fits the origin of the gamma with a minimisation procedure this
-        function expects that get_great_circles has been run already a
-        seed should be given otherwise it defaults to "straight up"
-        supperted functions to minimise are an M-estimator and the
+        ''' Fits the origin of the gamma with a minimisation procedure this
+        function expects that
+        :func:`get_great_circles<ctapipe.reco.FitGammaHillas.get_great_circles>`
```
kosack: it's not a problem, but you can also just put back-tics around the function like:

      `func_name` is a link to the function

 and sphinx will automatically detect what type it is and link it. It should look in the current namespace first (e.g. `ctapipe.reco.FitGammaHillas`, but you have to fully qualify things that are outside of course.

tino-michael: hm... I tried a lot of different ways to refer to the function, this was the only way that worked.
When I only used back-ticks, it created a box but not a link.
Maybe my local sphinx behaves differently than the online one?

tino-michael: no, indeed, only back-ticks works, too
good, I'll keep this in mind

kosack: at least it saves some typing :)

PR #327, opened 2017-02-09
added module for PS sensitivity

tino-michael:


kosack:
you should follow the following style guidelines so that this code looks like the rest (this comes from the CTA official style guidelines and PEP8):

 * functions and methods should be all lower-case with underscores
 * classes should be CamelCase (staring with upper-case)
 * module filenames should be all lower-case

cdeil:
Sensitivity computation is something we also wanted to add in Gammapy in the coming weeks (cc @bkhelifi @jjlk). To me it doesn't matter much whether the code is developed first here in ctapipe, we can discuss the best repo to maintain and evolve such code separately.

We do have a (I hope) well-implemented, -documented and -tested version of a significance function which you add here to ctapipe:
http://docs.gammapy.org/en/latest/api/gammapy.stats.significance.html
but the corresponding sensitivity functions aren't done yet:
http://docs.gammapy.org/en/latest/api/gammapy.stats.sensitivity.html
and there's some classes to estimate observation time for CTA (see [here](https://nbviewer.jupyter.org/github/gammapy/gammapy-extra/blob/master/notebooks/cta_simulation.ipynb)) but no sensitivity estimator yet.

@tino-michael - Do you have time to finish this pull request? I could do a review and leave suggestions.

From a very brief look, these would be my suggestions in addition to what @kosack mentioned above:
- Please give some info what this class does (e.g. what method is used to compute sensitivity) in the  docstrings (e.g. the class-level docstring).
- In the tests, add one that exercises a few concrete cases where you put numbers in and assert on the computed sensitivity. In my experience these simple tests are the best to have. If you want to put multiple cases you can use `pytest.parametrize`.
- Don't use `assert` outside tests (in actual code). Instead raise errors (e.g. ValueError for bad input values).

ctapipe/analysis/Sensitivity.py
```diff
@@ -0,0 +1,790 @@
...
+from scipy.optimize import minimize
+import scipy.integrate as integrate
+
+import numpy as np
+
+from matplotlib import pyplot as plt
+
+
+# tau, the superior circle number
+np.tau = 2*np.pi
```
cdeil: Suggest to remove `np.tau` and instead type out `2 * np.pi` everywhere.
The problem with `np.tau` is that if I see it in the code below I'm confused because `tau` doesn't exist in numpy. So `tau = 2 * np.pi` here would be better, but for the few letters `2 * np.pi` and the few uses I don't think the global variable (which makes people scroll up and search when reading the code) is warranted.

ctapipe/analysis/Sensitivity.py
```diff
@@ -0,0 +1,790 @@
...
+
+__all__ = ["SensitivityPointSource"]
+
+
+def convert_astropy_array(arr, unit=None):
```
cdeil: This function is really a one-liner and only called once. Suggest to remove it.
(I can have a closer look and suggest how to write the original function where you call this if there's some problem that makes the code there hard to read).

ctapipe/analysis/Sensitivity.py
```diff
@@ -0,0 +1,790 @@
...
+    -------
+    a : quantified numpy array
+    """
+
+    if unit is None:
+        unit = arr[0].unit
+    return np.array([a.to(unit).value for a in arr])*unit
+
+
+def crab_source_rate(energy):
```
cdeil: Hmm, we have this already in Gammapy:
http://docs.gammapy.org/en/latest/api/gammapy.spectrum.CrabSpectrum.html

@kosack - Did we discuss if `import gammapy` from `ctapipe` is OK yet?
(Clearly my preference would be that we work together and don't duplicate stuff.)
If not possible / desired we can find a sane way to share code without duplicating much effort.

ctapipe/analysis/Sensitivity.py
```diff
@@ -0,0 +1,790 @@
...
+    -------
+    flux : astropy quantity
+        differential flux at E
+
+    '''
+    #return 3e-7 * (energy/u.TeV)**-2.52 / (u.TeV * u.m**2 * u.s)
+    return 3e-7 * (energy/u.TeV)**-2.48 / (u.TeV * u.m**2 * u.s)
+
+
+def CR_background_rate(energy):
```
cdeil: We have this: http://docs.gammapy.org/en/latest/api/gammapy.spectrum.cosmic_ray_flux.html
I didn't check how this compares to your CR background rate.

In any case, ideally there should be a reference to a paper or webpage or something where this spectrum comes from ...

ctapipe/analysis/Sensitivity.py
MaxNoe: Brilliant. Maybe open a PR over at https://github.com/numpy/numpy ?

seriously though: don't monkey patch numpy.

cdeil: 🐵 🙈 🙉 🙊

tino-michael: This was meant of course more as a joke than a serious fix of numpy.
I will change it to use `2*np.pi` inline.

ctapipe/analysis/Sensitivity.py
tino-michael: Well, I do use the function in my other scripts, too, but since I didn't want to add my "helper_functions" collection here as a dependency, I simply copied the function definition.
The problem -- at the position where it is used -- is that `numpy.array` does not like lists of quantities (neither does matplotlib, btw), so the units have to be stripped and appended again to the np.array.
Anyway, I changed `make_mock_event_rate` to use `scipy.integrate` where I have to strip the units already in advance. So I no longer need `convert_astropy_array`.

ctapipe/analysis/Sensitivity.py
tino-michael: [also concerning the comment to `CR_background_rate`]
Of course, these rate definitions here are not intended to be the standard definitions to be used in ctapipe. I just wanted to have something to get things going. Hence also no proper references other than "random plot I found online". I have no problem referring to `gammapy` instead (or some dedicated `ctapipe` flux module if it is not desired to `import gammapy`)

tino-michael:
Other than the fact that this module still needs some clean-up and improved documentation, it is already working in that it produces sensitivity curves.
If your gammapy implementation is still just a skeleton with lots of "TODO"s, we probably should focus on this version here (except if you think that my whole approach is complete nonsense).

cdeil:
@tino-michael - I did not review the method implemented here yet, this would take me a few hours.

IMO the easiest would be if you continue with this PR a bit, and @bkhelifi puts up his PR on sensitivity curve computation in Gammapy (@bkhelifi - timescale in a week or two or longer?) and then we have a short call to discuss the methods / implementation and make the decision what gets merged into ctapipe and gammapy?

I also saw that @jknodlseder did a check on point-source sensitivity computation for CTA with ctools: https://cta-redmine.irap.omp.eu/attachments/download/2164/tn-sensitivity.pdf
It's a different method, so results not directly comparable. But @tino-michael and @bkhelifi - I think it would be great if all of you working on CTA sensitivity computation for now used the same IRFs / obs time and ran your script, so that results are comparable, and soon we have well-debugged and -understood sensitivity curves for CTA using the "new" codes (like ctapipe / gammapy / ctools).

ctapipe/analysis/tests/test_sensitivity.py
```diff
@@ -101,6 +82,28 @@ def test_SensitivityPointSource():
     np.testing.assert_allclose(eff_area_p.value, gen_area_p.value/2)
 
 
+def test_SensitivityPointSource_sensitivity():
```
tino-michael: @cdeil I added a specific test for the sensitivity. Is this what you had in mind?

ctapipe/analysis/tests/test_sensitivity.py
```diff
@@ -0,0 +1,219 @@
...
+    eff_a = Sens.effective_areas
+    eff_area_g, eff_area_p, eff_area_e = eff_a['g'], eff_a['p'], eff_a['e']
+
+    # "selected" events are half of the "generated" events
+    # so effective areas should be half of generator areas, too
+    np.testing.assert_allclose(eff_area_g.value, gen_area_g.value/2)
+    np.testing.assert_allclose(eff_area_p.value, gen_area_p.value/2)
+
+
+def test_SensitivityPointSource_sensitivity_MC():
```
tino-michael: @cdeil I added a specific test for the sensitivity. Is this what you had in mind?

cdeil: @tino-michael - Thanks!

I would suggest to add an assert_allclose for the computed sensitivity.
(not just that it "roundtrips" to the requested significance level)

ctapipe/analysis/tests/test_sensitivity.py
```diff
@@ -0,0 +1,219 @@
...
+
+
+def test_draw_events_with_flux_weight():
+
+    Emin = 20
+    Emax = 50
+    Nbin = 50
+    Ndraws = 10000
+
+    def weight_from_energy(e, gamma_new=3, gamma_old=2):
```
cdeil: Move this function out of the test function? (usually nested functions are only used for closures, which is not the case here) If it's only useful for tests, you can leave it in the test file of course.

ctapipe/analysis/tests/test_sensitivity.py
```diff
@@ -0,0 +1,219 @@
...
+
+    energy_sel_gamma = np.random.uniform(Emin, Emax, 50000)
+
+    indices = SensitivityPointSource.draw_events_from_flux_weight(
+                        {'g': energy_sel_gamma},
+                        {'g': weight_from_energy(energy_sel_gamma, 2, 0)},
+                        {'g': Ndraws})
+
+    assert len(energy_sel_gamma[indices['g']]) == Ndraws
+    # TODO come up with more tests...
```
cdeil: I'd suggest to always seed the RNG if you have tests that use RNG, so that results are always the same.
(tests where the results change on every run are painful to maintain and debug.)

Then when you always have the same result, you can actually assert on some number from your output, not just have these high-level consistency checks. In my experience there are always bugs and changes in method and code refactorings, and having tests with asserts on the actual results that come out of the functions is useful (even if now just establishing the status quo, not somehow "verified" numbers)

ctapipe/analysis/tests/test_sensitivity.py
```diff
@@ -0,0 +1,219 @@
...
+    # be the scaling factor that needs to be applied on Nsig to produce a 5 sigma result
+    # in the lima formula
+    ratio = sensitivity["Sensitivity"][0]/(crab_source_rate(1*u.TeV)).value
+    np.testing.assert_allclose([sigma_lima(ratio*Nsig+alpha*Nbgr, Nbgr, alpha)], [5])
+
+
+def test_SensitivityPointSource_sensitivity_data():
+
+    alpha = 1.
+    Non = 30
```
cdeil: Suggest to always follow the Python style guide, i.e. use `n_on` and not `Non`.
There's no need to do this via code review. E.g. [flake8](http://flake8.pycqa.org/en/latest/) will list you all of these superficial coding style violations.

ctapipe/analysis/sensitivity.py
```diff
@@ -0,0 +1,733 @@
...
+        calc_sens_errors : bool, optional (default: False)
+            calculates statistical error for every sensitivity point
+
+        Returns
+        -------
+        sensitivities : astropy.table.Table
+            the sensitivity for every energy bin of `sensitivity_energy_bin_edges`
+        """
+
+        assert sensitivity_energy_bin_edges is not None, \
```
cdeil: Never use assertions (except in tests), always raise errors.

ctapipe/analysis/sensitivity.py
```diff
@@ -0,0 +1,733 @@
...
+class SensitivityPointSource():
+    """
+    class to calculate the sensitivity to a known point-source
+    TODO extend sensitvity module
+        • add extended source?
+        • add pseudo experiment for low exposure times?
+    """
+    def __init__(self, mc_energies=None, energy_bin_edges=None,
+                 energy_unit=u.TeV, flux_unit=1/(u.TeV*u.m**2*u.s)):
+        """
```
cdeil: At least in Astropy and Gammapy, docstrings are always class-level and the `__init__` docstring is left empty.
I suspect (but didn't check) that the same pattern is used in ctapipe.
So suggest to move the content of this docstring to the class-level docstring.

ctapipe/analysis/sensitivity.py
```diff
@@ -0,0 +1,733 @@
...
+    return (sigma-X)**2
+
+
+#   ######  ##          ###     ######   ######
+#  ##    ## ##         ## ##   ##    ## ##    ##
+#  ##       ##        ##   ##  ##       ##
+#  ##       ##       ##     ##  ######   ######
+#  ##       ##       #########       ##       ##
+#  ##    ## ##       ##     ## ##    ## ##    ##
+#   ######  ######## ##     ##  ######   ######
```
cdeil:
```
 ('-.                                .-. .-')     ('-.   ,------.  
  ( OO ).-.                            \  ( OO )  _(  OO) '  .--.  ' 
  / . --. /           ,--. .-'),-----. ,--. ,--. (,------.|  |  |  | 
  | \-.  \        .-')| ,|( OO'  .-.  '|  .'   /  |  .---''--'  |  | 
.-'-'  |  |      ( OO |(_|/   |  | |  ||      /,  |  |        __.  | 
 \| |_.'  |      | `-'|  |\_) |  |\|  ||     ' _)(|  '--.    |   .'  
  |  .-.  |      ,--. |  |  \ |  | |  ||  .   \   |  .--'    |___|   
  |  | |  |      |  '-'  /   `'  '-'  '|  |\   \  |  `---.   .---.   
  `--' `--'       `-----'      `-----' `--' '--'  `------'   '---'   
```

ctapipe/analysis/sensitivity.py
```diff
@@ -0,0 +1,733 @@
...
+class SensitivityPointSource():
```
cdeil: Remove `()`

ctapipe/analysis/tests/test_sensitivity.py
tino-michael: good point

ctapipe/analysis/tests/test_sensitivity.py
tino-michael: I don't think that in the current state a function is even necessary... (previously, I wanted to give the callable to the `draw_events_from_flux_weight` method but now it takes the weights-array immediately)
I will put the calculation in-line

cdeil:
FYI: @bkhelifi and @jjlk have implemented a point source sensitivity estimate class in Gammapy in https://github.com/gammapy/gammapy/pull/1053 , it's going to go in soon.

@kosack - What's the status here? Do you want this in ctapipe? Would it help if I do another review?

My recommendation would be that we merge both PRs, to make it easy for people to use it and to compare results, with the understanding that later on we might want to refactor / merge the codes which to a certain degree are doing the same thing, i.e. debugging / extending / maintaining two versions is non-ideal.

kosack:
@cdeil  I agree we should merge it, though it may need one more review and currently the tests fail, so it needs some work. The new policy is that merges are not allowed until the Travis reports a green status

kosack:
The build seems to get killed by Travis in one of the sensitivity tests, perhaps due to too much memory usage. Is it possible it uses more than 3GB?

tino-michael:
when I run the functions locally (not the py.test), I got this error:
```Intel MKL FATAL ERROR: Cannot load libmkl_avx2.so or libmkl_def.so.``` 
after a quick google and reinstalling numpy, everything works fine

kosack:
`test_draw_events_from_flux_histogram` seems to be where the memory exceeds limits (I suspect it is because of previous tests that load pyhessio, which uses >2 GB of ram, so there is little left). Just reducing the number of bins in the histograms should probably work.

cdeil:
@kosack - Can you make it appear in the docs, please?

Not there at the moment:
https://cta-observatory.github.io/ctapipe/search.html?q=SensitivityPointSource&check_keywords=yes&area=default

PR #340, opened 2017-02-28
Zfits reader cocov

cocov:
Modified the io/zfits such that it works with latest ctapipe container structure and that it match the CameraToACTL zfits structure.
- added a sample file of 5 events
- added a unit test

Note that for ctapipe installation procedure, Etienne Lyard now produce a way to install the ProtoZFitReader from CameraToACTL through pip such that it would become easy for anyone who need to work with zfits input file to use ctapipe

Some TODO: not all the fields which can be implemented in zfits file are read yet, they will be implemented as they appear in the protozfitreader.


cocov:
One comment: I added some fields in the io.containers.RawCameraContainer class, let me know if this is not the proper approach, I could create a new class if needed

ctapipe/io/containers.py
```diff
@@ -75,6 +75,10 @@ class RawCameraContainer(Container):
     """
     Storage of raw data from a single telescope
     """
+    event_number = Item(-1, "telescope event number")
+    num_channels = Item(-1, "number of channels per pixel")
```
watsonjj: num_channels and num_pixels are filled in the InstrumentContainer as these values do not change per event.

Event_id exists in RawDataContainer. Is event_number intended to be different to this value? 

pixel_flags is new to me, what is it's purpose? could you add a description?

cocov: Thanks for your prompt comments.
Concerning the field appearing in the InstrumentContainer, I though this container will be deprecated, but indeed if it is not I can use it.

Pixel flag is indeed a specific to our camera. I just realise that, that would probably mean I should create camera specific DataContainer class. In our case it gives the status of each pixel in a given event. What do you think of this? When I look in the DataModel of CamerasToACTL there is indeed fields which are different from one camera to the other, which we should be able to readout. Maybe that would just be for an expert version (and correspond more to the R1 data format)

Concerning the event_number and event_id, that's a good question. To me right now it is the event counter from the DigiCam I am testing. I don't think that there is a mechanism that allow this event counter to be the same than the future overall cta event number. Since the Event_id in RawDataContainer counts for all telescope I was thinking it would be usefull to have the telescope "local" counter in the RawCameraContainer. But maybe if we go for camera specific container as discussed before I should put it there.

Let me know what you think about it

watsonjj: I believe this brings up some good points that need to be addressed. There are some properties that are specific to the GCT camera for which I created a new container which inherits from RawCameraContainer. I then create a new DataContainer class which uses the new containers, and fill this data container in my io file (analogous to hessio.py).

I do this in a separate project to ctapipe for the moment, which is unfortunately inside subversion instead of github so you cannot see what I mean. However I think this is the same idea you have described above.

@kosack is this how we should approach this for different cameras? and should these new containers all be inside containers.py?

kosack: >Concerning the field appearing in the InstrumentContainer, I though this container will be >deprecated, but indeed if it is not I can use it

The main reason we say that InstrumentContainer will be deprecated is that it contains info that will not be in the data files, but rather in a central database with a separate API.  Anything that describes the instumental layout or configuration will be in this database, and not repeated in each data file as it is now.  For now, it's fine to keep using the insturment container until this new API is ready, though, and it may even help us to not miss something useful.

kosack: Also, it's true that the R0 formats may be quite different per telescope, only DL0 should be common.

cocov: > For now, it's fine to keep using the insturment container until this new API is ready, though, and it may even help us to not miss something useful.

Ok, I'll use it then

cocov: > 2. use sub-classes to extend Containers to add instrument-specific info.
I quite like this solution. I'll look into it

cocov: I re-attributed the variables to the proper containers and added a DigiCamRawCameraContainer class deriving from RawCameraContainer in io/containers to treat the camera dependent info. 
 @kosack In the zfits.py I simply overwrite the RawCameraContainer items of the map in the RawDataContainer tel attribute.Does that sounds fine?

cocov:
I re-attributed the variables to the proper containers and added a DigiCamRawCameraContainer class deriving from RawCameraContainer in io/containers to treat the camera dependent info.
@kosack In the zfits.py I simply overwrite the RawCameraContainer items of the map in the RawDataContainer tel attribute.Does that sounds fine?

cocov:
I realise also that just before I did the pull request, R0,R1 were introduced in the zfits.py
I'll adapt to it

kosack:
Is this PR ready to merge? It now says there is a conflict in zfits.py that needs to be resolved, so you may need to pull once more from the upstream repo and commit it.

cocov:
@kosack  sorry for the delay, I was offline part of last week. I've commit a version with solved conflict

mackaiver:
The protozfitreader dependency should be added to the setup.py if its pip installable.

dneise:
I think this is superseded by #598 and can be closed.

PR #341, opened 2017-02-28
Charge resolution component

watsonjj:
Converted the charge resolution class to a Component, and updated the examples  of obtaining the charge resolution to be Tools, and also split into two parts: 

1 - extracting the charge resolution numpy array and saving the ChargeResolutionCalculator object into a pickle file
2 - plot multiple charge resolution numpy arrays by opening multiple ChargeResolutionCalculator pickle files


kosack:
I would suggest moving this code (and also the camera calibration) move from examples to ctapipe/tools, and add an entry_point in setup.py for them.  That way they become part of the command-line suite like `ctapipe-info`, `ctapipe-camdemo`, etc...

After all, they seem to be more than just simple examples, and fully working executables.

PR #354, opened 2017-04-04
Additional dl1

watsonjj:
Created additional functionality to the dl1 calibrator:
- Can now pass a cleaning method from waveform_cleaner.py to a dl1 calibrator 
- Created the extraction technique "AverageWfPeakIntegrator
- Cleaned up charge_extraction.py to be easier to follow


kosack:
We really need to do a full code review and clean up a lot of this - things like waveform cleaning should not be in calibration, but in `image`, since they are related to processing image cubes, not calibrating data. But for now, I think we can leave it here - i think the whole calibration module will be restructured in the future and the more general algorithms moved outside.

PR #359, opened 2017-04-05
Option for passing adc2pe file for r1 calibration

watsonjj:


kosack:
Looks fine - is the expected format of the file in the documentation? I guess it is something that can be loaded with `astropy.table`?

watsonjj:
The format of the file is very flexible - the classes that inherit from CameraR1Calibrator either define how the files are used, or do not use them at all (such as HessioR1Calibrator). For example, in my external test-bench project targetpipe, I have a TargetioR1Calibrator which loads these files (which are fits files in this case) into TargetCalib, our c++ calibration library. 

As a side note - The inplementation of the adc2pe file does not yet exist in TargetCalib, and so instead this is a numpy file, which is handled by TargetioR1Calibrator. But when we do implement this in TargetCalib, the user's higher level interface doesn't change.

I'm not aware of plans to have a common file type at this level, as this is the R1 calibration that will be applied by each camera team before storing as dl0.

Issue #346, opened 2017-03-15
proposal to speed up algorithms that use pixel neighbor lists


kosack:
I did a study of why the image cleaning algorithm (tailcuts) is slower than I would like, and found the issue is how the neighborlist is stored and applied.  The following shows several implementations of the tailcuts algorithm, using different representations of the neighbor list.  (conclusion below)

![py36](https://cloud.githubusercontent.com/assets/11677812/23952037/6590128e-0990-11e7-800c-bf8d14c2ff9f.png)

The conclusion is that using a 2D boolean matrix for the neighbors and *numpy* operations is significantly faster than even a simple for-loop based C implementation (the *cython-opt* method is equivalant to a c for-loop method with compiler optimizations turned on). The reason is likely the vectorizability of operations and modern CPU features.  Using the following neighbor matrix instead of an adjacency list of pixel-ids seems to be the fastest improvement:

![unknown](https://cloud.githubusercontent.com/assets/11677812/23952136/bce23044-0990-11e7-935b-3b672ac4703d.png)

Using this will likely also allow for speedups of other algorithms.  I plan on implementing it directly in the code in a future pull-request.

Full study here:
https://github.com/kosack/ctapipe/blob/improve_cleaning_speed/examples/notebooks/speed_up_image_cleaning.ipynb

kosack:
@watsonjj this could likely also speed up parts of the calibration that use nearest-neighbors. Though, it means some re-structuring of code that uses them.

watsonjj:
I've seen this, and will test/implement it in the charge extraction techniques as soon as I have time.

kosack:
Great, I'll modify the `CameraGeometry` class and the `image` module cleaning functions to use it, then you don't have to do much other than update your algorithms.

kosack:
@watsonjj once  PR #360 is merged, you can use the `geom.neighbor_matrix` attribute to speed up your algorithms (see `image.tailcuts_clean` and `image.dilate` for examples).  It may also simplify some other code, since you no longer need to cache the camera geometries youself, you can just keep loading or calling `.guess`, which is much faster now and automatically caches previously constructed geometries.

watsonjj:
@kosack I will try to look at this today. A problem I previously had with caching was if I was working with multiple files, as I was caching by telescope index. This is much nicer!

PR #360, opened 2017-04-05
Refactor and improve camera geometry and image cleaning

kosack:
Improvements
---------------

- based on speed-ups in issue #346
- moved `CameraGeometry` from `io` to `instrument`, and replaced the duplicate functionality in instrument
- sped up `CameraGeometry.guess()`, which now uses a cache to no longer recompute the same geometry multiple times
- neighbor lists are now lazily evaluated, for speed
- now provides a neighbor_matrix in addition to the neighbor list, for faster algorithms
- speed up `image.tailcuts_clean` and `image.dilate` by using `CameraGeometry.neighbor_matrix` (speed is faster than standard c implementation!) 

API changes
-------------
-  `image.tailcuts_clean` no longer takes a `pedvars` argument, which was redundant in most cases 
-  all references to `ctapipe.io.CameraGeometry` should now be c `ctapipe.instrument.CameraGeometry`


kosack:
@watsonjj  do you see any problems with this, or rather, does it break anything major in your code?

ctapipe/instrument/camera.py
```diff
@@ -33,19 +33,35 @@
 }
 
 
+_geometry_cache = {}
```
watsonjj: Is this an erroneous declaration?

kosack: no, it's used later in the guess() method. It's a class attribute to keep the list of previously constructed geometries, which are keyed by an integer hash that combines the x,y vectors and the focal length (which I think is enough to make them unique).  Originally, I had used only the number of pixels, but if we use the hexagonal-to-cartesian camera transforms, the number of pixels stays the same but the pixel ordering changes, so it caused some problems.

kosack: I'll add a comment to that line

watsonjj: But there are 2 `_geometry_cache = {}`, one global on line 36, and one as an attribute of CameraGeometry on line 60, are both required?

kosack: ahh, yes that is a mistake. The global one is not needed. I'll remove it

kosack:
if there's no major issue, I will merge this soon - I'd like to work on fixing a few other bugs/features that depend on this .

watsonjj:
I'm happy for this to be merged. I should implement the use of the neighbour_matrix in the charge extraction algorithms, correct? 

Is the old neighbour list going to be deprecated eventually? I see that neighbour_matrix relies on that list for its creation, is that still the best way for it to be created?

kosack:
> I'm happy for this to be merged. I should implement the use of the neighbour_matrix in the charge extraction algorithms, correct?

Yes that would be great. I already replaced a call in   `dl1.py` that calculated the neighbors again with a call to `CameraGeometry.neighbors`, and that simple change sped up the code (at least by a few seconds of runtime), now that CameraGeometry is cached. It still uses the neighbor list and not the matrix though, so changing the integrators to use it would improve the speed even more.

>Is the old neighbour list going to be deprecated eventually? I see that neighbour_matrix relies on that list for its creation, is that still the best way for it to be created?

I left both in there for now to not break everything, but we could eventually deprecate the neighbor list if it is not used. It could be useful in some cases.  It does not need to be created first in order to make the neighbor_matrix, I just did it that way for simplicity (the matrix could easily be generated directly).  For now, I think we can leave both, and recommend the matrix for fast calculations.  I think it may be nicer to have the list in case we want to write it to a  human-readable file though.

ctapipe/calib/camera/dl1.py
```diff
@@ -176,7 +176,7 @@ def check_dl0_exists(self, event, telid):
                 self._dl0_empty_warn = True
             return False
 
-    def get_neighbours(self, event, telid):
+    def get_geometry(self, event, telid):
         """
         Obtain the neighbouring pixels for this telescope.
```
MaxNoe: This is now lying documentation

watsonjj: I will fix that in my next PR

watsonjj:
@kosack Is it possible to change CameraGeometry to a Component? This would allow the radius size for the pixel neighbours to be changed, but I'm not sure if it makes sense to change CameraGeometry like that. Previously this could be changed by the radius traitlet in dl1.py.

PR #361, opened 2017-04-07
replace progress bar from tqdm package with astropy one to avoid extra dependency

kosack:
the new tools for charge extraction added a dependency on an external package `tqdm` for rendering a text progressbar. This PR replaces it with `astropy.utils.console.ProgressBarOrSpinner`, which has similar functionality without the added dependency

### DIFF
--- a/ctapipe/tools/extract_charge_resolution.py
+++ b/ctapipe/tools/extract_charge_resolution.py
@@ -8,7 +8,7 @@ from ctapipe.calib.camera.dl0 import CameraDL0Reducer
 from ctapipe.calib.camera.dl1 import CameraDL1Calibrator
 from ctapipe.calib.camera.charge_extractors import ChargeExtractorFactory
 from ctapipe.analysis.camera.chargeresolution import ChargeResolutionCalculator
-from tqdm import tqdm
+from astropy.utils.console import ProgressBarOrSpinner
 
 
 class ChargeResolutionGenerator(Tool):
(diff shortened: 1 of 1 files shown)


watsonjj:
I can understand we should avoid extra dependencies, but I am still an advocate of tqdm. It is an actively developed module, and is optimised to produce minimal overhead (60 ns) when displaying. It also features the ability to have layered progress bars if you needed them for nestled loops, and they are developing functionality to have multiple progress bars if you have multiple threads.

Looking online, I see a lot of related discussion about deprecating the astropy ProgressBar and replacing it with tqdm, and whether dependency for the sake of modularity is better.

MaxNoe:
:+1: for tqdm

kosack:
I'm not against adding it, but I want to keep dependencies to a minimum and tqdm is not even part of the standard anaconda distribution (though there is a package for it). 

Since we don't use it much yet, I'd say stick with astropy's tool for the current release (it has the same interface), and then if we find we need something better, we can use it (particularly if it becomes an astropy dependency).  Generally all scripts we run will be on a cluster anyhow, so we don't really need a progress bar except for some occasional user scripts. I'd prefer to keep things as simple as possible so far.  

We may on the other hand want something like a progress bar that can at regular intervals (not tied to the frequency it is called), output a status message to the logging system during long tasks.  E.g. we'd want a log that looks like:

```
2017-01-01T17:55:01: INFO: starting task x for 10000 events
2017-01-01T17:55:05: INFO: x is now at element 500/10000 (5%) rate=125 Hz
2017-01-01T17:55:10: INFO: x is now at element 1456/10000 (14%) rate=120 Hz
2017-01-01T17:55:15: INFO: x is now at element 1611/10000 (16%) rate=125 Hz
...
```

Is that something that can be achieved with tdqm?

kosack:
ok, I do like some of the features of tqdm and it seems they are adding logger support (At least there is an issue with a PR for it), so I think it's ok to have as a dep.  Could you open a new PR though and make sure you have updated `setup.py`, `environment.yml` and in the cta-conda-recipes repo, the list of depencies to install?  That all must be done in order to add a dependency - right now new users get a crash when using some parts of ctapipe, which is bad.

kosack:
For now, I will switch to the Astropy version, since I need a non-crashing version of ctapipe, but we can switch back to tqdm once the appropriate dependencies have been added.

PR #362, opened 2017-04-12
HDF5 pytables input/output from Containers

kosack:
This is an initial implementation of a simple I/O  system that takes Containers and creates HDF5 tables where one row in the table contains the container. It is intended as a building-block to make a more fancy I/O system or just as a technology study (there may be better ways to do it).  

* elements of a Container are translated to columns in the output table automatically
* the user can assign a transformation function to each column that is called on the value before writing it
* it does not yet use the common Serializer class yet, since I didn't quite like the interface (needs some review)
* it does *not* recurse into sub-containers, though this could be done. 

Use Case
----------

Pipeline needs a way to efficiently store data after calibration (DL1), DL2 event lists, etc.

Usage
-------

```python
    # make some dummy data in containers
    r0tel = R0CameraContainer()
    mc = MCEventContainer()
    mc.reset()
    r0tel.adc_samples = np.random.uniform(size=(50,10))
    r0tel.adc_sums = np.random.uniform(size=50)
    r0tel.num_samples = 10
    r0tel.meta['test_attribute'] = 3.14159
    r0tel.meta['date'] = "2020-10-10"

    # write the data to a file
    writer = SimpleHDF5TableWriter('test.h5', group_name='R0')

    for ii in range(100):
        r0tel.adc_samples[:] = np.random.uniform(size=(50,10))
        r0tel.adc_sums[:] = np.random.uniform(size=50)
        r0tel.num_samples = 10
        mc.energy = 10**np.random.uniform(1,2) * u.TeV

        writer.write("tel_001", r0tel)
        writer.write("tel_002", r0tel)  # write a second table too
        writer.write("MC", mc)

```

Which generates:
hdf5writer

TODO
------
* allow multiple containers to be joined into a table 
* either add `event_id` to each container, or make a way to add a _primary key_ when calling `write()`
* implement reading back into containers
* figure out what to do with things like `tels_with_data`, which are var-length lists and not easily supported (convert to fixed-length masks?)
* unify with Serializer base class, so we can implement other ones as well and speed test


kosack:
Reading now works like this:

```python
reader = SimpleHDF5TableReader("test.h5")
mc = MCEventContainer()

for cont in reader.read("/R0/MC", mc):
    print(cont)
```

```
{'alt': 0.0,
 'az': 0.0,
 'core_x': <Quantity 0.4060639535514794 m>,
 'core_y': <Quantity 0.5495818276293845 m>,
 'energy': <Quantity 77.34112211504227 TeV>,
 'h_first_int': 0.0,
 'tel': {}}
{'alt': 0.0,
 'az': 0.0,
 'core_x': <Quantity 0.2265416589346445 m>,
 'core_y': <Quantity 0.9771415161022963 m>,
 'energy': <Quantity 94.48982399127513 TeV>,
 'h_first_int': 0.0,
 'tel': {}}
{'alt': 0.0,
 'az': 0.0,
 'core_x': <Quantity -0.9818453098851436 m>,
 'core_y': <Quantity 0.8782971828013879 m>,
 'energy': <Quantity 29.2428288115323 TeV>,
 'h_first_int': 0.0,
 'tel': {}}
```


Right now both `cont` and `mc` above are pointers to the same structure (like in `hessio_event_source`, no copying is done), but we can discuss that. Also, multiple tables can be read at once by getting several generators.

kosack:
Since this doesn't interfere with anything, and I want to use it during the next workshop, I will merge it. The API may of course change

PR #368, opened 2017-04-21
Removed ctapipe-extras as submodule

kosack:
- no longer need to set `CTAPIPE_EXTRA_DIR` env variable
- resources are stored in a module called `ctapipe_resources` in the *ctapipe-extras* package (just do `conda install ctapipe-extras` to get it
- uses python package resources to be able to find the sample data files
- updated `utils.datasets.get_dataset` and examples/tests that use it to work with the new system, so no change to existing code should be needed (unless the code used the path to the resources and not the resource filename)
- deprecated `get_path()`, replaced by `get_dataset`, since it no longer is a "path", but the full filename

using:
```python
from ctapipe.utils.datasets import get_dataset

testfile = get_dataset("gamma_test.simtel.gz")
```
should work exactly as before.

You can also get to ctapipe resource files with:

```python
import ctapipe_resources

testfile = ctapipe_resources.get("gamma_test.simtel.gz")

# or

testfile = ctapipe_resources.test_events_file
```


kosack:
Note this breaks some things in the tests of `io.EventFileReader`, which seem to require a base path plus a filename.  Since the resources are not guaranteed to be in any particular directory, you can only get the full pathname to the resource, so the test should be updated.  @watsonjj, I don't understand how the EventFileReader works, so this may need some looking into by you (probably a simple fix)

PR #384, opened 2017-04-28
ImPACT dev version

ParsonsRD:
First pass at the implementation of ImPACT reconstruction. Includes file for interpolation of ImPACT templates and fitting. Also contains an array viewer.


ctapipe/reco/ImPACT.py
```diff
@@ -0,0 +1,661 @@
...
+
+    Because this application is computationally intensive the usual advice to use astropy units
+    for all quantities is ignored (as these slow down some computations), instead units within the
+    class are fixed:
+    - Angular units in radians
+    - Distance units in metres
+    - Energy units in TeV
+
+    """
+    def __init__(self, fit_xmax=True, root_dir="/Users/dparsons/Documents/Unix/CTA/ImPACT_pythontests/"):
```
kosack: may not be the best default for root_dir  :-) . Maybe default to "."?

ctapipe/reco/tests/test_ImPACT.py
```diff
@@ -0,0 +1,130 @@
...
+from ctapipe.utils.datasets import get_path
+import astropy.units as u
+from numpy.testing import assert_allclose
+from ctapipe.io.containers import ReconstructedShowerContainer, ReconstructedEnergyContainer
+
+
+class TestImPACT():
+
+    @classmethod
+    def setup_class(self):
```
kosack: I think this can just be `def setup()`, which is called automatically by pytest

PR #397, opened 2017-05-03
Reading of dst files

watsonjj:
Created dstio.py to read dst files into dl1
Created a DstioEventFileReader
Created a DstioR1Calibrator, which does nothing, except holds a staticmethod to handle the calibration into pe when dst files are read
Small optimises to hessio.py
Created empty test scripts for the new dst functionality. Before tests can be made, a test dst files needs to be generated

### DIFF
--- a/ctapipe/io/dstio.py
+++ b/ctapipe/io/dstio.py
@@ -1,6 +1,6 @@
 # Licensed under a 3-clause BSD style license - see LICENSE.rst
 """
-Components to read HESSIO data.  
+Components to read HESSIO DST data.  
 
 This requires the hessio python library to be installed
 """
@@ -12,6 +12,8 @@ from ..core import Provenance
 from astropy import units as u
 from astropy.coordinates import Angle
 from astropy.time import Time
+import numpy as np
+from ctapipe.calib.camera import DstioR1Calibrator
 
 logger = logging.getLogger(__name__)
 
@@ -27,13 +29,13 @@ except ImportError as err:
     raise err
 
 __all__ = [
-    'hessio_event_source',
+    'dst_event_source',
 ]
 
 
-def hessio_get_list_event_ids(url, max_events=None):
+def dst_get_list_event_ids(url, max_events=None):
     """
-    Faster method to get a list of all the event ids in the hessio file.
+    Faster method to get a list of all the event ids in the dst file.
     This list can also be used to find out the number of events that exist
     in the file.
 
@@ -58,21 +60,21 @@ def hessio_get_list_event_ids(url, max_events=None):
             event_id_list = []
             eventstream = pyhessio.move_to_next_event()
             for event_id in eventstream:
-                event_id_list.append(event_id)
-                counter += 1
-                if max_events is not None and counter >= max_events:
-                    pyhessio.close_file()
-                    break
+                if len(pyhessio.get_teldata_list()) > 0:
+                    event_id_list.append(event_id)
+                    counter += 1
+                    if max_events is not None and counter >= max_events:
+                        pyhessio.close_file()
+                        break
             return event_id_list
     except HessioError:
         raise RuntimeError("hessio_event_source failed to open '{}'"
                            .format(url))
 
 
-def hessio_event_source(url, max_events=None, allowed_tels=None,
+def dst_event_source(url, max_events=None, allowed_tels=None,
                         requested_event=None, use_event_id=False):
-    """A generator that streams data from an EventIO/HESSIO MC data file
-    (e.g. a standard CTA data file.)
+    """A generator that streams data from an dst MC data file.
 
     Parameters
     ----------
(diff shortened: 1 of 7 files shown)


ctapipe/io/dstio.py
```diff
@@ -0,0 +1,266 @@
...
+
+                if max_events is not None and counter >= max_events:
+                    pyhessio.close_file()
+                    return
+    except HessioError:
+        raise RuntimeError("dstio_event_source failed to open '{}'"
+                           .format(url))
+
+
+def _fill_instrument_info(data, pyhessio):
```
kosack: why can't use use the same method from hessio.py? The format is the same, so it is redundant to duplicate it here.

ctapipe/io/eventfilereader.py
```diff
@@ -341,6 +342,79 @@ def read(self, allowed_tels=None, requested_event=None,
         return source
 
 
+class DstioFileReader(EventFileReader):
```
kosack: following the software standards naming conventions, I'd suggest EventIODSTFileReader, or DSTIOFileReader (abbreviations and acronyms should be upper-case)

ctapipe/io/dstio.py
```diff
@@ -0,0 +1,266 @@
...
+    max_events : int, optional
+        maximum number of events to read
+
+    Returns
+    -------
+    event_id_list : list[num_events]
+        A list with all the event ids that are in the file.
+
+    """
+    logger.warning("This method is slow. Need to find faster method.")
```
kosack: This is a function we cannot guarantee will work anyhow, since it only works if the file is fully available and is not a stream. Even on the cluster, we may use some form of streaming IO like XROOTD, or similar, so may not always be a way to build such a list.   In other file format readers, this is referred to as building an index table, so we might want to rename the function to something like "build_event_index", and note that it only works if it is possible to loop multiple times over the input, and should not be relied upon.   In fact, in some formats like HDF5 (which we'll probably use for intermediate outputs), such an index can be built automatically by the hdf5 library.

kosack:
It would be good to also include a (small <20MB) test DST file. Is there one we could use? I can add it to `ctapipe_resources` and it can be used to test that this works.

ParsonsRD:
To be honest I'm not sure we need a separate DST reader, by default the pyhessio code will already read Konrad style DSTs fine.

watsonjj:
Sure, pyhessio does the reading. But DstioEventFileReader makes sure its read into the correct data level container (dl1 instead of r0).

kosack:
This does imply that we need to generalize the "event source" code at some point, and probably turn it into a base class with a well-defined interface, so all event_sources have the same functionality, rather than a single function.  In any case, I think Jason's solution is ok for now, which is to use a mixture of an `event_source` + the `EventFileReader` classes to solve the problem.

ctapipe/io/dstio.py
kosack: also, now that code has changed to fill in the SubarrayDescription class, so this code no longer works. It's better to just do `from hessio.py import _fill_instrument_info` and use it

ParsonsRD:
But what I mean is the code (pyhessio and ctapipe) will read and work with the level 0 DSTs that Konrad produces already with no problems. The Code just treats them like an ASTRI telescope with no trace and the signal already integrated. Every result I've showed so far has been using the DST0 files with no issue whatsoever, but it's much faster as we don't have to perform the signal integration step.

kosack:
Ah, ok, so I guess just running the `CameraCalibrater.calibrate()` basically does nothing but apply coefficients and put the final image in the right place? In that case, perhaps the DSTEventFileReader is really not needed...  Is the definition of "dst" in the monte-carlo basically just "trace-integrated"? or is there more processing applied?

watsonjj:
Yes it all works in its current state, and you are right they are essentially treated as ASTRI data. However there are two reasons why we still need this DstioEventFileReader.

Currently the default r1 calibration applied divides the pedestal by the number of samples. Unless the number of samples is set to 1 in the dst values, this will result in an incorrect pedestal subtraction.

Secondly when performing CameraCalibrater.calibrate(), charge integration **will** be attempted. And I believe an integration correction (based on the reference pulse) will be applied to this "one-value" integration. If the dst is read directly into dl1, then CameraCalibrater.calibrate() will literally do nothing, it will be faster, and it won't risk a incorrect alteration to the value.

On another note, to handle ASTRI, it might be right to automatically skip the dl1 integration altogether if there is only one sample in the waveform....

kosack:
> Currently the default r1 calibration applied divides the pedestal by the number of samples. Unless the number of samples is set to 1 in the dst values, this will result in an incorrect pedestal subtraction

Isn't the number of samples == the shape of the array, so it would divide by 1? That is a bit slower than reading it directly, but we should err on the side of less code to maintain for now...  We should not trust the num_samples value given by `pyhessio.get_num_samples()`, since it must by definition be equal to the shape of the array read out (if not there is an inconsistency). I think we had talked before about just getting rid of `adc_sums` altogether and just read everything in to `adc_samples`, which may have a single sample (like with ASTRI). I think right now, astri files get read into adc_sums (just because that's how they are stored in the MC), but that can be corrected in `hessio_event_source`.

 The whole event reading system is going to  need a review and refactoring soon, and already several of the event sources are broken as far as I can tell, so perhaps adding another is not useful at this point.  What do you think? Can we get away with sticking with `hessio_event_source` and just applying the calibration (perhaps modifying it so that it doesn't apply a crazy value?)

watsonjj:
Okay. I feel a litle nervous about it, in case something is applied to the data because it assumes it is a lower level than dl1. But the time being we could get away with it with some minor alterations to the code.

I will create a new PR where no integration or integration correction is applied if there is only 1 sample. Currently ASTRI is read into adc_sums and then in the r1 calibration, if there is nothing in adc_samples, it looks in adc_sums. I will change this logic to occur in `hessio_event_source`, effectively removing the need for adc_sums. Agreed?

watsonjj:
There is one more consideration:
`pyhessio.get_run_number()` crashes on some dst files. @ParsonsRD believes this occurs if the dst file consists of multiple combined runs. But I guess the fix should exist in pyhessio perhaps?

watsonjj:
The changes we have talked about in this PR are handled in #489

PR #403, opened 2017-05-04
Fix impact formating

kosack:
@ParsonsRD  take a look and see if this is ok with you (just an example)

- applied PEP8 and 80-col formatting to ImPACT.py
- made dependency on matplotlib weak (only loaded if a plot function is called)

Note that generally we should *not* plot things inside ctapipe algorithms, only dump output (matplotlib may not be on the grid environment)


ParsonsRD:
Hi @kosack, that looks good to me. Thanks for fixing that!

Yes the matplotlib dependency will go away on the next commit and instead these functions will just return info in the correct format to use in the event viewers.

PR #405, opened 2017-05-09
Table interpolator

ParsonsRD:
This is a modification of the lookup table interpolation class to generalise the the N dimensional interpolation for any table set. In order to do this the file format has been modified to a custom FITS based format with the description of the table dimensions contained within the fits header. Description of the required fits header is contained within the class comments.


kosack:
you might want to merge in the head again. I had moved `TableInterpolator` to utils a while ago

ParsonsRD:
Currently it is structured as a separate HDU for each image template. Now I agree it would be more efficient to store the whole thing as a single data cube (or hypercube I guess :-) ). However currently all entries in the cube are not necessarily filled, so I preferred to keep it as flexible as possible.

PR #410, opened 2017-05-10
rename hessio_event_source to simtelarray_event_source?

kosack:
More efforts to clarify some of the code:
------------------------------------

- to avoid confusion renamed the event source to be more clear about what data it applies to (not only HESS)
- also changed HessioFileReader → SimTelArrayFileReader
- also fixed small bug where the event source generator doesn't exit if
a single event is requested (causing some tests to fail due to an
unclosed file)

### DIFF
--- a/ctapipe/calib/camera/tests/test_calibrator.py
+++ b/ctapipe/calib/camera/tests/test_calibrator.py
@@ -1,13 +1,13 @@
 from ctapipe.calib.camera import CameraCalibrator
-from ctapipe.io.hessio import hessio_event_source
+from ctapipe.io.hessio import simtelarray_event_source
 from ctapipe.utils import get_dataset
 from numpy.testing import assert_allclose
 
 
 def get_test_event():
     filename = get_dataset('gamma_test.simtel.gz')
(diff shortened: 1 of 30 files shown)


watsonjj:
Should all references to "hessio" be changed to "simtelarray"?

kosack:
> Should all references to "hessio" be changed to "simtelarray"?

It's just a proposal, but I guess so.  The confusion is that this is really a reader for EventIO files from SimTelArray, with the particular EventIO dialect being HESSIO (called that for historical reasons). So the original name is not incorrect, but is just perhaps confusing for users who might think it is HESS-specific.  Of course once we have an official new format for CTA, I would guess *simtelarray* will be modified to use it, so then this reader may not be correct anymore (for example we might switch to the updated format called IACTIO, which is not the same as HESSIO, or even something totally different).   For that reason, I wasn't sure of the best name...   We could also go with `eventio_event_source`

kosack:
p.s. I did rename the FileReader subclass in this PR as well, which is now SimTelArrayFileReader

kosack:
I thing in the end we'll keep it as is , to avoid confusion

PR #412, opened 2017-05-19
Created function overlay_moments_update to update the position of the…

watsonjj:
… hillas ellipse without requiring a camera redraw

### DIFF
--- a/ctapipe/visualization/mpl.py
+++ b/ctapipe/visualization/mpl.py
@@ -97,6 +97,8 @@ class CameraDisplay:
         self.geom = geometry
         self.pixels = None
         self.colorbar = None
+        self.ellipse = None
+        self.ellipse_t = None
         self.autoupdate = autoupdate
         self.autoscale = autoscale
         self._active_pixel = None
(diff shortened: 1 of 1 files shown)


kosack:
I'm not sure this will be necessary.  The plan is to remove `overlay_moments` entirely and make it be a separate class of "overlay", so this will not be needed.  The idea is that CameraDisplay will not plot anything but cameras, but you can add an overlay for something like hillas parameters.  That will be much cleaner than having specialized functions:

```python
disp  = CameraDisplay(geom)
hil = HillasOverlay(hillas_parameters, linewidth=3)
disp.add_overlay(hil)  # may also want to specify a name so you can remove it later?
...

# update the display and overlay
disp.image = new_image
hil.data = new_params
```

This can then be used for many types of overlays on a camera.

kosack:
The problem with this implementation is that you can only have one ellipse that gets updated - we need to be able to overplot all hillas ellipses for stereo reconstruction, for example

watsonjj:
Agreed, I also foresaw this being an issue when you want multiple ellipses. However I needed this functionality recently as it would otherwise be slow to loop through a file and draw many images with their ellipses. Would it be possible to merge this function for the time being until the new overlay functionality exists?

Is someone assigned to work on this overlaying? Should I consider starting this?

PR #413, opened 2017-05-21
Waveform cleaner updates

watsonjj:
Updated CHECMWaveformCleaner to have to possible peak finders
Added helper function in charge extractor to quickly obtain window mask


kosack:
can you merge in the master again? That should fix all the tests (we now require Travis to say all tests pass before merging into master)

ctapipe/image/tests/test_waveform_cleaning.py
```diff
@@ -39,8 +39,24 @@ def test_checm_cleaner():
...
+    cleaner = CHECMWaveformCleanerAverage(None, None)
     cleaned = cleaner.apply(data_ped)
 
     assert_almost_equal(data_ped[0, 0, 0], -2.8, 1)
     assert_almost_equal(cleaned[0, 0, 0], -6.4, 1)
+
+
+def test_checm_cleaner_local():
+    telid = 11
+    event = get_test_event()
```
kosack: at some point we should define the test event as a global test fixture, so it is onyl read once (that will speed up the tests, since so many are opening and reading simtel files now). Then you just change your function to be `def test_checm_cleaner_local(test_event)`.  That can be done later though.

Issue #404, opened 2017-05-04
camera names need unification


kosack:
Right now we have the following hard-coded camera names (which are set in `instument/camera.py`) and are also written to `ctapipe_resources`:

- ASTRI (ASTRICam?)
- FlashCam
- NectarCam
- LSTCam
- GCT  (should this be GATE? - it is called both in various places)
- SCTCam (is this the correct name?)
- SST-1m (should be a camera name not a telescope name - is there a correct name? is it DigiCam)
- HESSI - HESS-I cameras

Please suggest updated/better names, since we need to fix them now (the camera name is the "camera_id" used in many parts of the code, and there is now a mismatch e.g. in the muon code)

@AMWMitchell @watsonjj @cocov

watsonjj:
GATE is what it was previously called in the pipeline, but I think it should be GCT. GATE is only the telescope structure, and has nothing to do with the camera...

AMWMitchell:
Hi,

Agreed - it should be either GCT or CHEC - wondered why it was GATE originally. The muon code might need updating to match. 
SST-1M should probably be DigiCam - as far as I’m aware, ASTRI don’t have any special name for their camera, so either ASTRI or ASTRICam is probably fine. 

Cheers,
Alison

GernotMaier:
Is it really necessary to use hardcoded camera names? None of the calibration methods/reconstruction routines depend on the cameras. It is the other way around, and it therefore  seems to be more a configuration issue.

When using name, keep in mind that cameras are still evolving. It is also not guaranteed e.g. that the first GCT camera is the same as the last one (build in 5-6 years? This might matter for the naming, but again it is probably better to have a configuration file ( Tel x -> use methods a,b,c).  

(maybe all of this is already taken into account)

watsonjj:
Gernot has a good point. Even now, there is a difference between GCT with the CHEC-M camera, and GCT with the CHEC-S camera. Also combinations between ASTRI and GCT may be used in the near future (CHEC-S on ASTRI structure, ASTRI camera on GCT structure), especially in MC.

watsonjj:
Perhaps my comment is an argument to really separate between camera and structure names... in that case there will be a CHECS and CHECM in `instument/camera.py`.

AMWMitchell:
[...]

PR #414, opened 2017-05-23
Fix camera names

kosack:
- unify camera names
- better loading and searching for known cameras (in case no simtel file is open)


kosack:
so far camera names look like this:
allcams

kosack:
see also issue #255

watsonjj:
Do you plan to change "GATE" to "CHEC" now? I noticed you reverted "GCT" back to "GATE".

kosack:
this fixes #404

kosack:
> Do you plan to change "GATE" to "CHEC" now? I noticed you reverted "GCT" back to "GATE".

I think GATE is gone now (at least it should be - do you see somewhere where it was missed?), it's only CHEC.  Note that GCT is the name of the telescope, but GATE is the name of the camera (this is the `camera_id`), so I think that makes the most sense.  Imagine for example putting a "GATE" camera on an Astri telescope, for example, and then it wouldn't make sense to call it GCT.  When I finish implementing the `TelescopeOptics` classes (which are like `CameraGeometry` but for the telescope), those will be associated with a telescope name and version (so 'GCT' would make sense there).

kosack:
Never mind, I see in the figure above that there is still a GATE! I'll track down where it's coming from... Perhaps some place I forgot to fix a name.

kosack:
I seem to be just crazy and accidently forgot which one was the camera name :) I'll fix it in the master.

kosack:
@watson is it correct that the camera is called CHEC? or has it be renamed GCTCam or something?

watsonjj:
Yes the camera is called CHEC :)

PR #415, opened 2017-05-23
Energy regressor and event classifier

tino-michael:
Implemented a scikit-learn-based energy estimator; it trains separate models for the various camera types on an image-by-image basis and combines all image-based predictions in one energy estimator for the event.

TODO:
- [x] unit tests
- [ ] develop an (energy-based) approach to combine the predictions of the different telescope types

Here is some exemplary behaviour on for gamma events using the ASTRI mini-array
green dots are medians and error bars are the 32 and 68 percentiles.

energy_migration_wave_-k-c1-m3-s2 2 3-n4_randomforestregressor
energy_relative_error_wave_-k-c1-m3-s2 2 3-n4_randomforestregressor


kosack:
try merging in the master again - that should fix the problems with travis

ctapipe/reco/energy_regressor.py
```diff
@@ -0,0 +1,383 @@
+import numpy as np
+
+from astropy import units as u
+
+from sklearn.ensemble import RandomForestRegressor
+
+
+def pick_any(my_dict):
```
MaxNoe: As from 3.6 on, python dictionaries are sorted (and this will most probably become a feature rather than an implementation detail) this will always return the first element, not an **arbitrary** element.

Also, this function is a oneliner:  `return next(iter(d.values()))` or if you want a random one: `random.choice(list(d.values()))`

ctapipe/reco/energy_regressor.py
```diff
@@ -0,0 +1,383 @@
...
+
+        if cam_id is None:
+            if len(self.reg_dict) == 1:
+                cam_id = next(iter(self.reg_dict.keys()))
+            else:
+                raise ValueError("you need to provide a cam_id")
+
+        return self.reg_dict[cam_id].predict(X)
+
+    def predict_dict(self, X):
```
MaxNoe: Maybe find a better name, like `predict_for_telescopes`

Also, not only the mean of the predicitions might be interesting, but also the std dev or other measures of centricity like the median.

ctapipe/reco/energy_regressor.py
```diff
@@ -0,0 +1,383 @@
...
+                    # QUESTION if there is no trained classifier for `cam_id`, raise an
+                    # error or just pass this camera type?
+                    raise KeyError("cam_id '{}' in X but no regressor defined: {}"
+                                   .format(cam_id, [k for k in self.reg_dict]))
+
+            predict_list.append(np.mean(res))
+
+        return np.array(predict_list)*self.energy_unit
+
+    def predict_dict_dict(self, X):
```
MaxNoe: Maybe `predict_telescopes`?

MaxNoe:
This seems really inefficient, as every event is predicted on its own, which means for loops and a lot of changes between slowish python and fast compiled code. It should get a lot faster by grouping the events by telescope id and then apply the models to all events of the same telescope type at once.

ctapipe/reco/energy_regressor.py
tino-michael: sure

ctapipe/reco/energy_regressor.py
tino-michael: • As the docstring says, I don't care which element I get; it doesn't have to be the first, it doesn't have to be random. I'm only interested in interfacing my class to the class of the elements -- and they should all have the same class.

• As far as we understand, `d.values()` creates a new object with a copy of all elements. The regressors can become very big and copying them *all* for this is overkill. If you assure me that `d.values()` does not allocate any unnecessary memory, I can put the `next(iter(…` version inline indeed.

MaxNoe: No, d.values() does not create copies of the underlying elements, it just keeps references.

Even list(d.values()) does not copy the elements, it keeps new references to them.

This can be checked by doing `list(d.values())[0] is list(d.values())[0]` which is True.

tino-michael:
That is what `.predict` is supposed to do -- it takes a list of lists as the default scikit-learn `predict` implementation. But then the responsibility to synchronise the telescopes afterwards to the events is upon the user.

ctapipe/reco/energy_regressor.py
tino-michael: cool, I will change it when I'm back from vacation

ParsonsRD:
Hi, by the energy reconstruction level we have already combined telescopes into a single event anyway, as the energy reconstruction step must take place after the direction and core reconstruction. Although I agree for some steps, such as image parameterisation we can probably save some cycles by running the code for all telescopes like this once we have reached reconstruction we have to have fully built events.

kosack:
Yes, I agree in the future we should avoid doing things event-wise that can be done on all events at once (as with a lot of our algorithms after reconstruction). The same with gamma/hadron separation, which should not be event-wise: we should write out the DL2 table, read it, apply gamma-hadron separation to all events at once, and write out DL3.

Of course, for energy reco, Tino is doing it per-telescope, so I'm not sure we can easily write out a table that includes all telescopes, so this way is ok.

tino-michael:
got rid of `pick_any` and added more statistics information to `predict_by_event`

tino-michael:
> Of course, for energy reco, Tino is doing it per-telescope

This implementation is independent of any actual features. So, in principle, you could still come up with a event-wise feature list and use the `.predict` function instead of `.predict_by_event`.
`.predict` behaves exactly like the default scikit-learn implementation, only that it has safe-guard for the dictionary keys. I'm not sure we even need this. We could just assume the user knows what he is doing, remove the `.predict` implementation so that it falls back to the default and have the exact scikit-learn implementation.

PR #421, opened 2017-06-06
Do not share `connexions` member variable among instances

tristan0x:
Default arguments should be not mutable.

### DIFF
--- a/ctapipe/flow/flow.py
+++ b/ctapipe/flow/flow.py
@@ -45,7 +45,7 @@ class PipeStep():
         Maximum number of element the router can queue
 '''
     def __init__(self, name,
-                 next_steps_name=list(),
+                 next_steps_name=None,
                  port_in=None,
                  main_connection_name=None,
                  nb_processes=1, level=0,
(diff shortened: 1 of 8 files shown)


tristan0x:
After looking at the Travis build log, I am not sure my patch is responsible for the failure, is it?

MaxNoe:
@tristan0x No it isn't

Also, why is this called `connexions`? Shouldn't it be `connections`?

tristan0x:
Hi @MaxNoe 

I agree with you. I just reused the previous member variable name.
Should I renamed it to `connections` ?

MaxNoe:
@tristan0x I would say so.

kosack:
there are actually a lot of language problems in this module (french vs english words). The software standards docs specify to use english spelling (or american depending on preference).  

So at least:
* connexion → connection
* processus → process

Using the refactor→rename functionality in PyCharm makes this quite easy to do

tristan0x:
I fixed the language problems in a separate PR: #425

PR #432, opened 2017-06-12
first try at a bugfix in geometry_converter

tino-michael:
@kosack can you try if this works for you? It seems good to me for at least Nectar-, Flash- and DigiCam.

### DIFF
--- a/ctapipe/image/geometry_converter.py
+++ b/ctapipe/image/geometry_converter.py
@@ -1,5 +1,5 @@
 import logging
-
+from collections import namedtuple
 import numpy as np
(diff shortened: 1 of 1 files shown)


kosack:
all the basic tests pass except for the whipple331 camera, but that's only because the computed phi is slightly larger than 1 deg. Probably we need more stringent tests (it only checks phi right now, but should check psi and others).

To run tests on all cameras, just edit `test_camera_geometry.py` and change the line that loads the cam_ids to be:
```python
cam_ids = CameraGeometry.get_known_camera_names()
```

PR #442, opened 2017-06-26
Improve containers to use __slots__

MaxNoe:
If I understood the idea correctly, containers should have:

* A number of fixed items, comparable to an ORM Model like peewee for example
* These Items should have read-only metadata and defaults
* It should not possible to create new instance members

This was currently implemented by raising an error in `__setitem__`, if the key was not a class member.
It involved a lot of look ups in the instance and class `__dict__`s.

I think I found a more concise way to do it, using a `metaclass` setting `__slots__`. 
Instances of `Item` will be added to the `__slots__` along with a `meta` dictionary for the metadata.
Every other class variable will become a read only class instance.

This is meant as a proposal and as bases for further discussion ;)
Let me know your ideas on this.

Working further on the Idea that a Container should be like a ORM Model, I think one should start to subclass `Item` to do static typing, which will help in the serialisation. 
So, for instance introduce `FloatItem`, `DoubleItem`, `StringItem`, `DateTimeItem` etc.

### DIFF
--- a/ctapipe/core/container.py
+++ b/ctapipe/core/container.py
@@ -136,7 +136,7 @@ class Container(metaclass=ContainerMeta):
 
     def items(self):
         """Generator over (key, value) pairs for the items"""
-        return ((k, getattr(self, k)) for k in self.__slots__)
+        return ((k, getattr(self, k)) for k in self._items.keys())
 
     def as_dict(self, recursive=False, flatten=False):
         """
@@ -204,7 +204,9 @@ class Container(metaclass=ContainerMeta):
         return "\n".join(text)
 
     def __getstate__(self):
-        return dict(self.items())
+        state = dict(self.items())
+        state['meta'] = getattr(self, 'meta')
+        return state
 
     def __setstate__(self, state):
         for k, v in state.items():
--- a/ctapipe/core/container.py
+++ b/ctapipe/core/container.py
@@ -1,208 +1,8 @@
 from collections import defaultdict
-from copy import copy
+from copy import deepcopy
 from pprint import pformat
 from textwrap import wrap
 
-
-class Container:
-    """Generic class that can hold and accumulate data to be passed
-    between Components.
-
-    The purpose of this class is to provide a flexible data structure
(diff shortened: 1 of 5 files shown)


kosack:
Thanks, this does look good. I originally tried doing it with `__slots__` in fact, but couldn't quite figure out how to do it correctly.  This is much cleaner.

kosack:
also +1 for implementing it as a metaclass (which is how it should have been done from the start)

kosack:
I'm actually happy to merge this in now, but I can wait if you want to keep this open and also implement the Item subclasses (or that can be a separate PR)

kosack:
By the way, I'm thinking of dropping/refactoring the `Serializer` and `Writer` classes and implementing them more like `TableWriter` (with a common base class), since it follows the actual use cases better and has features like preservation of units and column transforms, etc, so don't worry about breaking them (`TableWriter`/`TableReader` and subclasses should still work however)

ctapipe/core/container.py
```diff
@@ -37,72 +91,52 @@ class Container:
...
-    def meta(self):
-        """metadata key/values associated with this Container.
+            raise KeyError('No such Item: "{}"'.format(item))
 
-        When written to an output file, these will become headers, so
-        should represent data that does not change after the
-        `Container` is constructed.
-        """
-        return self._metadata
+    def __setitem__(self, item, value):
```
kosack: do we still need `__setitem__` and `__getitem__` if `__slots__` is used? I think the default behavior is to disallow new attribute creation

MaxNoe: This is to be able to use bracket access. But your right, the check shouldnt be necessary.

MaxNoe: It is not and I removed it

kosack:
it occurs to me that using descriptors (e.g. http://nbviewer.jupyter.org/urls/gist.github.com/ChrisBeaumont/5758381/raw/descriptor_writeup.ipynb)  might be a good way to implement the Item sub-classes.  In any case, I'll merge this change in now, and you can look into improving the Items separately.

PR #453, opened 2017-06-28
Improve container

### LINKED ISSUE #448: metadata broken in Container
I just noticed a problem in #442 (which is unfortunately already merged).   There is a misunderstanding with what the `.meta` attribute of containers was - it was *not* the Item metadata, but rather a set of header keywords, basically data that sould not become columns in an outtput file, but rather be headers.    Now it seems the item list (which was previously called "attributes") is mixed with "meta", which breaks writing HDF5 and FITS files..

Strangely the tests still pass, but just with warnings like `Keyword name 'tels_with_data' is greater than 8 characters or contains characters not allowed by the FITS standard; a HIERARCH card will be created.  keyword), VerifyWarning)`, and the output files are now full of crazy headers.

Can you change it back so that Container.meta is just a dictionary of headers?  Parts of the analysis also rely on that working.

Otherwise, I can revert the pull request.

MaxNoe:
follow up to #442 fixes #448

### DIFF
--- a/ctapipe/core/__init__.py
+++ b/ctapipe/core/__init__.py
@@ -1,10 +1,10 @@
 # Licensed under a 3-clause BSD style license - see LICENSE.rst
 
 from .component import Component
-from .container import Container, Item, Map
+from .container import Container, Field, Map
 from .factory import Factory
 from .provenance import Provenance
 from .tool import Tool, ToolConfigurationError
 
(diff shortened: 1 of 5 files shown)


MaxNoe:
I think this behaves now as intended:

* meta is an instance dictionary to store metadata, empty by default
* fields is a class instance dict of column names to `Field`s

PR #454, opened 2017-06-28
Parameterisation of image timing

ParsonsRD:
Added a simple function for extracting timing parameters from an image.


ctapipe/image/timing_parameters.py
```diff
@@ -0,0 +1,87 @@
...
+    "TimingParameters",
+    "gradient, intercept"
+)
+
+
+class TimingParameterizationError(RuntimeError):
+    pass
+
+
+def rotate_translate(pixel_pos_x, pixel_pos_y, phi):
```
kosack: Probably this should go somewhere more common at some point. It's also faster to use a matrix multiplication and the `utils.linalg.rotation_matrix_2d`

PR #457, opened 2017-07-03
ImPACT updates

ParsonsRD:
Fairly major refactoring of the ImPACT code adding several new features to the code:

- Optional addition of fit priors
- Selection of minimisation algorithm
- Helper functions for drawing likelihood surface


ctapipe/reco/ImPACT.py
```diff
@@ -18,13 +17,29 @@
...
+
+
+def energy_prior(energy, index=-1):
+
+    return -2 * np.log(np.power(energy, index))
+
+
+def xmax_prior(energy, xmax, width=30):
+
+    x_max_exp = 300 + 93 * np.log10(energy)
```
kosack: these numerical constants should be defined or documented in the docstring

kosack: Also, is energy a `Quantity`, or should it be in some particular unit? If it is assumed to be TeV, the unit name should be in the variable name like `energy_tev`, or you should use a reference energy:

```python
np.log10( energy/(1*u.TeV)) ) 
```

ctapipe/reco/ImPACT.py
```diff
@@ -377,33 +394,33 @@ def get_likelihood(self, source_x, source_y, core_x, core_y,
...
+        azimuth = self.array_direction.az
 
+        # Geometrically calculate the depth of maximum given this test position
         if self.fit_xmax:
             x_max = self.get_shower_max(source_x, source_y,
                                         core_x, core_y,
                                         zenith.to(u.rad).value) * x_max_scale
-
             # Calculate expected Xmax given this energy
             x_max_exp = 300 + 93 * np.log10(energy)
```
kosack: again, these bare constants are not allowed.  Should you use the `ShowerMaxEstimator` here?

kosack: Or at least make this a function "expected_x_max(energy)" that can be documented

ctapipe/reco/ImPACT.py
```diff
@@ -377,33 +394,33 @@ def get_likelihood(self, source_x, source_y, core_x, core_y,
...
-            if x_max_bin > 100:
-                x_max_bin = 100
-            if x_max_bin < -100:
-                x_max_bin = -100
+            if x_max_bin > 250:
+                x_max_bin = 250
+            if x_max_bin < -250:
+                x_max_bin = -250
         else:
             x_max_bin = x_max_scale * 37.8
```
kosack: another unidentified bare constant. Where does 37.8  come from, and what are its units?

ctapipe/reco/ImPACT.py
ParsonsRD: In this case I think we can just use the ShowerMaxEstimator. I forgot this exists, will change it.

ctapipe/reco/ImPACT.py
ParsonsRD: In this case the expected Xmax is based on what I use in the template generation, will add a function and document.

ctapipe/reco/ImPACT.py
ParsonsRD: Yep, this can be deleted anyway as I don't want to support this feature.

ctapipe/reco/ImPACT.py
kosack: In fact, I just noticed ShowerMaxEstimator didn't appear in the documentation due to an import problem, so I just fixed it... Otherwise it was easy to forget!

ctapipe/reco/ImPACT.py
```diff
@@ -19,20 +19,41 @@
...
+    ----------
+    energy: float
+        Energy of the shower in TeV
+        
+    Returns
+    -------
+    float: Expected depth of shower maximum
+    """
+    x_max_exp = 300 * (u.g*u.cm**-2) + \
+                93 * (u.g*u.cm**-2) * np.log10(energy.to(u.TeV))
```
MaxNoe: This will raise an error:
```
TypeError: Can only apply 'log10' function to dimensionless quantities
```

ParsonsRD: Aha yes I missed that one. Thanks!

ctapipe/reco/ImPACT.py
```diff
@@ -19,20 +19,41 @@
...
+    Parameters
+    ----------
+    energy: float
+        Energy of the shower in TeV
+        
+    Returns
+    -------
+    float: Expected depth of shower maximum
+    """
+    x_max_exp = 300 * (u.g*u.cm**-2) + \
```
ParsonsRD: Thats actually a good question :-), I pulled these numbers from my thesis, but they are actually wrong (I accidentally took the values for electrons). However, this does not actually change the result as it is just used to define the binning used in the template generation, I think for now I will comment that this is an incorrect value and needs to be updated. Then I can add a proper reference when I next update the image templates.

kosack:
several ImPACT tests are still failing

ParsonsRD:
OK tests should be fixed now. I will also add some (small) ImPACT template file to the ctapipe extras  repo today so we can properly run all tests.

kosack:
The tests pass, but the documentation has a bug in it somewhere:

```
Warning, treated as error:
/home/travis/build/cta-observatory/ctapipe/docs/reco/ImPACT.rst:48: (WARNING/2) Title underline too short.
Calculating Image Likelihood
++++++++++++++++++++++
```

kosack:
There are still doc errors - I suggest running `make doc` (or `cd docs; make html`) to generate the docs locally, and fix the errors  before committing - it seems the Travis build is configured to stop on the first error, rather than show all of them, so using it to debug will be tedious.

You may need to install the sphinx packages listed in `environment.yml` in the top-level directory to build the docs. Specifically:

  - sphinx
  - sphinx_rtd_theme
  - sphinx-automodapi

ParsonsRD:
Haha yes this has been driving may crazy, will try checking locally.

PR #458, opened 2017-07-03
Hillas intersection (algorithm 1)

ParsonsRD:
Simple Hillas axis intersection code, based on the implementation of read_hess. This code was created as a comparison the the HillasReconstructor.py code to check that the two approaches are functionally identical.


ctapipe/reco/hillas_intersection.py
```diff
@@ -0,0 +1,405 @@
...
+    '''
+
+    def __init__(self, atmosphere_profile_name="paranal"):
+
+        # We need a conversion function from height above ground to depth of maximum
+        # To do this we need the conversion table from CORSIKA
+        self.thickness_profile, self.altitude_profile = get_atmosphere_profile_functions(
+            atmosphere_profile_name)
+
+    def predict(self, hillas_parameters, tel_x, tel_y, array_direction):
```
tino-michael: Should we try to get the interface of the various `predict` functions as uniform as possible?
for comparison:
`HillasReconstructor.predict(self, hillas_dict, inst, tel_phi, tel_theta, seed_pos=(0, 0))`
`ImPACTReconstructor.predict(self, shower_seed, energy_seed)`

The telescope positions are in the `inst` object, my `tel_phi` and `tel_theta` should go in there as well...
The `seed_pos` isn't even used any more but we could add some default `XXX_seed=None` everywhere.
This would help switching out the reconstruction method on the fly without having to think too much about the function call itself. Or do you think we should just add `*args, **kwargs` at the end of the method ;)

Just a though.

kosack: That would be good to have the same interface.  The `inst` object is being phased out (see #456  that I will merge in today, I think, unless there are major issues), but it will still contain the necessary info, but wrapped in a SubarrayDescription object that has the telescope positions, camera geometries, optics descriptions, etc, already constructed (so no more need for `.guess` methods.  Therefore I'd suggest having a `SubarrayDescription` as an input to the constructor of a `Reconstructor`, then the `inst` object is no longer needed. 

e.g. 
```python
subarray = ... # a SubArrayDescription object
reco = HillasReconstructor(subarray)
```

kosack:
seems to be a problem with imports in the test:

```
ctapipe/reco/tests/test_hillas_intersection.py:1: in <module>
    from ctapipe.reco.hillas_intersection import intersect_lines
E   ImportError: cannot import name 'intersect_lines'
```

ctapipe/reco/hillas_intersection.py
ParsonsRD: Hi, yes I agree would be preferable to have a common interface. So what would the required inputs be? Hillas parameters, subarray object, direction seed & energy seed?

tino-michael: those and a position seed, I guess.
Don't you actually need one for your fit?

kosack: Just to comment on where the data should be, I think you can assume that all algorithms get an event data structure that contains all of the things that change *per event* (which would include the pointing positions, which should not be from "instrument", but from the event, since in general that changes in time, even if in the MCs it is currently fixed). 

The second piece of info would be the `SubarrayDescription`, which as I said above should be passed in to the contructor probably, since it doesn't change per event (it contains all of the fixed info).  Therefore having something like this:

```python
class SomeReconstructor(Reconstructor):
      def __init__(subarray : SubarrayDescription):
           super().__init__(subarray)
           ...

      def predict(input_data, tel_pointing, initial_fit_parameters):
         ..
```

Might be sufficient, where input_data is the hillas params or shower images as a dict(?) (or others if we have more methods) and tel_pointing is either a `Coordinate` or perhaps just a tuple of the theta,phi.   I guess that is not per-telescope, but the array-wise pointing in the NominalFrame? or do you need different pointings for each telescope?

kosack: I guess we should move this refactoring/common-input discussion to an issue rather than in this PR.

kosack:
We can refactor this to have a common interface with the other reconstructors separately (see #471), so for now I'll merge it

PR #459, opened 2017-07-03
added test unit for event classification with sklearn MLPClassifier

fvisconti:
Since I tested some nice behaviour with sklearn `MLPClassifier` on my own with ASTRI MC data, I just added some test unit  in ctapipe.
I noticed I couldn't use the `StandardScaler()`  available in `sklearn.preprocessing` module, since the `feature_list` in the code is a `dict`: ideas? Should I take the dict values for the scaler, or override the `StandardScaler` and let it accept the `dict` type input?


tino-michael:
maybe you can add a member function to `RegressorClassifierBase` that takes the `feature_list` as the `dict` that it is now and calls `StandardScalar` internally?

ctapipe/reco/event_classifier.py
```diff
@@ -3,6 +3,8 @@
 from astropy import units as u
 
 from sklearn.ensemble import RandomForestClassifier
+from sklearn.neural_network import MLPClassifier
+from sklearn.preprocessing import StandardScaler
```
tino-michael: I was playing around with different classifiers and regressors as well but intentionally only imported `RandomForestClassifier` here as a default. I would not add more classifiers here but import them in your own script and call `EventClassifier(classifier=MLPClassifier …)` from there. At least that was my intention and that is how I'm doing it when I play around with other classes...

fvisconti: Yes, I understand, that's why I only added a test for the `MLPClassifier`.

tino-michael: and -- concidering my other comment -- the `StandardScaler` import can also rather go to `RegressorClassifierBase`, no?

fvisconti: So, if I get it correctly, you suggest to import the `MLPClassifier` and `StandardScaler` modules in the test file? Or don't care about those test at all?

tino-michael: For `MLPClassifier` I'd say yes. Testing at least two different classifiers is probably a good idea anyway to show that the modularity works.

For `StandardScaler` I suggest to import it there where you add the functionality that uses it. So I'd say in `RegressorClassifierBase` since `MLPRegressor` is also something we can look into.

tino-michael:
If the classifier was trained with scaled features, should the prediction not also be performed the same? Do we have to store the range of the scaling to disk with the fitted model?
The predict test seems to have passed, so I wonder why...

MaxNoe:
Yes. This can be stored using an sklearn pipeline.

http://scikit-learn.org/stable/modules/generated/sklearn.pipeline.Pipeline.html

tino-michael:
cool, if I understand this correctly, we can use this `Pipeline` directly in-place as the `classifier` argument? Then it also should work with `save` and `load` just like that.

MaxNoe:
Yes, But I think one needs to use 
http://scikit-learn.org/stable/modules/generated/sklearn.preprocessing.StandardScaler.html#sklearn.preprocessing.StandardScaler
Instead of the plane `scale` function.

fvisconti:
Should I code a version with `Pipeline` and `StandardScaler`?

tino-michael:
Sure, please.
FYI I'm on vacation now for the next few days with sparse internet.

fvisconti:
I successfully tested a version with no `pipeline`.
Now, I tried _naively_ to write this one too:
    def test_pipeline_classifier():
        cam_id_list = ["FlashCam", "ASTRICam"]
        feature_list = {"FlashCam": [[1, 10], [2, 20], [3, 30], [0.9, 9],
                                     [10, 1], [20, 2], [30, 3], [9, 0.9]],
                        "ASTRICam": [[10, 1], [20, 2], [30, 3], [9, 0.9],
                                     [1, 10], [2, 20], [3, 30], [0.9, 9]]}
        target_list = {"FlashCam": ["a", "a", "a", "a", "b", "b", "b", "b"],
                       "ASTRICam": ["a", "a", "a", "a", "b", "b", "b", "b"]}

        estimators = [('scaler', StandardScaler()), ('clf', MLPClassifier())]

        clf = EventClassifier(classifier=Pipeline(estimators), cam_id_list=cam_id_list,
                                          clf__max_iter=400)
        clf.fit(feature_list, target_list)

        prediction = clf.predict_by_event([{"ASTRICam": [[10, 1]]},
                                                                 {"ASTRICam": [[2, 20]]},
                                                                 {"ASTRICam": [[3, 30]]}])
        assert (prediction == ["a", "b", "b"]).all()

        prediction = clf.predict_by_event([{"FlashCam": [[10, 1]]},
                                                                  {"FlashCam": [[2, 20]]},
                                                                  {"FlashCam": [[3, 30]]}])
        assert (prediction == ["b", "a", "a"]).all()` 

but then I get this error:
`E           TypeError: 'Pipeline' object is not callable`

I suppose we should change the `RegressorClassifierBase` to have it works..

Shall I meanwhile push skipping the `pipeline` test?

fvisconti:
Is this definitively approved?

ctapipe/reco/regressor_classifier_base.py
```diff
@@ -295,6 +296,39 @@ def load(cls, path, cam_id_list, unit=1):
 
         return self
 
+    @staticmethod
+    def scale_features(cam_id_list, feature_list):
```
tino-michael: Since it seems that we are going to use the `Pipeline` approach, I don't think this is still necessary.

ctapipe/reco/tests/test_event_classifier.py
```diff
@@ -1,6 +1,81 @@
...
+    feature_list = {"FlashCam": [[1, 10], [2, 20], [3, 30], [0.9, 9],
+                                 [10, 1], [20, 2], [30, 3], [9, 0.9]],
+                    "ASTRICam": [[10, 1], [20, 2], [30, 3], [9, 0.9],
+                                 [1, 10], [2, 20], [3, 30], [0.9, 9]]}
+    target_list = {"FlashCam": ["a", "a", "a", "a", "b", "b", "b", "b"],
+                   "ASTRICam": ["a", "a", "a", "a", "b", "b", "b", "b"]}
+
+    estimators = [('scaler', StandardScaler()), ('clf', MLPClassifier())]
+
+    clf = EventClassifier(classifier=Pipeline(estimators), cam_id_list=cam_id_list,
```
tino-michael: `classifier` expects a class-name not an already initialised instance. You need to use
`EventClassifier(classifier=Pipeline, steps=estimators …`
and then you need to modify the line that threw the previous error in `regressor_classifier_base.py`
to `self.model_dict[cam_id] = model(**deepcopy(kwargs))` (of course import deepcopy first). Otherwise, every camera_id uses the same model and we wouldn't want that ;)

fvisconti:
Thanks to @tino-michael for fixing the Pipeline usage.
Shall I be notified here when this pull request is merged?

tino-michael:
Yes, since you are in the list of participants, you should get an e-mail again when this is merged.
Cheers :-)

PR #478, opened 2017-07-25
Improve tools and prov

kosack:
a bunch of minor cleanups to the various command-line tools in `ctapipe.tools` as well as the Tool base class. Notibly:

- more verbose default log message (making debugging easier)
- tools now record and mention which files were output, using the provenance system  
- various small cleanups


kosack:
strange - somehow the history from the last master merge (from #457 ) is appearing here... Not sure why that is. There are only 4 commits different in this branch

PR #486, opened 2017-08-10
update to Geometry converter

tino-michael:
complete overhaul of how the rotated pixel positions are applied to the signal container;
instead of applying np.histogram also with the signal, creates an enumerated map that connects the entries between the 1D and 2D image;
significantly reduces the number of objects that need to be buffered;
the conversion back uses the same map, so that now the "unrotated" geometry is the exact same as the original (before, the order of the pixels in the arrays changed and one had to keep track of a third geometry object);
works for time-integrated images and collections of time-slices


ctapipe/image/geometry_converter.py
```diff
@@ -399,3 +483,217 @@ def convert_geometry_back(geom, signal, key, add_rot=0):
...
+    # if there is no time dimension, `signal` is a 1d array and `rollaxis` has no effect.
+    input_img_ext[:-1] = np.rollaxis(signal, axis=-1, start=0).ravel()
+
+    # now apply the transfer map
+    rot_img = input_img_ext[hex_to_rect_map]
+
+    # if there is a time dimension, roll the time axis back to the last position
+    try:
+        rot_img = np.rollaxis(rot_img, 0, 3)
+    except np.core._internal.AxisError:
```
kosack: This fails for me on the latest version of NumPy (you shouldn't used any module marked with an underscore, as that is private and subject to API change).

kosack: it seems it now throws a `ValueError` (which is the base class of the older `AxisError` anyhow, so catching a `ValueError` should work in both cases)

ctapipe/image/tests/test_geometry_converter.py
```diff
@@ -14,53 +14,61 @@
 from ctapipe.image.toymodel import generate_2d_shower_model, make_toymodel_shower_image
 from astropy import units as u
 
-#cam_ids = CameraGeometry.get_known_camera_names()
-cam_ids = ['FlashCam',]
 
-@pytest.mark.parametrize("rot", [3,])
+# cam_ids = CameraGeometry.get_known_camera_names()
+cam_ids = ['FlashCam', ]
```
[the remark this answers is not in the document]
kosack: I just tried it and it works: you can uncomment out the first line and delete the second, so that the test runs on all cameras

ctapipe/image/geometry_converter.py
tino-michael: I see... For me it was throwing this error when it wasn't working so I just caught it. I will change the catch to simple ValueError

ctapipe/image/tests/test_geometry_converter.py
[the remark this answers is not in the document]
tino-michael: Yeah, the full list is already used when you run the test script as main.

PR #489, opened 2017-08-21
Calibration Update

watsonjj:
Updates to calibration docs, and improvements to dst and astri handling


watsonjj:
Hi Karl, I'm hesitant to change it to the `CameraGeometry` as I intended the interface to the charge extractors to be only numpy arrays so they can be easily called in any circumstance.

The tests are failing because of the update to NeighbourPeakIntegrator. I will push an fix.

kosack:
Thanks that seems to have fixed it

PR #490, opened 2017-08-23
Improvements for NeighbourPeakIntegrator

watsonjj:
Created C extension neighbour_sum_c.cc which contains a function to sum all waveforms for a pixel and its neighbours. Using this for NeighbourPeakIntegrator makes it ~5 times faster!

As this involves a C extension, and it is the default choice when performing DL1 calibration, this PR will require all users to reinstall ctapipe so the extension is compiled. To do this I used `python setup.py develop`.


kosack:
Looks nice, and a good example of using a C extension.  It seems the build now fails, however, and I think it's just because the library is not being found correctly. I'm not sure yet why that is.  The travis  build runs `python setup.py install` to install the module - do you know if that is normally enough, or do we need to run "python setup.py build" first?

watsonjj:
I normally run `python setup.py develop` but I would have thought `python setup.py install` would be enough too. You should see the compilation messages when you run the command.

kosack:
Ok, I think I may have solved it.  The problem is that the library is installed in the package dir when `python setup.py install` is run, but when running the tests with `pytest`, it looks only at the local version of the module, not the installed one (unless you ran `make develop`, in which case they were the same and it worked).  To have it work correctly, the tests must be run as `python setup.py test` rather than `pytest` in general. I'll make a separate PR that fixes this in the travis config and setup.py and hopefully this will work from then on...

PR #494, opened 2017-08-31
fix test procedure when compiled c-code is used

kosack:
- fixed a bug in setup.py
- change test produre in Makefile to use `python setup.py test` rather
than `python -m pytest`, so the tests run on an installed package

see #490


kosack:
Strangely, I get a segfault at the end of "make test" on Travis, even though all the tests pass. No idea where it is coming from, and it doesn't happen when I run this locally on my Mac.

PR #498, opened 2017-09-05
Remove old overlays when plotting a new one

MaxNoe:
This removes old moment_overlay if a new one is added to make it possible to reuse a camera display


kosack:
thanks. I keep meaning to finish the updated version of this display that manages multiple overlays, but it's still in a branch that needs work!

PR #501, opened 2017-09-06
WIP: Implement first version of typed fields

MaxNoe:
Do not merge, this is WIP but I want to get Feedback early.

This is something that was already talked about. I started to implement typed fields, inspired by orm models for the containers.

Right now there is
* `Field` for arbitrary python objects
* `ArrayField` for numpy arrays
* `QuantityField` for astropy quantities
* `TimeField` for astropy.time.Time

There need to be more Field types, name any you deem useful.

I plan to rewrite `Map` so map is a `Map` of `Fields` and there should be a `SubContainerField` to have strictly typed `Map → Subcontainer` relations.

This will result in mostly statically typed Containers, which should make serialisation a lot easier.

### DIFF
--- a/ctapipe/core/__init__.py
+++ b/ctapipe/core/__init__.py
@@ -1,10 +1,22 @@
 # Licensed under a 3-clause BSD style license - see LICENSE.rst
 
 from .component import Component
-from .container import Container, Field, Map
+from .container import Container, Map
 from .factory import Factory
 from .provenance import Provenance
 from .tool import Tool, ToolConfigurationError
+from .field import Field, QuantityField, ArrayField
 
-__all__ = ['Component', 'Container', 'Tool', 'Field', 'Map', 'Factory',
-           'Provenance', 'ToolConfigurationError']
+
+__all__ = [
+    'Component',
+    'Container',
+    'Tool',
+    'Map',
+    'Factory',
+    'Provenance',
+    'ToolConfigurationError',
+    'Field',
+    'ArrayField',
+    'QuantityField',
+]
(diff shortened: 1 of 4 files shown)


kosack:
Thanks. Looks like a good start!

kosack:
For `ArrayField`s, we might want to think if *shape* is required, or optional, or whether or not we just have `num_dims` or something like that. Most of the time, we don't know the exact shape when the Container is defined, and it may be different for different telescopes. For example the Cherenkov image size is different for each camera.  

Or, we could pass that information in as a constructor argument(s) to the Container (e.g. `c = CameraContainer(npix=100)`, meaning that we have to construct them on-the-fly.

kosack:
just as a note, it seems python 3.7 will have something that basically does what the Container classes do in the standard library:   https://www.python.org/dev/peps/pep-0557/

dneise:
this might be interesting in this context:

Kai just brought my attention to this:
https://docs.python.org/3/library/typing.html#typing.NamedTuple

and Karl mentioned in this context this once:
https://www.python.org/dev/peps/pep-0557/#why-not-just-use-typing-namedtuple

kosack:
NamedTuple doesn't quite fit what we want, but there are other libraries that do similar things like the `attrs` package (and even pytables and sqlalchemy do very similar things with their respective ORMs). However, I think rolling our own is fine for now, since we need extra metadata beyond what we get from  the type annotation system (like units and UCDs).  I still like moving to these type of named fields, since it will simplify a lot of other problems.

kosack:
an even more exhaustive list, for inspiration: https://github.com/grundic/awesome-python-models

PR #504, opened 2017-09-08
Add containers for pointing and source information

MaxNoe:


kosack:
SourceContainer may not be very useful, since we will nearly always have more than one source in the FOV

kosack:
Also, are these "tracking" positions, or "pointing"?  The difference is tracking is where the telescope thinks it is, and pointing is after applying corrections.    There is also another way this could be done, rather than having tracking and pointing as values per-event as here, we could create a TrackingInterpolator class.  Note that tracking values are not really recorded per-event, but once per second or so, and normally will be stored in a separate FITS table with columsn like Alt, Az, Alt_err, Az_err, Alt_velocity, Az_velocity, etc....   THen one could load this into a `TrackingInterpolator` and call `get_position(event.gpstime)` to get the current tracking position. 

This is probably how it should be done in the future.  It's the same for all monitoring quantities like HighVoltage - you don't want to store that per-event, but in some table that can be interpolated in time.

MaxNoe:
I think of this as the high level container, which contains the values that should be used for reconstruction, meaning the transformation from sky to camera coordinates  and back.

How this value is calculated is a totally different story and might need more containers and a lot of telescope specific code. 
But imho we need this high level container now, to fill it with either MC Truth, or the already calibrated data from existing telescopes like FACT.

In my attempts to analyse FACT data with ctapipe, I fill this value from our auxiliary data fits files, containing the tracking position measured every 2 seconds. I select the closest point to the recorded event.
One could also interpolate, but this is not the question here, I think.

There should be a simple, high level container for the values that are needed for reconstruction and the low level calibration for those values should happen elsewhere.

kosack:
Ok, yes then I agree - how we fill it can be something more complex, but it can be somehow updated when necessary. If it's for reconstruction, then it is the pointing position after all necessary corrections, so should be fine as is.

MaxNoe:
I removed the `Source` container and added some comments in the docs of `TelescopePointingContainer`

MaxNoe:
> SourceContainer may not be very useful, since we will nearly always have more than one source in the FOV

Then one could have a list of source containers?

kosack:
Generally it's not needed to know what the source was during reconstruction, only during analysis (except in the case of Monte-Carlo point sources where you might want to select events around the simulated position, which is already possible).  The current data model for regular data has each observation linked with N proposals and M sources (and likewise a source can be linked with multiple proposals or obs_ids), so storing it here is unnecessary, it's an external database query.  Later parts of the pipeline might load a list of sources in the FOV for example when producing monitoring lightcurves or looking for flares, but that doesn't need to be in a container, since at that point in the analysis we will be using science tools not ctapipe.

kosack:
Should i merge this? I think it's fine now, though we need to modify hessio_event_source to fill it, but that can be done separately.

MaxNoe:
Yes

PR #507, opened 2017-09-11
add geometry_converter to documentation build

tino-michael:
• pull `geometry_converter` in `ctapipe/image/__init__.py`
• added `docs/image/geometry_converter.rst`
• added `geometry_converter` in `docs/image/index.rst`


kosack:
Now that the docs are being built for geometry_converter, it seems there are some syntax problems in them.  It's complaining that somewhere there is a section called "Args" that probably should be "Parameters"

tino-michael:
this "Args" problem seems to be from numpy and to be only a warning...
I think this is the actual problem:
```
Warning, treated as error:
/home/travis/build/cta-observatory/ctapipe/docs/image/geometry_converter.rst:19:toctree references unknown document 'api/ctapipe.image.geometry_converter.jit'
```

I use `numba.jit` to decorate some functions and it seems that sphinx expects some jit documentation that it doesn't find

PR #511, opened 2017-09-18
added sample_weights in ML fit function

tino-michael:


kosack:
seems to need an update - the number of arguments to `fit()` changed, breaking some tests.

tino-michael:
I did not realise that not all classes' `fit` call supports a `sample_weight` argument. I added some exception handling.

PR #518, opened 2017-09-27
Set the number of subplots from the number of known cameras 

orelgueta:
In the current master, this example fails due to the hardcoded values. Fix that and try to be more robust for the future.


examples/plot_all_cameras.py
```diff
@@ -2,17 +2,18 @@
...
     plt.style.use("bmh")
 
-    N = 2
-    M = 7
-    plt.figure(figsize=(15, 6))
-
     camera_names = CameraGeometry.get_known_camera_names()
+    M = 5
+    N = ceil(len(camera_names)/M)
+    plt.figure(figsize=(15, 6))
```
tino-michael: how about going all the way and doing it like
```
n_tels = len(camera_names)
n_cols = np.ceil(np.sqrt(n_tels)).astype(int)
n_rows = np.ceil(n_tels / n_cols).astype(int)
```

orelgueta: Sure, but I switched between cols and rows and rounded down for the number of rows. It helps get a nicer looking plot.

tino-michael: Sure, it was just a suggestion to make both dimensions dynamically sized :-)

PR #534, opened 2017-09-28
add muon test files and fix plot_muon_event call

AMWMitchell:
adding mst-fc and sst-1m muon test files at 100% optical efficiency. Minor fix to call to plot_muon_event


kosack:
ctapipe-extra is a separate package, so you should not commit files in it to the ctapipe repo.  You have to check out ctapipe-extra's git repo separately and modify it.  In fact there should not be a ctapipe-extra subdirectory in ctapipe (it was removed several versions ago).  

So unfortunately I cannot merge this in since it would add a fantom ctapipe-extra subdir with data files in it.

kosack:
in fact I'll just make this change manually since I'm cleaning up the code already

AMWMitchell:
OK - the minor fix was just removing one argument in the call to plot_muon_event .
Things got messy here - added the muon files to ctapipe-extra separate repository now.

PR #537, opened 2017-09-28
some fixes and cleanups to muon code

kosack:


kosack:
@AMWMitchell  this should include your fix to the muon plotting (but not the new files in ctapipe-extra)

AMWMitchell:
Thanks, the new files should be added correctly to ctapipe-extra now

PR #546, opened 2017-09-29
Fixed broken code

Sojourner92:
Muon diagnostic plots code included numerous code breaking
syntax errors in continued
lines. Also removed unused variables.


ctapipe/image/muon/muon_diagnostic_plots.py
```diff
@@ -125,14 +124,6 @@ def plot_muon_event(event, muonparams):
...
             signals = image * clean_mask
 
-            muon_incl = np.sqrt(
-                muonparams['MuonRingParams'][idx].ring_center_x ** 2. +
-                muonparams['MuonRingParams'][idx].ring_center_y ** 2.)
-
-            muon_phi = np.arctan(
-                muonparams['MuonRingParams'][idx].ring_center_y /
-                muonparams['MuonRingParams'][idx].ring_center_x)
-
```
kosack: are these simply not used? or is there some other reason for removing the calculation?

ctapipe/image/muon/muon_diagnostic_plots.py
```diff
@@ -173,19 +164,6 @@ def plot_muon_event(event, muonparams):
...
-            )
-            ring_dist = np.abs(
-                dist - muonparams['MuonRingParams'][idx].ring_radius
-            )
-            pixRmask = ring_dist < muonparams['MuonRingParams'][
-                                       idx].ring_radius * 0.4
-
-
             if muonparams['MuonIntensityParams'][idx] is not None:
                 signals *= muonparams['MuonIntensityParams'][idx].mask
```
AMWMitchell: Don't remove these lines, add at line 191/169:
else:
signals *= pixRmask

ctapipe/image/muon/muon_diagnostic_plots.py
AMWMitchell: Don't remove these lines, would prefer these were added to a container (e.g. add muon inclination to the muon ring parameters container)

AMWMitchell:
Please don't remove unused variables without considering first whether they should/could be used. 
This could lead to problems.

AMWMitchell:
Alternatively, move muon_incl and muon_phi to muon_reco_functions (which does fill containers) as the calculations can be done just as well there. 
In which case, removing is OK after all, as long as they are added and stored properly in muon_reco_functions.

Sojourner92:
Sorry, I thought being in a diagnostic plot code would mean that unused variables were not useful, I've made the requested changes to the muon container class and and propagated them through to the plotting code as requested.

AMWMitchell:
Thanks for fixing this - it's a fair assumption to make, and the code did need a bit of cleaning - at least now the inclination angle and ring orientation are calculated and stored in the right place.

kosack:
@AMWMitchell are you ok with this being merged? I don't see any obvious problem, but I'm not an expert with this code.  I'll wait until you click "approve"

PR #548, opened 2017-09-29
Impact reversion

ParsonsRD:
Fixed some code breaking issues introduced in commit #533. Function definitions had been changed without modifying function calls, call to migrad was removed so no minimisation was performed.

Some print statements also removed.


kosack:
Sorry, I didn't notice the change was more than cosmetic.  Can you add some unit tests to make sure this is checkedfor ? All tests passed. Though I guess we might need to add some very simplistic test model somewhere that we can load to test the more complex parts. It doesn't have to be real data, just some dummy model that can be used to test that the mechanism works, even if not a scientific test, similar to the test of the `EnergyRegressor`.

kosack:
I think it is correct that some things were confusingly not used (like the zenith and azimuth parameters to `image_prediction()`, which were overwritten by other data in the function call

ParsonsRD:
Yes fully agreed. That's my bad I'll try to write some better test functions. I just went on leave though so that'll have to wait until November :-).

kosack:
note sure why landscape.io is not running today, but I'll accept this anyhow

PR #549, opened 2017-09-29
added Qfactor calculation to EventClassifier class and testing function

fvisconti:
I used `pandas` for histogramming; it's not really needed here, but we discussed with @kosack that adding `pandas` is fine.
Tests can be expanded to **QFactor** values reliability, but have to think better how can it be tested.

Moreover, I use the fact that labels are `int`, otherwise it all breaks.


kosack:
We need to also then add pandas as dependencies to the `environment.yml` and to `setup.py` .  Alternately use `ctapipe.utils.Histogram`, but it's not quite the same use case.

fvisconti:
Dependencies added in setup.py and environment.yml, seems to work.
Would be nice to have a small (but not _too_ small!) dataset to test Q-factor values. Data needed are MC DL1 (Hillas parameters) protons and gammas.

fvisconti:
Why the code-quality/landscape check is taking so long?

kosack:
Not sure - sometimes it just doesn't run, or doesn't report status.  I'll ignore it at this point, since the code looks fine.

PR #554, opened 2017-10-05
Implement dl1 dumper

kosack:
This PR is just a test to write a simple writer for DL1 information, for testing purposes. It uses the HDF5TableWriter class to serialize the DL1Container to an output file.  It is  **not ready to merge** yet due to some issues:

1. the files aren't properly closed at the end (resulting in sometimes corrupt data, particularly if compression is enabled).  How to handle this needs some thought (make `HDF5TableWriter` use a context manager, or at least add a `close()` function to clean up properly?)

2. the output files are very slow to read, most-likely because the *chunksize* of the resulting HDF5 needs to be optimized (I'm not sure how to choose a good one)

3. of course the output data is probably not optimal (there needs to be a cleanup of what is stored in the DL1Container, since there is some replication, e.g. storing both "cleaned" and "image")

4. the "foreign keys"  (e.g *event_id*) are not in the Containers and so are not written, so needs some extension to the TableWriter to either add them or the ability to write multiple containers at once.

The output data currently looks like this, with one table per camera:

screen shot 2017-10-05 at 17 07 07

The data can then be accessed:

```python
import tables
data = tables.open_file("dl1_dump.h5")

t1 = data.root.DL1.tel_001
images = t1.col("image")  #images for all events

# histogram of just pixel 700 for all events, channel 0
plt.hist(images[:,0,700], log=True)  

#histogram for all events, all pixels, channel 0:
plt.hist(images[:,0,:].ravel(), log=True, bins=50)
```
ex

In principle, this can also be accessed by `pandas` (pandas.pydata.org) rather than the lower-level `pytables` library, but the *extracted_samples* array is in a format that it doesn't like for some reason.

some notes on chunk-shape:
* https://stackoverflow.com/questions/31383122/what-is-the-recommended-compression-for-hdf5-for-fast-read-write-performance-in
* http://www.pytables.org/usersguide/optimization.html
* https://support.hdfgroup.org/HDF5/doc/Advanced/Chunking/


dneise:
I also liked [Python and HDF5](http://shop.oreilly.com/product/0636920030249.do)  by Andrew Collette

PR #555, opened 2017-10-05
notebook examples for theta square plot and HDF5 images packing

Bultako:
I have done some notebooks showing how I would make images extraction in HDF5 format. For the serialization of D1CameraContainer I have set the following structure: /DL1/tel_id/channel_id/event_id

Attached screenshots of HDF5 images created by notebooks.

dl1_tel21_chan00_event23703

toymodel


kosack:
At the DL1 level there should *not* be 2 channels anymore (the correct channel should be selected in the R1 calibration), so that simplifies your output a bit.  Is it really efficient to store each image in a different dataset? I think that has extreme overhead, if I understand correctly, since each dataset gets it's own metadata. It's better to have a single dataset that has columns of "event_id, image, ...", where image is an ND array that stores all images (e.g. you end up with images shaped like `event_id,i,j`).  See for example #554 (though in that example I store the images as 1D arrays, so the result is 2D, but there's no reason it wouldn't work for 2D images like those output by the GeometryConverter)

Bultako:
The choice of having a different image for each event instead of packing all images in a single table dataset is *just* for visualization/inspection purposes in HDF5-related software (e.g. HDFView) I agree this has an overhead due to common metadata. Maybe I am missing something, since I am not quite familiar with this soft.

I will have a look on the possibility to pack images in a single dataset keeping the images inspection functionality.

I'll keep you updated.

kosack:
Ah, I see.  yes there must be a more efficient way to visualize them.  Is there anything like "SaoImage DS9" for HDF5 files? Basically what you want is a data cube visualizer where you can scan through a dimension with a slider-bar. It seems like something that might exist...

kosack:
Actually, an easy way would be to load the dataset using pytables (or h5py) in a Jupyter notebook and use `ipywidgets.interact()` to have it display. Something like this:

```python
from ipywidgets import interact
%matplotlib inline

datacube = load_hdf5_data_from_somewhere()  # assuming shape is (N, X, Y)

@interact(eventnum=(0,len(datacube))
def display(datacube, eventnum):
    plt.imshow(datacube[eventnum])

```

That would show you the image, and let you see others by moving the slider-bar.

Bultako:
I think that's a good feature to implement in notebooks.

We could load a datacube from a serialized dl1 container
`datacube = load_hdf5_datacube(hdf5_dl1=<filename>, axis=<axis>)` 

being axis 'tels' or 'events', we could have a different datacube.

Though I think people would still like to work with images packed in the serialized HDF5 DL1Container outside a notebook. I'll look for them for concrete needs in this aspect.

Bultako:
I have created notebook where all events images are serialized in a HDF5 table datasets for each telescope, and added interactive widget inspection for telescope data cubes  and events data cubes,

teldatacube

eventdatacube

PR #557, opened 2017-10-06
changed np.count_nonzero() to np.sum() in tailcuts_clean

kosack:
for some reason this had different behavior with different versions of
numpy, causing tests to fail in some cases.  Still not sure why, but on my machine (osx with 1.11.3) the test with min neighbors=2 failed for the last test case, while it worked on the travis build.


kosack:
Could this be due to rounding error when determining the neighbors of a square-pixel camera? Though in that case, I would expect the results to be the same between `count_nonzero(mask)` and `sum(mask)`, which should always give the same results, i would think.

PR #559, opened 2017-10-09
added to methods to EventClassifier

fvisconti:
* a hyperBinning method that output pandas groups
* a level_populations method to equalize gamma and hadrons based upon
binning

When doing gamma hadron separation, it is common to wrangle input data equalising the number of entries for gammas and hadron in all the requested hyper-bins, before training the classifier.
This is different from the sklearn train_test_split in the sense that this levels the two populations based upon a previous binning over any requested feature.

### DIFF
--- a/ctapipe/reco/event_classifier.py
+++ b/ctapipe/reco/event_classifier.py
@@ -1,12 +1,11 @@
 import numpy as np
 import pandas as pd
-
+import logging
 from astropy import units as u
 
 from sklearn.ensemble import RandomForestClassifier
 
 from .regressor_classifier_base import RegressorClassifierBase
-
 
 def proba_drifting(x):
     """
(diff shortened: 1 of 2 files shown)


ctapipe/reco/event_classifier.py
```diff
@@ -104,4 +103,127 @@ def compute_Qfactor(self, proba, labels: int, nbins):
...
+            feat_dict = featsToGroupBy[i]
+            bins = np.linspace(feat_dict['minf'], feat_dict['maxf'], feat_dict['nbins']+1)
+            binning_list.append(pd.cut(dfx[feat_dict['col']], bins))
+
+        groups = dfx.groupby(binning_list)
+
+        return groups
+
+    def level_populations(self, group_gamma, group_hadron, gamma_evts, hadron_evts):
+        """Equalize number of entries in each bin.
```
kosack: Keep in mind we may also be looking for electrons or other types events (Iron, etc) for some use cases, so perhaps naming this "signal" and "background" rather than "gamma" and "hadron" is better?  I think at least we can so far assume we have only 2 labels.

tino-michael:
you mean the `StandardScaler` efforts done in #459 ?

kosack:
Yes, I was just wondering the advantage of binning vs weighting.  It would be good for one or both of you to add some info to the user docs (in docs/reco) explaining how to use either method.

PR #580, opened 2017-11-21
Coord trafo

tino-michael:
added a module that contains some low-level coordinate transformation functions to bring various coordinate systems into the convention that is expected by my reconstruction tools -- i.e. right-handed x-y-z coordinate system with z->upwards and phi and theta in the standard (?) definitions: theta, angle to z-axis and phi, angle in the x-y-plane from the x-axis towards the y-axis.

By this approach, assuming that azimuth is measured from north towards east, the x-axis points north and y points west, so `phi = -azimuth`. 
The camera coordinates are mirrored along the x-axis, so that y points "up" and x to the "right", as it is expected by `pixel_position_to_direction`.


kosack:
could you move the `/utils/coordinate_transformations.py` file to `/coordinates/`? It makes more sense there

PR #536, opened 2017-09-28
New directory tree for the atmosphere calibration code

FrancaCassol:

PR #583, opened 2017-11-23
Atmo dev

bregeon:
I am willing to start putting code into the atmo-dev branch.
There are a number of simple generic scripts for atmosphere related reading/plotting of standard density and transmission files.
There is also an example of using numpy to reduce elastic lidar data.
This also contains a "canvas" for the documentation.

For now unlike what is proposed in #536, there is no subtree structure, I am actually not convinced it's absolutely needed.

The code is not yet "ctapipe" compliant in the sense that the code does not fully follows coding conventions, documentation is not complete and there are not unit tests yet.


bregeon:
waiting for codecov and travis to run... then I'll merge myself to check I really have the rights to do it.

bregeon:
hum, I realize that the PR title is "bregeon wants to merge 35 commits into cta-observatory:master from bregeon:atmo-dev"
this is clearly not what I am willing to do, I am willing to merge bregeon:atmo-dev into cta-observatory:atmo-dev.

bregeon:
Looks good, next is to advertise the existence of this branch and start to fix coding style, doc and write unit tests.

MaxNoe:
@bregeon I think it is not good to have long living development branches.

They tend to diverge, become harder and harder to merge and get messy. 

A branch should be as short lived as possible. Implement a change, open a PR, delete the branch directly after the merge. We should not advertise the existence of a branch here parallel to master as this is against the proposed workflow for this repository. See https://guides.github.com/introduction/flow/ and https://cta-observatory.github.io/ctapipe/getting_started/index.html

kosack:
I would say this is not intended to be long-lived. Once the relevant files are collected here, we can just merge it into the master and continue development there. Before that, however, the files will need to be updated to meet the style guide, etc, so several people may be involved (hence a common branch, rather than just Johan's fork).  Though since nothing in this branch conflicts with other parts of ctapipe, I'm happy even to merge it before things are fully finished, to ensure there is no divergence

PR #598, opened 2017-12-08
Zfits reader cocov

dneise:
Work in progress:

I would like to revive #340 a little bit, so I cloned @cocov s fork and made two little commits, to make the tests pass.

I currently skipped the test of this reader, since it depends on external packages. 
I still wanted to open this PR so people can have a look at comment on it.

I would assume PR #340 can be closed now, since it is superseded by this one, unless the original author @cocov  objects.


dneise:
I was having a look into the installation of the required external package "protozfitsreader" to make it a bit easier to install. At the moment installation can be done like this:

    conda create -n pz python=3.5
    source activate pz
    pip install git+https://github.com/dneise/protozfitsreader

But as you see, this needs python 3.5. Etienne is working on this at the moment and I do not want to interfere there.

dneise:
This depends on #613 
tests are failing unless #613 is merged

ctapipe/io/zfits.py
```diff
@@ -4,115 +4,151 @@
...
-        counter +=1
-
-        if max_events is not None and counter > max_events:
-            return
-
+    telescope_ids = Field([], "list of IDs of telescopes used in the run")
+    num_pixels = Field(Map(int), "map of tel_id to number of pixels in camera")
+    num_channels = Field(Map(int), "map of tel_id to number of channels")
+    num_samples = Field(Map(int), "map of tel_id to number of samples")
+    geom = Field(Map(None), 'map of tel_if to CameraGeometry')
```
kosack: I think `geom`, `cam`,  'telescope_ids`, and `optics` are not needed - they are all in the `SubarrayDescription`, which already works for SST-1M.    The only things that I see here that are *so far* SST-1M specific are the cluster and patch matrix definitions. All other info is already present elsewhere.

kosack: By the way, what is the difference between cluster and patch? I assume one is trigger related and the other camera module. IN which case, we should use the correct CTA terminology from the glossary.

From the glossary:
- module: sub-component of camera containing several pixels
- cluster: a unit including several modules (min 2)
- patch: couldn't find definition in the glossary (probably should be one, or is there another name for trigger groups like "sector"?)

kosack: Once I see how it works for SST1M, I think I'll just update the instrument module to have all this info - it shouldn't be in a Container (which are schemas for a row in an output table, and this data will not be written that way)

dneise: Thank you for the feedback. This is really helpful. The more stuff, I can remove the better.

dneise: > By the way, what is the difference between cluster and patch?
I had to look at that as well. Hopefully I get it right. It is basically described here on slide 6:
https://indico.cta-observatory.org/event/1362/contributions/14325/attachments/12437/14554/sst_1m_pipeline.pdf

This is all trigger related. 

The trigger sees a low resolution version of the camera, both in space and in dynamic range, but in real time of course. To make this low resolution image, 3 pixels are summed up and only the 8 least significant bits are transmitted. So instead of a 1296 X 16bit X 250 MHz = 5 Tbit/s stream, the Trigger "only" sees a ~860Gbit/s stream of data. On this data the trigger can take decisions on clusters of 7 patches or on 19 patches.

dneise:
Soo finally the tests are passing, **but** .. 

This reader only provides a part of the information the original SST1M reader provided. The SST1M reader also provided some instrument information, which I have not idea how to provide.

Anyway. I better like tiny steps than no steps at all, so I am wondering if @kosack or @watsonjj could review this again and tell me if this looks good from ctapipe point of view?

ctapipe/io/tests/test_zfits.py
```diff
@@ -0,0 +1,33 @@
+from pkg_resources import resource_filename
+import os
+
+example_file_path = resource_filename(
```
watsonjj: Is it possible to use `from ctapipe.utils import get_dataset` instead?

ctapipe/io/tests/test_zfits.py
```diff
@@ -0,0 +1,33 @@
+from pkg_resources import resource_filename
+import os
+
```
watsonjj: This test module needs to be skipped if protozfitsreader is not installed:

```python
import pytest
pytest.importorskip("protozfitsreader")
```

ctapipe/io/zfits.py
```diff
@@ -4,109 +4,96 @@
...
-    url : str
-        path to file to open
-    max_events : int, optional
-        maximum number of events to read
-
-
+__all__ = ['ZFitsEventSource']
+
+
+class ZFitsEventSource(EventSource):
```
watsonjj: If an instance of `ZFitsEventSource` is created, then there is no warning if protozfitsreader is not installed. It will only fail when you start reading from the EventSource.

I would suggest creating an `__init__` method with a `try...except` where it gives some extra information about if the import of protozfitsreader failed.

For example, in `HESSIOEventSource`:

```python
    def __init__(self, config=None, tool=None, **kwargs):
        super().__init__(config=config, tool=tool, **kwargs)

        try:
            import pyhessio
        except ImportError:
            msg = "The `pyhessio` python module is required to access MC data"
            self.log.error(msg)
            raise

        self.pyhessio = pyhessio
```

ctapipe/io/zfits.py
```diff
@@ -4,109 +4,96 @@
...
+        return (
+            (h['XTENSION'] == 'BINTABLE') and
+            (h['EXTNAME'] == 'Events') and
+            (h['ZTABLE'] is True) and
+            (h['ORIGIN'] == 'CTA') and
+            (h['PBFHEAD'] == 'DataModel.CameraEvent')
+        )
+
+    @property
+    def is_stream(self):
```
watsonjj: This EventSource reads files, no? It is not a stream. This property is used by `EventSeeker` to know if it is allowed to "seek" to an event, which is presumably allowed by this reader.

ctapipe/io/zfits.py
```diff
@@ -4,109 +4,96 @@
...
+        return True
+
+
+class SST1M_R0CameraContainer(R0CameraContainer):
```
watsonjj: 2 points:

Firstly, inheriting from `R0CameraContainer` does not inherit the Fields of `R0CameraContainer`.

Secondly, I think we should change the approach here. If there is event information you wish to store alongside the information in the default `R0CameraContainer`, that is not used in typical ctapipe execution, but could be useful for "expert/debugging" scripts, then it should be placed in its own container.

I suggest creating a new "SST1M" Container that is in addition to the default containers.

ctapipe/io/tests/test_zfits.py
```diff
@@ -0,0 +1,33 @@
...
+        'tests',
+        'resources',
+        'example_10evts.fits.fz'
+    )
+)
+
+FIRST_EVENT_NUMBER_IN_FILE = 97750287
+ADC_SAMPLES_SHAPE = (1296, 50)
+
+
```
watsonjj: Can you add some tests to check `ZFitsEventSource` is correctly returned from `EventSourceFactory` for the relevant input file?

dneise:
Thank you for the quick review.

ctapipe/io/tests/test_zfits.py
dneise: I think not. The file which is needed for this test is not in one of the places, where this function is searching I presume. Unless I misunderstood that function.

ctapipe/io/tests/test_zfits.py
dneise: thanks .. done 👍

ctapipe/io/zfits.py
dneise: I do not think these additional 9 lines (maintenance burden, noise for the reader) are worth it.
When `protozfitsreader` is not installed, the user will get a line like this:
```
ModuleNotFoundError: No module named 'protozfitsreader'
```
I agree a message like this: "The file:{path} you are trying to read is a fits.fz file. In order to read that file you need to have a package named `protozfitsreader` installed. You can find that package here:{url}"
Contains more information. But it is worth it? 
I'd like to hear a 2nd opinion here. 

please note: it costs me less time too implement these 9 lines, than writing this answer. I just assume, in the future many people (hundreds?) will read ctapipe code, thus every line we write too many, counts hundredfold.

ctapipe/io/zfits.py
dneise: It is a file, yes. But at the moment, the Python interface of the C++ protozfitsreader implemented [here](https://forge.in2p3.fr/projects/ctaactl/repository/show/CamerasToACTL/trunk/IO/Fits) does not expose random row access, so I assumed this means it is not compatible with the Seeker.

ctapipe/io/zfits.py
dneise: > 

It does not? Why does this work then?
```python:
In [1]: from ctapipe.io.containers import R0CameraContainer
In [2]: class Foo(R0CameraContainer):
   ...:     pass
In [3]: foo = Foo()
In [4]: foo.adc_samples = np.arange(10)   # works
In [4]: foo.a = 3                                       # should not work
AttributeError                            Traceback (most recent call last)
<ipython-input-4-a25649f06fba> in <module>()
----> 1 foo.a = 3
AttributeError: 'Foo' object has no attribute 'a'
 # and it does not work. 
# Foo is pretty much exactly what I would call a R0CameraContainer
```

dneise: > I think we should change the approach here. 

Can you make a proposal? I do not see how I can do this without inheriting from Containers, without repeating a lot of code.

ctapipe/io/tests/test_zfits.py
dneise: Ah .. well sort of.  I do not think, this PR should test that EventSourceFactory works as expected. But you are right, `is_compatible` has not been tested.

dneise: `is_compatible` was indeed broken and the test showed it nicely. Thank's for pointing that out

ctapipe/io/zfits.py
watsonjj: My main issue is that this error won't be raised when an `ZFitsEventSource` is created, but only once an attempt is made to begin reading from it. Maybe the difference isn't so large, but in my mind it seems more right - A `ZFitsEventSource` shouldn't be allowed to exist unless `protozfitsreader` is installed.

ctapipe/io/zfits.py
watsonjj: Na the default of `EventSeeker` is to loop through the files to find the correct event. Only if the `EventSource` has defined a `get_event_by_index` function will the `EventSeeker` utilise a more efficient approach. So therefore `ZFitsEventSource` is compatible with `EventSeeker`, and this property should return False (or be removed, as False is the default from the parent)

ctapipe/io/zfits.py
watsonjj: > It does not? Why does this work then?
Perhaps I am mistaken. I just remember encountering problems before when inheriting instead of creating from fresh.

> Can you make a proposal?

```python
class SST1MCameraContainer(Container):
    """
    Storage of raw data from a single telescope
    """
    pixel_flags = Field(ndarray, 'numpy array containing pixel flags')
    digicam_baseline = Field(ndarray, 'Baseline computed by DigiCam')
    local_camera_clock = Field(float, "camera timestamp")
    gps_time = Field(float, "gps timestamp")
    camera_event_type = Field(int, "camera event type")
    array_event_type = Field(int, "array event type")
    trigger_input_traces = Field(ndarray, "trigger patch trace (n_patches)")
    trigger_output_patch7 = Field(
        ndarray,
        "trigger 7 patch cluster trace (n_clusters)")
    trigger_output_patch19 = Field(
        ndarray,
        "trigger 19 patch cluster trace (n_clusters)")

    def fill_from_zfile_event(self, event):
        self.pixel_flags = event.pixel_flags
        self.digicam_baseline = event.baseline
        self.local_camera_clock = event.local_time
        self.gps_time = event.central_event_gps_time
        self.camera_event_type = event.camera_event_type
        self.array_event_type = event.array_event_type
        self.trigger_input_traces = event.trigger_input_traces
        self.trigger_output_patch7 = event.trigger_output_patch7
        self.trigger_output_patch19 = event.trigger_output_patch19

class SST1MContainer(Container):
    tel = Field(
        Map(SST1MCameraContainer),
        "map of tel_id to SST1MCameraContainer")


class SST1M_DataContainer(DataContainer):
   sst1m  = Field(SST1MContainer(), "SST1M Specific Information")
```

And then fill the R0 container separately. This ensures that the user can always expect the same contents to be found in the R0 Container, and not be confused about extra camera-specific information included. I think this separation is a better approach.

ctapipe/io/tests/test_zfits.py
watsonjj: Adding a test specifically for checking `ZFitsEventSource ` is returned from `EventSourceFactory` will help us to quickly identify problems such as where two different file formats are found to be compatible with the same `EventSource`, or a single file format is designated as compatible with two different `EventSource`s

ctapipe/io/zfits.py
dneise: Thanks for the clarification.

ctapipe/io/zfits.py
dneise: Ah I missed the main issue. Ok, the import protozfits is now in the `__init__`. My pylinter complains about an unused import now, which is why I did deliberately not do this in the first place....

ctapipe/io/tests/test_zfits.py
dneise: Ah .. I see ... I'll have to look at similar tests from other event sources to see how one can test for this.... I doubt it is possible.

dneise:
Sooo...  I followed @watsonjj s proposal regarding the splitting of the containers. 
Then I realized, that it was hopeless for me, to answer the following question: "Did I really fill all the fields of a certain (sub-)Container? Or did I miss some?"

So I started breaking out little functions, all responsible for filling only their little part of the world from the `protozfitsreader.Event`. 

At this point the code is failing again, for a reason I do not understand, maybe somebody can help me.

However, the next logical step after having all these little functions to fill containers from an event, is to implement unit tests for them, which will hopefully help me catch my mistake.

I think I like these functions a lot. I was just looking at the [`HESSIOEventSource`](https://github.com/cta-observatory/ctapipe/blob/master/ctapipe/io/hessioeventsource.py#L13) again and realized for myself, that I am utterly lost, when reading this code and trying to figure out, if this event source is filling all fields of R0 but only partly R1 or the other way round? 

I have a general question:

Obviously it is okay to fill Containers only partly. I mean, no EventSource has all the information already in hand, right? But on the other hand, it is obviously a bug, when an EventSource fills the waveform, but omits filling the trigger type, right? 

Am I mistaken or is this a problem? I mean, we are lacking a definition of: "How to fill a Container correctly", am I right? I'll open an issue for this.

Of course there is a pragmatical definition. A Container is filled correctly, when some *analysis* can work with it. But this means, I can not implement a test for my EventSource without having an analysis in mind, so I cannot judge when I am done implementing an EventSource, when I do not have a use case.

And .. while writing this ... I realize, that is exactly what I am doing wrong in this PR. There is no analysis (yet) which is using this reader. So there is nothing to measure it against. Thus I have the strong feeling to really call this reader done (as soon as the tests pass again and the Factory-Test-matter is solved) and in a next step really **use** it to find out where it is failing.

watsonjj:
All these different functions with similar names are confusing to follow, I think it would be easier to follow if the containers were filled in the main body of `_generator`.

The issue may very well be that the Fields are not properly inherited, therefore `SST1M_DataContainer` does not have an `r0` container. The way I would check this is to jump into the code, and look at the contents of an instance of SST1M_DataContainer.

I don't know if you are aware of this approach, but something I do when debugging is add the lines:

```python
from IPython import embed
embed()
```

To the point I wish to jump in, and then manually execute the test function:

```bash
ipython
from ctapipe.io.tests.test_zfits import test_loop_over_events
test_loop_over_events()
```

ctapipe/io/zfits.py
watsonjj: You could circumvent this by setting a member variable to the module in __init__, and then use that member variable in the `_generator` method instead of importing again.

watsonjj: Alternatively you could use importlib:

```python
import importlib
spec = importlib.util.find_spec("protozfitsreader")
found = spam_spec is not None
```

watsonjj:
Also to ensure `EventSourceFactory` can find `ZFitsEventSource` it needs to be imported at the top of eventsourcefactory.py (another reason to add a test to make sure `EventSourceFactory` correctly returns `ZFitsEventSource`)

dneise:
> The issue may very well be that the Fields are not properly inherited, therefore SST1M_DataContainer

That is indeed the case, but the question is **why**. I have never heard he expression: "not properly inherited".

dneise:
Okay found it. So the Container does inherit the fields somehow, but it does not assign any values to them. 
So when I tried to give a reference to `c.r0` to a function, which was supposed to fill this `c.r0`, `c` had not yet a member `r0` which I could get a reference to. 

So Now I **assign** a freshly filled `r0` to `c`. This I like much better and it has  the advantage that it works. :-D

watsonjj:
> That is indeed the case, but the question is why. I have never heard he expression: "not properly inherited".

No idea unfortunately. I've worked around it in the past because I assumed it was expected behaviour, but that clearly isn't the case. I think an issue should be created addressing this odd behaviour.

watsonjj:
> So Now I assign a freshly filled r0 to c

I don't think this is a full fix. The `SST1M_DataContainer` doesn't have all the other containers as `DataContainer` (such as r1). I think the containers need to be investigated as to why the Fields aren't inherited. But as a work around I would suggest filling `SST1M_DataContainer` with all the Fields you would expect:

```python
class SST1M_DataContainer(Container):
    r0 = Field(R0Container(), "Raw Data")
    r1 = Field(R1Container(), "R1 Calibrated Data")
    dl0 = Field(DL0Container(), "DL0 Data Volume Reduced Data")
    dl1 = Field(DL1Container(), "DL1 Calibrated image")
    dl2 = Field(ReconstructedContainer(), "Reconstructed Shower Information")
    trig = Field(CentralTriggerContainer(), "central trigger information")
    count = Field(0, "number of events processed")
    inst = Field(InstrumentContainer(), "instrumental information (deprecated")

   sst1m = Field(SST1MContainer(), "SST1M Specific Information")
```

watsonjj:
Additionally the current way you add the r0 container doesn't obey the convention expected by the Container class. I don't know exactly what this means, but could likely result in unexpected behaviour, such as the Field being read only.

dneise:
> I think the containers need to be investigated as to why the Fields aren't inherited. 
yes but not in this PR.

dneise:
> But as a work around I would suggest filling SST1M_DataContainer with all the Fields you would expect

Sounds horrible ...

watsonjj:
> yes but not in this PR.

Absolutely, but the workaround you use should at least adhere to the proper way Fields are used.

dneise:
Just fixing some Codacy issues ... I think @kosack still needs to comment, if there are still some changes to be made.

kosack:
> Obviously it is okay to fill Containers only partly. I mean, no EventSource has all the information already in hand, right? But on the other hand, it is obviously a bug, when an EventSource fills the waveform, but omits filling the trigger type, right?

That's a good question - I just added a few fields like trigger type, but didn't update `HESSIOEventSource`, so it's my fault. However, in reality we will have "optional" fields. We could mark which ones are required, and have a check if an event source doesn't do it correctly, or we can just assume all are optional, and then you have to always check for `None`...  

For the case of `trigger_type` once we have a good definition of it, we could just write the correct one to the container (it would just be a hard-coded value for monte-carlo sims probably)

kosack:
Is there a way we can get rid of `SST1M_DataContainer`? THe idea of the containers was to have a single common way for all telescopes, so having this class invites bad code to be written (I don't want a class per camera).  Could we simply add the required fields to the common classes and have them optionally filled?

kosack:
Also, ZFITS is currently used by the testbench of at least *NectarCam* and others I think. Will this code work for their files, or is it very SST-1M specific?  I can see if we can get a sample file from them perhaps.

watsonjj:
> Is there a way we can get rid of SST1M_DataContainer? THe idea of the containers was to have a single common way for all telescopes, so having this class invites bad code to be written (I don't want a class per camera). Could we simply add the required fields to the common classes and have them optionally filled?

I do worry that this will pollute the common Containers with optional Fields that are almost always empty and become potentially confusing to users as why they exist.

My intention for these additional containers was for them to contain Fields used in commissioning-debugging and R1 Calibration, tasks that will not be carried out in normal ctapipe operation, and the Fields only exist for ease of access for expert scripts that specifically only use that specific `EventSource`.

There are some Fields inside `SST1MCameraContainer` that should be moved to the common Containers, such as gps_time, which a field already exists for in CentralTriggerContainer.

kosack:
Can we just add then a `SST1MTestBenchData` container that contains everything *else* that is not in the main event data and put it in the DataContainer? 

I'd prefer to simply have that optionally in the main event container (unfilled) rather than use sub-classing, which will get confusing. E.g. all `EventSources` should return  `DataContainers` (which soon will be renamed back to EventContainer, once we remove the Instrument info), and that may contain an optional `SST1MTestBenchContainer`. Then if there is SST1M data in it, it would fill in the extra test bench container. We can do the same for other test benches. 

The main issue is that ZFITS as I mentioned already is used for several telescopes (but with different contents), so the ZFITSEventSource will need to decide which one is being read and fill in some extra data, depending on which telescope is used, but it should still work to run normal ctapipe algorithms, if one doesn't care about the extra info.  

Or perhaps I just don't get how the subclasses are being used?

MaxNoe:
Maybe add a `camera_specific` map field to the container?

watsonjj:
By creating the subclass of `DataContainer` (`SST1M_DataContainer`), we were essentially implementing the "DataContainer may contain an an optional SST1MTestBenchContainer". If a user was to look at the DataContainer from a hessio file they would see:

```
ctapipe.io.containers.DataContainer:
                          r0.*: Raw Data
                          r1.*: R1 Calibrated Data
                         dl0.*: DL0 Data Volume Reduced Data
                         dl1.*: DL1 Calibrated image
                         dl2.*: Reconstructed Shower Information
                          mc.*: Monte-Carlo data
                    mcheader.*: Monte-Carlo run header data
                        trig.*: central trigger information
                         count: number of events processed
                        inst.*: instrumental information (deprecated
                   pointing[*]: Telescope pointing positions
```

But if the file was a ZfitsEventSource:

```
ctapipe.io.containers.**SST1M_DataContainer**:
                          r0.*: Raw Data
                          r1.*: R1 Calibrated Data
                         dl0.*: DL0 Data Volume Reduced Data
                         dl1.*: DL1 Calibrated image
                         dl2.*: Reconstructed Shower Information
                          mc.*: Monte-Carlo data
                    mcheader.*: Monte-Carlo run header data
                        trig.*: central trigger information
                         count: number of events processed
                        inst.*: instrumental information (deprecated
                   pointing[*]: Telescope pointing positions
                       **sst1m.*: SST1M Specific Information**
```

The only difference is the additional optional container.

watsonjj:
We could change it so that the sst1m container always exists, but is unfilled when reading other file formats.

dneise:
I personally do not mind, if the SST1M specific part is inserted by means of inheritance or by means of composition.

In both cases, this can happen:

 * User does not expect the SST1M specific stuff to exists, and is confused by its existence.
 * A user expects the SST1M specific stuff to exist, but is reading a file lacking this information, so the analysis will raise a KeyError-like-Exception at some point.

I am a friend of Composition over Inheritance, similar to Karl. 

I think, we are at a point where we need a decision, and have to live with it until we receive a feature request, which cannot be implemented in the current design, so re-design is forced. Or until we have 3...5 of these telescope specific containers and see a pattern emerge. 

I can already tell you though, that this current version of the PR is **minimalistic** in the sense that not all the fields my friends from SST1M need, are present in this container.

However, I felt unable to justify the existence of fields which are very camera specific, will not yet be filled by the reader, and might not even fit into the current design.

watsonjj:
Very well, lets switch over to "Composition" and have the Container classes kept to containers.py :)

dneise:
I think these camera containers should be close to the thing, which is creating/filling them.
Proximity in scope can help people, to not overlook fields.

watsonjj:
I have no preference

dneise:
Sounds like you have.

Anyway. I would like to have a decision.

dneise:
> The main issue is that ZFITS as I mentioned already is used for several telescopes (but with different contents), so the ZFITSEventSource will need to decide which one is being read and fill in some extra data, depending on which telescope is used, but it should still work to run normal ctapipe algorithms, if one doesn't care about the extra info.

I think we should cross that bridge when we come to it. In preparation, I renamed this EventSource to `SST1MEventSource`, so that its specificity becomes clear. 

When we have more than one EventSource which are using the `protozfitsreader`, I think we should have a look where these share code and refactor accordingly.

watsonjj:
I have no strong feeling about where it should be placed, either seems reasonable.

dneise:
I will study how it looks ... so now I will follow Karls proposal and add a SST1M specific thing to `DataContainer` ... let's see how it looks.

dneise:
Okay. Now it should look more like what @kosack  had in mind. 

I had to move the Containers really into `io.containers.py`. Otherwise I had a circular import.
This way everybody who wants to see what might be in a container, can look into `containers.py` and does not have to search through all possible readers ... that's a pro I guess.

ctapipe/io/eventsourcefactory.py
```diff
@@ -4,6 +4,7 @@
 # EventFileReader imports so that EventFileReaderFactory can see them
 # (they need to exist in the global namespace)
 import ctapipe.io.hessioeventsource
+from .import zfits
```
watsonjj: Could you fix this import?

watsonjj: Sorry, what I meant was could you rename zfits.py to zfitseventsource.py, and use an absolute reference to the module `import ctapipe.io.zfitseventsource`?

dneise: rename file: very good.. I will ... better even to sst1meventsource.py, since I renamed the class to not imply any generality which might not exist.

dneise: absolute import .. why?

kosack:
>I had to move the Containers really into io.containers.py. Otherwise I had a circular import.
This way everybody who wants to see what might be in a container, can look into containers.py and does not have to search through all possible readers ... that's a pro I guess.

I'm not totally against using subclasses now that i look closer at how you are implementing it. As long as normal code that uses only the datacontainer still works, having the extra fields as sub-classes should be ok. I had though you had completely different return values from the ZFITSEventSource.

dneise:
I like both. What do you like best? I'll do it like that.

kosack:
It is nice to have it all in io/containers.py (at least for now), even if you use subclasses, so we can eventually have a global view and *hopefully* in the future, unify the outputs from all testbenches and cameras as much as possible. 

So I'd say, you can continue with using a subclass to add extra camera-specific fields, but put those subclasses in the same file so we can keep track.

dneise:
okay, thanks.

dneise:
One other thing, I think #653 should be merged asap. All tests seem to be failing due to that, right?

dneise:
Okay so ... can we merge this as soon as the test stop failing?

FrancaCassol:
Hi to all,
I see a big activity about the SST1MEventSource implementation, should I wait that all this is committed to the master in order to take it as example for a NectarCamEventSource implementation? (I spent some time to get a local Linux machine correctly installed for these developments)

dneise:
I think this PR here is close to be merged. So no need for you to wait, you can just go ahead.
I personally would be very interested to see your proposal.

dneise:
So ... tests are passing. Codacy complains mostly about things I was explicitely requested to do, so.. I do not see how to improve that at the moment.

Issue #594, opened 2017-12-04
What is the purpose of the `name` class.variable?


dneise:
I see in some classes something like this
```python
class XXX:
    name = 'XXX'
    # ...put implementation here ... 
```

So in this case `XXX.name == XXX.__class__.__name__`, but there must be some semantic difference between `self.__class__.__name__` and this class variable `name`.

Could you please clarify this?

An example would be: https://github.com/cta-observatory/ctapipe/blob/master/ctapipe/calib/camera/r1.py#L172

kosack:
those classes are sub-classes of `traitlets.config.Application` that asks for a name (which may be different from the class name).  I don't think it's required, and is somewhat confusing.  The only place I've really used it is when writing a command-line tool, to name it the same as the command-line rather than the class (e.g. there may be a Tool class called `MyTool`, but I would set the name to `ctapipe-mytool` so the `--help` message comes out right).  Otherwise, it doesn't seem to be used anywhere. 

See the docs for `traitlets.config` - there are  a number of "standard" class attributes that change the behavior of a traitlets-based class where configuration and provenance is tracked.

kosack:
see for example:

http://traitlets.readthedocs.io/en/stable/config.html

dneise:
I'm sorry, I do not understand the answer. You say these are all subclasses of `traitlets.config.Application`, but the example I've shown is not a subclass of that:

The example I've shown is this

https://github.com/cta-observatory/ctapipe/blob/89e82e3be4dff7d728fb55213655d199f1e71487/ctapipe/calib/camera/r1.py#L172

and this is a subclass of `traitlets.config.Configurable` not `Application`.

dneise:
You further say:

>  there are a number of "standard" class attributes that change the behavior of a traitlets-based class where configuration and provenance is tracked.

I do not find any mention of such standard class attributes in the traitlets documentation. On the page you nicely provided, the only place where a class attribute `name` is mentioned is here

http://traitlets.readthedocs.io/en/stable/config.html#python-configuration-files

The example looks like this:
```python
class MyClass(Configurable):
    name = Unicode(u'defaultname'
        help="the name of the object"
    ).tag(config=True)
    ranking = Integer(0, help="the class's ranking").tag(config=True)
    value = Float(99.0)
    # The rest of the class implementation would go here..
```

but I did not find anything in the text, that says, that `name` plays some kind of special role ... 
..confused....

dneise:
[...]

PR #599, opened 2017-12-08
Remove name from components

dneise:
This PR follows the discussion in #594. 
It was decided to remove the class variable `name` from Components, when it is not needed.

For factories I replaced occurrences of `self.name` with `self.__class__.__name__`


dneise:
In a sense, this is only a cosmetic PR, but the `name`s puzzled be a lot, so removing them might future readers of the code.

PR #607, opened 2017-12-13
better error handling for OpticsDescription

### LINKED ISSUE #605: instrument.optics: Does this test contradict the implementation?
Tests serve a docu to me, so I tend to believe what ever the test says is right, but here I was a bit puzzled:

**test**
```python
    with pytest.raises(KeyError):
        OpticsDescription.guess(0*u.m)
```

This seems clear, when we give a wrong focal length, we get a KeyError, plain and simple.

----
**implementation**: abbreviated
```python
class OpticsDescription
    @classmethod
    def guess(cls, equivalent_focal_length):
        tel_type = telescope_info_from_metadata(equivalent_focal_length)
        if tel_type == "unknown":
            logger.warning("setting tel_type to unknown")
        return cls(tel_type, equivalent_focal_length)
```
No key error ... just a warning.

Maybe deeper?
```python
def telescope_info_from_metadata(focal_length):
    return _FOCLEN_TO_TEL_INFO[round(focal_length.to('m').value, 2)]
```

yes, this raises the KeyError. It can not return the string "unknown". 
If it would return the string "unknown" instead of throwing KeyError, e.g. like this

```python
def telescope_info_from_metadata(focal_length):
    try:
        return _FOCLEN_TO_TEL_INFO[round(focal_length.to('m').value, 2)]
    except KeyError:
        return "unknown"
```

Then the "unknown" would bubble up into the `__init__`:
```python
    def __init__(self, tel_type, ...):
        if tel_type not in ['LST', 'MST', 'SST']:
            raise ValueError("Unknown tel_type %s", tel_type)
```

and throw a `ValueError` not a KeyError. 

-----

So the test says: on wrong key --> KeyError
implementation suggests ... could also be ValueError

What was the authors intention?

kosack:
- guess() now consistently throws either a KeyError (telescope wasn't found) or a ValueError (bad input), no longer "works" on bad key
- added more tests
- fixes #605

### DIFF
--- a/ctapipe/instrument/optics.py
+++ b/ctapipe/instrument/optics.py
@@ -5,6 +5,7 @@ Classes and functions related to telescope Optics
 import logging
 from ..utils import get_table_dataset
 import numpy as np
+import astropy.units as u
 
 logger = logging.getLogger(__name__)
 
@@ -43,14 +44,23 @@ class OpticsDescription:
     equivalent_focal_length: Quantity(float)
         effective focal-length of telescope, independent of which type of
         optics (as in the Monte-Carlo)
+
+    Raises
+    ------
+    ValueError:
+        if tel_type or mirror_type are not one of the accepted values
     """
 
+    @u.quantity_input
     def __init__(self, mirror_type, tel_type,
-                 tel_subtype, equivalent_focal_length,
-                 mirror_area=None, num_mirror_tiles=None):
+                 tel_subtype, equivalent_focal_length : u.m,
+                 mirror_area : u.m**2=None, num_mirror_tiles=None):
 
         if tel_type not in ['LST', 'MST', 'SST']:
             raise ValueError("Unknown tel_type %s", tel_type)
+
+        if mirror_type not in ['SC', 'DC']:
+            raise ValueError("Unknown mirror_type: %s", mirror_type)
 
         self.mirror_type = mirror_type
         self.tel_type = tel_type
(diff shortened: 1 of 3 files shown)


ctapipe/instrument/optics.py
```diff
@@ -43,15 +44,24 @@ class OpticsDescription:
     equivalent_focal_length: Quantity(float)
         effective focal-length of telescope, independent of which type of
         optics (as in the Monte-Carlo)
+
+    Raises
+    ------
+    ValueError:
+        if tel_type or mirror_type are not one of the accepted values
```
dneise: Can also raise UnitsError or TypeError

dneise:
One questions, what if the focal length "contradicts" the (tel_type, mirror_type, tel_subtype) tuple ?

ctapipe/instrument/optics.py
kosack: done. Also documented all members (some were missing)

kosack:
> One questions, what if the focal length "contradicts" the (tel_type, mirror_type, tel_subtype) tuple ?

I'm not sure I understand the question---you mean if constructed from  `__init__()`? In principle that is ok:  you can define any optics you like, they don't have to agree with what is in the lookup table, which is only there because the info is not stored in the monte-carlo files, so it has to be "guessed". If we had a more complete MC data file format that includes this information in a machine readable way, we could always construct from what is in the file.    In principle right now we even need to update these values for each MC production, since they may be adjusted.

Issue #590, opened 2017-11-29
io: reader for fits.fz files containing protobufs compressed with CTA compression


dneise:
Hello, 
yesterday in the SST1M analysis meeting it was agreed that SST1M wants to contribute as much as possible to ctapipe, and to gain as much as possible from ctapipe. 

SST1Ms raw data format is "fits.fz". There exists a dedicated reader for this file format, including Python bindings.It is maintained by @Etienne12345. 

The reader can for example be installed like this:

    conda create -n test_protozfits python=3.5
    source activate test_protozfits
    conda install numpy
    pip install http://www.isdc.unige.ch/~lyard/repo/ProtoZFitsReader-0.42.Python3.5.Linux.x86_64.tar.gz
    export LD_LIBRARY_PATH=$CONDA_PREFIX/lib/python3.5/site-packages:$LD_LIBRARY_PATH

And then tested (SST1M is currently looking into the possibility to add small example files to ctapipe-extra), like this:

    import protozfitsreader as pz
    f = pz.ZFile(path)
    # to stuff with f

Now the question is, how should one proceed to integrate this reader into the io package of ctapipe.

dneise:
Ah ... I just see there is already an open PR #340 about this issue. 

I'll review that one now, in order to understand what has already been done.

dneise:
We just talked to @Etienne12345 on the phone. He said he is planning to publish the reader as a conda package. I think this solves this issue. I just would keep it open until the package is ready.

watsonjj:
I believe this raises an important issue. Camera teams want to start reading their data into ctapipe, and members of the MC validation teams want to use ctapipe for common processing of the data between MC and real. However with the lack of a common data format, each camera team wants to read in their own custom data format.

However, as these data formats are unofficial and will not be maintained by ctapipe in the future, I have been hesitant adding these readers into the main ctapipe repository. I have not included the ctapipe reader for CHEC in the main repository.

I am starting to change my opinion on this approach. With these readers hidden away, camera teams that are new to ctapipe have no reference on how they should do it. Therefore I propose creating a ctapipe/io/unofficial directory, where I will write guidelines on creating a reader for a cameras custom data type, and where the readers will be stored. This will keep some order among the unofficial readers, and will be retired once the common data type is defined.

Karl, does this seem like an acceptable approach?

watsonjj:
Also in the guidelines I would advise camera teams to include installation instructions for their custom software in the documentation of their ctapipe reader.

[...]

PR #610, opened 2017-12-13
Guidelines for Unofficial Event Data Format File Readers

watsonjj:
I have created a module `ctapipe.io.unofficial` for the unofficial data format readers. Inside this module I have included the TargetioEventReader (used to read in CHEC’s data format), which is conditionally instanced based on if the machine has the required external packages imported.

Inside the ctapipe documentation I have outlined how one should create their own EventFileReader for a data format they wish to read into ctapipe, using TargetioEventReader for reference.

These guidelines are being created as a result of the discussion in #590.

I am open to feedback on these guidelines.

Things left to do for this pull request:

* add the installation method to .travis.yml so that the Travis CI can run the optional tests that require the external libraries. (Awaiting on cta-observatory/ctapipe-extra#9 so I have files I can test against on the Travis CI)
* ~~incorporate changes pending from other PRs (#609, #599)~~


MaxNoe:
This PR does way to much for such a simple thing as adding new guidelines!

dneise:
I have the same impression as Max it is quite big.

I mean, Jason did not only write guidelines, but he also delivers an example implementation. And I think this is okay.

But there is also code in here, which has in my understanding not too much to do with an event source.

watsonjj:
I admit there are a few small changes I made to EventFileSource and Factory to prepare them to be more widely used...

The majority of the code I have included is the TargetIOEventFileReader, and the low level reader and CameraR1Calibrator for that code, which as Dominik states, are an example implementation following the guidelines, an important inclusion.

MaxNoe:
I would like to first discuss these guidelines. It's very hard to do this on a PR with more than 1700 lines of diff and 30 files changed.

dneise:
> I admit there are a few small changes I made to EventFileReader and Factory to prepare them to be more widely used...

I find this confusing .. even *very* confusing... I must admit... There are some files touched in this PR, which creates fear in me. 

 -  ctapipe/core/factory.py
 -  ctapipe/core/tests/test_factory.py 
 -  ctapipe/tools/extract_charge_resolution.py
 -  ctapipe/calib/camera/calibrator.py
 -  ctapipe/calib/camera/r1.py
 -  ctapipe/calib/camera/tests/test_r1.py
 -  ctapipe/calib/camera/tests/test_calibrator.py

dneise:
Maybe we ... who want to review this PR should limit ourselves first to reviewing only the `*.rst`files. And ignore the modifications in the `*.py` files to make our task managable.

watsonjj:
> I find this confusing .. even very confusing... I must admit... There are some files touched in this PR, which creates fear in me.

Let me try to quell those fears... 

Firstly these are all modules I created. That does not dismiss the need for a proper review, but I at least am aware of their design and purpose.

> ctapipe/core/factory.py

I've added one new method, which allows the user to get the product of the factory with one line, instead of 3. This method does exactly what the 3 lines did. Any other changes are from your PR on removal of `name`.

> ctapipe/core/tests/test_factory.py

Test of new method

> ctapipe/tools/extract_charge_resolution.py

Use new factory produce method instead of old 3 line method

> ctapipe/calib/camera/calibrator.py

Use new factory produce method instead of old 3 line method

> ctapipe/calib/camera/r1.py

Create the new calibrator required for the R0 level calibration for files read from the targetio data format

> ctapipe/calib/camera/tests/test_r1.py

Test the new calibrator

docs/io/index.rst
```diff
@@ -9,20 +9,27 @@
...
+* EventFileReaders - High level class used to read from the file
+* hessio_event_source - Low level method to handle the filling of the
+  containers
 
 
 Container Classes
 =================
 
 Event data that is intended to be read or written from files is stored
 in subclasses of `ctapipe.core.Container`, the structre of which is
```
dneise: Here is a typo: "structre" --> "structure".

dneise:
I think the `check_modules_installed` story can be simplyfied in two steps:

1.) "Ask forgiveness not permission"

```python
necessary_modules = ['foo', 'bar', 'baz']
if check_modules_installed(necessary_modules):
    import foo
    import bar 
    import baz

    class FooBarBazDependend:
        pass
```

Is doing the same as:
```python
try:
    class FooBarBazDependend:
       import foo
       import bar 
       import baz

        pass
except ModuleNotFoundError:
    pass
```

But, there is less duplication, e.g. the string list of dependencies is not needed to be defined.
The entire function `check_modules_installed` did not need to be implemented, so some developer just got some free hours to play with their kids!

----

There is more .. .but let me write that into a 2nd post

docs/io/index_unofficial.rst
```diff
@@ -0,0 +1,232 @@
...
+or calling a method from the external reader library to check that the file is
+compatible.
+
+Additionally, each EventFileReader should contain a property to return the
+name of the `ctapipe.calib.camera.r1.CameraR1Calibrator` that should be used
+for calibrating the events. The CameraR1Calibrator that should be chosen
+depends on the source of the events. If a data format is to be used by
+different cameras that require different R1 calibration, then this method
+should correctly figure out the right one. If the data source is from data
+level R1 or above, return None or 'NullR1Calibrator'.
```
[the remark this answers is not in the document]
watsonjj: Yes, you are right that generic "Calibration" can be very confusing. But in these guidelines I try to make sure I specify "R1 Calibration", that is, the low-level camera specific calibration that is required to transform camera data from data level R0 to R1. I will try to emphasise this better.

You are right that its not immediately obvious why you need to consider calibration if all you want to do is read your data... I will try to reduce the emphasis on creating a calibrator, and just try to remind the user that they should read their data into the correct data level container, and that they can create a CameraR1Calibrator if they wish to perform their R1 Calibration.

docs/io/index_unofficial.rst
```diff
@@ -0,0 +1,232 @@
...
+Each unofficial data format can be wildly different, therefore these low-level
+modules can appear very different in how they function. This is fine as long
+as some guidelines are followed:
+
+* The `ctapipe.io.eventfilereader.EventFileReader` calls the low-level methods
+  such that the high-level interface remains the same as other
+  `ctapipe.io.eventfilereader.EventFileReader`.
+* The low-level methods fill the containers with the correct data, in a
+  similar way to other low-level reader methods (see `ctapipe.io.hessio` and
+  `ctapipe.io.unofficial.targetio.targetio` for examples)
```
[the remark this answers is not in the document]
watsonjj: Technically it should be... I'd like to hear input from others about that

watsonjj:
> I think the check_modules_installed story can be simplyfied in two steps:

Very valid suggestion...I will adopt this instead

dneise:
2.) "Help people to help themselves"

What happens at the moment if a person wants to read a file, but does not have the low level readers installed, in fact that person did not even know ctapipe does not bring all readers out of the box. (No Batteries included?)

The person gets a `ValueError` from here. 
https://github.com/cta-observatory/ctapipe/blob/1de13ea78823a4dfd1d49eb85bdb4efd2e24a10d/ctapipe/io/eventfilereader.py#L402-L416

ValueError? The person is missing a dependency, shouldn't this be a `ModuleNotFoundError`? 

Next .. what module is she missing? Do we give her any advice what she should google for? No. Why not? Don't we like our fellow developers enough to give them at least a little hint?

If we however implement the event file readers not like this:
```python
necessary_modules = ['foo', 'bar', 'baz']
if check_modules_installed(necessary_modules):
    import foo
    import bar 
    import baz

    class FooBarBazDependend:
        pass
```

but like this:
```
class FooBarBazDependend:
    def __init__(self, stuff, stuff, *args, **kwargs):
        import foo
        import bar 
        import baz
```

What happens now, when somebody tries to read a file, which just by chance needs `FooBarBazDependend`? The factory still knows, that `FooBarBazDependend` exists ... and it also still tries to make an instance of `FooBarBazDependend` in its `produce` method. But  at this point it fails, it cannot make an instance of `FooBarBazDependend` since the import statements result in a `ModuleNotFoundError`. The traceback explicitely says the name of the module which is missing. 

Now the person can head to the ctapipe issue tracker and seach for 
    
    ModuleNotFoundError: No module named 'bar'

And .. indeed there was somebody else before her, who had the same problem, she found the solution in 5 minutes ... she is still unhappy, we did not include our batteries, but at least she could solve the problem herself in reasonable time.

And look how much code we had to write, to support this feature. None. That is, less lines to be read, to be maintained, to be reviewed, to be understood.

watsonjj:
> What happens at the moment if a person wants to read a file, but does not have the low level readers installed

I understand the problem you are outlining, and I wish to find a solution, but I'm not sure this is it.

There are many different file readers, and we don't want them to add any dependency to ctapipe. The EventFileReaderFactory loops through the different types of EventFileReader and checks if that reader is compatible with a staticmethod. If that EventFileReader is not compatible, we wish to move onto the next EventFileReader without putting up any fuss. 

Your approach does work, except if the reader relies on a method from their external library to assess if a file is compatible with their reader. An error should not be raised to the terminal if the external library does not exist in this circumstance, as that reader is optional, and may not even be the intended reader for the file, so it should gracefully move onto the next EventFileReader to check if it is compatible.

watsonjj:
> ValueError? The person is missing a dependency, shouldn't this be a ModuleNotFoundError?

No, the user may have all the dependencies installed, but the file they have passed to EventFileReaderFactory is not compatible with any of the EventFileReaders.

watsonjj:
I think the best approach to handle this is not to raise a `ModuleNotFoundError`. And to keep it as it is and inform the user that "Cannot find compatible EventFileReader for:...".

To combat the problem you have described, we should include information about additional dependencies to read in other file formats in the installation instructions for ctapipe. That way the user is aware from the start of what they may need to do if they wish to read other data formats.

dneise:
>>  ValueError? The person is missing a dependency, shouldn't this be a ModuleNotFoundError?

> No, the user may have all the dependencies installed, but the file they have passed to EventFileReaderFactory is not compatible with any of the EventFileReaders.

How can these two cases be distinguished here?

dneise:
> There are many different file readers, and we don't want them to add any dependency to ctapipe. 

I got that.

> The EventFileReaderFactory loops through the different types of EventFileReader and checks if that reader is compatible with a staticmethod. If that EventFileReader is not compatible, we wish to move onto the next EventFileReader without putting up any fuss.

https://github.com/watsonjj/ctapipe/blob/b1dcd0a9d3da91e2bf868df4a31d793ed2ea55ec/ctapipe/io/unofficial/eventfilereader.py#L66-L74

Yes, you mean this part. I just realize the static method is checking if the dependencies are installed. 
Aren't these two different questions?

1.) Which reader can read this file?

and 

2.) Do I have all the dependencies installed for this reader?

watsonjj:
> >>  ValueError? The person is missing a dependency, shouldn't this be a ModuleNotFoundError?
> 
> > No, the user may have all the dependencies installed, but the file they have passed to EventFileReaderFactory is not compatible with any of the EventFileReaders.
> 
> How can these two cases be distinguished here? 

I don't think they can, the approach of checking if a file is compatible could be to call a method from the external library (as I've stated above). If that method cannot be called because that depenceny isn't installed, then I want it to treat that EventFileReader as incompatible and move on. The most common case will be that EventFileReader wasn't intended for my file anyway (as I never installed those dependencies).

dneise:
It is a static method, so in this proposal, 

```python
class FooBarBazDependend:
    def __init__(self, stuff, stuff, *args, **kwargs):
        import foo
        import bar 
        import baz
    @staticmethod
    def is_compatible(path):
        return path.endswith('.foo')
```

You can still loop over all classes and find out which one is compatible "without making a fuss*

watsonjj:
> It is a static method, so in this proposal, 
> 
> ```python
> class FooBarBazDependend:
>     def __init__(self, stuff, stuff, *args, **kwargs):
>         import foo
>         import bar 
>         import baz
>     @staticmethod
>     def is_compatible(path):
>         return path.endswith('.foo')
> ```
> 
> You can still loop over all classes and find out which one is compatible "without making a fuss*

Yes, but a different EventFileReader may want to do something like: 

```python
class FooBarBazDependend:
    def __init__(self, stuff, stuff, *args, **kwargs):
        import foo
        import bar 
        import baz
    @staticmethod
    def is_compatible(path):
        import foo
        compatible = foo.check_file_is_compatible(path)
        return compatible
```

Where `foo.check_file_is_compatible` does some operation that is specific to the data format to check if it is the correct format for the reader. Unless you suggest we outright disallow that?

dneise:
> Yes, but a different EventFileReader may want to do something like:

Yes, they might. But this is against the contract I was trying to define here.

I was trying to change this part: "use `if check_modules_installed(list_of_special_deps)`"
with this part: " `import` special deps within `__init__` only."

But it stays a contract, and if developers decide to break such contracts, it does not matter how it was phrased. 

We can only protect ourselves from such problems by testing this code, both with the necessary deps installed to see that it works, and **without** the deps installed to see that it breaks exactly in the way we want it to break.

watsonjj:
I worry that you might not be fully seeing my point.

Inside `EventFileReader.check_file_compatibility` the writer of the reader should define _some_ method to decide if a file is compatible with their reader. 

For targetio and hessio this is simple: we look if the filepath ends with the expected extension.

However, perhaps two different file formats have the same file extension, but read in the data differently. They would need to define a `EventFileReader.check_file_compatibility` that distinguishes between the two.

If it was a fits file format, you could check the headers for a particular indicator of the file format. This could be done entirely in python, so your proposal is still valid.

However, if its a less common file format, then the method to really be able to distinguish if it is a compatible file involves reading the file using the external library, and checking it was readable (in some fast way). Therefore in this situation the `EventFileReader.check_file_compatibility` method needs to import the external library, which will throw an error if the library is not installed, and break the EventFileReaderFactory loop. This is fine, except you want to separate between "file compatibility" and "dependencies not installed", when in this situation you can't.

My arguments against your approach are just future proofing, and "what-ifs". But I want to make sure you understand where my concerns are before we adopt your proposal.

dneise:
I think, I forgot some remarks before. These are personal opinions, but I phrase them like ultimate truths. But they are just opinions.

### `try-init-fail` not `is_compatible`
I personally think, the most clear way, to see if a file can be read with a certain reader, is to apply that reader to that file. So `Reader.is_compatible(some_path)`, I think is not as good as simply `Reader(some_path)` if it does not throw --> I can use that `Reader` for `some_path`.

It can happen, and I am sure it will happen, that `is_compatible` will return True, but actually the file cannot be read with that reader. 

**But** trying out all readers on a path, is a) slow and b) does not work at all, when all readers are not installed. 

### filenames have nothing to do with file contents. 
I think, using filenames, to guess if a file contains certain data, does in general not work. It works sometimes .. and sometimes means "very often" on windows, hence people think, it works. But it does not. 

-----

I conclude from these two opinions: 

We cannot find out if a reader is compatible with a file, if that reader is not already installed. And therefore the EventRederFactory cannot work without all readers being installed. 

It will simply tell users there is no reader for their file, but not that actually there is a reader, which they forgot to install. 

One can try to **copy** parts of the low level readers, which are able to find out if a certain file can be read by a certain reader which is not yet installed, into the high-level reader, which will always be installed. But this is code duplication, via package boundaries. When a low level reader is suddenly able to read also a different file format the high-level reader will not know about this and keep claiming: "No ... we cannot read that file".

ctapipe/utils/check_modules_installed.py
```diff
@@ -0,0 +1,10 @@
+from importlib.util import find_spec
+
+
+def check_modules_installed(module_name_list):
+    for module_name in module_name_list:
+        found = find_spec(module_name)
+        if found is None:
+            print("modules not found")
```
dneise: This should not `print` anything. It is a boolean function.

dneise: many boolean functions, have names starting with `is_` or maybe `are_` ...

watsonjj:
These are all valid points I've been considering. But that doesn't settle on a solution. I don't think it is feasible to try `EventFileReader(path)` due the the reasons that you mentioned.

I propose:

1. We import the dependencies in `EventFileReader.__init__` of the respective `EventFileReader`.
2. `EventFileReader.check_file_compatibility` performs a **quick** check to see if the file **might** be compatible (e.g. checking file extension). If there becomes a case in the future where a particular file format is passing two `EventFileReader.check_file_compatibility` then we will create a better check for those readers.
3. The `EventFileReader` can pass the compatibility check for the file, but then raises `ModuleNotFoundError` if the dependencies are not installed for that reader, along with the message of which packages need installing to be able to read that file format
4. If the creator of the `EventFileReader` wishes to use a method from their external module to evaluate if the file is compatible, that is against our guidelines, but we will assess that situation at the time. If it is insisted that it must be done that way, then they need to implement it in such a way that if the module is not found while checking for compatibility, then that `EventFileReader` is deemed not-compatible for that file. This removes the helpfull-ness of 3 for that `EventFileReader`.
5. We mention about these unofficial readers in the installation instructions for ctapipe, so users that have come to ctapipe to use their camera's data format are aware of what they should do immediately, and aren't met with surprises.

dneise:
Sounds like the best way, short of simply requesting that all readers need to be pip installable and we simply ship ctapipe with all readers.

dneise:
I have another question.

The camera/telescope specific readers provide camera/telescope specific data, I think some people call this "R0". And I think we already discussed, that these groups can define the layout for their R0 containers. 

So how does a typical usage scenario for the ReaderFactory look like? Given that as soon as I have used the Factory to get my `reader`, which was chosen depending on the path and start iterating over the events, I will get different things, depending on path. So the next thing, which looks at the events, must expect a lot of differently looking Containers.

watsonjj:
> I have another question.
> 
> The camera/telescope specific readers provide camera/telescope specific data, I think some people call this "R0". And I think we already discussed, that these groups can define the layout for their R0 containers. 
> 
> So how does a typical usage scenario for the ReaderFactory look like? Given that as soon as I have used the Factory to get my `reader`, which was chosen depending on the path and start iterating over the events, I will get different things, depending on path. So the next thing, which looks at the events, must expect a lot of differently looking Containers.

I'll start by clarifying some things. "R0 waveforms" are the waveforms obtained from the camera with no processing applied to them. They are raw. "R1 waveforms" are waveforms that have had the camera-specific calibration applied to them to so they are in a state in which they can be used for online analysis. As well as being sent down the online analysis pipeline, these "R1 waveforms" are also stored to disk (perhaps with some data reduction). They are now in the "DL0 waveforms".

DL0 is the format we will read into ctapipe in the final array implementation. However in this prototyping stage we actually save R0 (and perhaps R1 or DL0) files to disk, and we wish to read these data levels into ctapipe for analysis. Once in ctapipe, they are processed up the data level chain (R0 -> R1 -> DL0 -> DL1 ....), but at any stage the waveforms can be accessed.

At the R0 level, as the camera-specific low-level calibration has not been applied yet, some additional information may need to be stored in the container to allow that calibration to be applied. Every Reader's R0Container should contain the same attributes as the base R0Container (e.g. adc_samples -> where the waveforms are stored).

**Therefore anyone looking at R0 data of an unfamiliar camera can still see the waveforms in a common way, and any script designed to perform the analysis on any camera will not be hindered by the different container, because it contains additional information, but not different information. The base, important attributes are the same** This additional information is only important for debugging and calibration.

Once you are at the R1 level, every camera has the same containers.

I hope that explained it...

dneise:
Thank you for the nice explanation. I still don't get the difference between R1 and DL1, but also this was not my question. I must have expressed myself badly. Let me try again with an example.

Assume I need to implement a `Tool`, which is performing some `SST1MSpecificAnalysis` on events. I understand your guidelines so, that I have to do it this way:

```python
for event in EventReaderFactory.produce(path):
    event = SST1MSpecificAnalysis.process(event)
```
Assume I pointed it to a wrong file, one of a different Telescope type. 
This will raise something like `KeyError` or `ValueError`, telling me, that the `DataContainer` does not have a certain field, `SST1MSpecificAnalysis` is looking for. The reason, that I specified a wrong file path is effectively hidden by the fact, that the Factory does its best to give me a reader for that file. 


In case I have the liberty to do something like this:

```python
for event in zfits_event_source(path):
    event = SST1MSpecificAnalysis.process(event)
```

I would get an exception from the `zfits_event_source` telling me that it cannot read that file pointed to by `path`.

----

So I wonder if people like me should use the EventReaderFactory for doing Camera/Telescope specific analysis steps.

ctapipe/io/unofficial/targetio/camera.py
```diff
@@ -0,0 +1,195 @@
...
+
+        Parameters
+        ----------
+        camera_id
+        """
+        Borg.__init__(self)
+        if not self.__dict__:
+            self._id = None
+
+            self.dir = dirname(realpath(__file__))
```
dneise: I think this will not work, when the package is installed.

ctapipe/io/unofficial/targetio/camera.py
```diff
@@ -0,0 +1,195 @@
...
+which differ between CHEC-S, CHEC-M and single module files.
+"""
+import numpy as np
+from astropy import log
+from os.path import dirname, realpath, join
+from target_io import T_SAMPLES_PER_WAVEFORM_BLOCK as N_BLOCKSAMPLES
+from target_calib import MappingModule, MappingCHEC
+
+
+class Borg:
```
dneise: What is `Borg`?

watsonjj:
Firstly, that should not raise an error unless the analysis is looking for something in the R0 container that doesn't exist. I will use targetio as a proper example. If you look at `ctapipe.io.unofficial.targetio.contaners` the R0 container contains the additional attribute `first_cell_ids`. If I had a script that used `EventFileReaderFactory` whose purpose was for debugging `first_cell_ids` but I passed it a different file format than targetio, then an error (`KeyError` or `ValueError`) would be raised when it tries to access `first_cell_ids`.

As this is a very specific script in which is only intended to be used with targetio files, I would not use an `EventFileReaderFactory`, and would instead directly create an instance of `TargetioFileReader`.

Hopefully that make sense. 

Where it gets complicated though is in the R1 calibration. We want to have scripts that can read in R0 data, perform the R1 calibration, and then go on to do charge extraction and hillas parameterisation. And we want these scripts to be compatible with all file formats. Therefore we need to make sure the correct `CameraR1Calibrator` is chosen, which expects the correct fields in the R0 container. Therefore in this case we use both `EventFileReaderFactory` and `CameraR1CalibratorFactory`.

Which `CameraR1Calibrator` is returned from `CameraR1CalibratorFactory` can be decided by the user from the cmdline, or can be automatically chosen by passing `CameraR1CalibratorFactory` the `EventFileReader.r1_calibrator` attribute. To see an example of this implementation take a look at examples/calibration_pipeline.py in this PR.

ctapipe/io/unofficial/targetio/camera.py
watsonjj: It looks in the same directory as the python file for the numpy file containing the pixel positions. I think that does work when it is installed, as it also works when looking for the header file for c scripts such as `ctapipe.utils.neighbour_sum`.

ctapipe/io/unofficial/targetio/camera.py
watsonjj: So this is a class that essentially makes the Config a global object. Where-ever a new version of Config is created, it has the same settings as the others.

Its admittedly a horrible approach. But we have 2 prototype cameras, with 2 different configuration, with no way to tell them apart from the file alone. This allowed me to easily change the config for particular scripts without changing anything inside `TargetioFileReader`.

watsonjj: But as with anything global, it sets off alarm bells because of traceability. I was hoping as it only really affects pixel positions, it will only be fiddled with by CHEC people, and it only exists temporarily as we transition to our newer camera, that it could be accepted. If not, then I'll have to change it :(

dneise: Is the name a reference to Star Trek? Or is it an abbreviation, which is unknown to me? 
If it is a global config object, maybe one could call it `GlobalConfigObject` or so?

watsonjj: Not sure, but probably. Its talked about on various blogs/stackexchange and I copied it:
http://code.activestate.com/recipes/66531-singleton-we-dont-need-no-stinkin-singleton-the-bo/
https://www.safaribooksonline.com/library/view/python-cookbook/0596001673/ch05s23.html

ctapipe/io/unofficial/targetio/camera.py
dneise: To my understanding none of the package data files will be installed and thus also will not be found, usually people use `pkg_resources` to find package data.
c.f. http://setuptools.readthedocs.io/en/latest/setuptools.html#including-data-files

dneise:
> Firstly, that should not raise an error unless the analysis is looking for something in the R0 container that doesn't exist. I will use targetio as a proper example. If you look at ctapipe.io.unofficial.targetio.contaners the R0 container contains the additional attribute first_cell_ids. If I had a script that used EventFileReaderFactory whose purpose was for debugging first_cell_ids but I passed it a different file format than targetio, then an error (KeyError or ValueError) would be raised when it tries to access first_cell_ids.

Exactly that was what I meant by **SST1MSpecificAnalysis**. I will access stuff that only SST1M DataContainers have.

ctapipe/io/unofficial/targetio/camera.py
dneise: I'd still recommend to use super() to its __init__

dneise: Right now, this feels a bit like surprise to me. You can call an empty c-tor of `Config` and get a perfectly healthy instance. 

Somebody changes Config somewhere and you get a modified Config without even knowing it.

I assume this is exactly what you need here, for some reason. But for me this sounds pretty dangerous and difficult to understand.

watsonjj:
@dneise I have incorporated the suggestions you have had so far.

docs/io/index_unofficial.rst
```diff
@@ -0,0 +1,283 @@
...
+  C code, and remember to add your extension to ext_modules in setup.py.
+* **Your reader exists as an external C/C++ library**, therefore software such
+  as SWIG can be used to provide a Python wrapper to the library,
+  automatically creating interfaces to the functions in the library.
+
+
+EventFileReaders (`ctapipe.io.unofficial.eventfilereader`)
+==========================================================
+
+This class provides the high-level interface for reading data from each data
```
dneise: What does "this class" refer to?

docs/io/index_unofficial.rst
```diff
@@ -0,0 +1,283 @@
...
+format. Read about the purpose of EventFileReaders at
+:ref:`EventFileReaderClasses`.
+
+Location for every unofficial `ctapipe.io.eventfilereader.EventFileReader`.
```
dneise: This sentence is missing a verb

docs/io/index_unofficial.rst
```diff
@@ -0,0 +1,283 @@
...
+found.
+
+The reader must also implement a fast method for
+`ctapipe.io.eventfilereader.EventFileReader.check_file_compatibility`. This
+method attempts to find a reader which *may* be compatible with the input
+file, and can be as simple as looking for a unique file extension in the file
+path or attempting to find a header keyword in the file. It is strongly
+advised that this method does not depend on the external reader software, as
+this would introduce a dependency when choosing the correct reader in the
+`ctapipe.io.eventfilereader.EventFileReaderFactory`.
```
dneise: I think we are mixing topics here, or features. As we have discussed already, using a reader within that factory makes sense only if that reader is able to provide containers which are expected by the user of the factory.
But at this point this important fact has not even been mentioned.

Therefore, I propose:

 * rename `EventFileReader.check_file_compatibility(path)` to `EventFileReader.is_compatible(path)`
 * make the abstract method `is_compatible` explicit and by default return `False`.

This way, all new readers, whose developers do not yet know about that factory, simply work. The factory will just ignore them. Until the developers read further about that factory.

docs/io/index_unofficial.rst
```diff
@@ -0,0 +1,283 @@
...
+
+The currently existing EventFileReaders,
+`ctapipe.io.eventfilereader.HessioFileReader` and
```
dneise: Why is the HessioFileReader not in unofficial? Is it official? If so, why is it not called CtaFileReader?

docs/io/index_unofficial.rst
```diff
@@ -0,0 +1,283 @@
...
+templates to create a new EventFileReader for a different data format.
+
+
+Low-Level Reader-Specific Modules
+===========================
+
+A seperate module should be created for the low-level method used for filling
+the event containers for a specific data format. The naming convention for
+this module is "\*io.py", where "\*" is the short-name for the data format.
+For example: `ctapipe.io.hessio` and `ctapipe.io.unofficial.targetio.targetio`.
```
dneise: Really? We put sub-modules into the `io` module and invent a rule, that sub-module names have to **end on `io`** as well?

docs/io/index_unofficial.rst
```diff
@@ -0,0 +1,283 @@
...
+  `ctapipe.io.eventfilereader.EventFileReader`.
+* The low-level methods fill the containers with the relevant data, e.g. the
+  waveforms are stored in "data.r0.tel[].adc_samples" (assuming they are
+  waveforms of data level R0, see the next section for an explanation of data
+  levels and see `ctapipe.io.hessio` and
+  `ctapipe.io.unofficial.targetio.targetio` for examples)
+* The unique 'origin' string is added to the meta dictionary of the container
+  so the data source of the events can be obtained by other algorithms, e.g.
+
+  >>> data.meta['origin'] = "targetio"
```
dneise: A rule is missing here, how the origin string should be created. Should it be the module name? Or the class-name of the reader who created this? 

And why should it be a string? What will this string be used for? If is it used for dynamically executing some code, based on the "origin", when instead of putting a string there, one might want to hook an function-, class- or module-reference there, so the dynamic executing becomes much easier and safer.

docs/io/index_unofficial.rst
```diff
@@ -0,0 +1,283 @@
...
+container:
+
+* If the file contains raw events, that still require low-level calibration
+  to be useful for offline analysis, then they should be read into the r0
+  container.
+* If the file contains events which have their low-level calibration fully
+  applied to them, then they should be read into the r1 container.
+* If the file contains events which have their low-level calibration fully
+  applied to them AND they are in a proposed format of DL0 (perhaps including
+  a data reduction approach), then they should be stored in the dl0 container.
```
dneise: "DL0" and "dl0" both used in the same sentence without making clear what the distinction is between capitalized and lower case version.

dneise: Also this is a circular description. According to this sentence, a DL0 container contains data which is in a proposed DL0 format. 

Btw. "**a** proposed DL0 format", this implies there are multiple proposals and all are valid in this context.

docs/io/index_unofficial.rst
```diff
@@ -0,0 +1,283 @@
...
+See this link for more details about the `CTA High-Level Data Model
+Definitions SYS-QA/160517
+<https://jama.cta-observatory.org/perspective.req?projectId=6&docId=26528>`_
+(CTA internal).
+
+
+Containers and R1 Calibration
+=============================
+
+This is an optional section. It is included here because the R1 calibration
```
dneise: According to what I understood about the factory, this is actually optionlly only if one does not want to use the factory, since the first interface, where different file formats *have to agree* is R1. At least this is what I understood. 

So this part, just like the `EventFileReader.is_compatible`method is mandatory in context of the factory. 

Therefore, I would find it good, to discuss both features together.

docs/io/index_unofficial.rst
```diff
@@ -0,0 +1,283 @@
...
+then this method should correctly return the right one. By default, a new
+reader will return 'NullR1Calibrator'.
+
+
+Instrument Configuration
+========================
+
+As the `ctapipe.instrument` module and "instrument database" are not fully
+developed, the camera configuration information (such as pixel positions) are
+still usually obtained from inside the hessio file.
```
dneise: "the hessio file" ... I do not understand that. What file do you refer to here?

docs/io/index_unofficial.rst
```diff
@@ -0,0 +1,283 @@
...
+protoptype teams have created their own data formats in which
+their camera events are stored.
+
+For data analysis, MC verification, and increasing familiarity with using
+`ctapipe`, it is desirable to allow reading of all these data formats into
+`ctapipe`. However it is important to keep in mind the following points in the
+creation of a new unofficial reader:
+
+* These readers are unofficial and temporary - every camera team will be
+  adopting the official CTA data format once it is defined.
```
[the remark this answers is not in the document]
watsonjj: I disagree that it creates such alarm. This documentation is meant to be informative, it is very relevant to include this information, to ensure the person reading it has a full explanation of the intentions of ctapipe.

I am still adamant that compatibility with these unofficial data formats will be dropped in the future (for reasons we have already discussed at length). I think it is important we don't hide this information.

It is required by CTA that the camera teams will adopt the final data format. As ctapipe is CTA software, I am just stating the truth in this bullet point.

docs/io/index_unofficial.rst
```diff
@@ -0,0 +1,283 @@
...
+reader, this documentation has been created, containing the guidelines to be
+followed. These guidelines can be summarised as:
+
+* Each reader should have a corresponding
+  `ctapipe.io.eventfilereader.EventFileReader` in
+  `ctapipe.io.unofficial.eventfilereader` to ensure the common high-level
+  interface for reading data is maintained.
+* Care should be taken in the imports of external modules. There should be no
+  dependencies added to ctapipe on the external libraries required for these
+  unofficial readers.
```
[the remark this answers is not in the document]
watsonjj: These two points are intended as a summary. The expectation is the person reading will continue reading to learn the details. I have added a link to the relevant chapters where they can find more information.

docs/io/index_unofficial.rst
```diff
@@ -0,0 +1,283 @@
...
+* **You have a C/C++ inplementation of your reader in a simple script**,
+  therefore you could provide an interface to that functionality by using
+  "ctypes". The C/C++ scripts can exist inside your reader module, and are
+  compiled as part of the ctapipe installation. See
+  `ctapipe.utils.neighbour_sum` as an example of how to interface Python with
+  C code, and remember to add your extension to ext_modules in setup.py.
+* **Your reader exists as an external C/C++ library**, therefore software such
+  as SWIG can be used to provide a Python wrapper to the library,
+  automatically creating interfaces to the functions in the library.
+
```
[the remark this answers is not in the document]
watsonjj: I think a lot of people coming to ctapipe are complete beginners with Python, and have no experience on how to wrap C/C++ code into Python, or if its even possible (when I started with ctapipe I didn't know about these possibilities).  

This section gives some hints to such a person of possible approaches to incorporate their existing code. I don't think it should be moved, as this is very relevant to this section. 

Perhaps a new section in the ctapipe documentation could be created to cover the full subject of wrapping C/C++ code (in addition to this small introduction, I could also reference that chapter). But I am no expert in all the methods available.

docs/io/index_unofficial.rst
watsonjj: If the creator of the Reader has correctly followed these guidelines and the existing Reader implementations, then their Reader **will** provide containers which are expected in the generic scripts which utilise EventFileReaderFactory to process waveforms without caring about the data format they come from. Such as examples/calibration_pipeline.py.

Allowing the creator to ignore EventFileReaderFactory and check_file_compatibility will cause confusion when trying to read their data format into this script, as they will be told that no EventFileReader is compatible.

I think its fine to expect the creator to setup the full implementation of the EventFileReader.

watsonjj: The only containers that could be data format specific are the R0 containers. But they should still contain the same fields as R0CameraContainer (adc_samples, adc_sums), so the common waveform processing scripts will always find the correct fields irregardless of data format, so their should be no problem with using the factory.

docs/io/index_unofficial.rst
watsonjj: As we have previously discussed, technically it should be moved to unofficial.

docs/io/index_unofficial.rst
watsonjj: Its just the convention we have been using so far.

docs/io/index_unofficial.rst
watsonjj: Yeah, I don't think this is used anywhere yet, except debugging and testing that the events have the expected origin.

The string should just be the name of the data format.

docs/io/index_unofficial.rst
watsonjj: The lowercase refers to the name of the sub-container in the event container. I have given the lowercase code highlighting now.

It's not circular, but I am outlying the obvious just for clarity. If an unofficial data format is a proposed DL0 format (i.e. a format being proposed to CTA for storing the Data-Level-0 information to disk) then the information should be read into the `dl0` container.

watsonjj: I will improve the explanation in the docs

docs/io/index_unofficial.rst
watsonjj: No, it is separate to the `EventFileReaderFactory`, and only needs to be considered by the creator if the data format they are reading will be stored in the `r0` container.

docs/io/index_unofficial.rst
watsonjj: hessio is the name of the data format used to store the output from sim_telarray

docs/io/index_unofficial.rst
```diff
@@ -0,0 +1,287 @@
...
+into ctapipe. Some of the possible forms and appropriate inplementations are:
+
+* **Your reader can be entirely programmed in Python**, in which case your
+  reader does not depend on any external packages, and you can include that
+  code in your reader's `Low-Level Reader-Specific Modules`_.
+* **Your reader exists as an external Python module**, in which case it could
+  be made available on conda for easy (optional) installation.
+* **You have a C/C++ inplementation of your reader in a simple script**,
+  therefore you could provide an interface to that functionality by using
+  "ctypes". The C/C++ scripts can exist inside your reader module, and are
```
MaxNoe: Or `cython` or `swig` or `cffi` writing c glue code.

docs/io/index_unofficial.rst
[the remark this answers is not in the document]
MaxNoe: I don't really see the differences in the 4 bullet points, I think it's just two?

Existing python code and existing C++ code.

docs/io/index_unofficial.rst
```diff
@@ -0,0 +1,287 @@
...
+  C code, and remember to add your extension to ext_modules in setup.py.
+* **Your reader exists as an external C/C++ library**, therefore software such
+  as SWIG can be used to provide a Python wrapper to the library,
+  automatically creating interfaces to the functions in the library.
+
+
+EventFileReaders
+================
+
+The module `ctapipe.io.unofficial.eventfilereader` contains the
```
MaxNoe: Thats not good!

There should not be a single file that contains `EventFileReaders` for different data formats.
I think this will be much easier to maintain if there are submodules for each data format instead of a single file mixing different file formats.

```python
from ctapipe.io.hessio import HessioEventFileReader
from ctapipe.io.zfits import ZfitsEventFileReader
```

docs/io/index_unofficial.rst
```diff
@@ -0,0 +1,287 @@
...
+
+Each reader is required to import their
+`Low-Level Reader-Specific Modules`_ inside their `__init__` method,
+therefore bypassing the need for the external reader software to be installed
+unless someone has chosen to use this specific
+`ctapipe.io.eventfilereader.EventFileReader`, in which case a
+`ModuleNotFoundError` will be raised if the external software has not been
+found.
+
+The reader must also implement a fast method for
```
MaxNoe: define `fast`

docs/io/index_unofficial.rst
```diff
@@ -0,0 +1,287 @@
...
+`ctapipe.io.eventfilereader.EventFileReader.is_compatible`. This
```
MaxNoe: Then this has to be an `@abstractclassmethod` in the EventFileReaderClass.

docs/io/index_unofficial.rst
```diff
@@ -0,0 +1,287 @@
...
+method attempts to find a reader which *may* be compatible with the input
```
MaxNoe: I think you made a mistake here. A method `EventFileReader.is_compatible` should only return a bool, wheiter a given path is compatible or not. It should not find a compatible reader. It  may be used somewhere else to find a compatible reader.
E.g. like this:
```python
for reader in readers:
    try:
        if reader.is_compatible(path):
            break
    except ModuleNotFoundError:
        pass
else:
    raise IOError("No compatible reader found for path")
```

docs/io/index_unofficial.rst
MaxNoe: Again, just get rid of the `unofficial` submodule and deprecate formats as better alternatives become available.

docs/io/index_unofficial.rst
MaxNoe: I don't think it was a convention to call the submodule `*io`, I think it was just an chance that all readers until now are for formats that end on `io` like hessio and targetio.

MaxNoe: And i also do not like to have io at least two times in there if it is not necessary.

docs/io/index_unofficial.rst
```diff
@@ -0,0 +1,287 @@
...
+various components involved in the event building (such as containers) to be
+grouped together for a particular data format.
+
+Each data format can be wildly different, therefore these low-level
+modules can appear very different in how they function. This is fine as long
+as these rules are followed:
+
+* The `ctapipe.io.eventfilereader.EventFileReader` calls the low-level methods
+  such that the high-level interface remains the same as other
+  `ctapipe.io.eventfilereader.EventFileReader`.
```
MaxNoe: As it is a Base class of `EventFileReader` this does not need to be a rule as it is enforced syntactically.

docs/io/index_unofficial.rst
MaxNoe: I think @dneise is confused by the definit article. Is there any one single hessio file you refer to here? Then it should be linked. 

If not, then it should be reworded to something like "Instrument information is read from the inputfile in case of hessio"

docs/io/index_unofficial.rst
```diff
@@ -0,0 +1,287 @@
...
+added to the `data_files` dictionary in setup.py.
+
+For an example of one approach to achieve this, see
+`ctapipe.io.unofficial.targetio.camera`.
+
+
+Tests and Documentation
+=======================
+
+Every unofficial reader must be thoroughly tested and documented in their
```
MaxNoe: No need to repeat it here, this is true for the full ctapipe project, so automatically also for every submodule.

docs/io/index_unofficial.rst
```diff
@@ -0,0 +1,287 @@
...
+The method to install the external software should be added to the .travis.yml
+file, so that the Travis CI can perform all tests, including the unofficial
+reader tests. Additionally, a small test file that is in an unofficial data
+format should be committed to ctapipe-extra.
+
+Installation instructions for external software should be included in the
+docstrings of the EventFileReader.
+
+
+TargetioFileReader
```
MaxNoe: I don't this this should reside here. This looks like it belongs into the class docstring of TargetioFileReader.

mackaiver:
I'm wondering whether it makes more sense to simply have event readers for the different cameras implemented and managed in their own little repositories within the cta-observatory organization.
This project is big enough as it is. 

These readers would simply implement against the API as provided by ctapipe ( EventSource or whatever we should call that).  This also reduces the need for writing detailed guideline-lines. 

I also do not see the need for something like an EventSourceFactory or similar things. 
I understand these are convenience functions for autmagically using the right module for reading the files right? Once the file formats have converged this wont be necessary. And for now I think you can just choose the proper reader beforehand. Honestly I think this produces way to much overhead.

MaxNoe:
@mackaiver I fully agree. People now looking into data know which reader to use, later there will only be one reader. So no need for a factory.

dneise:
> I'm wondering whether it makes more sense to simply have event readers for the different cameras implemented and managed in their own little repositories within the cta-observatory organization.
This project is big enough as it is.

Well this PR and also #598 (which is currently WIP by me) have their start in the following question: How would camera teams contribute to ctapipe? Or how can ctapipe be useful for camera teams? 
The answer somehow was: Well we start by being able to read the data files of the different cameras. 
Hence people started to try and get their readers into ctapipe.

There is an issue about this, but I can't find it, it should be mentioned in the PR text at the top, but it is not. 

Your proposal @mackaiver if I understand you, goes in the opposite direction and not have the camera teams file readers inside ctapipe. This re-opens then the question: How should camera teams make use of ctapipe?

watsonjj:
@MaxNoe @mackaiver I strongly disagree with your views on `EventFileReaderFactory`. The whole reason this discussion about adding additional readers has become important is because:

1. MC validation teams want to create scripts which apply the same processes to the waveforms between simulated and real-data.
1. People (such as myself) are starting to write scripts to process on-sky data. That data does not yet exist, but will soon. Meanwhile the pipeline to process them can be developed using simulation data. 

Without the ability to allow these scripts to read in any compatible data format, and process the data without caring about the data format, then these scripts would need to be duplicated for every data source you wish to use. And it removes the easy sharing of such pipeline scripts between camera teams, reducing collaboration and increasing duplication.

Besides, there is nothing stopping people doing exactly what @mackaiver said, its exactly what I've been doing up to this point. However it is time we have guidelines which are clear and accessible for those people who want to know how to achieve this for their own data format. 

I have just seen @dneise comment, and thats exactly my point. The issue where we discussed this is #590. I believe we should continue with this approach and work on these guidelines.

watsonjj:
@MaxNoe Thank you for your comments!

I agree we should remove the low-level submodules, and move the filling of the containers to inside the EventFileReader.

I have changed my mind and now agree each EventFileReader should have their own submodule, and we should remove the `unofficial` submodule.

These are some big changes, not only to this PR but also to `ctapipe.io`. I will begin by creating a new PR which makes changes to the master branch's `ctapipe.io`, to move the low-level reading from its own module to its EventFileReader, and to prepare EventFileReaders for the addition of future new Readers. This is a change @kosack mentioned in #590 which I think should be implemented now before we consider new Readers.

I hope everyone agrees with this approach.

docs/io/index_unofficial.rst
```diff
@@ -0,0 +1,287 @@
...
+================================
+
+Presumably you already have some software that exists to read in your data
+format. What form that reader exists in dictates how easy it is to incorporate
+into ctapipe. Some of the possible forms and appropriate inplementations are:
+
+* **Your reader can be entirely programmed in Python**, in which case your
+  reader does not depend on any external packages, and you can include that
+  code in your reader's `Low-Level Reader-Specific Modules`_.
+* **Your reader exists as an external Python module**, in which case it could
```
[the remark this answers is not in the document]
watsonjj: As @dneise mentioned, perhaps someone should create a separate page in the documentation of how to wrap c code into python, and we reference that page here. I am definitely not an expert on these approaches, and only wrote the ones I have used.

mackaiver:
> Well this PR and also #598 (which is currently WIP by me) have their start in the following question: How would camera teams contribute to ctapipe? Or how can ctapipe be useful for camera teams?
> The answer somehow was: Well we start by being able to read the data files of the different cameras.
> Hence people started to try and get their readers into ctapipe.
> 
> There is an issue about this, but I can't find it, it should be mentioned in the PR text at the top, but it is not.
> 
> Your proposal @mackaiver if I understand you, goes in the opposite direction and not have the camera teams file readers inside ctapipe. This re-opens then the question: How should camera teams make use of ctapipe?



> Well this PR and also #598 (which is currently WIP by me) have their start in the following question: How would camera teams contribute to ctapipe? Or how can ctapipe be useful for camera teams?
> The answer somehow was: Well we start by being able to read the data files of the different cameras.
> Hence people started to try and get their readers into ctapipe.

Alright. Makes sense. 

> Without the ability to allow these scripts to read in any compatible data format, and process the data without caring about the data format, then these scripts would need to be duplicated for every data source you wish to use.

I'm sorry I did not word that correctly.  I'm not proposing something which is orthogonal to that idea. In fact I agree we should support reading all the file formats. Whether these modules are inside ctapipe or outside should not make difference to my understanding. Zfits and others are external dependencies which should be optional right? I agree with you that it might be easier to have it all in one place however. 

Maybe you can explain the idea of the factory again? From what I understand all that's needed is an abstract  base class (like the one you already implemented). Subclasses can be selected and imported dynamically. Or were you going for something like a [strategy pattern](https://en.wikipedia.org/wiki/Strategy_pattern)?

kosack:
By the way, this sort of behavior can be achieved using a plugin discovery system:
https://packaging.python.org/guides/creating-and-discovering-plugins/

For example, it seems that the *Namespace Package* is somewhat similar to what Jason is suggesting.  If we define the `ctapipe.io.plugins` (which may be better than `unofficial`) as a namespace package, we can either have extra IO readers in external packages or internal, and they can be automatically discovered. 

The most important thing is to come up with a nice clean API.  The `x_event_source` api is simple, but maybe not flexible enough, and the `EventFileReader` hierarchy probably needs some simplification, so perhaps we can converge on merging the two concepts into something simpler that just defines a common way to fill in data from a file. 

What are all the use cases? I can thing of a few:

* loop over events in an input file/stream (the main use case for reconstruction)
* go to a particular event and load it (maybe useful for debugging/test bench) - not all formats allow random event access, so this may be implemented also as a *seek* (loop until event ID found, and then read rest of data for that event)
* randomly access events? (probably not needed for most cases, but nice for e.g. a debugging GUI to step forward and backward through a file)

I think things like defining the output directory, etc, should be separated into another class, since they have nothing to do with reading the data.

kosack:
Also, if we stick with the "factory/registry" route, that can be cleanly achieved using metaclasses:

```py3
class ReaderMeta(type):
    _REGISTRY = {}
    def __new__(cls, name, bases, attrs):
        new_cls = type.__new__(cls, name, bases, attrs)
        cls._REGISTRY[name] = new_cls
        return new_cls

class A(metaclass=InputMeta):
    pass
class B(metaclass=InputMeta):
    pass
```

now loop over all registered classes, instantiate them, etc:

```py3
>>> for name, theclass in InputMeta._REGISTRY.items():
>>>    instance = theclass()
>>>    print("Constructed a '{}': {}".format(name, instance))

Constructed a 'A': <__main__.A object at 0x11071d9e8>
Constructed a 'B': <__main__.B object at 0x11071d358>
```

docs/io/index_unofficial.rst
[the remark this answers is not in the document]
dneise: > I think a lot of people coming to ctapipe are complete beginners with Python...
Well in that case, I agree that such a nice introduction should be here. But maybe not deep down in the part about readers, but somewhere more prominent? It is very general after all.

docs/io/index_unofficial.rst
dneise: There are camera types, which do not sample waveforms, right?

watsonjj: Sure, but they are treated as waveforms with 1 sample, and are handled specially when needed (e.g. in the dl1 calibration no processing is done to the values).

dneise:
@kosack wrote:
> The most important thing is to come up with a nice clean API.

and I totally agree, At the moment I still not fully understand the interface proposed here. 

Let me make a proposal for a simple interface:
```python
import numpy as np

from ctapipe.core import Map, Field
from ctapipe.io.containers import(
    R0CameraContainer,
    R0Container,
    DataContainer,
)


class CustomReader:
    def __init__(self, path):
        import external_deps
        self.file = external_deps.open(path)

    def __iter__(self):
        return self

    def __next__(self):
        for low_level_event in self.file:
            event = CustomDataContainer()
            # somehow fill event from `low_level_event`
            yield event


class CustomR0CameraContainer(R0CameraContainer):
    custom_field = Field(np.ndarray, 'something custom')


class CustomR0Container(R0Container):
    tel = Field(
        Map(CustomR0CameraContainer),
        "map of tel_id to CustomR0CameraContainer"
    )


class CustomDataContainer(DataContainer):
    r0 = Field(CustomR0Container(), "Raw Data")
```

----

A custom reader, should accept a `path` for to be constructed. And when iterated over it should spit out `DataContainer`s. That is the smallest useful interface, I could think of. It is easy to explain.

This interface is unfortunately impossible, due to the `traitlets` decision. So the next step would be:
```python
from traitlets import Unicode

class CustomReader(Component):

    path = Unicode(
        '',
        help='Path to input file'
    ).tag(config=True)

    def __init__(self, config, tool, **kwargs):
        super().__init__(config=config, parent=tool, **kwargs)
        import external_deps
        self.file = external_deps.open(self.path)

# the rest is the same
```

This interface is so easy (apart from the `traitlets` part) that it can be explained in one sentence. It is a useful reader, since It spits out stuff, the caller expects, i.e. `DataContainer`s.

Ah, I forgot, in order to ensure this interface, we let custom readers implement the interface specified by an abstract class, so people missing something in the docu, get an exception.

So this is the interface, easy to explain, and easy to read.
```python
from abc import abstractmethod
from traitlets import Unicode

class EventFileReader(Component):
    def __init__(self, config, tool, **kwargs):
        super().__init__(config=config, parent=tool, **kwargs)

    path = Unicode(
        '',
        help='Path to input file'
    ).tag(config=True)

    def __iter__(self):
        return self

    @abstractmethod
    def __next__(self):
        pass
```

And this would be an example implementation:
```python
from proposal_above import EventFileReader
import numpy as np

from ctapipe.core import Map, Field
from ctapipe.io.containers import(
    R0CameraContainer,
    R0Container,
    DataContainer,
)

class CustomReader(EventFileReader):
    def __init__(self, config, tool, **kwargs):
        super().__init__(config=config, parent=tool, **kwargs)
        import external_deps
        self.file = external_deps.open(self.path)

    def __next__(self):
        for low_level_event in self.file:
            event = CustomDataContainer()
            # somehow fill event from `low_level_event`
            yield event

class CustomR0CameraContainer(R0CameraContainer):
    custom_field = Field(np.ndarray, 'something custom')

class CustomR0Container(R0Container):
    tel = Field(
        Map(CustomR0CameraContainer),
        "map of tel_id to CustomR0CameraContainer"
    )

class CustomDataContainer(DataContainer):
    r0 = Field(CustomR0Container(), "Raw Data")
```

dneise:
@kosack & @watsonjj : I think this can be closed, right?

watsonjj:
Yes I plan to close it soon. The information in here will be deprecated by other PRs

kosack:
note: closed since implemented in #634

PR #613, opened 2018-01-09
Updates to EventFileReader

watsonjj:
This PR concerns updates to EventFileReader and HessioFileReader discussed in #590 and #610.

* EventFileReader has been cleaned-up (removing stuff like output_directory)
* EventFileReader has been made into a [sequence](https://docs.python.org/3.6/glossary.html#term-sequence) so that is is iterable and supports element access.
* The methods for retrieving events are defined in the base EventFileReader class. Other EventFileReaders can override `__getitem__` if they have a more efficient method instead of looping from the first event.
* The functionality of hessio_event_source has been moved into HessioFileReader._generator

### DIFF
--- a/ctapipe/io/eventfilereader.py
+++ b/ctapipe/io/eventfilereader.py
@@ -32,7 +32,7 @@ class EventFileReader(Component):
     to read.
 
     To create an instance of an EventFileReader you must pass the traitlet
-    configuration (containing the input_path) and the
+    configuration (containing the input_url) and the
     `ctapipe.core.tool.Tool`. Therefore from inside a Tool you would do:
 
     >>> reader = EventFileReader(self.config, self)
@@ -42,13 +42,13 @@ class EventFileReader(Component):
     ctapipe/examples/calibration_pipeline.py.
 
     However if you are not inside a Tool, you can still create an instance and
-    supply an input_path via:
+    supply an input_url via:
 
-    >>> reader = EventFileReader(None, None, input_path="/path/to/file")
+    >>> reader = EventFileReader(None, None, input_url="/path/to/file")
 
     To loop through the events in a file:
 
-    >>> reader = EventFileReader(None, None, input_path="/path/to/file")
+    >>> reader = EventFileReader(None, None, input_url="/path/to/file")
     >>> for event in reader:
     >>>    print(event.count)
 
@@ -58,13 +58,13 @@ class EventFileReader(Component):
     Alternatively one can use EventFileReader in a `with` statement to ensure
     the correct cleanups are performed when you are finished with the reader:
 
-    >>> with EventFileReader(None, None, input_path="/path/to/file") as reader:
+    >>> with EventFileReader(None, None, input_url="/path/to/file") as reader:
     >>>    for event in reader:
     >>>       print(event.count)
 
     Attributes
     ----------
-    input_path : str
+    input_url : str
         Path to the input event file.
     max_events : int
         Maximum number of events to loop through in generator
@@ -76,7 +76,7 @@ class EventFileReader(Component):
         * Observation ID
     """
 
-    input_path = Unicode(
+    input_url = Unicode(
         '',
         help='Path to the input file containing events.'
     ).tag(config=True)
@@ -108,15 +108,15 @@ class EventFileReader(Component):
 
         self.metadata = dict(is_simulation=False)
 
-        if not exists(self.input_path):
+        if not exists(self.input_url):
             raise FileNotFoundError("file path does not exist: '{}'"
-                                    .format(self.input_path))
-        self.log.info("INPUT PATH = {}".format(self.input_path))
+                                    .format(self.input_url))
+        self.log.info("INPUT PATH = {}".format(self.input_url))
 
         if self.max_events:
             self.log.info("Max events being read = {}".format(self.max_events))
 
-        Provenance().add_input_file(self.input_path, role='dl0.sub.evt')
+        Provenance().add_input_file(self.input_url, role='dl0.sub.evt')
 
     @staticmethod
     @abstractmethod
@@ -241,7 +241,7 @@ class EventFileReaderFactory(Factory):
 
     # Product classes traits
     # Would be nice to have these automatically set...!
-    input_path = Unicode(
+    input_url = Unicode(
         get_dataset('gamma_test.simtel.gz'),
         help='Path to the input file containing events.'
     ).tag(config=True)
@@ -258,14 +258,14 @@ class EventFileReaderFactory(Factory):
         if self.reader is not None:
             return self.reader
         else:
-            if self.input_path is None:
-                raise ValueError("Please specify an input_path for event file")
+            if self.input_url is None:
+                raise ValueError("Please specify an input_url for event file")
             try:
                 for subclass in self.subclasses:
-                    if subclass.is_compatible(self.input_path):
+                    if subclass.is_compatible(self.input_url):
                         return subclass.__name__
                 raise ValueError
             except ValueError:
                 self.log.exception("Cannot find compatible EventFileReader "
-                                   "for: {}".format(self.input_path))
+                                   "for: {}".format(self.input_url))
                 raise
--- a/ctapipe/io/eventfilereader.py
+++ b/ctapipe/io/eventfilereader.py
@@ -73,6 +73,7 @@ class EventFileReader(Component):
 
         self.source = self._generator()
         self.current_event = None
+        self.getevent_warn = True
 
     @staticmethod
     @abstractmethod
@@ -217,6 +218,9 @@ class EventFileReader(Component):
             The event container filled with the requested event's information
 
         """
+        if self.getevent_warn:
+            self.log.warning("Seeking to event... (long process)")
+            self.getevent_warn = False
         if not use_event_id:
             msg = "Event index {} not found in file".format(ev)
             for event in self.source:
@@ -235,7 +239,7 @@ class EventFileReader(Component):
         # Only need to calculate once
         if not self._num_events:
             self.reset()
-            self.log.info("Obtaining number of events in file...")
+            self.log.warning("Obtaining length of file... (long process)")
             count = 0
             for _ in self:
                 if self.max_events and count >= self.max_events:
--- a/ctapipe/io/eventfilereader.py
+++ b/ctapipe/io/eventfilereader.py
@@ -2,36 +2,32 @@
 Handles reading of different event/waveform containing files
 """
 from abc import abstractmethod
-from os.path import basename, splitext, dirname, join, exists
-import numpy as np
-from traitlets import Unicode, Int, CaselessStrEnum, observe
+from os.path import exists
+from traitlets import Unicode, Int, CaselessStrEnum
 from copy import deepcopy
 from ctapipe.core import Component, Factory
 from ctapipe.utils import get_dataset
-from ctapipe.io.hessio import hessio_event_source, hessio_get_list_event_ids
 
 
 class EventFileReader(Component):
     """
-    Parent class for EventFileReaders of different sources. A new
-    EventFileReader should be created for each type of event file read into
-    ctapipe, e.g. simtelarray files are read by the `HessioFileReader`.
+    Parent class for EventFileReaders of different sources.
+
+    A new EventFileReader should be created for each type of event file read
+    into ctapipe, e.g. sim_telarray files are read by the `HessioFileReader`.
+
+    EventFileReader provides a common high-level interface for accessing data
+    from different data sources. Creating an EventFileReader for a new data
+    source ensures that data can be accessed in a common way, irregardless of
+    the data source.
 
     Attributes
     ----------
     input_path : str
-    directory : str
-        Automatically set from `input_path`.
-    filename : str
-        Name of the file without the extension.
-        Automatically set from `input_path`.
-    extension : str
-        Automatically set from `input_path`.
-    output_directory : str
-        Directory to save outputs for this file
-
+        Path to the input event file.
+    max_events : int
+        Maximum number of events to loop through in generator
     """
-    origin = None
 
     input_path = Unicode(get_dataset('gamma_test.simtel.gz'), allow_none=True,
                          help='Path to the input file containing '
@@ -60,76 +56,21 @@ class EventFileReader(Component):
         """
         super().__init__(config=config, parent=tool, **kwargs)
 
-        if self.origin is None:
-            raise ValueError("Subclass of EventFileReader should specify "
-                             "an origin")
-
         self._num_events = None
-        self._event_id_list = []
 
         if self.input_path is None:
             raise ValueError("Please specify an input_path for event file")
-        self._init_path(self.input_path)
-
-    def _init_path(self, input_path):
-        if not exists(input_path):
+        if not exists(self.input_path):
             raise FileNotFoundError("file path does not exist: '{}'"
-                                    .format(input_path))
+                                    .format(self.input_path))
+        self.log.info("INPUT PATH = {}".format(self.input_path))
 
-        self.input_path = input_path
-        self.directory = dirname(input_path)
-        self.filename = splitext(basename(input_path))[0]
-        self.extension = splitext(input_path)[1]
-        self.output_directory = join(self.directory, self.filename)
-
-        if self.log:
-            self.log.info("INPUT PATH = {}".format(self.input_path))
-
-    @observe('input_path')
-    def on_input_path_changed(self, change):
-        new = change['new']
-        try:
-            self.log.warning("Change: input_path={}".format(new))
-            self._num_events = None
-            self._event_id_list = []
-            self._init_path(new)
-        except AttributeError:
-            pass
-
-    @observe('origin')
-    def on_origin_changed(self, change):
-        new = change['new']
-        try:
-            self.log.warning("Change: origin={}".format(new))
-        except AttributeError:
-            pass
-
-    @observe('max_events')
-    def on_max_events_changed(self, change):
-        new = change['new']
-        try:
-            self.log.warning("Change: max_events={}".format(new))
-            self._num_events = None
-            self._event_id_list = []
-        except AttributeError:
-            pass
-
-    @property
-    @abstractmethod
-    def origin(self):
-        """
-        Abstract property to be defined in child class.
-
-        Get the name for the origin of the file. E.g. 'hessio'.
-
-        Returns
-        -------
-        origin : str
-        """
+        if self.max_events:
+            self.log.info("Max events being read = {}".format(self.max_events))
 
     @staticmethod
     @abstractmethod
-    def check_file_compatibility(file_path):
+    def is_compatible(file_path):
         """
         Abstract method to be defined in child class.
 
@@ -149,214 +90,107 @@ class EventFileReader(Component):
 
     @property
     @abstractmethod
-    def num_events(self):
+    def camera(self):
         """
-        Abstract property to be defined in child class.
+        Name of the camera contained in the file. Read from the file or
+        inferred from the EventFileReader class if the data format is only
+        used by one camera.
 
-        Obtain the number of events from the file, store it inside
-        self._num_events, and return the value.
+        This property is used to choose the correct algorithms to process the
+        waveforms, as some cameras require different algorithms to others
+        (such as with R1 calibration and charge extraction).
 
         Returns
         -------
-        self._num_events : int
+        camera_name : str
         """
 
-    @property
-    @abstractmethod
-    def event_id_list(self):
-        """
-        Abstract property to be defined in child class.
-
-        Obtain the number of events from the file, store it as
-        self._event_id_list, and return the value.
-
-        Returns
-        -------
-        self._event_id_list : list[self._num_events]
-        """
+    def __iter__(self):
+        return self
 
     @abstractmethod
-    def read(self, allowed_tels=None, requested_event=None,
-             use_event_id=False):
+    def __next__(self):
         """
         Abstract method to be defined in child class.
 
-        Read the file using the processes required by the readers file-type.
-
-        Parameters
-        ----------
-        allowed_tels : list[int]
-            select only a subset of telescope, if None, all are read. This can
-            be used for example emulate the final CTA data format, where there
-            would be 1 telescope per file (whereas in current monte-carlo,
-            they are all interleaved into one file)
-        requested_event : int
-            Seek to a paricular event index
-        use_event_id : bool
-            If True ,'requested_event' now seeks for a particular event id
-            instead of index
+        Method where the filling of the `ctapipe.io.containers` occurs.
 
         Returns
         -------
-        source : generator
-            A generator that can be iterated over to obtain events
+        data : ctapipe.io.container
+            The event container filled with the event information
         """
 
-    def get_event(self, requested_event, use_event_id=False):
+    def __getitem__(self, item):
         """
-        Loop through events until the requested event is found
+        Obtain a particular event
 
         Parameters
         ----------
-        requested_event : int
-            Seek to a paricular event index
-        use_event_id : bool
-            If True ,'requested_event' now seeks for a particular event id
-            instead of index
+        item : int or str
+            If `item` is an int, then this is the event_index for the event
+            obtained. If `item` is a str, then this is the event_id for the
+            event obtained.
 
         Returns
         -------
-        event : `ctapipe` event-container
+        data : ctapipe.io.container
+            The event container filled with the requested event's information
 
         """
-        source = self.read(requested_event=requested_event,
-                           use_event_id=use_event_id)
-        event = next(source)
-        return deepcopy(event)
+        use_event_id = False
+        msg = "Event index {} not found in reader".format(item)
+        if isinstance(item, str):
+            item = int(item)
+            use_event_id = True
+            msg = "Event id {} not found in reader".format(item)
 
-    def find_max_true_npe(self, telescopes=None):
-        """
-        Loop through events to find the maximum true npe
-
-        Parameters
-        ----------
-        telescopes : list
-            List of telecopes to include. If None, then all telescopes
-            are included.
-
-        Returns
-        -------
-        max_pe : int
-
-        """
-        # TODO: Find an alternate method so this can be removed
-        self.log.info("Finding maximum true npe inside file...")
-        source = self.read()
-        max_pe = 0
-        for event in source:
-            tels = list(event.dl0.tels_with_data)
-            if telescopes is not None:
-                tels = []
-                for tel in telescopes:
-                    if tel in event.dl0.tels_with_data:
-                        tels.append(tel)
-            if event.count == 0:
-                # Check events have true charge included
-                try:
-                    if np.all(event.mc.tel[tels[0]].photo_electron_image == 0):
-                        raise KeyError
-                except KeyError:
-                    self.log.exception('[chargeres] Source does not contain '
-                                       'true charge')
-                    raise
-            for telid in tels:
-                pe = event.mc.tel[telid].photo_electron_image
-                this_max = np.max(pe)
-                if this_max > max_pe:
-                    max_pe = this_max
-        self.log.info("Maximum true npe inside file = {}".format(max_pe))
-
-        return max_pe
-
-
-class HessioFileReader(EventFileReader):
-    origin = 'hessio'
-
-    @staticmethod
-    def check_file_compatibility(file_path):
-        compatible = True
-        # TODO: Change check to be a try of hessio_event_source?
-        if not file_path.endswith('.gz'):
-            compatible = False
-        return compatible
-
-    @property
-    def num_events(self):
-        self.log.info("Obtaining number of events in file...")
-        if self._num_events:
-            pass
+        if not use_event_id and self.max_events and item >= self.max_events:
+            msg = "Event index {} outside of specified max_events {}"\
+                .format(item, self.max_events)
+        elif not use_event_id:
+            for event in self:
+                if event.index == item:
+                    return deepcopy(event)
         else:
-            self._num_events = len(self.event_id_list)
-        self.log.info("Number of events inside file = {}"
-                      .format(self._num_events))
+            for event in self:
+                if event.id == item:
+                    return deepcopy(event)
+
+        raise KeyError(msg)
+
+    def __len__(self):
+        if not self._num_events:
+            self.log.info("Obtaining number of events in file...")
+            count = 0
+            for _ in self:
+                if count >= self.max_events:
+                    break
+                count += 1
+            self._num_events = count
         return self._num_events
 
-    @property
-    def event_id_list(self):
-        self.log.info("Retrieving list of event ids...")
-        if self._event_id_list:
-            pass
-        else:
-            self.log.info("Building new list of event ids...")
-            ids = hessio_get_list_event_ids(self.input_path,
-                                            max_events=self.max_events)
-            self._event_id_list = ids
-        self.log.info("List of event ids retrieved.")
-        return self._event_id_list
 
-    def read(self, allowed_tels=None, requested_event=None,
-             use_event_id=False):
-        """
-        Read the file using the appropriate method depending on the file origin
-
-        Parameters
-        ----------
-        allowed_tels : list[int]
-            select only a subset of telescope, if None, all are read. This can
-            be used for example emulate the final CTA data format, where there
-            would be 1 telescope per file (whereas in current monte-carlo,
-            they are all interleaved into one file)
-        requested_event : int
-            Seek to a paricular event index
-        use_event_id : bool
-            If True ,'requested_event' now seeks for a particular event id
-            instead of index
-
-        Returns
-        -------
-        source : generator
-            A generator that can be iterated over to obtain events
-        """
-
-        # Obtain relevent source
-        self.log.debug("Reading file...")
-        if self.max_events:
-            self.log.info("Max events being read = {}".format(self.max_events))
-        source = hessio_event_source(self.input_path,
-                                     max_events=self.max_events,
-                                     allowed_tels=allowed_tels,
-                                     requested_event=requested_event,
-                                     use_event_id=use_event_id)
-        self.log.debug("File reading complete")
-        return source
-
-
-# External Children
-try:
-    from targetpipe.io.eventfilereader import TargetioFileReader, \
-        ToyioFileReader
-except ImportError:
-    pass
+# EventFileReader imports so that EventFileReaderFactory can see them
+from ctapipe.io.hessiofilereader import HessioFileReader
 
 
 class EventFileReaderFactory(Factory):
     """
     The `EventFileReader` `ctapipe.core.factory.Factory`. This
     `ctapipe.core.factory.Factory` allows the correct
-    `EventFileReader` to be obtained for the event file being read. This
-    factory tests each EventFileReader by calling
+    `EventFileReader` to be obtained for the event file being read.
+
+    This factory tests each EventFileReader by calling
     `EventFileReader.check_file_compatibility` to see which `EventFileReader`
     is compatible with the file.
+
+    Using `EventFileReaderFactory` in a script allows it to be compatible with
+    any data source that has an `EventFileReader` defined.
+
+    To use within a `ctapipe.core.tool.Tool`:
+
+    >>> reader = EventFileReaderFactory.produce(config=self.config, tool=self)
 
     Parameters
     ----------
@@ -373,7 +207,7 @@ class EventFileReaderFactory(Factory):
     Attributes
     ----------
     reader : traitlets.CaselessStrEnum
-        A string with the `EventFileReader.name` of the reader you want to
+        A string with the `EventFileReader.__name__` of the reader you want to
         use. If left blank, `EventFileReader.check_file_compatibility` will be
         used to find a compatible reader.
     """
@@ -407,7 +241,7 @@ class EventFileReaderFactory(Factory):
                 raise ValueError("Please specify an input_path for event file")
             try:
                 for subclass in self.subclasses:
-                    if subclass.check_file_compatibility(self.input_path):
+                    if subclass.is_compatible(self.input_path):
                         return subclass.__name__
                 raise ValueError
             except ValueError:
--- a/ctapipe/io/eventfilereader.py
+++ b/ctapipe/io/eventfilereader.py
@@ -4,7 +4,6 @@ Handles reading of different event/waveform containing files
 from abc import abstractmethod
 from os.path import exists
 from traitlets import Unicode, Int, CaselessStrEnum
-from copy import deepcopy
 from ctapipe.core import Component, Factory
 from ctapipe.utils import get_dataset
 from ctapipe.core import Provenance
@@ -45,52 +44,21 @@ class EventFileReader(Component):
     However if you are not inside a Tool, you can still create an instance and
     supply an input_path via:
 
-    >>> reader = EventFileReader(None, None, input_path=path)
+    >>> reader = EventFileReader(None, None, input_path="/path/to/file")
 
     To loop through the events in a file:
 
-    >>> reader = EventFileReader(None, None, input_path=path)
+    >>> reader = EventFileReader(None, None, input_path="/path/to/file")
     >>> for event in reader:
     >>>    print(event.count)
 
     **NOTE**: Every time a new loop is started through the reader, it restarts
     from the first event.
 
-    To obtain a particular event in a file:
-
-    >>> reader = EventFileReader(None, None, input_path=path)
-    >>> event = reader[event_index]
-    >>> print(event.count)
-
-    To obtain a particular event in a file from its event_id:
-
-    >>> reader = EventFileReader(None, None, input_path=path)
-    >>> event = reader["event_id"]
-    >>> print(event.count)
-
-    **NOTE**: Event_index refers to the number associated to the event
-    assigned by ctapipe (`event.count`), based on the order the events are
-    read from the file.
-    Whereas the event_id refers to the ID attatched to the event from the
-    external source of the file (software or camera or CTA array).
-
-    To obtain a slice of events in a file:
-
-    >>> reader = EventFileReader(None, None, input_path=path)
-    >>> event_list = reader[3:6]
-    >>> print([event.count for event in event_list])
-
-    To obtain a list of events in a file:
-
-    >>> reader = EventFileReader(None, None, input_path=path)
-    >>> event_indicis = [2, 6, 8]
-    >>> event_list = reader[event_indicis]
-    >>> print([event.count for event in event_list])
-
     Alternatively one can use EventFileReader in a `with` statement to ensure
     the correct cleanups are performed when you are finished with the reader:
 
-    >>> with EventFileReader(None, None, input_path=path) as reader:
+    >>> with EventFileReader(None, None, input_path="/path/to/file") as reader:
     >>>    for event in reader:
     >>>       print(event.count)
 
@@ -129,7 +97,6 @@ class EventFileReader(Component):
         """
         super().__init__(config=config, parent=tool, **kwargs)
 
-        self._num_events = None
         self._metadata = dict(is_simulation=False)
 
         if not exists(self.input_path):
@@ -141,11 +108,6 @@ class EventFileReader(Component):
             self.log.info("Max events being read = {}".format(self.max_events))
 
         Provenance().add_input_file(self.input_path, role='dl0.sub.evt')
-
-        self._source = self._generator()
-        self._current_event = None
-        self._has_fast_seek = False  # By default seeking iterates through
-        self._getevent_warn = True
 
     @staticmethod
     @abstractmethod
@@ -209,20 +171,19 @@ class EventFileReader(Component):
         generator
         """
 
-    def reset(self):
-        """
-        Recreate the generator so it starts from the beginning
-        """
-        self._source = self._generator()
-        self._current_event = None
-
     def __iter__(self):
-        # Always reset generator when starting a new iteration
-        self.reset()
-        for event in self._source:
+        """
+        Abstract method to be defined in child class.
+
+        Generator where the filling of the `ctapipe.io.containers` occurs.
+
+        Returns
+        -------
+        generator
+        """
+        for event in self._generator():
             if self.max_events and event.count >= self.max_events:
                 break
-            self._current_event = event
             yield event
 
     def __enter__(self):
@@ -230,144 +191,6 @@ class EventFileReader(Component):
 
     def __exit__(self, exc_type, exc_val, exc_tb):
         pass
-
-    def __getitem__(self, item):
-        """
-        Obtain a particular event
-
-        Parameters
-        ----------
-        item : int or str
-            If `item` is an int, then this is the event_index for the event
-            obtained. If `item` is a str, then this is the event_id for the
-            event obtained.
-
-        Returns
-        -------
-        event : ctapipe.io.container
-            The event container filled with the requested event's information
-
-        """
-        if self.is_stream:
-            raise IOError("Input is a stream, __getitem__ is disabled")
-
-        current = None
-
-        if not self._has_fast_seek and self._getevent_warn:
-            self.log.warning("Seeking to event... (potentially long process)")
-            self._getevent_warn = False
-
-        # Handling of different input types (int, string, slice, list)
-        use_event_id = False
-        if isinstance(item, int):
-            if self._current_event:
-                current = self._current_event.count
-            if item < 0:
-                item = len(self) + item
-                if item < 0 or item >= len(self):
-                    msg = ("Event index {} out of range [0, {}]"
-                        .format(item, len(self)))
-                    raise IndexError(msg)
-        elif isinstance(item, str):
-            item = int(item)
-            use_event_id = True
-            if self._current_event:
-                current = self._current_event.r0.event_id
-        elif isinstance(item, slice):
-            it = range(item.start or 0, item.stop or len(self), item.step or 1)
-            events = [self[i] for i in it]
-            return events
-        elif isinstance(item, list):
-            events = [self[i] for i in item]
-            return events
-        else:
-            raise TypeError("{} indexing is not supported".format(type(item)))
-
-        # Return a copy of the current event if we have already reached it
-        if current is not None and item == current:
-            return deepcopy(self._current_event)
-
-        # If requested event is less than the current event position: reset
-        if current is not None and item < current:
-            self.reset()
-
-        # Check we are within max_events range
-        if not use_event_id and self.max_events and item >= self.max_events:
-            msg = ("Event index {} outside of specified max_events {}"
-                .format(item, self.max_events))
-            raise IndexError(msg)
-
-        if not use_event_id:
-            return self._get_event_by_index(item)
-        else:
-            return self._get_event_by_id(item)
-
-    def _get_event_by_index(self, index):
-        """
-        Method for extracting a particular event for this file format by
-        event index.
-        If a file format allows random event access, this function can be
-        overrided for a more efficient method.
-
-        Parameters
-        ----------
-        index : int
-            The event_index to seek.
-
-        Returns
-        -------
-        event : ctapipe.io.container
-            The event container filled with the requested event's information
-
-        """
-        for event in self._source:
-            if event.count == index:
-                self._current_event = event
-                return deepcopy(event)
-        raise IndexError("Event index {} not found in file".format(index))
-
-    def _get_event_by_id(self, event_id):
-        """
-        Method for extracting a particular event for this file format by
-        event id.
-        If a file format allows random event access, this function can be
-        overrided for a more efficient method.
-
-        Parameters
-        ----------
-        event_id : int
-            The event_id to seek.
-
-        Returns
-        -------
-        event : ctapipe.io.container
-            The event container filled with the requested event's information
-
-        """
-        for event in self:  # Event Ids may not be in order
-            if self.max_events and event.count >= self.max_events:
-                break
-            if event.r0.event_id == event_id:
-                self._current_event = event
-                return deepcopy(event)
-        raise IndexError("Event id {} not found in file".format(event_id))
-
-    def __len__(self):
-        if self.is_stream:
-            raise IOError("Input is a stream, __len__ is disabled")
-
-        # Only need to calculate once
-        if not self._num_events:
-            self.reset()
-            self.log.warning("Obtaining length of file... "
-                             "(potentially long process)")
-            count = 0
-            for _ in self:
-                if self.max_events and count >= self.max_events:
-                    break
-                count += 1
-            self._num_events = count
-        return self._num_events
 
 
 # EventFileReader imports so that EventFileReaderFactory can see them
--- a/ctapipe/io/eventfilereader.py
+++ b/ctapipe/io/eventfilereader.py
@@ -71,6 +71,9 @@ class EventFileReader(Component):
 
         Provenance().add_input_file(self.input_path, role='dl0.sub.evt')
 
+        self.source = self._generator()
+        self.current_event = None
+
     @staticmethod
     @abstractmethod
     def is_compatible(file_path):
@@ -109,7 +112,7 @@ class EventFileReader(Component):
         """
 
     @abstractmethod
-    def __iter__(self):
+    def _generator(self):
         """
         Abstract method to be defined in child class.
 
@@ -119,6 +122,20 @@ class EventFileReader(Component):
         -------
         generator
         """
+
+    def reset(self):
+        """
+        Recreate the generator so it starts from the beginning
+        """
+        self.source = self._generator()
+        self.current_event = None
+
+    def __iter__(self):
+        # Always reset generator when starting a new iteration
+        self.reset()
+        for event in self.source:
+            self.current_event = event
+            yield event
 
     def __getitem__(self, item):
         """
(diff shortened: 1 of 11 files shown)


watsonjj:
@dneise I have implemented your suggestion for the EventFileReader, but with a few differences:

1. `__next__` does not work correctly if `yield` is used inside. Instead the abstract method (for filling the event container) is `_generator`, `source` is then an instance of that generator and `__next__` correctly loops through this generator
2. I needed a `reset` method to reset the generator, as otherwise the functionality of `__getitem__` is flawed; doing `reader[4]` and then `reader[2]` afterwards would not work as it has already gone past that element in the generator

If we are happy with this new layout, then I will update the examples that use EventFileReaders

dneise:
Ah yes .. my mistake, this should no go to `__next__`, but `__iter__` then it works. 
I often mix that up...

watsonjj:
Okay, that will change EventFileReader from an [iterator](https://docs.python.org/3/glossary.html#term-iterator) to an [iterable](https://docs.python.org/3/glossary.html#term-iterable). We would no longer need the separate `_generator` method and the `reset` method. But we will no longer be able to do `event = next(reader)`... which I think is okay.

kosack:
> Okay, that will change EventFileReader from an iterator to an iterable. We would no longer need the separate _generator method and the reset method. But we will no longer be able to do event = next(reader)... which I think is okay (we can still do event = reader[1] or for event in reader:).

That does seem to be a bit nicer.  We may need to change how `pyhessio` works to support this (right now it only exposes a generator). In that case we can get rid of things like hessio_event_source() (which just gets merged into the new scheme). For backward compatibility, we could always just have the `hessio_event_source()` method construct a FileReader and return the iterator. 

The other reason this may be nice, is that we can finalle separate the reading of *event* data and *instrument* configuration in the classes in a nicer way, so that `next(reader)` gives the next event, but `reader.instrument` gives you a properly filled in `instrument.SubarrayDescription` object or something (getting rid of the `event.inst` field is something i've been meaning to do for a while). That can all happen after the refactoring, of course.

watsonjj:
> That does seem to be a bit nicer.  We may need to change how `pyhessio` works to support this (right now it only exposes a generator). In that case we can get rid of things like hessio_event_source() (which just gets merged into the new scheme). For backward compatibility, we could always just have the `hessio_event_source()` method construct a FileReader and return the iterator. 

`pyhessio` working by exposing a generator is completely fine in this design. Each **new** iteration over reader, or calling of `__getitem__` correctly starts that generator from the beginning again, exactly as one would hope.

The other reason this may be nice, is 

> that we can finalle separate the reading of *event* data and *instrument* configuration in the classes in a nicer way, so that `next(reader)` gives the next event, but `reader.instrument` gives you a properly filled in `instrument.SubarrayDescription` object or something (getting rid of the `event.inst` field is something i've been meaning to do for a while). That can all happen after the refactoring, of course. 

Yes that does seem nicer. However as I just mentioned `next(reader)` will not work, as EventFileReader will be an [iterable](https://docs.python.org/3/glossary.html#term-iterable), not an [iterator](https://docs.python.org/3/glossary.html#term-iterator) (no `__next__` method defined). But I hope you agree this isn't a problem.

MaxNoe:
> Each new iteration over reader, or calling of __getitem__ correctly starts that generator from the beginning again, exactly as one would hope.

Having `__getitem__` suggests cheap random access, I would not provide `__getitem__` if it means that each time I call it, a new generator will be  created and looped to the requested event.

watsonjj:
> > Each new iteration over reader, or calling of __getitem__ correctly starts that generator from the beginning again, exactly as one would hope.
> 
> Having `__getitem__` suggests cheap random access, I would not provide `__getitem__` if it means that each time I call it, a new generator will be  created and looped to the requested event.

For data formats that do not allow random event access this would be the only choice to seek to a particular event. However for the data formats that do allow random event access their reader can override the `__getitem__` method to allow fast event access.

I understand your point that the default accessing is not cheap, but I think this is an acceptable approach for a common functionality.

watsonjj:
Either we have the accessing of particular events in the `__getitem__`, or as a `get_event()` method as we had before. Either one has the same problem that the accessing of the element is not cheap, which might not be immediately obvious to the user. But I prefer the first option as having the reader as a [sequence](https://docs.python.org/3/glossary.html#term-sequence) seems appropriate.

MaxNoe:
I agree that having `__getitem__` is a good thing for readers. One has to be careful with the the implementation though. Maybe a warning if the underlying file format or library does not allow random access. And only restarting the generator if the requested event is *before*  the current event.

watsonjj:
> I agree that having `__getitem__` is a good thing for readers. One has to be careful with the the implementation though. Maybe a warning if the underlying file format or library does not allow random access. And only restarting the generator if the requested event is *before*  the current event. 
 
A warning seems like a good idea.

In the [iterable](https://docs.python.org/3/glossary.html#term-iterable) design, the restarting of the generator is automatic, and the holding of a position would complicate things...

watsonjj:
Perhaps we could expose a methods in which a user can supply a list of events they want, then the reader can more efficiently return them?

watsonjj:
> It can be implemented efficiently to just reset a generator if necessary or even implement caching

In that case it would be better to go back to what I had in https://github.com/cta-observatory/ctapipe/pull/613/commits/cab25b1d02ddaf35b756a93ed765d42df88336c8

watsonjj:
Due to the way `pyhessio` is, you can only have one generator at a time as we can only have one file open in `pyhessio` at a time. If we are keeping reference of our location in the generator after doing `event = reader[5]`, then we can't create a fresh `for event in reader:` until we end that existing generator. We can do all that manually, like I did back in that commit I just mentioned, but then the class isn't as simple as it is currently, as it needs the `_generator` and `reset` method again.

watsonjj:
I have now included the functionality thats been requested so far:

1. The position in the file is remembered for faster seeking to further events
2. Starting a new iteration over the EventFileReader (`for event in reader:`) will reset the reader back to the first element first.
3. `__getitem__` also accepts slices and lists
4. `pyhessio.close_file()` is called, along with a warning, if any additional HessioFileReaders are created, to avoid HessioError. Otherwise the following would raise the error:
```
from ctapipe.io.hessiofilereader import HessioFileReader
reader = HessioFileReader(None, None)
event = reader[4]
reader = HessioFileReader(None, None)
event = reader[4]
```
5. A warning is shown (only once) for when a user uses the ineffient default seeking loop in `__getitem__`.

dneise:
Maybe my proposal to have the `__getitem__` in the baseclass was not good. 

If we say: Developers of file formats that do allow quick random access are free to implement the sequence interface, i.e. `__getitem__` and `__len__`.

Devs, who can implement `__len__` cheaply, may implement `__len__`, but not `__getitem__`.

This would lead to a world, where users can try `my_reader[123]` but it might fail. 
I assume direct event access is not part of your vanilla analysis, but rather happens when somebody is playing around with a file in a ipynb or so... 
So people will get at exception and think: "oh this is a crappy file format, which does not even allow direct event access, we better quickly use the final format, which gives us that feature."

I see that having readers, which do not present an identical interface is sub-optimal, but the base-class should only implement the **minimal** interface all readers can implement. 

If e.g. there is a reader, which does not only allow direct access, but also allows direct access by event meta.data, like trigger-id. We cannot force a developer not to implement that feature for their reader. 
However they cannot expect the to find the same feature in any other reader.

If we, by default, implement direct event access, i.e. we make it part of the reader-interface. Then people may expect this to work **without a warning**. 

So we should not make it part of the reader interface.

This means, when a factory spits out a reader, one should only use the features from the reader-interface.

watsonjj:
I think element access is a vital part of the reader and theres no reason not to include it even if its inefficient. If the high level interface access is allowed to be different between data formats, then it removes the point of having a high level interface.

Most of the time users will only want to iterate over the file. But there are some special cases where element access is useful, such as in a GUI that shows individual events.

I see no reason to exclude this functionality.

watsonjj:
If we didn't have the hessio format as our basic format, we wouldn't be having this discussion - it would be clear to have this functionality. Hessio is unlikely to be the final format, and majority of formats will probably allow random event access. I would appreciate it if we didn't discard this functionality (of which I have spent time improving its efficiency), and we moved on to the next topic as there is still a lot to be done.

MaxNoe:
> If we didn't have the hessio format as our first format, we wouldn't be having this discussion

No. There are a lot of sane data formats which do not provide efficient random access.

watsonjj:
It is a feature I require for my work, and I doubt I will be the only one who needs it. I insist we keep it.

kosack:
I think it's fine to leave the event-seeking functionality in `__get_item__` as you have it. We could always expose an attribute like "has_fast_seek" or something that tells the user that this format  has efficient random access, so that GUIs and things can decide how to use it . But I think it's nice to have a common way to do it.  I think we could even hack eventio to have faster random access by doing a first pass through the data, building an index table, and then using `fseek` to jump between events.  

As mentioned, the final format may support random access anyhow.  One other issue is if the input is a *stream* rather than a file, we could just automatically disable `__get_item__` and throw an exception.  Hessio already supports reading from a stream or via  URL (try it - it even works in ctapipe), so this may be necessary in those cases.

kosack:
The tests are currently only failing due to a bug in the conda-forge version of pyzmq, which we don't really need. I'll see if I can lock it to an earlier version.

MaxNoe:
> It is a feature I require for my work, and I doubt I will be the only one who needs it. I insist we keep it.

There is a difference in requiring every reader to implement it and having it optionally for readers who might be able to support it.

A base class should only require the minimum. The minimum would be from my perspective being an iterable over events. Random access is a nice to have feature, not a requirement. 
If you need random access, the most easy thing to do would be creating a list of events by looping over a reader, that is a perfectly fine thing to do for debugging purposes. Any real analysis will just loop over complete files in order.

watsonjj:
> If you need random access, the most easy thing to do would be creating a list of events by looping over a reader, that is a perfectly fine thing to do for debugging purposes. Any real analysis will just loop over complete files in order.

I work with both hessio and our custom camera files, and I require this functionality in both of them. It has been a functionality I have used since EventFileReader was created.

> There is a difference in requiring every reader to implement it and having it optionally for readers who might be able to support it.

If `__getitem__` is defined in the base class, then new readers don't need to implement it. Every reader (except if they are stream) is compatible with the default `__getitem__` method. And it is written efficiently so that other users who want to have this functionality don't have to create this loop themselves each time.

watsonjj:
Someone creating a new Reader doesn't need to care about `__getitem__`. The typical user doesn't care about `__getitem__`. But neither are affected by it existing for the users who need this functionality. 

There are two possibilities for event retrieval:
1. Either the event is obtained by looping through events (the default way), much like how a user would have to define themselves if they want this functionality.
OR
2. The event is obtained in a more efficient way defined by that particular sub-EventFileReader
Having it exist in this way in the base class allows the user to not care at all about the underlying way the event is obtained, either its from looping through the event or its from a more efficient method, which they benefit from without having to do any further consideration on their part.

And as Karl said, it may even be possible for random event retrieval to be improved for Hessio files. If `__getitem__` is already defined in this manner, then it is really simple to upgrade it, and users who use this feature don't need to change anything in their scripts to gain from the upgrade.

I really don't see a reason to remove this functionality...

kosack:
the tests all pass now, so let me know if you want me to review and merge this, or if you want to make more changes/comments.  @dneise, also let me know if you agree with this implementation (we can of course refactor further later).

ctapipe/io/hessiofilereader.py
```diff
@@ -0,0 +1,211 @@
+from astropy import units as u
+from astropy.coordinates import Angle
+from astropy.time import Time
+from ctapipe.core import Provenance
+from ctapipe.io.eventfilereader import EventFileReader
+from ctapipe.io.containers import DataContainer
+from ctapipe.instrument import TelescopeDescription, SubarrayDescription
+
+
+COUNT = 0
```
MaxNoe: This should be a class variable, not a global variable, move it below `class Hess...`

ctapipe/io/hessiofilereader.py
```diff
@@ -0,0 +1,211 @@
...
+
+
+class HessioFileReader(EventFileReader):
```
MaxNoe: Needs a Docstring

MaxNoe: I think `EventFileReader` should support `with EventFileReader() as reader`. The HessioReader should then call `close_hessio` in the `__exit__` method.

ctapipe/io/hessiofilereader.py
```diff
@@ -0,0 +1,211 @@
...
+                          "Previous hessio file will be closed.")
+            self.close_hessio()
+        COUNT += 1
+
+    @staticmethod
+    def is_compatible(file_path):
+        return file_path.endswith('.gz')
+
+    @property
+    def camera(self):
```
MaxNoe: Why is it called `camera`? At least HessIO files contain Array Events with different cameras.

ctapipe/io/hessiofilereader.py
```diff
@@ -0,0 +1,211 @@
...
+        return 'hessio'
+
+    def _generator(self):
```
MaxNoe: This can be `__iter__`, right?

MaxNoe: Ah, I see, this might create problems with the Only 1 file open constraint of pyhessio.

ctapipe/io/hessiofilereader.py
```diff
@@ -0,0 +1,211 @@
...
+            from pyhessio import close_file
+            from pyhessio import HessioError
+            from pyhessio import HessioTelescopeIndexError
+            from pyhessio import HessioGeneralError
+        except ImportError:
+            msg = "The `pyhessio` python module is required to access MC data"
+            self.log.error(msg)
+            raise
+
+        self.open_hessio = open_hessio
```
MaxNoe: I do not understand why we gain from assigning these functions and Exceptions to the reader, this looks like just adding noise.

ctapipe/io/eventfilereader.py
```diff
@@ -149,215 +98,202 @@ def check_file_compatibility(file_path):
 
     @property
     @abstractmethod
-    def num_events(self):
+    def camera(self):
```
MaxNoe: This confuses me. Why is this a property of the EventFIleReader? A File might contain array events with very different Cameras like it's the case with the current MCs.

ctapipe/io/eventfilereader.py
```diff
@@ -149,215 +98,202 @@ def check_file_compatibility(file_path):
...
-            select only a subset of telescope, if None, all are read. This can
-            be used for example emulate the final CTA data format, where there
-            would be 1 telescope per file (whereas in current monte-carlo,
-            they are all interleaved into one file)
-        requested_event : int
-            Seek to a paricular event index
-        use_event_id : bool
-            If True ,'requested_event' now seeks for a particular event id
-            instead of index
+        Generator where the filling of the `ctapipe.io.containers` occurs.
```
MaxNoe: I think we should have a single method that fills a single event container. Not just a method that creates a generator for all events.

ctapipe/io/eventfilereader.py
```diff
@@ -149,215 +98,202 @@ def check_file_compatibility(file_path):
...
-                           use_event_id=use_event_id)
-        event = next(source)
-        return deepcopy(event)
+        self.source = self._generator()
+        self.current_event = None
 
-    def find_max_true_npe(self, telescopes=None):
+    def __iter__(self):
+        # Always reset generator when starting a new iteration
+        self.reset()
```
MaxNoe: I think this is very counter intuitive. When I do:

```
reader = EventFileReader()
event = next(reader)
for event in reader:
    pass
```
Would this start with event 0 in the loop again?

ctapipe/io/eventfilereader.py
kosack: I tried implementing it that way, and it ended up being very complex (and would require a re-write of pyhessio).  It's not a fundamental problem, but we can maybe leave that to a future refactoring and keep it the way it is for now.

ctapipe/io/hessiofilereader.py
watsonjj: In the future there will be only one camera per file.

This exists so that scripts decide at run time the correct default operations to use according to the camera, for example the correct CameraR1Calibrator or ChargeExtractor. In my next PR I will create a class that takes a camera name and sets up the traitlet config so the correct Components are used for the camera.

It makes sense for the camera field to exist here as the camera type can be supplied from the file (either from a header or inferred from the data format).

Hessio files require the HessioR1Calibrator to be applied to them. So that will be set as the default R1Calibrator when given "hessio" as the input. This interface will be further improved in the future when we do have single cameras per file, and then camera-specific operations can be defined for simulation files.

ctapipe/io/eventfilereader.py
kosack: I guess that's the difference between a generator and an iterable. The latter would re-start and the former would continue... If we want these to be generator-like, it makes sense to have the user explicitly do a reset, but that can also be confusing for people not familiar with generators.

ctapipe/io/hessiofilereader.py
kosack: perhaps this is meant to give the file type? In which case it could be renamed? I'm not sure why it is there either.

ctapipe/io/eventfilereader.py
watsonjj: Firstly there is no `__next__` method defined. Again, EventFileReader is currently an [iterable](https://docs.python.org/3/glossary.html#term-iterable) not an [iterator](https://docs.python.org/3/glossary.html#term-iterator) .

But yes, every time you start a new `for event in reader:` it starts from event 0 again. 
If you didn't do that an operation like:

```
for event in reader:
   if event.count == 5: 
       break
for event in reader:
   if event.count == 2: 
       break
```

Wouldn't work. For me its more intuitive that the loop starts at the beginning of the file again when starting a new loop. Do you disagree?

watsonjj: When you do:

```
example_list = [3, 4, 6, 8, 9]
for el in example_list:
   if el == 8:
      break
for el in example_list:
   if el == 4:
      break
```

You expect it to work...

ctapipe/io/hessiofilereader.py
kosack: yes, another reason why it will be nice to have a more modern format :-)

watsonjj: Actually pyhessio would be fine with this as `__iter__` (as the `with` block is exited when a new `__iter__` is started. I had to move the generator from `__iter__` to its own function so I can keep reference to the generator. Therefore an operation like:
```
event = reader[3]
event = reader[3]
event = reader[4]
event = reader[2]
```
Continues from the last point and only starts from the beginning of the file again on the last line.

watsonjj: Of course, if we kept a reference to the generator, it does have a problem with the 1-file pyhessio limit, that is why I recreate the reference the generator in `__iter__` (`reset`).

ctapipe/io/hessiofilereader.py
watsonjj: @kosack If the `inst` container could somehow indicate that the file is a "simulated" camera from sim_telarray, and therefore HessioR1Calibrator should be used, then we could remove the `EventFileReader.camera` property, and rely on the `inst` container for such information.

ctapipe/io/eventfilereader.py
kosack: I think it's fine this way as long as it's documented.  In fact, there have been a few people confused that `hessio_event_source` didn't restart this way (since most people new to Python are not really familiar with generators)

ctapipe/io/hessiofilereader.py
kosack: Yes, I think we can add that functionality.  I can just add an "is_simulated" attribute to SubarrayDescription.  

FileReaders should only fill in the cameras that are in the file into SubarrayDescription, so then you can query it to see which camera is there (and we support both merged and unmerged camera files simultaneously).  I may have to rethink the naming, or add another description class, since "Subarray" should really refer to all the tels, not just the one in the file.

I have another suggestion: since all input files may have some generic headers (version number, etc) defined, why not just have a method or attribute called "metadata" that returns for now a dict, and you can keep an item like CAMERA in that? That way we can pyt other into in it, like version since the schema of the format may evolve.  In fact, it should really be "TELESCOPE" not "CAMERA" since it depends also on the OpticsDescription, etc. So it should be filled with a value that is the same as  what `str(TelescopeDescription)` gives you (like "SST-GCT:CHEC")

kosack: So more concretely, for now just have:

```py3
def metadata(self):
     return self.metadata
```
where `self.metadata` is filled in when the file is opened by the particular subclass, and can minimally be just `{'telescope': tel_identifier}`, and later we can also add things like the `{'obs_id' : 12345, is_simulation:True}`, etc. (to be defined)

kosack: or you could write that info into `event.meta` which is already there. We just have to define a common place to put file metadata.

watsonjj: Maybe `event.meta` isn't the best place, as you would need to retrieve an event container (by beginning to loop through the events) to be able to retrieve that information... And also it is information that does not change with event.

I prefer the `reader.metadata` and `reader.subarray` attributes.

ctapipe/io/hessiofilereader.py
watsonjj: Okay, I have removed the unused ones. I need to keep a reference to `open_hessio` and `close_file` for other methods in the class.

ctapipe/io/hessiofilereader.py
watsonjj: Should we create a metadata class instead of using a dictionary so the contents are consistent and predictable?

ctapipe/io/hessiofilereader.py
MaxNoe: Why? Why is the top level import not enough?

ctapipe/io/hessiofilereader.py
```diff
@@ -0,0 +1,211 @@
...
+            self.log.warn("Only one pyhessio reader allowed at a time. "
+                          "Previous hessio file will be closed.")
+            self.close_hessio()
+        HessioFileReader._count += 1
+
+        self._metadata['is_simulation'] = True
+
+    @staticmethod
+    def is_compatible(file_path):
+        return file_path.endswith('.gz')
```
dneise: What if other readers also read gzipped files, like "something.fits.gz" ? Then the factory **wrongly** tries to use this reader, if this reader happens to be asked first ..

ctapipe/io/hessiofilereader.py
```diff
@@ -0,0 +1,211 @@
...
+            counter = 0
+            eventstream = pyhessio.move_to_next_event()
+            if self.allowed_tels is not None:
+                self.allowed_tels = set(self.allowed_tels)
+            data = DataContainer()
+            data.meta['origin'] = "hessio"
+
+            # some hessio_event_source specific parameters
+            data.meta['input'] = self.input_path
+            data.meta['max_events'] = self.max_events
```
dneise: In my understanding when `something = DataContainer()` then something represents a single event.
Is that correct?

If so, why does a single event know how many events in total (`max_revents`) somebody is planning to maybe read from the file?

ctapipe/io/hessiofilereader.py
```diff
@@ -0,0 +1,211 @@
...
+        with self.open_hessio(self.input_path) as pyhessio:
+            # the container is initialized once, and data is replaced within
+            # it after each yield
+            Provenance().add_input_file(self.input_path, role='dl0.sub.evt')
+            counter = 0
+            eventstream = pyhessio.move_to_next_event()
+            if self.allowed_tels is not None:
+                self.allowed_tels = set(self.allowed_tels)
+            data = DataContainer()
+            data.meta['origin'] = "hessio"
```
dneise: What are people supposed to do with `data.meta["origin"]`? Could you give an example somewhere?

ctapipe/io/hessiofilereader.py
```diff
@@ -0,0 +1,211 @@
...
+
+            # some hessio_event_source specific parameters
+            data.meta['input'] = self.input_path
```
dneise: I think `data.meta["input_path"]` is more clear.

ctapipe/io/hessiofilereader.py
```diff
@@ -0,0 +1,211 @@
...
+                # tels_with_data and allowed_tels
+                if self.allowed_tels is not None:
+                    selected = tels_with_data & self.allowed_tels
+                    if len(selected) == 0:
+                        continue  # skip event
+                    data.r0.tels_with_data = selected
+                    data.r1.tels_with_data = selected
+                    data.dl0.tels_with_data = selected
+
+                data.trig.tels_with_trigger \
```
dneise: backslash line continuation is possibly dangerous, when some poor person adds a single whitespace. 

https://www.python.org/dev/peps/pep-0008/#maximum-line-length
> The preferred way of wrapping long lines is by using Python's implied line continuation inside parentheses, brackets and braces. Long lines can be broken over multiple lines by wrapping expressions in parentheses. These should be used in preference to using a backslash for line continuation.

ctapipe/io/hessiofilereader.py
```diff
@@ -0,0 +1,211 @@
...
+                    data.mc.tel[tel_id].altitude_raw = \
+                        pyhessio.get_altitude_raw(tel_id)
+                    data.mc.tel[tel_id].azimuth_cor = \
+                        pyhessio.get_azimuth_cor(tel_id)
+                    data.mc.tel[tel_id].altitude_cor = \
+                        pyhessio.get_altitude_cor(tel_id)
+                yield data
+                counter += 1
+
+                if self.max_events and counter >= self.max_events:
```
dneise: I think the base class should entirely deal with this whole `max_events` matter, so that explicit child classes do not have to bother about this, and devs to not have to repeat this little if in each `self._generator()` method. 

And I think it is possible. I will make a PR tomorrow.

dneise:
Thanks @kosack  for asking for my opinion. I am not really happy yet. You said, we can always later further improve, but this a definition of an interface, and some people want to implement against that interface, so we should get it "about right-ish". 

I think this interface is still a bit bloated.

watsonjj:
@dneise Thanks for the comments! I will include some of these changes in this PR. And address some of them in replies.

ctapipe/io/eventfilereader.py
```diff
@@ -2,36 +2,83 @@
...
+    are reader (for example you must use
+    `ctapipe.io.hessiofilereader.HessioFileReader` to read a hessio format
+    file). Alternatively you can use
+    `ctapipe.io.eventfilereader.EventFileReaderFactory` to automatically
+    select the correct EventFileReader subclass for the data format you wish
+    to read.
+
+    To loop through the events in a file:
+
+    >>> reader = EventFileReader(None, None, input_path=path)
```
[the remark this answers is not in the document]
watsonjj: Actually the correct way would be to have this Component initialised from inside a `ctapipe.core.tool.Tool`. In which case it would be: 
```
foo = SomeClass(self.config, self)
```
I was going to write that in the docstring, but I was worried that would be more confusing. I will change it, and point the reader to an example script using it. Does that sound better?

ctapipe/io/eventfilereader.py
```diff
@@ -2,36 +2,83 @@
...
+    To obtain a particular event in a file:
+
+    >>> reader = EventFileReader(None, None, input_path=path)
+    >>> event = reader[event_index]
+    >>> print(event.count)
+
+    To obtain a particular event in a file from its event_id:
+
+    >>> reader = EventFileReader(None, None, input_path=path)
+    >>> event = reader["event_id"]
```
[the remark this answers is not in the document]
watsonjj: Well, at least in sim_telarray, event_id does not increase in increments of 1. I think this is because only shower events that triggered the array are stored, but sim_telarray increments the id on all events (take a look at event.r0.event_id for a hessio file).

Having event_index allows the user to not care about this, and just get the next event.

But perhaps the user has noticed something peculiar about a particular event, perhaps in a different analysis software, allowing them to obtain an event using event_id allows them to ensure the same event is obtained between the programs.

ctapipe/io/eventfilereader.py
```diff
@@ -2,36 +2,83 @@
...
-    output_directory : str
-        Directory to save outputs for this file
-
+        Path to the input event file.
+    max_events : int
+        Maximum number of events to loop through in generator
     """
-    origin = None
 
     input_path = Unicode(get_dataset('gamma_test.simtel.gz'), allow_none=True,
```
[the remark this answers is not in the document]
watsonjj: Yes sounds good

ctapipe/io/eventfilereader.py
```diff
@@ -148,216 +148,207 @@ def check_file_compatibility(file_path):
...
         -------
-        self._num_events : int
+        dict
         """
+        return self._metadata
 
     @property
-    @abstractmethod
-    def event_id_list(self):
+    def is_stream(self):
```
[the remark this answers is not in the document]
watsonjj: I was hoping in a future PR someone could define a method to check if a string in being read, in which case `__getitem__` and `__len__` are disabled, because in that circumstance it is impossible for them to operate as expected.

ctapipe/io/eventfilereader.py
```diff
@@ -148,216 +148,207 @@ def check_file_compatibility(file_path):
...
-        requested_event : int
-            Seek to a paricular event index
-        use_event_id : bool
-            If True ,'requested_event' now seeks for a particular event id
-            instead of index
+    def __iter__(self):
+        # Always reset generator when starting a new iteration
+        self.reset()
+        for event in self.source:
+            self.current_event = event
```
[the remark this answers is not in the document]
watsonjj: We store current event so that:
```
event = reader[5]
event = reader[5]
```
Does not need to start the loop from the beginning to recover the event the second time. This is the only way possible, because the generator has already proceeded to the next event when we call the second line.

ctapipe/io/eventfilereader.py
```diff
@@ -148,216 +148,207 @@ def check_file_compatibility(file_path):
...
-class HessioFileReader(EventFileReader):
-    origin = 'hessio'
+        if self.is_stream:
+            raise IOError("Input is a stream, __getitem__ is disabled")
+
+        current = None
+
+        if not self.has_fast_seek and self.getevent_warn:
+            self.log.warning("Seeking to event... (potentially long process)")
+            self.getevent_warn = False
```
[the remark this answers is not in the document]
watsonjj: But as this is a Component it would be more appropriate to use the built-in logger, so the log messages are correctly handled according to how they are set-up. We can change this in the future if needed.

ctapipe/io/eventfilereader.py
```diff
@@ -148,216 +148,207 @@ def check_file_compatibility(file_path):
...
-        if self.max_events:
-            self.log.info("Max events being read = {}".format(self.max_events))
-        source = hessio_event_source(self.input_path,
-                                     max_events=self.max_events,
-                                     allowed_tels=allowed_tels,
-                                     requested_event=requested_event,
-                                     use_event_id=use_event_id)
-        self.log.debug("File reading complete")
-        return source
+        """
```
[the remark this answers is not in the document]
watsonjj: Ok.

I think id and index can be differentiated as the former is a unique identifier associated with the event itself, not necessarily in order or in single increments, and the latter is associated with the order the events are read from the file, and go up in single increments. I will add this in the EventFileReader docstring.

ctapipe/io/eventfilereader.py
```diff
@@ -148,216 +148,207 @@ def check_file_compatibility(file_path):
...
+            msg = "Event id {} not found in file".format(ev)
+            for event in self:  # Event Ids may not be in order
+                if event.r0.event_id == ev:
+                    self.current_event = event
+                    return deepcopy(event)
+        raise IndexError(msg)
+
+    def __len__(self):
+        if self.is_stream:
+            raise IOError("Input is a stream, __len__ is disabled")
```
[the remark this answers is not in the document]
watsonjj: This is the exact same argument you could use against removing `__getitem__` from the base class and forcing the sub-classes to define it. However in this case with the streams, it is physically impossible (without some extra functionality like caching). 

As it it only makes sense to use streams when using the pipeline processing, where we go through event by event, I don't think we should be concerned that this throws an error if someone tries to use a stream in their debugging script or GUI, they should clearly be using the file itself.

ctapipe/io/hessiofilereader.py
watsonjj: If that situation arises then we need to improve this function to better detect what a compatible file it. But we will do that when the situation occurs.

ctapipe/io/hessiofilereader.py
watsonjj: It could be removed as the reading is explicitly attached to an EventFileReader now (which explicitly tells you the "origin")

ctapipe/io/hessiofilereader.py
watsonjj: Maybe we should remove `data.meta` now as event reading is attached to an EventFileReader which has all this information.

The only reason you may want to keep it is if you want a function that takes `event` to have easy access to the information...

ctapipe/io/hessiofilereader.py
dneise: Sounds fair enough

ctapipe/io/hessiofilereader.py
watsonjj: We will be directing EventFileReader-creator to perform their import of external modules in the `__init__` method to not introduce dependencies on those modules.

To then be able to use that module elsewhere in the class, one must keep a reference to the import.

As pyhessio is a dependency of ctapipe, in this case we could import pyhessio at the top of this module instead of inside `__init__`, but I thought it better to keep things consistent between EventFileReaders, and besides maybe one day we will drop the dependency on pyhessio.

ctapipe/io/eventfilereader.py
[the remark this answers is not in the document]
dneise: > Well, at least in sim_telarray, event_id does not increase in increments of 1. 

Ah yes, this is indeed quite usual in IACT simulation. First N air showers are simulated, the M < N events are triggered and stored in a file. 
One might think events could be numbered consecutively, but for repeatability, the random seeds and all settings are stored somewhere, and in order to reproduce a single event later, one needs to know the "airshower-id" it is associated with and not the "trigger-id".

However this pattern goes on, right? After that, some kind of cleaning is applied and maybe a cut, which cuts away all events that did not "survive cleaning". 

So for simulated events, each event now has: "airshower-id", "trigger-id", "survived-cleaning-id" 

And we can go on: "survived-some-precuts", "survived-background-cuts", "survived-some-quality-cuts" ...

Any process, which selects some events out of many other events, results in a new "sub-stream", where one might say: 

"Hery! Why are the ids of these events not consecutively increasing but have "holes" in them?"

I do not think that the solution is to invent new beautiful ids after each cut, but to accept that event ids are simply not increasing in steps of 1. They are not so much counters but "telephone-numbers" each event has one .. it is unique ... but it is not increasing in steps of 1.

I would discourage people from inventing two different but unique ids for events. Since, how do people talk to each other then? 

"Hey have you looked at event 14532 in run 20190101_1432?" 

"what do you mean id=14532 or index=14532?"

"erm .. wait a sec ... dunno I closed the window"

Not so useful, is it?

ctapipe/io/hessiofilereader.py
watsonjj: I have reduced clutter now by only keeping a reference to the pyhessio module itself, and not its separate functions

ctapipe/io/hessiofilereader.py
watsonjj: Ok

ctapipe/io/hessiofilereader.py
kosack: yeah, in the future we can use a module like *magic* that detects file types by magic numbers or mime types. Depends on how many variants we end up with! At least here, you might want to check for `".simtel.gz"`

ctapipe/io/eventfilereader.py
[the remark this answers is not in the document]
watsonjj: Sure, but theres at least 2 clear identification of events:
event_id: The ID associated by the software or camera (or CTA trigger in the final array). So can be thought of an external ID unassociated with ctapipe.
event_index: The ID associated with the event by ctapipe, assigned in the order the events are read from the file.

I think this is clear enough. I don't think we can rely on event_id alone. 
1. Some sim_telarray files are combined with others into a final file, allowing for duplicate event_ids.
2. With `__gettitem__` I wish to be able to get the first or the nth event in the file, a system that doesn't work if I cannot get an event using event_index (i.e. order they are read from the file).
3. Theres nothing to say event_id 101 is directly after event_id 100. So we can't do anything smart in `__getitem__` to avoid looping from the start of the file again to find it. But event_index 4 is directly after event_index 3, so the loop to find an event based on event_index can operate smarter and just continue onto the next event in the generator without starting from the beginning in the file.

ctapipe/io/hessiofilereader.py
MaxNoe: EventIO file are quite easy to identifiy, they all start with two special magic bytes. gz files as well.

So we could do 
```python
with open(path, 'rb') as f:
    marker_bytes = f.read(4)
if  marker_bytes[0] == 0x1f and marker_bytes[1] == 0x8b:
    with gzip.open(path, 'rb') as f:
        marker_bytes = f.read(4)

return struct.unpack('I', marker_bytes)[0] == 3558836791
```

MaxNoe: First case is little endian, second case big endian

ctapipe/io/eventfilereader.py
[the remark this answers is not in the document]
MaxNoe: > This is the exact same argument you could use against removing __getitem__ 
We did.

We see here maybe another hierarchy would be appropriate:
`
class EventReader:
    pass

class EventFileReader(EventReader):
    pass

class EventStreamReader(EventReader):
    pass
````

File: ctapipe/io/hessiofilereader.py (hunk 12 above)
Comment (MaxNoe) [reply ctapipe/io/hessiofilereader.py]: could be a url in case of zmq input

Comment (MaxNoe) [reply ctapipe/io/hessiofilereader.py]: So maybe use fully qualified urls `file:path` `http://<>/``` etc

Comment (dneise) [reply ctapipe/io/hessiofilereader.py]: year in that case this should maybe be called `input_url` .. I just though `input` is not very specific.

File: ctapipe/io/eventfilereader.py (hunk 22 above)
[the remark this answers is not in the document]
Comment (watsonjj) [reply ctapipe/io/eventfilereader.py]: Yes...maybe another hierarchy... but the one you've suggested would require defining how the events are read for each data format twice.

How about a wrapping class (not sure if theres a python convention on how to do these):

```
class EventFileReader: 
   pass

class EventSeeker:
   def __init__(eventfilereader):
      pass

   def __getitem___(self):
      pass
```

EventFileReader is exactly as we have it, but without __getitem__. Each data format creates their own Reader that inherits this and defines how the events containers are filled

EventSeeker takes an EventFileReader, and defines __getitem__ as a loop through the EventFileReader (as it currenly exists in EventFileReader.__getitem__), unless that particular EventFileReader defines __getitem__ itself (because that data format has a random event retrieval method).

How does that sound?

File: ctapipe/io/eventfilereader.py (hunk 20 above)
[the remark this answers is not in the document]
Comment (MaxNoe) [reply ctapipe/io/eventfilereader.py]: Logging with level `warning` is something different than a true warning. Here a true warning is more appropriate.

File: ctapipe/io/eventfilereader.py (hunk 17 above)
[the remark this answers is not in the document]
Comment (MaxNoe) [reply ctapipe/io/eventfilereader.py]: Why are we using a class configurable trait here? I think this is the wrong approach as this would set the value for each instant of the class. I think it should just be a parameter to `__init__`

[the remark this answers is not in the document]
Comment (watsonjj) [reply ctapipe/io/eventfilereader.py]: Because these traits are how we pass command line arguments (or configuration files) to the Components.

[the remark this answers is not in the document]
Comment (watsonjj) [reply ctapipe/io/eventfilereader.py]: If one wants multiple input files in a script, one could create them as configurable traits in the script, and pass directly to the EventFileReader:
```python
reader1 = EventFileReader(self.config, self, input_path=path1)
reader2 = EventFileReader(self.config, self, input_path=path2)
```

Keeping the traitlet in EventFileReader means the user does not have to define it in the usual case where 1 file is being read with 1 reader.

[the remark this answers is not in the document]
watsonjj: I would say this falls under a discussion of traits/components/tools and could be discussed in a different PR.

ctapipe/io/eventfilereader.py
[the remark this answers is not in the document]
watsonjj: Then in that case EventFileReader is only used to loop through events, compatible for both files and streams, and is @dneise and your's desire to have a more basic EventFileReader definition. Whereas EventSeeker is defined for those who want `__getitem__` functionality and takes any EventFileReader as input.

ctapipe/io/hessiofilereader.py
watsonjj: Sounds good, I'll add that now

ctapipe/io/hessiofilereader.py
watsonjj: No need, I've made this change

watsonjj:
I have proceeded with creating the `EventSeeker` class. This change addresses the request for a more basic `EventFileReader` class (without a `__getitem__` method), but allows one to obtain that functionality by wrapping the reader in the `EventSeeker` class, providing an efficient and common way to extract particular events from any file format (as long as it is not a stream).

I hope everyone is now satisfied with this approach.

ctapipe/io/eventfilereader.py
```diff
@@ -148,216 +130,90 @@ def check_file_compatibility(file_path):
         """
 
     @property
-    @abstractmethod
-    def num_events(self):
+    def metadata(self):
```
dneise: why is `metadata` a property again and not just a simple attribute? 
Usually one would make it a property when one wants to do more than only returning it.

ctapipe/io/eventfilereader.py
```diff
@@ -2,38 +2,75 @@
...
+        Path to the input event file.
+    max_events : int
+        Maximum number of events to loop through in generator
     """
-    origin = None
 
-    input_path = Unicode(get_dataset('gamma_test.simtel.gz'), allow_none=True,
+    input_path = Unicode('', allow_none=False,
                          help='Path to the input file containing '
                               'events.').tag(config=True)
```
dneise: Just a side note:
`allow_none=False` is the default of `Unicode`. 

Indentation: I think this way of indenting long lines is sub-optimal
```python
input_path = Unicode('', allow_none=False,
                     help='Path to the input file containing '
                     'events.').tag(config=True)
```

Indenting like this:
```python
input_path = Unicode(
    '', 
    allow_none=False,
    help='Path to the input file containing events.'
).tag(config=True)
```

has the following advantages:
 * No need to modify indentation, when `Unicode` gets replaced by a different class  (i.e. stable under refactoring / renaming)
 * generally more space available for parameters, in this case, the help string does not need to be broken around.
 * in case of testing: single parameters can be commented out more easily, since each is on their own line.
```

File: ctapipe/io/eventfilereader.py (hunk 18 above)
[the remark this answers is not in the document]
Comment (dneise) [reply ctapipe/io/eventfilereader.py]: So .. erm .. if this feature is planned ot be introduced in the future only ... maybe ... then. Do you really want to include it into this PR?

Let me ask like this: Could you implement a test, which shows that this feature is working correctly?

Also, this class does not have `__getitem__` and `__len__`. So .. is this method correct here?

File: ctapipe/io/eventfilereader.py [hunk 25]
```diff
@@ -60,76 +97,21 @@ def __init__(self, config, tool, **kwargs):
...
-            raise ValueError("Subclass of EventFileReader should specify "
-                             "an origin")
-
-        self._num_events = None
-        self._event_id_list = []
-
-        if self.input_path is None:
-            raise ValueError("Please specify an input_path for event file")
-        self._init_path(self.input_path)
+        self._metadata = dict(is_simulation=False)
```
dneise: Can this class really already know if `is_simulation` is True or False?

watsonjj: At least at the moment only outputs from sim_telarray are stored in hessio files

watsonjj:
> So .. erm .. if this feature is planned ot be introduced in the future only ... maybe ... then. Do you really want to include it into this PR?
> 
> Let me ask like this: Could you implement a test, which shows that this feature is working correctly?
> 
> Also, this class does not have __getitem__ and __len__. So .. is this method correct here?

For some reason I can't reply directly to your comment @dneise 

Well, I want the functionality to be build reader (in `EventSeeker`) for when someone comes along who can fill this gap in my knowledge. This also says to a future programmer than we want this feature to be implemented.

As we do not actually use streams yet, its not urgent to implement in this PR

watsonjj:
@MaxNoe @dneise Is this PR now in a state you are happy with and it can be approved?

dneise:
> @dneise Is this PR now in a state you are happy with and it can be approved?
Honestly no.

But we are getting closer.  
(
Btw. in case you have the impression, I am doing this is order to make you angry or so .. this is not the case. And believe me I am not enjoying this at all
)

dneise:
* is_stream should be removed in my opinion.
 * I still don't get why `self.metadata = dict(is_simulation=False)`  is contained in `EventFileReader`
 * I still don't get why `config` and `tool` cannot have None as default args, since clearly a ready does not **need** config or tool to function properly. It only needs a path

dneise:
Ah and then there are missing tests.... gimme a sec to list them:

watsonjj:
> * is_stream should be removed in my opinion.
is_stream is a very valid property of the reader, and there are use-cases where one might want to check for it

>  * I still don't get why `self.metadata = dict(is_simulation=False)`  is contained in `EventFileReader`
It was agreed it would be useful for a place for metadata in the reader. How that is implemented may be changed in the future. The minimum it needs is to know if its a simulation file, as then one knows to apply `HessioR1Calibrator` to the events.

>  * I still don't get why `config` and `tool` cannot have None as default args, since clearly a ready does not **need** config or tool to function properly. It only needs a path
This is a question for how `config` and `tool` are handled across the whole project, and should not be addressed in this PR.

These are all minor points that can be changed in the future. You are welcome to create your own PR to suggest changes which we can discuss. 

> Ah and then there are missing tests.... gimme a sec to list them:

There are always the possibility for more tests, again you are welcome to add them. I believe the base functionality is tested and the coverage is satisfactory for this PR to be accepted. We should not delay this PR any further as there is still a lot of work on this topic to be done.

dneise:
A unit test, should use a thing as it should be used later in other people code, to show that it works.
Since the EventFileReader is an interface (i.e. an abstract class) one uses it by implementing it.

Only by using it as it is intended to be used, we can verify that it works as intended, e.g. like this:


```python

class TestReader(EventFileReader):
    '''
    uses input_path to create a generator,
    jsut a little bit different :-D
    '''
    def _generator(self):
        return range(int(self.input_path))

    def is_compatible(self, path):
        return False


def test_can_be_implemented():
    test_reader = TestReader(input_path='10')

def test_is_iterable():
    test_reader = TestReader(input_path='10')
    for _ in test_reader:
        pass

# Next I only list names of tests, I would implement if this was my PR.

def test_max_events_limit_iteration():
def test_metadata_is_emtpy():
def test_input_path_is_logged_as_info():
def test_max_events_path_is_logged_as_info():
def test_input_file_is_added_to_provenance():
def test_is_context_manager():
```

watsonjj:
Very well, sounds reasonable

dneise:
> You are welcome to create your own PR to suggest changes which we can discuss.

Well I think this code should not be merged like this, and I might very well be wrong here. 

So if people think this should be merged, they should go ahead. I have not been convinced otherwise, so I will not lie at you and say that I find this PR good. But you are right, I do not have to like it.

watsonjj:
We have clearly had a lot of disagreement on how to approach this important functionality throughout this process. But I believe I found an alternative that meets your main complaint? And the main functionality is now in a state where we can both agree is good?

You still disagree with two aspects that are not part of the main functionality of the EventFileReader, but both of these are matters we will address in the future. I hope you can meet me halfway here, for the sake of efficiency, and we can agree that some excellent progress and collaboration was made in this PR.

ctapipe/io/hessiofilereader.py
kosack: Actually, I think *path* is confusing, since we use the word to mean "a list of directories to search, separated by colons" (like `$PATH`) elsewhere in the code, and it doesn't mean that here.  Why not, `input_url`, since it can be more than just a filename? (URLs work in general, at least for pyhessio, and we could easily make the others work).

kosack: that comment actually applies to all other uses of `input_path` where it doesn't refer to a search path. For the Container metadata, it doesn't really matter, since we now store that in the Provenance system.

ctapipe/io/hessiofilereader.py
kosack: yeah, data.meta isn't so useful, since now we store the Component parameters in the provenence system, so the information gets recorded.  The idea was that anything in a Container's `.meta` dict would automatically be written to the headers of the output table by the `io.TableWriter`, but in this case that info is maybe not so useful.

ctapipe/io/eventfilereader.py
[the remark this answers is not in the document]
kosack: I agree- that's a refactoring for another PR though... I don't like the "None, None", but it comes from the underlying `traitlets.config` code (at least the first one, the second one is my fault and can be removed, since i don't think we need to keep a pointer to the parent anywhere anymore).  In traitlets, the first None tels the class not to try to read the config from the global configuration file or `sys.argv`, so usually you don't need it unless you're in a notebook or script where you are not inside a `Tool` base class.

ctapipe/io/eventfilereader.py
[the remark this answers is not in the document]
kosack: There's a CTA requirement that we have a unique ID for each event, but so far the MCs do not support that (though we could somehow combine the `mc_shower_id` and the `mc_event_id` into one unique identifier).  They will probably never increment by 1, since you may lose events (non-triggered, didn't read out due to dead time) at the DAQ level, or during data volume reduction.

ctapipe/io/hessiofilereader.py
dneise: While I fully agree that `input_url` should be used, simply because also urls can be specified here. I think using "path" for "a list of directories to be searched in" is confusing:

I have never seen this definition for "path". In Python it seams to be common to use the plural for a list, or any general container of stuff, hence I would recommend using `paths` in such a case. If once wants to convey, that theses paths should be used for searching something, maybe `search_paths` or `path_candidates`, when this group contains many paths, but only one will ultimately be correct.

Also: None of these have a similar definition of `path`.

* http://searchnetworking.techtarget.com/definition/path
* https://techterms.com/definition/path
* https://en.wikipedia.org/wiki/Path_(computing)

ctapipe/io/eventfilereader.py
```diff
@@ -60,76 +107,29 @@ def __init__(self, config, tool, **kwargs):
...
-        if self.origin is None:
-            raise ValueError("Subclass of EventFileReader should specify "
-                             "an origin")
-
         self._num_events = None
-        self._event_id_list = []
+        self._metadata = dict(is_simulation=False)
 
         if self.input_path is None:
             raise ValueError("Please specify an input_path for event file")
```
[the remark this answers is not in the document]
kosack: I'm not sure there is a way to specify required values (or non-keyword values) in traitlets - it's a bit annoying, but so far we're stuck with something like this I think.

ctapipe/io/eventfilereader.py
```diff
@@ -148,216 +139,74 @@ def check_file_compatibility(file_path):
...
 
 
-# External Children
-try:
-    from targetpipe.io.eventfilereader import TargetioFileReader, \
-        ToyioFileReader
-except ImportError:
-    pass
+# EventFileReader imports so that EventFileReaderFactory can see them
+import ctapipe.io.hessiofilereader
```
MaxNoe: I don't think this should be necessary. If this is necessary, it basically forbids external implementations of  `EventFileReader`, which is not good.

ctapipe/io/eventfilereader.py
```diff
@@ -382,19 +231,25 @@ class EventFileReaderFactory(Factory):
...
 
     # Product classes traits
     # Would be nice to have these automatically set...!
-    input_path = Unicode(get_dataset('gamma_test.simtel.gz'), allow_none=True,
-                         help='Path to the input file containing '
-                              'events.').tag(config=True)
-    max_events = Int(None, allow_none=True,
-                     help='Maximum number of events that will be read from'
-                          'the file').tag(config=True)
+    input_path = Unicode(
```
MaxNoe: I still disagree that this is a trait. I think this should be an argument to `__init__`. A `Tool` can have a trait `input_path` to make it available as config or commandline option but it should then pass it to `EventFileReader.__init__` or EventFileReaderFactory.__init__`. This is not configuration, this is an positional argument.

ctapipe/io/eventfilereader.py
watsonjj: I'd like to find an alternative approach to allow the EventFileReaderFactory to find all sub-classes of EventFileReader, but this is the only option I know so far.

I believe it would require a restructuring of how EventFileReaderFactory works, which is something we should do soon, but is a topic for a new PR.

ctapipe/io/eventfilereader.py
dneise: I agree with Max, but I also must admit, that I have problems to understand what should have traits and what not.
I have no experience with such a library and follow basically blindly.

However, I have the impression that such a library might be more suitable for projects, with a huge user base and awesome devs. Like ipython.

watsonjj: I don't really see your point against having it as a traitlet. Perhaps technically it is not a configuration, but it definitely makes using EventFileReader more efficient to use if the trait is defined in the class instead of the tool. In what situation are you going to use an EventFileReader without having to then create a `input_path` trait for the tool, every single time? It just seems like extra unnecessary work we put on the user...

If we do remove it as a traitlet, do you agree that the "default path" functionality could be kept in the `__init__`?

```python
if not input_path:
   input_path = get_dataset('gamma_test.simtel.gz')
```

watsonjj: Advantages of having it as a trait in the class:
* Disallows empty imput (`allow_none=False`)
* Automatically handles a default path
* User does not have to define a trait in the Tool each time they wish to pass a path from the command line, a functionality that majority of Tools will require.
* User can still override the traitlet configuration by passing the input_path in the class aruments (`reader = HessioFileReader(None, None, input_path="path/to/file")`)

Disadvantages of having it as a trait in the class:
* If there are multiple instances of the same reader in the script, every instance of the reader will have the same `input_path` set. However, a way around this would be to define the different input_path traitlets inside the tool and pass the paths individually to the readers in their initialisation arguments (see above). This is the exact same process that you would have to do if the trait didn't exist in the class.

Therefore I see removing the traitlet from the class as only removing functionality, and not achieving anything useful. I'd like to hear your arguments against having the trait in the class, and what it achieves.

dneise:
I think, we should merge this despite its potential issues. PRs are not supposed to deliver perfect code, but improve the existing code. 

The EventFileReader was a `Component` before this PR and afterwards, so if @MaxNoe  and me are correct with our assumption that `EventFileReader(Component)` should rather be `EventFileReader`... then this PR did not fix that issue. So what?

I believe one reason for the many discussions in the PR and its predecessor #610 was that it appeared to have some "finality". Guidelines are supposed to be valid for future devs. Code ... yeah .. code can be modified in future PRs if we have an idea how to improve it, but "Guidelines" ... that sounded so final.
And in a sense it is. EventFileReader defines an interface and all specific readers need to follow that interface. So if we make a mistake here and find out in 6 months from now, that people trying to implement their reader cannot follow this interface ... well then we might have to change the interface ... and then? Well we might have to touch all the other readers that were already implemented... and then? might have to touch a lot more code, since that code is using those readers.

But so what? How many readers are going to be there? For sure less than 10. 
How drastically will the interface change? For sure the readers will always be iterables. For sure they will always get some kind of url, when they are constructed. And fore sure they will keep spitting out Containers. 

So I think this interface is fine. 

Maybe we'll find out all readers should support direct event access by "id" and "index" and **add** these features to all specific readers which do not have them yet. This will probably not result in any downstream code modifications. Maybe all readers know their `__len__` in the future.

Don't get me wrong. I still think this PR is not perfect, but we can survive this.

I assume this project is hit much harder by PRs which are open for two weeks and cannot be merged because we fail to agree on stuff, than it is hit by possible future modifications.

watsonjj:
I fully agree @dneise.

I will just make the one more modification to change `input_path` to `input_url`.

ctapipe/io/eventfilereader.py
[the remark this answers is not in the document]
kosack: actually, the way it currently works, the "event" that is returned is a pointer, and there is only ever one instance, so if they do:

```py3
event = next(reader)
reader[5]

event is reader.current_event  
# always True, even if not assigned in the "for event in self.source" loop.
```
`event` will become the 5th event.   This was originally for speed considerations (so we don't have to re-alloc all the arrays and structures each time, rather just copy the data to them).  With this new interface and random-access, however, this may be confusing, so perhaps we should just go for a fully new event each time (with the associated speed reduction).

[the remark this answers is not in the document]
watsonjj: I think we should keep the functionality of event being a pointer, where the data it is pointed to is changed while you iterate through the generator. It is much more efficient, and the situation you have described is no longer a concern for `EventFileReader` as we have removed the `reader[5]` functionality.

Perhaps it should be included in the docstring that the returned event is a pointer, and if one wants to keep that particular event in memory, one should create a `copy.deepcopy` of it.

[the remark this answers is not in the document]
watsonjj: Actually in the current version of EventFileReader (where neither `event = next(reader)` or `reader[5]` is possible), this really isn't a concern.

```python
for event in reader:
   if event.count == 3:
      break
for event2 in reader:
   if event2.count == 5:
      break
assert(event.count == 3)
assert(event2.count == 5)
```

This works as it is a new generator that has been started in the second loop, so new memory has been assigned for the event. 

If we restore `event = next(reader)` functionality we should be careful, and include advice in the docstring, as a confusing situations could arise.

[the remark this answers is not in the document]
dneise: >  This was originally for speed considerations (so we don't have to re-alloc all the arrays and structures each time, rather just copy the data to them). With this new interface and random-access, however, this may be confusing, so perhaps we should just go for a fully new event each time (with the associated speed reduction).

How large was the performance impact here?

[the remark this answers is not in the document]
watsonjj: With some quick tests:
Current version (pointer): 80 events per second
Deepcopy the event on yield (dumb): 60 events per second
Create a new DataContainer for each new event: 78 events per second

The relative hit for HessioFileReader is low, as the reader itself is slow. However with TargetIOFileReader I can achieve file reading at around 1200 events per second if I remember correctly.

If it is true that the impact is very small on the efficiency, it certainly seems like something we could consider in the future. But I would advise to do some proper testing first.

dneise:
May I ask what the status here is?

watsonjj:
Ready to be merged I believe, however I don't have the authority to do so

Issue #603, opened 2017-12-11
problem with running  muon_reconstruction.py on NectarCAM muons 


FrancaCassol:
Hi, 
wenn I run muon_reconstruction.py on NectarCAM muons it crashes at the "chord_length" function when called by intersect_circle at line 126:

        if self.hole_radius > 0:
            hole_length = self.chord_length(
                self.hole_radius, r / self.hole_radius.value, angle
            )

Indeed in this case (r / self.hole_radius.value)=20=rho in chord_length and chord become negative, which is incompatible with the sqrt performed afterward:

       chord = 1 - (rho * rho * np.sin(phi) * np.sin(phi))
       if rho <= 1:
            chord = radius * (np.sqrt(chord) + rho * np.cos(phi))
        elif rho > 1:
            chord = 2. * radius * np.sqrt(chord)

Can somebody help on that?  I can pass a small simtel_array file if needed.

Thanks
Cheers

Franca

Sojourner92:
Hi Franca,
Could you please attach the simtel_array file you mentioned? I'm reasonably sure that the problem is on L173 of muon_reco_functions.py, a fixed (and wrong) mirror hole radius is being passed to the muon line integrator function which results in rho being 20 (when as its meant to be a fractional length it probably shouldn't be greater than 1).

FrancaCassol:
Hi Samuel,
yes! I was also guessing something like that, do you know which is the correct mirror hole radius? You can find the simtel_array file in https://www.cppm.in2p3.fr/~cassolb/data/
Thanks!
F.

Sojourner92:
Thanks for that. But yes, I'm having trouble finding the correct hole radius for the MSTs. In an ideal world I'd fix this problem for the three SST designs and the LSTs whilst I'm at it; does anyone happen to know what the correct hole radii for those are either?

FrancaCassol:
I can say 'yes' for what concerns NectarCAM muons: I could reconstruct them after the fix.
I plan to test muons simulated in other cameras in the coming weeks, but probably this issue can be considered close at the moment.

kosack:
Ok, thanks I'll close it. 

Soon, we can add some data files for testing muons... Alison had provided some, but I haven't yet added them to ctapipe-extra since it is growing big, and I think I'll just move test data to a file server in the future.

PR #615, opened 2018-01-15
Muonrecofix

Sojourner92:
Fix for issue #603. Most notable changes:

1. Added primary mirror hole radii to be taken in to consideration in the muon reconstruction for every telescope to muon_reco_functions.py (line 45), based on the latest MC parameter description files I could find.
2. Fixed an issue in muon_integrator.py arising due to non-astropy quantity objects being assigned astropy quantities.
3. Fixed an bug in the muon_reconstruction.py example script where detected muon rings were not being correctly logged.

As far as I can tell, this works properly now, but an expert in muon analysis (which I'm not) might want to double check the final results thrown out by muon_reconstruction.py to make sure they're sensible.


ctapipe/image/muon/muon_integrator.py
```diff
@@ -68,7 +68,8 @@ def __init__(self, mirror_radius, hole_radius, pixel_width=0.2,
         self.prediction = 0
         self.minlambda = 300.e-9
         self.maxlambda = 600.e-9
-        self.photemit = alpha * (self.minlambda**-1 - self.maxlambda**-1)  # 12165.45
+        self.photemit = 1.0 / 137.0 * (self.minlambda**-1 -
```
MaxNoe: Why replace a well-known constant with two magic numbers?

kosack: at least use `scipy.constants.alpha` (or `scipy.constants.inverse_alpha`)

kosack: or `astropy.constants.alpha` which is also there, and has references to where the value comes from automatically

ctapipe/image/muon/muon_integrator.py
```diff
@@ -68,7 +68,8 @@ def __init__(self, mirror_radius, hole_radius, pixel_width=0.2,
         self.prediction = 0
         self.minlambda = 300.e-9
```
kosack: if these are wavelengths, please require them to be astropy `Quantities`, so the unit is attached (e.g. define them as `300e-9 * u.m` or even `30 * u.angstrom`.

rlopezcoto:
Hi, 
I checked your fix but I got the error:

```sh
  File "muon_reconstruction.py", line 134, in <module>
    main()
  File "muon_reconstruction.py", line 91, in main
    muon_evt = analyze_muon_event(event)
  File "/home/rlopez/data_rlopez/cta/ctasoft/test/ctapipe/ctapipe/image/muon/muon_reco_functions.py", line 93, in analyze_muon_event
    dict_index = muon_cuts['Name'].index(str(teldes))
ValueError: 'SST-ASTRI:CHEC' is not in list
```

if I add 'SST-ASTRI:CHEC' to the dictionary entries in **muon_reco_functions.py** and use the same TailCuts, TotalPix... as in the 'SST-GCT:CHEC', that error is gone, but now I have an error related to the baseframe coordinates:

```sh
  File "muon_reconstruction.py", line 134, in <module>
    main()
  File "muon_reconstruction.py", line 91, in main
    muon_evt = analyze_muon_event(event)
  File "/home/rlopez/data_rlopez/cta/ctasoft/test/ctapipe/ctapipe/image/muon/muon_reco_functions.py", line 110, in analyze_muon_event
    rotation=geom.pix_rotation)
  File "/home/rlopez/data_rlopez/software/anaconda3/envs/cta-dev/lib/python3.6/site-packages/astropy/coordinates/baseframe.py", line 405, in __init__
    list(kwargs)))
TypeError: Coordinate frame got unexpected keywords: ['z']
```
do you get anything similar?

Sojourner92:
Hi,
No I haven't encountered anything similar (I think the script predates the astri/chec simulation studies), but if you post a simtelarray file I'll try to reproduce your error and see if I can fix it.

rlopezcoto:
Hi,
I tried with different files, but I get it just running the test file in the example:
ctapipe-extra/ctapipe_resources/gamma_test.simtel.gz

```sh
$python muon_reconstruction.py 
INFO: [SCRIPT] muon_reconstruction [__main__]
INFO:astropy:[SCRIPT] muon_reconstruction
Doing one-time thread-specific warning data initialization.
I/O block extended by 744748 to 1744748 bytes
Event Number 0
  File "muon_reconstruction.py", line 134, in <module>
    main()
  File "muon_reconstruction.py", line 91, in main
    muon_evt = analyze_muon_event(event)
  File "/home/rlopez/data_rlopez/cta/ctasoft/test/ctapipe/ctapipe/image/muon/muon_reco_functions.py", line 93, in analyze_muon_event
    dict_index = muon_cuts['Name'].index(str(teldes))
ValueError: 'SST-ASTRI:CHEC' is not in list

gzip: stdout: Broken pipe
```

as I said, to solve this is easy just by adding the 'SST-ASTRI:CHEC' dictionary entry, but then you have to solve mor problems and I wanted to know if you had encounter anything similar before digging deeper. Is it possible that you have an older gamma_test.simtel.gz file? I just checked it out from the repo.

Sojourner92:
Yeah, I wasn't running it with the example script, I was running the MuonDisplayerTool function with the data provided in #603 for NectarCAM and ipython. However, I don't get the same crash as you doing the same for the example file, albeit as I'm using the ASTRI parameters for the mirror radius etc as these will be more appropriate for an ASTRI:CHEC run. The parameters I'm using when I add the ASTRI: CHEC camera to the dictionary are:

Tailcuts=(5,7)
impact=(0.1,0.95)
Totalpix=2048
Minpix=164
AngPixelWidth=0.2
hole_rad=0.171 * u.m
cam_rad=2.86
sec_rad=1.8 * u.m
sct=True

I don't get a crash, but the tool isn't finding any muons, I think probably because the angular pixel width isn't correct. Does anyone know what value this should take for CHEC on ASTRI? There isn't a MC parameter description for that configuration.

rlopezcoto:
I run it with the parameters you wrote up there, but I still get the Coordinate frame error.
If the tool is not finding any muons, it might be because the example file is a gamma file.

rlopezcoto:
I found a problem that seems related to the rotation of the fitted circle, at least when plotted with the --display option (see images). 
screen shot 2018-01-22 at 19 43 00
screen shot 2018-01-22 at 19 49 25
screen shot 2018-01-22 at 19 49 47

It does not look like it is affecting the width measurement and maybe either the other muon measured parameters, so it is probably a rotation problem. If you want I can share with you a muon file that @AMWMitchell sent me.
Sorry, not sure if this bug is related to this PR or it was already there before, I'm reporting it here just in case, otherwise I can open an issue and investigate it.

AMWMitchell:
Not again - had this issue before. It looks like the definition of rotation angle has changed again. 
See plot_muon_event in muon_diagnostic_plots and try commentating out the if statement on LSTCam / NectarCam concerning the rotation angle. 
(might need to check how geom.pix_rotation is defined)

watsonjj:
Once this has been fixed, is there a test that can be created to check that this isn't broken again with a future update?

AMWMitchell:
Good point - although no easy tests spring to mind other than visually. 
Might have to do something like check that the pixels that pass the ring distance cut (pix distance to ring radius) in the nominal co-ordinate system also pass the same cut in the Camera co-ordinate system (using transformed ring centroid etc.)

rlopezcoto:
Commenting out lines 129 and 131 of muon_diagnostic_plots.py solved the issue. @Sojourner92, you may want to include it in your PR and we can maybe think later on a test to check this. 
screen shot 2018-01-23 at 08 19 02

kosack:
Is this ready to be merged, or is there more work being added?

Sojourner92:
Hi Karl,
No not yet, I've started having problems since I pulled the latest upstream master version. The camera dl1 image.charge_extractors modules are trying to import a function that no longer appears to exist on line 14 , 
from ctapipe.utils.neighbour_sum import get_sum_array.

Does anyone know anything about changes to this? It seems as though a python function has been replaced by a c++ wrapper, but I'm not sure how to fix the reference in the charge extractor file.

watsonjj:
That functionality has not be changed in the recent PRs. I suggest trying `python setup.py develop` to force a recompile of the c code

Sojourner92:
Thanks Jason, that appears to have fixed everything. Its now ready for merging then Karl.

PR #617, opened 2018-01-24
Refactor event source

kosack:
Following #613 , I made a few changes to fix some problems, mainly:

* removed the code for the old `hessio_event_source` and replaced it with a wrapper that works the same, but uses internally an HessioFileReader (so no more duplicated code)
* fixed some examples
* made `allowed_tels` a true traitlet config variable in the base class (and fixed the `HessioEventReader` so it works correctly for a null set)
* changed all (or most) tests that loaded a single event from gamma_test.simtel.gz to use a pytest *fixture* called `test_event` defined in `conftest.py`

I would like to also do the following (comments welcome):

* [x] According to the programming standards naming conventions, `EventFileReader` subclasses should be called e.g. `HESSIOEventFileReader`, not `HessioFileReader`. Since that's a bit long, how about we now just rename the base class: `EventFileReader` -> `EventSource` and then we have similar to what we had before: `HESSIOEventSource`, `TargetIOEventSource`, etc.?  that way we can also support e.g. the `ToyModelEventSource` as we had before?

* [x] make a generic helper function like `hessio_event_source()` called `event_source()` that internally constructs an `EventSource` using the factory class and returns a generator, to emulateold behavior and to have simple examples be shorter (it's use is only recommended for examples, so we can just do a quick:

```py3
for event in event_source("something.simtel.gz"):
     print(event.r0.tels_with_data)
```

* [ ] fix all examples and notebooks to use new scheme

I'll leave the following for a future pull request:

* [ ] I would like to also start to remove the `inst` container from the event, and have it be an attribute of the reader `reader.inst`.  Have to think about the right way since currently it is only filled on the first read event, but that could be changed. That way the `event` that is looped over is truly something that is related to events, not to the instrument config as well.  That means explicitly passing the `SubArrayDescription` in addition to the event to any algorithm that needs it (some already work that way).


kosack:
There still seems to be a problem with the Factory class: I removed input_url from `HessioFileReader`, since it is already in the base class. This works fine when using it independently of the base class, but the `test_event_file_reader_factory()` test fails, with the error that the `input_url` is empty (even though it is not). I suspect this has something to do with how the traits are copied in the Factory base class, but haven't figured it out.

watsonjj:
I agree with all the proposed changes in your first post.

I'm not sure about removing `input_url` from `HessioFileReader`. It then removes the default file for each reader.

I can have a look at altering the factory a little such that this problem is fixed. The reason it isn't working is because the config isn't being passed correctly. I will create a fix in a PR.

dneise:
>  I would like to also start to remove the `inst` container from the event, and have it be an attribute of the reader `reader.inst`.

Isn't it possible that analysis functions get only `event`s and have no access to the `reader`? Then how will these analysis functions get access to the `inst`rument description?

ctapipe/io/hessiofilereader.py
```diff
@@ -69,8 +62,6 @@ def _generator(self):
             # it after each yield
             counter = 0
             eventstream = file.move_to_next_event()
-            if self.allowed_tels is not None:
-                self.allowed_tels = set(self.allowed_tels)
```
dneise: why is this needed `self.allowed_tels = set(self.allowed_tels)`? Does the traitlets not take care that this is a set?

ctapipe/io/hessio.py
```diff
@@ -91,178 +43,12 @@ def hessio_event_source(url, max_events=None, allowed_tels=None,
...
-            subarray.positions[tel_id] = tel_pos
+    reader = HessioFileReader(None, None,
+                              input_url=url, **params )
 
-        except HessioGeneralError:
-            pass
 
-    return subarray
\ No newline at end of file
+    return (x for x in reader)
```
dneise: Why not return `reader`? It is also iterable and will emit `x` when iterated over.

examples/display_summed_images.py
```diff
@@ -37,58 +37,65 @@ class ImageSumDisplayerTool(Tool):
...
     def setup(self):
         # load up the telescope types table (need to first open a file, a bit of
         # a hack until a proper insturment module exists) and select only the
         # telescopes with the same camera type
-        data = next(hessio_event_source(self.infile, max_events=1))
-        data.inst.subarray.info(printer=self.log.info)
-        camtypes = data.inst.subarray.to_table().group_by('camera_type')
+
+        self.reader = HessioFileReader(None, None, input_url=self.infile,
+                                       max_events=self.max_events)
```
dneise: Indenting like this:
```python
self.reader = HessioFileReader(
    None, 
    None, 
    input_url=self.infile,
    max_events=self.max_events,
)
```
instead of this:
```python
self.reader = HessioFileReader(None, None, input_url=self.infile,
                               max_events=self.max_events)
```

Has some advantages:
 * stability over refactoring, when renaming `HessioFileReader` to `HessioFile` or `HessioEventSource` in the future, the 2nd line does not need to be re-indented.
 * the max-line length in pep8 is 80 chars, but most people like to read between 50...60 chars, hence the width of most magazine or newspaper columns. 
 * the latter is *discouraging* the use of explanatory keyword arguments, while the former can easily transition into:
self.reader = HessioFileReader(
    config=None, 
    tool=None, 
    input_url=self.infile,
    max_events=self.max_events,
)

Which is even better readable.

 * The latter way of indenting optically  "groups" the first 3 parameters, but they are no group.

examples/display_summed_images.py
```diff
@@ -37,58 +37,65 @@ class ImageSumDisplayerTool(Tool):
...
             if geom is None:
-                geom = data.inst.subarray.tel[self._base_tel].camera
+                geom = event.inst.subarray.tel[self._base_tel].camera
                 imsum = np.zeros(shape=geom.pix_x.shape, dtype=np.float)
                 disp = CameraDisplay(geom, title=geom.cam_id)
                 disp.add_colorbar()
                 disp.cmap = 'viridis'
 
-            if len(data.dl0.tels_with_data) <= 2:
+            if len(event.dl0.tels_with_data) <= 2:
```
dneise: is "2" a magic number here?

examples/display_summed_images.py
```diff
@@ -37,58 +37,65 @@ class ImageSumDisplayerTool(Tool):
...
                 continue
 
             imsum[:] = 0
-            for telid in data.dl0.tels_with_data:
-                imsum += data.dl1.tel[telid].image[0]
+            for telid in event.dl0.tels_with_data:
+                imsum += event.dl1.tel[telid].image[0]
 
             self.log.info("event={} ntels={} energy={}" \
```
dneise: According to pep8 line continuation with backslash is discouraged:

this could be reformated as:
```python
self.log.info("event={} ntels={} energy={}".format(
    event.r0.event_id,
    len(event.dl0.tels_with_data),
    event.mc.energy))
```

dneise:
I like this PR a lot. Removing this much code can only be great.

ctapipe/io/hessio.py
```diff
@@ -91,178 +43,12 @@ def hessio_event_source(url, max_events=None, allowed_tels=None,
     use_event_id : bool
         If True ,'requested_event' now seeks for a particular event id instead
         of index
-    """
-
-    warnings.warn("hessio_event_source has been deprecated. "
```
watsonjj: I would keep this warning, we do want this function to be transitioned away from

ctapipe/io/hessiofilereader.py
```diff
@@ -20,10 +19,6 @@ class HessioFileReader(EventFileReader):
     """
     _count = 0
 
-    input_url = Unicode(
```
watsonjj: Removing this removes the ability to set a default path for the reader. I think that is useful functionality to keep.

ctapipe/io/hessio.py
kosack: `next(source)` doesn't work in that case, which is the old behavior used in some scripts.  Eventually I want to just deprecate `hessio_event_source`, but it will take some time to update scripts. In the future, everything should use an EventFileReader

ctapipe/io/hessiofilereader.py
kosack: agreed, this is redundant now - it assures it's a set

ctapipe/io/hessio.py
kosack: I'll add an `@deprecated` decorator, as with other functions (actually that reminds me that we should define a ctapipe deprecated decorator, instead of using the astropy one, since the message is confusing when it is called an `AstropyDeprecationWarning`

ctapipe/io/hessio.py
dneise: Oh `next` does not work on a reader? I did not realize that. Can we repair that?

ctapipe/io/hessiofilereader.py
kosack: it was just confusing as it creates another attribute called input_url that shadows the parent class attribute, which is generally bad practice. We could just have a default value attribute that gets filled in if input_url is blank.  I don't really see having a default is very useful, though. The input shoudl be explicitly specified.

watsonjj: Okay, lets discard defaults

kosack:
I still have a strange factory problem: when doing this:

```py3
url = get_dataset("gamma_test.simtel.gz")
factory = EventFileReaderFactory(None,None, input_url=url)
print(factory.input_url) # prints correctly
factory.produce(None,None) #fails, saying input_url is empty
```

```
~/Projects/CTA/Working/ctapipe/ctapipe/io/eventfilereader.py in __init__(self, config, tool, **kwargs)
    124         if not exists(self.input_url):
    125             raise FileNotFoundError("file path does not exist: '{}'"
--> 126                                     .format(self.input_url))
    127         self.log.info("INPUT PATH = {}".format(self.input_url))
    128 

FileNotFoundError: file path does not exist: ''
In [1]: %debug
> /Users/kosack/Projects/CTA/Working/ctapipe/ctapipe/io/eventfilereader.py(126)__init__()
    124         if not exists(self.input_url):
    125             raise FileNotFoundError("file path does not exist: '{}'"
--> 126                                     .format(self.input_url))
    127         self.log.info("INPUT PATH = {}".format(self.input_url))
    128 

ipdb> self.input_url
''
```

Any idea why `input_url` isn't getting passed to the product? In this case, I don't pass a `config`, so perhaps that's the problem, but it should work without one, I would guess?

ctapipe/io/hessio.py
watsonjj: As was discussed in #613, EventFileReader was changed to be an [iterable](https://docs.python.org/3/glossary.html#term-iterable), not an [iterator](https://docs.python.org/3/glossary.html#term-iterator) (which has the `__next__` method). 

If you want to add this functionality to EventFileReader I would suggest opening a PR so we can discuss the implementation. Currently with EventFileReader we can do the following:
```
for event in reader:
   if event.count == 3:
      break
for event in reader:
   if event.count == 2:
      break
``` 
Where the it starts from the beginning again instead of the last position, which I think is more intuitive (its like looping through a list). If we include a `__next__` method I would argue we do it in a way that does not break this functionality.

watsonjj:
Because the correct way to use the Factory is:
```
url = get_dataset("gamma_test.simtel.gz")
reader = EventFileReaderFactory.produce(None,None, input_url=url)
print(factory.input_url) # prints correctly
```
as produce is a `classmethod`

dneise:
As `Factory.produce` is a class method ... and `produce` is maybe the single most important method of a factory, let me ask: Is there any reason to ever create an instance of a Factory?

If not ... why is a Factory a class?

dneise:
the current master creates 2 warnings for the factory tests:
```
(cta) dneise@lair:~/cta/ctapipe$ pytest ctapipe/core/tests/test_factory.py
================================================= test session starts =================================================
platform linux -- Python 3.6.4, pytest-3.3.2, py-1.5.2, pluggy-0.6.0 -- /home/dneise/anaconda3/envs/cta/bin/python
cachedir: .cache
rootdir: /home/dneise/cta/ctapipe, inifile: setup.cfg
plugins: cov-2.5.1, hypothesis-3.38.5
collected 5 items                                                                                                     

ctapipe/core/tests/test_factory.py::test_factory PASSED                                                         [ 20%]
ctapipe/core/tests/test_factory.py::test_factory_produce PASSED                                                 [ 40%]
ctapipe/core/tests/test_factory.py::test_false_product_name PASSED                                              [ 60%]
ctapipe/core/tests/test_factory.py::test_expected_args PASSED                                                   [ 80%]
ctapipe/core/tests/test_factory.py::test_expected_config PASSED                                                 [100%]

================================================== warnings summary ===================================================
ctapipe/core/tests/test_factory.py::test_expected_config
  /home/dneise/anaconda3/envs/cta/lib/python3.6/site-packages/traitlets/config/configurable.py:84: UserWarning: Config option `extra` not recognized by `ExampleFactory`.
    self.config = config
  /home/dneise/anaconda3/envs/cta/lib/python3.6/site-packages/traitlets/config/configurable.py:84: UserWarning: Config option `extra` not recognized by `ExampleComponent2`.
    self.config = config

-- Docs: http://doc.pytest.org/en/latest/warnings.html
======================================== 5 passed, 2 warnings in 0.46 seconds =========================================
```

watsonjj:
> If not ... why is a Factory a class?

Because then other classes like EventFileReaderFactory can inherit the Factory methods and functionality.

I am currently thinking about alternatives to the current Factory implementation. Karl suggested a metaclass approach.

If you have actual suggestions on how to achieve this Factory functionality (there are other Factories in ctapipe, like ChargeExtractorFactory) in some other way, I invite you to submit them...

kosack:
Ah, ok, I was confused since the factory class also had a input_url attribute. That makes more sense.

Nevertheless, I'm still getting some strange behavior: when I run the tests, the `EventFileReaderFactory` has an empty list of subclasses, even though when I contruct one manually it lists HessioFileReader.  I haven't looked into how it detects the subclasses, but if it is dependent on the order in which the imports are done, it could be a problem, since that isn't guaranteed.  I think we should switch to using a metaclass registry so that instances are automatically recorded when the file is read. I'll see if I can manage to do that...

kosack:
Seems to be a circular dependency problem: 
`hessiofilereader.py` imports `eventfilereader.py`, but `eventfilereader.py` also impots `hessiofilereader.py` before defining the factory class. I'm not sure how that even worked before, but now that I've cleaned up some import statments, it seems to have broken.

watsonjj:
We can avoid this circular dependency by moving EventSourceFactory to its own module. I will push a commit to this PR.

kosack:
There are still some issues with the EventSourceFactory -- depending on the order in which I import the subclasses, I either get no readers, or an error like `TypeError: can't pickle _thread.RLock objects` during the line `parent = deepcopy(self.__dict__['_trait_values']['parent'])`

Some of this may get fixed in #621 , so I may ignore them for now (the tests work, but most of the examples break with this error)

PR #620, opened 2018-01-29
fix broken docs

kosack:
latest sphinx had problems with using `quantity_input` decorator in constructor, so removed it (wasn't really needed)


MaxNoe:
Did it really have problems with `@u.quantity_input` or with the type decorators used?

So, does maybe `@u.quantity_input(focal_length=u.m**2)` work?

PR #622, opened 2018-01-29
fix bug caused by deepcopy in Factory constructor 

kosack:
changed deepcopy to copy (deepcopy was not needed and had problem with `TypeError: can't pickle _thread.RLock objects`)


kosack:
for some reason, this only caused problems with the example code, all tests worked, which means we probably are missing a test...

PR #625, opened 2018-01-29
Removed large binary files from examples/notebooks

kosack:
It seems some past PR included some large HDF5 files in the notebooks dir that were accidentally committed. This broke most of the code checker tools, since the repo is now too large. 

We may need to find a way to prune them out of the history as well, otherwise the repo size exceeds limits for landscape.io and others.  

No binary files should be added to this repo, only code. Binaries can be distributed separately in packages or a web server.


MaxNoe:
Maybe add `*h5`, `*.hdf` and `*.hdf5` to the `.gitignore`? And also stuff like `*.gz`, `*.fits`?

PR #629, opened 2018-01-30
Improve component constructor

kosack:
Component classes now have proper default values set for *config* and *tool* properties, so there is no longer a need to specify `comp = SomeComponent(None,None)`, and  now`comp = SomeComponent()` works. 

Updated all examples and uses to reflect this, which is much cleaner looking than before.  `Components` that are inside `Tool` base classes can still pass around their config info using the named parameters. 

example:
```py3

source = EventSourceFactory.produce(input_url="gammas.simtel.gz")
calib = CameraCalibrator()

for event in source:
    calib.calibrate(event)

...
```


dneise:
Thank you so much for this.

kosack:
closed this one due to a merge history problem.  Will re-open another PR with the same info

Issue #627, opened 2018-01-29
Redefinition of Instrument broke many compatibilities


rlopezcoto:
I guess you are aware of this since these changes are recent, but the redefinition of the instrument container in: ctapipe/ctapipe/io/containers.py
broke several compatibilities (apart from some of the examples already fixed) with examples and other files, e.g:

- examples/muon_reconstruction.py
- examples/plot_theta_square.py
- ctapipe/image/muon/muon_diagnostic_plots.py
- ...

is the plan to have a SubArray container similar to the previous Instrument one? How can we access now properties like optical_foclen, pixel_pos, ... that were defined in the Instrument Container?

Thanks!

PR #630, opened 2018-01-30
Instfix

rlopezcoto:
Related to the Issue reported in #627, apparently #623 did not completely solve the issue with the examples and other files due to the change in the hierarchy of classes. 

I changed the files:
- examples/plot_theta_square.py
- ctapipe/image/muon/muon_diagnostic_plots.py

to make them compatible with the new hierarchy definition.


kosack:
there seems to be a repo issue... I'm trying to solve it.  I think those files are already fixed in another PR though.

rlopezcoto:
ok, sorry it seems that something got messed up before doing the PR... I just made changes to the two files mentioned, that is why I closed the PR

PR #633, opened 2018-01-30
R1 Calibrator Changes

watsonjj:
Some cleanup of the R1 Calibrator code following the recent pull requests:

- Created NullR1Calibrator, which serves as a better default than HessioR1Calibrator, as other file formats would not want that used by default.
- Allowed the passing of eventsource to CameraR1CalibratorFactory. The Factory can then select the correct CameraR1Calibrator to use based on metadata or instrument information, unless overrided by the user.
- Allowed more customisation on the generic CameraCalibrator class
- Moved calib_scale into the HessioR1Calibrator
- Renamed HessioR1Calibrator to HESSIOR1Calibrator

### DIFF
--- a/ctapipe/calib/camera/calibrator.py
+++ b/ctapipe/calib/camera/calibrator.py
@@ -42,7 +42,12 @@ class CameraCalibrator(Component):
         ))
 
     """
-    def __init__(self, config=None, tool=None, origin='hessio', **kwargs):
+    def __init__(self, config=None, tool=None,
+                 r1_product=None,
+                 extractor_product=None,
+                 cleaner_product=None,
+                 eventsource=None,
+                 **kwargs):
         """
         Parameters
         ----------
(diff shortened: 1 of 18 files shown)


kosack:
This looks nice.  Is it also necessary to update any examples and notebooks where we had previously just constructed a `CameraCalibrator()` to do the calibration, and replace it with a `R1CameraCalibratorFactory`? or will the old way still work?

kosack:
Just a comment (not a problem with this PR): I think we'll need a better way to select which calibration method to use than basing it on the file format, since eventually there will be one format, and just a different method per telescope/camera type. So we could just choose based on the telescope or camera identifier in the SubarrayDescription or something like that.  Just need to keep that in mind for now.

watsonjj:
> This looks nice. Is it also necessary to update any examples and notebooks where we had previously just constructed a CameraCalibrator() to do the calibration, and replace it with a R1CameraCalibratorFactory? or will the old way still work?

As the default R1Calibrator has been changed from HESSIOR1Calibrator to NullR1Calibrator, the examples will need to be changed to be explicit... I will do that now.

> So we could just choose based on the telescope or camera identifier in the SubarrayDescription or something like that. Just need to keep that in mind for now.

Absolutely, in fact the passing of the eventsource was done such that the Factory can either use metadata or SubArrayDescription to choose the correct R1Calibrator. 

As you can see the current implementation uses the is_simulation entry from metadata to know if HESSIOFileReader should be used.

But in the future when the places for different items is more figured out, we could change this to use just the SubArrayDescription.

kosack:
might need to pull in or rebase on the latest master, since #635 changed some of the examples already, and they now conflict with your changes.

PR #634, opened 2018-01-30
Eventsource docs

watsonjj:
An addition to the io documentation concerning the creation of a new EventSource.

This short chapter is intended to replace the documentation changes in #610.


watsonjj:
@dneise You don't appear in the reviewer list, but I would also appreciate your review.

Issue #608, opened 2017-12-13
Refactor of name for container of waveforms


watsonjj:
I would like to propose a refactor of the name for the container of waveforms. Currently we have:

event.r0.tel[1].adc_samples
event.r1.tel[1].pe_samples
event.dl0.tel[1].pe_samples
event.dl1.tel[1].image

It is potentially misleading to call them "adc_samples" and "pe_samples" as these assume units that might not be the case. Therefore I would propose to rename them to "samples". This would also make the access of waveforms more consistent across data levels.

This would be a huge refactor that would presumably affect every ctapipe user's code. Therefore it should probably be delayed until we intend a huge refactor of the code. This could be the same time we remove the "tel" containers for the R0 -> DL1 containers (I presume this is an intended change once we move to single telescopes per file).

calispac:
In this case should : 
1. ```pe_samples.shape == (n_channels, n_pixels, n_samples)``` ? i.e. some kind of waveform where the max is the reconstructed p.e. charge
2. `pe_samples.shape == (n_channels, n_pixels)` ? 

In case of 2. I think it is more intuitive to have them named `pe_image`. Or not ?

watsonjj:
event.r0.tel[1].adc_samples --> event.r0.tel[1].samples
event.r1.tel[1].pe_samples --> event.r1.tel[1].samples
event.dl0.tel[1].pe_samples --> event.dl0.tel[1].samples

All of these are waveforms, so have shape (n_channels, n_pixels, n_samples)

The units depend on the low-level R1 calibration each camera applies on them. I'm not sure if there is a requirement to have the waveforms in p.e. at the dl0 level (they will at least be something proportional to p.e. i presume, such that the online analysis can use them)

calispac:
Ok I did not get you point first.
 It seems more user friendly to have the "main data" being named similarly across data levels. As the data level indicates some transformation of `samples`.

kosack:
I'm happy with the suggested names, though there will need to be some small change to support the fact that in the final DL1 we have only samples for a subset and images for the rest (so we need both DL0.image and DL0.samples. 

 In fact, I'm working on standardizing the definitions for all of CTA.  In my current doc I have "image" = a single image for a full camera, and "traces" = the time-evolution of a subset of pixels, but "samples" is perhaps better (any preference or better name?).

kosack:
> In case of 2. I think it is more intuitive to have them named pe_image. Or not ?

[...]

PR #638, opened 2018-02-01
Refactor containers

kosack:
This is in response to #608 

* Change `pe_samples`  and `adc_samples` to  `waveform` everywhere
* change `pe_sums` and `adc_sums` to `image` everwhere
* add some (so far unfilled) fields to `{R0,R1,Dl0}CameraContainer` for the camera trigger time and type
* updated docstrings to reflect where gain selection should appear (though it doesn't actually happen yet), and that the DL0 waveform may be a masked array, due to zero-suppression.


kosack:
@watsonjj does this seem ok? It will break other readers like TargetIO, which will need to be updated as well.

watsonjj:
Yep this is fine, I have an older branch of ctapipe that I use for TargetIO at the moment, and soon I will create a PR to add the reader to ctapipe this will incorporate these updates.

PR #640, opened 2018-02-12
Muondev

rlopezcoto:
Implementation of the angular containment within the camera.

Included in /image/muon/features.py I implemented a function that calculates what is the containment of a muon ring inside the camera (assuming a circular camera centered in 0,0).
I added the ang_containment field to the MuonRingParameter container in io/containers.py and included the calculation of ang_containment in the muon_reco_functions.py

### DIFF
--- a/ctapipe/image/muon/features.py
+++ b/ctapipe/image/muon/features.py
@@ -124,11 +124,27 @@ def ang_containment(
         cring_y,
         ):
 
-    '''
-    Estimate containment of a ring inside the camera
+    '''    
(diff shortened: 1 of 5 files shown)


kosack:
Also, the function name should not be abbreviated for clarity, perhaps `ring_containment()` or `ring_angular_containment()` so it's obvious what it does without reading the docstring.

rlopezcoto:
Yes, thanks for the quick review. I added a test unit and changed the function name to ring_containment

PR #641, opened 2018-02-12
Improve code formatting

kosack:
fix a ton of warnings about pep8 problems to attempt to improve score on codacy and highlight only real problems. 

This is somewhat annoying as it will look like large code changes, but we should probably strive to have few to no PEP8 warnings for the first release...  Any suggestions?


dneise:
Just out of curiosity: Which tool do you use for pep8-ing your code and do your use a custom config file for this? I think, all developers should use the same config.

In pep8 there are sometimes several alternatives allowed. For example when it comes to indentation for continuation lines:

I for example think this: 
```python
            color_list = [
                (0 / cmaxmin, 'darkblue'),
                (np.abs(min(signals)) / cmaxmin, 'black'),
                (np.abs(min(signals)) / cmaxmin, 'black'),
            ]
```
is better than this:
```python
            color_list = [(0 / cmaxmin, 'darkblue'),
                          (np.abs(min(signals)) / cmaxmin, 'black'),
                          (np.abs(min(signals)) / cmaxmin, 'black')]
```

since:
  * it is stable under refactoring of the name of "color_list"
  * during testing/playing around, one can easily comment out every of these lines.
  * one has more line width available for the list contents, so one is not so easily forced to choose a shorter name for the list itself
  * (weak: my editor has no feature to make the upper version for me)

kosack:
I've been just using autopep8. I also prefer the second indentation shown above - do you know if there is a way to tell autopep8 (or some other tool) to perfer that style? So far it seems autopep8 keeps the style of what was there before, so it doesn't make large changes to the indentation other than to align things. 

An yes, we should avoid backslash indentation - it's even marked as bad code by codacy.

dneise:
No, sorry. I never use autopep8. I rather tidy up un-pepped code manually, this way I give it another review and maybe I can not only pep8 the code, but also refactor it .... you know, break out a few functions ... rename a few variables.

So I've got no idea how to configure autopep.

kosack:
*yapf* seems to do nice reformatting of long lines as long as you specity it in the stylesheet or `--style` option:

```sh
yapf --diff --style='{dedent_closing_brackets: true}' prepare_event.py
```

```diff
-            yield PreparedEvent(event=event, hillas_dict=hillas_dict, n_tels=n_tels,
-                                tot_signal=tot_signal, max_signals=max_signals,
-                                pos_fit=pos_fit, dir_fit=dir_fit, h_max=h_max,
-                                err_est_pos=err_est_pos, err_est_dir=err_est_dir
-                                )
+            yield PreparedEvent(
+                event=event,
+                hillas_dict=hillas_dict,
+                n_tels=n_tels,
+                tot_signal=tot_signal,
+                max_signals=max_signals,
+                pos_fit=pos_fit,
+                dir_fit=dir_fit,
+                h_max=h_max,
+                err_est_pos=err_est_pos,
+                err_est_dir=err_est_dir
+            )
```

dneise:
looks perfect

Issue #644, opened 2018-02-15
Bug applying cuts for muons?


rlopezcoto:
Hi,

I found what seems to be a bug, but wanted to be sure before correcting it.  In l.152 of `muon_reco_functions.py`, there are the two following conditions:

        if(np.sum(pix_im > tailcuts[0]) > 0.1 * minpix
           and np.sum(pix_im) > minpix 

both of them are relating a sum over the content of `pix_im`, that has the pixel containments (pe, I guess). Then it is comparing it to something proportional to `minpix` that is a minimum number of pixels defined in a dictionary. My guess is that the comparison that wanted to be done is something like:

        if(len(pix_im[pix_im > tailcuts[0]]) > 0.1 * minpix
           and len(pix_im[pix_im > 0]) > minpix 

@AMWMitchell, is it alright if I correct it?

dneise:
Hey Ruben (no idea of the muon code) but I wanted to say one can quote code directly from the source file like this:
https://github.com/cta-observatory/ctapipe/blob/39cf9af6df8ffed682982aa40a0581d722db76a3/ctapipe/image/muon/muon_reco_functions.py#L152-L161

This way interested noobs like me can quicker find the entire file by following the link and study the context in detail... if they want.

Cheers.

rlopezcoto:
thanks for the tip, @dneise!

AMWMitchell:
Yes, well spotted - this should be comparing the number of pixels rather than the sum of the charge. Please do correct this, and check if it affects the number passing cuts (shouldn't make a large difference, but there might be fewer events). 
Thanks

rlopezcoto:
Solved in #645

PR #645, opened 2018-02-15
Muonfitdev

rlopezcoto:
Fix bug on the minimum number of pixels in a muon image reported in #644


ctapipe/image/muon/muon_reco_functions.py
```diff
@@ -153,8 +153,10 @@ def analyze_muon_event(event, params=None, geom_dict=None):
         # diameter of 0.11, all cameras are perfectly circular   cam_rad =
         # np.sqrt(numpix*0.11/(2.*np.pi))
 
-        if(np.sum(pix_im > tailcuts[0]) > 0.1 * minpix
-           and np.sum(pix_im) > minpix
+        if(len(pix_im[pix_im > tailcuts[0]]) > 0.1 * minpix
+           # > 0.1 * minpix above threshold
+           and len(pix_im[pix_im > 0]) > minpix
+           # Npix composing the ring > minpix
```
dneise: I think these comments show that this part of the code could profit from some documentation.
I personally like a refactoring process sometimes called: "breaking out functions" better than putting comments.

The reason is, that breaking out functions, often leads to code, which is better testable. 

In this case, instead of a comment, one could create a function e.g. similar to this one:
```python
def Npix_composing_ring(image):
    return np.sum(image > 0)
#and
def Npix_above_threshold(image, threshold=tailcuts[0]):
    return np.sum(image > threshold)

#then the if-clause would read:
if Npix_above_threshold(pix_im) > 0.1 * minpix and Npix_composing_ring(pix_im) > minpix:
    # do stuff
```

Comments tend to become "stale" and "smell ugly" after a time, especially when somebody changes the `0.1` in the code to `0.15` and forgets to update the comment.

Also with these new functions, you might want to go ahead and implement a little unit test, to make sure they really work as expected...

rlopezcoto: Thanks for the review.
Ok, I'll implement look into the proposed changes.

Just added a test module for ring_completeness and filled it in muon_reco_functions, please do not merge yet

rlopezcoto: I implemented the two proposed functions in `image/muon/features.py` (although I do not know if another location would be better for them...) and two test modules to test them.
I also modified the code in `muon_reco_functions.py` to call the functions and do not need to include comments.

ctapipe/image/muon/tests/test_features.py
```diff
@@ -14,6 +19,48 @@ def test_ring_containment():
...
+
+    assert((npix >= 0) and (npix <= len_array))
+
+
 if __name__ == '__main__':
 
     test_ring_containment()
+    test_ring_completeness()
+    test_npix_above_threshold()
+    test_npix_composing_ring()
```
dneise: I see these `__main__` blocks in python-test files here and in digicampipe as well and wonder why people are writing these... 

If one wants to execute only part of the tests, one can do:

    pytest ctapipe/image/muon/tests/test_features.py

for example ...

ctapipe/image/muon/tests/test_features.py
```diff
@@ -14,6 +19,48 @@ def test_ring_containment():
...
+    pe = np.random.uniform(0, 100, len(x))
+    ring_radius = 1. * u.m
+
+    ring_center_x = 0. * u.m
+    ring_center_y = 0. * u.m
+
+    ring_comp = ring_completeness(
+        x, y, pe, ring_radius,
+        ring_center_x, ring_center_y,
+        30, 30)
```
dneise: Sometimes functions, which need more than 3 or 4 parameters (apart from plotting functions, which are known to take a virtually unlimited amount of parameters :-D) are a hint, that there is a little class waiting to be implemented.

I could imagine a little `Ring` class (just a good old C-like-struct in python maybe a `collections.namedtuple`) might help here.

rlopezcoto:
Thanks for the review @dneise. Regarding the suggestion of a `Ring` class, could you maybe open an Issue on that? 
The reported bug was fixed and now the code is passing all the tests, so you can merge it anytime.

dneise:
Oh .. these comments where not meant as a review. I am fine with it and would not open an issue for this. It is a question of taste, I assume. And a question of style...

Sorry, I'll try to make a valid review in 5 ..15min

ctapipe/image/muon/muon_reco_functions.py
```diff
@@ -193,6 +196,15 @@ def analyze_muon_event(event):
                 muonintensityoutput.event_id = event.dl0.event_id
                 muonintensityoutput.mask = dist_mask
 
+                muonintensityoutput.ring_completeness = ring_completeness(
+                    x[pix_im is not False], y[pix_im is not False],
+                    pix_im[pix_im is not False],
```
dneise: `pix_im` is an "array with pixel content, usually pe" right? Why is this compared to a boolean here? 
Also: I think `not False` is  `True`.

if you are testing for `!= 0` here maybe this can be said explicitely.

Also:
```python
selection = pix_im is not False
x_selection = x[selection]
y_selection = y[selection]
pix_im_selection = pix_im[selection]
```

I looks to me like (x, y and pix_im) belong together, just as if they would form a unit, maybe it would be nicer to collect these in a structure array, give this array a meaningful name and give also `selection` a meaningful name. 
I just called it `selection` since I have no idea what is going on.

ctapipe/image/muon/tests/test_features.py
```diff
@@ -1,5 +1,10 @@
 import astropy.units as u
+import numpy as np
+import math as mt
```
dneise: is `import math as mt` common in the community? These `as something` imports effectively introduce new "global variables". And we all know globalb variables are bad ... but when a python-scientists reads `np.something` she just *knows* that np is numpy. So it is fine. 

Same goes for `plt`, `pd` and `u`... `mt` is new for me.

dneise: Unfortunately I have never found a list online, which lists these canonical module aliases ... So I don't know if `mt` is canonical.

ctapipe/image/muon/tests/test_features.py
```diff
@@ -14,6 +19,48 @@ def test_ring_containment():
...
+
+
+def test_npix_above_threshold():
+    len_array = 100
+    thr = 5.
+    pix = np.random.uniform(0, 100, len_array)
+
+    npix = npix_above_threshold(pix, thr)
+
+    assert((npix >= 0) and (npix <= len_array))
```
dneise: I would not use randomness for tests ... So I'd propose
```
pix = np.ones(len_array)
pix[[3, 7, 77]] = thr + 1

npix = npix_above_threshold(pix, thr)

assert npix == 3
```

You might want to test (i.e. document) corner cases: what if pix is an empty array? What result would you expect? zero? or some exception?

dneise:
I think my review is worthless :-(

ctapipe/image/muon/muon_reco_functions.py
rlopezcoto: x and y are the coordinates of all the pixels in the camera, while pix_im is the masked array of pixels inside a ring. Since the array is masked, I wanted to extract only the non-zero values and used `!=`, but pep8 complained. I changed it to `np.nonzero`

Apparently, `is not False` is not the same as `is True` :D

ctapipe/image/muon/tests/test_features.py
rlopezcoto: Not sure if `mt` is canonical, I reverted it to `math`

ctapipe/image/muon/tests/test_features.py
rlopezcoto: any particular reason not to use randomness here?
the assertion is very well defined

ctapipe/image/muon/features.py
```diff
@@ -154,3 +154,41 @@ def ring_containment(
...
+    thr: float
+        threshold for the pixels to be counted
+
+    Returns
+    ------
+    npix_above_threshold: float
+        Number of pixels above threshold
+    """
+
+    return len(pix[pix > thr])
```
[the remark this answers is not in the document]
rlopezcoto: yes, thanks

ctapipe/image/muon/tests/test_features.py
dneise: I think tests should be deterministic if they can. What if a user reports this test to fail on their machine? Hard to help this user.

Maybe it already helps when one sets the random seed, but I did not think about this in the first place.

ctapipe/image/muon/muon_reco_functions.py
dneise: > Apparently, is not False is not the same as is True :D

Well it is only the same, when comparing booleans. The usage for "True"  and "False" implies to the reader (at least I think so) that both operands of the binary `is` are booleans, which was not the case.

ctapipe/image/muon/tests/test_features.py
kosack: yes, we should remove these main blocks, mainly because the code coverage tools get confused by them and expect you to write unit tests for your unit tests if they are executable in this way.  The correct way to run the tests is with `pytest` (or `python -mpytest <script>`, which is identical).  

This is particularly important for unit tests that use text fixtures (common data items shared among tests, like `test_event` (defined in `conftest.py`) which is a common sample event one can use to test algorithms.

rlopezcoto: I got rid of the main block for the test (I think I included it because I saw it in other test files, but cannot remember which one...)

rlopezcoto:
I included another container to be filled the _pixel ring completeness_ (number of pixels within the muon ring width that are above a given threshold)

rlopezcoto:
travis-ci is complaining about the `test_json_2_fits.py` test:
```
 ctapipe/utils/tests/test_json_2_fits.py::test_jsonToFits FAILED          [ 98%]
```

although I did not touch anything for this test and it works on my machine (python 3.6.3)

dneise:
I assume the failing tests having nothing to do with your code, but are connected to #653

kosack:
> travis-ci is complaining about the test_json_2_fits.py test
It's due to an update to astropy (for which we don't have a version fixed, but probably should). It will resolve itself if #653 is merged in (should be soon).

rlopezcoto:
ok, #653 was merged and now the code is passing all the checks again.

PR #650, opened 2018-02-16
Enforce container cannot inherit twice

### LINKED ISSUE #649: Enforce particular Container inheritance rule
There is a rule in the container Docu that says:
https://github.com/cta-observatory/ctapipe/blob/39cf9af6df8ffed682982aa40a0581d722db76a3/ctapipe/core/container.py#L85-L86

I think this rule should be "must not" and not "should not". Since devs who still do it, will have problems that take them long hours to debug. I think such strange limitations must be enforced and not only documented. 

I make a proposal how to do it in a PR

dneise:
should fix #649

### DIFF
--- a/ctapipe/core/container.py
+++ b/ctapipe/core/container.py
@@ -46,6 +46,12 @@ class ContainerMeta(type):
     and no new fields can be added to a container by accident.
     '''
     def __new__(cls, name, bases, dct):
+        for base in bases:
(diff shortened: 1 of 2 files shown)


MaxNoe:
See #651, which enables inheritance instead of forbidding to use the broken inheritance

dneise:
perfect

Issue #642, opened 2018-02-13
Container construction is suboptimal due to deepcopy


dneise:
First an observation: Creating an empty DataContainer takes ~200us 
```
In [1]: from ctapipe.io.containers import DataContainer
In [2]: %timeit DataContainer()
189 µs ± 2 µs per loop (mean ± std. dev. of 7 runs, 10000 loops each)
```

when I remove the `deepcopy` from `Container.__init__` it speeds up factor of ~100
```
In [2]: %timeit DataContainer()
2.02 µs ± 25.2 ns per loop (mean ± std. dev. of 7 runs, 100000 loops each)
```

Nothing surprising here, `deepcopy` is known to be a costly operation.
Also, I am no fan of premature optimization, which leads to worse code readability or even worse usability.
But when slow code in module A leads to worse usability in module B, I think then some performance optimization, might be in order... or at least, it might be worth to talk about it.

I must admit, that I do not fully grasp, why the `deepcopy` is needed in that place. I would assume, that it is possible to define defaults, which are mutables. And in order to avoid the [mutable default arg gotcha](http://docs.python-guide.org/en/latest/writing/gotchas/#mutable-default-arguments) deepcopy was used here.

I am now taking a closer look at the Container to understand this. Any help is appreciated. Also, in case the same discussion has already been lead, when Container was implemented, and I am wasting time here, please just point me to the result.

Thanks a lot :-D

dneise:
Ah I forgot to mention the problem I saw, which in my understanding came from the fact that creating an empty DataContainer is costly.

So I saw this (short version):
```python
def event_source(path):
    container = DataContainer()
    for event in low_level_file_reader(path):
        fill_container_from_event(container, event)
        yield event
```

And then a user of the event source did this:
```python
events = [e for e in event_source(path)]
```

Of course all events where the same.

-----

keeping multiple events, is of course explicitely something we are not planning to do in ctapipe. We are thinking in streams, so there is absolutely no reason for any user to keep a couple of events in a list.

But ... I did anyway. I wanted to play around with them and not always loop over the file again. 

I think it makes code harder to understand, when the result of an operation, in this case an iteration, keeps under ownership of the operand. I currently do not remember a name for this principle, but I think I cam across it in the context of functional programming...

Imagine this:
```python
random = np.random.normal(size = 100)
large = np.where(random > 0)
# look at `large` here

# modify `random` a bit
random += 0.1

# would you expect `large` to have changed at this point?
```

[...]

PR #651, opened 2018-02-16
Enable field inheritance for containers, add test, fixes #648

### LINKED ISSUE #648: Containers do not inherit fields

MaxNoe:


dneise:
Do you also want to allow:
```
class ContainerA(Container):
    a = Field('a')
    great_field = Field(None)
class ContainerB(Container):
    a = Field('b')
    awesome_field = Field(None)
class ContainerC(ContainerA, ContainerB):
    c = Field(None)
```
???

MaxNoe:
I would, python does not.

Python does not allow inheritance from two classes who share the same base class

dneise:
Ok. thanks

kosack:
Seems fine to me, though so far we do not have containers inheriting from others since we used composition over inheritance, since they are supposed to be similar to database schema definitions,  but perhaps we will in the future (e.g it might be nice to have all containers that have an `event_id` and `time` to have a common base to avoid some duplication). In any case it should probably work sensibly.

kosack:
One question: does this slow down construction much? e.g. does it make the concerns of #642  worse?

MaxNoe:
@kosack 

This code runs once at class definition. (It is in the metaclass)

PR #652, opened 2018-02-20
remove sensitivity code, now in separate package

kosack:
the `ctapipe.analysis.sensitivity` code was deprecated and is now replaced by one of 2 packages:

* `irf_builder`  (a separate package for optimizing cuts and calculating sensitivity from ctapipe outputs, to which the old sensitivity code was moved)
* `gammapy.sensitivity`, code for binned sensitivity calculation


kosack:
Right now it's https://github.com/tino-michael/irf_builder, to be moved to cta-observatory shortly

PR #653, opened 2018-02-20
fix bug with json_to_fits and astropy 3.0

kosack:
astropy 3 is a bit more conservative when working with files, and the json_to_fits test function broke. This is a simple fix:

- pass string to json_to_fits, not a filehandle (no longer allowed in update mode for fitsio)
- rename a few functions to avoid deprecated names


<tempfile._TemporaryFileWrapper object at 0x1819803dd8>
```diff
@@ -0,0 +1 @@
...                                                                                                                                                                                                                                                                                                                                                                                                             XTENSION= 'TABLE   '           / ASCII table extension                          BITPIX  =                    8 / array data type                                NAXIS   =                    2 / number of array dimensions                     NAXIS1  =                    0 / length of dimension 1                          NAXIS2  =                    0 / length of dimension 2                          PCOUNT  =                    0 / number of group parameters                     GCOUNT  =                    1 / number of groups                               TFIELDS =                    0 / number of table fields                         VERSION =                    1                                                  EXTNAME = 'GLOBAL  '           / extension name                                 END
```
dneise: I think this tempfile should not be part of this PR?

kosack: oops, yes that was mistakenly committed. I'll remove it.

kosack: also, it seems to be a bug I introduced - the object name is being used as the filename...

PR #654, opened 2018-02-20
TargetIOEventSource

watsonjj:
Creating the `EventSource` for TargetIO files


kosack:
needs to be merged with the master, since there are some small conflicts due to the ZFITSEventSource code that was recently merged

MaxNoe:
You pushed a private ssh key. 

This should not happen.

watsonjj:
I am aware of how to properly encrypt private keys. However I decided it isn't that important in this case. It is a deploy key whose only ability is to unlock **read-access** to the target_software repository. The contents of that repository are not "top secret", but I'd rather dissuade people using it as a resource for the target libraries (they should download the source files from svn as I state in the installation instructions.

MaxNoe:
> I am aware of how to properly encrypt private keys. However I decided it isn't that important in this case. It is a deploy key whose only ability is to unlock read-access to the target_software repository.

Then why is this repository not public?????

Either it's private so also the key needs to stays private or it's public and thus the key is not needed.

watsonjj:
"Please note that encrypted files are not available for pull requests from forks."
- https://docs.travis-ci.com/user/encrypting-files/

watsonjj:
@kosack When can a new release of ctapipe_extra be made? Then the tests should pass.

kosack:
> @kosack When can a new release of ctapipe_extra be made? Then the tests should pass.

I should get to that soon, I hope! In any case, I want to start to phase out all test files in ctapipe-extra, since we now have a public webserver where we can put then (and then have them automatically downloaded as needed, as astropy does it). The main problem has been that the package is now getting too big for storage on free hosting systems like anaconda cloud.

kosack:
Related to that - should we rename the targetio files from `*.tio` to `*.tio.fits`? I notice they are actually FITS files, so should have that extension. They can also be compressed to `*.tio.fits.gz` files if we want to make them smaller (the fitsio and cfitsio libraries handle that automatically without having to decompress the files first).  

```
% ftlist targetmodule_ped.tcal hc    

        Name               Type       Dimensions
        ----               ----       ----------
HDU 1   Primary Array      Null Array                               
HDU 2   DATA               BinTable     1 cols x 64 rows            

  Col  Name             Format[Units](Range)      Comment
    1 CELLS              20352E               label for field   1


```

.travis.yml
```diff
@@ -44,17 +50,24 @@ before_install:
...
-
     - conda create --name cta-dev python=$TRAVIS_PYTHON_VERSION
     - conda env update -n cta-dev --file environment.yml
     - source activate cta-dev
     - ulimit -s 16000  # increase stack size limit, for libhessio
     - pip install travis-sphinx
     - pip install codecov
     # ----- SST1M:
-    - conda install numpy protobuf libgcc
+    - conda install protobuf libgcc
```
kosack: should we just add this to environment.yml?  I think we can also add pip packages there for SST-1m (under a `pip` heading).  That way everyone who installs the package using environment.yml will get TargetIO and SST1m support (maybe that's not necessary, though)

kosack:
In any case, there is now a new ctapipe-extra (0.2.13), so if you restart the build, you can check if this works

kosack:
seems to cause a segfault in ctapipe/calib/camera/tests/test_r1.py::test_targetio_calibrator

Not sure why, though!

kosack:
seems to cause a segfault in ctapipe/calib/camera/tests/test_r1.py::test_targetio_calibrator

Another comment: do you really need to build `cfitsio`? That adds a lot of time to the travis tests.  It's included in conda, so you can just add it to environment.yml as a dependency. The best would be to provide a pre-built targetio library as a conda package, as we do with pyhessio and others, to speed up the testing and make it easier for others.  All you have to do is make a meta.yml file, so it's pretty easy (you can just copy the one from pyhessio or ctapipe, which are in the cta-conda-recipes repo)

watsonjj:
> Related to that - should we rename the targetio files from *.tio to *.tio.fits?

I would request we don't do that... it is the convention within GCT that they have the .tio extension, and although it is essentially a fits file format, the data inside is structured in an unconventional manner meaning typical fits readers will not make sense of the data on their own, and instead you need the TargetIO library to read them.  The files themselves are already stored in uint16 to reduce size.

The .tcal file is a typical fits format (as you have demonstrated), but again, it is convention in GCT to keep these extensions so the different file types can be identified.

.travis.yml
watsonjj: This is not a good approach for the Target reader, the target_software repository only exists so TravisCI can perform the tests, no user should rely on these pre-compiled binaries. Instead they should install the TARGET libraries from source following the instructions at https://forge.in2p3.fr/projects/gct/wiki/Installing_CHEC_Software (which involves obtaining the latest source files from the cta svn, and performing cmake with swig on the three individual libraries, which also have dependencies on each other).

I have no idea how to make the TARGET libraries "conda installable" due to their complexity and the installation procedure involved, but I imagine it would be a huge undertaking that I do not have time for at the moment.

watsonjj:
I will investigate whats causing this error.

I will also try to use the conda packages for swig and cfitsio. I just need to figure out how to compile the TARGET software against them.

I would love to learn how to correctly package the software. My current method involves copying the pre-compiled binaries to the correct places, but the binaries are very dependant on OS, Python version, and also numpy version (due to the swig dependance), and depend on each other. This approach may work for now, but I'd love to improve this in the future when I have more time. However I haven't been able to find much information for this specific case.

kosack:
Ok, for now it may be best to stick to this method if TargetIO's build is so non-standard, though it does drastically increase the test complexity and execution time (We could always have target-based tests optional in the end). I'm worried that it is too fragile to maintain if a good build system is not used. 

If you do want to make a binary, usually as long as libcfitsio is in your LD_LIBRARY_PATH (which gets set automatically in a conda environment), you don't need to do anything special to link to it other than to add -lcfitsio to your build command, though for the header files, you may need to add a more specific includepath (which makes things more complex).  In the past for C++ code, I've used CMake, and used the *FindCFITSIO.cmake* package to have it located and linked automatically (see  the old `svn.in2p3.fr/cta/DATA/PIPELINES/Prototypes/ProtoReco/trunk/CMakeLists.txt` for example in the CTA SVN, which you can access via redmine)

If you don't use a nice builder like  CMake/Automake/SCons, it's of course a bit harder to do in a completely generic way, but you can always just rely on an environment variable in your Makefile that we can set in travis, or look into using libtool.
conda-build is quite nice in that if you can use it to run your build script, it often does the right thing automatically, and even makes your library relocatable, but often you need to start with a good build system to ensure things end up in correct places.

kosack:
Looking at your code, you are copying binaries directly for the target libraries, but then rebuilding cfitsio, etc, which may not work since the system is not the same. 

If this ends up being too complex, for now I'd recommend just removing the targetio stuff from the travis build, and adding the following to the test files:

```py3
import pytest

if not pytest.config.getoption("--have-targetio"):
    pytest.skip("--have-targetio is missing, skipping tests", allow_module_level=True)
```

Then, the default will be to skip those tests, and only run them if one specifies `pytest --have-targetio`

Of course then, it's up to you or somebody to make sure they still work, until we can build a nicer binary package that works everywhere.

watsonjj:
I will spend a bit longer trying to get it working with the pre-compiled binaries. I use a private github repository target_software_compilation to compile the TARGET libraries on Travis, for the same Travis build configuration that is used by ctapipe, and then upload the resulting binaries to the target_software repository, so in theory the binaries should be compatible for the Travis tests.

I also test the pre-compiled binaries in target_software using Travis, and as you can see the tests currently pass, so I should be able to get it working in ctapipe too :)

https://travis-ci.org/watsonjj/target_software

Thanks for the advice, I was a little lost on how best to approach this for a while.

watsonjj:
@kosack pytest is now completing fine and I have switched to using the anaconda builds of cfitsio and swig instead of building from source. The travis build is failing though for what looks like a sphinx issue.

kosack:
> The travis build is failing though for what looks like a sphinx issue.

Yes, `travis-sphinx` seems to be broken, though they seem to have fixed it today, so hopefully they will push a new version to pypy shortly

kosack:
after #663 works, I will merge this

Issue #658, opened 2018-02-22
Example for a simple analysis properly using `Tool`


MaxNoe:
Working on it

PR #659, opened 2018-02-22
Implement a path trait

MaxNoe:


MaxNoe:
@kosack The tests fail because of some unrelated sphinx error, can you check?

kosack:
>The tests fail because of some unrelated sphinx error, can you check?

seems that automodsumm either was updated, or needs to be updated to sync with sphinx. I'll try fixing the version number to an older one.

kosack:
This is nice. You may need to add a check that it is not a URL in the validator though, since so far things like `HESSIOEventSource` accept either URL for a remote file or a file path for local ones. If we change to a `Path` trait, then it would break that behavior.

kosack:
The sphinx problem should be fixed now.

MaxNoe:
> This is nice. You may need to add a check that it is not a URL in the validator though, since so far things like HESSIOEventSource accept either URL for a remote file or a file path for local ones. If we change to a Path trait, then it would break that behavior

Ah, I didn't know that. So having an URL trait might make sense as well

MaxNoe:
But adding this trait doesn't mean we have to use it everywhere, I was thinking about using it the example I talked about here #658

kosack:
> But adding this trait doesn't mean we have to use it everywhere
agreed - it's still quite useful.

dneise:
Could you describe what this is about?

dneise:
Do I understand correctly? 
 * `exists=False` means "The file/dir must not exist." .. 
 * `exists=True` means "The file/dir must exist." ..

How does: "I don't care if they exist" work? `exists=None`?

MaxNoe:
> How does: "I don't care if they exist" work? exists=None?
Yes, this is the default

dneise:
There are a couple of Python built in exceptions like [FileExistsError](https://docs.python.org/3/library/exceptions.html#FileExistsError) or [FileNotFoundError](https://docs.python.org/3/library/exceptions.html#FileNotFoundError)  

I was wondering if these are probably better suited to raise here, than a pretty generic `TraitsError`. ... but then again .. maybe TraitsError is just the right thing here.

dneise:
> How does: "I don't care if they exist" work? exists=None? - Yes, this is the default

Great thanks.

MaxNoe:
Oh yes, you are right. Haven't thought about it, just followed the example in the traitlets docs.

PR #661, opened 2018-02-22
Add basic tool for a hillas analysis

MaxNoe:
I tried to come up with a simple hillas analysis.

I'd say I think I failed. 

* I did not find a way to properly output the results (so it's jsonl for the moment)
* The code was labourious to create, I had to read a lot of code in ctapipe to get it running
* I think the config system is not what we want, e.g. there seems no way to create positional commandline arguments
* The need to define aliases for Traits directly defined at the Tool seems very much superflous

We should use this as a starting point for discussion how we can improve the steps.

### DIFF
--- a/examples/hillas_analysis.py
+++ b/examples/hillas_analysis.py
@@ -1,0 +1,73 @@
+'''
+This example shows how you should use ctapipe.core.tool.Tool
+for your analysis as example for a simple Hillas based analysis.
+'''
+from ctapipe.core.tool import Tool
+from ctapipe.core.traits import Unicode, Dict, List
+from ctapipe.utils import get_dataset
+from ctapipe.io.eventsourcefactory import EventSourceFactory
+from ctapipe.calib import CameraCalibrator
+from ctapipe.image.hillas import hillas_parameters
+from ctapipe.image.cleaning import tailcuts_clean
+
+import astropy.units as u
+
+import json
+
+
+class HillasAnalysis(Tool):
+    # class members are options, that can be set via commandline or config file
+
+    # inputfile, if None we will use the gamma test file
+    inputfile = Unicode(None, allow_none=True).tag(config=True)
+
+    # outputfile
+    outputfile = Unicode('', allow_none=False).tag(config=True)
+
+    classes = List([CameraCalibrator, EventSourceFactory])
+
+    aliases = Dict({'outputfile': 'HillasAnalysis.outputfile'})
+
+    def setup(self):
+        print('inputfile', self.inputfile)
+        print('outputfile', self.outputfile)
+        self.event_source = EventSourceFactory.produce(
+            tool=self, config=self.config,
+            input_url=self.inputfile or get_dataset('gamma_test.simtel.gz')
+        )
+        self.calibrator = CameraCalibrator(
+            tool=self,
+            config=self.config,
+            event_source=self.event_source
+        )
+        self.output_fh = open(self.outputfile, 'w')
+
+    def start(self):
+        for event in self.event_source:
+            self.calibrator.calibrate(event)
+
+            for telescope_id, dl1 in event.dl1.tel.items():
+                camera = event.inst.subarray.tels[telescope_id].camera
+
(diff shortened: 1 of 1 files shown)


kosack:
For dumping data, there's nothing perfect so far, but you can call `hillas_parameters()` with the `containter=True` option (which will soon become the default, once we merge in some other code and fix places that use the old structure), and then use the `io.HDF5TableWriter` to write out that container to a row in an output file. 

We're also working this week on starting to merge in all the code from Tino that does the full reconstruction chain (will need some refactoring), and that may serve as a more complete example to see how we want to refactor things. 

For the configuration, it's true there are no positional arguments, but I don't see that as being too bad (you can sort of do them with sub-applications), since in the end this is code that will run in an automated fashion on the grid, so the config will  be in a database, not run from the command-line.  That being said, there are a lot of things I don't like about `traitlets.config`, but most can be solved by writing some custom code.

examples/hillas_analysis.py
```diff
@@ -0,0 +1,73 @@
...
+    aliases = Dict({'outputfile': 'HillasAnalysis.outputfile'})
+
+    def setup(self):
+        print('inputfile', self.inputfile)
+        print('outputfile', self.outputfile)
+        self.event_source = EventSourceFactory.produce(
+            tool=self, config=self.config,
+            input_url=self.inputfile or get_dataset('gamma_test.simtel.gz')
+        )
+        self.calibrator = CameraCalibrator(
```
kosack: this should probably explicitly use a set of `Camera*Factory` classes rather than the `CameraCalibrator`, so all the various algorithm options are available. Ideally we should make that a simpler process.

MaxNoe: Can you give some more hints how to uses those?

mackaiver:
How would one go about adding directional reconstruction of events to this kind of tool?

kosack:
> How would one go about adding directional reconstruction of events to this kind of tool?

you can see `examples/plot_theta_square.py` which isn't quite perfect, but does it mostly right. I'm working right now on merging in the EventPreparer (from Tino's analysis) that encapsulates most of the steps needed to finish the telescope-wise event processing and write the DL1 parameters, and then it's just a matter of using the HillasReconstructor.   Right now, there is no really clean way to do it until I finish with that code.

examples/hillas_analysis.py
```diff
@@ -0,0 +1,73 @@
...
+            tool=self,
+            config=self.config,
+            event_source=self.event_source
```
watsonjj: This should be "eventsource" instead of "event_source"

mackaiver: I have to say I find this most surprising. Why are there even two parameters with similar names that behave so differently?  The `event_source` kwarg is not documented. Maybe we only allow one of these?

watsonjj: Its because of kwargs. "event_source" does nothing, but doesn't throw an error when its used. Not sure how to better handle that, the kwargs is required as part of the traitlet functionality. 

We could just explicitly throw an error if someone attempts to use "event_source", but thats one specific fix for what is a wider (potential) problem.

watsonjj: So that the traitlets can be configured when initialising the class

MaxNoe: kwargs is passed to `super().__init__`, `super()` of `CameraCalibrator` is `Component`. Which possible kwargs are we wanting to hand over to Component?

watsonjj: any of the CameraR1Calibrator (and its subclasses) traitlets. Currently the only ones that exist are the ones in TargetIOR1Calibrator (pedestal_path, tf_path...). But this is not the only class that adheres to this pattern, every class that is a component uses kwargs in this way so that the traitlets are correctly configured.

watsonjj: There are three ways traitlets can be set: passing a config file to the executable, setting them at the cmd line as arguments to the executable, or passing it as an argument in the initialisation of the class

mackaiver:
We decided to close this and try again.

PR #666, opened 2018-03-07
ctapipe.image.cleaning.tailcuts_clean speed up factor ~10

dneise:
Same as #665, but this time doing only the smallest possible diff to reach the result.

No additional tests, so no proof that this is faster than the previous version. Proof was shown in #665 

Regarding readability vs speed, I think some people might even like matrix notation better than the previous version, but who knows.

One peculiar thing is this:

Before:
```
boolean_vector = (Matrix & Vector).any(axis=1)
integer_vector = (Matrix & Vector).sum(axis=1)
```
Now:
```
boolean_vector = Matrix.dot(Vector)
integer_vector = Matrix.dot(Vector.view(np.byte))
```

This might actually confuse the readers ... "Why is this view needed here?? WTF?"
The point is obvious for us when I write it down here ... but a month from now I might have forgotten it ... see this code and remove the view, since I think it is superfluous. 

Luckily there is a test, that immediately fails, when this `view(byte)` is removed, so we have a test harness for this peculiar piece of code, which is really good.

Of course instead of `view(byte)` also one could have written `astype(int)` or so ... gives the same result. but view is faster than astype and a numpy boolean array is a char array in the background, so this view was the fasted version I found. It still costs ~430ns which is a lot, given that the view is not actually doing anything, it is just needed so that the result of the later dot-product is not cast

### DIFF
--- a/ctapipe/image/cleaning.py
+++ b/ctapipe/image/cleaning.py
@@ -30,11 +30,11 @@ def tailcuts_clean(geom, image, picture_thresh=7, boundary_thresh=5,
     picture_thresh: float or array
         threshold above which all pixels are retained
     boundary_thresh: float or array
-        threshold above which pixels are retained if they have a neighbor 
+        threshold above which pixels are retained if they have a neighbor
         already above the picture_thresh
     keep_isolated_pixels: bool
-        If True, pixels above the picture threshold will be included always, 
-        if not they are only included if a neighbor is in the picture or 
+        If True, pixels above the picture threshold will be included always,
+        if not they are only included if a neighbor is in the picture or
         boundary
     min_number_picture_neighbors: int
         A picture pixel survives cleaning only if it has at least this number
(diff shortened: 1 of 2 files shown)


dneise:
Now that I think about it, maybe it is worth to just grep for all use cases of neighbor_matrix and see if the sparse version can be dropped in, without bothering about profiling the possible impact on Performance... just assuming it might help.

Not good working style but hey... the neighbor matrix is really sparse by design.

dneise:
Turns out `neighbor_matrix` was only used in one more function `dilate()` in the same file. So I just used it there as well...hope that is file for you.

ctapipe/image/cleaning.py
```diff
@@ -56,8 +56,8 @@ def tailcuts_clean(geom, image, picture_thresh=7, boundary_thresh=5,
     else:
         # Require at least min_number_picture_neighbors. Otherwise, the pixel
         #  is not selected
-        number_of_neighbors_above_picture = np.sum(pixels_above_picture &
-                                                   geom.neighbor_matrix, axis=1)
+        number_of_neighbors_above_picture = geom.neighbor_matrix_sparse.dot(
+            pixels_above_picture.view(np.byte))
```
kosack: regarding your previous comment, you could just add some note in the comment text here as to why the view is used (to coerce it into an int), so people won't be confused later

dneise:
> In fact when I originaly implemented the neighbor_matrix, I had used a sparse matrix, but it made things much slower for some reason (a factor of around 20x if I recall) - not sure why it is different now! It could be due to differences in versions of numpy and whether or not some optimizations were enabled.

Interesting.... I remember having used sparse matrices in other projects of mine some years ago, and they performed as expected.

I assume you are aware of these [different types of sparse matrices](https://docs.scipy.org/doc/scipy/reference/sparse.html) and the fact that not all of them are good at dot-product... 

Anyway thanks for the quick merge.

PR #668, opened 2018-03-08
adjust sst1m eventsource to novel protozfits interface

dneise:
This might be interesting for other proto-zfits reading people like @FrancaCassol 

The novel protozfits interface `SimpleFile` is more general than the interface in v0.44.1. 
This means that operations that were previously hidden in `protozfitsreader` were now shifted into the eventsource in ctapipe. I think this makes more sense, since it is much more explicit.

A reader should, after all, only read the data from the file and not apply some arbitrary transformations on the data, however common they might be...

For example, it is necessary to sort the waveforms, since they are not stored in the order of pixel_ids but in some other order. Previously this sorting was done in the protozfitsreader, now it is done in the event source.

### DIFF
--- a/ctapipe/calib/camera/r1.py
+++ b/ctapipe/calib/camera/r1.py
@@ -14,13 +14,14 @@ Through the use of `CameraR1CalibratorFactory`, the correct
 of the data.
 """
 from abc import abstractmethod
-
+from traitlets import Unicode
 from ctapipe.core import Component, Factory
 from ctapipe.io import EventSource
 
 __all__ = [
     'NullR1Calibrator',
     'HESSIOR1Calibrator',
+    'TargetIOR1Calibrator',
     'CameraR1CalibratorFactory'
 ]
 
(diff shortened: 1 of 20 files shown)


ctapipe/calib/camera/r1.py
```diff
@@ -13,16 +13,20 @@
 `CameraR1Calibrator` can be obtained based on the origin (MC/Camera format)
 of the data.
 """
+import os.pyth
```
kosack: do you mean `os.path` here?

ctapipe/calib/camera/r1.py
```diff
@@ -308,6 +312,127 @@ def real_calibrate(self, event):
...
+        help='Path to a SST1M dark-baseline file (at the moment .npz)'
+    ).tag(config=True)
+
+    def __init__(self, config=None, tool=None, **kwargs):
+        super().__init__(config=config, tool=tool, **kwargs)
+        self.baseline_estimator = _SST1M_BaselineFromRandomTrigger(
+            n_bins=self.n_bins_for_baseline_estimation
+        )
+        if os.path.isfile(self.dark_baseline_path):
+            self.dark_baseline = np.load(self.dark_baseline_path)['baseline']
```
kosack: Any files you load must be recorded in the provenance system, so either use `utils.get_dataset()` to do it automatically (which also searches automatically everywhere in `CTAPIPE_SVC_PATH`, which can contain multiple directories), or add the following line afterward:

```py3
Provenance().add_input_file(self.dark_baseline_path, role='dl0.tel.svc.sst1m_baseline')
```

kosack: probably best to just use 
```py3
get_dataset(self.dark_baseline_path, role='dl0.tel.svc.sst1m_baseline')
```
since having an abstraction for reading files allows us to later switch to reading this from a database, or remote server without changing any code.

dneise:
I am sorry, Karl. I must have made a mistake. This PR should not have contained anything related to the R1Calibrator. I must have made a git-mistake.

dneise:
So... could this me merged? The SST1M EventSource which is updated here to make use of the novel interface has no users at the moment, so nobody will be affected.

However the novel interface of the underlying library "protozfits" seems to make more sense in the light of more people than only SST1M using fizs.fz files nowadays.

----

Regarding the the Codacity Review: It does not like the names of two functions, but the names were not modified in this PR, so if they were good before, they are still ok. The function names contain capital letters since they refer to class names, which are capital, and I do not want to cause confusion in the readers mind, by making class names lower case in function names.

----

Regarding the codecov. I do not fully  understand how to read the details of the report, but I realize that unit tests are missing. None of these "fill"-functions is tested. And they are bound to fail horribly as soon as either the containers structure, they are trying to fill, changes or when the "events" structure they are using changes. However, all these functions are at least used in the very basic `test_loop_over_events()` test. So if one of these functions tries to fill a field which was removed, or trying to read from a field, that was modified, this test will find the problem.

dneise:
As one can see in  c8f6f84 I forgot to adjust the package name from "protozfitsreader" to the new name "protozfits". This resulted in all tests being skipped! Which can potentially lead to quite a large disaster. 

While this was of course my very own mistake, I wanted to take the opportunity to make also others aware of this. I must say I was really devastated when I realized that I was about to happyly request to merge utterly broken code into the master, despite having tests. I was wondering if also others were aware that the our test harness might have gaping holes.

I have no time to fix these mistakes at the moment, but I wanted to make sure the tests fail, so that nobody my chance merges this PR.

dneise:
I do not understand what is going on here... it shows that the tests passed on travis, but this piece of code cannot work:
https://github.com/dneise/ctapipe/blob/c8f6f8404e3567bff0abfed11bb2ee4938188f91/ctapipe/io/tests/test_sst1meventsource.py#L4-L14

kosack:
It certainly looks like protozfits is installed, according to the logs.  Not sure why it wouldn't find it, unless its installer is broken.  Let me know when that is resolved, and I'll merge it in, since as you said, it doesn't really affect anyone else!

dneise:
@kosack indeed the installation does not work.

Problem might be different protobuf versions: 

here is travis of ctapipe:
https://travis-ci.org/cta-observatory/ctapipe/jobs/352752733#L852-L858

here is travis of protozfits:
https://travis-ci.org/cta-sst-1m/protozfitsreader/jobs/352758602#L578-L586

First I though  this is a result of us using different versions of miniconda ... protozfits uses alway the latest, while ctapipe uses a specific version (4.3.21) but that was not it ... looking further into this.

dneise:
@kosack Ah yes it was a problem with the conda version. 

travis recommends to do "conda update -q conda" after installing miniconda, and I do that in the `protozfits/.travs.yml` while it was commented out in the `ctapipe` version. I assume there was a good reason for it. 

I think the best way to proceed for me (and SST1M/digicam people in general) is to use the same version of conda as you use in ctapipe, so that such compatibility problems are reduced to a minimum.

So just for comfirmation, am I correct to assume that there is a reason to not use the latest version of conda and do not "conda update conda"? Or was this just done by accident?

dneise:
I missed these lines in the travis.yml

https://github.com/dneise/ctapipe/blob/e976f008a4922754c9e8f82c6ca053a99cd1987d/.travis.yml#L30-L32

In PR #677 I am proposing to use the latest version again. tests pass .. and I checked by eye ... the sst1m tests were not skipped.

dneise:
Last tests did fail due to some conda install timeout ... Actually just a restart of the build was needed, but I have not sufficient rights to restart a build...

dneise:
@kosack Ok. I think this can be merged, **if** you are happy with #677. Since this (at the moment) only works with the latest conda. I can work on making this work with older condas as well, but I'd rather only do that if really necessary.

kosack:
> travis recommends to do "conda update -q conda"

I think I had disabled that at some point due to a bug in conda that made it into the latest release and broke all builds, but it should be ok to put it back in now.   Otherwise, this looks fine now! Thanks for the detective work.

Issue #669, opened 2018-03-08
HDF5TableWriter and data types


calispac:
Hi all! 
I was testing a bit the implementation of the `io.HDF5TableWriter` and `io.HDF5TableReader`.

So I came up with two issues:

1. The writer does not seem to support `str` and `datetime` types
2. The writer + reader does not have the "`with`" feature  : 
```python

with HDF5TableReader('my.h5' ) as f:
     f.read('root/table', MyContainer)
```

To my understanding one has to do : 

```python
f = HDF5TableReader('my.h5')
f.read('root/table', MyContainer())
del(f)
```
The `__del__()` function in the table writer closes the file. If I make a mistake in the read() for instance:

```python
f = HDF5TableReader('my.h5')
f.read('root/NOT_THE_CORRECT_NAME', MyContainer())
del(f)
```
Then the reader fails without closing the file. In principle this is not an issue if one calls this piece of code with:

```
python my_piece_of_code.py
```

But working with Jupyter Notebook did cause an issue for me. I believe the cell I executed did not close the file. So then latter when I needed the file again I could not open it. 

Here is a notebook : https://nbviewer.jupyter.org/github/cta-sst-1m/digicampipe/blob/just_notebooks/notebooks/example_container_writer.ipynb

So regarding the two points:

1. What are the types supported? And if one type is supported by the `Container()` why is it not supported by the TableWriter?
2. Is an implementation of the with feature foreseen ? Did I just missed it ?

Thanks in advance

MaxNoe:
This is very much related to #445 and #501 

And yes, your points are correct.

kosack:
In principle it can support anything that pytables does - I just only implemented the bare minimum to start.  Right now the type mapping is done via a type map and in come cases some transform functions (e.g. to add or strip units, etc):

```
PYTABLES_TYPE_MAP = {
    'float': tables.Float64Col,
    'float64': tables.Float64Col,
    'float32': tables.Float32Col,
    'int': tables.IntCol,
    'int32': tables.Int32Col,
    'int64': tables.Int64Col,
    'bool': tables.BoolCol,
}
```


But as @MaxNoe mentioned, we plan on moving to a more "ORM-like" model where you specify the type as a subclass of `Field` (e.g. `Int32Field`), which allows us to add specializations and mapping functions in a more clear way.

> The writer does not seem to support str and datetime types

[...]

PR #670, opened 2018-03-08
Table writer context

calispac:
I suppose this solves an issue in #669.

I created abstract methods in the TableWrite and TableReader:

- close()
- open()

the __enter__() and __exit__() methods are handled by the parent class.


calispac:
I just need to run some tests. Not sure it works as supposed yet.

kosack:
could you also add a small unit test to  test this functionality in `ctapipe/io/tests/test_hdf5.py`?

calispac:
> could you also add a small unit test to test this functionality in ctapipe/io/tests/test_hdf5.py?
Sure!

MaxNoe:
I would completely remove the `__del__` method.

calispac:
I have added some more. I would like to test the fact that the __exit__() method really closes the file but I haven't found a way to do that...
Anyway it should be in the TableWriter and TableReader.

MaxNoe:
> I would like to test the fact that the exit()

The reference to the object still is there after the `with` block, so you can check if `writer._file.is_open` or whatever the underlying file provides

ctapipe/io/hdftableio.py
```diff
@@ -385,9 +428,18 @@ def __init__(self, filename):
...
         super().__init__()
         self._tables = {}
-        self._h5file = tables.open_file(filename)
+        self._h5file = self.open(filename)
         pass
 
+    def open(self, filename):
+
+        self._h5file = tables.open_file(filename)
+        return self._h5file
```
MaxNoe: why the return?

calispac: yes I think we can remove it

calispac: hum actully I don't like much that class attributes are definded outside of the __init__()

so I will make open() return the table without assigning it to the class attribute

MaxNoe: I don't think this makes sense. 

I think it's perfectly alright for a method called `open` to set internal state of a class

MaxNoe:
Sorry, wrong button. I should'n even be allowed to do this as a non-maintainer

ctapipe/io/hdftableio.py
calispac: ok!

calispac:
I believe it is ready. Just tell me if there is anything left.
Thanks for the fast review

MaxNoe:
Maybe a question for another PR, but does tables support appending? So could one add the mode to the args if `writer`? To make appending possible?

calispac:
> Maybe` a question for another PR, but does tables support appending? So could one add mode to the args of writer? To make appending possible?

I believe that looking at the current structure of Writer/Reader one should then have an "Appender". Therefore maybe this can be revised having only a HDF5Parser class which handles all the `mode`?

ctapipe/io/hdftableio.py
```diff
@@ -91,6 +99,16 @@ def write(self, table_name, containers):
         """
         pass
 
+    @abstractmethod
+    def open(self, filename):
```
kosack: need to add  **kwargs here, since the parameters of the overridden methods must match the base class

kosack: that's an old mistake, but might as well finish it now

MaxNoe: I this is not true. Python (at least not abc) does not check the method's signature:

```
from abc import ABCMeta, abstractmethod


class A(metaclass=ABCMeta):

    @abstractmethod
    def test(self):
        pass


class B(A):

    def test(self, name):
        print('Hello', name + '!')


b = B()
b.test('World')
```

PR #672, opened 2018-03-09
WIP: First shot at a SST1M R1Calibrator

dneise:
(This is based on #668. So the diff should become much smaller once #668 is merged.)

So .. R1Calibrator... I must admit I do not know what I am doing here, but I still wanted to give it a shot, sometimes it is easier to argue about something, if one has an example of how to now do it.

In order to understand what an R1Calibrator is supposed to do, I had a look at the already existing implementation, but I did not understand what they are doing.

So I read the (not yet published?) document named "R1 & DL0 Telescope Event Interfaces
and Prototype Evaluation". which I got by email, so I have no idea how to link this document, I could not even tell if I had the most recent version of it... anyway the version I had, says in chapter 3.1, that the R1 Calibration is supposed to do this:

 * remove any effects of ADC inhomogeneities in time (ring-sampler corrections)
 - [x] subtract a pixel-wise rough electronic + NSB noise pedestal value,  so that the values fluctuate
around 0 counts (and record it as MON or SVC data, so it can be un-applied or quality-checked
later)
 * if multiple gain channels exist, optionally select the appropriate channel such that the readout value
is within a linear range (or mark the pixel as saturated).
 * if a non-linear gain is used, apply the gain correction function
 - [x] apply a pixel-wise multiplicative gain correction factor to correct for differences in camera uniformity
 - [ ] apply a scalar scale and shift the signal so that it efficiently fits in a 16-bit signed integer value with
the required dynamic range (see Section 3.3)
 - [ ] record all applied corrections in SVC or MON streams

(Those, which have a bullet point and not a tick-box, do not apply for SST1M/digicam, but I though this list is so nice, I wanted to keep it complete.)

I think I did the first 2 points in this PR. I did not yet look-up how to write into SVC or MON streams, so I did not yet do it. 

I did not apply any special scaling so that the result fits into 16bit int rage, since it does so by design. The digicam ADC is 12 bit (unisgned int) and by removing the baseline and multiplying with the gain_drop (value between 0 and 1) I am bound to stay somewhere in a 12bit signed integer range. However, one might want to increase resolution and artificially multiply everything by 16 to make use of the 16bit range we have. However, this does not really add anything and I was not sure what is wanted.

I'd be happy for advice, if this is running in a somewhat right direction.


dneise:
Ah another reason, why I am feeling very unhappy about this is, that it is of course not yet clear how to best perform this R1 Calibration for digicam. We are in a prototype phase and people are working everywhere at the same time.

I personally think that sometimes one wants to see the effect of a modification in R1Calibration on high level output like ... dunno ... size distribution on an example data set. 

To study this, a telescope group would implement different R1Calibrators, but all the higher level stuff would stay the same and could be reused. (That is at least my dream ... )

On the other hand one always runs (at least) into the "cleaning-question", which goes like this: "When you modify how the baseline is defined, you also need to optimize the cleaning-levels for this case, otherwise the comparison is not useful at all".

So at least the cleaning levels (maybe the whole cleaning algorithm) needs to be adjusted for different R1-calibrations.

So my question is, how do other telescope groups perform these studies? What role does the `CameraR1CalibratorFactory` play in such a study, how could one utilize that factory to test different calibrators?  Are there examples for such studies?

kosack:
The R1 calibration is not well defined yet,  but the basic idea is that all cameras should put out data that is "good enough" to to a basic hillas analysis (for realtime purposes) and so that the data volume reduction can be performed in a generic way at the R1 level. 

That means at least applying some rough calibration (this is not the final calibration, so need not be perfect) so that camera images are uniform enough to determine where the shower is so we can drop non-shower pixels later.   It also means that the output should  still  be integers like in R0, not floats, so basically the pedestal and gain corrections just round to the nearest integer (this is necessary to store them in 16-bit ints in DL0).

What is done so far  for MCs in ctapipe is not quite this.  

Also keep in mind that the R1 calibration will almost certainly have to be implemented in the camera servers, so won't probably be in python (it's out of scope of ctapipe, except for monte-carlos), but it's useful to have it prototyped here so we can ensure we apply the same calibration to MCs as is done to the real data.

kosack:
Also the document you refer to is *not yet official* and was only circulated to the PC members as a status report, and still needs to go through a full review process with the camera groups.   Therefore do not assume anything in it is correct!. I'm also probably going to remove any mention of what R1 calibration means in that document, since that should be in a separate calibration specification.

dneise:
I #668 Karl made two very helpful comments regarding reading files. I will cite them here for completeness.

start quote:

---

Any files you load must be recorded in the provenance system, so either use utils.get_dataset() to do it automatically (which also searches automatically everywhere in CTAPIPE_SVC_PATH, which can contain multiple directories), or add the following line afterward:
```python
Provenance().add_input_file(self.dark_baseline_path, role='dl0.tel.svc.sst1m_baseline')
```
probably best to just use
```python
from ctapipe.utils import get_dataset
...

get_dataset(self.dark_baseline_path, role='dl0.tel.svc.sst1m_baseline')
```

since having an abstraction for reading files allows us to later switch to reading this from a database, or remote server without changing any code.

---

end quote.

dneise:
I have one question regarding the `role` string: How is this string designed? In your comment you propose: 
`"dl0.tel.svc.sst1m_baseline"` .. one might also assume something like `r1.tel.svc` or even `r0.tel.svc` since this baseline is *made from r0* level but *used for r1*.

kosack:
> I have one question regarding the role string: How is this string designed?

So far, we've just been using the data naming convention (DLx.{ARR,SUB,TEL}.{EVT,MON,SVC}.*).  We don't have a good convention yet, but I would suggest using the data level of where it is used, since that may be the most obvious... (though we should think about it)  So it may be R1 instead of DL0 if you are using it to compute the R1 waveforms.  I'd say go with what you think makes sense and we can write a note in the documentation about it, so we use the same convention afterward.

dneise:
Ah thanks for the clarification on this. I will soon fix that. 

So do I understand correctly, that (at the moment) these strings are not to be consumed by some software, but they are for humans? I assumed the purpose of the Provenance information was to provide machine readable information about the Provenance of a certain analysis result, so that it can be reproduced or the same treatment can be applied to similar input-data. 
In that case, I assume Provenance must be machine readable and it must be ensured that it is correct by a harness of tests-which make sure that no Provenance information is missing and that it is correct.

Anyway for the time being this seems to be not the case, so I'll try to come up with a useful name and that's it.

kosack:
>  these strings are not to be consumed by some software, but they are for humans? 

yes, the provenance is machine readable, and is supposed to show the chain of processing in a reproducible way.  The "role" is a required field from the W3C model, intended to describe the reason for the file (like a semantic name, not a file name), similar to a UCD.  How those roles are defined is up to CTA, and we don't yet have a dictionary for them, so so far I'm just using it to define which type of data are loaded. Really we should put "cal" or something in the name as well (e.g `r1.tel.svc.cal.XXX` so what we understand that this was calibration data for a telescope.

ctapipe/calib/camera/r1.py
```diff
@@ -309,6 +313,88 @@ def real_calibrate(self, event):
...
+            )
+        )
+    )
+
+    # measured in a toy-MC for the above mentioned nsb_rate
+    # the simulation takes some time, so we just put the results here
+    # https://github.com/cta-sst-1m/digicamtoy
+    # TODO: still have to find *the* exact source for these numbers
+    # but it is somewhere in that repo.
+    baseline_shift = np.array([
```
calispac: I think one could replace this by a formula at some point. We just need to check it with the data first. Could you mark this as a # TODO ?

mackaiver:
Is this still relevant?

PR #673, opened 2018-03-10
Feature/add gain selection

kosack:
I would like to include gain down-selection into the basic calibration.  In reality this may happen in-camera or at the DL0 level during data volume reduction (it is not yet decided). Nevertheless we need to have an implementation here, so that at least at the DL1 level, there is only a single calibrated image and parameters.

Implementation:
----------------
* Created classes for `GainSelector` (base class)
  * `NullGainSelector` (will leave both gains remaining)
  * `ThresholdGainSelector` (chooses gain based on threshold, defined in a calibration table called gain_selection_table.ecsv in *ctapipe-extra*)
* Created a `GainSelectorFactory` to allow the user to choose the method
* The `ThresholdGainSelector` has an option to either replace the full waveform with the high-gain one, if the threshold is exceeded, or to only replace samples above threshold (the latter assumes all calibration has been applied such that this will not introduce a discontinuity).  The default is to replace the full waveform, as done in current experiments.

```py3
selector = ThresholdGainSelector()
wf, mask = selector.select_gains(cam_id="NectarCam", multi_gain_waveform=waveform)
```

Where wf is now a single-gain waveform, and mask tells which channnel was used per pixel (if the mode is the default one), or *per sample*, if the mode is *select_by_sample*.

Questions/Todos:
------------------
* This must happen automatically for current MC files... so where should it go? In the `CameraR1Calibrator`, `CameraDL1Calibrator`, or `CameraDL0Reducer`? @watsonjj  do you have a preference?
* [ ] write more tests
* [ ] insert into calibration chain once the first question is answered.


watsonjj:
I imagine the two channels are not both stored in the DL0 level? If not then the selection should definitely occur in either the R1 or DL0 level. Which one out of those depends on if you think the gain selection is something that is more likely handled by the camera teams, or is handled in a common way by the OES. I propose that the R1 calibration is a more appropriate area for it.

Regarding the NullGainSelector, would it not make more sense for it to also return a single-gain-channel waveform, otherwise operations down the chain would not work properly? So I suggest it should return the first channel, as it is mostly going to be used by cameras that do not have 2 gain channels.

kosack:
>  is more likely handled by the camera teams, or is handled in a common way by the OES. I propose that the R1 calibration is a more appropriate area for it.

Just checked with Jim and he is advocating for the cameras doing it.  However, due to calibration, there may be special runs where both gains are read out, so R1 should support both cases. 

> NullGainSelector, would it not make more sense for it to also return a single-gain-channel waveform, otherwise operations down the chain would not work properly? So I suggest it should return the first channel, as it is mostly going to be used by cameras that do not have 2 gain channels

That sounds fine.

jacquemier:
+1 for gain selection in R1

kosack:
I think I'll merge this now, and add it to the R1CameraCalibrator in a separate step, since I have other pieces of the analysis chain to import from Tino's scripts and don't want to break anything yet.

Issue #679, opened 2018-03-15
EventSource seems to iterate again after loop has finished


kosack:
Running the following code produces an unexpected result:

```py3
source = event_source(filename, max_events=2)

for event in source:
    print(event.r0.event_id)

print("after loop:", event.r0.event_id)
```

output:
```
2105
2108
after loop: 2305
```
So the event after the loop over only 2 events is done, seems to be the 3rd event. This is certainly a bug, or at least not expected behavior for an iterable.

MaxNoe:
Is that true for all EventSources or only for e.g. hessio?

calispac:
I think the reason is because of the if inside the for loop: 

https://github.com/cta-observatory/ctapipe/blob/eeb1c6bd9ae949cf4d2c2c5b389076bbd0129653/ctapipe/io/eventsource.py#L191

calispac:
Nope sorry I went to fast...

dneise:
> Is that true for all EventSources or only for e.g. hessio?

I wrote 3 tests for the 3 event sources I know about (not sure if there are more already...)
https://travis-ci.org/cta-observatory/ctapipe/jobs/354194359#L1469-L1487

(tick shows passing tests)
 - [ ] Hessio
 - [ ] TargetIO
 - [x] FitsIO (i.e. actually only SST1M ...)

I assume at the moment, this might be connected to the design decision, to modify containers in place instead of emitting fresh ones. This decision can lead to little trip-wires... you see, when the event-source gives up ownership of an event after having it returned, there is virtually no chance to modify it after the loop. 

In the implementation of the sst1m_event_source I did deliberately break this design decision (at cost of runtime).  There is a fresh Container for every event. 
https://github.com/cta-observatory/ctapipe/blob/eeb1c6bd9ae949cf4d2c2c5b389076bbd0129653/ctapipe/io/sst1meventsource.py#L30

Of course one can fix this issue ... (I did not yet search for the exact source of the problem ... ) but I think it would be helpful to walk on a simple path and enjoy your life instead of walking on a path full of land mines, and just learn how to avoid them...

dneise:
I'm just trying this ... https://github.com/cta-observatory/ctapipe/pull/680/commits/2cc44ec23fd79a57269a961ede8b9bdc8bb2719e

Will  try your proposal afterwards .. then we can see what we like better ... just letting the tests finish... 10 minutes or so

dneise:
[...]

PR #680, opened 2018-03-16
Eventsource bug tests

dneise:
Tests regarding issue #679 

erm .. looks like this has a lot of commits ... but it is actually only b53346d.

The diff looks correct.


dneise:
This is superseded by #681

PR #681, opened 2018-03-16
Issue 679 tests and fix

dneise:
See #679 for details


kosack:
Thanks! This seems to do the trick.

PR #683, opened 2018-03-17
WIP: Add gain selection to r1 calibration

kosack:
I've attempted to make the data flow a bit more like what CTA is envisioning, where at the R1 level the gain-channel selection has already been performed.  For that the following change were make:

Main changes
--------------

* modified `HESSIOR1Calibrator` to include a GainSelector (threshold by default). I didn't know the correct gain-switch thresholds for most cameras, so put  some dummy values in `ctapipe-extra`. Those will need to be updated by experts 

* Modified all `ChargeExtractors` to assume a single gain channel (this simplified them a lot), and changed the options called *sig_amp_cut_LG* and *sig_amp_cut_HG* to a single option called *peak_detection_threshold* (which I think is more understandable)

* I changed the name of the `cleaned` Field in `R1CameraContainer` to be `waveform` as before (it now contains the gain-selected, cleaned, calibrated waveform)

*  added `waveform_full` and `gain_channel` mask to the R1Container, for times when both channels are still needed.

* updated the tests to expect a single gain past R1

Things that do not work so far
-------------------------------

* This has broken the `TargetIOR1CameraCalibrator` , which still expects 2 gains (i didn't want to touch that since I'm not sure exactly how it works.  This is probably the same for the `SST1MR1CameraCalibrator`, so I'll need some help with those. 

* same for the `CHECMWaveFormCleaner` - it breaks, since the input from the corresponding CameraCalibrator is still 2-channel.


ctapipe/calib/camera/gainselection.py
```diff
@@ -51,12 +50,12 @@ def pick_gain_channel(waveforms, threshold, select_by_sample=False):
             new_waveforms[gain_mask] = waveforms[1][gain_mask]
 
     elif waveforms.shape[0] == 1:
-        new_waveforms = np.squeeze(waveforms)
+        new_waveforms = np.squeeze(waveforms, axis=0)
```
watsonjj: Why use np.squeeze instead of waveforms[0]?

kosack: not sure - that was just copied from Tino's code.. I will change it to `waveforms[0]` since it makes more sense

dneise: Just out of curiosity, I see multiple mentions of something called 'Tinos Code'. Is this something one should know about?...

kosack:
I also defined a small Enum for the channel (`Channel.high`, and `Channel.low`) and made the option in the `SimpleGainSelector` be by name, not by index.

ctapipe/calib/camera/gainselection.py
kosack: >Just out of curiosity, I see multiple mentions of something called 'Tinos Code'. Is this something one should know about?...

Sorry that may be confusing: Tino (@tino-michael) is the one who created the full pipeline based on ctapipe that has been used so far to make sensitivity curves. To get it working, he had to create a few extra classes and scripts that are in his github repo, and they are not yet merged into ctapipe (this PR is the first step to doing that, but there will be several others once this is working).  It turned out to be not as simple as just copying his code in, since we wanted to move things around a bit, like having the gain selection in R1, rather than in Dl1 as it is in his code, hence some refactoring.

tino-michael: I think I used squeeze and not just the index because at some point there was an inconsistency somewhere with the time dimension. sometimes the whole dimension was gone, sometimes it was there but with length 1...
as in the different behaviour of `a = [1, 2, 3]` and `b = [[1, 2, 3]]`

kosack: I think that problem with missing time dimensions, etc, has been fixed now in the new EventSource classes and the calibration code... at least I haven't seen it anymore with any of the test data I used.

watsonjj:
I created a PR to your branch with the changes for the Target classes https://github.com/kosack/ctapipe/pull/2

dneise:
>  This is probably the same for the SST1MR1CameraCalibrator, so I'll need some help with those.
SST1M  does not have dual gain... so ... not sure there is a problem

kosack:
closed this since the calib refactoring will replace it

Issue #687, opened 2018-03-20
HDF5TableWriter/Reader improvements


calispac:
Hi all,

I am still playing around the `HDF5Table` classes. So far I have experienced some issues: 

1. Appending mode is not supported. 
```python 
with HDF5TableWriter(filename, group_name='group_1') as h5_table:
    h5_table.write('data'', datacontainer_1)
with HDF5TableWriter(filename, group_name='group_2') as h5_table:
    h5_table.write('data'', datacontainer_2)
```
AFAIK this will: 
* open the table
* write `datacontainer_1` in `/group_1/data/`
* close the table
* open the table and ERASE everything
*  write `datacontainer_2` in `/group_2/data/`

I think @MaxNoe pointed that out in this PR #670 

2. Writing container class is not supported: 

```python 
class ContainerA(Container):
    b = Field(ContainerB , 'a ContainerB field')
class ContainerB(Container):
   c = Field(int, 'a int field')

a = ContainerA()
with HDF5TableWriter(filename, group_name='group_1') as h5_table:
    h5_table.write('data'', a)
```
The former will not write `a` because `a` contains a field of type `Container` which is not supported.

For the moment I think it is fine because one can do : 

```python
with HDF5TableWriter(filename, group_name='group_1') as h5_table:
    h5_table.write('data'', a.b)
```

Could you confirm me that it is actually the case or am I just missing something ?
Concerning 1. I will try to play with the reader and set `mode='a'` in the `open()` function.
Concerning 2. I am not sure one can do this easily but maybe it is just a matter of adding  : 

```python
 if isinstance(value, Container):
       table_name = value.__class__.__name__
       self._create_hdf5_table_schema(table_name, value)
```

in the for loop of https://github.com/cta-observatory/ctapipe/blob/e4d9a307ec7bf2846bc5930de9b175ee9f5cf41a/ctapipe/io/hdf5tableio.py#L83-L152

kosack:
appending:
------------

Have you tried passing the `mode="a"` option when constructing it? Underneath HDF5TableWriter just passes all kwargs to `pytables.open()`, so in principle it should work, though I have not tried it. 

Writing container classes
--------------------------

[...]

PR #689, opened 2018-03-20
Append mode for HDF5Table

calispac:
This solves the first issue in #687.
I have updated the notebook to show the new feature.
I will find some time to write tests

I also added the possibility to write to an existing group_name. This will work only if the table name does not exist already. I am not sure if this feature is wanted.


kosack:
I still don't think this is necessary, since you can already pass `mode='a'` and it should work.  There are also many other parameters you can pass, like `filters=tables.Filters(complevel=5, complib='blosc')` if you want compression.  Perhaps we just need to update the documentation to be more explicit?

kosack:
see #690 - does this fix the problem?

calispac:
I just updated this PR to what was discussed and what is in this PR #690 . I can make the notebook a bit better to show what types are supported and how one could proceed to save different Containers and SubContainers.

kosack:
Also, all the tests seem to be failing (mainly because the first test, which writes out a file fails, so the others which read it also fail). Not sure why

calispac:
I think test first test was failing because `mode` was not set to `mode='w'`. By default pytables.open_file() sets `mode='r'`. See http://www.pytables.org/usersguide/libref/top_level.html?highlight=tables%20open_file#tables.open_file

I see that one can pass `root_uep` to this so I will try to make it work such that one can specify the location of the group

calispac:
I realized that in the reader `__init__()` one variable is called : 'table_name'. In fact this is not really a name but a location in the table.  Should we rename this variable?

calispac:
Concerning the Root Entry Point. By default `pytables` does not allow to write to a non existing group. e.g. : 

If one does this in an empty file:

```python
tables.open_file('my_group', root_uep='/a_non_existing_group_path')
```
an an `HDF5ExtError` is issued. See : https://www.pytables.org/usersguide/libref/file_class.html?highlight=file#tables.File.create_group

So my question is should we allow do as `pytables` does or should we allow to write to any location whether or not the group exists ?. In any case I think we can address this to a future PR.

PR #709, opened 2018-04-17
Cleanup

Sojourner92:
Various minor changes to ensure pep8 compatibility and improve our codacy score.


ctapipe/image/geometry_converter_astri.py
```diff
@@ -12,49 +12,49 @@ def astri_transformation_map():
     img_map = np.full([8 * 7, 8 * 7], -1, dtype=int)
 
     # Map values
-    img_map[0*8:1*8, 2*8:3*8] = np.arange(64).reshape([8,8])[::-1,:] + 34 * 64
```
kosack: I'm not sure this is actually easier to read, but I guess at least it stops pep8 from complaining

ctapipe/reco/ImPACT.py
```diff
@@ -479,7 +480,7 @@ def get_likelihood(self, source_x, source_y, core_x, core_y,
                                                self.spe,
                                                self.ped[tel_count])
             if np.any(prediction == np.inf):
-                print("inf found at ", self.type[tel_count], zenith,
+                print("inf found at ", self.tel_type[tel_count], zenith,
```
kosack: can you change this to use the logger, like `self.log.debug("info found at type=%s zenith=%.2f azimuth=%.2f energy=%.2f impact=%.2f x_max=%.2d", self.tel_type[tel_count], zenith, azimuth, energy, impact, x_max_bin)`

ctapipe/reco/hillas_intersection.py
```diff
@@ -220,7 +221,7 @@ def reconstruct_tilted(self, hillas_parameters, tel_x, tel_y,
 
         tx = np.zeros((len(tel_x), 2))
         ty = np.zeros((len(tel_y), 2))
-        for i in range(len(tel_x)):
+        for i in enumerate(tel_x):
```
kosack: I don't think this is correct.  Should be something like:

```py3
for i, _ in enumerate(tel_x):
    ...
```
since enumerate returns the index and the value as a tuple

ctapipe/image/geometry_converter_astri.py
dneise: In the "other recommendations" part of the PEP8 they actually say this is **not** recommended.

https://www.python.org/dev/peps/pep-0008/#other-recommendations

> If operators with different priorities are used, consider adding whitespace around the operators with the lowest priority(ies). 

Here `:` has lower priority than `*`. I assume some pep8 tools might not implement this correctly .... if this is the case for the tools we use here, we maybe wanna consider using different tools or to configure the tools we use differently?

Sojourner92: I was just using the pep8 command line tool.

kosack: This whole file is a bit of a hack anyhow, and will break if we change camera geometries (which is why I like the more generic geometry_converter better anyhow, so in any case it will disappear at some point and we shouldn't worry too much.  So Im' ok with at least stopping pep8 from complaining.

dneise: Oh I didn't know this ... anyway .. just wanted to mention one could write it like this .... as well:
```python

foo = [
    34, 35, 36,
    29, 30, 31, 32, 33,
    22, 23, 24, 25, 26, 27, 28,
    15, 16, 17, 18, 19, 20, 21,
    8, 9, 10, 11, 12, 13, 14,
    3, 4, 5, 6, 7]
bar = [
    0, 0, 0,
    1, 1, 1, 1, 1,
    2, 2, 2, 2, 2, 2, 2,
    3, 3, 3, 3, 3, 3, 3,
    4, 4, 4, 4, 4, 4, 4,
    5, 5, 5, 5, 5]
bar = [
    2, 3, 4,
    1, 2, 3, 4, 5,
    0, 1, 2, 3, 4, 5, 6,
    0, 1, 2, 3, 4, 5, 6,
    0, 1, 2, 3, 4, 5, 6,
    1, 2, 3, 4, 5
]
for a, b, c in zip(foo, bar, baz):
    img_map[
        a * 8:a+1 * 8,
        b * 8:b+1 * 8
    ] = np.arange(64).reshape([8, 8])[::-1, :] + c * 64
```

Issue #705, opened 2018-04-12
Docs version unknown


cdeil:
@kosack - Why is your docs version unknown?

![screen shot 2018-04-12 at 10 44 48](https://user-images.githubusercontent.com/852409/38666163-97bcb4cc-3e3e-11e8-9c43-31a187f00bc6.png)

kosack:
that's strange!  I'm not sure - it used to work correctly. Hopefully it will start working again when the next tagged release is made (hopefully this week)

kosack:
No, that was removed years ago, and it's worked fine since then.  We replaced it with a custom system that detects and sets the version based on the latest git tag.  On local copies the version gets set correctly, so it's something to do with the travis build. Not much has changed there, but it could be an updated dependency broke something. This was certainly working on the automatic docs as of a month ago. 

for example, i get:

```
ctapipe-info --version

*** ctapipe version info ***

version: 0.5.2.post509+git02986f2
```

And it works when I build the sphinx docs myself. 

![image](https://user-images.githubusercontent.com/11677812/38669050-a4a9830c-3e45-11e8-9af6-42e520a4dc67.png)

STSpencer:
Is this related to some of the travis CI tests failing? The most recent pull requests seem to be failing with the error:
The command "travis-sphinx -v --outdir=docbuild build --source=docs/" exited with 1.

thomasgas:
The docs displays the version nicely, so this is not an issue anymore.

PR #711, opened 2018-04-18
WIP: fix missing version number in generated docs

kosack:
As seen in #705 :

For some reason the docs generated by travis have the version set to "unknown", which means the version number is not being found from the git tags (normally it is taken from `git describe --tags --abbrev=7`).   This used to work correctly, so it needs some debugging...


kosack:
The problem seems to be also that the branch name is not known, which is strange:

```
git branch
* (HEAD detached at d3f9dcf)
  master
```

I'm not enough of an expert on how Travis checks out PRs, but I would have thought it would have fetched the remote branch.

This all may be fixed when I add a new tagged release, however, since it worked before the last one...

PR #714, opened 2018-04-20
Bokeh plotter

watsonjj:
Created an interactive event plotter that utilises the bokeh package (https://github.com/bokeh/bokeh)

This plotter shows both the camera and the waveform. It allows the selection of events, pixels, and time slices, and also allows the viewing and configuration of various calibration stages.

All EventSources are compatible with it, and all geometries are supported. However, all pixels are displayed as squares at the moment.

This tool requires a bokeh server to run:
`bokeh serve ctapipe/tools/bokeh/file_viewer`

To pass arguments to the tool:

`bokeh serve ctapipe/tools/bokeh/file_viewer --args -f /Volumes/gct-jason/data/170330/onsky-mrk501/Run05477_r1.tio`


watsonjj:
eventdisplayer

An animation of the file_viewer (some updates have been made since this gif's creation)

MaxNoe:
Looks good. Will checkout the code later. :+1:

ctapipe/tools/bokeh/file_viewer/main.py
```diff
@@ -392,5 +392,13 @@ def on_dl1_widget_change(self, _, __, ___):
...
 
 
-exe = BokehFileViewer()
-exe.run()
+def main():
+    exe = BokehFileViewer()
+    exe.run()
+
+
+if 'bk_script' in __name__:
```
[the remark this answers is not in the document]
watsonjj: To run a bokeh server you have to execute `bokeh serve` on the script instead of `python`. When you do this, `__name__` equals something like "bk_script_dc88af7ef51b4e0d85854896dd662ed0", where the third term seems to be random.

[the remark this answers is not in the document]
watsonjj: However, there may be a way to avoid having to use `bokeh serve` if I can understand how to utilise this: http://bokeh.pydata.org/en/latest/docs/user_guide/server.html#embedding-bokeh-server-as-a-library

ctapipe/utils/rgbtohex_c.cc
```diff
@@ -0,0 +1,18 @@
+/*
+C extension to convert and rgb array into a array of chars containing the
+hexidecimal string equivalent.
+*/
+
+#include <iostream>
+#include <stdint.h>
+#include <stdio.h>
+
+extern "C" void rgbtohex(const uint8_t* rgb, size_t n_pix, char* hex)
```
dneise: TL;DR: I believe in this particular case the C-extension is **not necessary**.

I believe we should avoid C-extensions as long as possible and I assume you believe so too (as opposed to writing C-extension just for fun). So I assume you had no chance to avoid this, since with pure Python this was performing horrible. In such a case I believe C-extensions are fine.

Still, I personally consider writing a C-extension so exceptional that a little discussion of it might not be too annoying. I would have liked the author to present maybe some numbers in a comment or in the PR-discussion to backup the decision to go for a C-extension. 

Since these numbers were missing, I tried to understand myself what was so slow here. For this I created a Python implementation of what I understood the C-implementation to do and profiled it by hand (using simply `%timeit` in my ipython session).

The result was a second implementation of `intensity_to_hex` which is simply named `intensity_to_hex2` which is pure python and happens to be even faster than the version using the C-extension. 

----

Here is the comparison:
```
In [1]: from rgbtohex import *

In [2]: array = np.random.rand(10000)

In [3]: assert (intensity_to_hex(array) == intensity_to_hex2(array)).all()

In [4]: %timeit intensity_to_hex(array)
2.99 ms ± 13.7 µs per loop (mean ± std. dev. of 7 runs, 100 loops each)

In [5]: %timeit intensity_to_hex2(array)
1.75 ms ± 10.6 µs per loop (mean ± std. dev. of 7 runs, 1000 loops each)
```

This is the original `intensity_to_hex` (rewritten a bit and added the runtimes per line):
```python
def intensity_to_hex(array, minval=None, maxval=None):
    hex_ = np.empty((array.size, 8), dtype='S1')   #   <1 us
    rgb = intensity_to_rgb(array, minval, maxval)  #  200 us
    rgbtohex(rgb, array.size, hex_)                # 1300 us
    return hex_.view('S8').astype('U8')[:, 0]      # 1300 us
```

This is the other variant:
```python
def intensity_to_hex2(array, minval=None, maxval=None):
    import codecs
    hex_ = np.zeros((array.size, 8), dtype='B')   #    3 us
    rgb = intensity_to_rgb(array, minval, maxval) #  200 us

    hex_encoded = codecs.encode(rgb, 'hex')       #   39 us
    bytes_ = np.frombuffer(hex_encoded, 'B')      #   <1 us
    bytes_2d = bytes_.reshape(-1, 8)              #   <1 us
    hex_[:, 0] = ord('#')                         #    4 us
    hex_[:, 1:7] = bytes_2d[:, 0:6]               #   48 us

    return hex_.view('S8').astype('U8')[:, 0]     # 1300 us
```

watsonjj: This is great! 

Yeah, I couldn't understand how to do this efficiently in Python (I had to use a for loop), so I resorted to the C extension. I love your solution. Thanks!

dneise:
I think this looks very nice. 

It seems there are quite some viewers popping up here and there these days, which I believe might be necessary, but I also believe viewers are often quite heavy in terms of LOCs and therefore they might be considered expensive in terms of maintenance. I believe ctapipe has 15k lines of code and with 1.7k lines, this PR adds a whopping 10% on top. 

I do not think that adding a lot of LOCs alone is a bad thing by itself, but let's assume for a moment that in 6 month from now we realize that of the 3 different existing viewers this particular one is not used by too many people... Then I personally would consider removing it maybe... or not if it turns out that is simply works without any maintenance.

That being said, I have no real clue of Bokeh, but I always wanted to learn it :-D

ctapipe/plotting/bokeh_event_viewer.py
```diff
@@ -0,0 +1,489 @@
...
+                else:
+                    self.image = samples[c, :, time]
+                self.fig.title.text = 'DL0 Slice (T = {})'.format(time)
+            elif v == 'dl1':
+                self.image = e.dl1.tel[t].image[c, :]
+                self.fig.title.text = 'DL1 Image'
+            elif v == 'peakpos':
+                peakpos = e.dl1.tel[t].peakpos
+                if peakpos is None:
+                    self.image = np.zeros(self.image.shape)
```
dneise: For the other cases you say:
```
if samples is None:
    self.image = None
```

but for peakpos, there is an exceptional handling of this case. Could you explain the difference?

mackaiver:
This looks really nice. Good Job. I'm just wondering whether this might be better located in a standalone package outside of ctapipe?

kosack:
*re @dneise* - Such a viewer could be even more generic, since Containers can be looped over and accessed by field name string.  One could  provide a dropdown with the base container name, telescope name, and field name which would automatically be "r0, r1, dl0, dl1...", "1,2,3,4...", "image", "waveform", or even simply search recursively for all fields in any subcontainer that have the same geometry as the image, or something like that.  Something to think about for a future feature. 

*re where it should go*:  I'm also not sure about whether this should be in ctapipe, or wether there should have a separate "ctapipe-viewers" repo.  It would take some time to set another repo up, but might make it more clean and easier to maintain.  Its a similar question for the "flow" module, which is a bit out of date and could be easily separated.  Perhaps for now, it's ok to have it in the main repo and we can separate things later?  


*Another comment*: I would suggest at the very least adding a top-level `Tool` for this in the `ctapipe/tools/` that can be used by running `ctapipe-event-viewer --input <filename>` or something like that, so that the code is trivially usable by end users. Otherwise, it will just be forgotten. That's a tool that would be quite nice and which we are missing so far.

watsonjj:
@dneise I like that "dream" too. I have incorporated the updates you have suggested, it should help make the utilisation of an automatically generated "view_options" easier a future update.

dneise:
> it should help make the utilisation of an automatically generated "view_options" easier a future update.
that's great ... I hope so too

watsonjj:
I have included all the comments so far. The file viewer can now be ran with either:
`python ctapipe/tools/bokeh/file_viewer.py`
OR
`ctapipe-event-viewer`

Codacy is complaining about jinja2.Environment, but I tried the change it suggests and it breaks bokeh entirely.

If everyone is happy, I am ready for this PR to be merged.

dneise:
At the moment it seems to break when I feed a digicampipe fits file into it:

For testing I use the `example_10_evts.000.fits.fz` one can fine here:
<https://github.com/cta-sst-1m/protozfitsreader/tree/master/protozfits/tests/resources>
```
(cta-dev) dneise@lair:~/cta/ctapipe$ ctapipe-event-viewer -f /path/to/example_10_evts.000.fits.fz
INFO [BokehFileViewer] (tool/initialize): ctapipe version 0.5.3.post117+git31c39a2
INFO [BokehFileViewer] (tool/run): Starting: BokehFileViewer
INFO: Obtaining SST1MEventSource from EventSourceFactory [BokehFileViewer.EventSourceFactory._product]
INFO: INPUT PATH = /home/dneise/sst_new/digicampipe/digicampipe/tests/resources/example_10_evts.000.fits.fz [BokehFileViewer.SST1MEventSource.__init__]
INFO: Obtaining NeighbourPeakIntegrator from ChargeExtractorFactory [BokehFileViewer.ChargeExtractorFactory._product]
INFO: Obtaining NullWaveformCleaner from WaveformCleanerFactory [BokehFileViewer.WaveformCleanerFactory._product]
INFO: Obtaining NullR1Calibrator from CameraR1CalibratorFactory [BokehFileViewer.CameraR1CalibratorFactory._product]
INFO: Using NullR1Calibrator, if event source is at the R0 level, then r1 samples will equal r0 samples [BokehFileViewer.NullR1Calibrator.__init__]
INFO: Applying no data volume reduction in the conversion from R1 to DL0 [BokehFileViewer.CameraDL0Reducer.__init__]
WARNING: Seeking to event by looping through events... (potentially long process) [BokehFileViewer.EventSeeker.__getitem__]
  File "/home/dneise/anaconda3-5.1.0/envs/cta-dev/bin/ctapipe-event-viewer", line 11, in <module>
    load_entry_point('ctapipe', 'console_scripts', 'ctapipe-event-viewer')()
  File "/home/dneise/cta/ctapipe/ctapipe/tools/bokeh/file_viewer.py", line 414, in main
    exe.run()
  File "/home/dneise/cta/ctapipe/ctapipe/core/tool.py", line 167, in run
    self.start()
  File "/home/dneise/cta/ctapipe/ctapipe/tools/bokeh/file_viewer.py", line 134, in start
    self.event_index = 0
  File "/home/dneise/cta/ctapipe/ctapipe/tools/bokeh/file_viewer.py", line 165, in event_index
    self.event = self.seeker[val]
  File "/home/dneise/cta/ctapipe/ctapipe/tools/bokeh/file_viewer.py", line 218, in event
    self.viewer.event = val
  File "/home/dneise/cta/ctapipe/ctapipe/plotting/bokeh_event_viewer.py", line 423, in event
    sub.change_event(val, self.telid)
  File "/home/dneise/cta/ctapipe/ctapipe/plotting/bokeh_event_viewer.py", line 96, in change_event
    self.event = event
  File "/home/dneise/cta/ctapipe/ctapipe/plotting/bokeh_event_viewer.py", line 89, in event
    self._update_geometry()
  File "/home/dneise/cta/ctapipe/ctapipe/plotting/bokeh_event_viewer.py", line 74, in _update_geometry
    self.geom = e.inst.subarray.tel[t].camera
KeyError: 1
```

watsonjj:
> At the moment it seems to break when I feed a digicampipe fits file into it

It appears you aren't filling the subarray container in sst1meventsource.py. So not a fault of the file viewer :)

watsonjj:
I apologise, I wrote that comment in a rush before I caught a bus so that you weren't left hanging without a response (after downloading the file and the protozfits software to double check the problem).

I was not aware this container is deprecated. To my understanding it is the only method available to obtain the pixel positions from the event container.

Yes, I should refer to us all as owners of the code. However, I (and the majority of contributors) aren't familiar with the organisation of SST1M, and were unqualified to make the sst1meventsource, and are subsequently unqualified to figure out the best method to supply the pixel positions that accompany the data. In the TargetIOEventSource, we utilise the TargetCalib library which has knowledge of the pixel positions of our camera.

Again, sorry for the careless response.

dneise:
Thank you for your kind understanding...

watsonjj:
Can this PR be accepted yet?

The only failures are those present in the master.

kosack:
yes, i think this is fine to merge once I get the tests working again. I'll do it shortly.

PR #715, opened 2018-04-21
Refactor of "reductor" to "reducer"

watsonjj:
In preparation for contributions of algorithms, I thought it best to refactor the name


kosack:
I agree!  I always wondered why the name was non-standard english 😄  Though I suppose reductor is still a word, but usually associated with chemical reductions.

PR #720, opened 2018-04-24
WIP: put TelescopeDescription input `inst.subarray`

dneise:
@watsonjj spotted a bug in the SST1MEventSource.

The problem came up in connection with the new bokeh plotter in #714.

As far as I understood, EventSources are supposed to fill also the `inst` Field, which is not done by the SST1MEventSource.

### Unit test still missing

### DIFF
--- a/ctapipe/io/sst1meventsource.py
+++ b/ctapipe/io/sst1meventsource.py
@@ -7,6 +7,7 @@ Needs protozfits v0.44.3 from github.com/cta-sst-1m/protozfitsreader
 import numpy as np
 from .eventsource import EventSource
 from .containers import SST1MDataContainer
+from ..instrument import TelescopeDescription
 
 __all__ = ['SST1MEventSource']
 
@@ -31,6 +32,7 @@ class SST1MEventSource(EventSource):
             data.count = count
             data.sst1m.fill_from_zfile_event(event, self._pixel_sort_ids)
             self.fill_R0Container_from_zfile_event(data.r0, event)
+            data.inst.subarray.tels[0] = TelescopeDescription.from_name('SST-1M', 'DigiCam')
             yield data
 
 


dneise:
Thanks for your comment.

I am however a bit puzzled by: "ctapipe contains no information (yet) about the pixel positions".
If I try this:
```
In [1]: import matplotlib.pyplot as plt
In [2]: from ctapipe.instrument import TelescopeDescription
In [3]: tel_desc = TelescopeDescription.from_name('SST-1M', 'DigiCam')
In [4]: plt.plot(-tel_desc.camera.pix_y, -tel_desc.camera.pix_x, 'd')
In [5]: plt.savefig('digicam_x_is_minus_y_and_y_is_minus_x.png')
```

I get this result ...

digicam_x_is_minus_y_and_y_is_minus_x

....which looks very much like the camera view, which is created by the digicam viewer which currently exists in digicampipe: <https://github.com/cta-sst-1m/digicampipe#example-viewer>.

However I had to do play around with the axes like this `x=-pix_y, y=-pix_x` to get *exactly* the same orientation. 

So ... may it be your information about missing pixel positions is outdated?

watsonjj:
Ah! Very true, my knowledge is outdated.

I assume these are the pixel positions extracted from MC. The only caution I have then is that the pixel ordering of MC might not be the same as the pixel ordering inside SST1M (it certainly isn't for CHEC), so you should check that.

dneise:
Yes I have to check that

dneise:
@watsonjj  you are right, just looking at the first 3 pixels shows clearly, that in digicampipe and in ctapipe the `pix_id` does not mean the same, but it seems also the area is a bit different. So maybe the positions are not only in a different order, but also shifted a bit ... I have to investigate that.

I am not sure how to proceed here. In `CameraGeometry.from_name` it is possible to specify a version. So I could simply create a new version of the `Digicam.geomcam.fits` file and read that one instead of the old one...

Alternatively I could just try to find the a mapping between the pixel ids, so that the result is closest to the result I would get in digicampipe... 

I am confused. 


https://github.com/cta-observatory/ctapipe/blob/c7d73cdf3c8b8f840523e58b7900ec263dbe1bad/ctapipe/instrument/camera.py#L209

```python
from digicampipe.utils import DigiCam
digicampipe_geometry = DigiCam.geometry.to_table()
ctapipe_geometry = Table.read('DigiCam.camgeom.fits')
digicampipe_geometry[:3]
```

&lt;Table length=3&gt;
<table id="table139967082579168" class="table-striped table-bordered table-condensed">
<thead><tr><th>pix_id</th><th>pix_x</th><th>pix_y</th><th>pix_area</th></tr></thead>
<thead><tr><th></th><th>mm</th><th>mm</th><th></th></tr></thead>
<thead><tr><th>int64</th><th>float64</th><th>float64</th><th>float64</th></tr></thead>
<tr><td>0</td><td>12.15</td><td>-498.05</td><td>400.0</td></tr>
<tr><td>1</td><td>-24.3</td><td>-477.01</td><td>400.0</td></tr>
<tr><td>2</td><td>0.0</td><td>-477.01</td><td>400.0</td></tr>
</table>

```python
ctapipe_geometry[:3]
```

&lt;Table length=3&gt;
<table id="table139967082653792" class="table-striped table-bordered table-condensed">
<thead><tr><th>pix_id</th><th>pix_x</th><th>pix_y</th><th>pix_area</th></tr></thead>
<thead><tr><th></th><th>m</th><th>m</th><th>m2</th></tr></thead>
<thead><tr><th>int64</th><th>float64</th><th>float64</th><th>float64</th></tr></thead>
<tr><td>0</td><td>0.4837000072</td><td>-0.0118000004441</td><td>0.000482051264597</td></tr>
<tr><td>1</td><td>0.463270008564</td><td>-0.0472000017762</td><td>0.000482051264597</td></tr>
<tr><td>2</td><td>0.463270008564</td><td>-0.0236000008881</td><td>0.000482051264597</td></tr>
</table>

dneise:
I assume there is some historical evolution involved here which predates my time in SST1M, I will have to check this.

dneise:
I think I made a mistake regarding the ids ... it seems the ids are in fact pretty much okay (after some shuffling around with the axes) ... however the positions are not the same. 
I've opened and issue in the SST1M repo, I believe is followed by people who might know more about it.
https://github.com/cta-sst-1m/CTS/issues/8

ctapipe/io/sst1meventsource.py
```diff
@@ -31,6 +32,10 @@ def _generator(self):
             data.count = count
             data.sst1m.fill_from_zfile_event(event, self._pixel_sort_ids)
             self.fill_R0Container_from_zfile_event(data.r0, event)
+            data.inst.subarray.tels[0] = TelescopeDescription.from_name(
```
kosack: it might be better to pass this to the SubarrayDescription constructor instead, but this should work for existing code.  If in the future we pre-compute some things in the constructor, however, it will fail. 


e.g. this may be safer: 
```py3

event.inst.subarray = SubarrayDescription(
    name="SST1MTestBenchArray", 
    tel_positions={0: [0,0]*u.m}, 
    tel_descriptions={0:TelescopeDescription.from_name("SST-1M","DigiCam")}
)
```
or similar

kosack: Also, using `TelescopeDescription.from_name()` of course assumes the `DigiCam.camgeom.fits.gz` and `optics.ecsv.txt`  files in your `CTAPIPE_SVC_PATH` (or ctapipe-extra package) correspond to the input file your give (i.e. the pixel numbers are correct, etc).  Since currently that was generated from Monte-Carlo, it may not be exactly right, so just keep that in mind. You might add a check in the code to make sure at least the number of pixels, etc, makes sense in case you load a file with a slightly modified camera geom.

kosack:
> So ... may it be your information about missing pixel positions is outdated?

Not outdated, but there is no guarantee the MC pixel positions correspond to your testbench camera pixels.  The MCs, as you found out use a very strange coordinate system where x and y are swapped for some reason, with North along the y-axis, and we don't yet support conversion between this "MC system" and the "EngineeringSystem", which is probably more like what you get if you look at the front of the camera, rather than through it from the back as in the MCs. 

So the point is if you use SST1m data and the default camera loaded up, which was generated from MCs, you may have a mismatch in pixel ids and camera rotation.

MaxNoe:
> I assume these are the pixel positions extracted from MC. The only caution I have then is that the pixel ordering of MC might not be the same as the pixel ordering inside SST1M (it certainly isn't for CHEC), so you should check that.

Good to hear that we (FACT) are not the only Collaboration with this :D

Issue #721, opened 2018-04-24
Direction Reconstruction Problems on Diffuse Gammas


kbruegge:
Hello Everyone. 

I'm having trouble reconstructing the direction on diffuse gammas.
Here is the code I use. I basically took it from the example scripts. The code below is executed for each event.

```
    reco = HillasReconstructor()
    
    features = {}
    params = {}
    pointing_azimuth = {}
    pointing_altitude = {}
    
    for telescope_id, dl1 in event.dl1.tel.items():
        camera = event.inst.subarray.tels[telescope_id].camera

        mask = tailcuts_clean(camera, dl1.image[0], boundary_thresh=5, picture_thresh=10)
        telescope_type_name = event.inst.subarray.tels[telescope_id].optics.tel_type
        
        image = dl1.image[0].copy()
        image[~mask] = 0
        
        h = hillas_parameters(
            camera,
            image,
            container=True
        )
        
        moments = MomentParameters(size=h.intensity, cen_x=h.x, cen_y=h.y, length=h.length, width=h.width, psi=h.psi )
        params[telescope_id] = moments
        pointing_azimuth[telescope_id] = event.mc.tel[telescope_id].azimuth_raw * u.rad
        pointing_altitude[telescope_id] = ((np.pi/2) - event.mc.tel[telescope_id].altitude_raw )* u.rad # this is weird to say the least. 
        

    reconstruction = reco.predict(params, event.inst, pointing_azimuth, pointing_altitude)
    d = {
        'mc_alt': event.mc.alt.rad,
        'mc_az': event.mc.az.rad,
        'mc_core_x': event.mc.core_x.value,
        'mc_core_y': event.mc.core_y.value,
        'mc_height_first_interaction': event.mc.h_first_int.value,
        'alt_prediction': reconstruction.alt.si.value,
        'az_prediction': reconstruction.az.si.value,
        'core_x_prediction': reconstruction.core_x.value,
        'core_y_prediction': reconstruction.core_y.value,
        'h_max_prediction': reconstruction.h_max.value,
    }
    return d 
```

This seems to work fine on point-source gammas. However for diffuse gammas it fails. 
See below a plot which shows the distribution of the distances (in degrees) between the `mc_az, mc_alt` and `az_prediction, alt_prediction` (to be exact the histogram shows `astropy.coordinates.angle_utilities.angular_separation`)

[...]

PR #723, opened 2018-04-24
Improve ArrayDisplay and SubarrayDescription

kosack:
The `visualization.ArrayDisplay` code was quite out of date, and mostly not useful. This PR simplifies it, makes it use the `SubarrayDescription` correctly, and adds some useful functionality.  It so far will break the `plotting.ArrayPlotter`, but that seems to be an unnecessary class and will be replaced with basic ArrayDisplay functionality.

ArrayDisplay Changes:
-----------------------

- Three things can be controlled: the telescope border color, the telescope size, and the fill color:
  - telescope sizes are back to being set to the mirror radius by default (that functionality seems to have disappeared in a previous refactoring for some reason), but you can set them to all the same size if you like. 
  - telescope borders are colored by telescope type by default
  - the internal fill color can be controlled by setting the `values` a-tribute, similar to `CameraDisplay.image`. 
- You now can overlay a vector field (e.g. for HillasParameters) in a way that updates very quickly (internally uses a `matplotlib.quiver` object).  This is controlled by the `set_vector_uv()`, `set_vector_rho_phi()` or `set_vector_hillas()` functions, all of which provide different ways to set the vector angle and length
- added `examples/array_display.py` and `examples/plot_array_hillas.py`, the latter plots the array, the core position, and the hillas ellipses for each event 

SubarrayDescription changes:
------------------------------

- added some helper functions to map tel_ids to indices, etc.
- added `telescope_types`, `camera_types`, and `optics_types` atributes, that give back lists of those quantities, helpful for selecting telescopes.
- added `get_tel_ids_for_type()` helper, that returns a list of tel_ids for a given telescope type.

Some things that need improvement:
-------------------------------------

- [ ] need support for different coordinate `Frames`. E.g. should be able to give ArrayDisplay a Frame, and have it convert the SubarrayDescription telescope positions automatically, so that the result is e.g. in the `TiltedFrame`

Some problems possible unrelated to this PR, but that should be looked into:
-----------------------------------------------------------------------------

- [ ] not sure right now why the Hillas vectors seem to intersect nicely at the core position in some MC files I tried, and not in others where the core position is far away form the intersection point (could be due to the coordinate frame, but it seems larger than the zenith angle offset). See e.g. the plot below:
- [ ] related to previous question, it seems in some MC files, the vector phi angle is 180 degs off. No idea why that would happen - it should always correspond to the assymmetry direction, I guess. Could simply plot the vectors as lines (2 vectors in opposite directions)

A LaPalmaRefSim MC:
image

A standard Prod3b MC:
image


mackaiver:
> not sure right now why the Hillas vectors seem to intersect nicely at the core position in some MC files I tried, and not in others where the core position is far away form the intersection point (could be due to the coordinate frame, but it seems larger than the zenith angle offset). See e.g. the plot below:
> related to previous question, it seems in some MC files, the vector phi angle is 180 degs off. No idea why that would happen - it should always correspond to the assymmetry direction, I guess. Could simply plot the vectors as lines (2 vectors in opposite directions). Could also be coordinates - seems to be different for North and South arrays.

Could this be related to #721 ?

MaxNoe:
I had a nice laugh when I read:

> `SkyCoord` in the `GroundFrame`

PR #724, opened 2018-04-24
add script to execute all the ipynbs in the examples/notebooks folder

dneise:
reading the discussion in #716 made me want to test our jupyter notebooks.

In this PR I make a first proposal, I do not like it myself too much, but I thought maybe this sparks a discussion. 

What I propose to do is, convert the ipynb files to python scripts and then simply execute them and see if the return value is zero or not.

In this first version I swallow all the stderr and stdout, since all of the scripts are very verbose. And I saw no nice way to get some overview. 

I assume one could still catch the stderr and in case the return value is not zero, I could still try to find the last traceback and print it ...

The main problem is, `%matplotlib inline`, which is not working in a ipython session. So all our ipynbs which to `%matplotlib inline` fail this test of mine, even when they are perfectly valid.

----

A much better alternative might be this:  https://github.com/zonca/pytest-ipynb
but it needs us to rewrite our notebooks a little... 

Anyway I'm curious what you think?


dneise:
Here one can see the output:
https://travis-ci.org/cta-observatory/ctapipe/jobs/370764934#L1921-L1999

MaxNoe:
This is great! Will hopefully preventlying examples/docs. However I think we should not silence `stderr` just stdout.

MaxNoe:
You could also execute directly using:

```
jupyter nbconvert --to notebook --execute mynotebook.ipynb
```

dneise:
Thanks @MaxNoe ! I did not know about `--execute` and was unable to find it. The script is now much shorter and it can execute correctly all the plotting stuff.

dneise:
On my laptop I ran into this issue:
https://github.com/jupyter/notebook/issues/3407

and I've seen the same on travis. So I decided to follow the `fix` mentioned in that issue above and downgraded to tornad0 4.5.2.

watsonjj:
This is nice. Just a thought: Could you put this into a pytest file so its automatically ran by pytest anyway in travis. That would have the benefit of automatically indicating to the user locally if they need to change the notebooks for their changes. Would that work?

dneise:
> Just a thought: Could you put this into a pytest file so its automatically ran by pytest anyway in travis? That would have the benefit of automatically indicating to the user locally if they need to change the notebooks for their changes. Would that work?

I've got no idea yet, how to put this nicely into pytest... But you are right, I myself am sure I will forget to execute this script... I barely remember to `pytest` locally.

 I believe we should maybe try out these  <https://github.com/zonca/pytest-ipynb> at some point. This integrates nicely with pytest. But this means potentially modifying our notebooks, which might not be feasible at the moment.

mackaiver:
Hello.

I like the idea of testing notebooks. 👍 

There is also this little library which looks interesting to me: https://github.com/opengeophysics/testipynb I've never used it myself though.

dneise:
Okay ... this is I think the end from my side. I have not more time to spend on this matter. So I would like to give it into the hands of the community now.

I added a way to skip tests of notebooks, which are known to fail (similar to pytests xfail).

So now I hope the build will pass even with the two notebooks which are still failing.
If the build passes .. I hope we can maybe merge this even though there are still a huge number of features missing here ...

dneise:
okay tests are passing and output seems reasonable to me:
https://travis-ci.org/cta-observatory/ctapipe/jobs/370960158#L1956-L1967

dneise:
I have no idea what codacy is whining about again ...

kosack:
> pytest file so its automatically ran by pytest anyway in travis

we can just add it to .travis anyhow - if the script returns a non-zero value, it should get marked as failed.

kosack:
> I have no idea what codacy is whining about again ...

It thinks ctapipe is a python2.7 project, so always complains about syntax with error 999... I can't find anywhere in their docs where you can set the python version, except one place where you set it for pyflakes, and I did that, but nothing changed. 

 So far, I've been disappointed in landscape.io (constantly failed to process pull requests), and codacy is better but has these problems.   ANyhow, this PR looks fine

PR #726, opened 2018-04-25
Container cleanup

watsonjj:


kosack:
Note, some of this will need updating when I merge in #683 (which still needs work, however)

PR #729, opened 2018-04-29
Changes to SST1MEventSource

watsonjj:


watsonjj:
ctapipe/io/tests/test_sst1meventsource.py should pass following #730

ctapipe/io/sst1meventsource.py
```diff
@@ -7,7 +7,7 @@
 import numpy as np
 from .eventsource import EventSource
 from .containers import SST1MDataContainer
-from ..instrument import TelescopeDescription
```
dneise: did this not work?

watsonjj: Sorry, I imagine it did. My previous PR was before your PR and in merging I lost that import, and pycharm's autofill chose the other import approach

ctapipe/io/sst1meventsource.py
```diff
@@ -18,38 +18,76 @@ def __init__(self, config=None, tool=None, **kwargs):
...
-            self.fill_R0Container_from_zfile_event(data.r0, event)
-            data.inst.subarray.tels[0] = TelescopeDescription.from_name(
-                optics_name='SST-1M', 
-                camera_name='DigiCam'
-            )
-            yield data
 
+            data.inst.subarray.tels[telid] = self._tel_desc
+
+            # R0Container
```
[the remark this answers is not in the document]
dneise: This is why I said "nice example".  This is exactly what happens ... it is accurate for some time .. until it is not ... often one simply forgets to update comments when code changes ... or the comments are simply "too far up" .. so one does not even see them when updating code. 

Therefore I rather like to the method of "breaking out little functions" ... instead of commenting. (These functions might be easier to unit test, but this is another point) ... however, for some reason I seem to realize, when I put "r1" stuff into a function called "fill_R0_stuff" ... much better than I realize if I put "r1" stuff below a comment that said "# filling R0 stuff" ... 

But never mind ... this was me nit picking again... sorry.

ctapipe/io/sst1meventsource.py
dneise: interesting ... thx.

dneise:
I think this is very nice. Thank you very much

PR #730, opened 2018-04-29
Correct dl1.py for using inst.num_channels for the integration correction

watsonjj:


watsonjj:
This fix is valid for the current master branch, but will be revised for #683

PR #731, opened 2018-04-29
Changes to NectarCAMEventSource

watsonjj:


watsonjj:
Something odd seems to be happening when reading a NectarCAM file for a second time (see the error message that occurs in even the first commit, before any changes to the NectarCAMEventSource class). @dneise  Any idea why this error occurs (it happens inside protozfits/simple.py)

dneise:
I believe this is connected to an issue I ran into nearly a month ago .. but I was too lazy to even completely describe the issue. I am very sorry. 

https://github.com/cta-sst-1m/protozfitsreader/issues/25

kosack:
Can you merge in the master now and fix the conflicts? I merged #734, so perhaps that will resolve this

watsonjj:
A segfault is thrown when one of these events go through `ctapipe.utils.neighbour_sum.get_sum_array`, likely due to the dual channels. Its probably not worth fixing as there will be only one channel at that level when #683 is merged.

PR #734, opened 2018-04-30
install protozfits v0.44.5 and require it for nectarcam

dneise:
I believe this fixes the problems seen in #731. I hope it does not introduce new problems .. so let's wait for the tests to pass.


dneise:
Maybe I should add:

https://github.com/cta-sst-1m/protozfitsreader/pull/28

This is the PR in protozfitsreader which results in v0.44.5, there were in my opinion a few interesting questions raised by Jason and maybe others are interested as well the discussion over there.

PR #735, opened 2018-05-02
Removed `TooFewTelescopes` Exception class in `HillasReconstructor.py`

vuillaut:
Removed entirely the `TooFewTelescopes` Exception class in `HillasReconstructor.py` that was not called and replaced the `TooFewTelescopesException` (note that the name of the class was differing from the intended one) call by a simpler `Exception` with custom message.

I don't think a custom Exception class is required here. Otherwise I'd suggest:
```
class TooFewTelescopes(Exception):
    def __init__(self, hillas_dict):
        Exception.__init__(self, "need at least two telescopes, have {}".format(len(hillas_dict)))
```

and raising
```
TooFewTelescopes(hillas_dict)
```


vuillaut:
@dneise do you suggest I change it for `ValueError` ?

dneise:
Yes I believe as a general ruple the most specific exception possible should be chosen.

In this particular case, I assume having sometimes not enough telescopes is actually expected and does not indicate a user error.

So I suppose in an actual use case this function will be wrapoed into something like a try-except-log block, to ensure that the analysis of many events also succeeds when one event contains not enough telescopes. 

Now this except wants to catch this expected case not all cases where something could go wrong, say an IOError or so, so choosing a specific exception is important here.

Ah bye the way, did you see any uses of this function in the ctapipe codebase, where now an `except TooFewTelescopes` is dangling around, which needs to be adjusted as well?

dneise:
Ah and whatever you decide, I assume one would want to adjust the docstring of this function to avoid lying documentation.

dneise:
Ah, and the `__all__` list would also need to be adjusted

kosack:
I strongly disagree that a custom class is not needed - at the end, we want to keep track of all exception statistics, or have different branches when different exceptions are caught.

kosack:
Imagine a case where another exception is throwns (e.g. a real error in the code, or some bad input) - In normal analysis, I would want to catch TooFewTelescopeExeptions, maybe count them, but continue processing. However, I want to stop if some unknown/unexpected exception occurs.  With this change, that is not posssible!

```py3
for event in source:
    try: 
        reconstruct_event:
    except TooFewTelesopesException as ex:
        ...
        #contineu, since this is fine.

```

vuillaut:
@kosack I see, then I will create a custom exception class following the recommendations from @dneise and yourself and modify the Reconstructor accordingly.
Note: This Exception class could potentially be used in other reconstructions - so maybe it should be placed in another place than `HillasReconstructor.py`...

kosack:
yes, the exception should probably be in `ctapipe.reco.reco_algorithms.py` with the base class

vuillaut:
Hi.
So I basically just moved the exception class in `reco_algorithms.py` and called it in predict.

dneise:
@kosack  I believe the main point of this PR was to fix a typo. In the mean time this typo has been fixed in the master. So I assume this PR can be closed.

dneise:
The main topic of this PR was to fix a typo, which was fixed in the meantime

PR #736, opened 2018-05-03
Muon reco changes

rlopezcoto:
- Ring size container was previously created but not filled in the muon reco functions.

- As discussed during the meeting, added the possibility of analyzing muon rings without applying image cleaning. Set the cleaning option to True to make not to mess with the results people were getting in the past. This option, together with all those included at the beginning of muon_reco_functions.py should eventually be moved to an input card (@AMWMitchell are you planning to do it? I could help out if you want).


kosack:
Can you add a simple test case to the unit tests with this option disabled?

kosack:
Ok, I'll merge it now then, and we can leave refactoring the whole muon code to a later point

PR #741, opened 2018-05-16
Fix camera axis alignment in HillasReconstructor

mackaiver:
I know the coordinate transformations will be reworked soon but here it is anyways. This fixes the directional reconstruction. Now the reconstruction seems to work well on point-like as well as diffuse gammas.


kosack:
Is there a way we can write a simple unit test for this so we are sure we don't break it again?

PR #743, opened 2018-05-17
Fix broken build

kosack:
- Added more info to calibrated_data_exploration
- Removed some superfluous stuff that broke with latest Pandas (0.23) in provenance.ipynb (should fix master build failure)


kosack:
this fixed the notebook tests, but the build now fails due to travis-sphinx due to this error: https://github.com/Syntaf/travis-sphinx/issues/39

Issue #742, opened 2018-05-16
HillasParameterContainer should be filled with np.nan if higher moments are not calculated


MaxNoe:
0.0 are valid values for these variables, so should not be used in case no value was ever calculated.

PR #744, opened 2018-05-18
Change requested in #742: init Hillas skewness and kurtosis to NaN

STSpencer:
Change to Hillas parameter container such that higher order hillas paramaters (skewness and kurtosis) are initialized to nans, such that they take that value if they are not calculated (#742).


MaxNoe:
Hi.

Thanks for starting to work on this :+1: 

I think all values should be initialised with nan. And the algorithms in `image/hillas.py `fill the container with 0, just setting the default is not enough. 

I was planning to do this myself, do you want to continue?

STSpencer:
Yeah I'm happy to do it; I think that should get the allocations in the four Hillas parameters methods in hillas.py correct.

kosack:
Is this ready to be merged? or are you still making changes?

ctapipe/image/hillas.py
```diff
@@ -118,20 +118,25 @@ def hillas_parameters_1(geom: CameraGeometry, image):
     phi = np.arctan2(mean_y, mean_x)
 
     # Higher order moments
-    sk = cos_delta * (pix_x - mean_x) + sin_delta * (pix_y - mean_y)
+    if length == 0:
+        skewness = np.nan
```
kosack: it may be better to simply init these to nan at the top, not in an *if* branch, since then you need another unit test to test this branch (hence why codecov is complaining about this change).  Same for other places you added if-statements. 

Also comparing length to exactly 0 may not be a good idea, as it's a floating point quantity.

kosack:
Still has branches that are not tested in the if statements, e.g. there is no test where the inputs fail and a HillasParameterizationError is correctly returned.  However, since the Hillas code may change soon (e.g. #748), it's not so critical so I'll merge it.

PR #745, opened 2018-05-21
Trigger pixel reader

thomasarmstrong:
Small change to add trigger pixel id to R0 container. NOTE This requires new release of pyhessio where I have added the get_trig_pixels functionality, so do not merge until this happens...

Also up for discussion if it is needed, for me it is useful for performing studies on muon flagging in the camera.


kosack:
If pyhessio has been updated and it's version number bumped, i could create a new conda package for it, and you should then update the requirement in `setup.py` and `environment.yml` to ensure pyhessio has the right version.

PR #746, opened 2018-05-23
Add example script for a simple event writer

jjlk:
Hi @kosack ,
This PR adds an example script and a configuration file to save image information (e.g. moment) and event information (e.g. MC energy and id) in HDF5 format. I used the `ctapipe.core.Tool` class capabilities to handle the configuration file and the event processing (setup, start, finish). Some pre-selection cuts, wich are defined in the configuration file, are also applied using the `ctapipe.utils.CutFlow` class.

What do you think?

Afterwards, I plan to start from this script to make an other tool to write DL1+ data that could be used to fed a regressor and a classifier for energy regression and event discrimination, respectively. The goal will be to launch the script on the GRID with @bregeon scripts.

Cheers

### DIFF
--- a/examples/simple_event_writer.json
+++ b/examples/simple_event_writer.json
@@ -1,0 +1,9 @@
+{
+  "Analysis": {
+    "allowed_tels": [1, 2, 3, 4]
+  },
+  "Preselect": {
+    "image_amplitude": {"min": 50},
+    "n_pixel": {"min": 3}
+  }
+}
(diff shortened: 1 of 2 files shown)


kosack:
Thanks! Once we have some advanced classes merged in like the EventPreparer (which will allow you to select image cleaning thresholds per telescope type, etc), we can eventually make this more powerful and move it to ctapipe.tools as an official tool, perhaps.

PR #748, opened 2018-05-25
Implement hillas features usen eigh

MaxNoe:
The values for length slightly disagree (2 %)

I think this version is much more readable then the 1-4 versions


kosack:
I like it - it definitely is more readable.

kosack:
Do you know why the length is a bit different? Rounding error?  If so, you could just increase the threshold in the test. It's a bit strange, since length is such a low-order moment that it shouldn't be a problem

MaxNoe:
> Do you know why the length is a bit different?

Yes, i think it is the question of sample variance vs. variance. So the question of `1/N` vs `1/(N - 1)`. I kann set `ddof=0`, than it should be the same

MaxNoe:
Yes, with `ddof=0` the results are the same as  the other methods. However, with `ddof=1` the results should be better for small showers.

PR #750, opened 2018-06-01
cleaned up version of LST_reader from franca cassol

dneise:
This is just a cleaned up version of https://github.com/cta-observatory/ctapipe/pull/749 from @FrancaCassol

### DIFF
--- a/ctapipe/io/containers.py
+++ b/ctapipe/io/containers.py
@@ -23,6 +23,10 @@ __all__ = ['InstrumentContainer',
            'TargetIOCameraContainer',
            'SST1MContainer',
            'SST1MCameraContainer',
+           'LSTContainer',
+           'LSTCameraContainer',
+           'NectarCAMContainer',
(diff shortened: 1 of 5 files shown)


dneise:
I am not sure, if I managed to correctly cherry-pick all the commits. So this is basically a test.

I did: 

    git cherry-pick --allow-empty --strategy-option=ours 90f77fb732434c0c0ad36fd3d531a33590c0093f^..e118351f37791bb85610a36fb48ef7be3a715a59

Then I got an error message, since there is a merge commit in the range of these commits and I seem to be unable to get this correctly done ... but still ... after this command I had (seemingly) all the modifications which franca had done .. so I wanted to give it a try.

dneise:
Nope ... this is not identical to what franca has done. The Codacy report clearly shows that this is an older state, so I was missing some commits. 
I am going to close this PR now and leave the field to people who know more about cherry picking.

PR #751, opened 2018-06-01
trying if ctapipe also works with py3.7

dneise:
I wanted to see if ctapipe tests pass with py3.7 

I think supporting also py3.7 was once mentioned in order to have access to the new Dataclasses. 
Maybe supporting py3.7 is nothing that we can get quickly, but I thought it is still worth it to know how far we are away and what the problems might be.

However I do not consider this PR somehow high priority.


kosack:
Travis is failing here only because TravisCI doesn't come with a python 3.7 distribution, but we don't actually use Travis's python anyhow (we immediately overwrite it with anacondas's version in the script).  I wonder if there is a nice way to avoid having travis try to install python as well?

dneise:
Ah, I guess something like this might work. That actually allows you to test also on OSX, if you like. 

https://github.com/cta-sst-1m/protozfitsreader/blob/931af6fc45dabb0aa901cba5a39826a7c3448cc3/.travis.yml#L1-L20

dneise:
Okay let me go ahead and utilize this "travis matrix" also here in ctapipe. Should be quite interesting...

dneise:
I think it might be sub-optimal that the `install.sh` from the target software expects to be running in a travis environment...

dneise:
Current status:

 - protozfits v.1.4.0 does not support Py.3.7 
    --> using `protozfitsreader/archive/work_on_py3.7.tar.gz`
   expecting this branch to be merged within this week (OSX binaries need to be recompiled...then done)
 - Target software does not yet support Py3.7. 
    I cannot comment on the plans there, maybe @watsonjj knows more? 
    --> solution for the moment: commented out the Target installation.
 - `xvfb` workaround for matplotlib backends does not work on OSX
    --> I think there are better solutions for this, like: `export MPLBACKEND=Agg` 
    (from: https://stackoverflow.com/a/48728877)

kosack:
There are guidelines for which wrapping to use in the developer documentation, but I'm happy to update them if there are better suggestions.  The suggestion is to use ctypes for simple bindings. 

I think the only thing forbidden was SWIG, which comes from a higher-up CTA requirement.  Pybind11 should be ok  though.

dneise:
> I think the only thing forbidden was SWIG, which comes from a higher-up CTA requirement. Pybind11 should be ok though.

Interesting, could you elaborate on that higher up requirement? Is it a license issue?

mackaiver:
Waiting for the #864 merge.

dneise:
Moving this branch from my personal fork to the organisation, so all devs can easier play with it.

PR #763, opened 2018-06-19
Use HillasParametersContainer only

MaxNoe:


MaxNoe:
Tests are passing now

PR #764, opened 2018-06-25
Regression features in `RegressorClassifierBase`

vuillaut:
I added dictionaries to the class so that we can track what features have been used to train the model.
This will be particularly handy when the model is saved on disk and used elsewhere.


kosack:
could you add a unit test as well?

PR #765, opened 2018-06-27
Pointing correction transform matrix

ParsonsRD:
This is just a simple class based on discussions with @kosack that creates the affine transform matrix for applying pointing corrections. I'm not sure this is the best way to do this though so would welcome discussion :-).


ctapipe/coordinates/pointing_corrections.py
```diff
@@ -0,0 +1,58 @@
...
+        Get dummy transformation matrix, returns diagonal unit array
+
+        Returns
+        -------
+        matrix: affine translate matrix for camera corrections
+        """
+        return np.matrix(np.diagflat(np.ones(3)))
+
+
+class HESSStylePointingCorrection(PointingCorrection):
```
kosack: I think this is fairly similar to how all telescopes do it, not specific to HESS, so perhaps rename this something like "StandardPointingCorrection". In the docs you should also put the formula, and specify the order of operations, since S x R x T is not the same transform as T x R x S.   E.g. write how one would transform a vector [alt,az] using this matrix (that also depends if these are column or row-ordered matrices).

ctapipe/coordinates/pointing_corrections.py
```diff
@@ -0,0 +1,58 @@
...
+        rotation: float
+            Image rotation
+        scale: float
+            Scaling of focal length
+        """
+        self.x_trans, self.y_trans = x_trans, y_trans
+        self.rotation = rotation
+        self.scale = scale
+
+    def get_camera_trans_matrix(self):
```
kosack: don't use abbreviations: rename to "get_camera_transformation_matrix".  It could be a property or a lazyproperty as well.

ctapipe/coordinates/pointing_corrections.py
```diff
@@ -0,0 +1,58 @@
...
+        """
+        Get transformation matrix in the telescope frame, applying translation rotation
```
kosack: should mention that this is in *homogeneous coordinates* , e.g. it's a transform for a 2D vector, with an artificial 3rd coordinate added (see e.g. https://en.wikipedia.org/wiki/Homogeneous_coordinates)

kosack: and https://en.wikipedia.org/wiki/Transformation_matrix#Affine_transformations

kosack:
is anyone still working on this? I think it should be merged once small changes are made.

ParsonsRD:
Sorry @kosack I pretty much forgot about this. Will take another look next week.

ctapipe/coordinates/pointing_corrections.py
MaxNoe: I would rather not use something like `HESSStyle` or `Standard` but something that actually describes what it does.

ctapipe/coordinates/pointing_corrections.py
```diff
@@ -0,0 +1,58 @@
...
+        and scaling as in the HESS software
+
+        Returns
+        -------
+        matrix: affine translate matrix for camera corrections
+        """
+        c, s = np.cos(self.rotation), np.sin(self.rotation)
+        print(c, s, self.scale)
+        matrix = np.matrix([[c * self.scale, -s* self.scale, self.x_trans * c - self.y_trans * s],
```
MaxNoe: This is a standard rotation matrix for 3d around the z-axis, right?

I think there are numpy/astropy functions to create those.

ParsonsRD:
Starting working on this again after the coordinates update.

ctapipe/coordinates/pointing_corrections.py
ParsonsRD: Hi, yes I guess something must exist in astropy to do this. Did you come across any code for this in the coordinate refactor?

PR #767, opened 2018-07-05
Adding an example notebook no how to convert hex geometry to square and back

vuillaut:
It was not clear for me at first how to convert back from square to hexagonal geometry using the key (and the hidden buffer).
So I modified a bit the docstring of the conversion function and made an example notebook.


mackaiver:
Nice. I didn't even know this was in there. I don't quite understand why manually indexing the buffer is neccessary in the first place.

vuillaut:
@mackaiver I am not sure to understand why it's necessary either actually...
It was implemented by @tino-michael I think.

tino-michael:
My idea was to give the user a more fine-grained control over the uniqueness of the entries in the buffer (up to not using the buffer at all).
But if you want, by all means, remove the `key` argument from the function and make it automatically use `geom.cam_id`.

mackaiver:
> But if you want, by all means, remove the key argument from the function and make it automatically use geom.cam_id.

Yeah guess using `geom.cam_id` makes sense. 👍

kosack:
Perhaps a small 2 line example of converting forward and back should be added to the docstring as well, if it's confusing to use (that way people don't have to go looking in all the notebooks)

vuillaut:
Alright, I added example code in the docstring.

I also tried to use cam_id as default key but it would mean changing the default key of the square geometry from "LSTCam_square" back to simply "LSTCam" which would be a bad thing overall imo.
I think the first implementation was good, just lacked a bit of documentation for a newbie to use.

PR #770, opened 2018-07-30
Unstructured interpolator

ParsonsRD:
Fairly major update to ImPACT code. Removes all loops, significantly increases speed and adds a lot of new options. 

Also added xmax member to event.mc, as now this info is available in pyhessio (need to be sure this passes all tests in the master branch). 

Finally had to add a new line to apply the camera rotation in camera.py, otherwise this info does not seem to be properly copied...


ParsonsRD:
Hi @kosack, yes I think I changed some coordinate transforms internally and it broke some tests. I would also like to add some tests where I use a sample set of templates (with size ~10Mb). Where would be the best place for me to put this in the file structure.

kosack:
>  I would also like to add some tests where I use a sample set of templates (with size ~10Mb). Where would be the best place for me to put this in the file structure.

So far we've been putting all test data in the `ctapipe-extra` package, outside of ctapipe, so for now that's probably best.  Eventually we could put that sort of thing in a public webserver (like the one we have for CTA data already), and have it download automatically as needed, to avoid blowing up the size of ctapipe-extra, which was only meant for configuration info, not big data files.

kosack:
I'll commit this now, and you can add the new tests in a separate PR

Issue #695, opened 2018-03-28
Image truncation or leakage parameter missing?


rlopezcoto:
Hi,

I've been trying to find how truncated images are handled but could not find any parameter that is containing that information after the shower image analysis. Does anybody know where is this implemented or wether there is no implementation at all?

Thanks

kosack:
Right now it has not yet been merged into the main ctapipe code -there are only some basic mathods in @tino-michael's repo, including edge-pixel rejection. If you'd like to help, its appreciated! It's not clear to me yet what method is best.  The easiest of course is just putting a cut on the hillas centroid distance.

rlopezcoto:
Sorry, I could not find anything in any of the branches of the repo, @tino-michael could you point me to the modified files?

tino-michael:
It's not a branch in my ctapipe fork, it's in my `ImageCleaning` module in a separate repo:
https://github.com/tino-michael/tino_cta/blob/master/tino_cta/ImageCleaning.py

MaxNoe:
Got implemented in #783, function is `ctapipe.image.leakage`

PR #775, opened 2018-08-24
add leakage parameter #695

ThomaJu:
add 4 leakage parameter

### DIFF
--- a/ctapipe/image/leakage.py
+++ b/ctapipe/image/leakage.py
@@ -16,24 +16,24 @@ def leakage(geom, image):
 
     Parameters
     ----------
-        geom: `ctapipe.instrument.CameraGeometry`
-            Camera geometry information
-        image: array
-            pixel values
-
+    geom: ctapipe.instrument.CameraGeometry
+        Camera geometry information
+    image: array
+        pixel values
 
     Returns
     -------
-        leakage_pixel1:   float
-            Number of shower-pixel on the border divided by all shower-pixel
-        leakage_pixel2:   float
-            Number of shower-pixel in the second row of the border
-             divided by all shower-pixel
-        leakage_intensity1:   float
-            Number of photo-electrons in the border-pixel divided by all photo-electrons
-        leakage_intensity2:   float
-            Number of photo-electrons in the second row of the border-pixel
-             divided by all photo-electrons
+    leakage_pixel1:   float
+        Number of shower-pixel on the border divided by all shower-pixel
+    leakage_pixel2:   float
+        Number of shower-pixel in the second row of the border
+        divided by all shower-pixel
+    leakage_intensity1:   float
+        Number of photo-electrons in the border-pixel divided by all photo-electrons
+    leakage_intensity2:   float
+        Number of photo-electrons in the second row of the border-pixel
+        divided by all photo-electrons
+
     """
 
     max_value = max(np.sum(geom.neighbor_matrix, axis=0))
--- a/ctapipe/image/leakage.py
+++ b/ctapipe/image/leakage.py
@@ -31,6 +31,7 @@ def leakage(geom, image):
              divided by all photons
 
     """
+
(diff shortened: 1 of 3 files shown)


ctapipe/image/leakage.py
```diff
@@ -0,0 +1,53 @@
...
+            Number of photons in the second row of the border-pixel divided by all photons
+    '''
+    max_value = max(np.sum(geom.neighbor_matrix, axis=0))
+    leakage_border_pixel1 = 0
+    leakage_border_pixel2 = 0
+    leakage_border_photon1 = 0
+    leakage_border_photon2 = 0
+    size = 0
+    pixel_count = 0
+    for i in range(len(image)):
```
mackaiver: Maybe its faster if you only iterate over the non-zero pixel? I don't know. Shouldn't make a huge difference.

ctapipe/image/leakage.py
```diff
@@ -0,0 +1,53 @@
...
+            leakage_border_pixel1 += 1
+            leakage_border_photon1 += image[i]
+        for j in range(len(geom.neighbor_matrix[i])):
+            if geom.neighbor_matrix[i][j] == 0:
+                continue
+            if np.sum(geom.neighbor_matrix[j]) != max_value:
+                leakage_border_pixel2 += 1
+                leakage_border_photon2 += image[i]
+                break
+    return leakage_border_pixel1 / pixel_count, leakage_border_pixel2 / pixel_count, leakage_border_photon1 / size, leakage_border_photon2 / size
```
mackaiver: @kosack Should these values go into a container or something? Like the hillas parameters?

ctapipe/image/leakage.py
```diff
@@ -0,0 +1,59 @@
+"""
+Leakage calculation
+"""
+
+import numpy as np
+__all__ = ['leakage']
+
+
+def leakage(geom, image):
+    r"""Calculating the leakage-values for a given image.
```
kosack: Can you define in the docstring what is the definition of a leakage parameter?

ctapipe/image/leakage.py
```diff
@@ -0,0 +1,59 @@
...
+            pixel values
+
+    Returns
+    -------
+        leakage_border_pixel1:   float
+            Number of shower-pixel on the border divided by all shower-pixel
+        leakage_border_pixel2:   float
+            Number of shower-pixel in the second row of the border
+             divided by all shower-pixel
+        leakage_border_photon1:   float
```
kosack: I guess you mean photo-electrons and not photons here? It might be better to call it "intensity" as we do elsewhere.

kosack:
Probably this could be sped up a lot of numpy operations rather than for-loops or even my just using numba.jit, if speed is a problem.

kosack:
A somewhat more general question: for things like this, are we better off return a Container class with the names defined? In that case we can just add that container to the list of Containers to be written by a TableWriter, and automatically the results are added to the output file.  It is more complex though.

kosack:
Please fix the docstring so that the docs build correctly (I think it's just a missing blank line):
```
Warning, treated as error:
/home/travis/build/cta-observatory/ctapipe/ctapipe/image/leakage.py:docstring of ctapipe.image.leakage:11:Block quote ends without a blank line; unexpected unindent.
```

ctapipe/io/containers.py
```diff
@@ -669,3 +670,17 @@ class HillasParametersContainer(Container):
...
     kurtosis = Field(nan, 'measure of the tailedness')
+
+
+class LeakageContainer(Container):
+    """
+    Leakage
+    """
+    leakage_pixel1 = Field(nan, ("shower-pixel on the border"
+                                 "devided by all shower-pixel"))
+    leakage_pixel1 = Field(nan, ("shower-pixel in the second row of the border"
```
mackaiver: this should be leakage_pixel2

kosack:
superceded by #783

PR #776, opened 2018-09-04
Lst reader

FrancaCassol:
Added to LstEventSource class the possibility to read events from several files in parallel, this is necessary because the LST Event builder writes the events of a single run 4 four different fits files. The reader can now read them and loop on the events with the right order. 

Since EventSource foresees only  one input_url, in the case of multiple input files the name of the files must finish with suffix *000.fits.fz, *001.fits.fz, etc... and the user must give as input_url the name  of the first file (*000.fits.fz). The program will search for the other files. 
In the case of only one input file the input_url can have any form.


dneise:
@FrancaCassol I believe one would like to install the most recent version of `protozfits` for this:

    https://github.com/cta-sst-1m/protozfitsreader/archive/v1.4.0.tar.gz

So `environment.yml` would need to be modified

dneise:
Codacy is complaining about this:
https://github.com/FrancaCassol/ctapipe/blob/dbaa23c37a9d07c913ef5aa671b4ce919628fe57/ctapipe/io/lsteventsource.py#L218-L222

The thing is, line continuation with a backslash `\` is discouraged by many style guides. 
The reason I believe is, that a backslash at the end means something different depending on **invisible trailing characters**, so it is a line continuation if there is a line break after it ... but it is not a line continuation if there is a blank space after the `\`. Which is hard to spot and can easily lead to nasty bugs.

Most style guides I have seen, suggest to use implicit line continuation in brackets `()`.
So something like this:
```python
# verify that the configuration_id of all files are the same
# in the CameraConfig table
for path in paths:
    assert (
        self._camera_config[path].configuration_id == 
        self._camera_config[paths[0]].configuration_id
    )
```

dneise:
[This is what pep8 writes about it](https://www.python.org/dev/peps/pep-0008/#maximum-line-length)


> The preferred way of wrapping long lines is by using Python's implied line continuation inside parentheses, brackets and braces. Long lines can be broken over multiple lines by wrapping expressions in parentheses. These should be used in preference to using a backslash for line continuation.

> Backslashes may still be appropriate at times. For example, long, multiple with-statements cannot use implicit continuation, so backslashes are acceptable:

FrancaCassol:
Thanks Dominik! I was indeed stuck on this point ;-)

dneise:
No problem. I hope Codacy will be happy now ... :-)

FrancaCassol:
Hi Karl,
is there still something wrong with this pull?
Thanks

Franca

PR #777, opened 2018-09-07
Fixing core reconstruction

kpfrang:
Discussing with @TarekHC, @mackaiver and @moralejo we found that there was a bug in the calculation of the core position. I've tracked this down to being a wrong sign in one component of the norm vector in the HillasPlanes. In order to not break the direction reconstruction I added two lines which only change the y-component's sign of the norm vector in the estimation of the core position and should not affect anything else.

In this plot right here, you can clearly see that the old reconstruction of the position (blue diamond) was nonsense and the fixed estimate in red is clearly getting close to the true position (black).
attachment

### DIFF
--- a/ctapipe/reco/HillasReconstructor.py
+++ b/ctapipe/reco/HillasReconstructor.py
@@ -327,7 +327,9 @@ class HillasReconstructor(Reconstructor):
             # circle.norm to the ground gives higher weight to planes
             # perpendicular to the ground and less to those that have
             # a steeper angle
-            A[i] = circle.weight * circle.norm[:2]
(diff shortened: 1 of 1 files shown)


mackaiver:
Nice one. We should check if thats actually a bigger problem. Maybe connected to #721

MaxNoe:
I don't think this should have been merged, as Kai found the underlying reason which was the azimuth rotation, not a minus sign for y.

kosack:
Ah, that wasn't clear from the discussion.  I could revert it, or just wait for a general fix. Which would you prefer?

PR #778, opened 2018-09-07
Impact intersection

mackaiver:
While discussing with @TarekHC, @moralejo and @kpfrang I also came up with a fix for the impact parameter. A quick comparison between #777 and this PR on a small sample (1000 events) can be seen below.

I will try and get more statistics next week. 
impact_compare

### DIFF
--- a/ctapipe/reco/HillasReconstructor.py
+++ b/ctapipe/reco/HillasReconstructor.py
@@ -60,6 +60,20 @@ def normalise(vec):
         return vec
 
 
+def line_line_intersection_3d(uvw_vectors, origins):
+    C = []
+    S = []
+    for n, pos in zip(uvw_vectors, origins):
+        n = n.reshape((3, 1))
+        norm_matrix = n@n.T - np.eye(3)
+        C.append(norm_matrix@pos)
+        S.append(norm_matrix)
+
+    S = np.array(S).sum(axis=0)
+    C = np.array(C).sum(axis=0)
+    return np.linalg.inv(S)@C
+
+
 class HillasReconstructor(Reconstructor):
     """
     class that reconstructs the direction of an atmospheric shower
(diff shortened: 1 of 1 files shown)


mackaiver:
I added the same method to estimate max height. Seems to be better than the current method. At least on a small sample (1000 events). Also it is __much__ faster since its only matrix operations.

mackaiver:
Alright I added some changes to make this work properly.  Note that I fixed the coordinate conversion by changing the sign of the *azimuth* direction when conversion to cartesian. Since the astropy convention is spinning in the other direction. 

See the attached plots. Where I compare old reconstrunction  (orange )to this fix (yellow). 

h_max
impact

TarekHC:
Hi @mackaiver 

With old reconstrunction method you mean before the fix? It probably makes more sense to compare the method you implemented with the one @kpfrang fixed in #777 and #785. The changes are so minimal, that it would probably be easy to add them and compare the methods again.

mackaiver:
@TarekHC @kpfrang did the plots you asked for. For now just ~3000 events each.

h_max
impact



Runtime on my machine is improving by about 50%  for the matrix method.

kosack:
So what is the final decision about these?  Will somebody make the change in the coordinate code, or should we merge one of these PRs?  If not, we should close these ones. I guess now that the coordinates are done in a more central way, it might be better to do it there rather than to hack minus signs in several places, but it's up to you which you think is cleaner.

mackaiver:
I think it should be fixed in the HorizonFrame class. This might take some time however to test it. Maybe we can also discuss next week in person.

kosack:
I'd like to at least merge this one (which seems more general than #785, even if not as general as a fix in the coordinate transform), as a stop-gap so that we can make some end-to-end tests with correct h_max very soon. We can then open another issue to do this fix in a better way. 

Do you have any problem with that?

mackaiver:
Fine for me.  Not sure why the Travis tests failed though.

kosack:
Ok the travis tests work now - can you just add the 3D vector return as @jjlk mentioned, and I will merge it and we can do some tests on the grid.

PR #782, opened 2018-09-10
Add test that shows slicing breaks cam geom and fix it

MaxNoe:


ctapipe/instrument/tests/test_camera.py
```diff
@@ -150,3 +152,16 @@ def test_slicing():
     assert sliced2.pix_id[0] == 5
     assert sliced2.pix_id[1] == 7
     assert len(sliced2.pix_x) == 5
+
+
+def test_slicing_rotation():
```
kosack: please change this test to use the pre-loaded test event fixture, rather than opening a new event file. You can just add:

```py3
def test_x(test_event):
    event = deepcopy(test_event)
```

(I need to make a small update to the test fixtures so that the deepcopy is not needed).  The fixture is defined in `conftest.py`

MaxNoe: I need to make sure it's an LST event, because only there it is breaking. Is this the case for the fixture?

MaxNoe: Maybe it would be better to test it for all geometries

kosack: yes, it might be better to just load dummy cameras using `CameraGeometry.from_name` and fill them with some info, rather than opening a file (which is memory intensive and slow), since you don't need real data for this.

kosack:
```py3
@pytest.mark.parametrize("cam_id", cam_ids)
def test_X(cam_id):
    geom = CameraGeometry.from_name(cam_id)
    
```

MaxNoe: The Problem was that from_name did not show the problem, but the result of the hessioreader did

kosack: Ah, right that is because the  FITS versions of the geometries were written out after the camera rotation was applied (so the resulting rotation is 0)... Perhaps we can make an artificial test for that, like use "from_name", and call `rotate` to add an artificial non-zero rotation, and that should again break the code.

kosack: the following might work:

```py3
g = CameraGeomety.from_name("LSTCam")
g.rotate("25d")
g.cam_id = "LSTCam_rotated"
```
Then you get g as:
```
CameraGeometry(cam_id='LSTCam_rotated', pix_type='hexagonal', npix=1855, cam_rot=-25.0 deg, pix_rot=75.893 deg)
```

and that can be used in the tests

MaxNoe:
merge? We just went into this again.

ctapipe/instrument/tests/test_camera.py
```diff
@@ -6,6 +6,8 @@
     _get_min_pixel_seperation,
 )
 import pytest
+from ctapipe.io.eventsourcefactory import EventSourceFactory
+from ctapipe.utils import get_dataset_path
```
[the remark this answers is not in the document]
MaxNoe: done

MaxNoe:
done

MaxNoe:
@kosack See #802, would be nice if this gets merged

PR #783, opened 2018-09-10
Leakage

MaxNoe:
Follow up to #775, using numpy


MaxNoe:
borders

MaxNoe:
Yes, that's the idea.

kosack:
Ok i'll close #775

kosack:
can you re-open this PR? I will apply it again since github seems to have reverted it along with #789

PR #786, opened 2018-09-14
Adding class for subtract pedestals from binary file for LST

pawel21:
I have created class ```LSTR1Calibrator``` in ctapipe/calib/camera/r1.py to subtract pedestals from binary file (https://github.com/pawel21/ctapipe/blob/low_level_lst_camera_calibration/examples/pedestal_value20180705.dat).
Example program to call calibration process:
```python
from traitlets import Dict, List
from ctapipe.calib import CameraR1CalibratorFactory
from ctapipe.core import Tool
from ctapipe.io.lsteventsource import LSTEventSource
from ctapipe.io import EventSeeker

class R1Tester(Tool):
    name = "LST R1Tester"
    description = "Test LST R1 Calibrator"

    aliases = Dict(
        dict(
            r1='CameraR1CalibratorFactory.product',
            pedestal_path='CameraR1CalibratorFactory.pedestal_path',
        )
    )
    classes = List([
        CameraR1CalibratorFactory,
    ])

    def __init__(self, **kwargs):
        super().__init__(**kwargs)
        self.eventsource = None
        self.r1_calibrator = None
        self.plotter = None

    def setup(self):
        kwargs = dict(config=self.config, tool=self)

        self.r1_calibrator = CameraR1CalibratorFactory.produce(
            eventsource=None, **kwargs
        )

    def start(self):
        list_file = ['/home/pawel1/Pulpit/Astrophysics/CTA/dragon/data/dane_lst/20180705/Date20180705.0010.fits.fz']
        inputfile_reader_event0 = LSTEventSource(input_url = list_file[0], max_events = 15)

        for i, ev in enumerate(inputfile_reader_event0):
            self.r1_calibrator.calibrate(ev)

    def finish(self):
        pass

if __name__ == '__main__':
    exe = R1Tester()
    exe.run()
```

Run & Output:
```
(cta-dev) python calibration_pipeline_lst_r1_pr.py --r1 LSTR1Calibrator --pedestal_path pedestal_value20180705.dat 
INFO [R1Tester] (tool/initialize): ctapipe version 0.6.0.post21+git7df9f4c
INFO [R1Tester] (tool/run): Starting: LST R1Tester
INFO [R1Tester.CameraR1CalibratorFactory] (factory/_product): Obtaining LSTR1Calibrator from CameraR1CalibratorFactory
INFO [R1Tester.LSTR1Calibrator] (r1/_load_calib): Load binary file with pedestal version 1: pedestal_value20180705.dat 
INFO [R1Tester.LSTR1Calibrator] (r1/_load_calib): Number of modules in file: 259
INFO [R1Tester] (tool/run): Finished: LST R1Tester
INFO [R1Tester] (tool/run): Output: 
```

### DIFF
--- a/ctapipe/calib/camera/r1.py
+++ b/ctapipe/calib/camera/r1.py
@@ -15,7 +15,7 @@ of the data.
 """
 from abc import abstractmethod
 import numpy as np
-from ...core import Component, Factory
+from ...core import Component, Factory, Provenance
 from ...core.traits import Unicode
 from ...io import EventSource
 
@@ -23,6 +23,7 @@ __all__ = [
     'NullR1Calibrator',
     'HESSIOR1Calibrator',
     'TargetIOR1Calibrator',
+    'LSTR1Calibrator',
     'CameraR1CalibratorFactory'
 ]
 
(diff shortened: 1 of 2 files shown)


kosack:
If you want this to be a commandline tool, you also need to add an entry_point in `setup.py` as follows:

```
entry_points['console_scripts'] = [
    ...
    'ctapipe-lst-subtract-peds = ctapipe.tools.r1_tester:main'
]
```
And it may be best to put this code in the `ctapipe/tools` directory, which is where command-line tools go, rather than calib (which is for the library functions, not executables)

ctapipe/calib/camera/r1.py
```diff
@@ -191,6 +192,144 @@ def calibrate(self, event):
...
+
+    def _load_calib(self):
+        """
+        If a pedestal file has been supplied, create a array with
+        pedestal value . If it hasn't then point calibrate to
+        fake_calibrate, where nothing is done to the waveform.
+        """
+        Provenance().add_input_file(self.pedestal_path)
+        if self.pedestal_path:
+            with open(self.pedestal_path, "rb") as binary_file:
```
kosack: can you use a standard file format, preferably a FITS table for this rather than a custom binary format? We would like to avoid any hand-coded files, and soon will also require a set of metadata headers that must be included in any file. That's difficult with a custom format.

ctapipe/calib/camera/r1.py
```diff
@@ -191,6 +192,144 @@ def calibrate(self, event):
...
+        if event_number_of_modules == self.number_of_modules_from_file:
+            for nr_module in range(0, event_number_of_modules):
+                first_cap = self._get_first_capacitor(event, nr_module)
+                for gain in range(0, 2):
+                    for pixel in range(0, self.n_pixels):
+                        for roi in range(0, self.roisize):
+                            position = int((roi + first_cap[gain, pixel]) % self.size4drs)
+                            val = (
+                                event.r0.tel[0].waveform[gain, pixel + nr_module * 7, roi]
+                                - int(self.pedestal_value_array[nr_module, gain,
```
kosack: This code is very non-pythonic (it's written more like C), and will thus be very slow - it should be done using numpy vector operations, and that would probably simplify it to only a few lines.

pawel21: Thanks for feedback. Now, I known, I am working on it, soon I will send changes. What do you think about using `numba` to speed-up code?

ctapipe/calib/camera/r1.py
pawel21: We would like use the same file format as camera server can use, and unfortunately camera server can't use FITS table (because zFits decoder is not implement inside the camera server). The best for this reason would be simply binary file. What data format is allowed in ctapipe?

MaxNoe: > FITS table (because zFits decoder is not implement inside the camera server)

Why not use a standard-conform fits file that is readable by e.g. libcfitsio (and thus also astropy)?

Using custom binary formats never turned out well in the end for maintainability.

pawel21: Thanks a lot for comments. You are right. I will use fits file to store pedestal values.

kosack: I'd suggest just a ECSV text file then - that can be read by essentially everything, including astropy.table.  It's what we already use for many files in ctapipe. 

http://docs.astropy.org/en/stable/api/astropy.io.ascii.Ecsv.html

pawel21:
I am working on new version of code to subtract pedestal using new reader and a bit another approach.

PR #787, opened 2018-09-14
Minor changes: mostly deprecationwarning fixes

mireianievas:
* deprecation warning fixes
* escaping characters defined as raw strings
* astropy-healpix in environment.yml
* removal of unneeded? second loop

Number of warnings gone down from >100 to ~40 in pytest.


environment.yml
```diff
@@ -7,6 +7,7 @@ dependencies:
   - zeromq=4.2.2
   - ctapipe-extra=0.2.*
   - astropy
+  - astropy-healpix
```
kosack: where is healpix used in ctapipe? I would guess it is not a required dependency

mireianievas: You are right, it does not appear in any .py or .ipynb file. At some point when I was installing ctapipe I got an error because of it not being installed and I just added it, but it does not seem to be used and anyway I do not remember what was the error I got.

PR #790, opened 2018-09-27
Fixing the toy image generator

MaxNoe:
When implementing the concentrations features, I found that the toy image generator was severely broken.

* The given intensity is not reproduced, the showers are much, much too bright. 

* Width and Length are taken as variance, not standard deviation

* The cleaning for the test event is far from optimal.

Here is the image that is used for the hillas tests:

example_after

The maximum pixel value is over 9000!
Just kidding, it's over 2000.


MaxNoe:
This is the test image after the fix:
example_after

MaxNoe:
The hillas tests are now failing because some methods report a `psi` rotated by 180 degrees.

ctapipe/image/tests/test_geometry_converter.py
```diff
@@ -12,19 +12,33 @@
 cam_ids = CameraGeometry.get_known_camera_names()
 
 
+def create_mock_image(geom):
```
dneise: maybe one could expose the "magic numbers" as parameters of this function, like:
```
def create_mock_image(
    geom,
    centroid=0.3,
    width=0.03,
    length=0.10,
    psi="25d",
    intensity=0.5,
    nsb_level_pe=3,
):
```

dneise: or maybe one could somehow comment on this particular choice for these 6 parameters?

MaxNoe: The function is defined in the test, so is not a user facing function. The whole does also not work with astropy units. So if we would like to generalize this function, more work is needed anyway.

dneise: ok

MaxNoe:
The tests are passing again, because I coerced the psi angles in the test to the [-90, 90] interval. 

Maybe we should think about adding this in the hillas_functions themselves but we will also not keep 5 hillas_functions, right?

MaxNoe:
Somehow the installation of pyqt fails on travis for python 3.6

PR #797, opened 2018-10-10
Add unit to h_max in HillasReconstructor

jjlk:
Hi @mackaiver , @kosack ,
This short PR add a unit (meter) to the height of the shower mawimum computed by `estimate_h_max` (for consistency with other methods).

Don't you want also to rename the python file: HillasReconstructor.py to hillas_reconstructor.py for consistency with the other files?


mackaiver:
> Don't you want also to rename the python file: HillasReconstructor.py to hillas_reconstructor.py for consistency with the other files?

Yup. Renaming is a good idea. 👍

PR #798, opened 2018-10-10
speed up unit tests that use test_event fixture

kosack:
This speeds up the tests that use the `test_event` fixture in that the `test_event` is now only loaded once per pytest run.  

This also avoids some of the usual crashes during testing, since pyhessio does less file access

### DIFF
--- a/ctapipe/calib/camera/tests/test_calibrator.py
+++ b/ctapipe/calib/camera/tests/test_calibrator.py
@@ -1,5 +1,5 @@
-from copy import deepcopy
 from numpy.testing import assert_allclose
+
 from ctapipe.calib.camera import (
     CameraCalibrator,
     HESSIOR1Calibrator,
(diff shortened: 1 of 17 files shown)


dneise:
Would something like this help to avoid the deepcopy in the tests?
```python
@pytest.fixture(scope='module')
def module_wide_test_event():
    """ an example event for algorithm testing"""
    filename = get_dataset_path('gamma_test.simtel.gz')

    with HESSIOEventSource(input_url=filename) as reader:
        seeker = EventSeeker(reader)
        event = seeker['409']

    yield event


@pytest.fixture(scope='function')
def test_event(module_wide_test_event):
    """deep-copied test event"""
    from copy import deepcopy
    return deepcopy(module_wide_test_event)


```

kosack:
I'll try that out - it may work.

kosack:
Now it does the deepcopy internally (as per @dneise 's suggestion)

PR #800, opened 2018-10-11
Fix unknown section example warning

MaxNoe:


MaxNoe:
Docs build without warnings now

PR #801, opened 2018-10-12
Implement number of islands

LukasNickel:
* added number_of_islands to image/cleaning.py
(Calculates the number of connected clusters on a pixel mask)
* added unit test with LSTCam geometry

### DIFF
--- a/ctapipe/image/cleaning.py
+++ b/ctapipe/image/cleaning.py
@@ -93,3 +93,42 @@ def dilate(geom, mask):
         input mask (array of booleans) to be dilated
     """
     return mask | geom.neighbor_matrix_sparse.dot(mask)
+
+
+def number_of_islands(geom, mask):
+    """
+    Calculate and return the number of connected clusters of a pixel mask given 
+    the camera geometry. 
+    This can be used to seperate between gamma and hadronic showers.
+
+    Parameters
+    ----------
+    geom: `~ctapipe.instrument.CameraGeometry`
+        Camera geometry information
+    mask: ndarray
+        input mask (array of booleans)
+    """
+    hits = np.where(mask)[0]
+    # store information about marked pixels to avoid double counting
+    marked = np.ones(hits.shape[0]) * -1 
+    i = 0
+    count = 0
+
+    for hit in hits:
+
+        # count any new triggered pixel towards cluster count
+        if hit not in marked:
+            marked[i] = hit
+            i += 1
+            count += 1
+
+        # search for neighbors that triggered and arent marked already
+        neighbors = np.where(geom.neighbor_matrix[hit] & mask)[0]
+        neighbors_mask = np.isin(neighbors, marked)
+        new_neighbors_count = np.sum(np.invert(neighbors_mask))
+
+        # mark any new neighbors so that they wont be counted again
+        marked[i:i + new_neighbors_count] = neighbors[np.invert(neighbors_mask)]
+        i += new_neighbors_count
+
+    return count
--- a/ctapipe/image/cleaning.py
+++ b/ctapipe/image/cleaning.py
@@ -5,6 +5,7 @@ Image Cleaning Algorithms (identification of noisy pixels)
 __all__ = ['tailcuts_clean', 'dilate']
 
 import numpy as np
+from scipy.sparse.csgraph import connected_components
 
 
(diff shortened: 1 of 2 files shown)


ctapipe/image/cleaning.py
```diff
@@ -93,3 +93,42 @@ def dilate(geom, mask):
...
+    Parameters
+    ----------
+    geom: `~ctapipe.instrument.CameraGeometry`
+        Camera geometry information
+    mask: ndarray
+        input mask (array of booleans)
+    """
+    hits = np.where(mask)[0]
+    # store information about marked pixels to avoid double counting
+    marked = np.ones(hits.shape[0]) * -1
```
MaxNoe: this could be a python `set` or a boolean mask of length `n_pixels`. I don't think an integer array with the indices is the best choice here.

ctapipe/image/cleaning.py
```diff
@@ -93,3 +93,42 @@ def dilate(geom, mask):
         input mask (array of booleans) to be dilated
     """
     return mask | geom.neighbor_matrix_sparse.dot(mask)
+
+
+def number_of_islands(geom, mask):
```
MaxNoe: maybe it would be interesting to also return the clusters, to be able to visualize them

ctapipe/image/tests/test_cleaning.py
```diff
@@ -142,3 +142,17 @@ def test_tailcuts_clean_with_isolated_pixels():
                                          boundary_thresh=5,
                                          keep_isolated_pixels=True)
         assert (result == mask).all()
+
+
+def test_number_of_islands():
```
MaxNoe: You should also test for clusters that have a radius > 2, to check if the neighbor adding works.

ctapipe/image/cleaning.py
```diff
@@ -93,3 +93,42 @@ def dilate(geom, mask):
...
+    mask: ndarray
+        input mask (array of booleans)
+    """
+    hits = np.where(mask)[0]
+    # store information about marked pixels to avoid double counting
+    marked = np.ones(hits.shape[0]) * -1 
+    i = 0
+    count = 0
+
+    for hit in hits:
```
MaxNoe: I'm not sure this works for cluster with a radius > 2. You need to treat the neighbors of the neighbors recursively for that.

LukasNickel:
When I worked on this bigger clusters were counted correctly. In hindsight this might need the pixels to follow a specific camera geometry tho.
I will check on this once I find the time. If we also want to know which pixels connect to a cluster, it might be best to just rework it completely.

ctapipe/image/cleaning.py
```diff
@@ -93,3 +93,39 @@ def dilate(geom, mask):
...
+    Array entries correspond to cluster assignment, bool of array 
+    corresponds to input pixel mask.
+
+    Parameters
+    ----------
+    geom: `~ctapipe.instrument.CameraGeometry`
+        Camera geometry information
+    mask: ndarray
+        input mask (array of booleans)
+    """
```
MaxNoe: Documentation of output Parameters is missing

LukasNickel: True
Will add that asap

kosack:
This is written in a very non-pythonic way, so it likely quite inefficient, though perhaps it could be sped up quite a bit with a numba `@jit` decorator?

I suspect there is already code to count islands somewhere in the scientific python stack, since we use this technique already. @jjlk do you use a better method for this?

MaxNoe:
I think https://docs.scipy.org/doc/scipy/reference/generated/scipy.sparse.csgraph.connected_components.html#scipy.sparse.csgraph.connected_components is what we need

LukasNickel:
> I think https://docs.scipy.org/doc/scipy/reference/generated/scipy.sparse.csgraph.connected_components.html#scipy.sparse.csgraph.connected_components is what we need

That looks pretty good. I didn't expect a scipy function for an arbitrary geometry, but sparse graphs seems to be the keyword here. I'll play around with that function and upload a cleaner scipy version later

MaxNoe:
Looks good!

kosack:
@MaxNoe  do you still want changes? If not, you can accept your review and I'll merge this,

PR #825, opened 2018-11-19
First version of notebook for LST analysis bootcamp

MaxNoe:
Please review and comment about things, you think I should mention.

Plan is to add some words about 

* saving DL2 data
* Running the aict_tools on the output
* Say a bit more about development and future work


mackaiver:
Maybe you can show how to make a theta-square-like plot on the gammas? Not sure if this could be helpful.

MaxNoe:
Added the theta square plot

PR #826, opened 2018-11-20
Use sparse neighbor matrix

MaxNoe:
Speed up of ~7


dneise:
changing one line and getting a speedup of ~7 ? nice ... 
Could you comment on this a bit? Or is there an issue explaining what is going on?

kosack:
should we just  change `CameraGeom.neighbor_matrix` to be a sparse matrix by default? Is there any code left that uses the full matrix?  Even if so, it can be cast to an array to un-sparisfy the matrix. Then we can get rid of the `sparse_matrix` attribute altogether.

kosack:
> changing one line and getting a speedup of ~7 ? nice ...
 Could you comment on this a bit? Or is there an issue explaining what is going on?

I think it's just that the operation is O(n^2) where n is the number of pixels if you have a full matrix, but since the matrix is really sparsely filled since there are only at most 7 neighbors of a pixel (so a row in the matrix will only have up to 7 non-zero entries), any operation that works on the sparse version will be something closer to O(n).  

Sparse matrices are stored in such a way that they only store filled entries.  

IT's not totally clear that this speedup is always true, since in most cases doing operations on aligned data structures can be optimized by the CPU and the sparse version may pose some problems for that, but clearly numpy doesn't do that optimization at this point.

MaxNoe:
> Could you comment on this a bit? Or is there an issue explaining what is going on?

The scipy function we uses always works on sparse matrices, if the input is not sparse, it is converted to a sparse matrix. We already have it, so the conversion is avoided.

PR #827, opened 2018-11-20
Charge Resolution Update

watsonjj:
Revisited the charge resolution classes in ctapipe. They are much simpler now, and conform to more common standards, such as the utilisation of pandas HDFStore to store the values calculated.

A charge resolution is now built from events iteratively by building a list of pandas dataframes, which are combined once the list is above 1GB in memory (or at the end of the calculation). This essentially allows infinite events to be contributed to the charge resolution calculation.

Furthermore, the switch to use dataframes as opposed to a numpy array of fixed size allows for more flexible cases, such as a non-integer true charge which might arise when using the "average expected illumination" in place of true charge to generate a charge resolution from lab measurements.


ctapipe/plotting/charge_resolution.py
```diff
@@ -0,0 +1,358 @@
...
+
+    Parameters
+    ----------
+    array : ndarray
+
+    Returns
+    -------
+    float
+    """
+    return np.sqrt(np.sum(np.power(array, 2)) / array.size)
```
dneise: I was wondering why here `np.power(array, 2)` was used instead of `array**2`. I hardly ever saw the former version and find the latter a little bit simpler to read. So I assumed the former version must be somehow superior, I assumed it was maybe faster ... so I looked at the execution time ... 
```python
import numpy as np
a = np.random.normal(size=100000)
assert (np.power(a, 2) == a**2).all()  # assert there is no difference in the result
%timeit a**2  # 55 µs ± 318 ns
%timeit np.power(a, 2)  # 692 µs ± 6.09

b = (np.random.normal(size=100000) * 200).astype('i4')
assert (np.power(b, 2) == b**2).all()
%timeit b**2   # 25.7 µs ± 285 ns
%timeit np.power(b, 2)  # 182 µs ± 1.3 µs
```

So it seems on my machine `np.power()` is a bit slower than `**`, but returns the same result. So I am a bit puzzled what the reason for using `np.power` here might be? (maybe it performs different on different python and numpy versions?) 

Please note, I'm not trying to say one should use one version or the other due to performance differences. I think performance should be optimized last. So if `np.power` is used here for improved readability or for other reasons I believe it is totally fine. I am just curious.

ParsonsRD: Hi, thats weird I alway kind of assumed internally these are calling the same function and would take the same time! I'd be kind interested to find out what is happening internally there to cause the difference. 
Personally I use np.power because I am used to seeing C code and it is a similar syntax so helps readability for me, but I have no strong feelings either way.

dneise: Ah .. so the `pow()` in C is also slower than simple multiplication, see for example https://baptiste-wicht.com/posts/2017/09/cpp11-performance-tip-when-to-use-std-pow.html

The same principle applies of course to CPython, so *pow()*, which is named `np.power()`, should be slower than simple multiplication, but it is able to work with arbitrary powers. 

It seems the `**` operator, while being able to work with arbitrary powers just like `pow`, tries to be smart about the special case `power==2`, where it does not delegate to the lower level `np.power()` function but to simple multiplication. However already at `power==3` both have exactly the same behaviour, see below:
```python
In [1]: import numpy as np

In [2]: a = np.random.normal(size=1000) + 5  # only positive values in a

In [3]: %timeit a*a
872 ns ± 3.8 ns per loop (mean ± std. dev. of 7 runs, 1000000 loops each)

In [4]: %timeit a*a*a
1.81 µs ± 11.4 ns per loop (mean ± std. dev. of 7 runs, 1000000 loops each)

In [5]: # simple multiplication is optimized in contrast to arbitrary powers, see below

In [6]: %timeit np.power(a, 2)
8.29 µs ± 21.1 ns per loop (mean ± std. dev. of 7 runs, 100000 loops each)

In [7]: %timeit np.power(a, 3)
56.4 µs ± 172 ns per loop (mean ± std. dev. of 7 runs, 10000 loops each)

In [8]: # but simple multiplication cannot do this:

In [9]: %timeit np.power(a, 2.5)
56.3 µs ± 129 ns per loop (mean ± std. dev. of 7 runs, 10000 loops each)

In [10]: # Pythons ** operator seems to delegate to simple multiplication in case the exp == 2

In [11]: %timeit a**2
1.56 µs ± 7.79 ns per loop (mean ± std. dev. of 7 runs, 1000000 loops each)

In [12]: %timeit a**3
56.7 µs ± 218 ns per loop (mean ± std. dev. of 7 runs, 10000 loops each)

In [13]: # the statement above is exactly as slow as np.power(a, 3)
``` 

So as a rule of thumb (for Python), I would say: 

> Always use `base**exp` instead of `np.power(base, exp)`. 

(I personally would despite the numbers above not start optimizing runtime using multiplication for higher exponents, unless it is proven that this is really the bottleneck.)

ctapipe/plotting/charge_resolution.py
```diff
@@ -0,0 +1,360 @@
...
+    n_bins : int
+        Number of bins to allow in range
+
+    Returns
+    -------
+    pd.DataFrame
+    """
+    true = df['true'].values
+    min_ = true.min()
+    max_ = (true.max() // 500 + 1) * 500
```
dneise: Would it be possible to explain the "magic number" `500` in the line above?

ctapipe/plotting/charge_resolution.py
```diff
@@ -0,0 +1,360 @@
...
+
+        self.ax.set_xlabel("True Charge (p.e.)")
+        self.ax.set_ylabel(
+            r"Fractional Charge Resolution $\frac{{\sigma_Q}}{{Q}}$"
+        )
+
+        if not self.output_path:
+            raise ValueError("Output path must be specified")
+
+    def _set_file(self, path):
```
dneise: Could this method be renamed to reflect its purpose a little bit clearer? It seems this method opens a file, reads some contents from the file and stores these contents as instance variables for later use.
Maybe something like "read_charge_resolution_from_hdfstore()" or so?

ctapipe/plotting/charge_resolution.py
watsonjj: I will remove this "magic number" tweaking. I don't think theres any reason for it, I just missed it.

ctapipe/plotting/charge_resolution.py
```diff
@@ -0,0 +1,360 @@
...
+            Label for the figure's legend
+        """
+        self._set_file(path)
+        df_binned = bin_dataframe(self._df_pixel, self.n_bins)
+        agg = {'charge_resolution': ['mean', 'std'], 'true': 'mean'}
+        df_agg = df_binned.groupby(['bin']).agg(agg)
+        x = df_agg['true']['mean'].values
+        y = df_agg['charge_resolution']['mean'].values
+        yerr = df_agg['charge_resolution']['std'].values
+        self._plot(x, y, None, yerr, label)
```
dneise: I believe explicitly providing `None` for the 3rd argument is very C-like, while in Python one would rather write this like:
`self._plot(x,y, yerr=yerr, label=label)`

This `foo=foo` might feel a bit repetitive, therefore some people might write this part as:
```python
self._plot(
    x=df_agg['true']['mean'].values,
    y=df_agg['charge_resolution']['mean'].values,
    yerr=df_agg['charge_resolution']['std'].values,
    label=label,
)
```

ctapipe/plotting/charge_resolution.py
watsonjj: I don't think the long function name is necessary, but I will add more to the description. Its a private method that the user never needs to call.

ctapipe/plotting/charge_resolution.py
dneise: Cool thx

ctapipe/plotting/charge_resolution.py
dneise: I was more convert about your fellow devs than a "user"

dneise: The example is deliberately long .. I wanted to express what I mean by "a little bit clearer", by using a example. Not saying that this example is necessarily the best choice.

ctapipe/plotting/charge_resolution.py
watsonjj: Okay I've tried to address this a little in my latest commit

ctapipe/tools/extract_charge_resolution.py
```diff
@@ -19,36 +20,35 @@
 
 class ChargeResolutionGenerator(Tool):
     name = "ChargeResolutionGenerator"
```
dneise: **Note OT**: This comment is not specific to this PR, I just realized it while reading this diff.

this code:

    class MyTool(Tool):
        name = "MyTool"

Seems a bit repetitive to me.  Also this line might spark the following questions in a readers mind:

 * does every `Tool` need to define a `name` class variable?
 * if so: does this `name` have to be identical to the class name or can it be chosen freely?
 * why would one use `name` and rather than `self.__class__.__name__` are there certain caveats here, where the class name is somehow altered, so that it is safer to use `self.name`?

In a word a reader might mentally trip over this line and spend some minutes to figure out the meaning of this line in the ctapipe ecosystem. And I believe it would be a nice gesture to spare the reader this time if possible.

ctapipe/tools/extract_charge_resolution.py
```diff
@@ -19,36 +20,35 @@
...
-    description = "Generate the a pickle file of ChargeResolutionFile for " \
-                  "a MC file."
```
dneise: Again, this is not specific to this PR, I just realized it again, while reading this diff. 

As we all know, it is quite fundamental in Python to describe a class or function using a *docstring*. Docstrings are so fundamental in Python that they are introduced in the [Python Tutorial at the same time when they explain how to define a function](https://docs.python.org/3/tutorial/controlflow.html#defining-functions). 

So in a readers mind the line above might spark questions like:

 * Do all Tools need a member named `description`
 * Why is no docstring used for this?
 * Is it safe to add a docstring to a `Tool` or will it somehow be modified using `description`

Again, I think a reader might trip over this and spend some minutes perusing the ctapipe docu and the class definition of `Tool` to understand why `description` is used and not `__doc__`.

watsonjj:
Thanks for the comments, I think I correctly addressed them

ctapipe/tools/extract_charge_resolution.py
watsonjj: I thought the tool's command-line name is defined in setup.py?

Issue #828, opened 2018-11-21
Impact reconstruction still fails for different azimuth directions


MaxNoe:
Impact reconstruction on current master still fails dependent on pointing azimuth:

The files I used are here (standard cta password):
https://nextcloud.e5.physik.tu-dortmund.de/index.php/s/nA5SrraRjjCZRPc

Azimuth 180deg:

Azimuth 0deg:
![impact2](https://user-images.githubusercontent.com/5488440/48830721-8d288680-ed75-11e8-9035-f623f3b2f4f2.png)
![impact1](https://user-images.githubusercontent.com/5488440/48830722-8d288680-ed75-11e8-8367-8c99c3f90e6e.png)

I'm using this code to do the analysis:
```python
import astropy.units as u

from ctapipe.io import EventSourceFactory
from ctapipe.calib import CameraCalibrator

from ctapipe.image.cleaning import tailcuts_clean, number_of_islands

from ctapipe.reco import HillasReconstructor


# unoptimized cleaning levels, copied from 
# https://github.com/tudo-astroparticlephysics/cta_preprocessing
cleaning_level = {
    'LSTCam': (3.5, 7.5, 2),  # ?? (3, 6) for Abelardo...
    'FlashCam': (4, 8, 2),  # there is some scaling missing?
    'CHEC': (2, 4, 2),
}

allowed_cameras = {'LSTCam', 'CHEC', 'FlashCam'}


input_url = '/home/maxnoe/Downloads/gamma_20deg_180deg_run7360___cta-prod3-merged_desert-2150m-Paranal-3HB89-NGFD.simtel.gz'
event_source = EventSourceFactory.produce(input_url=input_url, max_events=8)

calibrator = CameraCalibrator(
    eventsource=event_source,
)

reco = HillasReconstructor()

for event in event_source:
    print(f'Id: {event.count}', f'E = {event.mc.energy:1.3f}', f'Telescopes: {len(event.r0.tel)}', sep=', ')
    
    calibrator.calibrate(event)
    
    # mapping of telescope_id to parameters for stereo reconstruction
    hillas_containers = {}
    pointing_azimuth = {}
    pointing_altitude = {}
    time_gradients = {}
    
    for telescope_id, dl1 in event.dl1.tel.items():
        camera = event.inst.subarray.tels[telescope_id].camera
        
        if camera.cam_id not in allowed_cameras:
            continue
            
        image = dl1.image[0]
        peakpos = dl1.peakpos[0]
        
[...]

PR #830, opened 2018-11-21
Fix impact reco

MaxNoe:
Anyone with working knowledge of the 3d reconstruction should check this carefully.

* I added transformations for `GroundFrame` to `TiltedGroundFrame` for the telescope positions and a `project_to_ground` for the result. See discussion in #828 

* I also replaced `phi` with `psi` in the calculations, that made the diffuse gammas a lot better.

### DIFF
--- a/ctapipe/reco/HillasReconstructor.py
+++ b/ctapipe/reco/HillasReconstructor.py
@@ -7,7 +7,7 @@ Contact: Tino Michael <Tino.Michael@cea.fr>
 
 from ctapipe.reco.reco_algorithms import Reconstructor
 from ctapipe.io.containers import ReconstructedShowerContainer
-from ctapipe.coordinates import HorizonFrame, CameraFrame
+from ctapipe.coordinates import HorizonFrame, CameraFrame, GroundFrame, TiltedGroundFrame, project_to_ground
 from astropy.coordinates import SkyCoord, spherical_to_cartesian, cartesian_to_spherical
 from itertools import combinations
 
@@ -92,8 +92,9 @@ class HillasReconstructor(Reconstructor):
         super().__init__(config=config, parent=tool, **kwargs)
         self.hillas_planes = {}
 
-    def predict(self, hillas_dict, inst, pointing_alt, pointing_az, seed_pos=(0, 0)):
-        """The function you want to call for the reconstruction of the
+    def predict(self, hillas_dict, inst, pointing_alt, pointing_az):
+        '''
+        The function you want to call for the reconstruction of the
         event. It takes care of setting up the event and consecutively
         calls the functions for the direction and core position
         reconstruction.  Shower parameters not reconstructed by this
@@ -101,23 +102,21 @@ class HillasReconstructor(Reconstructor):
 
         Parameters
         -----------
-        hillas_dict : python dictionary
+        hillas_dict: dict
             dictionary with telescope IDs as key and
             HillasParametersContainer instances as values
         inst : ctapipe.io.InstrumentContainer
             instrumental description
-        pointing_alt:
-        pointing_az:
-        seed_pos : python tuple
-            shape (2) tuple with a possible seed for
-            the core position fit (e.g. CoG of all telescope images)
+        pointing_alt: dict[astropy.coordinates.Angle]
+            dict mapping telescope ids to pointing altitude
+        pointing_az: dict[astropy.coordinates.Angle]
+            dict mapping telescope ids to pointing azimuth
 
         Raises
         ------
         TooFewTelescopesException
             if len(hillas_dict) < 2
-
-        """
+        '''
 
         # stereoscopy needs at least two telescopes
         if len(hillas_dict) < 2:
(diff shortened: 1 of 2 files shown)


MaxNoe:
Test on three files, very low statistics:
impact_compared

MaxNoe:
This is before the changes:
impact_before

MaxNoe:
So point-like is worse, I don't know why.

thomasgas:
> So point-like is worse, I don't know why.

I was also looking at this. 
I give you my idea: for point like sources phi and psi are **almost** equal. When you use phi instead of psi you are basically saying that your ellipse points **exactly** to the center of the camera, whereas it's real angle is psi.
So, i guess that using phi gives better results for point-like sources because it's like cheating...but not too much :smile:

MaxNoe:
@all Could we maybe get this reviewed and merged before the weekend?

Would be nice to have this working for the LST Bootcamp

MaxNoe:
I fixed the documentation, the typo and added the example script I used as I think it will be helpfull

mackaiver:
> I give you my idea: for point like sources phi and psi are **almost** equal. When you use phi instead of psi you are basically saying that your ellipse points **exactly** to the center of the camera, whereas it's real angle is psi.
> So, i guess that using phi gives better results for point-like sources because it's like cheating...but not too much 😄

That is the reason yes.  Code looks good.

MaxNoe:
So were are sure that using `psi` is more correct than `phi` and that is just "worked" before because for point-like gammas φ ≈ ψ?

ctapipe/reco/HillasReconstructor.py
```diff
@@ -193,8 +196,8 @@ def inititialize_hillas_planes(
 
         self.hillas_planes = {}
         for tel_id, moments in hillas_dict.items():
-            p2_x = moments.x + 0.1 * u.m * np.cos(moments.psi)
-            p2_y = moments.y + 0.1 * u.m * np.sin(moments.psi)
+            p2_x = moments.x + moments.length * np.cos(moments.psi)
+            p2_y = moments.y + moments.length * np.sin(moments.psi)
```
vuillaut: Only the direction of the vector is of use here, that's why its length was set to an arbitrary value of 0.1*u.m.
Does `moments.length` carries more information here?
Either way, it is fine, unless:
- `moments.length` has a 0 or `nan` value (but not psi which is really what matters here)
- `p2` is used for graphical representation of the direction, in which case `moments.length` might be small in many cases. But that could be that problem could be left to the visualization.

vuillaut:
@MaxNoe yes, phi is the angle between (x,y) and the camera centre, it does not give you the direction of the ellipse, unless `psi=phi` which happen to be the case for prod3 point source where telescopes are pointing exactly towards the source.

ctapipe/reco/HillasReconstructor.py
mackaiver: >     * `p2` is used for graphical representation of the direction, in which case `moments.length` might be small in many cases. But that could be that problem could be left to the visualization.

good point. maybe leaving a constant 0.1 here makes more sense.

mackaiver:
>  ps: It might be a good idea to give these angles some more meaningful names, what do you think?

Yes. Definitely.

MaxNoe:
> yes, phi is the angle between (x,y) and the camera centre,

I did know their definitions, I wasn't sure which one is needed to estimate impact.

> ps: It might be a good idea to give these angles some more meaningful names, what do you think?

It was  already discussed some time ago here: https://github.com/cta-observatory/ctapipe/pull/771

vuillaut:
`phi` could be called `direction`
I don't have a clever name for `psi`. In polar coordinates system it is called `phi`  😅 
Should I open an issue and we deal with that later?

MaxNoe:
Carl suggested names in #771 which I think are fine and more meaningful

vuillaut:
Ha yes our messages crossed.
Good suggestion indeed, thanks for pointing this out.


Thomas

ParsonsRD:
> `phi` could be called `direction`
> I don't have a clever name for `psi`. In polar coordinates system it is called `phi` 😅
> Should I open an issue and we deal with that later?

I couldn't agree more with this comment, phi and psi is confusing and there is no convention for this naming. I think in HESS they are named the opposite way around!

MaxNoe:
I processed some more diffuse gammas, here are the results:

Overall:
impact_all

Containments vs. multiplicity
impact_multiplicty

Containment vs. energy
impact_energy

ctapipe/reco/HillasReconstructor.py
MaxNoe: The vectors are weighted with `intensity * length / width` so having a `nan` or `0` length will in any case have a invalid reconstruction.

But I can put the `0.1 m` back. I was worried that it is camera dependent and could be too large for small cameras. Might this be the case? Or does it not matter?

MaxNoe: I added the 0.1 meters back

PR #833, opened 2018-11-22
Prefix containers

MaxNoe:
This allows writing multiple containers using a prefix for each column in the hdfwriter.

This enables writing duplicated keys (e.g. `ReconstructedShower.alt`, `MCEvent.alt`).

By default the prefix is `cls.__name__.lower().replace('Container', '')`


vuillaut:
The way I understand it, the user can change the prefix as he likes, right?

MaxNoe:
> The way I understand it, the user can change the prefix as he likes, right?

Not really intended. The prefix is a class member of the container.
But as this is python, you could change this class member.

ctapipe/core/container.py
```diff
@@ -59,10 +59,16 @@ def __new__(cls, name, bases, dct):
                 for k, v in b.fields.items():
                     dct['fields'][k] = v
 
-        for k in items:
+        for k in field_names:
```
dneise: If you like, I believe `for name in field_names` might be easier to read than `k`

ctapipe/io/hdf5tableio.py
```diff
@@ -72,7 +74,15 @@ class HDF5TableWriter(TableWriter):
...
     """
 
-    def __init__(self, filename, group_name, mode='w', root_uep='/', **kwargs):
+    def __init__(
+        self,
+        filename,
+        group_name,
+        add_prefix=False,
+        mode='w',
+        root_uep='/',
```
dneise: what is a "uep" ?

kosack:
>> The way I understand it, the user can change the prefix as he likes, right?

>Not really intended. The prefix is a class member of the container.
>But as this is python, you could change this class member.

This is definitely the behavior we need - the main use case is that we will have for example multiple image cleanings and multiple hillas parameterizations that will be stored in the output file, so we want to be able to change the prefix for each one.  E.g. if you have multiple image cleanings, for example (loose and tight), you might have a dict inside your container like:

```py3
cont.hillas['tight'] = hillas_parameters(...)
cont.hillas['loose'] = hillas_parameters(...)
```

then you want to have outputs like:

`tight_length, tight_width` (or perhaps `tight_hillas_width` if you combine multiple prefixes)

ctapipe/io/hdf5tableio.py
```diff
@@ -105,6 +116,12 @@ def close(self):
 
         self._h5file.close()
 
+    def build_prefix(self, container):
```
kosack: Shouldn't this prefix-building code go in the base class? I would assume it's the same idea for any output type (e.g. FITS or other table format).  This way, the prefixes will only work with HDF5 outputs.  Of course we haven't implemented any other TableWriters, but I intend to do at least an AstropyTableWriter and a pure FITSTableWriter

MaxNoe: yes, good point

MaxNoe:
I updated this to enable instance level prefixes and moved the `build_prefix` function to the base `TableWriter`.

Containers now have a class variable `containter_prefix` which is used to initialize the instance level `prefix`.

Unfortunately this was not possible differently, since python does not allow class variables to initiliaze `__slots__`.

ctapipe/io/hdf5tableio.py
thomasgas: "The user entry point to the object tree attached to the HDF5 file is represented in the root_uep attribute". From http://www.pytables.org/usersguide/libref/file_class.html

mackaiver: We want to get rid of `root_uep`. We want to merge it with the `group_name`.

ctapipe/io/tests/test_hdf5.py
```diff
@@ -28,24 +32,45 @@ def test_write_container(temp_h5_file):
     r0tel.meta['test_attribute'] = 3.14159
     r0tel.meta['date'] = "2020-10-10"
 
-    writer = HDF5TableWriter(str(temp_h5_file), group_name='R0',
-                             filters=tables.Filters(
-        complevel=7))
-    writer.exclude("tel_002", ".*samples")  # test exclusion of columns
+    with HDF5TableWriter(
+            str(temp_h5_file),
```
dneise: is `str(temp_h5_file)` really needed here? Looks like `temp_h5_file` is already a string.

Also the name `temp_h5_file` looks wrong to me ... it is a path not a file-like

MaxNoe: I also do not like the whole test setup here. Tests depend on each other.The first test writes into the same temporary file as another reads. Order is important here. Parallel execution of tests might not be possible, right?

kosack: I agree. That's my fault - the tests were poorly written. We should separate them so that they do not depend on each other

kosack: the `str(temp_h5_file)` was just to get the filename as a string, since pytables wants just a string, not whatever the tmpfile fixture gives back (which is perhaps a pathlib Path?). There's probably a better way.

ctapipe/io/tests/test_hdf5.py
```diff
@@ -59,58 +84,52 @@ class C2(Container):
         d = Field('d', None)
 
     with tempfile.NamedTemporaryFile() as f:
```
kosack: You should not use `tempfile` in tests, and instead of pytest's built-in fixtures for that.  
I think pytest recommends using it's own system, since tempfile may not do the right thing in self-contained testing sessions, if I remember right. The correct usage is here:  
https://docs.pytest.org/en/latest/tmpdir.html

dneise: done

ctapipe/io/tests/test_hdf5.py
dneise: done

PR #835, opened 2018-11-22
MAGIC data readers implementation

IevgenVovk:
Implementation of the MAGIC data readers as derivatives of the event source factory. These include readers both from the native MAGIC data format (ROOT) and converted HDF5 files. Also included are MAGIC camera definitions, necessary to work with these data.

Note: reading of the native ROOT data requires additional "uproot" package (https://github.com/scikit-hep/uproot).


README.rst
```diff
@@ -49,3 +49,11 @@ Developers should follow the development install instructions found in the
 `documentation <https://cta-observatory.github
 .io/ctapipe/getting_started>`_.
 
+
+If you're installing *ctapipe* manually, you'll also need these external dependencies:
```
MaxNoe: I would rather give `pip installs here`, as `git clone` is not enough:

```
pip install https://github.com/cta-observatory/pyhessio/archive/master.tar.gz
pip install https://github.com/cta-observatory/ctapipe-extra/archive/master.tar.gz
pip install https://github.com/cta-sst-1m/protozfitsreader/archive/master.tar.gz
```

ctapipe/io/containers.py
```diff
@@ -111,6 +112,20 @@ class DL1CameraContainer(Container):
...
     )
+    badpixels = Field(
+        None, "numpy (nd-)array of bools indicating (a specific type of) bad pixels"
+    )
+
+
+class WeatherContainer(Container):
+    """Storage of event-wise weather information (MAGIC implementation).
+    """
+    air_temperature = Field( None, "Outside air temperature"
```
MaxNoe: strange linebreaks here, the parenthese should be on the same line

ctapipe/io/containers.py
```diff
@@ -381,6 +396,8 @@ class TelescopePointingContainer(Container):
     """
     azimuth = Field(nan * u.rad, 'Azimuth, measured N->E', unit=u.rad)
     altitude = Field(nan * u.rad, 'Altitude', unit=u.rad)
+    ra = Field(nan * u.rad, 'Right ascension of camera center pointing from raw data', unit=u.rad)
```
MaxNoe: I don't think we should duplicate information here. Shouldn't this be handled by coordinate transforms?

ctapipe/io/magiceventsource.py
```diff
@@ -0,0 +1,1337 @@
+
+
+import glob
+import re
+
+import uproot
```
MaxNoe: With this, uproot is now a required dependency of ctapipe.

So it has to be added to the setup.py and environment.yaml.
Or you make it optional by wrapping it in something like `try... except ImportError`

ctapipe/io/magiceventsource.py
```diff
@@ -0,0 +1,1337 @@
...
+            import h5py
+        except ImportError:
+            msg = "The `h5py` python module is required to access the MAGIC data or MCs"
+            self.log.error(msg)
+            raise
+
+        self.h5py = h5py
+        self.file = h5py.File(self.input_url)
+
+#         if MAGICEventSource._count > 0:
```
MaxNoe: obsolete code, should be removed

ctapipe/io/magiceventsource.py
```diff
@@ -0,0 +1,1337 @@
...
+
+
+class MAGICEventSourceHDF5(EventSource):
+    """
+    EventSource for MAGIC raw data converted to hdf5.
+
+    This class utilises `h5py` to read the hdf5 file, and stores the
+    information into the event containers.
+    """
+    _count = 0
```
MaxNoe: is count really needed? hdf5 allows multiple instances of `h5py.File`

ctapipe/io/magiceventsource.py
```diff
@@ -0,0 +1,1337 @@
...
+
+    @staticmethod
+    def is_compatible(file_path):
+        import h5py
+        # check general format:
+        if not h5py.is_hdf5(file_path):
+            return False
+        # crude check if hdf5 file contains MAGIC raw data:
+        with h5py.File(file_path, "r") as file:
+            intrument_attr = file.attrs['instrument']
```
MaxNoe: this might raise a `KeyError` if it's not a MAGIC file, so you rather should do:

```python
f.attrs.get('instrument')
```

ctapipe/io/magiceventsource.py
```diff
@@ -0,0 +1,1337 @@
...
+
+
+    @staticmethod
+    def is_compatible(file_path):
+        import h5py
+        # check general format:
+        if not h5py.is_hdf5(file_path):
+            return False
+        # crude check if hdf5 file contains MAGIC raw data:
+        with h5py.File(file_path, "r") as file:
```
MaxNoe: Don't overwrite builtins. Use e.g. `hdf_file` or simply `f`

ctapipe/io/magiceventsource.py
```diff
@@ -0,0 +1,1337 @@
...
+            intrument_attr = file.attrs['instrument']
+            if intrument_attr.tostring() == b"MAGIC":
```
MaxNoe: what is `f.attrs['instrument']` supposed to be? I tried writing both a `str` and a `bytes`, neither have a `tostring` method and why compare with bytes `v'MAGIC'`?

ctapipe/io/magiceventsource.py
```diff
@@ -0,0 +1,1337 @@
...
+                return True
+            else:
+                return False
+            
+    def __exit__(self, exc_type, exc_val, exc_tb):
+        self.file.close()
+
+    def _generator(self):
+        with self.h5py.File(self.input_url, "r") as file:
```
MaxNoe: don't overwrite builtins (`file`)

ctapipe/io/magiceventsource.py
```diff
@@ -0,0 +1,1337 @@
...
+            # the container is initialized once, and data is replaced within
+            # it after each yield
+            counter = 0
+            #eventstream = file.move_to_next_event()
```
MaxNoe: left over hessio code

ctapipe/io/magiceventsource.py
```diff
@@ -0,0 +1,1337 @@
...
+                eventstream = file['dl1/event_id']
+                if file.attrs['data format'] == b"mono M1":
+                    tels_in_file = ["M1"]
+                elif file.attrs['data format'] == b"mono M2":
+                    tels_in_file = ["M2"]
+                elif file.attrs['data format'] == b"stereo":
+                    tels_in_file = ["M1", "M2"]
+                
+            elif file.attrs['dl_export'] == b"r0":
+                if list(file.keys()) == ["MAGIC1_rawdata"]:
```
MaxNoe: better use `'MAGIC1_rawdata' in file`

ctapipe/io/magiceventsource.py
```diff
@@ -0,0 +1,1337 @@
...
+                    eventstream = file['MAGIC1_rawdata/EvtHeader/StereoEvtNumber']
+                    events_per_tel[0] = np.array(eventstream, dtype = dt)
+                    tels_in_file = [1]
+                elif list(file.keys()) == ["MAGIC2_rawdata"]:
```
MaxNoe: better use `'MAGIC2_rawdata' in file`

ctapipe/io/magiceventsource.py
```diff
@@ -0,0 +1,1337 @@
...
+                    eventstream = file['MAGIC2_rawdata/EvtHeader/StereoEvtNumber']
+                    events_per_tel[1] = np.array(eventstream, dtype = dt)
+                    tels_in_file = [2]
+                elif list(file.keys()) == ["MAGIC1_rawdata", "MAGIC2_rawdata"]:
```
MaxNoe: better use `"MAGIC1_rawdata" in file and "MAGIC2_rawdata" in file`.

Than this has to be the first if

ctapipe/io/magiceventsource.py
```diff
@@ -0,0 +1,1337 @@
...
+                    events_per_tel[0] = np.array(file['MAGIC1_rawdata/EvtHeader/StereoEvtNumber'], dtype = dt)
+                    events_per_tel[1] = np.array(file['MAGIC2_rawdata/EvtHeader/StereoEvtNumber'], dtype = dt)
+                    tels_in_file = [1, 2]
+                    # order MC mono events into event stream:
```
MaxNoe: Why are we drawing random numbers here in an `EventSource`?

ctapipe/io/magiceventsource.py
```diff
@@ -0,0 +1,1337 @@
...
+# 
+                obs_id = file.attrs['RunNumber']
+                data.count = counter
+                data.r0.obs_id = file.attrs['RunNumber']
+                data.r0.event_id = event_id
+                data.r1.obs_id = file.attrs['RunNumber']
+                data.r1.event_id = event_id
+                data.dl0.obs_id = obs_id
+                data.dl0.event_id = event_id
+#
```
MaxNoe: remove obsolete code

ctapipe/io/magiceventsource.py
```diff
@@ -0,0 +1,1337 @@
...
+                        data.r0.tel[tel_id].num_trig_pix = file['MAGIC'+str(tel_id) +'_rawdata/EvtHeader/NumTrigLvl2'][nevent]
+                        
+                        # add MC information:
+                        if data.meta['is_simulation'] == True:
+                            # energy of event should be the same in both telescopes, so simply try both:
+                            data.mc.energy = file['MAGIC'+str(tel_id) +'_rawdata/McHeader/Energy']["Energy"][nevent] * u.TeV
+                            data.mc.core_x = file['MAGIC'+str(tel_id) +'_rawdata/McHeader/Core_xy']["Core_x"][nevent] * u.m
+                            data.mc.core_y = file['MAGIC'+str(tel_id) +'_rawdata/McHeader/Core_xy']["Core_y"][nevent] * u.m
+                            data.mc.h_first_int = file['MAGIC'+str(tel_id) +'_rawdata/McHeader/H_first_int']["H_first_int"][nevent] * u.m
+
```
MaxNoe: remove obsolete code

ctapipe/io/magiceventsource.py
```diff
@@ -0,0 +1,1337 @@
...
+                data.r0.tels_with_data = tels_with_data
+                data.r1.tels_with_data = tels_with_data
+                data.dl0.tels_with_data = tels_with_data
+                data.trig.tels_with_trigger = tels_with_data
+                data.inst.subarray = magic_subarray
+                
+                yield data
+                counter += 1
+        
+        return
```
MaxNoe: empty return is not needed

ctapipe/io/magiceventsource.py
```diff
@@ -0,0 +1,1337 @@
...
+
+    def _build_subarray_info(self, file):
```
MaxNoe: This method is still from hessio and is probably never called

ctapipe/io/magiceventsource.py
```diff
@@ -0,0 +1,1337 @@
...
+        ----------
+        run_number: int
+            The run number to use.
+
+        Returns
+        -------
+
+        """
+
+        input_path = '/'.join(self.input_url.split('/')[:-1])
```
MaxNoe: I think this line should be `input_path = os.path.dirname(self.input_url)`

ctapipe/io/magiceventsource.py
```diff
@@ -0,0 +1,1337 @@
...
+        this_run_mask = input_path + '/*{:d}*root'.format(run_number)
```
MaxNoe: `os.path.join`

MaxNoe:
https://github.com/cta-observatory/ctapipe/issues/813

moritzhuetten:
@MaxNoe (@IevgenVovk ): Thanks for your detailed review, we'll take care of adding unit tests! Concerning your first general point, we maybe should discuss the next days in Padua. I think that it shouldn't be a big problem to add the event source to the core code as (i) it is just a single file not harming anything else, and (ii) and more importantly, from my point of view, it is already hard to keep track of the many floating-around ctapipe "extensions" (for example, we had no idea about that repository you mention; the optics/camera definitions are already in ctapipe-extra, and then there would be another dependency to actually load the data). So keeping things together should help people to know what's already there (and for me, the ctapipe/io is the natural place where to search first). But I admit, at the end this might be a rather political decision. Do you think this needs more time to discuss at some higher-order level? In any case, we should merge your (Master student's) and our efforts!

README.rst
dneise: I'd even like to give people who want to install *all optional dependencies* an easy and easy to remember way to do so.

ctapipe/io/magiceventsource.py
```diff
@@ -0,0 +1,1337 @@
...
+
+    def __init__(self, config=None, tool=None, **kwargs):
+        super().__init__(config=config, tool=tool, **kwargs)
+
+        try:
+            import h5py
+        except ImportError:
+            msg = "The `h5py` python module is required to access the MAGIC data or MCs"
+            self.log.error(msg)
+            raise
```
dneise: `h5py` is a default anaconda package In Py3.5 and Py3.6 anaconda for Linux (both 32bit and 64bit versions). 
One can find a list of all packages coming with anaconda here: 

https://docs.anaconda.com/anaconda/packages/py3.6_linux-64/

In addition ctapipe has made `h5py` an explicit dependency by stating it in the `environment.yml`:

https://github.com/cta-observatory/ctapipe/blob/master/environment.yml#L14

So I think there is no need to treat h5py as an optional dependency here.

ctapipe/io/magiceventsource.py
dneise: where just `f` would violate pep8

ctapipe/io/magiceventsource.py
dneise: This function is very long, maybe try to break out some smaller functions.

ctapipe/io/magiceventsource.py
```diff
@@ -0,0 +1,1337 @@
...
+            NOTE: The file mask of the data to read can be passed with
+            the 'input_url' parameter.
+        """
+
+        file_list = glob.glob(kwargs['input_url'])
+        file_list.sort()
+
+        # EventSource can not handle file wild cards as input_url
+        # To overcome this we substitute the input_url with first file matching
+        # the specified file mask.
```
dneise: This is  a general issue with `EventSource` right? If so, well spotted. Maybe this issue should be fixed in EventSource then.

ctapipe/io/magiceventsource.py
```diff
@@ -0,0 +1,1337 @@
...
+        }
+        # MAGIC telescope description
+        optics = OpticsDescription.from_name('MAGIC')
+        geom = CameraGeometry.from_name('MAGICCam')
+        self.magic_tel_description = TelescopeDescription(optics=optics, camera=geom)
+        self.magic_tel_descriptions = {1: self.magic_tel_description, 2: self.magic_tel_description}
+        self.magic_subarray = SubarrayDescription('MAGIC', self.magic_tel_positions, self.magic_tel_descriptions)
+
+    def __exit__(self, exc_type, exc_val, exc_tb):
+        return
```
dneise: an empty  `__exit__`  is inherited from `EventSource` no need to explicitly write it down

ctapipe/io/magiceventsource.py
```diff
@@ -0,0 +1,1337 @@
...
+            try:
+                with uproot.open(file_path) as input_data:
+                    if 'Events' not in input_data:
+                        is_magic_root_file = False
+            except ValueError:
+                # uproot raises ValueError if the file is not a ROOT file
+                is_magic_root_file = False
+                pass
+
+        return is_magic_root_file
```
dneise: This can be sped up if you like:
instead of `is_magic_root_file = False`, just `return False`.
and `return True` in the end. 

But only if you like.

ctapipe/io/magiceventsource.py
```diff
@@ -0,0 +1,1337 @@
...
+        Parameters
+        ----------
+        file_name: str
+            A file name to process.
+
+        Returns
+        -------
+        int:
+            A run number of the file.
+        """
```
dneise: What if the original MAGIC file has been renamed e.g. for testing purposes from its original file name to `test_magic_file.root`. I assume in that case this method would raise an exception. 
Would it be possible to fail gracefully in this case and return maybe `nan` or `-1` or `"unkown"` or something like that?
So that one can read events from a file, even if the run-number cannot be found out?

ctapipe/io/magiceventsource.py
```diff
@@ -0,0 +1,1337 @@
...
+        # Loop over the available data runs
+        for run_number in self.run_numbers:
+
+            # Removing the previously read data run from memory
+            if self.current_run is not None:
+                if 'data' in self.current_run:
+                    del self.current_run['data']
+
+            # Setting the new active run
+            self.current_run = self._set_active_run(run_number)
```
dneise: `self.current_run` seems to be only used locally in this method and not in other methods of the same class, so there seems to be no need for it to be an instance variable. I think it can just be a local.

ctapipe/io/magiceventsource.py
```diff
@@ -0,0 +1,1337 @@
...
+                pointing_zd = input_file['Events']['MPointingPos.fZd'].array() - input_file['Events']['MPointingPos.fDevZd'].array()
+                pointing_az = input_file['Events']['MPointingPos.fAz'].array() - input_file['Events']['MPointingPos.fDevAz'].array()
+                pointing_ra = (input_file['Events']['MPointingPos.fRa'].array() - input_file['Events']['MPointingPos.fDevHa'].array()) * 15.
+                pointing_dec = input_file['Events']['MPointingPos.fDec'].array() - input_file['Events']['MPointingPos.fDevDec'].array()
+            else:
+                # Getting the telescope drive info
+                drive_mjd = input_file['Drive']['MReportDrive.fMjd'].array()
+                drive_zd = input_file['Drive']['MReportDrive.fCurrentZd'].array()
+                drive_az = input_file['Drive']['MReportDrive.fCurrentAz'].array()
+                drive_ra = input_file['Drive']['MReportDrive.fRa'].array() * 15.
```
dneise: If you could explain the magic number 15 here to a future reader of the code, I assume that would help readability a lot. 
One might put a comment or even nicer give a to this number, like:
```
number_of_foos_per_bar = 15
drive_ra = input_file['Drive']['MReportDrive.fRa'].array() * number_of_foos_per_bar
```
here the variable name `number_of_foos_per_bar` implies to the reader that `"MReportDrive.fRa"` is obviously measured in bars, while `drive_ra` is measured in foos and therefore the multiplication with a conversion constant was necessary. 

Anyway some kind of explanation of a magic number would be great I guess.

ctapipe/io/magiceventsource.py
```diff
@@ -0,0 +1,1337 @@
...
+        dict:
+            A dictionary of pedestal event IDs in M1/2 separately.
+        """
+
+        mono_ids = dict()
+        mono_ids['M1'] = []
+        mono_ids['M2'] = []
+
+        m1_data_condition = self.event_data['M1']['trigger_pattern'] == 128
+        m2_data_condition = self.event_data['M2']['trigger_pattern'] == 128
```
dneise: The "magic number" `128` is now repeated here for the 4th or 5th time ... I think it might be good to replace them by a nicely named costant.

ctapipe/io/magiceventsource.py
```diff
@@ -0,0 +1,1337 @@
...
+
+        event_id = self.mono_ids[telescope][mono_event_num]
+        file_num = np.digitize([event_id], self.event_data[telescope]['file_edges'])
+        file_num = file_num[0] - 1
+
+        return file_num
+
+    def get_pedestal_event_data(self, pedestal_event_num, telescope):
+        """
+        This method read the photon content and arrival time (per pixel)
```
dneise: If you like, you could remove this repetitive "this method" part from all function (or method) docstrings and try to use the "imperative" form for the first line of a docstring.

I think both PEP8 and PEP257 mention that the imperative form is preferred for the first line of a docstring, but I found this page described it also very nicely.
https://blog.dolphm.com/pep257-good-python-docstrings-by-example/

dneise:
I think this PR indeed looks very good. I just mentioned a few minor things.

ctapipe/io/containers.py
IevgenVovk: Some telescopes (e.g. MAGIC) provide this information right away. Can remove it if you insist.

ctapipe/io/magiceventsource.py
IevgenVovk: There has to be at least two files - one for Magic-I and another for Magic-II; these are normally linked to each other via the run number. So if there run number is not there, proper data reading can not be guaranteed anyway.
if the run number is set to nan/-1/unknown etc, the corresponding files can not be found further in _set_active_run(). Maybe a specially tailored exception would be a better option here? Say, like so:
 ```
try:
    run_number = int(parsed_info[0])
except IndexError:
    raise IndexError('Can not identify the run number of the file {:s}'.format(file_name))
```

File: ctapipe/io/containers.py (hunk 3 above)
Comment (MaxNoe) [reply ctapipe/io/containers.py]: FACT also handles pointing in equatorial coordinates and transforms to horizontal as needed.
So it might make sense to keep all four and fill all of them from the data available.

File: ctapipe/io/magiceventsource.py (hunk 26 above)
Comment (dneise) [reply ctapipe/io/magiceventsource.py]: Thank you for the clarification. Indeed after reading more of this code I understood how big the influence of the files path to the reader is. So one cannot simply rename a MAGIC file and assume it can still be read, the name (or complete path?) is a part of the file format as it contains important information as you pointed out. 

So I would like to mark this comment as **resolved**, but github does not have this feature.

**Comment by dneise** [comment]
Hello @IevgenVovk thanks for the changes so far. They look very good indeed. Just making `uproot` an optional dependency did apparently not work. 

Maybe you'd like to have a look at `hessioeventsource.py` or `sst1meventsource.py` with me, both of them make use of an optional dependency. I think one should note that `try import optional_dependency except NameError: log; raise` does not make a dependency into an *optional* dependency. 
What makes a dependency optional is to design the entire code in this case the code of a particular `EventSource` in a way that everything that can work, also does work, even when the optional dependency is missing. Of course an exception must be raised and will be raised whenever a user tries to do something which cannot work at all without the optional dependency.

---

Here you can see the `__init__` of the `HESSIOEventSource`, the import is wrapped into a `try ... except` but still .. if `pyhessio` is not installed ... an exception will be raised as soon as `__init__` is called ... just accompanied with a nice log message:

https://github.com/cta-observatory/ctapipe/blob/74b0838aa99c34a13601e4caeca1f9c75a1e4f51/ctapipe/io/hessioeventsource.py#L22-L30

Here you see another example of the same idea, this time for `SST1MEventSource`. The `try .. except` is missing, but still the fact that the `import ...` is in the `__init__` function and not in the head of the file ensures that the file can for example be imported even the the optional dependency is not installed.

https://github.com/cta-observatory/ctapipe/blob/74b0838aa99c34a13601e4caeca1f9c75a1e4f51/ctapipe/io/sst1meventsource.py#L15-L19

**is_compatible?**

One thing is important here, both `HESSIOEventSource` and `SST1MEventSource` took care to avoid using their optional dependencies within the `@staticmethod is_compatible(file_path)` because it is important that this method will run (and return either True or False ... and not raise an exception) even when the optional dependency is missing: 

This is how `HESSIOEventSource` did it, you see no use of the optional dependency `pyhessio` in this method:

https://github.com/cta-observatory/ctapipe/blob/74b0838aa99c34a13601e4caeca1f9c75a1e4f51/ctapipe/io/hessioeventsource.py#L42-L53

And this is how `SST1MEventSource` did it, they replaced the option dependency `protozfits` by the `fits`-library from the non-optional `astropy` package for this method:

https://github.com/cta-observatory/ctapipe/blob/74b0838aa99c34a13601e4caeca1f9c75a1e4f51/ctapipe/io/sst1meventsource.py#L79-L101

----

# TLDL;

Please do not use `uproot` in the `is_compatible` method. I would say: If you have to use it, wrap it into a try except block and return False. But maybe ... **maybe** ... you could check the filename? since the filename is fundamentally important for a MAGIC file .. one can maybe already say: Well ... I did not even look *into* the file .. but the filename is already so broken ... this cannot be a valid MAGIC file. and return False. 

And only if the filename looks good you try to import `uproot`.

Maybe this would help?

**Comment by dneise** [comment]
Very nice, we are close to the point where the current set of unit tests are passing. As soon as the existing tests all pass again, one might want to add new tests, dedicated for the MAGIC files and make them pass.

The only thing still making one test fail is around line 371 in `magiceventsource.py`:

    >     self.log.error(msg)
    E           AttributeError: 'MAGICEventSourceROOT' object has no attribute 'log'

This is easy to fix. `self` does not *yet* have a `log` since this is added by the `__init__` of the parent class `EventSource`. So only after this line: 

    super().__init__(config=config, tool=tool, input_url=file_list[0], **kwargs)

One can start using members which are added by the parent. 

So all one needs to do is move some lines around and one is good to go into the next round. :+1:

File: ctapipe/io/magiceventsource.py [hunk 31]
```diff
@@ -0,0 +1,1352 @@
...
+            intrument_attr = file.attrs['instrument']
+            if intrument_attr.tostring() == b"MAGIC":
+                return True
+            else:
+                return False
+            
+    def __exit__(self, exc_type, exc_val, exc_tb):
+        self.file.close()
+
+    def _generator(self):
```
MaxNoe: I just learned, that this should be used by users as public API, so it should not start with an `_`.

Maybe a fitting name would be `iter_stereo_events`

ctapipe/io/magiceventsource.py
```diff
@@ -0,0 +1,1352 @@
...
+        this_run_mask = input_path + '/*{:d}*root'.format(run_number)
+
+        run = dict()
+        run['number'] = run_number
+        run['read_events'] = 0
+        run['data'] = MarsDataRun(run_file_mask=this_run_mask)
+
+        return run
+
+    def _generator(self):
```
MaxNoe: I think this should be removed as it just calls another generator

ctapipe/io/magiceventsource.py
```diff
@@ -0,0 +1,1352 @@
...
+        generator instance.
+
+        Returns
+        -------
+
+        """
+
+        return self._stereo_event_generator()
+
+    def _stereo_event_generator(self):
```
MaxNoe: If this is intended to be called by users, it should not start with an underscore.

I would propose to call it `iter_stereo_events`

ctapipe/io/magiceventsource.py
```diff
@@ -0,0 +1,1352 @@
...
+
+                # Setting the instrument sub-array
+                data.inst.subarray = self.magic_subarray
+
+                yield data
+                counter += 1
+
+        return
+
+    def _mono_event_generator(self, telescope):
```
MaxNoe: I would propose to call this `iter_mono_events`

ctapipe/io/magiceventsource.py
```diff
@@ -0,0 +1,1352 @@
...
+
+                # Setting the instrument sub-array
+                data.inst.subarray = self.magic_subarray
+
+                yield data
+                counter += 1
+
+        return
+
+    def _pedestal_event_generator(self, telescope):
```
MaxNoe: I would propose to call this `iter_pedestal_events`

ctapipe/io/magiceventsource.py
MaxNoe: no. It could also overwrite `__iter__` to call  `iter_stereo_events()`

MaxNoe: Requiring a generator does not make sense for every eventsource, imho.

MaxNoe: An EventSource could also be a simple iterable, not a generator

watsonjj: > An EventSource could also be a simple iterable, not a generator

Sure, but the iterable should act as a generator. ctapipe works on a single event container at a time...the functionality of the EventSource is that it returns one event container at a time.

ctapipe/io/magiceventsource.py
```diff
@@ -0,0 +1,1352 @@
+
+
+import glob
+import re
+
+import h5py
```
kosack: this adds another dependency of H5py which is not used in ctapipe so far (we use pytables for HDF5 access, and I would like to keep the required dependencies to a minimum).  Can you make that optional as with uproot?

ctapipe/io/magiceventsource.py
kosack: I'm going to suggest at the framework meeting that we switch to pathlib for all file handling. It's so much nicer.
```py3
import pathlib

input_path = pathlib.Path(input_url)
input_path.exists()  # -> True
input_path.parent # -> the parent dirname
input_path.name # -> the filename 
input_path.stat().st_size # -> file size
```
etc.

ctapipe/io/magiceventsource.py
MaxNoe: As the reading using uproot works well, do you need both event sources?

dneise: h5py is by default anaconda and it is already in the evironment.yml in the master:
https://github.com/cta-observatory/ctapipe/blob/e7152482461d1f386f3a998e6ac567e21bcb93f7/environment.yml#L14

ctapipe/io/magiceventsource.py
IevgenVovk: Awaiting for your coming discussion, I've implemented Max's proposal for now.

ctapipe/io/magiceventsource.py
IevgenVovk: Fixed.

ctapipe/io/magiceventsource.py
IevgenVovk: Removed...

ctapipe/io/magiceventsource.py
IevgenVovk: We didn't want to touch the EventSource as this could have affected others. Maybe this can be fixed with the next patch?

ctapipe/io/magiceventsource.py
IevgenVovk: Done.

ctapipe/io/magiceventsource.py
IevgenVovk: Done.

ctapipe/io/magiceventsource.py
```diff
@@ -0,0 +1,1337 @@
...
+        int:
+            A run number of the file.
+        """
+
+        mask = ".*\d+_M\d+_(\d+)\.\d+_.*"
+        parsed_info = re.findall(mask, file_name)
+
+        run_number = int(parsed_info[0])
+
+        return run_number
```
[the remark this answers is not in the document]
IevgenVovk: I felt it's more natural to have this method's code in each class to improve its clarity. The price - code duplication - is minor...

ctapipe/io/magiceventsource.py
IevgenVovk: Done.

ctapipe/io/magiceventsource.py
IevgenVovk: Fixed.

ctapipe/io/magiceventsource.py
IevgenVovk: Thanks for pointing this out. I'll be sure to follow this in future commits.

ctapipe/io/magiceventsource.py
IevgenVovk: OK.

ctapipe/io/magiceventsource.py
IevgenVovk: OK.

ctapipe/io/magiceventsource.py
IevgenVovk: OK.

dneise:
:+1:  .. I think the tests should pass now  ... this is exciting

moritzhuetten:
Just going through the comments to the MAGICEventSourceHDF5 class. After the discussions at the LST bootcamp last week I am not sure whether it is helpful to merge that one into the main repo. There are efforts to define a unified (HDF5) format (including @labsaha at UCM) to read DL1 data (DL0,DL2,...) and once such a format is defined, the corresponding reader would probably obtain its own instrument-independent reader class. In turn, the presence of such a reader at this point at this place might be confusing. Let us know if we (i) should comply with the comments for this pull request and possibly remove the HDF5 reader later, (ii) completely remove it now or (iii) leave a generic skeleton MAGICEventSourceHDF5 class. Thanks!

MaxNoe:
@moritzhuetten If that's the case, I would remove it now I guess.

moritzhuetten:
@MaxNoe ok, will remove it.

IevgenVovk:
Just a heads-up that after a discussion within the team and the comments raised here we've removed the HDF5 reader from our pull request.

dneise:
Is more work planned for this PR? Or should it be closed?

moritzhuetten:
What about keeping it open until the plugin design for event sources is ready (and then move it to a plugin, as you suggested)? Just as a reminder to correctly finish this task at some point.

ctapipe/io/magiceventsource.py
MaxNoe: > And what is saved by requesting this eventsource subclass to override __iter__ instead of defining _generator?

That we use the standard python why? Principal of least suprise.

MaxNoe: > In what circumstance would an eventsource not utilise a generator? 

A generator is something very specific. An EventSource imho only needs to be an iterable of events. 

In the most simple case, i would say an EventSource only needs to implement the iterator protocol.

watsonjj: This is more a question of the EventSource design, not how the MAGIC EventSource should be implemented. It would be more appropriate to discuss this separately.

When designing the EventSource, it was requested that the base EventSource class handles the max_events check, therefore avoiding every EventSource developer having to define it. I think this was a good suggestion, and the best way I saw to do that was to have the max_events checked in `__iter__`. This therefore required a separate function to `__iter__` that could be overridden in the EventSource subclass - hence `_generator`.

If you have improvements to suggest to EventSource then it would be more helpful to open a PR or issue on the subject. The changes you are requesting here go against the current EventSource design.

MaxNoe:
The plugin system is now merged. I would suggest to move this implementation into it's own repository called `ctapipe_io_magic`

noahhdf:
Is there ongoing work for something like this `ctapipe_io_magic` plugin? I would really like to read some magic files using this implementation. If not, I will try to take care of it

IevgenVovk:
Yes, it's been delayed due to some other commitments we've had earlier, but we're working on it.

IevgenVovk:
Heads-up: https://github.com/cta-observatory/ctapipe_io_magic

moritzhuetten:
Hi @dneise, hi all, after @IevgenVovk has committed the ``ctapipe_io_magic`` plugin, this PR now can  probably be closed.

Issue #727, opened 2018-04-25
coordinate classes misused in many places


kosack:
I noticed that a lot of places where `ctapipe.coordinate` classes are used, they are not used correctly as described in the astropy docs.  In the misused cases, the coordinate *Frame* is used to constuct the coordinates, not the generic `SkyCoord`  class (which is what all coordinates should be). 

All coordinates should be `astropy.coordinates.SkyCoord` instances, and it is not correct to construct the coordinates using the frame directly.   The correct way to construct a coordinate (or array of them) is:

```py3
az, alt  = 180 * u.deg, 80 * u.deg

pointdir = SkyCoord(az=az, alt=alt, frame='altaz')
tilted = TiltedGroundFrame(pointing_direction=pointdir)

# build a list of coordinates in the GroundFrame:
tel_coords = SkyCoord(x=[-1,0,1], y=[0,1,-1], z=[0,0,0], frame=GroundFrame)

# transform them to our new frame
tel_coords.transform_to(tilted)

# transform to a modified frame:
tilted.pointing_direction = SkyCoord(alt=10*u.deg, az=90*u.deg, frame='altaz')
tel_coords.transform_to(tilted)
```

We should update all places where coordinates are used to correctly construct them as `SkyCoords` and keep frames separate, as this may break code in future updates of astropy.

ParsonsRD:
No I don't and this is not what I suggest, in fact it's quite the opposite. The vast majority of the functionality and the checks performed by the astropy coordinates are unnecessary, that's why it's so slow. I instead propose to write a bare bones implementation of the coordinate transformations which does what we want and nothing else, is easier to use (for example each coordinate frame should already contain enough information to convert of any other compatible system) and isn't dominated in execution time by either the instantiation of the coordinate frame or checking whether the units we use are compatible.

I agree that premature optimisation is not a great idea, but this code has sat in the repository for > 1 year and isn't used in many sections of the code because it's either too hard to use or far too slow!

MaxNoe:
> that's why it's so slow.

That's not correct. The main part why astropy transformations are slow is the ludicrous precision with which the conversions are performed and a missing optimization of the calculation of time dependent parameters.

There is active work done in providing a way to optimize for speed rather precision on micro arc second level, which is discussed here:

https://github.com/astropy/astropy/pull/6068

kosack:
[...]

PR #842, opened 2018-11-29
Add test how it should be

MaxNoe:
This test shows how one should use our coordinate frames using skycoord

### DIFF
--- a/ctapipe/coordinates/tests/test_coordinates.py
+++ b/ctapipe/coordinates/tests/test_coordinates.py
@@ -1,5 +1,19 @@
 import numpy as np
 import astropy.units as u
+from astropy.coordinates import SkyCoord
+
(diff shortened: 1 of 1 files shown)


thomasgas:
Just to add a comment: this PR goes in the way of having the coordinate frames behaving correctly (and being used correctly, see #727) with the astropy SkyCoord class.
Solutions are:
1. use SkyCoord from astropy with small fixes with respect to the present implementation.
2. refactor coordinates in order to make them fast (a first implementation came from Dan: #753) 

Both approaches needs to be consistent across the whole ctapipe (which is not the case right now). Tests, notebooks and classes using coordinates have to be updated accordingly.
It would be nice if also @kosack and @ParsonsRD could comment on this.

PR #857, opened 2018-12-02
SPE Fitting

watsonjj:
I would like to contribute the fitting classes I use for SPE spectra.

Initially contributed are a base class and two subclasses, one for PMTs and the other for SiPMs (using the formula defined in Gentile 2010).

The base class is defined to be flexible, allowing the easy definition of a new subclass containing a different SPE fitting formula to be contributed.

The base class can simultaneously fit N illuminations, allowing for a better handle on the fit parameters (as many should not change with average illumination).

The iminuit package is used to perform the fit, and the confidence on the resulting fit parameters is also extracted from the covariance matrix.

### DIFF
--- a/ctapipe/analysis/camera/spe_fitter/__init__.py
+++ b/ctapipe/analysis/camera/spe_fitter/__init__.py
@@ -1,0 +1,1 @@
+from .base import SPEFitter
--- a/ctapipe/analysis/camera/spe_fitter/base.py
+++ b/ctapipe/analysis/camera/spe_fitter/base.py
@@ -1,0 +1,334 @@
+from abc import abstractmethod
+import iminuit
+import numpy as np
+from scipy.stats.distributions import poisson
+import yaml
+import warnings
+
+
+class SPEFitterMeta(type):
+    def __call__(cls, *args, **kwargs):
+        obj = type.__call__(cls, *args, **kwargs)
+        obj._post_init()
+        return obj
+
+
+class SPEFitter(metaclass=SPEFitterMeta):
+    def __init__(self, n_illuminations, config_path=None):
+        """
+        Base class for fitters of Single-Photoelectron spectra. Built to
+        flexibly handle any number of illuminations simultaneously.
+
+        Parameters
+        ----------
+        n_illuminations : int
+            Number of illuminations to fit simultaneously
+        config_path : str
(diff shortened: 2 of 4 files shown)


watsonjj:
Developments required:

- [ ] Transform the class into a Component
- [ ] Consider a replacement for the metaclass
- [ ] Create a Tool (located in tools or examples?) that demonstrates the use of the class
- [ ] Use multiprocessing to fit each pixel in parallel
- [ ] Add functionality to incrementally contribute to the histogram before fitting (to be more in line with how ctapipe reads an event at a time)
- [ ] Possible to get jit to work inside a class to avoid the global parameters?
- [ ] Okay to keep config file, or should parameters be converted into traitlets?
- [ ] Obtain a implementation of the Borel fitting approach for SiPMs
- [ ] Add tests
- [ ] Improve documentation

kosack:
One (very general) question about the overall workflow for diagnostics: do we want to do things like histogramming in side the event loop at all, or should we separate it into 2 steps?  In all of our current pipeline diagnostics, we simply write out a DL1 or DL2 event file first, and then use something like Pandas to do the histogramming on the events as a separate step operating on that file (note that multiple files may be chained for statistic reasons). 

That is somewhat more flexible (with the expense of larger output files), since you can run other diagnostics or stages of the pipeline in the same way without having to re-do the event-loop and calibration. 

Of course for per-pixel data, this may not be so simple, since Pandas doesn't support vector columns (though astropy.table and the HDF5 pytables we use do, but doesn't have histogramming built in), so we do still need some custom code to deal with fitting per-pixel quantities.   If the output is first an HDF5 table written by a HDF5TableWriter for each event, there are many utilities that can rapidly pass through that data and make histograms.  I much prefer that to making many custom event loops.

So my suggested workflow is:
1. run event loop to write out a DL1 file for a particular camera (can also write a separate file/dataset for each camera)
2. generate a pixel-wise histogram from the processed event data, writing out the histograms
3. fit the histograms 

The last 2 steps can also be combined so that the histograms are just passed to the next stage in-memory, to avoid some output if we like, but it's not required.

watsonjj:
I will pick this up again once the DL1 writer is in place

watsonjj:
SPE fitting will be implemented in #1415

PR #862, opened 2018-12-06
Implemented FACT image cleaning

LukasNickel:
As pointed out on the recent LST Analysis Bootcamp we might want to access different cleaning methods from MAGIC and FACT within ctapipe.

This cleaning method is based on the the one used at FACT, a java implementation can be at the FACT github repo:
https://github.com/fact-project/fact-tools

The results for simple gamma showers seem to be comparable with the tailcuts_clean method with these cleaning levels used for both methods:
cleaning_level = {
    'LSTCam': (3.5, 7.5, 2),
    'NectarCam': (3, 5.5, 2),
    'DigiCam': (2, 4.5, 2),
}

- performs cleaning based on pixel value and timing information
- default values are rather loose, higher picture and boundary thresholds could prove useful
- added unit test

### DIFF
--- a/ctapipe/image/cleaning.py
+++ b/ctapipe/image/cleaning.py
@@ -227,7 +227,7 @@ def fact_image_cleaning(geom, image, arrival_times, picture_threshold=4,
 
     # Step 3
     pixels_above_boundary = image >= boundary_threshold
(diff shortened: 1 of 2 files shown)


mackaiver:
Add reference to http://dx.doi.org/10.17877/DE290R-17773.

PR #864, opened 2018-12-07
SimTelEventSource using pyeventio

MaxNoe:
As an alternative to pyhessio, we add this EventSource based on pyeventio

We have a tests that shows that for all events in gamma_test.simtel.gz and gamma_test_large.simtel.gz both deliver the same results.


mackaiver:
We decided to replace the pyhessio with pyeventio and simply replace it in the EventSource. this means we only have to change the code in the event source.

PR #812, opened 2018-10-25
Updates of NectarCam and LSTCam real data eventsource class

FrancaCassol:
Added the instrument information to the real data containers. The camera geometry files have been adde to the ctapipe-extra repository. They are:
- LSTCam-001.camgeom.fits.gz = present geometry of the LST1 camera (001 indicate the version)
- NectarCam-001.camgeom.fits.gz = present geometry of the NectarCam prototype (001 indicate the version)

The optics is the one of MC data.

Issue #747, opened 2018-05-24
MON & SVC data in ctapipe 


FrancaCassol:
Hi,
writing the reader for LSTCam data I run into a conceptual problem: LST data files include a CameraConfiguration header (one per file) that contains data that could be consider SVC or MON data, and can be read/set one time per file. Wondering on where to place them into the containers, I realized that in ctapipe the EVT, MON and SVC concepts are still missing. We have for example

r0.tel[m].waveform 
and not 
r0.tel[m].evt.waveform 

so, in the case of LST specific data, I plan to introduce the evt and svc containers, like

lst.tel[0].evt.xxx --> for event specific data
lst.tel[0].svc.yyy --> for camera specific data

what do you think, is there anything wrong with that?

Cheers

Franca

kosack:
so far, the event structure is supposed ot be for only EVT data, though we have mixed in some SVC header info (e.g. `mcheader`, and `inst.subarray`) that will be removed in the next version, and handled separately from the event data.  

The current way we do things is that all header info is part of the instrument config, and eventually only things that change per event should be in the event container.  *however*, the containers do have a concept of meta-data headers, which you can access via e.g. `event.meta`, or `event.r0.meta`, and anything in those meta dicts will be written to the output file's header.  So probably for now that is where you want to put this info. It is free-form, so no need for any structure to it, just key=value pairs.  

Can you give examples of what data this is? That would help to structure things better.

FrancaCassol:
Hi Karl,
too late ;-)  I implemented the SVC data in a LSTServiceContainer included in container.py that is filled in lsteventsource.py, you can see there the fields implemented till now (some are still not completely explicited, e.g. algorithms or pre_proc_algorithms, and others are still missing because I wait for the new EventBuilder version)... but no problem to change. Some of the data are indeed instrument data (e.g. num_pixels which can now be reached as event.lst.tel[telid].svc.num_pixels) and other are more software-like data (e.g. data_model_version which can be reached as  event.lst.tel[telid].svc.data_model_version)... Let me know what do you suggest

PR #870, opened 2018-12-11
Flat-field camera corrections

FrancaCassol:
New code to calculate flat-field coefficients from flasher calibration events.

For that, new containers are defined in container.py
- MonitorDataContainer : general monitor data container
- FlatFieldContainer and FlatFieldCameraContainer: containers for the flat-field calculated coefficients

The code is based on a FlatFieldFactory (place in calib/camera) which permits to select a specific FlatFieldCalculator.

For the moment one calculator is implemented:
- FlasherFlatFieldCalculator : based on the best algorithm described by S. Fegan in MST-CAM-TN-0060
 
Also, a tool is provided in ctapipe/examples:
- calc_flatfield.py : which selects the waveform cleaner, the charge integrator and the desired  FlatFieldCalculator. It performs a loop on the events and writes the  FlatFieldCameraContainer in a hdf5 file.

NB: this PR will work only when  PR #812 will be accepted (help on that side is very welcome)


kosack:
Since FF values are SVC data and do not change per event hopefully, does it make sense to use a Container to store them rather than just passing them to the CameraCalibrator as input? That may be simpler. 

However, for MON quantities like pedestals that *do* change in time on second or so timescales,
I always had in mind we would have some sort of interpolator class that reads MON quantities like pedestals vs time from an input table, and can output an interpolated value for a single event based on the timestamp.  I guess that could be used to fill these containers per event.  Is that the idea?

so when processing events, we'd have something like the following (oversimplified pseudocode):
```py3
# read in pre-calculated pedestals
ped_interpolator = PedestalInterpolator("pedestals_obs012345.fits.gz")
...

# fill in the mon container for an event
for tel_id in tels:
    event.mon.pedestal.tel[tel_id] = ped_interpolator[tel_id](event.dl0.time)
```

I think probably we need to decide on a common way to handle MON and SVC data.  I'll open an issue about that separately.  

For now, could you describe the data flow here? It's not clear to me.

kosack:
ah, I see #747 probably answered my question (I forgot that).  I guess for now this method is ok, but eventually we will move away from using a container for SVC data, since it is not changing by event.

FrancaCassol:
The idea is that FF will be continuously updated during the data taking (thanks to  interleaved flasher events at some tens of Hz), like the pedestals values. So, they must better be considered as MON data. The container will be filled per "flat-field event" (that corresponds to a certain statistics (100?1000?) of flasher events and/or to a certain time of data taking (15-30 s?)). The flat-field container  is then useful both to pass the FF values to the CameraCalibrator and to easily write on disk the flat-field events for archiving... Something similar should happen to the pedestals interleaved events. 
Both FF and pedestals will also be used for selecting bad pixels (temporary failing or with high NSB, i.e from stars) that will be either totally excluded or (if isolated) interpolated with neighbouring pixels (to keep trace of the bad pixels I introduced the pixel_status array in the r0 container).  

Concerning  the time interpolation you suggest, it will be probably useful if the MON data change more rapidly than the time they are updated, otherwise we can just use the last values contained in the MON containers (that should in principle correspond to some tens of seconds before, hopefully this is enough ...).

.travis.yml
```diff
@@ -54,6 +54,7 @@ install:
     - pip install travis-sphinx
     - pip install codecov
     # ----- SST1M:
+    - conda install protobuf
```
dneise: line 58 will install protobuf, so we do not need line 57

examples/calc_flatfield.py
```diff
@@ -0,0 +1,112 @@
+"""
+Extract flat field coefficients from flasher data files.
```
dneise: I believe it would be good to mention in this doc string here, how this example should be executed. 
Don't hesitate to be very specific in an example. E.g. mention real and exact file-names. Even if a possible user, might not have that file. They still gained some knowledge here, namely which file they need to ask for.

ctapipe/calib/camera/flatfield.py
```diff
@@ -0,0 +1,247 @@
...
+
+class FlasherFlatFieldCalculator(FlatFieldCalculator):
+    """
+    Class for calculating flat field coefficients witht the
+    flasher data. Fills the MON.flatfield container.
+    """
+
+    def __init__(self, config=None, tool=None, **kwargs):
+        """
+        Parent class for the flat-field calculators.
```
dneise: This is lying documentation as a result of a copy+paste action I assume. We as developers must be very careful not to fall into copy & paste traps, because there is a rule, that documentation must be written. 

Look at this particular case: the `config` and `tool` parameters are found all over our code base. Does it really add value, if we copy and paste their description into every Tool and Component we have? 

If we remove the description of `config` and `tool` what is left?
```
Parent class for the flat-field calculators.
Fills the MON.flatfield container.
kwargs
```

Of which the 1st line is a lie. The 2nd is already contained in the class-doc-string. The 3rd line is actually interesting, as it could explain what are the other parameters for this Component... but unfortunately it was omitted.
I assume a doc-string which is visually already very long is assumed to be complete ... another trap which we should try to avoid, I believe.

dneise:
@FrancaCassol I am currently reviewing this PR **actively**, that means when ever I find things which are easy to fix. I do so right away and push them into your branch. This should speed up the entire process a bit.

For you this means: if you want to work further on this branch, just do not forget to:
 * git pull ... 
 * work just as before .. git add .. git commit ..
 * and git push

If you forget to git pull before ... your git will prevent you from pushing .. until you pulled.

dneise:
EventSources have a parameter called `max_events`, and it is defined as:
```
max_events : int
        Maximum number of events to loop through in generator
```

The new FlatfieldCalculator introduced in this PR also uses `max_events` as a parameter name, but with a different meaning:
```
    max_events = Int(10000, help='Define the maximum number of events per '
           ' coefficient flat-filed calculation').tag(config=True)
```

I think it should be renamed like `sample_size` or so.

ctapipe/calib/camera/flatfield.py
```diff
@@ -0,0 +1,247 @@
...
+        integral, peakpos = self._extract_charge(event, tel_id)
+
+        # remember the charge
+        self.trace_integral[self.count] = integral
+
+        # remember the time
+        self.trace_time[self.count] = peakpos
+
+        # keep the mask of not working pixels (to be improved)
+        self.trace_mask[self.count] = [pixel_status == 0, pixel_status == 0]
```
dneise: why do we need a list of 2x `pixel_status == 0` again? I forgot ... why is this duplication needed?

dneise: Ah, got it, not needed at all .. can be replaced by:

    self.trace_mask[self.count, :] = pixel_status == 0

FrancaCassol:
> @FrancaCassol I am currently reviewing this PR **actively**, that means when ever I find things which are easy to fix. I do so right away and push them into your branch. This should speed up the entire process a bit.
> 
> For you this means: if you want to work further on this branch, just do not forget to:
> 
>     * git pull ...
> 
>     * work just as before .. git add .. git commit ..
> 
>     * and git push
> 
> 
> If you forget to git pull before ... your git will prevent you from pushing .. until you pulled.

Thanks a lot Dominik, 
I will go through your suggestions and perform a new push (after the pull ;-))

dneise:
Okay Franca ... however, you do not have to do anything at this stage .. I am working on it.
Currently I am trying to break out a few functions, to avoid having bodies of if-clauses in the order of 50 lines.
I think this is hard to do for the original author, since the one who wrote the code often finds it very clear to read...

dneise:
@FrancaCassol I believe there is some problem here

If we have a look at the test, which nicely shows how to **use** the new class, which is introduced in this PR:
https://github.com/cta-observatory/ctapipe/blob/a9063a47cf48f3cf2dd958215af6bbe3bef099b9/ctapipe/calib/camera/tests/test_flatfield.py#L15-L20

We can nicely see that `calculate_relative_gain(event, tel_id)` will be called with the same event for multiple different `tel_id`s. (The fact, that the example file used for the test only contains one single telescope is from the look at the code not obvious and must be considered a special case)

Inside `calculate_relative_gain` the waveforms of different telescopes are mixed:

We see here the waveform is taken out of the event container:
https://github.com/FrancaCassol/ctapipe/blob/e1f69ec9906b467a8978af81d9ce10c6873b9095/ctapipe/calib/camera/flatfield.py#L164

From that point on `waveform` is used for the analysis, e.g. it is integrated to get the charge. This charge is then accumulated for multiple events using an internal `self.count`er, which counts up every time   `calculate_relative_gain` is called. So **if** we take the example from the test seriously, we must assume this function is called with the same event once per telescope, which will either end in wrong results or will raise and exception because the number of pixels for different telescopes differs.

Of course every function returns "wrong" results, when it is called with wrong input and in this case, giving multiple different telescope ids to a single `FlasherFlatFieldCalculator` is simply wrong input. So the implementation of `calculate_relative_gain` is actually kind of fine, just a bit surprising maybe.

I think it would be less surprising if the user would specify the `tel-id` at construction time of the `FlasherFlatFieldCalculator` once and not with every call to `calculate_relative_gain`, since this implies that there are use-cases where people would actually like to mix events from multiple Telescops together in one single `FlasherFlatFieldCalculator`. Is this the case?

Also the test should be adjusted, in order to clearly convey to the reader, that one needs **one FlasherFlatFieldCalculator per tel_id** on not one per event stream as it is now. 

Do you agree with this impression of mine, or did I misunderstand something?

dneise:
Ah .. another possibility would of course be to move the loop over tel_id completely into `calculate_relative_gain` and then return flatfielding constants for every tel_id we ever saw. 
If this was the goal, then this goal is currently not reached.

FrancaCassol:
Hi Dominik, 
sorry for my late reply, I am back now on the subject after a somehow long absence ;-) Yes, I totally agree with you, I prefer the first option: the tel-id should be defined at construction time and avoid most probably  unrealistic loops on several telescopes at this point. How do we proceed? Should I make the changes?

dneise:
> Should I make the changes?

Yes please, if you have the time.

FrancaCassol:
Hi Dominik,

is the code correct now?

Thanks

dneise:
If it's "correct" I cannot tell, but I think it is better now. Less likely to go wrong when used on a multi-teleskope file.

examples/calc_flatfield.py
```diff
@@ -0,0 +1,103 @@
...
+from ctapipe.io import HDF5TableWriter
+from ctapipe.core import Tool
+from ctapipe.io import EventSourceFactory
+
+from ctapipe.image import ChargeExtractorFactory, WaveformCleanerFactory
+from ctapipe.calib.camera.flatfield import FlatFieldFactory
+from ctapipe.io.containers import MonitorDataContainer
+
+
+class FlatFieldGenerator(Tool):
```
MaxNoe: Generator has a very specific meaning in Python. This class is not a Generator in that sense.
It would be better to rename it. (maybe to  `FlatFieldHDFWriter`? )

MaxNoe: Generator has a very specific meaning in Python. This class is not a Generator in that sense.
It would be better to rename it. (maybe to `FlatFieldHDFWriter`? )

FrancaCassol:
Hi,

can this PR be accepted? 

Thanks 

Cheers

Franca

dneise:
Ah I had approved already ... but new commits invalidate my approval. ... let me quickly look what has changed.

FrancaCassol:
Hi Dominik, 
Waiting for the PR to be accepted, I just corrected some style errors as suggested travis. It was probably a bad idea. I am sorry for that. Actually, it is not clear to me where the problem is now.
Do you think is it possible to approve and accept the PR in one go? That would avoid to make this type of mistakes ;-) 

Please, let menage to finish with it in a short time (I have colleagues that are waiting for that). Thanks!

dneise:
> I just corrected some style errors as suggested

That is always a good idea.

But something seems to have been broken by accident... sorry I was distracted... I will now look

FrancaCassol:
Dominik, wait to accept I think there is something wrong with the BaselineWaveformCleaner, I want to verify before

dneise:
Good morning. For some reason the tests fail ... I am not yet sure what the reason is, but we need to fix this before we can merge this PR anyway.

dneise:
I still have no idea where this error comes from

FrancaCassol:
For me also is very strange, I don't remember I made important content changes in my last push ...
 I am now testing the code with MC data and I released that there are some differences with real data.
For example, the event.r0.tel.trigger_time is not set, I will use the event.trig.gps_time for them (which  is still not available for data). I will make a new push as soon as I will have checked that everything is ok...

dneise:
The current master seems to have a problem, this is why the tests for this PR also fail ... so this is most probably not your fault at all. I am currently investigating what happened.

FrancaCassol:
Thanks Dominik, I wish you to find easily  the problem.

I see that you made a merge in my repository, can I make a pull it to my local repository and then a push again? (I have some corrections to push to the PR coming from the MC test)

dneise:
> can I make a pull it to my local repository and then a push again

yes, sure.

dneise:
I think we need #925 before we can expect the tests to work again

dneise:
Very good @FrancaCassol  .. I think we are close to have the master fixed again .. so that this PR can finally also be merged.

FrancaCassol:
Dominik, I am going to move the test_pedestal.

dneise:
Okay do whatever you like ... just for the moment do not (yet) expect the tests to pass

FrancaCassol:
Hi,
I have now a "strategy" question: I have  a new version of the LST reader (necessary for a full interpretation of the data, e.g. from TIB) that I would like to pull, can I merge it to this PR?

dneise:
Ah Hello Franca, regarding the new LST reader.

Generally, the smaller a PR is the easier and quicker it should be to merge it, i.e. the more topic one mixes within one PR the longer it takes to discuss it. 

In this specific case it is even more difficult. ctapipe would like to move these optional dependencies, like the SST1M, the Target and also the LST into their own packages, which will then be plugins for ctapipe. 

For you as the author of the reader this means you have complete control over your own plugin package.
I will soonish start to make an examples based on the SST1M reader so people can have a look at it.

dneise:
@FrancaCassol  The tests fail .. but I do not know why. 

There seems to be an issue with the order(?) of tests, I believe. I can re-create the failing tests on my laptop when I do just:

    pytest

but when I only execute the failing test file like this:

    pytest ctapipe/image/tests/test_waveform_cleaning.py

then the tests do not fail. This is very puzzling for me. Maybe you have an idea where this issue might come from?

FrancaCassol:
Hi Dominik,

> Generally, the smaller a PR is the easier and quicker it should be to merge it, i.e. the more topic one mixes within one PR the longer it takes to discuss it.

I agree totally with you, but I think the problem is the time that a PR takes to be accepted. I was waiting for this PR to be accepted before asking a new one, but now I have such an amount of new l (initially small) changes that the next PR will be again rather important. I even wonder if it make sense to continue with the present one because it is outdated with respect to my code... should I  better start with a new one?
 
> In this specific case it is even more difficult. ctapipe would like to move these optional dependencies, like the SST1M, the Target and also the LST into their own packages, which will then be plugins for ctapipe.
> 
> For you as the author of the reader this means you have complete control over your own plugin package.
This is interesting ... are we going to discuss on that at the next A&S WG meeting?  

> I will soonish start to make an examples based on the SST1M reader so people can have a look at it.
Wonderful :-)

FrancaCassol:
> then the tests do not fail. This is very puzzling for me. Maybe you have an idea where this issue might come from?

For me it happens the same ... strange, it looks like the example_event changes ...

kosack:
Yes, sorry this PR is being slow mainly because there is current work to move all EventSource "plugins" (e.g. those that need optional external libraries like Zfits) to separate repos, to make testing faster in the main ctapipe repo, and to avoid having to install so many dependencies.  So soon this code will be external (probably a single module like ctapipe_io_zfits that contains both lst and nectarcam readers).

FrancaCassol:
Hi Karl,
let's say that this is the present source of delay ;-)

Just to be sure to understand: the message is: "wait till a bit"?

FrancaCassol:
It was decided to perform a successive PR, when the code will be more advanced and tested in the context of the lst-chain repository.

PR #880, opened 2018-12-11
Remove hessio.py, use HESSIOEventSource consistently

MaxNoe:


mackaiver:
change them to use the `event_source`helper function.

PR #655, opened 2018-02-20
Additional metadata from pyhessio

morcuended:
Added new parameters from pyhessio to MCHeaderContainer (as asked in [image-extractor#14](https://github.com/cta-observatory/image-extractor/issues/14)):
  - CORSIKA and sim_telarray version
  - energy range of primary particle
  - magnetic field
  - spectral index
  - prod site information

Check [this](https://github.com/morcuended/ctapipe/blob/master/examples/notebooks/Inspect_Provenance_Metadata_Info.ipynb) notebook.

Note that some of the implemented fields (*prod_site_array, prod_site_coord , prod_site_subarray*) are still not retrieving the required information from pyhessio.

PR #895, opened 2018-12-12
Additional metadata from pyhessio

dneise:
#655 was very good in my opinion and we did not implement in the `SimTelEventSource` in the first place.


dneise:
NB: I like what Gernot said [here](https://github.com/cta-observatory/ctapipe/pull/655#issuecomment-368258835) wrt #655.

Issue #505, opened 2017-09-08
Why do CameraFrame and HorizonFrame need an `ArrayPointing`


MaxNoe:
Transformation from Camera Coordinates to sky coordinates should only require the pointing of the single telescopes, or not?

kosack:
I think the coordinate transforms are a bit backward and need reworking (see e.g. #416)

ParsonsRD:
Short answer, they don't. However if they are defined here then a direct transformation can be performed all they way through all the systems. e.g. from camera directly to Horizon.

Agreed this needs restructuring though.

MaxNoe:
Why does a transform from `CameraFrame` to `AltAz` need an `ArrayPointing`? 

The individual pointing of the Telescope should be enough.

ParsonsRD:
Yes, it doesn't but some of the frames in between do. Currently the coordinates transform doesn't bypass any systems, such as the system centred on the nominal array pointing direction, but that is easy to change in the astropy structure.

MaxNoe:
fixed in #896

thomasgas:
Solved in #896

Issue #416, opened 2017-05-23
CameraGeometry should store pixel coordinates as SkyCoord


kosack:
I would like to store the camera coordinates inside `CameraGeometry` as a `SkyCoord` (not really sky, but that's the base class for astropy coordinates, even if confusingly named), but as of right now, this is difficult because when talking about a `CameraGeometry`, we do not necessarily have the focal length of the telescope in which the camera is installed.  

Therefore, I would like to have `CameraFrame` be a base frame that does not need any parameters (it's just pixel pos in meters, relative to the center of the camera).  Then to transform to an angular system where the pixels are positions on the sky (e.g. the TelescopeSystem), I would then add information about the focal length only.  This system should be the same as CameraSystem, but angular coordinates, still centered on the center of the telescope's FOV.

Finally, I would later go to a AltAz or nominal by adding information on the pointing position of the telescope, etc.

@ParsonsRD is there a reason why it is not done this way? Right now I need to have the focal length to specify a CameraSystem coordinate, but I don't have that info, and it shouldn't be needed.

kosack:
I guess, to me it makes sense like this:
```
 CameraFrame ---[telescope optics]---> TelescopeFrame ---[pointing]---> NominalFrame
```

kosack:
This would then allow us to store the frame type in the CameraGeometry class (without loading a telescope optics description), and support the various frames described in #255

ParsonsRD:
Hi, sorry was on vacation for the last week. I think thats fine and shouldn't cause any serious problems.

MaxNoe:
I recently implented a working set of Coordinate Transformation for FACT here: 
https://github.com/fact-project/pyfact/blob/master/fact/coordinates/camera_frame.py

I learned quite a lot about how astropy frames are supposed to work and what members they need to have to make transformation work painlessly.

I started working on implementing these in ctapipe, but right now I cannot give a timeframe when I will open a PR for this. 

In the mean time, i'd like to know if there are definitions of the camera coordinate system and telescope coordinate system somewhere.
This would be very helpful for writing tests and the actual implementation.

kosack:
I think @tino-michael is debugging that right now for his reconstruction, and found for example the extra 90 deg rotation that we are missing to go from the sim_telarray coord sys to the sky, etc.  He's not using astropy coordinates yet, but once he has them finalized, they should be translated into them.

kosack:
see #580

labsaha:
[...]

PR #896, opened 2018-12-12
Major coordinate refactoring, fixes #899, #900, #505, #416

### LINKED ISSUE #899: Use `SkyOffsetFrame` for TelescopeFrame and NominalFrame
It sound like we could just use astropy's `SkyOffsetFrame` for `TelescopeFrame` and `NominalFrame`, as these are basically AltAz Frames offset from the telescope pointing position or an arbitrary sky position.

MaxNoe:
Starting the cleaning up of the coordinates together with @thomasgas


kosack:
take a look at #416 if you haven't already

MaxNoe:
It is now possible to go directly from ICRS to `CameraFrame`:

```py3
from ctapipe.coordinates import CameraFrame, HorizonFrame

obstime = Time('2013-11-01T03:00')
location = EarthLocation.of_site('Roque de los Muchachos')
horizon_frame = HorizonFrame(location=location, obstime=obstime)

# simulate crab "on" observations
crab = SkyCoord(ra='05h34m31.94s', dec='22d00m52.2s')
telescope_pointing = crab.transform_to(horizon_frame)

camera_frame = CameraFrame(
   focal_length=28 * u.m,
   telescope_pointing=telescope_pointing,
   location=location, obstime=obstime,
)

ceta_tauri = SkyCoord(ra='5h37m38.6854231s', dec='21d08m33.158804s')
ceta_tauri_camera = ceta_tauri.transform_to(camera_frame)

camera_center = SkyCoord(0 * u.m, 0 * u.m, frame=camera_frame)

# assert ceta tauri is in FoV
assert camera_center.separation_3d(ceta_tauri_camera) < u.Quantity(0.6 * u.m)
```

MaxNoe:
I think this is basically ready for review.
I made a number of significant changes:

* `NominalFrame` and `TelescopeFrame` are now `UnitSphericalFrames`, which made defining them a lot easier, the implementation is close to what astropy defines as `SkyOffsetCoordinates`. x and y are true angles in these frames.

* Astropy provides specific attributes for frames, e.g. `TimeAttribute` or `CoordinateAttribute`, these are now used where appropriate.

* `HorizonFrame` is simply an alias for astropy's `AltAz` and could be removed completely

* I implemented a hack, that allows usage of `HorizonFrame` without setting `obstime` and `location`. If they are not set, transformations assume they are the same for the `from` and `to` frame. 
At the moment a warning is raised if transformations are done without setting `obstime` or `location`, we could think of dropping the warning, it's quite annoying in the tests.

ctapipe/coordinates/telescope_frame.py
```diff
@@ -0,0 +1,97 @@
...
+)
+
+from .horizon_frame import HorizonFrame
+
+
+class TelescopeFrame(BaseCoordinateFrame):
+    """
+    Telescope coordinate frame.
+
+    Cartesian system describing the angular offset of a given position in reference
```
mackaiver: non cartesian

ctapipe/coordinates/telescope_frame.py
```diff
@@ -0,0 +1,97 @@
...
+    to the pointing direction of a given telescope.
+
+    This makes use of small angle approximations of the position of interest and
```
mackaiver: fix lying docs.

ctapipe/coordinates/tests/test_coordinates.py
```diff
@@ -9,13 +13,81 @@ def test_cam_to_nominal():
...
-    nom_frame = NominalFrame(array_direction=array_pointing)
+    nom_frame = NominalFrame(origin=array_pointing)
     cam.transform_to(nom_frame)
 
 
+def test_icrs_to_camera():
+    from ctapipe.coordinates import CameraFrame, HorizonFrame
+
+    obstime = Time('2013-11-01T03:00')
+    location = EarthLocation.of_site('Roque de los Muchachos')
```
mackaiver: Does this test need an internet connection? Or is `of_site` also working offline?

ParsonsRD:
Hi, great work! I fully agree with the changes made here. I think the change of the nominal system for something like a sky offset frame makes most sense. I was mentioned yesterday on the ASWG call that tests are beginning for divergent pointing and we should be sure that this works nicely with the coordinates classes.

Also it's great that we can directly transform to RA DEC from these coordinates. When this is done I think the astropy code is applying corrections to account for the refraction of light through the atmosphere, these corrections can get quite large at large zenith angle. However in IACTs the light is not traversing the whole atmosphere but only the final ~10km, which we found in HESS can make a significant difference. So we should make sure we can add our own atmospheric profiles here so the correction is done correctly.

This is not a big deal for now, but I will make an issue about it just to make sure it's not forgotten.

MaxNoe:
Shower images transformed to nominal and hexbinned:

nominal

MaxNoe:
Travis fails because bulding the docs produces a warning that there is no API definition for `HorizonFrame` which is simply an alias for `astropy.coordinates.altaz`.

Anyone has an idea how to solve that, besides changing everything to just use altaz?

ctapipe/coordinates/camera_frame.py
```diff
@@ -0,0 +1,110 @@
...
+def camera_to_telescope(camera_coord, telescope_frame):
+    '''
+    Transformation between CameraFrame and TelescopeFrame.
+    Is called when a SkyCoord is transformed from CameraFrame into TelescopeFrame
+    '''
+    x_pos = camera_coord.cartesian.x
+    y_pos = camera_coord.cartesian.y
+
+    rot = camera_coord.rotation
+    if rot == 0:
```
kosack: does it make sense to compare a float to exactly 0? This rotation angle may be computed elsewhere and subject to rounding error.  maybe `np.isclose()` should be used instead

kosack:
*The following might be addressed here or in a future PR:*

Can we also fix the issue that the camera coordinates are now stored in the Corsika system with the x-coordinate and y-coordinate flipped? It would be nice to at least define a few extra camera systems for engineering purposes  that will allow us to plot the camera in a system where for example Y is up when the telescope is at 0 deg elevation and one is looking at the back of front of the camera (taking into account 1 or 2-mirror systems).  At the very least, we should make it easy to plot the camera in a more common system with X-Y swapped.  (note this also implies perhaps rethinking the visualization to allow multiple systems to be used in CameraDisplay, but that would be a separate PR)

ParsonsRD:
I can't remember why I made HorizonSystem, it may have been in case we had to implement the refraction correction ourselves. But if this code needs changing it might be better to contribute to astropy. I think we can delete it and just use altaz.

thomasgas:
> _The following might be addressed here or in a future PR:_
> 
> Can we also fix the issue that the camera coordinates are now stored in the Corsika system with the x-coordinate and y-coordinate flipped? It would be nice to at least define a few extra camera systems for engineering purposes that will allow us to plot the camera in a system where for example Y is up when the telescope is at 0 deg elevation and one is looking at the back of front of the camera (taking into account 1 or 2-mirror systems). At the very least, we should make it easy to plot the camera in a more common system with X-Y swapped. (note this also implies perhaps rethinking the visualization to allow multiple systems to be used in CameraDisplay, but that would be a separate PR)

@kosack This is related to the issue #875 assigned to me from the Dortmung meeting. I'm now working on a notebook to plot the pixels of the cameras together with their IDs. 
I guess we can start to work on a document for the coordinates: @MaxNoe can add to the docs some details regarding the different reference frames and their astropy implementation and I'll add plots for the telescopes and the cameras, both in 2d and 3d with labels and arrows.
When this document is ready, reviewed and approved we can use it to create the docs...sounds right?

ctapipe/coordinates/camera_frame.py
vuillaut: Actually, what is the purpose of this test?
If `rot==0`, don't you get anyway
```
x_rotated = x_pos
y_rotated = y_pos
```
?

MaxNoe: It's to not call sin and cos to spare some computing time

MaxNoe: I don't like it either, it was there before.

ParsonsRD: Yeah this was to save computing time in the mist common case that no rotation is applies, feel free to get rid if you think it confuses the code. Ultimately the rotation shouldn't be needed, but for some cameras the camera coordinates passed needed to be rotated (this may even be fixed already).

MaxNoe: It works. In most cases the camera rotation is exactly zero, so no problems with the float comparison

MaxNoe:
I tested the coordinate transforms using FACT data we took in 2015.

We pointed the telescope to bright stars and recorded random triggers.
I calculated a value we (and I think veritas) calls PedVar, which is proportional to the amount of background light.

One can clearly see the stars in the FoV (most notably the brightest star in the center, we observed Aldebaran in this example).

I transformed all stars from the hipparcos catalog into the `CameraFrame` using ctapipe.


For this to work, I had to create a new `CameraGeometry` for FACT, where I changed `x -> -x`, `y -> -y`.
Is this expected? The FACT coordinate frame is defined as: Looking from the dish onto the camera, x is right, y is up.

MaxNoe:
I found the mistake, x in telescope frame was along the azimuth but x in camera frame is along altitude for the hess coordinate frame.

To avoid this confusion, I renamed `x` and `y` of the telescope frame and nominal frame to `delta_az` and `delta_alt`, this also makes clear that these are offset frames using angles.

kosack:
> For this to work, I had to create a new CameraGeometry for FACT, where I changed x -> -x, y -> -y.
>Is this expected? The FACT coordinate frame is defined as: Looking from the dish onto the camera, x is right, y is up.

Currently we use the Corsika system for the cameras, in which X is up, so this points to the fact that we need to define a transform between the corisika and more traditional systems in order to make this work in general. it should not require re-defining the camera geometry (except in this case it seems the FACT convention is different from the Corsika one, so they should at least all be in the same system)

MaxNoe:
Ah sorry, this was all cleared up here #875 and via mail with @moralejo. Should be all consistent now, together with the fix of FACT's geometry in ctapipe-extra

kosack:
Are we just waiting for another review of this, or is there more to discuss?

MaxNoe:
You asked for updated docs.

I wrote down what I know about the current definitions, but not more.

Issue #905, opened 2018-12-14
sst-1m labelled as SC


thomasgas:
Using a simtel file from Prod3b (`gamma_20deg_0deg_run100___cta-prod3_desert-2150m-Paranal-merged.simtel.gz`) I just found out that all telescopes SST-1M are labelled as Schwarzschild-Couder (SC) instead of Davies-Cotton (DC).
With this code:
```
optics = events[2].inst.subarray.tel[430].optics
camera = events[2].inst.subarray.tel[430].camera

print(camera)
print(optics)
print(optics.mirror_type)
```
I get:
```
DigiCam
SST-1M
SC
```
And this is true for all the telescopes:
```
event_table = event.inst.subarray.to_table()
event_table[event_table['tel_subtype'] == "1M"]["mirror_type"]  == "SC"
```
is true for all the 158 SST-1M in the hyper-array in Prod3b.

thomasgas:
@bregeon and/or @GernotMaier can you comment on this?

MaxNoe:
This does not come from the mcs themselfs but from the guessing lookuptable in ctapipe:

https://github.com/cta-observatory/ctapipe/blob/16457f090dd50d4c2c2b2b0a12dfabeb2517c7f5/ctapipe/instrument/optics.py#L12

thomasgas:
ok thanks @MaxNoe! i'll do a PR to solve this issue.

thomasgas:
solved with #908.

PR #908, opened 2018-12-18
Fix SST-1M to be DC and not SC. Solves #905

thomasgas:


MaxNoe:
But LST is not Davies Cotton afaik, its parabolic

thomasgas:
> But LST is not Davies Cotton afaik, its parabolic

You're right. So i guess that by now "DC" is kind of an alias for "1-mirror-telescope" and "SC" is for "2-mirror-telescope". Which is clearly wrong...
what shall we do with this LUT? fill it with correct info or create it from the mc?

kosack:
This info is there since one is not always reading an MC file, so you need a lookup for it. Also at the time, it wasn't possible to read it from the MCs - is there now some accessor? 

For the names, I agree it's not ideal - we can easily change the optics type names, however, since they are not fixed or used anywhere yet.  The main reason was to be able to adapt algorithms to different types, which could be simply the number of mirrors, but also may have other aspects such as off-axis corrections.  I'm happy if we update these to be correct, or use better names.    

* SC: Swartzschild-Couder
* DC: Davies Cotton
* might need another for modified Davies-Cotton (e.g. LST) - is there a name for that optics type?

thomasgas:
so do you want to merge this or do you want to change LST optics.mirror_type in order to be consistent?

MaxNoe:
Apply this simple fix now, i'd say, rest later

GernotMaier:
For completeness: the 1-mirror MST has a 'modified DC design' (something between DC and parabolic).

I don't think this matters anywhere (unless one needs e.g. the smearing in time due to the optics or if one wants to fold the expected optical PSF into the image reconstruction), and knowing that it is a one or two-mirror instrument should be enough.

PR #910, opened 2018-12-20
Do not set autodownload, fixes doc build

MaxNoe:


kosack:
do you know what changed to cause this to break in the doc build? It's worked for years. Though I agree it's probably  not needed in the docs.

MaxNoe:
The server was simply down for about three days.

MaxNoe:
See https://github.com/astropy/astropy/issues/8304

Issue #915, opened 2019-01-08
Issues with Factories


dneise:
I had the impression our Component Factories had some issues (mainly a gut feeling), so I took a closer look and I believe I identified a few real issues, which are more than just my gut telling me that our current implementation is too complex. However the main driver of this issue is a persistent feeling of complexity.

I understand our current implementation of `Factory` in `core/factory.py` is driven mainly by the following use case: "Users of `Tool`s should be able to choose and configure specific implementations of certain tasks by name". 

Another important use case is certainly letting the factory decide what is the "correct" implementation for that task at hand, this is e.g. the case when the correct event source is automatically selector from the url, or when the correct extractor is selected based on the given event source.

So I set out to find an alternative implementation, which fulfills these use cases while not using meta programming.

The other issues I found in the process were these:

In case multiple factory products define traitlets of the same name, then `FactoryMeta` silently keeps only one of them and silently drops the rest. I do not know what happens if traitlets of the same name but different type are defined. 

As a consequence of this users cannot inspect different defaults of factory products, since the factory exposes only the first.

Well I worked a bit on an alternative implementation and will open a PR now for discussion.
I do not assume that this first try will already be better than our current implementation, just different. I believe it can be healthy to see different versions in order to have a fruitful discussion.

watsonjj:
Related to #821

watsonjj:
> In case multiple factory products define traitlets of the same name, then FactoryMeta silently keeps only one of them and silently drops the rest. I do not know what happens if traitlets of the same name but different type are defined.

It would be possible to simplify the existing `Factory` implementation considerably, and fix this issue you have pointed out, if we don't rely on the `Factory` to replicate and pass the traitlets to the Components. Instead, one would need to directly specify the traitlets, identically to how its now done in #915.

I am still really in favour of keeping the Factory class separate to a Component. I don't think every Component should be automatically be given Factory behaviour...

MaxNoe:
#916 got merged, so closing this

Issue #844, opened 2018-11-29
Use a plug-in system for Factories?


kosack:
- right now the Factories are somewhat fragile (depend on order of  imports)
- would be nice to have a nicer convention for plugins for things like  IO sources, etc, that could even be in separate packages  (e.g. ctapipe-magic or ctapipe-hess packages would install readers  for those telescopes).

Similar things exist for other large python frameworks, so there is probably a nice way to do this

MaxNoe:
Done.

Issue #821, opened 2018-11-19
Why Factories?


MaxNoe:
At the moment, to get a correct analysis, one has to specify the product of Factories:

```python
event_source = EventSourceFactory.produce(
        input_url=input_file,
        max_events=n_events if n_events > 1 else None,
        product='HESSIOEventSource',
)
```

What then is the difference to directly using `HessioEventSource`?

watsonjj:
This surprises me... the whole point of the factory is that it finds the correct EventSource based on the input_url. The `product` traitlet exists only to force a particular return object.

watsonjj:
If the `EventSourceFactory` is not correctly returning `HESSIOEventSource` for the simtelarray file (without specifying the `product`), then the `is_compatible` function needs revisiting. Something may have changed to the simtelarray file definition....

MaxNoe:
Ok, seems to have been false alarm. Sorry.

@mackaiver always put the keys there and I remember there was an issue at least with `CameraCalibrator` doing something wrong if not provided with the products. 

On the current master, it seems to work.

kosack:
Should we close this issue, since it seems to be a misunderstanding? I thikn the use case for Factories is quite clear and reasonable (since we have now 4+ data formats to support)

kosack:
You can always use the shortcut `event_source()` helper function as well to avoid thinking about factories and products, though it's less flexible when used in a Tool since the Config doesn't propegate.

MaxNoe:
I should like another Interface.

I think it would be great if `EventSource(path)` would return the correct subclass.

watsonjj:
I considered such a functionality when designing the factories. But I found it important to separate the base class from the factory class. One reason was to avoid the subclasses of EventSource and ChargeExtractor from inheriting the factory behaviour.

Personally, I don't see what is gained over the `event_source()` helper function. But I look forward to seeing your proposed implementation.

kosack:
Hiding the Factory design pattern in the base class would be very confusing - that's not how most python classes work, and almost certainly would create circular dependencies, since the base class would need to know about all sub-classes (but sub-classes, which are plug-ins can be in different files, that require the base class to be loaded first), so I'm not sure how easy it would be to make something like that.

kosack:
In any case, this may be connected to #844

MaxNoe:
closed by #916

Issue #217, opened 2016-10-14
Factory classes


watsonjj:
A mechanism for handling factory classes needs to be created so that a particular container can be chosen based on user input.
E.g. Which ChargeExtractor should be used (LocalPeakIntegration, GlobalPeakIntegration, etc.)

How cmdline arguments are handled based on these factory classes need to be considered. Essentially some cmdline arguments (specific to a particular ChargeExtractor Container) only exist depending on which ChargeExtractor is chosen.

watsonjj:
An idea for a factory class can be found in `ChargeExtractorFactory` at #219 in `ctapipe/calib/camera/charge_extractor.py`.

watsonjj:
A Factory class now exists in core/factory.py

PR #916, opened 2019-01-08
Alternative Take on Factories

dneise:
This PR is a "design study" aiming to reduce apparent code complexity in the context of `Factory`. As mentioned by Max, myself and some other devs had problems understanding that part of the code. 

Goals are: 
 
 - remove complexity (less meta-programming)
 - remove lines of code
 - keep features

What are the features I want to keep working in particular? 

 - Construct specific implementations by name, from user-config
 - Construct specific implementations as needed, like finding the right EventSource given a file-path.
 - Retain a meaningful `--help-all` for users of tools, so they know what's going on.

This PR does not aim to add or remove tested functionality. If it does, this is a mistake. Untested functionality might be modified (Since I simply have no idea, what aspect of the current master implements a wanted feature, and what is just there "by accident")

This PR is related to #915, #821, #217

And very very loosely connected to  #844

---

**EDIT**

I think this study is now ready for closer review. I tried to make the diff as small and thus readable as possible. Still the main modifications should be listed here:

 - new `from ctapipe.core.subclass_from_name(base_cls, name)` function
 - `camera.r1.camera_r1_calibrator_for_eventsource`: makes an R1 Calibrator, given an eventsource
 - `ctapipe.io.event_source()`: makes an EventSource, given a file path.
 - modifications in Tools, see discussion below.

## Discussion

During this design study some things, which we did in the master within the `Factory` class fell apart into maybe 3 different parts now. Which I believe is a sign that in some sense we violated the [SRP](https://en.wikipedia.org/wiki/Single_responsibility_principle) with the `Factory` implementation.

These 3 parts are:
 1. create products by name
 1. create products by special considerations
 1. form a traitlet frontend/stand-in for its products in Tool creation

### creating products by name

`product = getattr(namespace, 'name')(*args, **kwargs)` is all, what is needed to create a product from a name, given one knows the namespace. So a little convenience function `subclass_from_name` was created which took over this part from the former Factory.
The current master Factory knows *defaults* for cases where a product should be created by name, **without actually giving a name**. I think this is not good, so I removed this behavior.

### creating products by special considerations

An example is here: creating an EventSource from a files path. This is unrelated from making a an EventSource from its name. Also it is unrelated from implementing a command line interface. Its implementation is a loop over all possible EventSources evaluating the  `is_compatible` staticmethod. The implementation is 3 lines and can be understood without following an inheritance tree.

### form a traitlet frontend/stand-in for its products in Tool creation

This is use case, for tool developers only, i.e. people who develop command line applications to be used by non-developers. The main features needed to provide useful command line interfaces (CLI) are: 

 1. allow to select certain *algorithms* **by name**
 1. allow to **configure** these algorithms

The *algorithms* are implemented of course as specific implementations of certain interfaces, e.g. the `GlobalPeakIntegrator` is a special implementation of the `ChargeExtractor` interface. 

Most of these features are already provided to us by the `traitlets.config` machinery. A tool developer must however be able to express his intend, to allow the tool user the configuration of a whole list of algorithms, for this a convenient way of listing algorithms was needed. In our master branch this intend is expressed by stating e.g. the `ChargeExtractorFactory` in the `classes` list. I propose to construct this list using a helper function: `classes_with_traits(ChargeExtractor)`. Alternatively the tool developer is of course free to list the algorithms which should be exposed to the tool user explicitely. It might be possible that certain algorithms do not make sense for certain tools. In that case neither the `ChargeExtractorFactory` nor the `classes_with_traits` function is of great help.

In addition to listing all the available algorithms one thing more is needed, the tool user must be allowed to type in a name on the command line, which is then used to instantiate this algorithm. The `traitlets.config` machinery needs one traitlet to exist for every command line option. So the tool developer needs a method to create such a traitlet of type `CaselessStrEnum`. For this also a convenience function has been created:
`enum_trait(ChargeExtractor)`. In the master creating this traitlet was not necessary, since the e.g. `ChargeExtractorFactory` already exposed a string traitlet named `product`.


watsonjj:
This looks very interesting! I have a few questions:

1. What are the limitations you have encountered with this alternative, compared to the current `Factory`?
2. Currently it seems there isn't a way to specify the chosen subclass from the command-line via the traitlets (unless one creates a traitlet manually in the tool), is that correct? 
3. More of a statement than a question, I see you have issue using the input_url traitlet of EventSource in `from_config` without first initialising an instance. I am trying to think of a workaround, but I'm unable to at the moment.
4. Do the traitlets of subclasses (to ChargeExtractor for example) get correctly set from the command line. In other words, in a script can one activate an alias for the `lwt` traitlet of `NeighbourPeakIntegrator`, and specify a value for it on the command line. Will that value be passed to `NeighbourPeakIntegrator` if it is obtained from `ChargeExtractor.from_name`?
5. Would this have trouble with subclasses that are defined in other files, or as plugin packages?
6. Do you see a problem with developers being able to do something wrong like `NeighbourPeakIntegrator.from_name`? As it has inherited that method.

watsonjj:
Thanks for this Dominik, I appreciate the effort you've put into showing an alternative. I admit the current `Factory` implementation is not completely perfect, and I like the simplifying of the code in your approach and the avoiding of the issue of duplicate product traitlet names that you highlighted in #915 . However, I must say that keeping a distinction between what is a `Factory` class and what is a `Component` class is still appealing in my mind.

I look forward to the further discussion and input from others about the direction for Factories

watsonjj:
> Currently it seems there isn't a way to specify the chosen subclass from the command-line via the traitlets (unless one creates a traitlet manually in the tool), is that correct?

> I am not sure I understand the question correctly.

Sorry, let me clarify. In `BokehFileViewer` you've replaced
```python
extractor='ChargeExtractorFactory.product',
```
with 
```python
extractor_name = tool_utils.enum_trait(ChargeExtractor)
...
extractor='BokehFileViewer.extractor_name',
```
i.e. instead of having the traitlet for choosing the `Component` defined in the factory class itself, you create the traitlet in every tool you need it. Maybe this is the only sensible way for this factory approach, but its something extra we will require the `Tool` creator to remember to do themself.

dneise:
> ... its something extra we will require the Tool creator to remember to do themself. 

Indeed this is the case. 

> However, I must say that keeping a distinction between what is a Factory class and what is a Component class is still appealing in my mind.

I think here we differ in our opinions. I believe in Python factories can be as simple as a dict-lookup (which is the implementation of from_name) or a little if-clause, often there is no need to role out the complete abstract factory pattern just to accomplish a certain goal. In particular a thing should not need to have any knowledge about somebody planning to construct it using a factory, as should a thing-developer not need to have any knowledge that somebody else might want to utilize a factory to construct said thing. 
With our current implementation I had the feeling that the Factories were pretty loud about their existence and developers had to keep them in mind, but maybe that is just a personal impression. Also end-API-users were exposed to an implementation detail, which the use of the factory pattern ultimately is. I believe this should not need to be the case. 

In the current implementation, there is no need for the helper function `event_source` anymore end users can immediately call `EventSource.from_url` which has exactly the same interface as `event_source()` and a in my opinion a fairly understandable name, albeit a bit longer (20 instead of 12 characters).

Anyway, I believe attaching `from_url` and `from_name` to the baseclass, so that it is inherited by all specific implementations where its existence is utterly useless and thus very confusing is a clear flaw of this version here, which needs to be overcome.

ctapipe/calib/camera/r1.py
```diff
@@ -110,6 +104,19 @@ def check_r0_exists(self, event, telid):
                 self._r0_empty_warn = True
             return False
 
+    @classmethod
+    def from_eventsource(cls, eventsource=None, *args, **kwargs):
```
MaxNoe: maybe call it `for_eventsource`? I know that `from_` is very widespread for alternative constructors, but semantically, we want a given calibrator **for** our eventsource

ctapipe/io/eventsource.py
```diff
@@ -198,3 +196,37 @@ def __enter__(self):
...
+        compatible_cls = cls.cls_from_url(url)
+        return compatible_cls(input_url=url, *args, **kwargs)
+
+    @classmethod
+    def cls_from_url(cls, url):
+        '''find compatible EventSource sub-class for `url`'''
+        for subcls in cls.__subclasses__():
+            if subcls.is_compatible(url):
+                return subcls
+        raise ValueError
```
MaxNoe: Add an error message here

ctapipe/io/eventsource.py
```diff
@@ -198,3 +196,37 @@ def __enter__(self):
 
     def __exit__(self, exc_type, exc_val, exc_tb):
         pass
+
+    @classmethod
+    def from_url(cls, url, *args, **kwargs):
```
MaxNoe: Maybe also call it `for_`? Same reason as above.

dneise:
I did some research again regarding this question:

> Would this have trouble with subclasses that are defined in other files, or as plugin packages?

I looked at the example of the `EventSourceFactory` in the current master. At the top of the file, these imports are needed: 
```
# EventFileReader imports so that EventFileReaderFactory can see them
# (they need to exist in the global namespace)
import ctapipe.io.hessioeventsource
from . import sst1meventsource
from . import nectarcameventsource
from . import lsteventsource
import ctapipe.io.targetioeventsource
```

I removed them from this place and imported them later (in this case, I imported them at the top of `file_viewer.py`) and as expected, this was too late, the `EventSourceFactory` did not know anything about these specific EventSources. So in case new EventSources are developed and added to ctapipe, EventSource developers must not forget to import them in the top of `eventsourcefactory.py`. 

The current implementation behaves differently (a bit), here it does not matter when these imports are done as long as they are done and as long as they are done before the call to `EventSource.from_name`. This might or might not be an advantage, since now we have a little more freedom to detect new EventSources. 

(I must admit I totally forgot about these imports) so for some time, this current version here, did not know anything about these other event sources. 

However, now I've put these imports into `ctapipe.io.__ini__.py` and e.g. this works

    ctapipe-event-viewer -f ~/anaconda3/envs/cta/lib/python3.6/site-packages/protozfits/tests/resources/example_SST1M_20180822.fits.fz

---

This means, as soon as we start working with real plugins for certain Components, we seem to be less restricted as to when we do the plugin-detection. It is not a use-case for us, but it would even be possible to let the user specify names of python packages on the tool command line, which are then imported and treated as plugins, and that would still work...

ctapipe/calib/camera/r1.py
dneise: sounds good to me.

dneise:
Thanks for your nice words @MaxNoe. So working on this is maybe not a total waste of time...

As I am just reminded by CodeCov, that I am missing some tests, I will add these now. Maybe in that process further problems show up. :-D

ctapipe/io/eventsource.py
MaxNoe: Actually, I wouldn't call it url if it does not support urls.

watsonjj: It is referring to the `input_url` though

watsonjj:
> > ... its something extra we will require the Tool creator to remember to do themself. 
> 
> Indeed this is the case. 

Could something different be done here, to be more similar to the existing factory? Could there be a traitlet added to every Component called `subclass` or `product` that allow immediate selection of a subclass when the `extractor='ChargeExtractorFactory.product',` alias is included in the tool. This way the tool developer does not need to remember to do:

```python
extractor_name = tool_utils.enum_trait(ChargeExtractor)
...
extractor='ChargeResolutionGenerator.extractor_name',
...
    classes = List(
        [
            SimTelEventSource,
            CameraDL1Calibrator,
            ChargeResolutionCalculator
        ] + tool_utils.classes_with_traits(ChargeExtractor)
    )
...
extractor = ChargeExtractor.from_name(self.extractor_name, **kwargs)
```

And instead does something like

```python
extractor='ChargeResolutionGenerator.product',
...
    classes = List(
        [
            SimTelEventSource,
            CameraDL1Calibrator,
            ChargeResolutionCalculator,
            ChargeExtractor
         ]
    )
...
extractor = ChargeExtractor.from_config(**kwargs)
```

dneise:
At the moment sphinx fails for py3.5 I am not sure if this is my fault. It seems to fail while loading "intersphinx directories" which is a hint that this might not be my fault, right?
sphinx_fail

ctapipe/calib/camera/calibrator.py
```diff
@@ -25,27 +30,15 @@ class CameraCalibrator(Component):
...
-        ped='CameraR1CalibratorFactory.pedestal_path',
-        tf='CameraR1CalibratorFactory.tf_path',
-        pe='CameraR1CalibratorFactory.adc2pe_path',
-        extractor='ChargeExtractorFactory.extractor',
-        extractor_t0='ChargeExtractorFactory.t0',
-        window_width='ChargeExtractorFactory.window_width',
-        window_shift='ChargeExtractorFactory.window_shift',
-        sig_amp_cut_HG='ChargeExtractorFactory.sig_amp_cut_HG',
-        sig_amp_cut_LG='ChargeExtractorFactory.sig_amp_cut_LG',
-        lwt='ChargeExtractorFactory.lwt',
```
watsonjj: Could you restore all of these aliases?

I guess it would need to refer to the actual class where these aliases are created? For example in place of `ped='CameraR1CalibratorFactory.pedestal_path',` one would do `ped='TargetIOR1Calibrator.pedestal_path',`

I suppose some traitlets may need to be moved to their parent classes to avoid the need to repeat this for every R1Calibrator that uses a `pedestal_path`. This is for a different PR.

ctapipe/calib/camera/r1.py
```diff
@@ -313,65 +320,7 @@ def real_calibrate(self, event):
...
-            if es:
-                if es.metadata['is_simulation']:
-                    return 'HESSIOR1Calibrator'
-                elif es.__class__.__name__ == "TargetIOEventSource":
-                    return 'TargetIOR1Calibrator'
-            return 'NullR1Calibrator'
+__all__ = [
+    cls.__name__
+    for cls in non_abstract_children(CameraR1Calibrator)
+]
```
kosack: is it possible to put this boilerplate code in a metaclass as we did before? Then, it's transparent to a user, and we don't have to remember to cut and paste this in.

watsonjj:
> Could something different be done here, to be more similar to the existing factory? 

Continuing on from my comment here, if this is implemented, then we don't necessarily need `from_name`, `from_url`, and `from_config`. We just need a single `from_config`, and that classmethod does all the factory behaviour using the traitlets.

kosack:
A couple of comments:

- Can you give a short top-level description in this PR of what this new implementation does and what problems it solves? (perhaps edit the top-level PR description, which currently says nothing useful).  I haven't done a detailed code review, but It still seems like a kind of "factory" design pattern, but with a different implementation.  I guess the difference is now the factory is not a class, but a class method (e.g. `for_url()` or `from_name()`)?

- I think the issue of plugins can perhaps be separated from this - a plugin loader could run before the Tool is fully constructed, so that we register the subclasses properly. That is almost certainly necessary if we have plugin sub-classes in separate python modules that are not known by the core package.  I don't think that changes the design here, though.

ctapipe/calib/camera/calibrator.py
kosack: I think we should write down what we want here: I would say, having the ability to set configuration parameters from the base class and any extra ones in the sub-class is a nice thing, and the developer should not have to repeat them in each case. But maybe this points more to a problem with the configuration system.  I'm still in favor of eventually dropping the implementation of traitlets.config.Application and rolling our own simpler one (still using traitlets, since I do like them).  E.g. just defining that any traitlet in a class is a configurable parameter.

ctapipe/calib/camera/r1.py
kosack: What I mean by that is a new developer should not have to paste that loop into each class that needs it, they should just set a metaclass to "Producible" or something like that, and the metaclass automatically sets __all__

ctapipe/calib/camera/calibrator.py
dneise: I think these aliases cannot be restored with this current approach, since at Tool creation time it is not known which of the ChargeExtractors will be selected by the user.

This is a major difference between the current version in the master and this proposal here: A user can only configure things by naming them explicitly. I believe this is an advantage, since errors are more visible in the command line call.

As an example, this is a valid (but probably wrong) invocation using the current master:

    ctapipe-event-viewer \
    --ChargeExtractorFactory.t0=10  \
    --ChargeExtractorFactory.sig_amp_cut_LG=5.3

No existing ChargeExtractor has both `t0` and `sig_amp_cut_LG`.

Contrasting this branch here, forces the user to write this:
    ctapipe-event-viewer \
    --SimpleIntegrator.t0=10  \
    --GlobalPeakIntegrator.sig_amp_cut_LG=5.3

Which gives a user another moment to pause and say ... hey, I might have made a mistake here.

dneise:
> I think the issue of plugins can perhaps be separated from this.

I agree. It came up in a question, so it was discussed, but it is a separate issue.

ctapipe/calib/camera/calibrator.py
watsonjj: From your comment @dneise, it does sound like it can be restored, it just has to be more verbose than the current implementation:

```python
        ped='TargetIOR1Calibrator.pedestal_path',	
        tf='TargetIOR1Calibrator.tf_path',	
        pe='TargetIOR1Calibrator.adc2pe_path',	
        simpleintegrator_t0='SimpleIntegrator.t0',	
        window_width='WindowIntegrator.window_width',	
        window_shift='WindowIntegrator.window_shift',	
        sig_amp_cut_HG='PeakFindingIntegrator.sig_amp_cut_HG',	
        sig_amp_cut_LG='PeakFindingIntegrator.sig_amp_cut_LG',	
        lwt='NeighbourPeakIntegrator.lwt',
```
then one can do `ctapipe-event-viewer --window_width 20`
Could you see if that works?

watsonjj: > E.g. just defining that any traitlet in a class is a configurable parameter.

I believe this is possible, as in @dneise comment:
```
ctapipe-event-viewer \
--ChargeExtractorFactory.t0=10  \
--ChargeExtractorFactory.sig_amp_cut_LG=5.3
```

Its only when you wish to set an alias does one need to write out the extra stuff - which I think is important as it avoids confusion between identical traitlets. Does one also need to have the class included in the `classes` List to be able to do the `--SimpleIntegrator.t0=10`?

watsonjj: > I think we should write down what we want here: I would say, having the ability to set configuration parameters from the base class and any extra ones in the sub-class is a nice thing

This is what I think could be convenient, is this you meant too @kosack?

The following should be valid to set a Charge Extractor window width:

```python
ctapipe-event-viewer --WindowIntegrator.window_width=10 #Works in current version
ctapipe-event-viewer --PeakFindingIntegrator.window_width=10 #Does not yet work (I think)
ctapipe-event-viewer --LocalPeakIntegrator.window_width=10 #Does not yet work (I think)
```

As all these classes are subclasses of `WindowIntegrator`, where the `window_width` traitlet is defined.

However I don't think the following should be possible as it might cause problems, as different subclasses of this class may have their own version of the traitlet which operates differently. I think this behaviour partially exists with the current Factory implementation, which @dneise is trying to avoid:

```python
ctapipe-event-viewer --ChargeExtractor.window_width=10
```

ctapipe/calib/camera/r1.py
dneise: I am sorry there seems to be a misunderstanding here, the in the master, this part looks like this:
https://github.com/cta-observatory/ctapipe/blob/c5f609b8bf31d5fe8d17577b020949d40ffe44b2/ctapipe/calib/camera/r1.py#L22-L25

The advantage of having this loop is only, that a developer of a new CameraR1Calibrator does not need to remember to put the name of the new class into `__all__`, but we can of course revert back to what is in the master.

> a new developer should not have to paste that loop into each class that needs it

Absolutely. And they do not ... they can do whatever they want. 

I see putting this loop into `__all__` is confusing for reviewers and it is unrelated to the actual PR. I will remove it.

ctapipe/calib/camera/calibrator.py
watsonjj: So interestingly, even this works to change the `window_width` in both this PR and the master:

```
ctapipe-event-viewer --Component.window_width=20
```
You can tell by scrolling down on the viewer to see the value its been set to. It does result in a bunch of warnings for all the Components that don't have a window_width traitlet. But it seems that as long as the class is a subclass of what you specify on the command line, it will try to set the traitlet value.

dneise:
> Can you give a short top-level description in this PR 

I intended the issue #915 which caused this PR to explain my intentions. Maybe the issue text was a little bit too long. So I updated the PR message.

ctapipe/calib/camera/calibrator.py
dneise: >  ... it does sound like it can be restored ...

Ah .. I didn't get that. In that case I will happily do that ... soon.

ctapipe/core/component.py
```diff
@@ -80,3 +94,24 @@ def __init__(self, parent=None, config=None, **kwargs):
             self.log = getLogger(
                 self.__class__.__module__ + '.' + self.__class__.__name__
             )
+
+    @classmethod
```
watsonjj: @MaxNoe Here the code of Component is modified (relevant to conversation in #917)

MaxNoe: But this is just a utility method to get a subclass by name, right?

This is not "Implementing a Factory-like" behaviour.

ctapipe/io/eventsource.py
dneise: done

ctapipe/core/component.py
dneise: this is outdated, I'll close it here.

ctapipe/calib/camera/calibrator.py
dneise: I restored these aliases in the last commits (travis is running the tests currently) so I assume this conversation could be closed.

ctapipe/calib/camera/r1.py
dneise: This is fixed now

ctapipe/core/component.py
```diff
@@ -1,10 +1,45 @@
...
+    subclasses = base.__subclasses__() + [
+        g for s in base.__subclasses__()
+        for g in non_abstract_children(s)
+    ]
+    non_abstract = [g for g in subclasses if not isabstract(g)]
+
+    return non_abstract
+
+
+def subclass_from_name(base, name=None, *args, **kwargs):
```
MaxNoe: add `@lru_cache`?

dneise: I would refrain from optimizations at this point. I think, the usual use case for this function is once in the lifetime of a tool at the beginning, so not much to cache here. But we can certainly do this when we have the impression this is a time-sink.

watsonjj: Also if you cache this, then it might not detect new children that have been added to the global space since the previous call

dneise: So we resolve this comment

dneise:
@watsonjj I believe I have implemented your requested changes .. would you like to re-review?

ctapipe/tools/extract_charge_resolution.py
```diff
@@ -64,7 +69,7 @@ def setup(self):
 
         self.eventsource = SimTelEventSource(**kwargs)
 
-        extractor = ChargeExtractorFactory.produce(**kwargs)
+        extractor = ChargeExtractor.from_name(self.extractor_product, **kwargs)
```
watsonjj: This won't work anymore, right?

dneise: So we are missing a test for this tool. Well spottet! Thanks .. I will try to add one .. 
Or does the missing of the test imply that this tool is not really needed and should be removed?

watsonjj: I believe I improve the testing for this tool in #827

dneise:
Thank you Jason. I am looking forward to #917. 

I am sure we can agree on something.

I personally am happy, that I learned in this study, that meta-programming is not urgently needed to combine our use-cases and `traitlets.config`. Due to the time I spent on this study here I believe I do now understand our existing Factory in the master, as well as what I did here and I am positive I will also understand #917 
So if we cannot agree I am happy to close this PR here.

watsonjj:
> - speed ups in some EventSources is_compatible methods from seconds to ms.

@dneise Could you contribute these in another PR? They are very useful

dneise:
ah sure ...

dneise:
> 1. multiple standards?

"This is a feature not a bug" is a lame meme, but here it is true.
As I wrote above. I believe our current Factory implementation violates SRP in 
some sense. This is visible e.g. by the generality of the name `produce()`
and especially visible when it is used as `SomeFactory.produce(**kwargs)`.
So in order to restore SRP in this context, I decided to replace one general method
by multiple specific ones.


> 3. Using these factories in a Tool is now more complicated.

I think it is more explicit and uses more lines. I do not think it is more 
complicated.


> 4. not a standard for creating a new custom factory

There is no standard, since no standard is needed. All one needs is a function
like this:

```python
def make_thing_from_color(color, *args, **kwargs):
    if color == 'blue':
        return Whatever(*args, **kwargs)
    elif color == 'red': 
        return Dunno(*args, **kwargs)
    else:
        raise ValueError('color must be either blue or red')
```

Or similar. **If** this factory is needed in order to write a tool, then the 
tool auther of course needs to make sure `Whatever` and `Dunno` (which do not have 
to have a common base class! they just share similar behavior) are listed in 
`classes` of that tool, but **only** if the author wants the user to be able to
configure these things. It is totally valid for the tool author to prevent 
the user from configuring these Components in detail, but beeing able 
to specify `color`. 

> 5. we have less power to direct and influence a standardised factory behaviour.

Indeed. We have no power anyway. ctapipe is a library. People can use it 
to build their own projects. We have no power at all.
Illusions of power have in my opinion no place in open source software development.

> 6. There is some attraction to the function subclass_from_name [...], but we might want to limit which Components can have Factory access.

C'mon that function has 5 lines ... if the use case you describe arises, **then**
a function can be written, for that usecase. This is not a valid argument 
against that function.

> 7. The handling of which Component is returned by default is a little unintuitive.

Yes. Introducing the class variable `_default_name` was not a good idea. So it was removed.

> 8. selecting an EventSource based on the input_url is more clunky

I think not:  `event_source(input_url=...)` .. done.

watsonjj:
I have just committed a test where some arguments are passed via the factory.

The usecase I envision for this is the following. I create a Tool which extracts the charge from the waveform. I want to be able to switch between ChargeExtractors freely via the cmdline. This functionality is given to me via the factory. However, if my charge extraction method that I choose has a configurable window width and shift, I want it to be always 20, 3.

As you can see, if I choose the FullIntegrator ChargeExtractor, an error occurs, as the FullIntegrator does not have the window_width and window_shift traitlets. However, I want to be able to use any ChargeExtractor I want.

This was possible in the previous Factory approach. And in #917 the `Factory._clean_kwargs_for_product` method keeps this functionality. Do we still want this to be possible?

Perhaps its not a job of the factory, but it is a job of the configuration system to properly handle this usecase. The configuration system can correctly handle a trait being set, even if the ChargeExtractor does not have that trait (in which case it prints a warning). But the configuration system does not allow you to pass an argument if the Comonent does not have a corresponding trait - it raises the error you see in this test.

If it is an issue of the configuration system, perhaps we should keep this functionality in the factory for now, and address improvements to the configuration separately, after which the factory no longer needs this.

dneise:
Thanks Jason. This is a very good question which in my opinion shows how difficult the traitlets config package is to use. Thanks also for not only adding this test but also explaining where the `kwargs` should come from, namely from the traitlets config machinery.

So ... I have to say, the test is wrong :-S ... since the `self.config` object we get will not look like this:
```
    kwargs = dict(
        window_width=20,
        window_shift=3,
    )
```

since `self.config` neither is a `dict` but a `traitlets.config.Config` (or so) nor will it be so shallow. It will look like this:
```python
    config = Config(
        {
            'ChargeExtractor': {
                'window_width': 20,
                'window_shift': 3,
            }
        }
    )
```

So it will have this additional `ChargeExtractor` level. 

And if we recreate this `self.config` object, which we would get in a real `Tool` also in this new test like this:
```python
def test_charge_extractor_factory_args():

    config = Config(
        {
            'ChargeExtractor': {
                'window_width': 20,
                'window_shift': 3,
            }
        }
    )

    local_peak_integrator = subclass_from_name(
        ChargeExtractor,
        'LocalPeakIntegrator',
        config=config,
    )
    assert local_peak_integrator.window_width == 20
    assert local_peak_integrator.window_shift == 3

    subclass_from_name(
        ChargeExtractor,
        'FullIntegrator',
        config=config,
    )
```

then, I believe, everything works as expected.

dneise:
Ah .. I misunderstood the use-case. So the subject of this usecase is a Tool-Developer not a Tool-User.

So yes, I believe a Tool Developer would have to modify the self.config object if they wanted to modify the configuration somehow. The would indeed not be able to say:
```
    subclass_from_name(
        ChargeExtractor,
        'FullIntegrator',
        window_width=20,
    )
```

But I do not see this as a problem or so ... albeit inconvenient.

mackaiver:
I like the approach taken here. To me this is much easier to read and understand. I like methods like `subclass_from_name` cause they are very explciit in what they do compared to the previous approach.

ctapipe/io/eventsource.py
```diff
@@ -198,3 +226,10 @@ def __enter__(self):
 
     def __exit__(self, exc_type, exc_val, exc_tb):
         pass
+
+
+def from_config(config, *args, **kwargs):
```
watsonjj: Should restore tool argument, explicit arguments are better to have

ctapipe/io/eventsource.py
```diff
@@ -198,3 +226,10 @@ def __enter__(self):
...
+    '''return EventSource instance from traitlets.Configuration'''
+    if not isinstance(config.EventSource.input_url, str):
+        config.EventSource.input_url = EventSource.input_url.default_value
+    return event_source(config.EventSource.input_url, *args, **kwargs)
```
watsonjj: config needs to be passed to event_source for tests to pass

ctapipe/io/eventsource.py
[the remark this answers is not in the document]
dneise: Noooo :-D Single Responsibility Principle.
No seriously. Sure it could .. but I don't think it is a good idea to do so. `event_source` has a pretty clear interface. It takes a `url` an returns an instance of a compatible based only on that `url`. (We ignore that one can provide more parameters which are just forwarded into the particular EventSources __init__, since it is transparent) 

`from_config` is a complex thing dealing with traitlets.config.LazyConfig objects. It might break for all sorts of reasons and it might need to be modified for all sorts traitlets connected reasons. Single Responsibility means not only "one purpose" but also "one reason to change" ...

dneise: `tool` was not used in the body and was only forwarded to the client here. If we ever decide to get rid of `tool` or to rename `tool` this method does not need to change.

ctapipe/io/eventsource.py
```diff
@@ -3,11 +3,43 @@
...
+from traitlets import Unicode, Int, Set, TraitError
+from ctapipe.core import Component, non_abstract_children
 from ctapipe.core import Provenance
+from traitlets.config.loader import LazyConfigValue
 
-__all__ = ['EventSource', ]
+__all__ = ['EventSource', 'event_source']
+
+
+def event_source(input_url, *args, **kwargs):
```
watsonjj: I'm not sure what is gained by allowing positional arguments from event_source and from_config to be passed to EventSource. Perhaps only allow named arguments (kwargs) for both methods?

ctapipe/io/eventsource.py
```diff
@@ -198,3 +227,12 @@ def __enter__(self):
 
     def __exit__(self, exc_type, exc_val, exc_tb):
         pass
+
+
+def from_config(config, *args, **kwargs):
```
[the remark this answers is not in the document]
dneise: You are right. Clear names are better than short and unclear ones. I personally would do `eventsource.from_this` and `eventsource.from_that`.

But in this case many people are already used to the convenience `event_souce()` function. It is being used more than 100 times in our code base in more than 30 files (not including pure documentation files like rst or html or so).

While `eventsource.from_config` is hardly used at all. It is only used correctly in the file_viewer.py In the other 3 examples it is used incorrectly (both in the master and in this branch here)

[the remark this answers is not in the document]
watsonjj: I can understand the argument for not changing `event_source`.

I would still like to see `from_config` changed to `event_source_from_config`, and the tools and examples updated to import that method directly.

ctapipe/io/eventsource.py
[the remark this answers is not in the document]
watsonjj: Okay, this makes sense

ctapipe/io/eventsource.py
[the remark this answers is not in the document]
dneise: Okay gonna do it ... soonish

dneise:
> Should there be a deprecation cycle for the existing factory, with instructions on how to update a script to use these changes?

I think **if** the ctapipe core devs think this is a good API, we should find a away to best communicate the API to potential users. But honestly so far, I did not think about whether or not a deprication cycle was needed or not. 

For many users I assume nothing will change, since `event_source()` did not change. For all Tool developers however a lot will change. So maybe yes. What do others think?

dneise:
Honestly I have no idea how much ctapipe using code is out there. I know there is a whole pipeline somewhere developed in secrecy.... which is of course fine, if ctapipe was as stable as say numpy... but for now, I believe core devs could potentially learn a lot, if they had access to a decent part of the scripts and applications using ctapipe out in the wild.

ctapipe/calib/camera/r1.py
```diff
@@ -9,24 +9,51 @@
...
     'NullR1Calibrator',
     'HESSIOR1Calibrator',
     'TargetIOR1Calibrator',
-    'CameraR1CalibratorFactory'
+    'CameraR1Calibrator',
+    'camera_r1_calibrator_for_eventsource',
 ]
 
 
+def camera_r1_calibrator_for_eventsource(eventsource, **kwargs):
```
kosack: Shouldn't this be a constructor for CameraR1Calibrator, following the convention that factories are really just specialized constructors of the base class?  e.g. `cal = CameraR1Calibrator.for_event_source(source)`

dneise:
> General question: would it be cleaner and easier to understand if we included a helper constructor for each class like CameraR1Calibrator.from_name() rather than calling subclass_from_name directly? It seems to me that having constructor methods is the way most python code does it (notably astropy)

Yes. This was exactly what we had about a week ago in this PR. It was subsequently rejected, since I was unable to solve this:

If `CameraR1Calibrator` has a method named `from_name`, then also all its children will inherit that method. Consequently one would be able to write non-sensical code like `TargetIOR1Calibrator.from_name('HESSIOR1Calibrator')`.

If I remember correctly @MaxNoe did not think this was a major issue, @watsonjj thought this was a major issue and I was later convinced that Jason was right in this point.

ctapipe/calib/camera/r1.py
dneise: See my comment below.

dneise:
I think `event_source(url)` should stay .. it is useful.

watsonjj:
Okay, but how about it calls `EventSource.from_url` inside?

kosack:
> If we do switch, would we change event_source into EventSource.from_url and event_source_from_config to EventSource.from_config? I think this would be nice, but in this case I would suggest the event_source function goes through a deprecation cycle (with a warning and instructions) due to its wide usage.

I understand the arguments now - though I guess in the end I'm fine either way. It is a bit confusing to have the constructor propagated to the sub-classes, but it is also more clear that you are constructing an EventSource  if you use `EventSource.from_url(pathname, ...)`. In that case yes we would have to deprecate event_source since many people use it.  I suppose something like  `TargetIOEventSource.from_url(filename)` would still work, it would just give semantically confusing results if given a simtel file...

kosack:
> Okay, but how about it calls EventSource.from_url inside?

yes, that's basically what it did anyhow. Originally the whole EventSource code was in that function an it was a generator, then it changed to just wrap constructing an EventSourceFactory.  It could stay around.

kosack:
By the way, do URLs still work? Could be better to just call it `from_file` and accept anything resolvable by `pathlib.Path().expanduser().resolve()` or `urllib.request.urlopen()`

kosack:
that's for another PR though

kosack:
We also then need  clear documentation of how to create other things that were previously Factory classes.  In the end there may be many hierarchies we want to construct from name or other attributes (`GainSelector`, `Reconstructor`, `FlatFieldingCorrector`, etc.)

watsonjj:
> We also then need clear documentation of how to create other things that were previously Factory classes. In the end there may be many hierarchies we want to construct from name or other attributes (GainSelector, Reconstructor, FlatFieldingCorrector, etc.)

In my understanding, all subclasses of Component will have the `from_name` method. But I agree we could have a documentation section on how to add special factory methods to a class, similar to `EventSource.from_url`.

watsonjj:
With @dneise's permission, I have implemented the switch to using classmethods as discussed.

The `event_source` function seems quite unnecessary now, but I have left it in.

I have noticed that there are discrepancies in how the `config` and `tool`/`parent` arguments are structured the initialisations of different `Components`. This has caused an issue with the new factory approach. I will create a separate PR that attempts to fix this disparity, as it is clearly its own issue that needs to be addressed irregardless of this PR.

PR #917, opened 2019-01-09
Factory V3

watsonjj:
As an alternative suggestion to #916, I would like to propose some improvements to the existing `Factory` that:

- [x] Keeps the distinction between a Factory and a Component, therefore avoiding every Component containing Factory functionality due to inheritance
- [ ] Addresses the issues highlighted in #915 
- [ ] Avoids confusion in the setting of traitlets
- [ ] Simplifies the code in the `Factory` definition
- [ ] Improves on the Component discovery, removing the Factory's dependance on the Components being defined before it is (potentially preparing it for a plugin system)

This is a work-in-progress and a test of concept. I will see how successfully I can achieve these goals. Other goals may also be identified and added to this list as I progress.

### DIFF
--- a/ctapipe/core/tests/test_factory.py
+++ b/ctapipe/core/tests/test_factory.py
@@ -193,3 +193,27 @@ def test_default_passing():
     ExampleFactory.value.default_value = old_default
     obj = ExampleFactory.produce()
     assert (obj.value == old_default)
+
+
+def test_component_definition_after_factory():
+    """
+    Test if a component defined after the factory definition will be
+    obtainable by the factory
+    """
+    class ExampleComponent5(ExampleComponentParent):
+        value = Int(1234445, help="").tag(config=True)
+
+    assert "ExampleComponent5" in ExampleFactory.product.values
+    obj = ExampleFactory.produce(product='ExampleComponent5')
+    assert (obj.value == 1234445)
+
+def test_component_definition_after_factory_help_message():
+    """
+    Test if a component defined after the factory definition will be
+    inside the help message
+    """
+
+    class ExampleComponent5(ExampleComponentParent):
+        value = Int(1234445, help="").tag(config=True)
+
+    assert "ExampleComponent5" in ExampleFactory.class_get_help()


watsonjj:
> The other proposal does not do that. It does not include factory behaviour for all Components, just for specific subclasses of components that need it..

That is not what I understood from the code. Could you help clarify @dneise?

From what I understand, the Factory functionality has been added to every Component via the `from_name` classmethod. It now means that any sub-Component can be returned from a base-Component.

It still stands that in #916 there is no longer a distinct Factory class, with a clear singular purpose to return a Component. Whereas the purpose of the Component is kept solely to perform operations on some data, not return other Components.

MaxNoe:
> That is not what I understood from the code. Could you help clarify @dneise?

Check again, the code of `Component` is not modified. Only the code of `EventSource` and `CameraCalibrator`

MaxNoe:
Or more precise, there is only a utility method to get a subclass of a certain component by name.
But the Factory-like behaviour is not present in component.

watsonjj:
> Or more precise, there is only a utility method to get a subclass of a certain component by name.
But the Factory-like behaviour is not present in component.

I would argue that is the factory behaviour...  a class of a different type is being returned depending on the input given...

watsonjj:
In the latest commit, I have restored the `produce` classmethod, which is labelled as deprecated. This method restores the old way of using the previous Factory, along with the passing of traits from the Factory to the Component. This means that any user's scripts will still work - there is no break of API. The user is warned that this way of using Factory is deprecated, and they should switch to the new method.

The improvements to Factory (such as the removal of the requirement that Components are imported before the Factory) are still present for all calls to the Factory.

dneise:
I still think the implementation in this PR is complicated. Let me show you what I mean:
https://youtu.be/olqOG-sbZLY

watsonjj:
Brilliant! I loved this. You clearly highlighted problems, and I agree with them. Sometimes its easy to get too detached from how a newcomer would react to code one has grown too familiar with. I will try to save Jerry from being bothered too much.

dneise:
In addition to what I said above. I still think the Factory implementation as it is in the master and also in this PR is adding complexity in terms of lines of code, as this PR added 608 lines of code and removed 366, which is a net gain of ~300 lines. 

Whereas #916 added 657 lines and removed  1140 lines, which is a net loss of ~500 lines. 

Now, this argument is not clean, since lines of code are not at all a measure for complexity. Quite often more lines can make stuff more explicit and thus *easier* to understand. Maybe another measure for complexity could be: How difficult is it to understand the implementation of `event_source()`

I'm making another video about this... stay tuned

thomasgas:
> I still think the implementation in this PR is complicated. Let me show you what I mean:
> https://youtu.be/olqOG-sbZLY

Best explanation ever!!!

watsonjj:
@dneise There were some errors in the documentation I had committed, which I noticed thanks to your video. These should be fixed. The docstrings should also be a lot more helpful now.

With regards to your comment about lines of code, I believe this PR does reduce the amount of primary **code** compared to its master, and hopefully its complexity. But there is a lot more documentation, some more tests, and some new functions to improve clarity about how the factory functions. There is also the retaining of the deprecated `produce` method, which adds around 100 lines, but are intended to be temporary and removed in a future version.

watsonjj:
@dneise Thank-you for the second video.

I have committed some changes that make the Factory and EventSourceFactory have a more intuitive generation process. Specifically, `_get_product_name` is now called separately, instead of inside `_get_constructor`. Instead, `_get_constructor` (which is renamed to `_get_product_constructor`) takes the name as an argument.

These changes mean that in the case that an input_url did not result in the expected EventSource, a new developer should not need to delve into the Factory class to figure out how to debug the issue. Instead the sequence of events that lead up to the error are more easy to follow.

I have also removed the functionality to pass kwargs to the product, as you have done in your PR (after our discussion regarding this). This decision allowed me to remove some of the more complex code of the Factory (`_clean_kwargs_for_product`).

watsonjj:
Closed in favour of #916

PR #918, opened 2019-01-09
Add get_pixel_id method

moritzhuetten:
Added CameraGeometry.get_pixel_id() method to return the pixel number(s) corresponding to a (or an array of) x,y position(s), to be used like:
```
>>> geom = CameraGeometry.from_name("LSTCam")
>>> x, y = 0.1 * u.m, 0.1 * u.m
>>> geom.get_pixel_id(x, y)
138
>>> x, y = [0.1, 0.3] * u.m, [0.1, 0.4] * u.m
>>> geom.get_pixel_id(x, y)
array([138, 876])
```

### DIFF
--- a/ctapipe/instrument/camera.py
+++ b/ctapipe/instrument/camera.py
@@ -204,7 +204,7 @@ class CameraGeometry:
         return np.ones(pix_x.shape) * area
 
     @lazyproperty
-    def all_pixels_kdtree(self):
+    def kdtree(self):
         """
         Pre-calculated kdtree of all pixel centers inside camera
 
@@ -556,7 +556,7 @@ class CameraGeometry:
 
         points_searched = np.dstack([x.to_value(u.m), y.to_value(u.m)])
 
-        kdtree = self.all_pixels_kdtree
+        kdtree = self.kdtree
         dist, pix_indices = kdtree.query(points_searched)
         pix_indices = pix_indices.flatten()
 
--- a/ctapipe/instrument/camera.py
+++ b/ctapipe/instrument/camera.py
@@ -533,9 +533,9 @@ class CameraGeometry:
 
         return longi, trans
 
-    def get_pixel_id(self, x, y):
+    def position_to_pix_index(self, x, y):
         '''
-        Return the camera pixel number which contains a given position (x,y) 
+        Return the index of a camera pixel which contains a given position (x,y) 
         in the camera frame. The (x,y) coordinates can be arrays (of equal length),
         for which the methods returns an array of pixel ids. A warning is raised if the 
         position falls outside the camera.
@@ -547,8 +547,8 @@ class CameraGeometry:
 
         Returns
         -------
-        pixel_ids: Pixel number or array of pixel numbers. Returns -1 if position falls 
-                   outside camera
+        pix_indices: Pixel index or array of pixel indices. Returns -1 if position falls 
+                    outside camera
         '''
 
         if np.any(~np.isclose(self.pix_area.value, self.pix_area[0].value), axis=0):
@@ -557,8 +557,8 @@ class CameraGeometry:
         points_searched = np.dstack([x.to_value(u.m), y.to_value(u.m)])
 
         kdtree = self.all_pixels_kdtree
-        dist, pixel_ids = kdtree.query(points_searched)
-        pixel_ids = pixel_ids.flatten()
+        dist, pix_indices = kdtree.query(points_searched)
+        pix_indices = pix_indices.flatten()
 
         # Check if the position lies inside the camera. It is first checked if any border
         # pixel numbers are returned. If not, everything is fine. If yes, the distance of 
@@ -570,30 +570,30 @@ class CameraGeometry:
 
         border_mask = self.get_border_pixel_mask()
         # get all pixels at camera border:
-        borderpix_ids = np.where(border_mask)[0]
+        borderpix_indices = np.where(border_mask)[0]
 
-        borderpix_ids_in_list = np.intersect1d(borderpix_ids, pixel_ids)
-        if borderpix_ids_in_list.any():
+        borderpix_indices_in_list = np.intersect1d(borderpix_indices, pix_indices)
+        if borderpix_indices_in_list.any():
             # Get some pixel not at the border:
-            inside_pix_id = np.where(~border_mask)[0][0]
+            insidepix_index = np.where(~border_mask)[0][0]
             # Check in detail whether location is in border pixel or outside camera:
-            for borderpix_id in borderpix_ids_in_list:
-                index = np.where(pixel_ids == borderpix_id)[0][0]
+            for borderpix_index in borderpix_indices_in_list:
+                index = np.where(pix_indices == borderpix_index)[0][0]
                 # compare with inside pixel:
                 xprime = points_searched[0][index, 0] \
-                    - self.pix_x.to_value(u.m)[borderpix_id] \
-                    + self.pix_x.to_value(u.m)[inside_pix_id]
+                    - self.pix_x.to_value(u.m)[borderpix_index] \
+                    + self.pix_x.to_value(u.m)[insidepix_index]
                 yprime = points_searched[0][index, 1] \
-                    - self.pix_y.to_value(u.m)[borderpix_id] \
-                    + self.pix_y.to_value(u.m)[inside_pix_id]
+                    - self.pix_y.to_value(u.m)[borderpix_index] \
+                    + self.pix_y.to_value(u.m)[insidepix_index]
                 dist_check, index_check = kdtree.query([xprime, yprime])
-                if index_check != inside_pix_id:
+                if index_check != insidepix_index:
                     logger.warning(" Coordinate ({} m, {} m) lies outside camera"
                                    .format(points_searched[0][index, 0], 
                                            points_searched[0][index, 1]))
-                    pixel_ids[index] = -1
+                    pix_indices[index] = -1
 
-        return pixel_ids if len(pixel_ids) > 1 else pixel_ids[0]
+        return pix_indices if len(pix_indices) > 1 else pix_indices[0]
 
 
 # ======================================================================
--- a/ctapipe/instrument/camera.py
+++ b/ctapipe/instrument/camera.py
@@ -204,6 +204,21 @@ class CameraGeometry:
         return np.ones(pix_x.shape) * area
 
     @lazyproperty
+    def _pixel_circumferences(self):
+        """ pixel circumference radius/radii based on pixel area and layout
+
+        """
+
+        if self.pix_type.startswith('hex'):
+            circum_rad = np.sqrt(2.0 * self.pix_area / 3.0 / np.sqrt(3))
+        elif self.pix_type.startswith('rect'):
+            circum_rad = np.sqrt(self.pix_area / 2.0)
+        else:
+            raise KeyError("unsupported pixel type")
+
+        return circum_rad
+
+    @lazyproperty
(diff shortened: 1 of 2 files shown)


MaxNoe:
Awesome.


How does this behave for points outside the camera?

ctapipe/instrument/camera.py
```diff
@@ -517,6 +517,54 @@ def get_shower_coordinates(self, x, y, psi):
...
+
+        Returns
+        -------
+        pixel_ids: Pixel number or array of pixel numbers. Returns -1 if position falls outside camera
+        '''
+        
+        pixel_centers = np.stack([self.pix_x.to_value(u.m),self.pix_y.to_value(u.m)]).T
+        points_searched = np.dstack([x.to_value(u.m),y.to_value(u.m)])
+
+        kdtree = KDTree(pixel_centers)
```
MaxNoe: I would suggest to build the kdtree in `__init__`, as it is used by more than one method.

ctapipe/instrument/camera.py
```diff
@@ -517,6 +517,54 @@ def get_shower_coordinates(self, x, y, psi):
...
+        pixel_ids = pixel_ids.flatten()
+        
+        # Check if the position lies inside the camera. It is first checked if any border
+        # pixel numbers are returned. If not, everything is fine. If yes, the distance of 
+        # the given position to the closest pixel center is translated to the pixel center of the,
+        # camera center, pos -> pos', and it is checked whether pos' lies within the camera center 
+        # pixel. If not, pos lies outside the camera. This approach does not need to know the  
+        # particular pixel shape.
+        
+        borderpix_ids = np.where(self.get_border_pixel_mask())[0]
```
MaxNoe: I think this can be vectorized by just using the mask

moritzhuetten:
> How does this behave for points outside the camera?

Returns -1 and prints a warning.

ctapipe/instrument/camera.py
kosack: agreed - right now we always make the assumption that you cannot modify the geometry (many functions cache a result).  It could also go as a `lazyproperty` like the neighbor matrix,  since most of the time you don't need it.

moritzhuetten:
At the moment, the Outside-of-camera-check is exact (for arbitrary pixel shapes), but presumes that camera_pixels[0] lies in the center of the camera (which may not be always the case). A much simpler, but only approximate alternative would be to check if the position distance to the nearest pixel center is smaller than than the radius of the pixels' circumcircle.

ctapipe/instrument/camera.py
moritzhuetten: put it into lazyproperty in 7114978

moritzhuetten:
Dropped assumption on central pixel in e95ff10, still being exact. However, the code might be much simpler and & readable if the inside-camera-check is only approximate. Or doing an exact check in a different, smarter way.

ctapipe/instrument/camera.py
```diff
@@ -203,6 +203,21 @@ def _calc_pixel_area(pix_x, pix_y, pix_type):
...
+        Pre-calculated kdtree of all pixel centers inside camera
+
+
+        Returns
+        -------
+        kdtree
+
+        """
+
+        pixel_centers = np.stack([self.pix_x.to_value(u.m), self.pix_y.to_value(u.m)]).T
```
MaxNoe: `np.column_stack` gives directly what you want here without having to transpose.

ctapipe/instrument/camera.py
```diff
@@ -518,6 +533,69 @@ def get_shower_coordinates(self, x, y, psi):
...
+        x: astropy.units.Quantity (distance) of horizontal position(s) in the camera frame
+        y: astropy.units.Quantity (distance) of vertical position(s) in the camera frame
+
+        Returns
+        -------
+        pixel_ids: Pixel number or array of pixel numbers. Returns -1 if position falls 
+                   outside camera
+        '''
+
+        if not np.all(self.pix_area == self.pix_area[0], axis=0):
```
MaxNoe: Use `if np.any(self.pix_area != self.pix_area[0])`. `all` always needs to test every value, `any` exits at the first `True` value it encounters.

Also this looks like it should be a `@lazyproperty`, like `all_pixel_areas_equal` or  better name you can come up with.

ctapipe/instrument/camera.py
```diff
@@ -518,6 +533,69 @@ def get_shower_coordinates(self, x, y, psi):
...
+
+        borderpix_ids_in_list = np.intersect1d(borderpix_ids, pixel_ids)
+        if borderpix_ids_in_list.any():
+            # Get some pixel not at the border:
+            inside_pix_id = np.where(~border_mask)[0][0]
+            # Check in detail whether location is in border pixel or outside camera:
+            for borderpix_id in borderpix_ids_in_list:
+                index = np.where(pixel_ids == borderpix_id)[0][0]
+                # compare with inside pixel:
+                xprime = points_searched[0][index, 0] \
```
MaxNoe: for calculations, use implicit line continuations in parenthese instead of explicit line continuations, like this:

```
a = (
    b 
    + c
)
```

MaxNoe: Also, converting units is kind of expensive, so you should do `self.pix_y[index].to_value(u.m)` instead of converting all values and then only using one.

ctapipe/instrument/camera.py
```diff
@@ -518,6 +533,69 @@ def get_shower_coordinates(self, x, y, psi):
 
         return longi, trans
 
+    def get_pixel_id(self, x, y):
```
kosack: I'd suggest renaming this - it's not clear what the input is ("get" could mean just lookup the pixel_id from the pixel *index*). Maybe `position_to_pixel_id(x,y)`?

ctapipe/instrument/camera.py
```diff
@@ -518,6 +533,69 @@ def get_shower_coordinates(self, x, y, psi):
...
+        position falls outside the camera.
+
+        Parameters
+        ----------
+        x: astropy.units.Quantity (distance) of horizontal position(s) in the camera frame
+        y: astropy.units.Quantity (distance) of vertical position(s) in the camera frame
+
+        Returns
+        -------
+        pixel_ids: Pixel number or array of pixel numbers. Returns -1 if position falls
```
kosack: Does this give the pixel ID (the name of the pixel), or the pixel *index* the index into the arrays in the `CameraGeometry` (these are not the same thing)

given a pixel *index* **i**, the x-position of that pixel is`geom.pix_x[i]`, and the pixel_id of that pixel is `geom.pix_id[i]`. 

You should avoid the word "pixel number" and don't confuse the concepts.

ctapipe/instrument/camera.py
kosack: also might want to use a tolerance, in case of any rounding error in case the area of the pixel is calculated somewhere. (or `np.isclose()` - is there an `anyclose()` function?),

ctapipe/instrument/camera.py
```diff
@@ -203,6 +203,21 @@ def _calc_pixel_area(pix_x, pix_y, pix_type):
 
         return np.ones(pix_x.shape) * area
 
+    @lazyproperty
+    def all_pixels_kdtree(self):
```
kosack: is there a case where this is not for all pixels? Otherwise, perhaps just name this property `kdtree`

ctapipe/instrument/camera.py
kosack: Looking at the code, it seem to compute the `pixel_index`, not the `pixel_id`

ctapipe/instrument/camera.py
MaxNoe: `np.any(~np.isclose(...`

kosack:
> Dropped assumption on central pixel in e95ff10, still being exact. However, the code might be much simpler and & readable if the inside-camera-check is only approximate. Or doing an exact check in a different, smarter way.

Couldn't that just be a bounding box? E.g.  `(pix_x.min()-pix_rad, pix_x.max()+pix_rad)`, that's an easy and rough way to avoid computing if it's in the camera. You could compute a bounding circle as well, which might be better since most cameras are pretty circular, so `r < hypot(pix_x, pix_y).max()` or similar.

ctapipe/instrument/camera.py
moritzhuetten: Renamed it to simply kdtree in f1a7b92d. If finally satisfied with this PR, maybe have a look at already existing _find_neighbor_pixels method() to work with the same kdtree (seems to also use all pixels, but slightly differently implemented).

ctapipe/instrument/camera.py
moritzhuetten: done in e43e781.

ctapipe/instrument/camera.py
moritzhuetten: done in 704d7cc and 5c9c185. I marked the lazyproperty internal (leading underscore). Correct me if that's a wrong use-case.

ctapipe/instrument/camera.py
moritzhuetten: renamed method to position_to_pix_index() in 49e3572.

ctapipe/instrument/camera.py
moritzhuetten: Indeed, it's the pixel index. Wasn't aware of the difference in ctapipe, but of course makes sense to differentiate, thanks for pointing this out. Renamed the method to position_to_pix_index(), because thought it's useful to let the user decide if s/he wants to obtain the id in a separate step. Also clarified the variable naming to avoid confusing expression 'number'.

ctapipe/instrument/camera.py
```diff
@@ -518,6 +533,69 @@ def get_shower_coordinates(self, x, y, psi):
...
+                   outside camera
+        '''
+
+        if not np.all(self.pix_area == self.pix_area[0], axis=0):
+            logger.warning(" Method not implemented for cameras with varying pixel sizes")
+
+        points_searched = np.dstack([x.to_value(u.m), y.to_value(u.m)])
+
+        kdtree = self.all_pixels_kdtree
+        dist, pixel_ids = kdtree.query(points_searched)
```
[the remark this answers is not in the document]
moritzhuetten: `distance_upper_bound` implemented in a34e9bd. I left the precise check for now (only done for points between circumference and actual surface), but the corresponding code can simply be removed (is only called if points haven't been sorted out already by `distance_upper_bound`). For many queried points, `distance_upper_bound` seems a factor 3 faster than the current exact check.

> At a certain size, it also might be faster to build a kdtree for the points that are looked for.

Absolutely, but I fear this might turn into some over-engineering?

ctapipe/instrument/camera.py
moritzhuetten: done in fd5d48f.

moritzhuetten:
> > Dropped assumption on central pixel in [e95ff10](https://github.com/cta-observatory/ctapipe/commit/e95ff10b3002ae1304f1ded493f80f9ff9b4dd0a), still being exact. However, the code might be much simpler and & readable if the inside-camera-check is only approximate. Or doing an exact check in a different, smarter way.
> 
> Couldn't that just be a bounding box? E.g. `(pix_x.min()-pix_rad, pix_x.max()+pix_rad)`, that's an easy and rough way to avoid computing if it's in the camera. You could compute a bounding circle as well, which might be better since most cameras are pretty circular, so `r < hypot(pix_x, pix_y).max()` or similar.

I implemented @MaxNoe's suggestion to do it approximately with the circumference on the individual pixel basis. As commented above, can easily simply remove the complicated chunk of code now.

PR #920, opened 2019-01-10
Try py3.7

dneise:
This is the continuation of PR #751. See also the discussion there as to why current the TARGET plugin is not installed for some systems and some python versions.


.travis.yml
```diff
@@ -59,9 +59,9 @@ before_install:
 
 install:
     - conda create --name cta-dev python=$PYTHON_VERSION pip
-    - conda env update -n cta-dev --file environment.yml
+    - travis_wait 20 conda env update -n cta-dev --file environment.yml
     - source activate cta-dev
-    - ulimit -s 16000  # increase stack size limit, for libhessio
+    - ulimit -s 50000  # increase stack size limit, for libhessio 50MB
```
MaxNoe: I think this is 50G:

```
$ ulimit -a | grep stack
stack size              (kbytes, -s) 8192
```

dneise: Yep you are right ... thanks ...

dneise: let's see if 50GB helps with this test

dneise:
superseded by #925 - closed

PR #923, opened 2019-01-10
Trying out Travis Caching

dneise:
In this PR I am playing with travis cache for the first time .. maybe it helps speeding up the conda build.
If you are interested .. come an play


dneise:
The cache is deleted every 28 days.

The cache can be manually deleted in this menu:
screenshot from 2019-01-10 15-38-17

Every person (as opposed to core-devs)  opening PR can switch the cache off by setting `cache: false` in the environment.yml ...

I think there are more ways.

dneise:
Regarding the environment.yml .. I don't know.

dneise:
Okay ... few insights so far
 - in order for caching to work **at all** "Build pushed branches" must be switched on, so I switched it on again for the time being.
 - packing and pushing the cache normally has a timeout at 180sec, I set it to 1000sec now .. let's see.

kosack:
Well, it does seem to reduce the test processing time from 30-50 minutes (which has blown up recently) to only 12 minutes, so I say this is a very good thing to do if possible!

The tests originally ran in 6 minutes, but since we added the Target, ZFITS, and other tests, they are quite a bit slower.

dneise:
I think we need #925 before we can further work on this

kosack:
> I think we need #925 before we can further work on this

done.

dneise:
Caching seems not to help significantly anymore since we stopped conda from up-/downgrading python behind our backs. One build takes ~16..18min now:

| task | time [sec] | comment
| -- | -- | -- | 
| before_install | 250 | install miniconda ~40; conda env update ~200 |
| pytest --cov=ctapipe | 130 | - |
|  travis-sphinx | 40 | - |
| execute ipynbs | 440 | 

When we do not go for travis caching, we can switch off the PR build again and thus avoid 2 builds on travis which safes time.

If we take a closer look on the notebooks we might be able to reduce the runtime of some of them. Maybe .. I did not yet take a closer look.

`execute_ipynbs.py` now also prints the runtime of the notebooks into the travis build output so we can identify long running notebooks quicker... starting with the next build.

dneise:
Closing this, since hardly any effect.

PR #925, opened 2019-01-17
remove optional deps in order to fix tests

dneise:
Currently the build of the master fails. This PR intends to fix this by removing optional dependencies which are not conda packages


dneise:
We found that `numpy>=1.15.4` results in conda installing a py3.6 even if we installed a py3.5 previously in that environment. So we did not really test on py3.5 since some time ... we just thought we did.

So I also removed the py3.5 test

MaxNoe:
Don't we need `codecov` for the `pytest --coverage` call?

dneise:
I do not know that

dneise:
Sorry @MaxNoe I did might have misunderstood the question. Did I modify something regarding codecov?

MaxNoe:
> Did I modify something regarding codecov?

Yes, you removed `pip install codecov` from the travis yaml

dneise:
ah yes now I see ... I removed one line too many

dneise:
I think it is needed in the `after success` part .. so I will observe the current build to see if the tests pass now .. and only then re-add the line I removed accidentally.

dneise:
I removed some tests for pyhessio earlier, but since we omitted OSX these tests might actually run again. Also CodeCov complained about these missing tests.

PR #927, opened 2019-01-23
speed up is_compatible; makes event_source factory faster

dneise:
This is a commit from #916 where @watsonjj  said, this is useful and should be in the master anyway.


dneise:
FYI coverage went down, since I added untested code.
This untested code will eventually leave ctapipe once we do the plugins, but for now I did not want to loose it.

dneise:
@watsonjj could your approve again quickly? Another commit removed your previous review .... I'm sorry.

PR #929, opened 2019-01-23
detect ctapipe_io_ plugins in eventsource factory

dneise:
preparation for io plugins containing dedicated EventSources

### DIFF
--- a/ctapipe/io/__init__.py
+++ b/ctapipe/io/__init__.py
@@ -6,10 +6,6 @@ from .hdf5tableio import HDF5TableReader, HDF5TableWriter
 from .tableio import TableWriter, TableReader
 
 
-# this needs to be the last ever import ever
-from .eventsourcefactory import EventSourceFactory, event_source
-
-
 __all__ = [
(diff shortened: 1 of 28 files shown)


kosack:
thanks for starting this! Once we have this interface, we can then require that the tests for external readers are in those readers' repos, rather than in ctapipe itself.

MaxNoe:
Why was this closed?

dneise:
Ah .. sorry, should have explained it.
So .. somehow .. I did not research how... this already made its way into the master:

https://github.com/cta-observatory/ctapipe/blob/master/ctapipe/io/__init__.py#L10

dneise:
So this is not needed anymore

dneise:
Here I am trying out if it works
https://github.com/cta-observatory/ctapipe/pull/981

MaxNoe:
See:
blame

dneise:
yeah .. but I am not aware how this stuff ended up in that PR it was unrelated... but well .. it did not harm.

PR #930, opened 2019-01-23
Remove sst1m optional dependency

dneise:
This PR is needed to show what changes would be needed in ctapipe in order to use a first plugin.
(Unless I made a mistake of course) 

In order to make it work, two lines had to be removed from `ctapipe.io.__init__.py` to avoid circular imports between the plugin and the EventSourceFactory. Accordingly the plugin cannot be tested against the master but against a version where these lines were removed and a lot of imports were changed. This is this PR.

The plugin can be seen here:
https://github.com/cta-observatory/ctapipe_io_sst1m

I seem to be unable to enable travis for that plugin.


watsonjj:
Very nice, but I think we should try to find a way around this circular dependency without removing the imports...if possible.

watsonjj:
I am also trying to find a solution

watsonjj:
It seems to be less of a circular dependency or import loop problem, and more of a restriction of the current EventSourceFactory, which requires the EventSources to be in the global namespace before its creation.

The following works as expected:
```python
(cta) 15:05 [Jason ~/Software/ctapipe_io_sst1m]  (master)  $ ip
Python 3.6.2 |Continuum Analytics, Inc.| (default, Jul 20 2017, 13:14:59)
Type 'copyright', 'credits' or 'license' for more information
IPython 6.1.0 -- An enhanced Interactive Python. Type '?' for help.
Using matplotlib backend: Qt5Agg

In [1]: from ctapipe.io.eventsourcefactory import EventSourceFactory
ctapipe_io_sst1m.__init__
Start of SST1MEventSource
END of SST1MEventSource
Obtaining EventSourceFactory.subclasses

In [2]: EventSourceFactory.subclasses
Out[2]:
[ctapipe.io.simteleventsource.SimTelEventSource,
 ctapipe_io_sst1m.SST1MEventSource,
 ctapipe.io.hessioeventsource.HESSIOEventSource,
 ctapipe.io.lsteventsource.LSTEventSource,
 ctapipe.io.nectarcameventsource.NectarCAMEventSource,
 ctapipe.io.targetioeventsource.TargetIOEventSource]
```

However if I instead try to run the ctapipe_io_sst1m tests:
```python
(cta) 15:06 [Jason ~/Software/ctapipe_io_sst1m]  (master)  $ ip
Python 3.6.2 |Continuum Analytics, Inc.| (default, Jul 20 2017, 13:14:59)
Type 'copyright', 'credits' or 'license' for more information
IPython 6.1.0 -- An enhanced Interactive Python. Type '?' for help.
Using matplotlib backend: Qt5Agg

In [1]: from ctapipe_io_sst1m.tests.test_sst1m_eventsource import test_EventSourceFactory
ctapipe_io_sst1m.__init__
Obtaining EventSourceFactory.subclasses
Start of SST1MEventSource
END of SST1MEventSource
tests

In [2]: from ctapipe.io.eventsourcefactory import EventSourceFactory

In [3]: EventSourceFactory.subclasses
Out[3]:
[ctapipe.io.simteleventsource.SimTelEventSource,
 ctapipe.io.hessioeventsource.HESSIOEventSource,
 ctapipe.io.lsteventsource.LSTEventSource,
 ctapipe.io.nectarcameventsource.NectarCAMEventSource,
 ctapipe.io.targetioeventsource.TargetIOEventSource]
```

This is because the first import is ctapipe_io_sst1m, which hasn't finished before it starts importing ctapipe.io, and therefore EventSourceFactory due to `from .eventsourcefactory import EventSourceFactory, event_source`. And as ctapipe_io_sst1m didn't finish importing, it wasn't ready to be included in the Factory subclass list.

This is obviously dodgy behaviour of the EventSourceFactory, but that behaviour is changed in #917. I have tested this PR with the Factory from #917, with the import lines kept, and all seems to work without issues.

So I think the options are:

1. Proceed with this PR, with the import line removed. The import line can be restored after the Factories are changed
2. Consider PR #917, and wait for the update to factories before proceeding with this PR
3. Maybe someone has an elegant fix for this problem that I was unable to find?

kosack:
If I understand correctly, you are removing the sst1mevent source and moving it to another package (as a plugin)? If  so , then this is fine - there should be no import of it in ctapipe, only automatically upon plugin detection.

ctapipe/io/__init__.py
```diff
@@ -18,7 +14,5 @@
     'TableWriter',
     'TableReader',
     'EventSeeker',
-    'EventSourceFactory',
     'EventSource',
-    'event_source'
```
kosack: is it necessary to remove event_source here? It is nice to be able to just type `from ctapipe.io import event_source`, since the fact that there even exists a factory can be hidden to the user. Having to explicitly import it from the sub-package is a bit cumbersome.

dneise: What do you mean by "With this plugin scheme"? Do you know a different scheme which would work?

watsonjj: I don't, I just didn't want to rule out the possibility. I think this is the right plugin scheme to use, and we should change the Factory such that it works better with it.

dneise:
this is not needed anymore

PR #932, opened 2019-01-25
clean up environment.yml

kosack:
- remove *conda-forge* from channel list and only apply it to packages that need it (to avoid upgrading all packages to bleeding-edge versions)
- remove *ctapipe-extra* from pip install, go back to using conda package instead

### DIFF
--- a/environment.yml
+++ b/environment.yml
@@ -22,7 +22,6 @@ dependencies:
   - numpy>=1.15.4
   - numpydoc
   - pandas
-  - pyhessio
   - pytest
   - pytest-cov
   - psutil
--- a/environment.yml
+++ b/environment.yml
@@ -1,15 +1,16 @@
+# A conda environment with all useful package for ctapipe developers
 name: cta-dev
 channels:
   - default
   - cta-observatory
-  - conda-forge
 dependencies:
+  - ctapipe-extra=0.2.16
   - zeromq
   - pyzmq>=17  # needed for correct function of notebooks on OSX
   - astropy
   - bokeh
   - cython
-  - gammapy
+  - conda-forge::gammapy
   - graphviz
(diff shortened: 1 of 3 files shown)


environment.yml
```diff
@@ -1,15 +1,16 @@
...
-  - conda-forge
 dependencies:
+  - ctapipe-extra=0.2.16
   - zeromq
   - pyzmq>=17  # needed for correct function of notebooks on OSX
   - astropy
   - bokeh
   - cython
-  - gammapy
+  - conda-forge::gammapy
```
dneise: nice. I did not know this was possible

kosack:
If the py3.6 and py 3.7 yaml files are now used for testing, we could clean those up as well and simplify them down to just what is needed to test (since environment.yml contains everything needed for a full interactive setup for a developer, which is a lot more than what is a true dependency)

dneise:
Ah right ... good idea.

dneise:
the py3.6 and 3.7.yaml are mostly copies of each other. Does anybody know a solution around having these copies?

kosack:
though it doesn't look like there is a large number of packages we can remove... other than maybe gammapy, since it's not used.

kosack:
Does yaml allow something like `#include`?

dneise:
nope: https://stackoverflow.com/a/15437697/1532474

I guess we would need some little tool creating these files from a template and e.g. the PYTHON_VERSION env variable .. could be easily done ... 
is it worth it?

kosack:
May not be worth so much effort.

kosack:
for now, I just updated the 3 yml files. We can make it better later, but this should offer some speed up for people who have the full anaconda installed at least since it won't try to upgrade everything.

PR #935, opened 2019-01-25
Add additional tests to Component

watsonjj:
Add some additional tests to `Component`.

Note `test_change_default_subclass_override`, which fails. Do we want a method to be able to change the default of a traitlet for all subclasses? Or should we accept this behaviour for the time being?


dneise:
I think this behavior is exactly what one would expect, so it should stay as it is.

watsonjj:
Something seems to be wrong with pytables in the master

PR #937, opened 2019-01-28
fix pandas to 0.23.4 for the time being

dneise:
I think pandas 0.24.0 is the problem. Let's see what travis says ...


dneise:
Done

PR #940, opened 2019-01-28
fixed examples that use pandas to read hdf5 

### LINKED ISSUE #938: pandas 0.24.0 breaks ctapipe
## Reminder: 

As we can see in #937 pandas 0.24.0 broke the master. It works with pandas 0.23.4
We should not too long stick to an old version of pandas but rather find out what the problem is.

kosack:
changed `read_hdf(filename)` to `read_hdf(filename, key='xxx') everywhere`.  This avoids bug in latest pandas where finding the first dataset automatically fails.

Fixes #938 (with workaround to avoid using automatic dataset detection)


kosack:
Oh, I also made a small edit to the LST bootcamp notebook: in it the single-tel parameters were labeled as DL2, but that is actually DL1, following the CTA definitions. Doesn't change the code at all though.

PR #941, opened 2019-01-28
Example notebooks in Sphinx documentation

kosack:
This PR makes it possible to add notebooks into the documentation directly using `nbsphinx`.  It moves some of the current examples/notebooks into `docs/examples` and `docs/tutorials`. I also reorganized a bit the documentation tree (to put the API docs at a lower-level)

The features are:

* the notebooks are executed during the doc build (and if fail the tests fail)
* the notebooks now *must* be stored "Stripped" (e.g. with no output), which is nicer for git
* the docs look like this!:

image

This is still a bit of a work in progress, as I need to migrate more of the notebooks to this new scheme (it takes some re-editing), but I think already it's useful.


docs/tutorials/ctapipe_handson.ipynb
```diff
@@ -0,0 +1,659 @@
...
+    "%matplotlib inline"
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "data = utils.get_dataset_path(\"gamma_test.simtel.gz\")\n",
```
dneise: I think the return value of `get_dataset_path` is a path-like or string pointing to a file containing *data* .. it is not the *data* itself. Therefore I think this should be named `path`

docs/tutorials/ctapipe_handson.ipynb
```diff
@@ -0,0 +1,659 @@
...
+   "outputs": [],
+   "source": [
+    "r0tel.waveform.shape"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "note that this is (N_chan, N_pix, N_samp)"
```
dneise: The output of Box 7 uses this terminology: `n_channels x n_pixels, n_samples`. I think a few more letters might not hurt here. Also I always wondered: **why** does the container help use a an `x` between the first two dimensions and a comma `,` between the last two?

docs/tutorials/ctapipe_handson.ipynb
```diff
@@ -0,0 +1,659 @@
...
+    "Now let's write out program"
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "data = utils.get_dataset(\"gamma_test_large.simtel.gz\") \n",
```
dneise: deprecated: --> `get_dataset_path`

docs/tutorials/ctapipe_handson.ipynb
```diff
@@ -0,0 +1,659 @@
...
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "import pandas as pd\n",
+    "\n",
+    "hillas = pd.read_hdf(\"hillas.h5\", key='/dl1/hillas')\n",
```
dneise: Assumed, somebody just sent me `hillas.h5` via email .. how would I know the key?

dneise: Maybe one could show a bit how dataset exploration works in ctapipe

mackaiver:
Nice work. While I didn't  look at each notebook in detail I think its a good idea 👍

docs/tutorials/ctapipe_handson.ipynb
kosack: yes, i think many of these noteboooks will need improvement, for now I'm just moving them over from the old dir and not making large mods other than bug fixes.

To look inside a hdf5 file there are many options `h5ls` probably is the easiest

PR #943, opened 2019-01-30
Correct filling of photo_electron_image container in SimtelEventSource

watsonjj:
The data.mc.tel[tel_id].photo_electron_image container is not correctly filled. This is a difficult task to test, because for some reason the test files do not contain an array_event['photoelectrons'] table. This has been an issue with these test files for a long time.

This PR fixes the accessing of the photoelectron information from the simtel file. It is important to also note that the array_event['photoelectron'] keys correspond to telescope index, **NOT** telescope id.

### DIFF
--- a/ctapipe/io/simteleventsource.py
+++ b/ctapipe/io/simteleventsource.py
@@ -181,6 +181,7 @@ class SimTelEventSource(EventSource):
             telescope_events = array_event['telescope_events']
             tracking_positions = array_event['tracking_positions']
             for tel_id, telescope_event in telescope_events.items():
+                tel_index = self.file_.header['tel_id'].tolist().index(tel_id)
                 telescope_description = self.file_.telescope_descriptions[tel_id]
 
                 data.mc.tel[tel_id].dc_to_pe = array_event['laser_calibrations'][tel_id]['calib']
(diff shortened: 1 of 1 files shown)


kosack:
thanks for catching that! It definitely would have broken some benchmarks.  Yes, this points to a  strong need to clean up the testing data. It will eventually be completely removed from ctapipe-extra, and be downloaded on-the-fly instead similar to gammapy.  Right now I'm a bit afraid to remove gamma_test.simtel.gz, even though it's so outdated (prod2!), since a lot of the calibration tests seem to rely on specific values from it. Once those are make more general (e..g. with dummy inputs rather than a specific event), we can migrate.

ctapipe/io/simteleventsource.py
```diff
@@ -205,12 +206,11 @@ def __generator(self):
...
-
-                data.mc.tel[tel_id].photo_electron_image = array_event.get(
-                    'photoelectrons', {}
-                ).get(tel_id)
-                if data.mc.tel[tel_id].photo_electron_image is None:
-                    data.mc.tel[tel_id].photo_electron_image = np.zeros((n_pixel, ), dtype='i2')
+                data.mc.tel[tel_id].photo_electron_image = (
+                    array_event.get('photoelectrons', {})
+                               .get(tel_index, {})
+                               .get('photoelectrons', np.zeros(n_pixel, dtype='float32'))
```
dneise: is `np.zeros` correct here? or should it be `np.nans` since we have no idea what the values are ... instead of being sure they were zeros.

dneise: I know it was also `np.zeros` before this PR .. so this PR is in principle not responsible for fixing this .. but still.

watsonjj: np.zeros was what was returned in the case of HESSIOEventSource, but I don't have a strong opinion about what to return...

For normal simtel files (prod3+), the np.zeros will never be returned anyway. Only in the case of the prod2 test files

PR #944, opened 2019-01-31
modernize formatting code for python 3.6+

kosack:
Since we dropped support for Python 3.5, we can now use some of the features of 3.6 everywhere. This PR upgrades the use of `str.format` to use f-strings instead (cleaner).  The edits were not made manually, rather I used the pyupgrade refactoring tool (https://github.com/asottile/pyupgrade)

### DIFF
--- a/ctapipe/calib/camera/gainselection.py
+++ b/ctapipe/calib/camera/gainselection.py
@@ -153,7 +153,7 @@ class ThresholdGainSelector(GainSelector):
         self.log.debug("Loaded threshold table: \n %s", tab)
 
     def __str__(self):
-        return "{}({})".format(self.__class__.__name__, self.thresholds)
+        return f"{self.__class__.__name__}({self.thresholds})"
 
     def select_gains(self, cam_id, multi_gain_waveform):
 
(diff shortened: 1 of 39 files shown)


kosack:
This is pretty uncontroversial I think and ready for review - the poor codecov is unrelated - it just detects all of the coverage problems in the affected files

kosack:
ah, it seems codacy still is not treating this as modern python code (hence it now complains about the use of f-strings). I think we can ignore that for now and I'll try to see if there is a way around it (they claim it's fixed https://github.com/codacy/codacy-pylint/issues/41 , but I seems not to be)

kosack:
It seems no option I try makes codacy understand that this is python3.6. I could simply disable Prospector, perhaps. Even disabling the error F999 seems to have no effect.

kosack:
I'll contact their support, but for now let's just accept this as is.

kosack:
I received an answer from Codacy, and it seems they are still using an outdated version of Prospector that doesn't support python 3.6+ syntax.  They suggested just disabling that pattern (Prospector_F999), so I did so.  Otherwise this PR is ready to go

Issue #873, opened 2018-12-11
Check if HillasReconstructor impact position reconstruction works for divergent pointing.


kbruegge:
Make sure the core position / impact position works for divergent pointing. Come up with unit tests for it.

Issue #924, opened 2019-01-10
predict in reco_algorithms


thomasgas:
All reconstruction algorithms are expected to overwrite this method:

https://github.com/cta-observatory/ctapipe/blob/5849db486586bc94737d171e611cc782bbe927fa/ctapipe/reco/reco_algorithms.py#L11

But the reconstruction algorithms don't overwrite this method in the required way:
https://github.com/cta-observatory/ctapipe/blob/59931737dee8b27c5e36655a538c87aee9c2c440/ctapipe/reco/ImPACT.py#L618
https://github.com/cta-observatory/ctapipe/blob/59931737dee8b27c5e36655a538c87aee9c2c440/ctapipe/reco/HillasReconstructor.py#L95
https://github.com/cta-observatory/ctapipe/blob/59931737dee8b27c5e36655a538c87aee9c2c440/ctapipe/reco/hillas_intersection.py#L52

How do we want to deal with this issue?
Substituting `tels_dicts` with `**args` solves this but it's a pretty bad fix (i guess).
Any ideas?

kosack:
Since the interface is really different for each, it's clear that this method probably is not appropriate for a virtual method, at least in the current implementation. It's more of just a guideline that there should be some method named predict for each one, but doesn't enforce any interface for that. 

This points to a need to refactor the Reconstructor class hierarchy.  I think it still makes sense to have a base class for these, since  they could share a lot of things, particuarly the output, which should be the same for each.  The question I would have is if we can refactor all Reconstructors to have a more common interface for `predict()`, that works with all types of reconstructors?  Or just take it out of the base class, and say that all reconstructors should define a predict() method that is appropriate? 

I guess the original idea was to make it as similar to scikit-learn classes as possible (e.g. having a `train()` and `predict()` method, so that a user/developer isn't confused as to how to use it.

thomasgas:
In #946 I've refactored a bit the reconstructors so that the same input can be given to both of them (was not the case before). I moved some stuff inside the `HillasIntersection` class, so that the predict method would be something like:
`def predict(self, hillas_dict, inst, array_pointing, telescopes_pointings=None):`
And if the `telescopes_pointings` are not provided, they are copied from the `array_pointing`, so that the code is independent from this input. This is not a final fix since we should have a better handling of the pointing container somehow...

maxnoe:
[...]

PR #946, opened 2019-02-01
Reconstruction fix (also for Divergent Pointing) 

thomasgas:
I'm trying to analyze some MC in divergent pointing mode. 
The reconstruction was not working out of the box due to few fixes needed to the HillasIntersection class after the coordinates refactoring in #896: seems to be working correctly now...I'm still working on this.
I've not checked the HillasReconstructor in divergent mode (see #873) yet. I just know that divergent pointing is not allowed now with HillasReconstructor.


MaxNoe:
Can you create Test cases for the issues you fixed so this does not happen again?

thomasgas:
> Can you create Test cases for the issues you fixed so this does not happen again?

The HillasIntersector class is not very well tested, so I think I can add some tests.

kosack:
> divergent pointing is not allowed now with HillasReconstructor

What do you mean by not allowed? It used to work (in a very old version, before coordinates were implemented anyhow).  In principle, the methodology in HillasReconstructor should be better for divergent pointing, since no Nominal system is needed. The HillasIntersection one was just a copy of the HESS implementation, I think, and is not used anywhere.  We should probably decide on a single method.

MaxNoe:
> divergent pointing is not allowed now with HillasReconstructor

Multiple code path in the HillasReconstructor as it is now assume a common pointing position.
I found this out when debugging the diffuse reconstruction issue and added an exception for divergent pointing.

It it is possible to support divergent pointing in principle, then these code paths must be changed.
For me, there was no obvious, simple fix, so I just added the exception to make it clear that it currently does not support divergent pointing.

ParsonsRD:
Hi, I agree Karl the HillasReconstructor should work better here where the assumption of a flat nominal system starts to become a bad assumption. I created the HillasIntersection code as a baseline comparison as this is (as far as I know) the way all current instruments do their Hillas reconstruction (and also HillasReconstructor gave weird results at the time). I also agree that we should move towards removing HillasIntersection, but we should probably show the HillasReconstructor gives similar or better results at some point.

kosack:
We should probably add a cross-check test between the two implementations, at least for some simple case (non-divergent for now I guess until both support it).

kosack:
> Also: I think we should just drop HorizonFrame and replace it by just using AltAz directly from astropy.coordinates

Agreed - if I recall correctly in HESS, there was a difference between AltAz and Horizon, but it was just related to the refraction correction (it was applied when going from Horizon to AltAz).  I think in Astropy,  that correction is applied when going from AltAz to ICRS (but we should verify), so there is no difference between Horizon and AltAz

kosack:
yes, it looks like AltAz has an optional *temperature* and *pressure* option, which implies refraction gets corrected later. (http://docs.astropy.org/en/stable/api/astropy.coordinates.AltAz.html)

thomasgas:
I run the analysis on some data produced with divergent pointing and it seems it's working well with the 2D reconstruction method.

ctapipe/reco/hillas_intersection.py
```diff
@@ -82,7 +82,7 @@ def predict(self, hillas_parameters, tel_x, tel_y, array_direction):
         sky_pos = SkyCoord(
             az=src_x * u.deg,
             alt=src_y * u.deg,
-            frame=HorizonFrame()
+            frame=HorizonFrame(location=array_direction.location, obstime=array_direction.obstime)
```
MaxNoe: Isn't it then `SkyCoord(az=..., alt=..., frame=array_direction.frame)`?

thomasgas: yes! in this way all the information in the should be passed.

MaxNoe:
So the tests are passing, but diff coverage is zero. So there are no tests covering `HillasIntersection`.

kosack:
Yes, there need to be some unit tests if we want to keep this class - it's one of the reasons it has not been used. At least similar ones to those in HillasReconstuctor.  Ideally, we could have common tests for all reconstructors, but so far the interfaces are not quite uniform.

thomasgas:
> So the tests are passing, but diff coverage is zero. So there are no tests covering `HillasIntersection`.

Yes I know...I have in my to-do list to add some tests to this class...if someone wants to help, that would be great! 
Plus we need some tests to compare the output of the two reconstruction methods.

thomasgas:
I'm testing the output of the two reconstructors and it seems that the 3d gives better results that the 2d one. 
I'm ok for dropping the 2d as soon as we are sure about this and if I find a way of implementing the reconstruction for divergent pointing with the 3d method.
It's also easier to use the 3D method than the 2D one, since in the 3d method there's no need to use particular coordinate frames.

ParsonsRD:
Great! So does it give a significantly better angular resolution or just small improvements? Either way is good, but if it's much better it might be nice to make a plot of this to show to consortium to show that we are able to improve on "standard method" even with a fast Hillas based analysis.

thomasgas:
> Great! So does it give a significantly better angular resolution or just small improvements? Either way is good, but if it's much better it might be nice to make a plot of this to show to consortium to show that we are able to improve on "standard method" even with a fast Hillas based analysis.

I tested it on few data and the 3D one is 50% to 100% better :D but i need more data and different pointings in order to be sure about this. And if someone else could do the same check, it would be great.

ParsonsRD:
> tested it on few data and the 3D one is 50% to 100% better

Wow! That is impressive, if that holds for all events we should definitely show this somewhere, as I'm not sure that this method has been used by IACTs before.

thomasgas:
I changed the title since I'm doing some fixes also to ImPACT. 
The changes in ImPACT are just to solve bugs and decrease the computing time, whereas those in HillasReconstructor are needed for the divergent pointing.

kosack:
What is the status of this PR? It's still marked as WorkInProgress, but seems to be nearly ready (other than some style and coverage issues)?

thomasgas:
> What is the status of this PR? It's still marked as WorkInProgress, but seems to be nearly ready (other than some style and coverage issues)?

I should add some unit tests for the HillasIntersection reconstructor and then it can be merged. If somebody can help in writing the unit tests, please do it :)

kosack:
What sort of tests are needed? we certainly need to add some more reconstruction tests. Perhaps one could construct some dummy Hillas parameters sets, hand-coded to give a certain result, and check if that result comes back?

thomasgas:
I took some time to refactor a bit the `HillasIntersection` class in order to use as the `HillasReconstructor` is used. While writing the unit tests I also fixed some bugs and I think that now it's working properly again.
- The idea is to calculate the x and y of the cog in the `CameraFrame` as it is done usually and then port the points everything in the `NominalFrame`
- I've tested both classes on the same datasets and the performances of the 2D method is slightly worst wrt the 3D one...still checking though (I tested on some diffuse file, but I'd like to run it on more diffused gammas).
- Since I feel more confident to use the 3D method for the divergent pointing, I'm not gonna fully implement the 2D method to be used in divergent mode.
- Now both classes gives the `h_max`, while before one was giving `h_max` and the other was calculating `x_max`: I left that commented since there's still the issue of calculating the `h_max` wrt the sea level (see #1030).
- Codacy complains for the misuse of the predict method...
Comments are welcome!

thomasgas:
I'm still missing some unit tests for ImPACT but I'll do it later in another PR, otherwise this becomes too big.

thomasgas:
The two reconstructors can also be easily added to the **cta-benchmarks** since the interface is the same.

kosack:
Is `divergent_mode` really needed as an option? I would guess that in the general case where telescope have a pointing error, the correction you make is the same as what  you would do for divergent pointing, so perhaps you want to do the divergent correction always? We only ever have the case where the `array_pointing == telescope_pointings` in an ideal monte-Carlo, and there may soon be MCs with real pointing-errors included, so perhaps that special case should not be the main one? You could then even get rid of array_pointing and calculate it from some barycenter of the telescope pointings.

thomasgas:
> Is `divergent_mode` really needed as an option? I would guess that in the general case where telescope have a pointing error, the correction you make is the same as what you would do for divergent pointing, so perhaps you want to do the divergent correction always? We only ever have the case where the `array_pointing == telescope_pointings` in an ideal monte-Carlo, and there may soon be MCs with real pointing-errors included, so perhaps that special case should not be the main one? You could then even get rid of array_pointing and calculate it from some barycenter of the telescope pointings.

Yes I can remove the flag...it makes more sense to always do the divergent correction when the pointing is not exactly parallel (pointing corrections or divergent mode): I can just check that if the `telescope_pointings` is provided, the correction is applied.
Concerning the `array_pointing` I also thought at some point to get rid of that and use a barycenter of the telescope pointings, but it needs a bit of investigation. Having another reconstructor which also does the reconstruction of the impact point and of the position in the sky in one shot (and in 3D) would naturally remove the need for an `array_pointing`. That's material for separate PR.

thomasgas:
More details on the fix for the divergent mode with the `HillasReconstructor` can be found here: https://indico.cta-observatory.org/event/2114/contributions/19047/attachments/15148/19178/ASWG_Divergent_Montpellier.pdf

ctapipe/reco/tests/test_hillas_intersection.py
```diff
@@ -42,5 +52,229 @@ def test_parallel():
...
+
+            try:
+                moments = hillas_parameters(geom, pmt_signal)
+                hillas_dict[tel_id] = moments
+            except HillasParameterizationError as e:
+                print(e)
+                continue
+
+        if len(hillas_dict) < 2:
+            continue
```
MaxNoe: We should make sure that at least one event was tested!
If no events survive, we will not test the hillas intersection. Please add a check that at least one, or a known number of events were indeed tested.

ctapipe/reco/HillasReconstructor.py
```diff
@@ -108,9 +108,11 @@ class that reconstructs the direction of an atmospheric shower
     def __init__(self, config=None, parent=None, **kwargs):
         super().__init__(config=config, parent=parent, **kwargs)
         self.hillas_planes = {}
+        self.divergent_mode = False
+        self.corrected_angle_dict = {}
 
-    def predict(self, hillas_dict, inst, pointing_alt, pointing_az):
-        '''
+    def predict(self, hillas_dict, inst,  array_pointing, telescopes_pointings=None):
```
kosack: I'd suggest updating the base class as well either to have the same signature, or to remove this as an abstract function. In the former case, it might also imply a change to `ImPACTReconstructor` to have it use the same basic arguments. In that case it's not `hillas_dict` but `image_dict` or something, so maybe coming up with a more generic name would allow all to use the same interface (`telescope_features_dict` or something)

ctapipe/reco/HillasReconstructor.py
```diff
@@ -155,24 +159,27 @@ class are set to np.nan
...
+
         self.initialize_hillas_planes(
-            hillas_dict,
-            inst.subarray,
-            pointing_alt,
-            pointing_az
+                hillas_dict,
+                inst.subarray,
+                telescopes_pointings,
+                array_pointing
```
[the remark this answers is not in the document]
thomasgas: yep

ctapipe/reco/HillasReconstructor.py
```diff
@@ -203,38 +210,39 @@ class are set to np.nan
...
-            needs to contain at least the same keys as in `hillas_dict`
+        telescopes_pointings: dictionary
+            dictionary of pointing direction per each telescope
+        array_pointing: SkyCoord[AltAz]
+            pointing direction of the array
         """
+
         self.hillas_planes = {}
-        horizon_frame = AltAz()
+        horizon_frame = list(telescopes_pointings.values())[0].frame
```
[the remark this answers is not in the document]
thomasgas: ah good idea!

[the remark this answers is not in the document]
thomasgas: `horizon_frame = telescope_pointings[k].frame`, right?

ctapipe/reco/hillas_intersection.py
```diff
@@ -44,14 +45,22 @@ class HillasIntersection(Reconstructor):
     for multiplicity 2 events.
     """
 
-    def __init__(self, atmosphere_profile_name="paranal"):
+    def __init__(self, atmosphere_profile_name="paranal", weighting="Konrad", config=None, parent=None, **kwargs):
```
kosack: Make these parameters traits of the Component rather than explicit in the constructor (see other Components).  Then all you need is config, parent, and kwargs here. 

e.g. 
```py3
from ctapipe.core import traits

class HillasIntersection(Reconstructor):
    atmosphere_profile_name = traits.CaselessStrEnum(
        ['paranal',], 
        default_value="paranal", 
        help="name of atmosphere profile to use"
    ).tag(config=True)
    
    weighting = traits.CaselessStrEnum(
        ['konrad','hess'],
        default_value='konrad',
        help = 'Weighting Method name'
    ).tag(config=True)

...
...
```

ctapipe/reco/tests/test_hillas_intersection.py
thomasgas: I copied this from test_HillasReconstructor.py...should I change both files?

MaxNoe: guess so

MaxNoe: If it's copied, can we just have one test that loops over the reconstruction methods?

thomasgas: > If it's copied, can we just have one test that loops over the reconstruction methods?

sure

MaxNoe: you could use this parameterize feature of pytest

thomasgas: I was wondering how to do it :D thanks for the suggestion...I'm having a look at it right now

thomasgas: I just pushed a new commit adding an assert over the minimum number of reconstructed events. Could you check that? If it's fine I'll just put that in a single test which tests both reconstruction methods.

ctapipe/reco/tests/test_HillasReconstructor.py
```diff
@@ -82,7 +82,8 @@ def test_h_max_results():
     # np.testing.assert_allclose(fitted_core_position.value, [0, 0], atol=1e-3)
 
 
-def test_reconstruction():
+@pytest.mark.parametrize("max_events, reco_num_events_expected", [(10, 0)])
```
MaxNoe: This two paremeters can be fixed, right? I thought about using something like this:
```
@pytest.mark.parametrize("reconstructor", [HillasIntersection, HillasReconstructor])
```

thomasgas: I wrote a new test which run both reconstructions.
I used a feature called pytest.fixture to collect the functions and to pass them to the test. If I ask to have at least 40 telescopes in the `hillas_dict`, the test fails...which means it works properly.

ctapipe/reco/hillas_intersection.py
thomasgas: done.

ctapipe/reco/HillasReconstructor.py
thomasgas: Discussion started in #924 . Shall we continue there?
I might only say that by now ImPACT is a bit different so we might think of its refactoring later on.

thomasgas:
I implemented the suggested modifications (a part for the predict method which I think deserves some discussion).
I added some the warnings also to `hillas_intersection` and moved the warnings to the common reco_algorithms. Now both reconstructors import the warnings from there.
There is a test which loops over both reconstructors. We might clean a bit the duplicate tests and just keep the common one, moving the common tests there.

ctapipe/reco/HillasReconstructor.py
thomasgas: @kosack what do you think about this?

kosack: yes, we can deal with that later

thomasgas:
A part from codacy complaining, is there any other problem in merging this PR? could someone review it? Should I merge the `master` into this branch again before merging the PR?

kosack:
I think we're just waiting for @MaxNoe (who's probably quite busy) or @ParsonsRD  to give an ok. I Don't see any major issues.  merging in master shouldn't be necessary since it doesn't touch any other changed files,  but you could do it if you like. 

The codacy issues look ok - it's mainly complaining about missing docstrings in the tests, which aren't really critical, so I'm ok ignoring that.

ParsonsRD:
OK I'll try to take a look today

kosack:
Ok, since we have 2 reviews and Max's requests seem to be addressed, I'll merge this.

PR #947, opened 2019-02-02
Improve Component argument consistency

watsonjj:
There are inconsistencies in the arguments for initialising different Component classes.

For example in some places one will see:
```python
def __init__(parent=None, config=None):
```
And in other places:
```python
def __init__(config=None, tool=None):
```

I have chosen to correct the former to the latter, as the latter appears to be used by the majority of sub-`Components` that people will have called in their own scripts.

This change is required to allow the `Component.from_name` functionality in #916 to work consistently.


MaxNoe:
mmh.

Is `tool` really the better choice here? I think it should be the more general `parent`.

watsonjj:
> mmh.
> 
> Is `tool` really the better choice here? I think it should be the more general `parent`.

If that is the consensus, then I will switch. I just wanted to avoid changing the arguments to classes that people may have used in their own scripts.

watsonjj:
I also have never seen a case where `parent` was anything other the instance of a `Tool`

dneise:
I think `tool` as parameter name is a little bit more explicit here. I remember when I first started looking at Components (but had never looked into `Tool` development yet) I was very confused by these arguments to init. 

`parent` sounded to my like this Component wanted an instance of a ParentClass of itself ... and I had no idea why I should do this. The word parent for me always carries with it this notion of inheritance. 

`tool` on the other hand could at least hint that this is connected with the `Tools` we have in ctapipe. So a newbie might start looking in the right direction. 

However .. I have one question. Readability aside for a moment.... Wouldn't this also work?
Instead of 
```python
class ChargeExtractor(Component):
    def __init__(self, config=None, tool=None, **kwargs):
        ''' doc string explaining config and tool  and neighbours'''
        super().__init__(config=config, tool=tool, **kwargs)
        self.neighbours = None
```

we could also do:
```python
class ChargeExtractor(Component):
    def __init__(self, **kwargs):
        ''' doc stringexplaining neighbours'''
        super().__init__(**kwargs)
        self.neighbours = None
```

right? This would hide everything regarding `Tools` and hiding stuff is a bad thing, but it would also reduce noise for people looking at Components from a non Tool-developer stance.

I think `Component.__init__`  would then change to:
``` python
class Component(Configurable, metaclass=AbstractConfigurableMeta):
    def __init__(self, **kwargs):
        super().__init__(
            parent=kwargs.get('tool', None), 
            config=kwargs.get('config', None), 
            **kwargs
        )
```

the advantage here is ... if we ever want to modify something related to these two parameters, we only need to modify `Component` and all the explicit use cases ... but not all the things inheriting from Component.

dneise:
I think it would be great, if we could agree on something. In the moment it is inconsistent and Jason went out to fix this. I think this is an improvement all by itself.

PR #951, opened 2019-02-05
Avoid underscore replacement empty prefixes.

mackaiver:
The underscore replacment would match every string when the prefix was empty.

### DIFF
--- a/ctapipe/io/hdf5tableio.py
+++ b/ctapipe/io/hdf5tableio.py
@@ -151,7 +151,7 @@ class HDF5TableWriter(TableWriter):
                     continue
 
                 if isinstance(value, Quantity):
-                    if self.add_prefix:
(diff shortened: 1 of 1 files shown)


ctapipe/io/hdf5tableio.py
```diff
@@ -151,7 +151,7 @@ class Schema(tables.IsDescription):
                     continue
 
                 if isinstance(value, Quantity):
-                    if self.add_prefix:
+                    if self.add_prefix and container.prefix:
                         key = col_name.replace(container.prefix + '_', '')
```
MaxNoe: I think what we really want to do here (and there might be other places where this is needed) is:

```python
import re

key = re.sub(f'^{prefix}_', '', col_name)
```

This makes sure, `prefix_` is only removed at the beginning of the column name.

Issue #952, opened 2019-02-05
Issue with installation of ctapipe


chaimain:
https://github.com/cta-observatory/ctapipe/blob/37f70ce678a43fc1ccc956572a25e5892e73a4a9/setup.py#L63

While installing ctapipe a check was conducted on the version of pyhessio even though the new pyeventio is used now, as per the FAQ.
I tried to install the latest version of pyhessio separately as per the instructions, given in https://cta-observatory.github.io/pyhessio/getting_started/index.html and that caused an error while doing the pytest - 
> _pytest.config.ConftestImportFailure: (local('local_address/ctapipe/conftest.py'), (<class 'ValueError'>, ValueError('numpy.ufunc size changed, may indicate binary incompatibility. Expected 216 from C header, got 192 from PyObject',), <traceback object at 0x7fec64c0ab48>)

I then removed the pyhessio package and edited the setup.py file of ctapipe to comment out the check for pyhessio, but I still receive this error which I am unable to discern why.
I have python 3.6.8 currently.

thomasgas:
Sorry for being late in replying to you.
> [ctapipe/setup.py](https://github.com/cta-observatory/ctapipe/blob/37f70ce678a43fc1ccc956572a25e5892e73a4a9/setup.py#L63)
> 
> Line 63 in [37f70ce](/cta-observatory/ctapipe/commit/37f70ce678a43fc1ccc956572a25e5892e73a4a9)
>  'pyhessio>=2.1', 
> 
> While installing ctapipe a check was conducted on the version of pyhessio even though the new pyeventio is used now, as per the FAQ.

I think this comes from some unit test used to make sure that the pyhessio and pyeventio gives the same output: in order to do this you still need to install pyhessio.

> I tried to install the latest version of pyhessio separately as per the instructions, given in https://cta-observatory.github.io/pyhessio/getting_started/index.html and that caused an error while doing the pytest -
> 
> > _pytest.config.ConftestImportFailure: (local('local_address/ctapipe/conftest.py'), (<class 'ValueError'>, ValueError('numpy.ufunc size changed, may indicate binary incompatibility. Expected 216 from C header, got 192 from PyObject',), <traceback object at 0x7fec64c0ab48>)
> 
> I then removed the pyhessio package and edited the setup.py file of ctapipe to comment out the check for pyhessio, but I still receive this error which I am unable to discern why.
> I have python 3.6.8 currently.

[...]

PR #957, opened 2019-02-08
make pyhessio not necessary for normal users

dneise:
pyhessio is nowadays only needed to verify pyeventio works correctly. So some users might not even want to install it...


dneise:
Any opinions here from non pyeventio developers?

thomasgas:
> Any opinions here from non pyeventio developers?

So pyhessio would be installed only if one develops something and have to run tests which will then will require to download pyhessio, right?
If so, I'm ok for dropping the installation of pyhessio.

dneise:
Not exactly.

The tests can still be run also without having pyhessio installed. Just the 4 tests which explicitely need pyhessio will automatically be skipped for people (users or developers who do not develop something which needs pyhessio) who do not have pyhessio installed.

This automatic skip of tests is done using pytest.importorskip().

So any user can still: 
 - install ctapipe and
 - perform the tests using pytest.

They just do not need to have pyhessio installed and also do not need to alter their stack size in order to make pyhessio work.

dneise:
Thus PR might have helped in cases like #952

thomasgas:
Ok now I understood the skip in pytest. So i definitely agree!

dneise:
On travis we do still install pyhessio in order to stay sure that pyeventio and pygessio deliver the exact same results for selected simtel files.

These tests are needed to gain trust into pyeventio, but strictly speaking they should not be part of ctapipe, but of pyeventio.

PR #961, opened 2019-02-11
Add mc header information to SimTelEventSource

mackaiver:
One can access the MC header information from the SimTelEventSource now. No need to get the first event and check the header then. Since the header information never changes in a file I thought this might be convinient.

### DIFF
--- a/ctapipe/io/simteleventsource.py
+++ b/ctapipe/io/simteleventsource.py
@@ -1,7 +1,7 @@
 import warnings
 import numpy as np
 from ctapipe.io.eventsource import EventSource
-from ctapipe.io.containers import DataContainer
+from ctapipe.io.containers import DataContainer, MCHeaderContainer
 from astropy import units as u
 from astropy.coordinates import Angle
 from astropy.time import Time
@@ -25,6 +25,9 @@ class SimTelEventSource(EventSource):
             self.file_.telescope_descriptions,
             self.file_.header
         )
+
+        self.mc_header_information = MCHeaderContainer()
+        self._add_mc_header_information(self.mc_header_information)
 
     @staticmethod
     def prepare_subarray_info(telescope_descriptions, header):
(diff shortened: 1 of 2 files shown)


dneise:
Oh but wait .. there exist files with multiple runs inside. People using such files might be confused ... or not ... don't know. How do we want `self.mc_header_information` to behave during iteration? should it update once a new mc run header is encountered?

One issue where this was mentioned can be found here:
https://github.com/cta-observatory/pyeventio/issues/132

mackaiver:
> Oh but wait .. there exist files with multiple runs inside. People using such files might be confused ... or not ... don't know. How do we want `self.mc_header_information` to behave during iteration? should it update once a new mc run header is encountered?

Ah. I wasn't aware such files exist. Thanks for the heads up.  
Is there a way to test wether a file contains multiple runs without reading the entire file first?

dneise:
> Is there a way to test wether a file contains multiple runs without reading the entire file first?

If there is one, I am not aware of it.

MaxNoe:
> If there is one, I am not aware of it.

I think they are called `merged` in the cta file naming convention and the run headers come at the beginning of the file, right?

MaxNoe:
Like this file for example:
gamma_20deg_180deg_run99___cta-prod3_desert-2150m-Paranal-merged_cone10.simtel.gz

It has three run headers and three inputcards right at the beginning of the file:

```
 RunHeader[2000](size=8048, only_subobjects=False, address=5938880)
 MCRunHeader[2001](size=140, only_subobjects=False, address=5946944)
 MCRunHeader[2001](size=140, only_subobjects=False, address=5947100)
 MCRunHeader[2001](size=140, only_subobjects=False, address=5947256)
 InputCard[1212](size=29776, only_subobjects=False, address=5947412)
 InputCard[1212](size=29776, only_subobjects=False, address=5977204)
 InputCard[1212](size=29776, only_subobjects=False, address=6006996)
```

dneise:
Hm ... Thomas armstrong provided here an example file https://github.com/cta-observatory/pyeventio/issues/132#issuecomment-457192545

And if I remember correctly there was one run after the other in that file. So the header of the 2nd run came after the 1st run and not before, but I might be mistaken.

MaxNoe:
@ParsonsRD told me, that these files are merged from corsika runs with the same showers, but different subarrays.

dneise:
> So the header of the 2nd run came after the 1st run and not before, but I might be mistaken.

yep here it is .. 1st run header comes after 25kB ... then a lot of events .. and the histograms part ... and then it starts ... again with the history part .. basically just like two files were simply binary appended.

```bash
12:55 $ eventio_print_structure run1011_merged.dst0.simtel.gz | grep -i run
 RunHeader[2000](size=296, only_subobjects=False, address=25376)
 MCRunHeader[2001](size=140, only_subobjects=False, address=25688)
 RunHeader[2000](size=284, only_subobjects=False, address=9282164)
 MCRunHeader[2001](size=140, only_subobjects=False, address=9282464)
```

MaxNoe:
So it's probably not a good idea to implement this, right?

On  the other hand, @bregeon told me, the times of "Frankenstein" simtel files should be over, so probably, we could still add this.

Opinions?

moralejo:
On Fri, 22 Feb 2019 at 10:16, Maximilian Nöthe <notifications@github.com>
wrote:

> So it's probably not a good idea to implement this, right?
>
> On the other hand, @bregeon <https://github.com/bregeon> told me, the
> times of "Frankenstein" simtel files should be over, so probably, we could
> still add this.
>
> Opinions?
>

Can it be implemented such that by default it throws an error if a second
run header is found - and then those Frankenstein files can be read, if
ever needed, using an option which just turns the error into a warning?

-- 

Abelardo Moralejo Olaizola
Institut de Física d'Altes Energies
Tel : +34 931641662
Fax: +34 935811938
Avís - Aviso - Legal Notice - (LOPD) - http://legal.ifae.es

-- 
Avís -
Aviso - Legal Notice - (LOPD) - http://legal.ifae.es 
<http://legal.ifae.es/>

dneise:
I think we cannot control what files people are generating. So I doubt such times are over. 
If we say, we officially do not support such files in ctapipe, then we should fail loudly and early of we encounter one of them. Early is not really possible, but loudly.

dneise:
But anyway the current state ... before this PR .. was pretty fool proof, no? so let's just keep it as it is. 
It is less convenient, but this is the price we have to pay, I think.

MaxNoe:
> But anyway the current state ... before this PR .. was pretty fool proof, no? so let's just keep it as it is.
It is less convenient, but this is the price we have to pay, I think. 

agreed

PR #966, opened 2019-02-12
Use eventio v0.17.1

dneise:
* Can read more kinds of simtel files (e.g. where multiple runs got merged)
* Can read calibration events


MaxNoe:
This needs the not yet released version 0.17.0 of pyeventio and 0.2.17 of ctapipe-extra with the calibration events test file.

MaxNoe:
eventio 0.17.0 is now released, waiting for a new release of ctapipe-extra

dneise:
`ctapipe/io/tests/test_simteleventsource.py::test_calibration_events` fails ...

dneise:
Nice !

setup.py
```diff
@@ -65,7 +65,7 @@
           'pandas',
           'bokeh>=1.0.1',
           'scikit-learn',
-          'eventio==0.11.0',
+          'eventio==0.17.0',
```
mackaiver: should this be 0.17.1?

ctapipe/io/containers.py
```diff
@@ -428,6 +428,7 @@ class TelescopePointingContainer(Container):
 class DataContainer(Container):
     """ Top-level container for all event information """
 
+    event_type = Field('data', "Event type")
```
mackaiver: what are the options here? Is it data vs calibration event? Or is it observed vs simulated event?

dneise: Ah that's indeed a good question, maybe this PR makes sense in this context.
https://github.com/cta-observatory/ctapipe/pull/978

By defining an `enum.Enum` or `enum.IntEnum` the choices are somehow obvious...

MaxNoe: For now, data and calibration can come from eventio. But I was not really pleased with that choice. Physics and specific calibration names would be better I think.  I will change that in eventio once we agree on the names.

MaxNoe:
Setup.py is fixed, @cta-observatory/ctapipe-core-devs please review again

dneise:
@MaxNoe this PR was opened by me, so you could review already .. :-D

MaxNoe:
Ah, right.

PR #969, opened 2019-02-18
Passing masked array to hillas_parameters

watsonjj:
The error `ValueError: aweights cannot be negative` is raised when passing a masked numpy array to `hillas_parameters`.  This is due to the line `cov = np.cov(delta_x, delta_y, aweights=image, ddof=0)` where the image is the masked numpy array.

I could have sworn this used to work! This PR will attempt to allow this functionality.

### DIFF
--- a/ctapipe/image/hillas.py
+++ b/ctapipe/image/hillas.py
@@ -8,8 +8,6 @@ import astropy.units as u
 import numpy as np
 from astropy.coordinates import Angle
 from astropy.units import Quantity
-
-from ctapipe.instrument import CameraGeometry
 from ..io.containers import HillasParametersContainer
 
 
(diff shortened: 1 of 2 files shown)


watsonjj:
This was originally tested for since #601, but the tests were removed in #904. I can't see any discussion or justification for their removal.

MaxNoe:
Are you sure the fault is the masked array and not what the error actually says? Negative values in `image`?

watsonjj:
> Are you sure the fault is the masked array and not what the error actually says? Negative values in `image`?

See the test. `image_zeros` is working fine, but `image_masked` is not. It could be that the mask is not being considered in the np.cov aweights handling, and so the error is thrown on the negative values. But the problem still stands - masked arrays are not correctly handled anymore

MaxNoe:
I remember we discussed some things about masked arrays at the developer meeting in dortmund and found some pretty severe performance impacts for no or very small benefits.

Maybe we decided to drop them completely, but I don't remember.

ctapipe/image/hillas.py
```diff
@@ -70,7 +96,11 @@ def hillas_parameters(geom, image):
     # The ddof=0 makes this comparable to the other methods,
     # but ddof=1 should be more correct, mostly affects small showers
     # on a percent level
-    cov = np.cov(delta_x, delta_y, aweights=image, ddof=0)
+    try:
+        cov = np.cov(delta_x, delta_y, aweights=image, ddof=0)
+    except ValueError:
```
MaxNoe: The ValueError will also be raised in other cases. E.g. when the weights really are negative. 

Do we really want to support masked arays here? What is the advantange?

Many things, including simple sums and so on, turned out much slower using masked arrays, so why bother?

watsonjj:
I have compared approaches for restoring masked array compatibility with the intention to not impact the execution time for other methods not involving numpy arrays. These are my findings:

Selected - refers to when a subset of the array is indexed
Zeros - refers to when the masked pixels are set to zero
Masked - refers to when the the image is transformed into a masked array

Original (no numpy array compatibility):
selected - 433 µs ± 2.07 µs per loop (mean ± std. dev. of 7 runs, 1000 loops each)
zeros - 619 µs ± 2.03 µs per loop (mean ± std. dev. of 7 runs, 1000 loops each)

With a try and except clause when calling `np.cov`:
selected - 418 µs ± 926 ns per loop (mean ± std. dev. of 7 runs, 1000 loops each)
zeros - 607 µs ± 5.17 µs per loop (mean ± std. dev. of 7 runs, 1000 loops each)
masked - 1.14 ms ± 3.87 µs per loop (mean ± std. dev. of 7 runs, 1000 loops each)

When applying `np.ma.filled` at the start of the function, irregardless if it is a masked array or not:
selected - 404 µs ± 4.66 µs per loop (mean ± std. dev. of 7 runs, 1000 loops each)
zeros - 591 µs ± 4.07 µs per loop (mean ± std. dev. of 7 runs, 1000 loops each)
masked - 612 µs ± 3.47 µs per loop (mean ± std. dev. of 7 runs, 1000 loops each)

Applying `np.ma.filled` to a non-masked array contributes less than a ms to the execution time. I therefore have concluded that this is an acceptable approach. As a bonus, it ensures compatibility with masked arrays should be easy to keep in future developments of this function.

I have added descriptions in the docstring about the different forms image can have, and the corresponding performances.

PR #973, opened 2019-02-20
Waveform cleaner

FrancaCassol:
Add simple waveform-cleaner to get a baseline subtracted waveform


ctapipe/image/waveform_cleaning.py
```diff
@@ -67,6 +67,26 @@ def apply(self, waveforms):
         return waveforms
 
 
+class BaselineWaveformCleaner(WaveformCleaner):
+    """
+    Basic waveform cleaner that subtracts the waveform baseline as
+    estimated  in a window of "baseline_width" samples, shifted of
+    "window_shift" samples
+    """
+    baseline_width = Int(10, help='Define then number of samples for estimating the '
```
calispac: If `window_shift` is not equal to zero then it does not represents the `baseline_width` anymore right ?.
I would suggest to call the two parameters : `sample_start` `sample_end`

ctapipe/image/waveform_cleaning.py
```diff
@@ -67,6 +67,26 @@ def apply(self, waveforms):
...
+                                  'baseline').tag(config=True)
+
+    window_shift = Int(0, help='Define the shift of the window on which calculate '
+                               'the baseline.').tag(config=True)
+
+    def apply(self, waveforms):
+        # Subtract initial baseline
+        baseline_sub = waveforms - np.mean(waveforms[:, :, self.window_shift:self.baseline_width],
```
MaxNoe: I think this should be `self.window_shift:self.window_shift + self.baseline_width` to get a window of length `self.baseline_width`

FrancaCassol: Yes, surely, sorry. I added a test function. Should I change the arguments to sample_start sample_end?

MaxNoe: I have no opnion on what would be better, both solutions would be fine with me.

ctapipe/image/waveform_cleaning.py
```diff
@@ -78,7 +78,8 @@ class BaselineWaveformCleaner(WaveformCleaner):
 
     def apply(self, waveforms):
         # Subtract baseline
-        baseline_sub = waveforms - np.mean(waveforms[:, :, self.sample_start:self.sample_end],
+        baseline_sub = waveforms - \
```
MaxNoe: Please don't use explicit line contiuations with `\`, use parenthese.

Maybe even split like this:
```
baseline = np.mean(
    waveforms...,
)

return waveforms - baseline```

**Comment by FrancaCassol** [comment]
It would be very helpful to me if this very simple PR could be rapidly accepted (it would avoid the copy of the waveform_cleaning.py module in lstchain). Thanks!

File: ctapipe/image/waveform_cleaning.py [hunk 4]
```diff
@@ -67,6 +67,24 @@ def apply(self, waveforms):
...
+    Basic waveform cleaner that subtracts the waveform baseline
+    estimated as the mean waveform value in the interval [baseline_start,baseline_end]
+    """
+    baseline_start = Int(0, help='Start sample for baseline estimation').tag(config=True)
+
+    baseline_end = Int(10, help='End sample for baseline estimation').tag(config=True)
+
+    def apply(self, waveforms):
+        # Subtract baseline
+        baseline = waveforms - np.mean(
```
watsonjj: Very minor note, but its not really "baseline" you obtain here, but rather "waveform_corrected" instead

FrancaCassol: Hi Jason, actually it seems to me that the name an the description of this very simple and short function is enough clear.

I have now a question.  Who is supposed to accept PRs in ctapipe? For such simple type of PRs (reviewed by at least 3 persons) it would be correct to avoid long delay times.

watsonjj: Let me be clearer. The variable you obtain on line 81 is currently called baseline. But it is not baseline, it is the waveform subtracted by the baseline, and is misleading.

Indeed, this review should be accepted very quickly.

FrancaCassol: Ah, you are right, I misunderstood :-) 
Yes, l change it rapidly

FrancaCassol:
Please, could this PR be taken into account?

PR #974, opened 2019-02-20
Fix set_line_hillas

kpfrang:
The current implementation of set_line_hillas did not work for huge values of `m`. This is an quick fix which calculates the actual start end end points of the line exactly. This works also in the case of (almost) vertical lines as you can see in the plot attached.

event0

### DIFF
--- a/ctapipe/visualization/mpl_array.py
+++ b/ctapipe/visualization/mpl_array.py
@@ -237,15 +237,13 @@ class ArrayDisplay:
         coords = self.tel_coords
         c = self.tel_colors
 
+        r = np.array([-range, range])
         for tel_id, params in hillas_dict.items():
             idx = self.subarray.tel_indices[tel_id]
             x_0 = coords[idx].x.value
             y_0 = coords[idx].y.value
-            m = np.tan(Angle(params.psi))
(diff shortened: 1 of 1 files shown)


ctapipe/visualization/mpl_array.py
```diff
@@ -242,11 +242,11 @@ def set_line_hillas(self, hillas_dict, range, **kwargs):
             x_0 = coords[idx].x.value
             y_0 = coords[idx].y.value
             m = np.tan(Angle(params.psi))
```
MaxNoe: Wouldn't using polar coordinates be even more concise?
```
x = params.x + np.cos(params.psi) * np.array([-range, range])
y = params.y + np.sin(params.psi) * np.array([-range, range])
```

MaxNoe: You can even put `r` befor the loop.

MaxNoe:
While you are at it, can you replace the `.value` with `.to_value(u.m)`, this makes sure it's always meters and not just uses the internally stored unit.

MaxNoe:
Nice, let's just wait for travis

kpfrang:
There was some accidental typo, which is fixed now.

Issue #972, opened 2019-02-19
Different event types / trigger types


maxnoe:
As far as I can see, we have no information in the event container, what kind of event we have.

The next eventio version will also be able to read calibration events (e.g. pedestals).

Where should this information go and how?

First idea would by a field `event_type` for the `DataContainer`.

Should that be a number or a string?

dneise:
I would prefer the most explicit, but I assume people wanna optimize for size here. So I'd say an int key, for a `event_type_int_2_string_map` dict, which must be available as "meta" information. 

:-D And no code in the wild must ever *use* the ints ... every code must use the strings, to make sure the code is robust against mapping changes. (of course we must avoid changes in the mapping). 

Can we story python enums in containers? that might solve the issue entirely...

MaxNoe:
> event_type_int_2_string_map
that's an enum

dneise:
Sorry Max .. I wrote "`event_type_int_2_string_map` dict"   .. so that was .. for me a dict, since I wrote "dict".

By enum I mean this: https://docs.python.org/3/library/enum.html

dneise:
I'll just try it out.

MaxNoe:
This came up today again.

For reference, simtel files can contain `CalibrationEvents` that can already be read by pyeventio and can be inserted into the event stream of the SimTelEventSource by setting `skip_calibration_events=False`.

Simtel has 4 or more types of calibration events:

>  *  @param type  Dark (0), pedestal (1), in-lid LED (2) or laser/LED (3+) data.

From `sim_conv2hess.c` line 2638.

How do we want to deal with event / trigger types here for now?

FrancaCassol:
Hi,
I would suggest to copy them in the R0CameraContainer::trigger_type (as in data)

MaxNoe:
As this information is independent of data level (true for telescope event as a whole), r0 is not the right place I think. What happens when this information is needed at the DL1 level with r0 discarded?

FrancaCassol:
I guess the trigger type (R0/R1) is kept  somewhere in the DL1 format (e.g. in the event/telescope/parameters table)

kosack:
[...]

PR #978, opened 2019-02-20
allow enums in containers and support in tableio

dneise:
Connected slightly to #972

just, in case we want to use `enum.Enum` somewhere I thought we might want to be able to serialize them as well. 

Not sure this is the best way to do it. I've read somewhere that HDF5 natively supports storing enumeration types, so that it deals with the back and forth trafo from the numeric value to the actual enum instance. But for the moment I did this in ctapipes own io layer.

### DIFF
--- a/ctapipe/io/hdf5tableio.py
+++ b/ctapipe/io/hdf5tableio.py
@@ -1,3 +1,4 @@
+'''Implementations of TableWriter and -Reader for HDF5 files'''
 import enum
 from functools import partial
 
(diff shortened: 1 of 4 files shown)


dneise:
Codacy is just complaining about invalid syntax again -> f-strings. Apart from that, I think it might be fine.

MaxNoe:
There is a class `IntEnum` in `enum`. Might this obliviate the need to write custom code to write it out?

dneise:
> Might this obliviate the need to write custom code to write it out?

Might ...  but I think converting back into an IntEnum upon reading is impossible without storing the enum class in the meta data. And this must be done upon writing it, no? So what is the point?

MaxNoe:
Ah sure.

kosack:
> Codacy is just complaining about invalid syntax again -> f-strings. Apart from that, I think it might be fine.

I contacted them and got the response that they do not yet support python 3.6+   (!!) but are looking into it. So generally we can ignore those errors. I thought I had disabled them in fact, but perhaps it has to be done for each branch?

dneise:
@cta-observatory/cta-devs I think this might be mergeble

PR #981, opened 2019-02-21
remove sst1m event source and use plugin instead

dneise:
let's see if plugins work already


dneise:
Okay so this works. Now the plugin needs to be pip installable and we need to decide if we want to test the plugins somehow in ctapipe, or we do not need that, as long as the plugins test themselves against ctapipe.

At the moment , the plugin used here, does not test itself.

dneise:
Okay so the sst1m plugin now has travis running and tests itself against the current master of ctapipe. 
https://github.com/cta-observatory/ctapipe_io_sst1m
So we do not need to test the plugin here. 
This is the particular test, where the plugin tests, if it is correctly detected by `ctapipe.io.event_source()`
https://github.com/cta-observatory/ctapipe_io_sst1m/blob/master/ctapipe_io_sst1m/tests/test_sst1m_eventsource.py#L62

I think all plugins should perform such a test... but we of course cannot force that. 

For this test of the plugin, the plugin needs of course install a certain version of ctapipe, which is a pain, since we hardly ever tag a version...
So to mitigate this, the plugin tests itself against a certain commit: "the master of 22.02.2019" .. really not nice.

ctapipe/io/utils.py
```diff
@@ -1,6 +1,8 @@
 import gzip
 
-
+# This function is currently needed by some EventSource, which will soon
+# be removed and converted into plugins.
+# Then this can be removed from ctapipe
 def is_fits_in_header(file_path):
```
MaxNoe: This function does not really test for fits compatiblity. `FITS` is not a required keyword. However, the FITS standard clearly states:

> The SIMPLE keyword  is required to  be  the  first  keyword in the primary header of all FITS files.

So, we only need to read the first 6 bytes and check if they are `b'SIMPLE'`

dneise: nice.

kosack: should also probably rename it to `is_fits_format()` or something like that, if you are no longer testing for "fits" in the header.

MaxNoe:
Could you try rebasing this against the current master? It shows a lot of changes unrelated to sst1m event source

dneise:
Nope, I am too stupid for a rebase, I think. I opened a new PR #1011

Issue #971, opened 2019-02-19
Stop guessing telescope descriptions


MaxNoe:
I think we should just use all the information simtel files provide.

Correct me if I'm wrong, but I think for prod3, all necessary information is there.
The only thing me might want to do, is to guess a human readable name, e.g. `astri`.

Is something else missing?

MaxNoe:
Done in #982

Issue #738, opened 2018-05-03
Proposal: allow for selectable sets of datasets/configuration


kosack:
Right now we store all defaults and configuration tables in the `ctapipe-extra/ctapipe_resources` package, and the user can override them by setting the `$CTAPIPE_SVC_PATH` environment variable (which takes a directory or list of directories where to look for service files).  

This is fine for simple use cases, but is getting cumbersome when we switch simulation versions, or even later when we have multiple sites/configs.  I propose that we think of developing a more flexible system, which can allow sub-sets to be enabled by the user using a single name describing the full set of data. 

The simplest realization is to start to structure the `ctapipe_resources` directories, and allow for these different data versions (note in CTA parlance, these are *SVC* (service) data).  Perhaps:

```
ctapipe_resources/
    prod3b/
    prod4/
    ...
```

In each sub-dir, we could further organize things to make it more clear:

```
prod3b/
     cameras/
          NectarCam.camgeom.fits.gz
          FlashCam.camgeom.fits.gz
          ...
          gain_channel_thresholds.ecsv
          optics.ecsv.txt
    atmosphere/
          paranal.atmprof.fits.gz
          atmprof26.ecsv.txt
    data/
          gamma_test.simtel.gz
          gamma_test_large.simtel.gz
```

Then we provide a method to set the current data sub-set `ctapipe.set_svc_data("prod3b")` or something like that if they want to switch from the default. From an API standpoint, not much would change, since `utils.get_dataset_path()` and related functions would remain the same, just automatically look in whichever one is specified, and they already have support for searching subdirectories. 

Users could still specify custom paths using `$CTAPIPE_SVC_PATH`, but those should point to paths with sub-dirs in them with new names (e.g `my_custom_svc_data/`, which would be enabled by `ctapipe.set_svc_data("my_custom_svc_data")`.

Any better ideas?

kosack:
Note: this concept is not so different from how the IRFs are being organized for the science tools (currently using a *caldb* structure, but that's a bit cumbersome)

thomasarmstrong:
Hi Karl, 

With the configurations, would it make any sense to connect with the yaml files currently being constructed for the MC? Although I guess the problem might be that parameters are stored in non-ideal formats...
 https://forge.in2p3.fr/projects/cta_analysis-and-simulations/repository/show/Simulations/MCModelDescription/trunk/configReports

STSpencer:
[...]

PR #982, opened 2019-02-21
Refactor instrument  module

MaxNoe:
Working on #971 

* Only guesses telescope names, not full camera geometries or optics
* Refactoring of the members of `TelescopeDescription`, `OpticsDescription`
* Make those classes hashable, so set operations work and they can be used as dict keys


dneise:
Btw, I was curious if one can find the telescope names somewhere in the eventio file, so that guess could potentially avoided completely. It found them in the "History" part, I think, but only in comments looking like this:

```
ECHO ************************************************************************
ECHO Configuration for single-reflector SST-1M 5 in Prod-4.
ECHO ************************************************************************
ECHO 
ALTITUDE 2000. % m (Typically replaced on the command line).
ATMOSPHERIC_TRANSMISSION atm_trans_2000_1_10_0_0_2000.dat
TRIGGER_TELESCOPES 1           % Unless an array trigger file sets up more specific triggers.
[.. more configuration ..]
```

I guess this is not a good way to read the telescope name from the file, but I wanted to mention it anyway.

MaxNoe:
The tests now failing are due to  testing for magic numbers in the old simtel file.

Any advice on how to replace those values?

mackaiver:
> The tests now failing are due to testing for magic numbers in the old simtel file.
> 
> Any advice on how to replace those values?

If we find no other way to check for correct values then these tests can be removed imo.

kosack:
>I guess this is not a good way to read the telescope name from the file, but I wanted to mention it anyway.

I know that I had mentioned that it would be nice to have the names to Konrad, and I think he said he would add that in a future version. It may be there already somewhere. In any case, we still probably want to map names there to "standard" ones we define, so there is no confusion.

kosack:
> If we find no other way to check for correct values then these tests can be removed imo.

was the example file updated? If so, yes we should remove all tests that assume specific values. It's better to give the tests a hand-made dummy event where we know the correct outcome, rather than using a real event (which can be tested at a higher level if necessary).  Probably that should be a separate PR.

kosack:
Maybe not necessary for this PR, but it would be nice to add a "guess" function for the SubarrayDescription itself, so that at least the Prod ID and the array layout could be shown (not sure if that is info directly in the MC headers or not). Right now all subarrays are called "MonteCarloArray", which is not so useful to identify them.

MaxNoe:
> I know that I had mentioned that it would be nice to have the names to Konrad, and I think he said he would add that in a future version. It may be there already somewhere. In any case, we still probably want to map names there to "standard" ones we define, so there is no confusion.

I just produced small test files using the current developement version of simtel_array (prod4 baselines la palma and paranal) and I found no changes in the containers that describe the telescopes.

MaxNoe:
> was the example file updated? If so, yes we should remove all tests that assume specific values. It's better to give the tests a hand-made dummy event where we know the correct outcome, rather than using a real event (which can be tested at a higher level if necessary). Probably that should be a separate PR.

Yes, I changed the `example_event` to come from the prod3 `gamma_test_large.simtel.gz`

MaxNoe:
I removed the magic number tests, test pass now.
I opened an issue here: #988 to get real tests for those features.

MaxNoe:
Tests are passing, please review @watsonjj @kosack @dneise @ParsonsRD

ctapipe/instrument/guess.py
```diff
@@ -0,0 +1,55 @@
...
+
+GuessingKey = namedtuple('GuessingKey', ['n_pixels', 'focal_length'])
+GuessingResult = namedtuple(
+    'GuessingResult', ['type', 'name', 'camera_name', 'n_mirrors']
+)
+
+
+TELESCOPE_NAMES = {
+    GuessingKey(2048, 2.28): GuessingResult('SST', 'GCT', 'CHEC', 2),
+    GuessingKey(2368, 2.15): GuessingResult('SST', 'ASTRI', 'ASTRICam', 2),
```
watsonjj: Can an entry be added for:
`GuessingKey(2048, 2.15): GuessingResult('SST', 'ASTRI', 'CHEC', 2),`

mackaiver:
So the difference between the two test fles is the MC production? Prod2 vs Prod3? Maybe we should rename that test file to something conveying more information.

MaxNoe:
Yes, but this would be fore ctapipe-extra. I can also produce more sensible test files not, I got a recent simtel running.

ctapipe/instrument/guess.py
MaxNoe: done

MaxNoe:
Please review again, I added @watsonjj comment.

kosack:
> Yes, but this would be fore ctapipe-extra. I can also produce more sensible test files not, I got a recent simtel running.

It would be nice to have some smaller nicer test files and get rid of the old ones in ctapipe_extra which is becoming too large already: at least something like 10+ gamma rays for north and south configurations, and some muon-only files. For deeper/larger test files, we can just make them automatically download from the server, rather than put them in the package.   In the future, we will move to selectable configurations as in #738, so this is just a temporary fix.

PR #992, opened 2019-02-27
Remove LSTEventsource,

MaxNoe:
Got moved to https://github.com/cta-observatory/ctapipe_io_lst


ctapipe/io/nectarcameventsource.py
```diff
@@ -242,3 +247,83 @@ def fill_r0_container_from_zfile(self, event):
             event
         )
 
+class MultiFiles:
+
+    """
+    This class open all the files in file_list and read the events following
+    the event_id order
```
kosack: Could use a better description and class name - does this merge multiple camera files? Or chain together multiple observations? I guess the former?  Does it provide an EventSource interface? Does it keep all files in memory?

MaxNoe: I would not do this here, the nectarcam got also removed, see #993

MaxNoe: So, though this class was moved here, because nectarcam dependent on lstcam internals, (and also sst1m internals, btw), we should ignore this and just merge both PRs.

Issue #967, opened 2019-02-14
HillasReconstructor and PSF


jjlk:
Hi everyone, @thomasgas 

The geometrical reconstruction with the method implemented in the `HillasReconstructor` class gives interesting results when one computes the point spread function (PSF) for event with  multiplicity 2. You can see below:
- Left, PSF with a 68 % containment radius as a function of true energy for gamma-rays seen by the LST subarray for different event multiplicities
- Right, corresponding number of events 

![psf_prod_lsts_array_south_zen20_az0](https://user-images.githubusercontent.com/15717586/52783492-54393e80-3052-11e9-85c6-97b1c4a4b498.png)

I'm using ctapipe 0.6.0 and for both plots, the events come from ON-axis gamma-ray simulation at the Southern site, for 20 deg zenith and North azimulth (production called, Prod3_Paranal_Baseline_NSB1x_gamma_North_20deg_DL0). I applied a cleaning base on tail-cuts (3/6) and in order to reconstruct the shower I selected the images with the following cuts:
 - intensity >= 50 p.e.
 - #pixels >=3
 - 0.1 <= ellipticity <= 0.6
 - nominal distance (distance between the center of the camera and the barycentre of the charges) <= 80 % of the camera radius

I didn't find an explaination for the 200 GeV 'bump' for events with multiplicity 2 by looking at the image or the shower properties. Note that the effect seems also present for multiplicty 3 but with a lower intensity.

@thomasgas can reproduce this behaviour independently.

Is there an expert that would be willing to have a look at this problem? I'll be happy to help by  providing high-level diagnostics but I guess that a closer look at the reconstruction implementation will be necessary.

Thanks!

thomasgas:
Hi Julien! Thanks for reporting this here!
My first guess at this is that when the impact point on the ground is more or less along the line connecting the two triggered telescopes, then the reconstruction is worst: maybe we see this just at lower energies due to the small images that we have on the cameras, but I'm just doing some speculation.
Experts are welcome!

vuillaut:
Well, it's not at lowest energies.
There are more events below 100GeV and the reconstruction is better.

thomasgas:
> There are more events below 100GeV and the reconstruction is better.

Ah yes you are right

thomasgas:
hi @jjlk, do you still have this issue with 2 LSTs? 
Can @kpfrang weighting method using a LUT for the distance of closest approach be useful?

kpfrang:
[...]

Issue #772, opened 2018-08-03
Corrupted weighting in HillasReconstructor for small images


kpfrang:
Playing around with `HillasReconstructor`, I realized that a rather large amount of events are not reconstructed. I tracked down the problem to be the weighting of the `HillasPlane` which currently is taken as `weight=moments.intensity * (moments.length / moments.width)`.  
However, depending on the the Hillas parametrization this can lead to conflicts. For small images, where the pixels after image cleaning are all in a line, the width will be either `nan` or `0` as you can see right here:

![image](https://user-images.githubusercontent.com/15632125/43634449-615927c2-970c-11e8-8126-11ef516bf55f.png) ![image](https://user-images.githubusercontent.com/15632125/43634454-6413bee6-970c-11e8-9501-20703088be85.png)

For the weighting, the means that those planes will have `nan` or `inf` as weights respectively, and those mess up the reconstruction e.g. when calculating the weighted sum of the crossings.

I think, one might still be able get some information from those images, so there might be better ways than simply throwing them away in `HillasReconstructor`. 
- there should be no `nans` as the width for correctly parametrized images but only numbers,  I guess. This would already remove the possibility of `nan` as weight
- instead of 0 one could use some very small positive number as minimum width, e.g. a fraction of the pixel size
- closely related to the last point one could put a limit on the maximum weight of the planes to avoid `inf`
- in case there is no other suited solution, one should definitely remove those images in `HillasReconstructor` so that the reconstruction can be performed with rest of images

One problem I see with the solutions would be that you somehow will set the weight 'by hand' and consequently the weights given to them might not be appropriate.  
I'm currently playing around with an other possibility of avoiding this by using different weighting methods for the planes, e.g. from a look up table.

vuillaut:
Hi
- I agree, width should not be equal to `nan` is the first case you show
- the weighting formula could be change to `intensity * (1 + length)/(1 + width)`
- in all case, one bad image giving `nan` or `inf` parameters (for any reason) should be simply thrown without throwing the event

MaxNoe:
> I agree, width should not be equal to nan is the first case you show

I disagree, `NaN` is the correct value for width in this case.

vuillaut:
> I disagree, NaN is the correct value for width in this case.
How so?
And why the event on the left would have `width = nan` and the one on the right gets `width = 0` ?
There is just a rotation between the two.

MaxNoe:
[...]

PR #994, opened 2019-02-28
Fix HillasReconstructor in case of too few telescopes

LukasNickel:
@mackaiver , @MaxNoe and me faced an issue where the HillasReconstructor would fail even though there were two telescopes left after cleaning. It turned out that you could encounter situations where one of the hillas ellipses had a width of np.NaN because there were only a few pixels remaining which are all aligned.
This would render the weights useless and leave the h_max estimation with a singular matrix.

We decided that the HillasReconstructor should check for valid telescopes (that is telescopes with a defined width of the hillas ellipse) and that we would not even try to reconstruct other events (which should be very few if any).

Is there any reason against this approach?

### DIFF
--- a/ctapipe/reco/HillasReconstructor.py
+++ b/ctapipe/reco/HillasReconstructor.py
@@ -134,8 +134,10 @@ class HillasReconstructor(Reconstructor):
         # filter warnings for missing obs time. this is needed because MC data has no obs time
         warnings.filterwarnings(action='ignore', category=MissingFrameAttributeWarning)
         
-        # stereoscopy needs at least two telescopes
-        if len(hillas_dict) < 2:
+        # stereoscopy needs at least two telescopes with a valid width of the hillas ellipse
+        valid_telescopes = sum([1 if hillas_dict[x]['width'].value is not np.NaN else 0 for x in hillas_dict])
+
+        if valid_telescopes < 2:
             raise TooFewTelescopesException(
                 "need at least two telescopes, have {}"
                 .format(len(hillas_dict)))
--- a/ctapipe/reco/HillasReconstructor.py
+++ b/ctapipe/reco/HillasReconstructor.py
@@ -133,14 +133,16 @@ class HillasReconstructor(Reconstructor):
 
         # filter warnings for missing obs time. this is needed because MC data has no obs time
         warnings.filterwarnings(action='ignore', category=MissingFrameAttributeWarning)
-        
-        # stereoscopy needs at least two telescopes with a valid width of the hillas ellipse
-        valid_telescopes = sum([1 if not np.isnan(hillas_dict[x]['width'].value) else 0 for x in hillas_dict])
+
+        # stereoscopy needs at least two telescopes with valid widths
+        valid_telescopes = sum([1 if not np.isnan(hillas_dict[x]['width'].value)
+                                else 0
+                                for x in hillas_dict])
 
         if valid_telescopes < 2:
             raise TooFewTelescopesException(
                 "need at least two telescopes, have {}"
-                .format(len(hillas_dict)))
+                .format(valid_telescopes))
 
         self.initialize_hillas_planes(
             hillas_dict,
--- a/ctapipe/reco/HillasReconstructor.py
+++ b/ctapipe/reco/HillasReconstructor.py
@@ -29,10 +29,14 @@ import numpy as np
 from astropy import units as u
 
 
-__all__ = ['HillasReconstructor', 'TooFewTelescopesException', 'HillasPlane']
+__all__ = ['HillasReconstructor', 'TooFewTelescopesException', 'HillasPlane', 'InvalidWidthException']
 
 
 class TooFewTelescopesException(Exception):
+    pass
+
+
+class InvalidWidthException(Exception):
     pass
 
 
@@ -129,20 +133,27 @@ class HillasReconstructor(Reconstructor):
         ------
         TooFewTelescopesException
             if len(hillas_dict) < 2
+        InvalidWidthException
+            if any width is np.nan or 0
         '''
 
         # filter warnings for missing obs time. this is needed because MC data has no obs time
         warnings.filterwarnings(action='ignore', category=MissingFrameAttributeWarning)
 
-        # stereoscopy needs at least two telescopes with valid widths
-        valid_telescopes = sum([1 if not np.isnan(hillas_dict[x]['width'].value)
-                                else 0
-                                for x in hillas_dict])
-
-        if valid_telescopes < 2:
+        # stereoscopy needs at least two telescopes
+        if len(hillas_dict) < 2:
             raise TooFewTelescopesException(
                 "need at least two telescopes, have {}"
-                .format(valid_telescopes))
+                .format(len(hillas_dict)))
+
+        # check for np.nan or 0 width's as these screw up weights
+        if any([np.isnan(hillas_dict[tel]['width'].value) for tel in hillas_dict]):
+            raise InvalidWidthException(
+                "A HillasContainer contains an ellipse of width==np.nan")
+
+        if any([hillas_dict[tel]['width'].value == 0 for tel in hillas_dict]):
+            raise InvalidWidthException(
+                "A HillasContainer contains an ellipse of width==0")
 
         self.initialize_hillas_planes(
             hillas_dict,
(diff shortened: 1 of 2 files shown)


vuillaut:
Hi.

Checking for valid telescopes is indeed a good approach. Crucial parameters can fail for many reasons.

I also remember suggesting in some discussion about Hillas parameters that the width should be equal to 0 and not `nan` in case of aligned pixels.

LukasNickel:
I think u refer to #772, right?

ctapipe/reco/HillasReconstructor.py
```diff
@@ -133,12 +133,16 @@ class are set to np.nan
 
         # filter warnings for missing obs time. this is needed because MC data has no obs time
         warnings.filterwarnings(action='ignore', category=MissingFrameAttributeWarning)
-        
-        # stereoscopy needs at least two telescopes
-        if len(hillas_dict) < 2:
+
+        # stereoscopy needs at least two telescopes with valid widths
+        valid_telescopes = sum([1 if not np.isnan(hillas_dict[x]['width'].value)
```
MaxNoe: I think you could directly build a new dict here with all the valid telescopes and only use those later.

E.g. `valid_hillas = {tel_id: hillas for tel_id, hillas in hillas_dict.items() if np.isfinite(hillas['width'].value}`

thomasgas:
Just a question: what happens in case of we have 3 telescopes triggered and one of them has a NaN width? is h_max well reconstructed in this case?
I would also add a comment above those lines saying something like "This is done in case the width is NaN and h_max cannot be reconstructed"...or something like this...so that in the future we forget why we put it there.

LukasNickel:
> Just a question: what happens in case of we have 3 telescopes triggered and one of them has a NaN width? is h_max well reconstructed in this case?
> I would also add a comment above those lines saying something like "This is done in case the width is NaN and h_max cannot be reconstructed"...or something like this...so that in the future we forget why we put it there.

Im not sure actually. If we only go with the valid telescopes as @MaxNoe suggested above we would obviously avoid that (potential) problem

ctapipe/reco/HillasReconstructor.py
LukasNickel: I think this would be the cleaner approach for now but it kind of comes back to #772.
Do we still use these images by adjusting the weights or do we not use them at all?

vuillaut: Exactly, thanks for digging it out.
My feeling is that 3 aligned pixels still give information about the direction and should not simply be thrown out - but I have not done a study on that ;-)
I see the problem with weights, however, and have no clear fix to handle this.

thomasgas:
> Im not sure actually. If we only go with the valid telescopes as @MaxNoe suggested above we would obviously avoid that (potential) problem

I agree we should treat those images in a certain way (and we can discuss about this).
I'm just wondering that happens to those passing this condition even though the width in one on them might be NaN. 
In any case I would approve this PR: if you could also have a look at what happens to the reconstruction if we use those images by setting the width to a certain small value (which of course should make sense) instead of NaN, it would be great!

LukasNickel:
> > Im not sure actually. If we only go with the valid telescopes as @MaxNoe suggested above we would obviously avoid that (potential) problem
> 
> I agree we should treat those images in a certain way (and we can discuss about this).
> I'm just wondering that happens to those passing this condition even though the width in one on them might be NaN.
> In any case I would approve this PR: if you could also have a look at what happens to the reconstruction if we use those images by setting the width to a certain small value (which of course should make sense) instead of NaN, it would be great!

I will try to find some time for this :)

ctapipe/reco/HillasReconstructor.py
kosack: yes, it's nice to keep track of what telescopes were dropped. Generally, i think a real pipeline would use a CutFlow to select telescopes before Hillas parameterization (e.g. with a minimum width and length or number of surviving pixels), but the code should still do reasonable things without that separate step, so testing for at least NaN is good. It should only skip telescopes if there is a real problem, and leave it up to the user to make any looser cuts for robustness.

kosack: and by real problem I mean e.g. a singular matrix. In that case you could still reconstruct a direction if possible, and set the h_max to NaN

kosack:
Reiterating my comments from above, we should avoid putting any "cuts" in the code, other than absolute failure cases (e.g. NaN or Inf preventing a sensible value).  The user should explicitly control all cuts before the hillas code is executed (e.g they might replace the telescope dict with a subset after rejecting telescopes for any reason)

kosack:
And we must make sure any ignored telescopes are recorded somehow, so we don't get the case of events that are marked as 3 tel in the output that only really used 2 telescopes (which will throw off any benchmarks we do)

MaxNoe:
> Reiterating my comments from above, we should avoid putting any "cuts" in the code, other than absolute failure cases (e.g. NaN or Inf preventing a sensible value). The user should explicitly control all cuts before the hillas code is executed (e.g they might replace the telescope dict with a subset after rejecting telescopes for any reason
)

That would mean to fail loudly, if any of the telescopes contain unusable data.
I like that a lot! This makes it much easier, to design systems, that keep track where and why telescope data was discarded.

So this means, the `Reconstructor` should fail early and load, if any of its input data is invalid.

LukasNickel:
@kpfrang 
Did you actually succeed with other weighting methods for the planes?

kosack:
> So this means, the Reconstructor should fail early and load, if any of its input data is invalid.

Yes, I think that's cleaner. I worry about adding any "hidden" cuts inside classes like Reconstructors, since it will clearly lead to some unexpected behavior later when people forget they are there (e.g. expecting the telescope multiplicity to be what the reconstructor used, when sometimes it is not).

mackaiver:
So this should fail as soon as there any nans in the hillas_dict. Makes sense to me. 👍

thomasgas:
I would like to solve this PR asap since I just run into this same problem today. 
I had 28 telescopes triggered in a huge event and with just one of them having a `width = 0` the reconstruction was giving `nan`. How to solve this:
- discard all telescopes with `width = 0` (should be done in the cleaning and not hard-coded then);
- if `width = 0` (which is unlikely with more than 3 pixels remaining...), use @dneise suggestion (https://github.com/cta-observatory/ctapipe/issues/772#issuecomment-410658085) to use a minimum value for `width` from the pixel geometry;
- discard all telescopes with less than 5 (or configurable number of) pixels remaining after cleaning

We leave this selection to the user (replace width or apply some cleaning) and decide that in the HillasReconstructor we keep only those telescopes not breaking the reconstruction? Of course if the reconstructor drops a telescope, this should be reported to the user (with explanations).

Since my problem is with `width = 0` and not `width = NaN`, I would add the `0`-case to the exclusion of a telescope from the reconstruction because asking for isnan doesn't solve my problems.

ctapipe/reco/HillasReconstructor.py
```diff
@@ -129,20 +133,27 @@ class are set to np.nan
...
-        if valid_telescopes < 2:
+        # stereoscopy needs at least two telescopes
+        if len(hillas_dict) < 2:
             raise TooFewTelescopesException(
                 "need at least two telescopes, have {}"
-                .format(valid_telescopes))
+                .format(len(hillas_dict)))
+
+        # check for np.nan or 0 width's as these screw up weights
+        if any([np.isnan(hillas_dict[tel]['width'].value) for tel in hillas_dict]):
```
MaxNoe: Better in one loop and use a generator expression?

```
any(np.isnan(h.width.value) or h.width.value == 0 for h in hillas_dict.values())
```

LukasNickel: Fine with me. I wanted to seperate the error messages but we can combine them if you feel like we dont gain anything with it

thomasgas: I would prefer to print two different messages in case of `w=0` or `w=NaN`.
Ok to use one generator expression per each case.

kpfrang:
> @kpfrang
> Did you actually succeed with other weighting methods for the planes?

What I did was investigating what we gain using weights from lookup tables for the distance of closest approach between the semi-major axis of the image and the true direction projected in the camera.

Besides improving the overall reconstruction, it also helped to save those images. For this investigation I always used `width = 0` for such images (back then this was not always consistent). The LUTs were able to deal with `0` as one of the parameters was the ratio `width / length`. If `width = NaN` however,  the LUTs obviously would also fail to come up with a weight for those images.

I did not dig too deep into this, but I didn't find a problem in the reconstruction when I kept those images.

thomasgas:
What about using those weights to try to fix the reconstruction for just 2 LSTs? Would it solve the issue that @jjlk showed in #967?

HealthyPear:
Hi all,

regarding this (still open) discussion, now that the protopipe code is out, you could give a look to event_preparer.py.

I couldn't yet elaborate on what Julien left me, but I believe that from line #133 of that file there is code very similar to what you suggest - perhaps it can be possible to get inspired from it.

PR #996, opened 2019-02-28
Refactor toymodel, add SkewedGaussian

MaxNoe:
* Uses units
* much more flexible and nicer to use imho
* Fix tests of muon fitter (wasn't testing anything really)
* Does not implement gaussian pdf ourselves


ctapipe/image/toymodel.py
```diff
@@ -8,95 +8,196 @@
...
+
+        """
+        self.x = x
+        self.y = y
+        self.width = width
+        self.length = length
+        self.psi = psi
+        self.skewness = skewness
+
+    def _momemts_to_parameters(self):
```
kosack: Typo: moments?

PR #1002, opened 2019-03-05
Make ctapipe-extra truly optional

MaxNoe:
ctapipe-extra was not listed in setup.py as required but was actually required in several places.

This should make it a truly optional dependency


kosack:
should it be optional? It defines a lot of required defaults. They can of course be overwritten by users using $CTAPIPE_SVC_PATH, but many things won't work without at least the defaults defined.  Of course, it will be refactored soon to be a bare minimum, but I think it's still required for normal operation.

MaxNoe:
> It defines a lot of required defaults.

It does? Where are those used?

Currently, `ctapipe_resources` is not mentionend in `install_requires`.

MaxNoe:
I tested several things now, e.g. doing hillas reconstructiion on mc.

Nothing needed ctapipe-extra.

I also grepped through the code for stuff like `get_dataset_path`, only the tests use it, or the tools for example data. 

I think we should go forward and merge this.

kosack:
I guess it will break ctapipe-camdemo and most of the tutorials however, so  it's still "somewhat" required. I'd prefer to keep it at least in the conda package as a dependency, and we could always rename ctapipe-extra to ctapipe-resources (which is the only package inside it so far, so would make more sense), and then the semantic problem is gone.

MaxNoe:
@kosack it already has.

dneise:
@kosack  this is the current setup.py in the master:

https://github.com/cta-observatory/ctapipe/blob/cc215e2f7233973ff1b604f90c96ac8cea447fcd/setup.py#L76-L80

MaxNoe:
Ok, I added a small fix for setup.py, please review again @kosack @dneise

PR #1008, opened 2019-03-14
Update requirements

MaxNoe:


kosack:
the build is failing only due to travis-sphinx (so unrelated).

Also, Is it normal to get so many `UnknownPixelShapeWarnings`?

MaxNoe:
> Also, Is it normal to get so many UnknownPixelShapeWarnings?

Yes, since a few tests still open the prod2 file, each time this file is opened, this warning is emitted, since prod2 file do not save the pixel shape.

kosack:
Ah, ok then we can just accept that until we update the tests to a more modern file

kosack:
For the build failure, see #1010

MaxNoe:
I fixed it by requiring nbconvert > 5.4 which works with tornado 6

MaxNoe:
Merge?

PR #1011, opened 2019-03-15
remove sst1m_event_source

dneise:
#981 was too messy. This PR simply removes two files in a single commit and thus is much cleaner.


MaxNoe:
Nice! Very easy to review/approve

PR #1012, opened 2019-03-15
Remove targetio event source and calibrator

MaxNoe:
Now here: https://github.com/cta-observatory/ctapipe_io_targetio


kosack:
I'll prepare a "before" release now.    Question: what should the new version be? I guess 0.7.0 for this one, since there was significant refactoring of Factories since 0.6.2, but that's somewhat internal, so perhaps 0.6.3 is still ok, and leave 0.7 for the plugin event sources?

MaxNoe:
I don't think it makes sense to do a release now. We are halfway through removing the event sources.

We already removed nectar, lst and sst1m. This is the last one.

So we should either make a release *before* we started to remove any event source, or after we removed all.

kosack:
Ok that's fine then - I think 0.6.2 is fine anyhow, and even though it has a few less event sources, most users will not notice.  I'll wait for the full change for 0.7 then. It's cleaner since its a full overhaul of the I/O.

bregeon:
Well, if I may releases are free, so if someone needs a release se should just do it. :grinning:

MaxNoe:
Eure, but it should make sense. So I would make a release for the last PR we merged before starting to remove the event sources.

Issue #963, opened 2019-02-11
Robust pixel neighbour calculation


watsonjj:
The pixel neighbour calculation may need revisiting. Currently it defines neighbours based on pixel separation. Right now, the defined separation seems to result in a different number of neighbours for different pixels on the CHEC-S camera. From examples/display_integrator.py:

![image](https://user-images.githubusercontent.com/17825673/52570941-e18f4f80-2e14-11e9-8e95-c71ebd9a7ff6.png)

I did not expect more than four neighbours for each pixel. Perhaps a more robust definition of a pixel neighbour should be implemented. Or perhaps it is fine to have N pixels within a separation to be defined as a neighbour. I will look into it.

watsonjj:
This occurs due to the curved focal plane of the camera, which results in a changing pixel seperation with distance from the camera centre. The current neighbour definition does not account for this.

MaxNoe:
I cannot reproduce this locally with the CHEC camera geometry that comes with ctapipe, can you provide a file where this happens?

MaxNoe:
![n_neightbors_chec](https://user-images.githubusercontent.com/5488440/54520430-fbb6d300-4968-11e9-8235-532105b284b1.png)

watsonjj:
Yeh sure. It occurs for the new CHEC geometry in Prod 4. An example simtel file can be obtained from this link:
https://www.dropbox.com/s/dlb05ntljpjhfil/run1.simtel.gz?dl=0

MaxNoe:
My robust method results in this for the CHEC camera:

![neighbors_chec](https://user-images.githubusercontent.com/5488440/54522463-95807f00-496d-11e9-9a91-8368e11968f4.png)

Is this wanted?

watsonjj:
No, usually we do count cross-module pixels as neighbours

MaxNoe:
If I increase the tolerance to 0.35 I get the "correct" neighbors.

It's important that this threshold is smaller than sqrt(2) to not count the diagonals as pixels.

On the master it was `0.4`, but only evaluated on the central pixel, not for each pixel. This is why it broke for curved sensor plane.

watsonjj:
I am happy to take a look at your alternative solution. I took the approach of defining the pixels in terms of row and columns, as that tends to be how we do this sort of thing in the CHEC group. The utility function for obtaining row and column could be useful to have (perhaps adapted to be more robust in a similar fashion as you have done for the neighbours).

[...]

PR #1014, opened 2019-03-15
Fix pixel neighbour definition for curved focal planes (with rectangular pixels)

watsonjj:
As demonstrated in #963, the current definition of pixel neighbours is not robust for cameras with curved focal planes. This is because pixel separation changes with distance from the camera centre.

As the CTA cameras with curved focal planes are also the cameras with pixels aligned on a rectangular grid, this PR will address this issue by defining the neighbours based on the row and columns of the pixels instead of the coordinates.

### DIFF
--- a/ctapipe/instrument/camera.py
+++ b/ctapipe/instrument/camera.py
@@ -686,19 +686,21 @@ def _rectangular_pixel_row_column(pix_x, pix_y):
 
     """
     # Estimate the maximum number of rows and columns of pixels
-    dist = _get_min_pixel_seperation(pix_x, pix_y) / 2
-    max_nrow = int(np.ceil((pix_y.max() - pix_y.min()) / dist))
-    max_ncol = int(np.ceil((pix_x.max() - pix_x.min()) / dist))
+    dist = _get_min_pixel_seperation(pix_x, pix_y)
+    max_nrow = int(np.ceil((pix_y.max() - pix_y.min()) / dist)) + 1
+    max_ncol = int(np.ceil((pix_x.max() - pix_x.min()) / dist)) + 1
 
     # Bin the pixel positions on a 2D grid
     hist_x = np.histogram2d(pix_x, pix_y, weights=pix_x, bins=[max_ncol, max_nrow])[0]
     hist_y = np.histogram2d(pix_x, pix_y, weights=pix_y, bins=[max_ncol, max_nrow])[0]
+    hist_xc = np.histogram2d(pix_x, pix_y, bins=[max_ncol, max_nrow])[0]
+    hist_yc = np.histogram2d(pix_x, pix_y, bins=[max_ncol, max_nrow])[0]
     # Find row and col with a complete number of pixels along them
     full_col = np.bincount(hist_x.nonzero()[0]).argmax()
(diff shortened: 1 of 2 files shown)


MaxNoe:
I don't understand why the histogram approach does not exhibit the same problem as the radius approach. Both should suffer in the same way from changing pixel separations on a curved plain.

Maybe it's time to generalize CameraGeometry, maybe even subclass it to a FlatCameraGeometry and a CurvedCameraGeometry.

All the information necessary is contained in the simtel files and is available in the SimTelFile of the SimTelEventSource.

watsonjj:
> I don't understand why the histogram approach does not exhibit the same problem as the radius approach. Both should suffer in the same way from changing pixel separations on a curved plain.

It absolutely could for extreme curvatures, but there is more leeway with the histogram binnings, whereas the neighbour definition defines a single threshold.

> Maybe it's time to generalize CameraGeometry, maybe even subclass it to a FlatCameraGeometry and a CurvedCameraGeometry.

I think this could be nice.

This was more intended as a quick, simple fix for a problem I identified, without too much overhaul of the geometry class. An improvement on the current way it's done.

MaxNoe:
I think I found a better solutiion using the kdtree, that is even more general.

I'll open another PR.

MaxNoe:
Can you explain the approach here? Why is it necessary to call histogram2d multiple times with weights and so on?

watsonjj:
I bin the pixel positions onto a square grid. The bins of the square grid needs to be small enough so each pixel has its own bin, but not too small such that the change of the y (or x) pixel coordinate across a single row (or column) due to the focal plane curvature does not result in mistakingly assigning the pixel to a different row (or column).

Because of the module gaps, this will result in some empty bins, which should not be counted as its own row. This is the reason for the second pair of calls for histogram2d.

This part could probably be done better. The aim of this segment of code is to obtain an array of x coordinates for a row that fully spans the camera, and the an array of y coordinates for a column that fully spans the camera.

MaxNoe:
See my other approach in #1015

kosack:
Doesn't this mean perhaps we should store the curved pixel positions in another coordinate system, so that their distances are not changing?

kosack:
for instance the angular distance between pixels I guess doesn't change? or is it just the inter-module distances that change, and each module is a plane? I don't know enough about how the focal plane on those cameras is physically implemented.

MaxNoe:
> Probably we need better documentation somewhere describing the curved planes

simtel files contain u v information for those pixels. I think Everythong we need to know is there.

kosack:
> contain z and angle

That's good to know - I though it was only z, which isn't enough.  However that means they are not stored as I thought they were (e.g. X-Y position on the focal plane, which might be curved), rather they are X-Y in real space, not accounting for the curvature.  To be certain, we should look in read_cta in simtelarray to see how the analysis is done there - I would have assumed they were in a system that makes the analysis the same for curved and flat planes, but if all coordinates are given, that may not be the case and a correction might be necessary.

watsonjj:
I will close this PR as #1015 was merged.

Discussion of curved focal plane handling has continued in #1018.

PR #1015, opened 2019-03-18
Fix neighbors

MaxNoe:
Alternative approach to #1014 

By using the L1 norm for rectangular grids, this works like before, but much better for square grids.

I also changed the hierarchy of `neighbors`, `neighbor_matrix`, and `neighbor_matrix_sparse`.

The actual data is now in `neighbor_matrix_sparse` and the rest is computed from that.
chec

### DIFF
--- a/ctapipe/image/geometry_converter_hex.py
+++ b/ctapipe/image/geometry_converter_hex.py
@@ -397,8 +397,8 @@ def convert_geometry_hex1d_to_rect2d(geom, signal, key=None, add_rot=0):
         new_geom = CameraGeometry(
             cam_id=geom.cam_id + "_rect",
             pix_id=ids,  # this is a list of all the valid coordinate pairs now
-            pix_x=grid_x * u.meter,
-            pix_y=grid_y * u.meter,
+            pix_x=u.Quantity(grid_x.ravel(),  u.meter),
+            pix_y=u.Quantity(grid_y.ravel(),  u.meter),
             pix_area=pix_area * u.meter ** 2,
             neighbors=geom.neighbors,
             pix_type='rectangular', apply_derotation=False)
--- a/ctapipe/instrument/camera.py
+++ b/ctapipe/instrument/camera.py
@@ -337,27 +337,30 @@ class CameraGeometry:
         if self.pix_type.startswith('hex'):
             k = 7
             radius = 1.1
+            norm = 2  # use L2 norm for hex
         else:
+            norm = 1  # use L1 norm for rectangular
+
             # if diagonal should count as neighbor, we
             # need to find at most 8 neighbors with a max distance
             # < than 2 * the pixel size, else 4 neigbors with max distance
             # < sqrt(2) pixel size
             if diagonal:
                 k = 9
-                radius = 2
+                radius = 2.95
             else:
                 k = 5
-                radius = np.sqrt(2)
+                radius = 1.95
 
         for i, pixel in enumerate(self._kdtree.data):
-            d, n = self._kdtree.query(pixel, k=k)
+            d, n = self._kdtree.query(pixel, k=k, p=norm)
 
             # remove self-reference
             d = d[1:]
             n = n[1:]
 
             # remove too far away pixels
-            mask = d < (0.98 * radius * np.min(d))
+            mask = d < radius * np.min(d)
             neighbors[i, n[mask]] = True
 
         return neighbors.tocsr()
--- a/ctapipe/instrument/camera.py
+++ b/ctapipe/instrument/camera.py
@@ -318,7 +318,7 @@ class CameraGeometry:
     @lazyproperty
     def neighbors(self):
         '''A list of the neighbors pixel_ids for each pixel'''
-        return [np.where(r)[0] for r in self.neighbor_matrix]
+        return [np.where(r)[0].tolist() for r in self.neighbor_matrix]
 
     @lazyproperty
     def neighbor_matrix(self):
--- a/ctapipe/instrument/camera.py
+++ b/ctapipe/instrument/camera.py
@@ -10,6 +10,7 @@ from astropy.coordinates import Angle
 from astropy.table import Table
 from astropy.utils import lazyproperty
 from scipy.spatial import cKDTree as KDTree
+from scipy.sparse import lil_matrix, csr_matrix
 import warnings
 
 from ctapipe.utils import get_table_dataset, find_all_matching_datasets
--- a/ctapipe/instrument/camera.py
+++ b/ctapipe/instrument/camera.py
@@ -345,9 +345,9 @@ class CameraGeometry:
             norm = 1  # use L1 norm for rectangular
 
             # if diagonal should count as neighbor, we
-            # need to find at most 8 neighbors with a max distance
-            # < than 2 * the pixel size, else 4 neigbors with max distance
-            # < sqrt(2) pixel size
+            # need to find at most 8 neighbors with a max L1 distance
+            # < than 3 * the pixel size, else 4 neigbors with max distance
+            # < 2 pixel size
             if diagonal:
                 k = 9
                 radius = 2.95
--- a/ctapipe/instrument/camera.py
+++ b/ctapipe/instrument/camera.py
@@ -10,7 +10,7 @@ from astropy.coordinates import Angle
 from astropy.table import Table
 from astropy.utils import lazyproperty
 from scipy.spatial import cKDTree as KDTree
-from scipy.sparse import csr_matrix
+from scipy.sparse import lil_matrix, csr_matrix
 
 from ctapipe.utils import get_table_dataset, find_all_matching_datasets
 from ctapipe.utils.linalg import rotation_matrix_2d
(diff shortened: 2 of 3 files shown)


MaxNoe:
I fixed the comment.

ctapipe/image/geometry_converter_hex.py
```diff
@@ -397,8 +397,8 @@ def convert_geometry_hex1d_to_rect2d(geom, signal, key=None, add_rot=0):
         new_geom = CameraGeometry(
             cam_id=geom.cam_id + "_rect",
             pix_id=ids,  # this is a list of all the valid coordinate pairs now
-            pix_x=grid_x * u.meter,
-            pix_y=grid_y * u.meter,
+            pix_x=u.Quantity(grid_x.ravel(),  u.meter),
+            pix_y=u.Quantity(grid_y.ravel(),  u.meter),
```
kosack: will this break any of the code from the machine-learning studies who later write this out as a 2D image? Perhaps we should at least store the original dimensions as metadata in the case the original  geometry was 2D. (I agree we shouldn't normally support 2D though, but having the metadata would allow one to `reshape` back to 2D later)

ctapipe/instrument/camera.py
```diff
@@ -145,7 +157,7 @@ def _calc_pixel_area(pix_x, pix_y, pix_type):
         Note this will not work on cameras with varying pixel sizes.
```
kosack: I just noticed we never check for this case where it doesn't work... could be good to raise an exception or warning. The Whipple490 camera can be used to test that it fails (it has mixed pixel radii)

ctapipe/instrument/camera.py
```diff
@@ -308,32 +320,53 @@ def __str__(self):
...
-        dist = _get_min_pixel_seperation(self.pix_x, self.pix_y)
-
-        neighbors = _find_neighbor_pixels(
-            self.pix_x.value,
-            self.pix_y.value,
-            rad=1.4 * dist.value
-        )
-
-        return neighbors
+        '''A list of the neighbors pixel_ids for each pixel'''
```
kosack: use # for comments, not string

ctapipe/instrument/tests/test_camera.py
```diff
@@ -87,7 +91,10 @@ def test_neighbor_pixels(cam_id):
         assert n_neighbors.count(5) == 0
         assert n_neighbors.count(6) == 0
 
-    assert n_neighbors.count(1) == 0  # no pixel should have a single neighbor
+    # whipple has inhomogenious pixels that mess with pixel neighborhood
+    # calculation
+    if cam_id != 'Whipple490':
```
kosack: here can use an `AssertRaises` if you implement the exception as described above.  Then you can test if a slice of the camera still works (e.g. `geom[0:378]`)

ctapipe/image/geometry_converter_hex.py
kosack: to be more clear, the main use case of the GeometryConverter code is to produce 2D images that can be fed to image processing routines, so we need to support that somehow.

ctapipe/instrument/camera.py
MaxNoe: This is a docstring?

ctapipe/instrument/tests/test_camera.py
MaxNoe: The neighbors it calculates are not completly wrong. Just the constrained that no pixel should have just one neighbour is not met. So Maybe don't throw an exception.

MaxNoe: See here:

whipple

ctapipe/image/geometry_converter_hex.py
MaxNoe: > will this break any of the code from the machine-learning studies who later write this out as a 2D image?

I have no clue. All the tests pass. I did not change any test code concerning the geometry converters.

MaxNoe: Also, I think most of the behaviour was broken before.

If there was a camera roation set, but pxiel shape was not 1d, it was silently ignored, no warning, no error.

ctapipe/instrument/camera.py
MaxNoe: You cannot really throw an error here, because this function is only called, when the pixel area is unknown.

Maybe we should just remove it, I think it's not used anymore. Pixel area is known from the simtel files.

MaxNoe:
I just found that the neighbor matrix is not symmetrical. I will investigate this.

watsonjj:
So why these radius values? What was wrong with them before?

MaxNoe:
> So why these radius values? What was wrong with them before?

They make no difference for the regular grids and look better for the irregular. 
They are closer to the maximum radius where you would start to get the second ring of hexagons.

(1.4...) (not sqrt(2)).

ctapipe/instrument/camera.py
```diff
@@ -308,32 +321,74 @@ def __str__(self):
...
+                k = 9
+                norm = 2
+                radius = 1.95
+            else:
+                k = 5
+                radius = 1.5
+                norm = 1
+
+        for i, pixel in enumerate(self._kdtree.data):
+            d, n = self._kdtree.query(pixel, k=k, p=norm)
```
dneise: kdtree.query() has a parameter called `distance_upper_bound`. Could this be used to get rid of the whole `mask` issue in line 378? 
https://docs.scipy.org/doc/scipy-0.14.0/reference/generated/scipy.spatial.KDTree.query.html

ctapipe/instrument/camera.py
```diff
@@ -308,32 +321,74 @@ def __str__(self):
...
+            # remove too far away pixels
+            mask = d < radius * np.min(d)
+            neighbors[i, n[mask]] = True
+
+        # filter annoying deprecation warning from within scipy
+        # scipy still uses np.matrix in scipy.sparse, but we do not
+        # explicitly use any feature of np.matrix, so we can ignore this here
+        with warnings.catch_warnings():
+            warnings.filterwarnings('ignore', category=PendingDeprecationWarning)
+            if (neighbors.T != neighbors).sum() > 0:
```
dneise: Isn't this `.sum() > 0` actually an `.any()` ? In english this would then read roughly like: 

if the transposed neighbors is not equal to neighbors, anywhere, then...

ctapipe/instrument/camera.py
watsonjj: I wondered this, but then you lose the per-pixel distance_upper_bound setting. I think this is why Max has done it this way.

ctapipe/instrument/camera.py
```diff
@@ -308,32 +321,74 @@ def __str__(self):
...
+        Parameters
+        ----------
+        diagonal: bool
+            If rectangular geometry, also add diagonal neighbors
+        '''
+        neighbors = lil_matrix((self.n_pixels, self.n_pixels), dtype=bool)
+
+        if self.pix_type.startswith('hex'):
+            k = 7
+            radius = 1.4
```
dneise: Ah sorry one more thing. I believe these 3 `radius` are somehow magic numbers. Maybe it would be good to shortly explain their choice in a comment. Maybe each `radius` needs and individual comment maybe not? 

I think you already commented on them in this discussion, but this will be lost after the PR is merged.

ctapipe/instrument/camera.py
watsonjj: More specifically - the distance set as an upper bound is based on the neighbours he obtains for that pixel, so it can't be specified for the `distance_upper_bound` argument priorly.

MaxNoe: Exactly. The maximum distance is only known after the query is done.

ctapipe/instrument/camera.py
MaxNoe: Sparse matrices do not have any, unfortunately.

dneise: ok  ... sorry .. should have checked that myself.

ctapipe/instrument/camera.py
```diff
@@ -79,17 +83,26 @@ def __init__(self, cam_id, pix_id, pix_x, pix_y, pix_area, pix_type,
...
         self.pix_rotation = Angle(pix_rotation)
         self.cam_rotation = Angle(cam_rotation)
-        self._precalculated_neighbors = neighbors
+        self._neighbors = neighbors
+
+        if neighbors is not None:
+            if isinstance(neighbors, list):
+                lil = lil_matrix((self.n_pixels, self.n_pixels), dtype=bool)
+                for pix_id, neighbors in enumerate(neighbors):
+                    lil[pix_id, neighbors] = True
```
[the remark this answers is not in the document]
MaxNoe: TypeError: 'coo_matrix' object does not support item assignment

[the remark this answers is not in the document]
MaxNoe: And lil is faster than csr

ctapipe/instrument/camera.py
MaxNoe: No, this is a list of neighbor ids for each pixel (np.where on the flags)

ctapipe/instrument/camera.py
```diff
@@ -308,32 +321,74 @@ def __str__(self):
...
+            norm = 2  # use L2 norm for hex
+        else:
+
+            # if diagonal should count as neighbor, we
+            # need to find at most 8 neighbors with a max L2 distance
+            # < than 2 * the pixel size, else 4 neigbors with max L1 distance
+            # < 2 pixel size
+
+            if diagonal:
+                k = 9
```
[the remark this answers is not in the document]
MaxNoe: See updated code.

ctapipe/instrument/camera.py
MaxNoe: done

ctapipe/instrument/camera.py
```diff
@@ -308,32 +321,74 @@ def __str__(self):
...
+
+        for i, pixel in enumerate(self._kdtree.data):
+            d, n = self._kdtree.query(pixel, k=k, p=norm)
+
+            # remove self-reference
+            d = d[1:]
+            n = n[1:]
+
+            # remove too far away pixels
+            mask = d < radius * np.min(d)
```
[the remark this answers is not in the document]
MaxNoe: This variable is used only in the line below.

ctapipe/image/geometry_converter_hex.py
kosack: >I have no clue. All the tests pass. I did not change any test code concerning the geometry converters.
I guess we need to look at the other projects that use this and see what test should be added...

ctapipe/instrument/camera.py
kosack: I think it was just for some of the camera definitions I collected from other telescopes where the area wasn't specified, but then it was calculated and written back out to FITS files, so perhaps it's not needed anymore.

ctapipe/instrument/tests/test_camera.py
kosack: It works better than I thought. I never really checked with that camera.  I guess we can just keep it that way then until it becomes a real problem. Probably this will rarely if never be an issue anyhow

ctapipe/image/geometry_converter_hex.py
MaxNoe: Or just wait until people expecting different behaviour open issues or PRs.

dneise: I think we cannot start guessing what people might want.

Issue #1013, opened 2019-03-15
Problem with "parent" argument in eventsource


FrancaCassol:
Hi, 
when I call the ctapipe_io_lst source from a tool:
`flatfield_tool.run(argv=['--config','/astro/users/cassol/soft/python/cta-lstchain/lstchain/tools/flatfield_param.json'])`

I get an error concerning the "parent" and the "config" arguments. 

```
/scratch4/franca/soft/ctapipe/ctapipe/io/eventsource.py
 in __init__(self, config, parent, **kwargs)
    142         kwargs
    143         """
--> 144         super().__init__(config=config, parent=parent, **kwargs)
    145 
    146         self.metadata = dict(is_simulation=False)

/scratch4/franca/soft/ctapipe/ctapipe/core/component.py in __init__(self, config, parent, **kwargs)
     95         """
     96         if parent is not None and config is not None:
---> 97             raise ValueError('Only one of `config` or `parent` allowed')
     98         super().__init__(parent=parent, config=config, **kwargs)
     99 
ValueError: Only one of `config` or `parent` allowed
```

The Component argument description seems outdated (there is "tool"  not   "parent"). 
Can perhaps somebody help?
Thanks

MaxNoe:
Hi Franca, I already opened a PR in the LST eventsource to fix this: https://github.com/cta-observatory/ctapipe_io_lst/pull/6

You maybe need to click the "watch" button in this repository, so you receive notifications.

kosack:
Perhaps we should rights to merge things in that repo to Franca (or whoever is maintaining it)?

kosack:
In any case, I just merged it

MaxNoe:
> Perhaps we should rights to merge things in that repo to Franca (or whoever is maintaining it)?

I gave her Admin permissions a few days ago

FrancaCassol:
Thanks Karl,
unfortunately with my tool I continue to get the error : 

ValueError: Only one of `config` or `parent` allowed

```
flatfield_tool.run(argv=['--config','/astro/users/cassol/soft/python/cta-lstchain/lstchain/tools/flatfield_param.json'])
```
```
---------------------------------------------------------------------------
ValueError                                Traceback (most recent call last)
<ipython-input-3-0f1913781b4d> in <module>()
      1 #
----> 2 flatfield_tool.run(argv=['--config','/astro/users/cassol/soft/python/cta-lstchain/lstchain/tools/flatfield_param.json'])

[...]

PR #1016, opened 2019-03-18
Fix component docs

MaxNoe:
See #1013


MaxNoe:
Codacy complains about 0% diff hit, I only changed comments ....

Issue #851, opened 2018-11-29
define how to handle things that need a configuration per-telecope in a standard way


kosack:
Many algorithms have telescope or camera-specific parameters that need to be tuned. Right now our config system doesn't handle that well, and some people have hard-coded scripts that read YAML files, tables, etc.

Let's pick a standard way. 

It should support:
* setting parameters for all cameras of the same type/version
* setting parameters for a *specific* camera by `tel_id`
* setting parameters for a group of telescopes specifid by tel_id. 
* May be useful to have some execution ordering to the configuration so one can e.g. first set defaults for all telescopes, change a few parameters for a group of telescopes(e.g. by type), and then override a few parameters for specific telescopes.

watsonjj:
> May be useful to have some execution ordering to the configuration so one can e.g. first set defaults for all telescopes, change a few parameters for a group of telescopes(e.g. by type), and then override a few parameters for specific telescopes.

I think a default configuration is correct, which might specify some things per type, and others per telescope. But I think the user should then be allowed to override the defaults in any way they wish. I.e. if the user configures all telescope types in some way, this also overrides the default "per telescope" configuration for that telescope type.

This should be very simple to implement, and most intuitive. The configurations are just applied in the order they are specified. E.g.
```python
config = {**default_config, **user_config}
```

watsonjj:
Okay, actually probably more complicated than that, as you will need some way of overwriting the "per telescope" configuration with the "per telescope type" configuration... but I think this entire topic is a step in the right direction

kosack:
see some discussion in #1022

kosack:
Just as another use-case, here is an example of what is used in Julien's pipeline for configuring 2 cleaning types (tailcuts + wavelet) as a yaml file.  
```yaml
cleaning_reco:  # note that there are more than one image cleaner defined, this is just the reco one
  tail:
   thresholds:  # picture, boundary
    - LSTCam: [6, 3]
    - NectarCam: [6, 3]
   keep_isolated_pixels: False
   min_number_picture_neighbors: 2

  wave:
   # Directory to write temporary files
   #tmp_files_directory: '/dev/shm/'
   tmp_files_directory: './'
   options:
    LSTCam:
     type_of_filtering: 'hard_filtering'
     filter_thresholds: [3, 0.2]
     last_scale_treatment: 'drop'
     kill_isolated_pixels: True
     detect_only_positive_structures: False
     clusters_threshold: 0
[...]

PR #1022, opened 2019-03-21
Add a component to make cleaning configurable

MaxNoe:
In many examples and pipelines, one can see code like this:

```
cleaning_levels = {
    'LSTCam': (5, 3, 2),
    'CHEC': (10, 4, 2),
}

levels = cleaning_levels[camera.cam_id]
tailcuts_clean(camera, image, picture_thresh=levels[0], boundary_threshold=levels[1])
```

I tried to make a Component that should make this configurable via traitlets

### DIFF
--- a/ctapipe/image/cleaning.py
+++ b/ctapipe/image/cleaning.py
@@ -2,16 +2,65 @@
 Image Cleaning Algorithms (identification of noisy pixels)
 """
 
-__all__ = ['tailcuts_clean', 'dilate']
+__all__ = ['tailcuts_clean', 'dilate', 'fact_image_cleaning', 'ImageCleaning']
 
 import numpy as np
 from scipy.sparse.csgraph import connected_components
+from ctapipe.core import Component
+from traitlets import Dict, CaselessStrEnum
 
 
-def tailcuts_clean(geom, image, picture_thresh=7, boundary_thresh=5,
-                   keep_isolated_pixels=False,
-                   min_number_picture_neighbors=0):
+class ImageCleaning(Component):
+    '''
+    A component to make image cleaning configurable using traitlets.
+    '''
 
+    key_type = CaselessStrEnum(
+        default_value='camera_id',
+        values=['camera_id', 'telescope_id'],
+        help='If the keys in settings are telescope ids or camera ids'
+    ).tag(config=True)
+    settings = Dict(
+        help='A dict mapping either camera_id or telescope_id to cleaning settings.'
+        ' This are keyword arguments for the actual cleaning function plus the'
+        ' the key algorithm which can be either "tailcuts" or "fact"'
+    ).tag(config=True)
+
+    def __call__(self, geom, image, arrival_times=None, telescope_id=None):
+
+        if self.key_type == 'camera_id':
+            kwargs = self.settings.get(geom.cam_id, {})
+        elif self.key_type == 'telescope_id':
+            if telescope_id is None:
+                raise ValueError(
+                    'telescope_id is required if using key_type="telescope_id"'
+                )
+            kwargs = self.settings.get(telescope_id, {})
+
+        algorithm = kwargs.pop('algorithm', 'tailcuts')
+
+        if algorithm == 'tailcuts':
+            return tailcuts_clean(geom, image, **kwargs)
+
+        elif algorithm == 'fact':
+            if arrival_times is None:
+                raise ValueError(
+                    'If using algorithm="fact", arrival_times must not be None'
+                )
+            return fact_image_cleaning(geom, image, arrival_times, **kwargs)
+
+        else:
+            raise ValueError(f'Unknown algorithm {algorithm}')
+
+
+def tailcuts_clean(
+    geom,
+    image,
+    picture_thresh=7,
+    boundary_thresh=5,
+    keep_isolated_pixels=False,
+    min_number_picture_neighbors=0,
+):
     """Clean an image by selection pixels that pass a two-threshold
     tail-cuts procedure.  The picture and boundary thresholds are
     defined with respect to the pedestal dispersion. All pixels that
--- a/ctapipe/image/tests/test_cleaning.py
+++ b/ctapipe/image/tests/test_cleaning.py
@@ -2,6 +2,7 @@ import numpy as np
 from numpy.testing import assert_allclose
 from ctapipe.image import cleaning
 from ctapipe.instrument import CameraGeometry
+import pytest
 
 
 def test_tailcuts_clean_simple():
@@ -43,6 +44,58 @@ def test_dilate():
     # dilate a third row
     dmask = cleaning.dilate(geom, dmask)
     assert dmask.sum() == 1 + 6 + 12 + 18
+
+
+def test_component():
+    from ctapipe.image import ImageCleaning, tailcuts_clean, fact_image_cleaning
+
+    geom = CameraGeometry.from_name('LSTCam')
+    image = np.random.poisson(12, geom.n_pixels)
+    arrival_times = np.random.uniform(0, 30, geom.n_pixels)
+
+    # test default is tailcuts with defaults
+    cleaning = ImageCleaning()
+    comp_mask = cleaning(geom, image)
+    tail_mask = tailcuts_clean(geom, image)
+    assert np.all(comp_mask == tail_mask)
+
+    # test different thresholds
+    cleaning = ImageCleaning(settings={
+        'LSTCam': {'picture_thresh': 10, 'boundary_thresh': 5}
+    })
+    comp_mask = cleaning(geom, image)
+    tail_mask = tailcuts_clean(geom, image, picture_thresh=10, boundary_thresh=5)
+    assert np.all(comp_mask == tail_mask)
+
+    # test different algorithm
+
+    cleaning = ImageCleaning(settings={
+        'LSTCam': {'algorithm': 'fact', 'picture_threshold': 10, 'boundary_threshold': 5}
+    })
+    comp_mask = cleaning(geom, image, arrival_times)
+    fact_mask = fact_image_cleaning(
+        geom, image, arrival_times, picture_threshold=10, boundary_threshold=5
+    )
+    assert np.all(comp_mask == fact_mask)
+
+    # test unknown algorithm
+    with pytest.raises(ValueError):
+        # test different algorithm
+        cleaning = ImageCleaning(settings={
+            'LSTCam': {'algorithm': 'awesome_cleaning'}
+        })
+        comp_mask = cleaning(geom, image)
+
+    # test by telescope_id
+    cleaning = ImageCleaning(
+        key_type='telescope_id',
+        settings={
+            1: {'picture_thresh': 10, 'boundary_thresh': 5}
+        },
+    )
+    comp_mask = cleaning(geom, image, telescope_id=1)
+    tail_mask = tailcuts_clean(geom, image, picture_thresh=10, boundary_thresh=5)
+    assert np.all(comp_mask == tail_mask)
 
 
 def test_tailcuts_clean():
--- a/ctapipe/image/__init__.py
+++ b/ctapipe/image/__init__.py
@@ -1,5 +1,14 @@
-from .hillas import *
-from .cleaning import *
+from .hillas import (
+    hillas_parameters,
+    HillasParameterizationError,
+    camera_to_shower_coordinates,
+)
+from .cleaning import (
+    tailcuts_clean,
+    ImageCleaning,
+    dilate,
+    fact_image_cleaning,
+)
 from .pixel_likelihood import *
 from .charge_extractors import *
 from .waveform_cleaning import *
@@ -8,3 +17,14 @@ from .muon import *
 from .geometry_converter import *
 from .leakage import *
 from .concentration import concentration
+
+
+__all__ = [
+    'HillasParameterizationError',
+    'hillas_parameters',
+    'camera_to_shower_coordinates',
+    'tailcuts_clean',
+    'fact_image_cleaning',
+    'dilate',
+    'ImageCleaning',
+]


watsonjj:
I think its a nice starting point, but in the end we have many per-telescope-type configuration parameters we need to pass to the pipeline, not just the tailcuts. For example, the charge extraction algorithm to use.

It would be nicer to have a single class to handle the per-telescope configuration for all Components. It could contain the defaults for each telescope type, and the ability to read a config file within which one can specify the configuration per telescope type.

ctapipe/image/cleaning.py
```diff
@@ -2,16 +2,65 @@
...
+        values=['camera_id', 'telescope_id'],
+        help='If the keys in settings are telescope ids or camera ids'
+    ).tag(config=True)
+    settings = Dict(
+        help='A dict mapping either camera_id or telescope_id to cleaning settings.'
+        ' This are keyword arguments for the actual cleaning function plus the'
+        ' the key algorithm which can be either "tailcuts" or "fact"'
+    ).tag(config=True)
+
+    def __call__(self, geom, image, arrival_times=None, telescope_id=None):
```
watsonjj: Could you pass the event container to this higher-level class? Then you can obtain the arrival times as you wish (and other cleaning methods might like to get other items from the container)

ctapipe/image/cleaning.py
```diff
@@ -2,16 +2,65 @@
...
+        if self.key_type == 'camera_id':
+            kwargs = self.settings.get(geom.cam_id, {})
+        elif self.key_type == 'telescope_id':
+            if telescope_id is None:
+                raise ValueError(
+                    'telescope_id is required if using key_type="telescope_id"'
+                )
+            kwargs = self.settings.get(telescope_id, {})
+
+        algorithm = kwargs.pop('algorithm', 'tailcuts')
```
watsonjj: Why is this not an argument to the __init__? Its a bit hidden like this

ctapipe/image/cleaning.py
```diff
@@ -2,16 +2,65 @@
...
+
+        if algorithm == 'tailcuts':
```
watsonjj: We adopt a factory approach for other algorithms, wouldn't it make sense to keep the same design across the pipeline? By having a Tailcut and FactImageCleaning subclass to the ImageCleaning baseclass? 

I can understand why you would do it like this (the functions exist, the classes don't, they are simple methods, and theres only two of them at the moment) but I'm just suggesting this from a consistency point of view.

ctapipe/image/cleaning.py
watsonjj: We haven't used `__call__` for other classes in the pipeline. Should we be consistent in some way? Either changing them to use `__call__`, or this one to use some member function.

dneise: I think we probably should have used `__call__` in other places, c.f. #1024

ctapipe/image/cleaning.py
dneise: I am not sure I understand Jasons point here correctly or if I have another point. I would propose to refrain from inventing new names for things (even if they are shorter and nicer) and thus avoid mental mapping, i.e. when the user wants the function `fact_image_cleaning` to be called, they should in my opinion have to write this:

    settings={
        'LSTCam': {'algorithm': 'fact_image_cleaning', 'picture_threshold': 10, 'boundary_threshold': 5}
    }

and not the shorter form:

    settings={
        'LSTCam': {'algorithm': 'fact', 'picture_threshold': 10, 'boundary_threshold': 5}
    }

Another reason for using this longer form would be: Assume you are a young bachelor starting here, and somebody gives you an already working analysis with such a config file as a starting point for your work. If you are interested in understanding the cleaning, it would be pretty cool if you could just search the whole codebase for the name after "algorithm" and actually have a single hit which happens to be the exact function you are looking for...

ctapipe/image/tests/test_cleaning.py
```diff
@@ -45,6 +46,58 @@ def test_dilate():
...
 
+def test_component():
+    from ctapipe.image import ImageCleaning, tailcuts_clean, fact_image_cleaning
+
+    geom = CameraGeometry.from_name('LSTCam')
+    image = np.random.poisson(12, geom.n_pixels)
+    arrival_times = np.random.uniform(0, 30, geom.n_pixels)
+
+    # test default is tailcuts with defaults
+    cleaning = ImageCleaning()
```
dneise: Minor point: I think it would be nice to use a verb form for the instance name here, instead of a gerund, since the instance is supposed to feel like a simple function and we (try to) use verbs for functions...

ctapipe/image/cleaning.py
watsonjj: I agree with @dneise on his point.

My point was that this `if... else...` clause is very similar use case to the factory approach we use for other components like the `ChargeExtractor`. It would be good to stay consistent with design across ctapipe.

watsonjj: I must say, I have been very inspired by a talk on KM3Pipe at PyGamma19 this week. Their pipeline is kept quite simple, and I like their configuration approach for their "modules". I include the link here for those who are curious: https://indico.cern.ch/event/783425/contributions/3328264/

MaxNoe: Yes, we met tamas at last years Asterics Workshop. 

We briefly talked about the possibility of just using km3pipe and writing cta modules.
I like this design a lot.

kosack:
There was a previous implementation of this (not in ctapipe, but in the external pipeline code), but I like this approach better.  

We should think to the future when all configurable objects will be specified in a YAML file, so something like this could be configured like:
```yaml
  GeometryImageCleaner:  
    _default_:
      algorithm: 'TailCutsImageCleaner'
      picture_threshold: 10.0
      boundary_threshold: 5.0
    ASTRICam:
      algorithm: 'WaveletImageCleaner'
      noise_threshold: 3.0

   EnergyImageCleaner:
       ...
```

Remember also that some image cleaning functions needs some extra input, like the time gradients, etc.

kosack:
By the way this pattern (of passing in a dict of type and parameters) is something we can apply to all per-telescope algorithms.

ctapipe/image/cleaning.py
kosack: This was a long conversation in the past, and we ended up keeping the explicit functions (for semantic reasons), but I think having a uniform interface is fine, then algorithms that are only functions can be passed around like algorithms that are configurable Components.

kosack:
Another comment: the ability to set a default in the config is quite useful (see my example above), especially if we have to do per-tel_id configurations, or have lots of telescope types. Then one only has to configure the special cases. 

A future development would be to allow one to specify configuratoin in more complex ways such as ranges of tel_ids. 

All this points to the need for a parent class that handles per-telescope configuration for all per-telescope algorithms, so perhaps we could take out some of this code from this ImageCleaner class, and make a `PerTelescopeComponent` class, such that we can do the configuration in common.  e.g. 

```py3
class ImageCleaner(PerTelescopeComponent):
     ...

class ChargeExtractor(PerTelescopeComponent):
    ...
class GainSelector(PerTelescopeComponent):
```

dneise:
I like this PR a lot, it reminded me of an idea I had some time ago ... sorry for the long wall of text, but I'd like to hear your opinion on that.

First some "observations"

1.)  [we put algorithms into functions if possible ✓](https://cta-observatory.github.io/ctapipe/development/code-guidelines.html#writing-algorithms)
2.) we use `Components` to configure these functions, this PR is an example for such a component ✓ 

So in a sense, the "meat" of ctapipe is not the components but the functions, the components are just "necessary evil" which is needed to get things nicely working together, using config, logging, provenance and such. 

Now the "idea" or question or whatever...:

Wouldn't it be nice if the "bloat ware" (Components) would be as small as possible? I was wondering if we could maybe (certainly not today but some time later) avoid writing components like `ImageCleaning` but instead of actually writing them, just having them constructed from some kind of "description" similar to `collections.namedtuple`. 

So in my dream we would say something like this:

```python
ImageCleaning = make_configurable_component_from_functions(
    options=[tailcuts_clean, fact_image_cleaning]
)
```

And that's all, `make_configurable_component_from_functions()` would basically *inspect* the functions for their common interface (in this case `geom, image` is common) and the rest would be pushed into `kwargs` just like you did in this PR.

Now... I know ... writing such a `make_configurable_component_from_functions` might be a waste of time if in the end we only create two or three Components from a list of functions. So I **do not propose to write such a thing today**, but maybe we want to keep this in the back of our minds? I would also not propose to write such a `make_...()` function today since according to some agile design principles, one should rather first write a couple of these components and use them and only much *later* observe that they have something in common and refactor the code in order to get rid of them in an independent refactoring PR.

ctapipe/image/cleaning.py
MaxNoe: Event? Or Dl0 container?

watsonjj: event

MaxNoe: THe nice thing about `__call__` is, it doesn't matter if it's a configurable class or a function. The call is the same.
For example we found that the `HillasReconstructor` could be a simple function, but `TemplateReconstructor` cannot.

If TemplateReconstructor implements `__call__` instead of predict, both have the same interface, e.g.

```
reconstruct_stereo = simple_stereo_hillas
# or
reconstruct_stereo = TemplateReconstruction(templates='templates.hdf5')

but:
```
reconstruct_stereo(event)
```

File: ctapipe/image/cleaning.py (hunk 2 above)
Comment (MaxNoe) [reply ctapipe/image/cleaning.py]: Because this way, it can be configured *per telescope*

File: ctapipe/image/cleaning.py [hunk 5]
```diff
@@ -2,16 +2,65 @@
...
-
-def tailcuts_clean(geom, image, picture_thresh=7, boundary_thresh=5,
-                   keep_isolated_pixels=False,
-                   min_number_picture_neighbors=0):
-
+from ctapipe.core import Component
+from traitlets import Dict, CaselessStrEnum
+
+
+class ImageCleaning(Component):
```
kosack: Rename this to `ImageCleaner` so that we are using the same grammar as other objects in ctapipe (Reconstructor, ChargeExtracter, etc).  "Image Cleaning" is a general concept of denoising, and "Image Cleaner" is something specifically that does denoising.

kosack:
We should collect some of these nice ideas eg in #851

ctapipe/image/cleaning.py
kosack: The only think I don't like about the modular approach they use is the same as for the HESS code: it's hard to then run them in a more simple way (simple for-loop) and debug them, but a lot of it is very nice. Originally the concept for ctapipe (back when it was also mostly C++) was such a modular flow-based approach. We even had ctapipe-flow for that as well, but in the end it's sometimes a bit overkill, and we can wrap in such a system later.

ctapipe/image/cleaning.py
kosack: agreed. And we can keep the grammar semantics as you show:
`clean_image = ImageCleaner(..)` (i.e the instance is named with a verb, while the class is a noun)

ctapipe/image/cleaning.py
watsonjj: I agree with avoiding the modular approach. The things we will wish to do in our pipeline typically have a set order, and its more clear to define what is done at each stage or data level. The freedom we allow the user to set is exactly which algorithm implementation is used for each of the fixed stages (`ChargeExtractor`, `ImageCleaner`, ...).

But for the configuration system, I like the KM3Pipe approach.

MaxNoe: >  it's hard to then run them in a more simple way (simple for-loop) and debug them, but a lot of it is very nice.

What Tamas showed me on his laptop, it's very easy to debug, because he has stuff like drop into ipython at any point for debugging, if I remember correctly

kosack:
Just as another use-case, here is an example of what is used in Julien's pipeline for configuring 2 cleaning types (tailcuts + wavelet) for 2 different image cleaners (reco + energy) as a yaml file.   

```yaml
cleaning_reco:  # note that there are more than one image cleaner defined, this is just the reco one
  tail:
   thresholds:  # picture, boundary
    - LSTCam: [6, 3]
    - NectarCam: [6, 3]
   keep_isolated_pixels: False
   min_number_picture_neighbors: 2

  wave:
   # Directory to write temporary files
   #tmp_files_directory: '/dev/shm/'
   tmp_files_directory: './'
   options:
    LSTCam:
     type_of_filtering: 'hard_filtering'
     filter_thresholds: [3, 0.2]
     last_scale_treatment: 'drop'
     kill_isolated_pixels: True
     detect_only_positive_structures: False
     clusters_threshold: 0
    NectarCam:  # TBC
     type_of_filtering: 'hard_filtering'
     filter_thresholds: [3, 0.2]
     last_scale_treatment: 'drop'
     kill_isolated_pixels: True
     detect_only_positive_structures: False
     clusters_threshold: 0

cleaning_energy:
 ...
```

Obviously this doesn't fit directly in our traitlets scheme, but it's not too bad. You could imagine defining the 2 cleaners with classes:

```py3
class RecoImageCleaner(ImageCleaner):
    pass
class EnergyImageCleaner(ImageCleaner):
    pass
```

Then they are individually configurable via traitlets.config.

kosack:
What is the status of this PR? Is it still being worked on? I think it's still a really needed to have this functionality soon, so the question is: there were a lot of discussions - can we come up with a final design, or accept the one proposed here? It doesn't have to be perfect, but is a nice first example of how we can configure complex multi-implementation/multi-telescope algorithms.

MaxNoe:
Unfortunately, I will have to cut ctapipe developement short for now (there is a PHD thesis to be written),
but maybe @LukasNickel can have a look, he is working on cleaning anyway in the moment.

kosack:
Ok, thanks for the info (and yes your PhD is more important!)  We can discuss this elsewhere then.  There is also 2 other implementations that I know of of an ImageCleaner class... in protopipe and PyWI.  Neither are ideal, though, so a nice simple one is needed.

kosack:
I'm going to close this for now, since I've made an implementation that will be PR'd shortly, so we can continue on that PR.    In that version, I implemented a TelescopeParameter trait type that allows one to specify default values, per-telescope type values, and per-telescope-ID values (and mix an match them).  These get translated all to per-telescope-ID values later, and then are applied.

MaxNoe:
Sounds good!

PR #1023, opened 2019-03-21
ChargeExtractor tests and simplification

watsonjj:
This PR aims to improve the test for charge extractors, and begins the major refactoring described in #1026. Namely, the `ChargeExtractors` are significantly simplified with the intention that they more heavily use a composition approach instead of inheritance in the future.


dneise:
Since the title is "Major refactoring" I am already frightened :-D

So in case you wanna make review easier, here just some thoughts:

If one only modifies tests or adds tests, but not the code base, this is quite simple to review I think.
If one does not touch the tests, but refactors the code base, this is quite easy to review as well. 
(I did not follow my own advice in the past, I know)

So if you happen to be done already (or partly done) with refactoring and adding some tests, we could already merge it.

watsonjj:
I get where you are coming from. I just thought this process would be a bit messy if split into many pull requests. Not sure what is the best way in Github to handle a major pull request with many sub-pull-requests. I will try to split it into stages.

MaxNoe:
If the single diffs are easy to review, that's maybe enough.

dneise:
Cool, thx. I can already say, that I like the current version better than the master. The master tested only the bare minimum, i.e. if the extractors could be called at all ... but it never tested the results.

dneise:
> If the single diffs are easy to review, that's maybe enough. 

yeah ... you are right. I should start reviewing that way and not only look at the end result.

watsonjj:
Okay, this is the first PR for the refactoring of the ChargeExtractors. I have now created the issue #1026 to organise the large amount of planned changes.

I began this PR by creating some tests for the return values of the charges, to make sure I maintain results in my changes. I have then simplified the implementations of the ChargeExtractor.

watsonjj:
This PR should not break API either

watsonjj:
Python 3.7 failing in travis due to pandoc version (I think)

MaxNoe:
Travis passed...

ctapipe/image/charge_extractors.py
```diff
@@ -2,18 +2,67 @@
...
+    # Ensure window is within waveform
+    # width[width > n_samples] = n_samples
+    # start[start < 0] = 0
+    # sum_check = start + width > n_samples
+    # start[sum_check] = n_samples - width[sum_check]
+
+    start = peakpos - shift
+    end = start + width
+    ind = np.indices(waveforms.shape)[2]
+    integration_window = (ind >= start[..., None]) & (ind < end[..., None])
```
dneise: even though it is longer, I would prefer the more explicit `np.newaxis` instead of `None`.

dneise:
I was asked to review, but I still see a couple of `#TODO`s in the code and also some comments, which look like they are meant to imply todos...

Just a question (and yes I know we do not want to optimize prematurely) I saw this:
```python
    charge = np.ma.array(waveforms, mask=~integration_window).sum(2).data
```

and wondered what was the reason again no to use this 
(it is 3x faster on my machine with random data and waveform shapes (2, 1000, 100)):
```python
    charge = (waveforms * integration_window).sum(axis=2)
```
I just tried it out .. and no tests failed when using that.

watsonjj:
The TODOs are placeholders to remind me for the next steps in #1026 

>  wondered what was the reason again no to use this

Thanks for suggesting that and profiling it, I will use it.

dneise:
Ah sorry `np.einsum` is of course even faster, since it avoids the intermediate copy., but I believe `einsum` had some internal precision issues ... (internally casting everything to float32 or so, I do not remember). 
Therefore I could not test `==` but only `np.allclose`.

### profiling 

(`d` is some random data, `m` is a mask (masking away ~80%), shape is (2,1000,100)
```
In [13]: %timeit y = np.einsum('...i,...i', d , m)                                                                                                     
188 µs ± 786 ns per loop (mean ± std. dev. of 7 runs, 10000 loops each)

In [14]: %timeit x = np.ma.array(d, mask=~m).sum(2).data                                                                                                 
947 µs ± 3.2 µs per loop (mean ± std. dev. of 7 runs, 1000 loops each)

In [15]: %timeit X = (d * m).sum(axis=-1)                                                                                                                
305 µs ± 1.3 µs per loop (mean ± std. dev. of 7 runs, 1000 loops each)

In [16]: np.allclose(x, X)                                                                                                                               
Out[16]: True

In [17]: np.allclose(x, y)                                                                                                                               
Out[17]: True
```

watsonjj:
Okay... I will use the `X = (d * m).sum(axis=-1)` form as it is the most readable, while providing some (albeit premature) optimisation.

MaxNoe:
Why multiply and not index? I find it much more readable to use the index form.

watsonjj:
> Why multiply and not index? I find it much more readable to use the index form.

Need to keep the shape/dimensions for the sum. Indexing flattens the array. Could reshape back, but theres the potential that one could have different number of samples in integration window for different pixels.

kosack:
Should this be tagged work in progress? or is it basically done?

watsonjj:
This part is done

watsonjj:
Ah, I have realised a problem with this approach - configuring the window_width and window_shift traitlets cannot be easily done anymore.

Before, one could specifiy the following in a `Tool`:
```
    aliases = Dict(
        dict(
            window_width='WindowIntegrator.window_width',
            window_shift='WindowIntegrator.window_shift',
        )
    )
```
But now, because the different width and shift traitlets are attached to each ChargeExtractor independently, one would have to include the traitlet for each ChargeExtractor.

The only way I see to fix this is keep some inheritance of a common `WindowIntegrator` class, which defines these traits

kosack:
Just a comment, I think it is ok to use `einsum` if it's a real speed gain, but only if we include a comment explaining exactly what it's doing (e.g. write it in tensor notation so we can understand the expectation without reading the manual for `einsum`, which is always a bit confusing due to the syntax).

kosack:
Of course, from what I've learned at PyGamma,  you can also get some of the benefits of einsum by using Numba on a numpy function, keeping the readability (but then you have to ensure the function has simple ndarray inputs and outputs)

watsonjj:
> using Numba

I intend to do an optimisation study towards the end of #1026, which will include using numba.

watsonjj:
I'm not sure what is up with codacy, but this PR is ready for reviews and merge

watsonjj:
I have adopted the approach mentioned in #1029, which suggests to more sparingly use the aliases feature. This therefore allows me to remove the abstract classes once again, returning to the simple Component-inheritence (only inherit from ChargeExtractor) I had intended for this PR.

ctapipe/image/charge_extractors.py
```diff
@@ -2,18 +2,60 @@
...
+        Shape (if numpy array): (n_chan, n_pix)
+    shift : ndarray or int
+        Window size of integration window.
+        Shape (if numpy array): (n_chan, n_pix)
+
+    Returns
+    -------
+    charge : ndarray
+        Extracted charge.
+        Shape: (n_chan, n_pix)
```
dneise: The docu regarding Returns is not correct: it returns charge **and** integration window

ctapipe/image/charge_extractors.py
```diff
@@ -105,589 +128,161 @@ def extract_charge(self, waveforms):
...
-    def _obtain_peak_position(self, waveforms):
-        nchan, npix, nsamples = waveforms.shape
-        return np.full((nchan, npix), self.t0, dtype=np.intp)
+        # TODO: remove integration window return
+        peakpos = np.zeros(waveforms.shape[:2], dtype=np.intp)
+        window = np.ones(waveforms.shape, dtype=np.bool)
+        return waveforms.sum(2), peakpos, window
 
 
-class PeakFindingIntegrator(WindowIntegrator):
```
dneise: was `PeakFindingIntegrator` removed without replacement? If so, why?

dneise: Ah I see, `PeakFindingIntegrator` was abstract and only used in order to derive form it, right?

ctapipe/image/tests/test_charge_extraction.py
```diff
@@ -12,95 +14,96 @@
...
 
-    data = example_event.r0.tel[telid].waveform
-    nsamples = data.shape[2]
-    ped = example_event.mc.tel[telid].pedestal
-    data_ped = data - np.atleast_3d(ped / nsamples)
-    data_ped = np.array([data_ped[0], data_ped[0]])  # Test LG functionality
+def test_full_integration(camera_waveforms):
+    waveforms, _ = camera_waveforms
+    integrator = FullIntegrator()
+    integration, _, _ = integrator.extract_charge(waveforms)
```
dneise: the return values of `extract_charge` seem to be called `charge, peakpos, window` ... why not keep `charge` but replace it by `integration` here?

ctapipe/image/tests/test_charge_extraction.py
```diff
@@ -12,95 +14,96 @@
...
+
+    assert_allclose(integration[0][0], 545.945, rtol=1e-3)
+    assert_allclose(integration[1][0], 970.025, rtol=1e-3)
```
dneise: can one explain where these numbers come from? Could it help future readers of this ?

ctapipe/image/charge_extractors.py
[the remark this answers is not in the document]
thomasgas: The documentation says that they are the same (https://docs.scipy.org/doc/numpy/reference/generated/numpy.expand_dims.html), but I also prefer `np.newaxis`, even though if you profile them on a 1d array:
1. `x[..., None]` takes 200 ns
2. `x[..., np.newaxis]` takes 306 ns
3. `x[:, None]` takes 310 ns
4. `x[:, np.newaxis]` takes 360 ns

ctapipe/image/charge_extractors.py
watsonjj: Correct

ctapipe/image/tests/test_charge_extraction.py
watsonjj: These values are just the charge extracted for the first pixel in each channel for this ChargeExtractor

ctapipe/image/tests/test_charge_extraction.py
watsonjj: Ok, I've refactored these to `charge` now

ctapipe/image/charge_extractors.py
```diff
@@ -2,18 +2,67 @@
...
 import numpy as np
-from traitlets import Int, Float
+from traitlets import Int
 from ctapipe.core import Component
 from ctapipe.utils.neighbour_sum import get_sum_array
 
 
+def extract_charge_from_peakpos_array(waveforms, peakpos, width, shift):
+    """
+    Build the numpy array of bools defining the integration window.
```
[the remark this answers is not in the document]
watsonjj: Fixed

ctapipe/image/charge_extractors.py
watsonjj: Fixed

ctapipe/image/charge_extractors.py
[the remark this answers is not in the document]
dneise: OMG thomas :-D this is a difference of 100ns, and the line directly below this takes **atleast** 100us, that is 3 orders of magnitude more. Let us not start hunting fractions of us here yet.

ctapipe/image/tests/test_charge_extraction.py
dneise: Right .. erm .. how were they extracted? With the same algorithm we are testing? If so this is somehow circular, no? (Sometimes this is the best we can do, I think)

watsonjj: Yeah, it is a bit. They are put here to simply test against future commits, to make sure the expected values do not change, or at least make the contributor aware that their changes has perhaps had an unexpected impact further down the chain.

dneise: thx

ctapipe/image/charge_extractors.py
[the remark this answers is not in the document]
thomasgas: eheh don't worry about it :smile: I'm not that crazy :laughing:

watsonjj:
@kosack Could you restore your approval for this PR? Or @thomasgas?

PR #1027, opened 2019-03-22
Fix tests for Tools, fixes #1214

### LINKED ISSUE #1025: Some tools are broken
Seems we are missing some tests, since the update to Tool and Component recently seems to have broken some tools.  For example:

```sh
$ ctapipe-reconstruct-muons --input=gamma_test_large.simtel.gz

INFO [MuonDisplayerTool] (tool/initialize): ctapipe version 0.5.2.post636+gitcc215e2
INFO [MuonDisplayerTool] (tool/run): Starting: ctapipe-reconstruct-muons
  File "/Users/kosack/miniconda3/envs/cta-dev/bin/ctapipe-reconstruct-muons", line 11, in <module>
    load_entry_point('ctapipe', 'console_scripts', 'ctapipe-reconstruct-muons')()
  File "/Users/kosack/Projects/CTA/Working/ctapipe/ctapipe/tools/muon_reconstruction.py", line 124, in main
    tool.run()
  File "/Users/kosack/Projects/CTA/Working/ctapipe/ctapipe/core/tool.py", line 165, in run
    self.setup()
  File "/Users/kosack/Projects/CTA/Working/ctapipe/ctapipe/tools/muon_reconstruction.py", line 71, in setup
    config=self.config, parent=self, eventsource=self.source
  File "/Users/kosack/Projects/CTA/Working/ctapipe/ctapipe/calib/camera/calibrator.py", line 78, in __init__
    super().__init__(config=config, parent=parent, **kwargs)
  File "/Users/kosack/Projects/CTA/Working/ctapipe/ctapipe/core/component.py", line 100, in __init__
    'Only one of `config` or `parent` allowed'
ValueError: Only one of `config` or `parent` allowed If you create a Component as part of another, give `parent=self` and not `config`
```

### LINKED ISSUE #1214: Tools should give a reasonable return value to the shell
Right now in `ctapipe.core.Tool.run()` different types of errors are caught and the activity status in the Provenance are updated.  However, there are no return values: it should also return 0 on success, and a standard error code if not (See e.g. http://tldp.org/LDP/abs/html/exitcodes.html)

See also: https://github.com/cta-observatory/cta-lstchain/issues/271

dneise:
Fixes #1025


dneise:
So first I make sure the test fails.

ctapipe/core/tool.py
```diff
@@ -170,23 +174,28 @@ def run(self, argv=None):
             Provenance().finish_activity(activity_name=self.name)
         except ToolConfigurationError as err:
             self.log.error(f'{err}.  Use --help for more info')
+            return 2  # wrong cmd line parameter
```
kosack: will this prevent the `finally` block from executing?

kosack: I think we should make sure it still runs, since that is where provenance is written (which can still be written if the tool failed, with the appropriate status recorded). That is perhaps useful if we use the provenance info to track if jobs complete correctly or not

kosack:
probably should rename this PR to show that it fixes all tools!

watsonjj:
Can we also test `tool.run(['--help'])` for every tool? It can often raise some errors when traitlets have been changed, but not updated in the tool.

ctapipe/core/tool.py
kosack: I think having a single "return" point is better form anyhow (minimize complex branching), so just refactoring this to set a `return_value` and returning it at the end should suffice.

MaxNoe: No, finally will still execute. See https://docs.python.org/3.7/tutorial/errors.html#defining-clean-up-actions

> A finally clause is always executed before leaving the try statement, whether an exception has occurred or not. When an exception has occurred in the try clause and has not been handled by an except clause (or it has occurred in an except or else clause), it is re-raised after the finally clause has been executed. The finally clause is also executed “on the way out” when any other clause of the try statement is left via a break, continue or return statement.

watsonjj:
I have created #1034 to test for the help message of tools

ctapipe/core/tool.py
kosack: Ah, that's not at all what I expected! Ok, it's fine then! I had to actually try it myself.

kosack:
this is mostly ready to merge, but there are a few minor conflicts

ctapipe/tools/tests/test_tools.py
```diff
@@ -11,13 +11,13 @@
...
         argv=shlex.split(
-            f'--events={GAMMA_TEST_LARGE} '
+            f'--input={GAMMA_TEST_LARGE} '
             '--max_events=2 '
         )
     )
+    assert return_code == 0
 
     with pytest.raises(SystemExit):
         tool.run(['--help-all'])
```
kosack: this seems to be failing (it could be a line that is there due to a bad merge, since `tool` doesn't exist in the new version)

ctapipe/tools/muon_reconstruction.py
```diff
@@ -120,4 +119,4 @@ def finish(self):
 
 def main():
     tool = MuonDisplayerTool()
-    tool.run()
+    return tool.run()
```
MaxNoe: I think this needs to be `sys.exit(tool.run())`, not `return` https://docs.python.org/2/library/sys.html#sys.exit

ctapipe/core/tool.py
```diff
@@ -202,23 +206,28 @@ def run(self, argv=None):
...
```
MaxNoe: maybe using `sys.exit` everywhere here in `Tool.run` would be the simplest solution.

 The tests could then do something like this:
```python
try:
    tool.run()
except SystemExit as e:
    assert e.code == 0
```

kosack: or just put the "return" in the `finally:` block, and make a variable like `exit_code=2`, since we don't really want to exit exactly inside that except block (things in finally should still run)

kosack: but using `sys.exit()` is also fine, as long as it happens later.

kosack: As long as the `finally:` block still runs, I think that is ok, and is probably more clear. I didn't realize `sys.exit()` raises an exception rather than immediately exists (as in c)!

kosack: The goal being to also automatically store the exit information in the provenance (returning immediately, breaks that)

MaxNoe: That should probably solver by using provenance as context Manager und in run?

Exit geht's the exception und ist automatically reraised if Exit does not return true

kosack: yes, i agree - making provenance a context manager would be much nicer. I want to clean it up anyhow (the current interface is not great).  For this PR, i think it should still work, since we still get to the finally block if there is an error.

kosack:
I think doing a pull from upstream master would solve the travis issues in this PR (which are not related to the changes, but to a fix that was made in a previous PR)

MaxNoe:
The branch is in this orga, not in dominiks fork, so you can just push to it.

Unfortunately, dominik is no longer active in gamma ray astronomy

kosack:
Ok thanks, will do

ctapipe/core/tool.py
```diff
@@ -229,6 +239,9 @@ def run(self, argv=None):
             with open("provenance.log", mode="w+") as provlog:
```
MaxNoe: This opens a hardcoded file and overwrites it without any warning. 
This is probably not really what we want here, right?

kosack: probably not, it was just a hack for now to get the provenance info somewhere. 

 At minimum it should probably append, not overwrite, and perhaps the name should be a config option, or at least prepend the tool name (so you get `mytool-provenance.log`).   In the future, it will just be attached to output files, but so far that is not implemented.

kosack:
The only annoying thing I see now with this, is that if a user wants to use a Tool in a script rather than running it as a comand-line application, they need to catch SystemExits, otherwise their script dies. But not sure there is a cleaner way.

MaxNoe:
Actually, we could only call `sys.exit` if the exit code is not 0. That would raise only if there is an error, which is good and in all other cases, the default exit code of python will be 0.

MaxNoe:
From the travis log, it seems traitlets applications have an exit method that raises sys.exit which is called by default.

MaxNoe:
Ok, now I am satisfied with this.

* A new function run_tools catches the `SystemExit` and returns the status code

* A semantic change: the finish method is not called under error conditions anymore.
The charge extractor used this for writing the file, even e.g. if an exception occured in `start`.
I think it is also hioghly surprising, that on error some possibly complex code is started.

MaxNoe:
@kosack codacy complains that I use asserts in unit tests, do we really want to continue using that?

kosack:
> @kosack codacy complains that I use asserts in unit tests, do we really want to continue using that?

It has a lot of annoying things like that, but it is quite configurable. I just didn't find a good way to ignore unit tests. Basically it just runs prospector and pyflakes.  We used to use another system, but it had even worse problems. Until CTAO provides one, let's just stick with it and try to update it's configuration.

watsonjj:
@MaxNoe Do you have a moment to fix the conflicts, so we can push to get this merged?

MaxNoe:
Ahyes, sorry I thought this was already merged

MaxNoe:
@kosack @watsonjj this should be done now

PR #1032, opened 2019-03-26
Replace C function get_sum_array with a Numba implementation

watsonjj:
This PR removes the ctypes-wrapped function `get_sum_array` that was used by the NeighbourPeakIntegrator, and replaces it with the numba function `neighbor_average_waveform`. This new function provides almost identical performance as the C function.

This change therefore removes the only C code that needed to be compiled when installing ctapipe.


watsonjj:
It seems codecov does not consider jitted functions...

dneise:
This version (not using jit or so) is 50% slower than the jitted version:
```python

def neighbor_average_waveform2(
    waveforms,
    neighbor_matrix_sparse,
    lwt
):
    '''
    neighbor_matrix_sparse:  from
    `ctapipe.instrument.CameraGeometry.neighbor_matrix_sparse`
    '''
    number_neighbors = np.array(neighbor_matrix_sparse.sum(axis=1))

    result = np.zeros_like(waveforms)
    for channel_id in range(waveforms.shape[0]):
        w = waveforms[channel_id]
        result[channel_id] = (
            (neighbor_matrix_sparse.dot(w) + lwt * w) / number_neighbors
        )
    return result
```

watsonjj:
Are you against using Numba?

dneise:
Nope .. just played around with different implementations in order to see what's what. Was impressed by how nicely the dot product for sparse matrices performs in this case and wanted to share it.

watsonjj:
Cool. I had considered sparse arrays, but I unfamiliar with them. Nice to see how it can be done :)

dneise:
One more question regarding this function. It seems to be needed only in `NeighbourPeakIntegrator`. I must say I never looked at that integrator very much before today. I was wondering .. is it used by any existing telescope? 

Naively I would have expected the order of operations "1st: average neighbors,  2nd: find timing" to be reversed, so "1st find the arrival time on every pixel, then average neighbors timing, to determine integration window"... so I was wondering: is there a paper about this kind of extraction method, which goes into detail about the method?

dneise:
Ah one more thing ... we currently have 3 representations of the neighbors: 
 - `neighbor_matrix_where`,
 - `neighbor_matrix_sparse` and
 - `neighbor_matrix`

I just checked it quickly, so I might have overlooked something, but its seems: 
 - `neighbor_matrix_where` is only used here, i.e. in charge_extractors.NeighbourPeakIntegrator`
 - `neighbor_matrix_sparse` is used in `image.cleaning` (but might be useful here)
 - `neighbor_matrix` is not used at all.

So one could .... maybe ... in principle .... reduce the number of different neighbor representations from 3 to one and just call it `.neighbor_matrix` ... or `.neighbors`.

But that's not urgent or so. This PR is all good, I believe.

kosack:
> It seems codecov does not consider jitted functions...

I guess it's because that python code is never really "called", it's introspected and then compiled to another function that is then run. Probably no way to fix that ourselves.

kosack:
actually, it seems from here (https://stackoverflow.com/questions/26875191/analyzing-coverage-of-numba-wrapped-functions) it suggests we set NUMBA_DISABLE_JIT during the coverage checks to have it report a correct coverage result.  Not sure it's worth it though.

watsonjj:
> One more question regarding this function. It seems to be needed only in `NeighbourPeakIntegrator`. I must say I never looked at that integrator very much before today. I was wondering .. is it used by any existing telescope?
> 
> Naively I would have expected the order of operations "1st: average neighbors, 2nd: find timing" to be reversed, so "1st find the arrival time on every pixel, then average neighbors timing, to determine integration window"... so I was wondering: is there a paper about this kind of extraction method, which goes into detail about the method?

It's a common charge extraction technique to use for IACTs due to its reduced charge bias. Its the default charge extractor for ctapipe.

Unless I'm severely mistaken, taking the argmax to find the peak time in each pixel, and then getting the average of those times is not equivalent to getting the average waveform, and then extracting the argmax of it.

> So one could .... maybe ... in principle .... reduce the number of different neighbor representations from 3 to one and just call it .neighbor_matrix ... or .neighbors.

I agree that it would be nice to achieve this in the future.

MaxNoe:
> It's a common charge extraction technique to use for IACTs due to its reduced charge bias. Its the default charge extractor for ctapipe.

I'm pretty sure neither FACT nor MAGIC use this kind of neighbor information for charge extraction.
That this is the default in ctapipe does not mean anything, right?

So please, could you provide any references?

Issue #1024, opened 2019-03-22
Use `__call__` where apropriate


dneise:
We have some components in ctapipe, which are in principle callables and offer only one real interface, which is to call them:

Possible examples are:
 - `CameraCalibrator.calibrate()`
 - `CameraDL1Calibrator.calibrate()`
 - `CameraDL0Reducer.reduce()`
 - `ChargeExtractor.extract_charge()`
 - `DataVolumeReducer.reduce_waveforms()`
 - `WaveformCleaner.apply()`
 - `IntensityFitter.fit()`
 - `Reconstructor.predict()`

In #1022 Max thankfully starting tidying this up. This issue is just a reminder for us to tidy the rest up as well.

Issue #1026, opened 2019-03-22
Major refactoring of calibration chain


watsonjj:
Transfer of #1023 into an issue, to organise the different pull requests involved in the major refactoring.

## 1. PR #1023
- [x] Remove dependance of tests on files from ctapipe_extra (PR #1023)
- [x] Add values to unit tests to ensure behaviour is kept for upcoming refactoring (PR #1023)
- [x] Refactor `ChargeExtractors` into a composition approach for simplification (replacing the inheritance approach) (see #907) (PR #1023)

## 2. PR #1032 
- [x] Replace C function used in the neighbor charge extractor with a Numba implementation

## 3. PR #1033 
- [x] Create `ImageExtractor` class to replace `ChargeExtractor` and `WaveformCleaner`
- [x] Make `ImageExtractor` return two parameters per event - charge and peak time (floating point)
- [x] A simple peak time extraction is using the weighted average (see #1021)
- [x] Switch to using `__call__` (see #1022 and #1024)

## 4. PR #1038
- [x] Optimisation study, and consideration of using numba for charge extraction

## 5. PR #1045 
- [x] Consider more robust calculations of pulse_time (see #1033)
- [x] Optimisation and improvement of pulse time extraction

## 6. PR #1046 
- [x] Retire CameraR1Calibrator, make filling r1 waveform container the responsibility of the `EventSource`

## 7. PR #1048
- [x] Simplify calibration chain into a single class, which is composed of individual calibration steps (gain selection, data reduction, image extraction, charge calibration)

## 8. PR #1095 
- [x] Gain selecting (replacing #683)

## 9. PR #1146 
- [x] Combine `extract_pulse_time_around_peak` and `sum_samples_around_peakpos`

## 10. PR #1155
- [x] Improve passing of neighbors to algorithm (see #909)
- [x] Enable individual calibration config per telescope (#703 #702)

## 11. PR #1160
- [x] Add flat field, absolute signal, and pedestal calibration to DL1Calibrator

## 12. PR #1242 
- [x] Move integration_correction from CameraCalibrator to ImageExtractors
- [x] Apply sampling_rate in ImageExtractors

## 13. PR #1283 
- [x] Update documentation and notebooks to describe new calibration chain

# Delegated to other issues:
- [x] Restructure calib modules and location of ImageExtractors
   - #1205 and #1037
- [x] Classes to fill calibration fields in event
   - #1226 and #1147 and #1037 
- [x] Extract pulse width from waveform
[...]

Issue #907, opened 2018-12-18
Peakpos calculation


thomasgas:
the peakpos image that we get out now in ctapipe is not very useful for small images since the pixels that contains photons from the shower are very few and they all have the same time value (not very useful to compute a gradient).
@kosack you also explained to me some details (that now i don't remember quite well) but it would be nice if you write down here how it's done and how it could be implemented to have a more useful peakpos image to be used to compute the time gradient.

MaxNoe:
You could take the weighted average of the waveform, I think this is done in MAGIC.

So `np.average(np.arange(num_samples), weights=waveform)`

For the production case, you should not create the arange array in each iteration, but once.

kosack:
The problem right now is the current peakpos (and for that matter the trace integration) is ulta simplistic, and does no interpolation. The idea, I believe (@watsonjj would be able to comment more) was to keep it to simple sums and implement interpolation as a pre-processing step in a WaveformCleaner, effectively generating more time bins by oversampling/filtering. 

However, as we discussed at the meeting, that is somewhat of a waste of CPU time and memory, since it then makes the pixel data much larger, so we should consider implementing a more fancy methods for integration and time calculations that are floating-point (i.e. not tied to integer sample numbers).  The question is how easy it is to do without being too slow or too complex code.    The simplest would be to create a 1D interpolator for each pixel's waveform, and then do true numerical integration.  However, some knowledge of signal processing and the hardware is needed to choose the optimal interpolation (in the end it may actually be more accurate to apply the correct filter to oversampled data).  I'd guess that for GHz samplers, simple linear interpolation may be fine, but for the 250 MHz ones, more complex signal filtering may be better.

watsonjj:
Originally the idea was that waveform cleaning (which could include things like filtering and upsampling) was a separate step before the charge extraction. Therefore different combinations between WaveformCleaners and ChargeExtractors could be flexibly chosen. However, I have also come to the realisation that even this is a bit strict as some charge extraction approaches don't really conform to this split. Also conforming to this split may cause a waste in CPU time and memory, as you said. 

I personally have been considering this for a while, and I have some up with two potential improvements to the current charge extraction procedure:
1. further splitting of the steps (Cleaning, PeakFinding, Upsampling, Extraction)
2. perhaps we abandon allowing flexible combination of techniques at runtime, and instead have a list of hardcoded ChargeExtractors containing all of the steps. For example, a CCNeiChargeExtractor class might do:

[...]

PR #1033, opened 2019-03-26
Replace WaveformCleaner and ChargeExtractor with WaveformExtractor

watsonjj:
This PR replaces the `WaveformCleaner` and `ChargeExtractor` with `ImageExtractor`.

The `ImageExtractor's` purpose is to extract the `charge` and `pulse_time` from a waveform. It can also apply any sort of waveform cleaning, upsampling, interpolation etc. to extract these parameters. Essentially, an `ImageExtractor` is some combination of the functions defined in extractor.py to obtain the `charge` and `pulse_time`.

The `extract_charge` method of the old `ChargeExtractor` has been replaced to use `__call__` instead, as per #1024.

An initial algorithm for extracting the `pulse_time` (via a weighted average of the waveform) is also included in this PR. This approach may be replaced with a different `pulse_time` calculation if it is identified as providing a better time resolution.

The containers, calibration, and tools have also been updated to handle the new DL1 contents.


kosack:
Perhaps there is a better name... right now "WaveformExtractor" seems like it extracts waveforms, and you would then get back a waveform.    

At the end this algorithm returns estimated charges, right?  Is that intended to be before or after calibration?  I guess before (even if we did it backwards before), but if before it's actually more still a ChargeExtractor (even if there is some internal smoothing), or if there is calibration, it could be an "IntensityExtractor", since the final output is a calibrated light intensity.

watsonjj:
The `WaveformExtractor` returns both charge and pulse_time.... I am open to a better name. But `IntensityExtractor` doesn't fully capture it either. I did start with `WaveformReducer`, as we are reducing the waveform down to two parameters, but I felt this was too similar to `DataVolumeReducer`.

watsonjj:
> Is that intended to be before or after calibration? 

Before flat-fielding / calibration into p.e.

Yeh, the simtelarray waveforms are somewhat calibrated into p.e. in their R1Calibrator. But not fully (as the full calibration depends on the charge extraction scheme used, and the correct calib scale). There needs to be an additional calibration step added after the charge extraction, as discussed in the calibration call a few months back. I intend to add it.

watsonjj:
Some ideas for the name:

1. SignalExtractor
2. ImageExtractor
3. CherenkovImageExtractor
4. WaveformEvaluator
5. WaveformReducer
6. ChargeExtractor (and we just accept that it also returns pulse_time)

kosack:
`ImageExtractor` isn't too bad, but it may be too general. It does extract "images" of charge and time, but that's not so clear from the name. I guess there is no perfect one!

watsonjj:
@FrancaCassol This PR has remove the waveform_cleaner module. I have moved your `BaselineSubtractor` class into the `subtract_baseline` found inside extractor.py. I have also created the `BaselineSubtractedNeighborPeakWindowSum` extractor, which from uses the `subtract_baseline` function before extracting the charge using the neighbouring pixels.

The idea in this new approach is that an `ImageExtractor` performs all of the operations required to extract the charge and time from the waveforms, including any waveform cleaning. I hope this makes sense.

watsonjj:
@kosack @dneise @MaxNoe @FrancaCassol It would be really appreciated to have this reviewed so I may continue :)

dneise:
Sorry for reviewing late and then asking stupid questions. Why was this API change needed/wanted again?

I mean this part:

>This PR replaces the WaveformCleaner and ChargeExtractor with ImageExtractor.

watsonjj:
The aim is to greatly simplify the charge extraction process. The first discussion on this can be found at #907.

To summarise, this seperation of waveform cleaning and charge extraction is unnecessary, as the two are often very closely coupled. An image extractor should define the waveform operations it wants to use in order to extract the charge (and time). 

One example is a cross correlation charge extraction technique. Applying the cross correlation to a waveform with a reference pulse results in a trace which defines how correlated the waveform is with the pulse for each position in the waveform. This could be considered a waveform cleaning. However, the value of each sample of this result is already an integration of charge, therefore one only needs to select the correct sample. It does not make sense to allow the user to select a window for this charge extraction technique. See my ASWG presentation in Barcelona for more details on the cross correlation technique if you are interested.

I don't think the freedom to flexibly chose a waveform cleaner and charge extractor (and mix and match at will) is necessary at run time. Instead, I think it is better to explicitly define individual charge extractors that are deemed to be useful implementations, which the user can select from.

watsonjj:
And a further point, the `ChargeExtractor` was replaced with `ImageExtractor`, as its purpose is to extract images of both charge and time from the waveform.

dneise:
Wow ..long text ... but thanks a lot

ctapipe/image/extractor.py
```diff
@@ -104,16 +106,69 @@ def neighbor_average_waveform(waveforms, neighbors, lwt):
...
+
+    Returns
+    -------
+    pulse_time : ndarray
+        Floating point pulse time in each pixel
+        Shape: (n_chan, n_pix)
+
+    """
+    samples_i = np.indices(waveforms.shape)[2]
+    pulse_time = np.average(samples_i, weights=waveforms, axis=2)
```
dneise: Performance note: line 127 can be pulled out of this function and be done only once per telescope. 
I found it interesting to see, that for waveform shapes like (2, 1000, 300) lines 127 and 128 both took equally long, ~1.7ms or so on my machine. So moving line 127 out of this, should reduce runtime by ~50%.

I found an ugly hack to make `sample_i` in a faster way using something called `stride_tricks` ... I would not recommend using it ... but it might be interesting to know about.
```
In [65]: %%timeit
    ...: i = np.indices(waveforms.shape)[2]
    ...: result = np.average(i, weights=waveforms, axis=2)
3.55 ms ± 16.8 µs per loop (mean ± std. dev. of 7 runs, 100 loops each)

In [63]: %%timeit 
    ...: S = waveforms.shape
    ...: I = np.arange(S[2])
    ...: i = np.lib.stride_tricks.as_strided(I, (S[0], S[1], I.size), (0, 0, I.itemsize))  # a 3D view, no copy of the data
    ...: result = np.average(i, weights=waveforms, axis=2)
1.53 ms ± 5.35 µs per loop (mean ± std. dev. of 7 runs, 1000 loops each)
```

watsonjj: Yep this is a good point. I also noticed that np.indices is surprisingly slow in my investigations for #1038. It would be my intention to replace this function with a numba for-loop version in a future PR, which should result in a huge improvement in performance.

I did not know about `stride_tricks`, thank-you for bringing it up.

ctapipe/image/extractor.py
```diff
@@ -104,16 +106,69 @@ def neighbor_average_waveform(waveforms, neighbors, lwt):
...
+    pulse_time[pulse_time < 0] = 0
+    pulse_time[pulse_time > waveforms.shape[2]] = waveforms.shape[2]
```
dneise: I'm not sure ... is it possible to have so funny weights, that the pulse_time can be outside of the range `[0...waveforms.shape[2]]`? ... I think not... unless ... of course .. when the weights are negative, then the average can be pulled out of range, but I think this should be avoided. 

So I played a little bit around with this kind of average ... I used this kind of signal:
image

As one can see the actual pulse start time and the result of this `np.average(sample_i, weights=waveforms)` is clearly correlated as expected. 

But as soon as the baseline moves around problems occur:
image

image

image

image

So I think there needs some care to be taken with respect to these weights. And one should not simply say ... ah ... when the pulse_time is negative .. then we set it to zero ... I think when the pulse_time is out of range, this simply is a strong hint, that something was not good about the range of these weights.

dneise: Ah sorry .. I just see, I labelled my axis wrongly: What I called `average_time` is actually called `pulse_time`. 

So `dneise_average_time = pulse_time = np.average(samples_i, weights=waveform)`

watsonjj: Yeah you are absolutely right. This implementation for obtaining pulse time is very simple, but not very robust. The limit I added for the pulse time to be within the waveform readout was done to make it easier to plot this. Thinking about it now, it could be better to set such unrealistic pulse times to nan (what do you think?).

Or do you suggest only use samples with positive weights in the calculation?

I am inclined to suggest that contributions are made for better pulse time algorithms in future pull requests, which can replace the current function used in the ImageExtractors. Unless you are adamantly against accepting this implementation temporarily.

dneise: Maybe not cut at all then. The result of this implementation is (even in the best case) only correlated with and not identical to the arrival time (50% of rising edge) .. so it is difficult to say what the allowed range is ... I think .. so maybe avoiding a cut and leave the cut to the user of this implementation is ok.

dneise: No no .. simple implementations are good .. just this cut was puzzling me.

watsonjj: Without the cut, it made plotting tools (such as the ctapipe-event-viewer) useless for checking the pulse time, due to the z axis range. I would rather not impose a limit inside this plotting tool, as then it is a misleading visualisation of the data. 

I preferred having these cuts in the algorithm because it's not physical to have a pulse time outside the window.

I would prefer to improve the implementation rather than outright remove the cuts, should you request it.

Would setting it to nan be more acceptable?
How do your plots look if you only consider positive samples in the weighted average?

dneise: When cutting away the negative weights the correlation looks a bit better, but mind the difference in x and y axis ranges .. y goes from (15, 70) while x is from (0, 50). I think this is an effect of the pulse shape... and this is independent of the offset or the noise ...

dneise: image

watsonjj: Okay, that seems somewhat better, thanks. However, with this implementation we would have to handle the case where there are no positive samples, which would involve setting the pulse_time to some value such as zero.

Could we reserve consideration of better pulse_time algorithms for a different PR?

watsonjj: I have added it as a task in #1026

dneise: Indeed if you set out of range results to nan that might be better, since the user knows there was indeed a cut.

dneise: > Could we reserve consideration of better pulse_time algorithms for a different PR?

totally. You asked me do produce these plots.

ctapipe/image/waveform_reducer.py
```diff
@@ -147,21 +165,20 @@ def extract_charge(self, waveforms):
...
-        window = np.ones(waveforms.shape, dtype=np.bool)
-        return waveforms.sum(2), peakpos, window
+    def __call__(self, waveforms):
+        charge = waveforms.sum(2)
+        pulse_time = extract_pulse_time_weighted_average(waveforms)
+        return charge, pulse_time
 
 
-class SimpleIntegrator(ChargeExtractor):
+class UserWindowSum(WaveformReducer):
```
dneise: I would remove the `User` here.

watsonjj: I disagree, all of these extracts do a window sum, but for this one the window is specified manually by the user

dneise: So ... `GlobalPeakWindowSum` needs to be called `UserGlobalPeakWindowSum` because the user has to manually the width and the shift?

watsonjj: Not really...the user does not manually select a fixed window position in that case, it is calculated from the waveform

dneise: okay fine.

MaxNoe: I strongly oppose calling anything `User`. 
Name it like what it needs or does. `StaticWindowSum` or similar.

User does not tell anything and most people will not share the idea that a user of ctapipe should be concerned with integration schemas for raw data.

watsonjj:
I have renamed the UserWindowSum and set out-of-range pulse_times to zero.

@kosack @dneise Could you please restore your approvals?

watsonjj:
Setting the out-of-bounds pulse-times to nan can cause problems down the chain (some algorithms are not happy with nans existing in an array). This is currently causing an error in travis. I suggest that the pulse time is set to -1 for the time being, and we create a better pulse_time calculation in a different PR.

PR #1038, opened 2019-04-03
Create numba ufunc for sum of samples within charge extraction window

watsonjj:
This PR contains commits waiting for approval in #1033. This PR should not be considered until that PR has been merged.

This PR replaces the `extract_charge_from_peakpos_array` function with `sum_samples_around_peakpos`. This is a numba function that more efficiently extracts the charge from the waveform within the window defined by the peak position, window width, and window size.

It is defined using the `@guvectorize` decorator, which creates a numpy universal function, enabling the passing of both scalars and arrays for the peak position, window width, and window shift arguments.

The way the `@guvectorize` works (in this case) is that you define the operation that is applied for each channel and pixel. The operation is then optimised using numba's just-in-time compilation.

I have profiled `sum_samples_around_peakpos` against the previous `extract_charge_from_peakpos_array` function. The new function provides a **factor of 60** reduction in execution time (from a couple of ms, to tens of µs).

It also provides a factor of 2 reduction in execution time upon the `waveforms[:, :, start:end].sum(2)` operation performed in `GlobalPeakWindowSum`.

The execution time for the extractors before this change were:
```
GlobalPeakWindowSum
6.15 ms ± 57.2 µs per loop (mean ± std. dev. of 7 runs, 100 loops each)
LocalPeakWindowSum
12 ms ± 209 µs per loop (mean ± std. dev. of 7 runs, 100 loops each)
NeighborPeakWindowSum
14.7 ms ± 293 µs per loop (mean ± std. dev. of 7 runs, 100 loops each)
```

The execution time for the extractors following this change are:
```
GlobalPeakWindowSum
3.26 ms ± 30.6 µs per loop (mean ± std. dev. of 7 runs, 100 loops each)
LocalPeakWindowSum
3.28 ms ± 28.3 µs per loop (mean ± std. dev. of 7 runs, 100 loops each)
NeighborPeakWindowSum
5.67 ms ± 33 µs per loop (mean ± std. dev. of 7 runs, 100 loops each)
```

The bottleneck is now the pulse_time calculation, which I could address in a different PR.


ctapipe/image/extractor.py
```diff
@@ -21,48 +21,60 @@
...
+
+@guvectorize(
+    [
+        (float64[:], int64, int64, int64, float64[:]),
+        (float32[:], int64, int64, int64, float64[:]),
+    ],
+    '(s),(),(),()->()',
+    nopython=True,
+)
+def sum_samples_around_peakpos(waveforms, peakpos, width, shift, ret):
```
kosack: isn't `peakpos` now renamed `pulse_time`? perhaps updating the terminology everywhere is useful (just doing a replace all should work)

kosack: Ah, ok.  Perhaps `peak_index` is better?  it's not clear from the name that it is different from the peak time

watsonjj: Okay, I have switched to `peak_index`

thomasarmstrong:
I have had a look over the code and have run some tests to compare these changes to the extraction methods in the current master branch. Everything seems good to me, in fact this pull request also fixes a current bug in the GlobalPeakWindowSum which causes it to fail if there is only one channel.  I would suggest that the PR is approved.

(One note, there is a typo in the sum_samples_around_peak description "The ret argument is required by numpy to _creae_ the numpy array which is")

ctapipe/image/extractor.py
```diff
@@ -21,48 +21,60 @@
...
+    peak position, window width, and window shift.
+
+    This function is a numpy universal function which defines the operation
+    applied on the waveform for every channel and pixel. Therefore in the
+    code body of this function:
+        - waveforms is a 1D array of size n_samples.
+        - peak_index, width and shift are integers, corresponding to the
+            correct value for the current pixel
+
+    The ret argument is required by numpy to creae the numpy array which is
```
MaxNoe: typo in creae

watsonjj: Fixing this tiny typo results in all reviews requiring restoration, and all test being ran. Annoying!

MaxNoe: yes, is there somthing smarter? Probably not, right?

MaxNoe: Maybe we should just deactivate the automatic dismissal of reviews and just do it by hand if the author think changes require new reviews.

watsonjj:
@kosack @MaxNoe Could you restore your review here

PR #1039, opened 2019-04-05
Update bokeh plotters to handle nan

watsonjj:
The bokeh plotter will now correctly plot nan as a transparent pixel

### DIFF
--- a/ctapipe/utils/rgbtohex.py
+++ b/ctapipe/utils/rgbtohex.py
@@ -26,7 +26,7 @@ def intensity_to_rgb(array, minval=None, maxval=None):
     if minval is None:
         minval = np.nanmin(array)
     if maxval is None:
-        maxval = np.nanmin(array)
+        maxval = np.nanmax(array)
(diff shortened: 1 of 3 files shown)


MaxNoe:
> Checked changes on local master branch, behaves as expected. Should we make sure that mpl_camera behaves the same way when supplied with nan values?

This is not the default and would require creating custom colormaps and converting everything into masked arrays. I don't know, if that is a good idea.

kosack:
>Checked changes on local master branch, behaves as expected. Should we make sure that mpl_camera behaves the same way when supplied with nan values?

I think it already does, but couldn't hurt to add a unit test at least to ensure it doesn't crash

MaxNoe:
> I think it already does, but couldn't hurt to add a unit test at least to ensure it doesn't crash

It does not crash, but it also does not use a special color or transparency by default.

MaxNoe:
Checking this, it does not really handle nans well, I will open a PR.

PR #1045, opened 2019-04-12
Create improved algorithm for pulse time

watsonjj:
We utilise robust peak finding algorithms in the charge extraction. We should use them in the pulse time extraction as well.

This PR creates the `extract_pulse_time_around_peak` numba ufunc, which calculates the weighted average of the pulse within the window defined by the `ImageExtractor`.

### DIFF
--- a/ctapipe/image/extractor.py
+++ b/ctapipe/image/extractor.py
@@ -12,7 +12,7 @@ __all__ = [
     'BaselineSubtractedNeighborPeakWindowSum',
     'extract_charge_from_peakpos_array',
     'neighbor_average_waveform',
-    'extract_pulse_time_around_peakpos',
+    'extract_pulse_time_around_peak',
     'subtract_baseline',
 ]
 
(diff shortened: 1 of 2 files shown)


ctapipe/image/extractor.py
```diff
@@ -124,11 +150,19 @@ def extract_pulse_time_weighted_average(waveforms):
...
+
+    num = 0
+    den = 0
+    for isample in prange(start, end):
+        if 0 <= isample < n_samples:
+            num += waveforms[isample] * isample
+            den += waveforms[isample]
+
+    # TODO: Return pulse time in units of ns instead of isample
+    ret[0] = num / den if den > 0 else peak_index
```
kosack: I think you might be missing a `return` statement here

kosack: that will also fix the errors in Codacy (since it complains about this)

watsonjj: No, this is how `@guvectorize` works, the return value is stored in the `ret` array

watsonjj: Like filling a vector in c++ that was passed to a function by reference

kosack: ah, I see - then we can ignore the codacy errors

PR #1046, opened 2019-04-12
Suggestion: Retire CameraR1Calibrator

watsonjj:
Due to the nature of the r1 calibration being very camera specific, it resulted in the `CameraR1Calibrator` being dependent on the `EventSource` being used. As each camera also has its own file format, it was possible to enable automatic selection of the correct `CameraR1Calibrator` with the `from_eventsource` factory method.

However, if the two are so closely connected, I suggest the `EventSource` is itself responsible for filling the r1 waveforms container. This is more in line with the foreseen operation of the pipeline, as ctapipe would not normally read R0 waveforms, nor would it be responsible for applying the R1 calibration. This also removes the awkward passing of the EventSource to `CameraR1Calibrator` and `CameraCalibrator`.

This is also now a feasible suggestion due to the recent separation of the camera-specific event sources being moved into the plugins.

The r0 container will still be kept, and can be filled as wished, still allowing investigations into R0 waveforms.

I am therefore suggesting retiring the `CameraR1Calibrator` in this PR as a step towards simplifying the calibration chain.


MaxNoe:
Awesome :+1: 

Will have a look later, but this sounds very very reasonable.

kosack:
Sounds good to me as well.

FrancaCassol:
Yes, I approve also this change. Good idea.

watsonjj:
> I noticed nothing in the documentation changed - do you need to update any of the notebooks or top-level docs? Otherwise, this looks fine.

As there will be a lot of changes to the calibration chain as part of #1026, I was hoping to do a single pass through the documentation and notebooks at the end, updating them with the new calibration procedure, and only address errors raised in travis for the meantime.

PR #1048, opened 2019-04-14
Consolidate camera calibration in a single class

watsonjj:
As #1046 was well received, I have proceeded with further simplifying the calibration chain. This PR consolidates the calibration chain into a single class. The `CameraCalibrator` class is repurposed from a "helper" class, into the main calibration chain class.

The `CameraDL0Reducer` and `CameraDL1Calibrator` are deleted. The individual calibration levels are still apparent inside the class (as individual functions), providing structure and direction.

The arguments to `CameraCalibrator` now operate similar to how the `CameraDL1Calibrator` worked, where the classes responsible for each of the calibration steps are individually constructed outside the `CameraCalibrator`, before being passed to the `CameraCalibrator` to be applied in the correct order to fill the event container. If a calibration-step-class is not passed to the `CameraCalibrator` the default classes are used.

By changing to this approach, it makes the process of adding additional steps in the calibration chain more easier and clear. Two additional steps that I intend to be add in future PRs are a gain selector and a charge calibrator (flat-field and absolute calibration), for which additional arguments to the class will be added.

Some TODOs exist in the `CameraCalibrator` indicating changes I intend for future PRs. I wanted to approach this in smaller steps to prevent errors in the results of the calibration chain, and to make reviews more digestible. This PR therefore keeps the calibration operations identical to the existing chain for the time being.


watsonjj:
Also the `calibrate` method was replaced with `__call__` in line with #1024

thomasarmstrong:
Do we no longer need the check_x_exists functionality to prevent containers being overwritten?

watsonjj:
> Do we no longer need the check_x_exists functionality to prevent containers being overwritten?

You are right, I've restored them now

PR #1054, opened 2019-04-24
Cache camera geometries by camera name

MaxNoe:
Just using `lru_cache` is not possible, as numpy arrays and dicts are not hashable.

We could use `cachetools` but here is a solution by hand.

### DIFF
--- a/ctapipe/io/simteleventsource.py
+++ b/ctapipe/io/simteleventsource.py
@@ -21,12 +21,41 @@ from eventio.file_types import is_eventio
 __all__ = ['SimTelEventSource']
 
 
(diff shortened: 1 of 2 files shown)


kosack:
> Just using lru_cache is not possible, as numpy arrays and dicts are not hashable.

Yes, I think that's why I originally added a custom `__hash__` function to CameraGeometry, but it was kind of a hack.

MaxNoe:
> Yes, I think that's why I originally added a custom __hash__ function to CameraGeometry, but it was kind of a hack.

The CameraGeometry **is** hashable, the inputs you need to create one are not. This is a different problem.

PR #1056, opened 2019-04-25
Use sparse matrix to calculate neighbor_matrix_where

MaxNoe:


ctapipe/instrument/camera.py
```diff
@@ -407,7 +407,10 @@ def neighbor_matrix_where(self):
         -------
         ndarray
         """
-        return np.ascontiguousarray(np.array(np.where(self.neighbor_matrix)).T)
+        return np.column_stack([
```
watsonjj: You could store `self.neighbor_matrix_sparse.tocoo()` in a temporary variable instead of calculating twice

watsonjj:
You need to change:
```
@njit([
    float64[:, :, :](float64[:, :, :], int64[:, :], int64),
    float64[:, :, :](float32[:, :, :], int64[:, :], int64),
], parallel=True)
def neighbor_average_waveform(waveforms, neighbors, lwt):
```

to

```
@njit([
    float64[:, :, :](float64[:, :, :], int32[:, :], int64),
    float64[:, :, :](float32[:, :, :], int32[:, :], int64),
], parallel=True)
def neighbor_average_waveform(waveforms, neighbors, lwt):
```
in ctapipe/image/extractor.py

ctapipe/image/extractor.py
```diff
@@ -79,6 +79,7 @@ def sum_samples_around_peak(waveforms, peak_index, width, shift, ret):
 
 @njit([
     float64[:, :, :](float64[:, :, :], int64[:, :], int64),
```
watsonjj: It would be better to switch the signatures to either:

```python
    float64[:, :, :](float64[:, :, :], int32[:, :], int64),
    float64[:, :, :](float32[:, :, :], int32[:, :], int64),
    float64[:, :, :](float64[:, :, :], int64[:, :], int64),
    float64[:, :, :](float32[:, :, :], int64[:, :], int64),
```

Waveforms can be float32 or float64 in my experience

PR #1060, opened 2019-04-25
Fix HDF5TableWriter to ensure user-defined col transforms are applied

kosack:
There were not done *before* constructing the table metadata,
so for columns that were not serializable but where the user added
a proper transform, it failed.

For example, this failed:

```py3
with HDF5TableWriter(filename=outfile, 
                     group_name="dl1_event", 
                     mode="a", 
                     add_prefix=False) as writer:
    
    writer.add_column_transform(
        table_name='sub_trigger', 
        col_name='tels_with_trigger', 
        transform=tel_list_transform
    )
    
    for event in source:
        writer.write('sub_trigger', event.trig)
```

Since the writer would try to generate the table schema before it transformed the value (and in this case tels_with_trigger was not serializable without first transforming it)


kosack:
I think I found one more case where this fails, so don't merge this yet.  Also need to add a unit test to ensure it doesn't regress

kosack:
> Could you add this as a test case then?

Done. there's now a check that the column transform is applied before the table schema is generated.

ctapipe/io/hdf5tableio.py
```diff
@@ -147,32 +141,35 @@ class Schema(tables.IsDescription):
...
 
                 if self._is_column_excluded(table_name, col_name):
-                    self.log.debug(
-                        "excluded column: %s/%s", table_name, col_name
-                    )
+                    self.log.debug("excluded column: %s/%s", table_name, col_name)
                     continue
 
+                # apply any user-defined transforms first
+                value = self._apply_col_transform(table_name, col_name, value)
```
kosack: Note: this line is the main change here, the rest is just re-formatting.

ctapipe/io/tests/test_hdf5.py
```diff
@@ -351,6 +351,32 @@ def create_stream(n_event):
...
+    with HDF5TableWriter(tmp_file, group_name='data') as writer:
+        # add user generated transform for the "value" column
+        cont.value = 6.0
+        writer.add_column_transform('mytable', 'value', my_transform)
+        writer.write('mytable', cont)
+
+    # check that we get a length-3 array when reading back
+    with HDF5TableReader(tmp_file, mode='r') as reader:
+        for data in reader.read("/data/mytable", SomeContainer()):
+            print(data)
```
vuillaut: maybe not necessary print in a test ;-)

PR #1062, opened 2019-04-29
Update eventio to 0.20

MaxNoe:
@kosack can you build the conda packages, then I change back to conda

This brings mainly speed improvements for reading gzipped files and a little improvement for all files.

### DIFF
--- a/ctapipe/io/simteleventsource.py
+++ b/ctapipe/io/simteleventsource.py
@@ -17,6 +17,8 @@ from traitlets import Bool
 
 from eventio.simtel.simtelfile import SimTelFile
 from eventio.file_types import is_eventio
+from gzip import GzipFile
+from io import BufferedReader
 
 __all__ = ['SimTelEventSource']
 
@@ -51,6 +53,10 @@ def build_camera_geometry(cam_settings, telescope):
 
 class SimTelEventSource(EventSource):
     skip_calibration_events = Bool(True, help='Skip calibration events').tag(config=True)
+    zcat = Bool(
+        True,
+        help='Use zcat for faster reading of gzipped files, only forward reading possible'
+    ).tag(config=True)
 
     def __init__(self, config=None, parent=None, **kwargs):
         super().__init__(config=config, parent=parent, **kwargs)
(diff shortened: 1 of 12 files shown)


kosack:
Sure, I'll make new packages (perhaps not today, though)

ctapipe/io/simteleventsource.py
```diff
@@ -51,6 +53,10 @@ def build_camera_geometry(cam_settings, telescope):
 
 class SimTelEventSource(EventSource):
     skip_calibration_events = Bool(True, help='Skip calibration events').tag(config=True)
+    zcat = Bool(
```
dneise: All good .. just one random thought.

is `zcat` maybe a bit too implicit as a parameter name? Could `only_forward_seeking` or `no_back_seek`  or so, be maybe a bit simpler to understand? 
My point is, that `zcat` is somehow an implementation detail ... and this might not be good as a parameter name here...

MaxNoe: Yes, I agree. 

I think `seekable` would be better, and then directly fail if it's not possible.

kosack:
Ok, all packages are there on anaconda cloud, so you should be able to use the conda versions

ctapipe/io/simteleventsource.py
```diff
@@ -64,15 +74,22 @@ def __init__(self, config=None, parent=None, **kwargs):
...
+            raise IOError('back seekable was required but not possible for inputfile')
 
         self._subarray_info = self.prepare_subarray_info(
             self.file_.telescope_descriptions,
             self.file_.header
         )
         self.start_pos = self.file_.tell()
 
+    @property
+    def is_stream(self):
```
kosack: should this be a general property of `EventSource` rather than `SimTelEventSource`? (same comment also for the `back_seekable` trait)

MaxNoe: It is. This is inherited from EventSource but overriden

MaxNoe:
@kosack the pip installs were still in place

Issue #1041, opened 2019-04-08
Discussion on separation of Event, Monitoring, and Service/header data


kosack:
The Problem:
-------------
The concept of an `EventSource` has been quite successful and easy to use in ctapipe, since it allows one to write clear loops over events, where the event data is part of an iterable object. However, we now need to think about other data streams that are not tied directly to each Cherenkov event.  

When processing data, we don't only need the data for an event, but several other pieces of information

* **Monitoring data**, which includes things like the pointing position of each telescope, or various calibration coefficients that change on ~ 1s timescales.
* **Service Data**
  * **Instrument description**, which is fixed during an observation (e.g. the `SubarrayDescription`)
  * **Observation configuration**: the "run header" information associated with an observation 
  * **Simulation Configuration**: the "mc header" information associated with an observation that comes from a Monte-Carlo simulation, which defines the monte-carlo parameters

These info come at different rates than the events (much more slowly).  However, some form of synchronization to the event data is needed.

###  for Monitoring data

The monitoring (MON) data comes rates around 1HZ, but it is not necessarily correct to simply synchronize the "latest" MON data with the current event: normally one would like to do some interpolation, so how the monitoring data is accessed may need some thought.  This means there may be two types of MON data: 

* the "raw" MON data that is just a measurement each second
* interpolated MON data that has the correct value at the exact instance of the event time. 

These do not need to be stored in the same way, e.g. one might first read the raw MON data, construct an interpolator, and then use it to fill `event.mon` info (event-wise interpolated MON data)

Note also that there are also two further types of MON data:
* MON  data read from a pre-existing file (which could be simply loaded, and an interpolator function constructed, so that one can generate per-event interpolations)
* MON data generated during the event loop (e.g. pedestals), so there is no pre-existing file to read and interpolate -- must be done on-the-fly

### For Service Data:

In general these are all static information for a given observation, but in the case of Monte-Carlo files, it *is possible* for these to change in the middle of a file (e.g. if multiple runs are merged).  The question is whether or not we support these changing for a single `EventSource` loop, or assume they are header info that is read once at the start. 

### Parallelization and distributed processing

Right now parallelizing the event loop (e.g. `some_parallel_system.map(processing_function, source)`) means sending around all of the event, mon, and service data every time an event changes (even though the mon and service info doesn't change).  This is probably very inefficient, since that is a lot of data to transmit.  Separating the streams makes this much cleaner.

# Implementation Options

[...]

PR #1067, opened 2019-05-03
Camera calibration code

FrancaCassol:
I propose in this PR a fist version of the monitoring containers related to the camera calibration and
a first version for the pedestal and flat-field Components (they were already partly reviewed in a previous removed PR [https://github.com/cta-observatory/ctapipe/pull/870](https://github.com/cta-observatory/ctapipe/pull/870) )

The new Monitoring containers are: 

```
FlatFieldContainer
PedestalContainer
PixelStatusContainer
WaveformCalibrationContainer
MonitoringCameraContainer
MonitoringContainer
```
In order to handle them before a final agreement on the data model  (see [https://github.com/cta-observatory/ctapipe/issues/1041](https://github.com/cta-observatory/ctapipe/issues/1041) ), I introduced also a EventAndMonDataContainer which joins the monitoring container to the event container (and is probably temporary)

```
class EventAndMonDataContainer(DataContainer):
    """
    Data container including monitoring information
    """
    mon = Field(MonitoringContainer(), "container for monitoring data (MON)")

```

The flat-field and pedestal Components are placed in the camera/calib directory and are meant to fill the FlatFieldContainer and PedestalContainer container


ctapipe/io/containers.py
```diff
@@ -611,3 +618,189 @@ class TimingParametersContainer(Container):
...
+    Container for the pixel calibration coefficients
+    """
+    time = Field(0, 'Time associated to the calibration event', unit=u.s)
+    time_range = Field(
+        [],
+        'Range of time of validity for the calibration event [t_min, t_max]',
+        unit=u.s
+    )
+
+    dc_to_phe = Field(
```
kosack: perhaps rename to dc_to_pe since use "PE" for the abbreviation for photo-electrons everywhere else

ctapipe/io/containers.py
```diff
@@ -611,3 +618,189 @@ class TimingParametersContainer(Container):
...
+    )
+
+    n_phe = Field(
+        None,
+        "np array of photo-electrons in calibration signal (n_chan, n_pix)"
+    )
+
+    pixel_status_mask = Field(
+        None,
+        "Boolean np array (mask) of final calibration data analysis (n_chan, n_pix)"
```
kosack: I don't understand what this is from the description. Is it pixels that have a correct calibration? Does True mean "good"?

ctapipe/io/containers.py
```diff
@@ -611,3 +618,189 @@ class TimingParametersContainer(Container):
...
+
+class WaveformCalibrationContainer(Container):
+    """
+    Container for the pixel calibration coefficients
+    """
+    time = Field(0, 'Time associated to the calibration event', unit=u.s)
+    time_range = Field(
+        [],
+        'Range of time of validity for the calibration event [t_min, t_max]',
+        unit=u.s
```
kosack: How is the time defined here? I guess seconds since the start of the observation? Should we make that an absolute time?

ctapipe/io/containers.py
FrancaCassol: ok

ctapipe/io/containers.py
FrancaCassol: No, True are the pixels that must be masked (not good pixels). 
This makes things easier if one want to use the numpy.ma module. Is this ok? 
But I agree, this mask stuff is often ambiguous. I will write  more clearly in the description

MaxNoe: Please try to also find a better variable name, not only a better description.

e.g. `unusable_pixels`

ctapipe/io/containers.py
FrancaCassol: This is for the moment the event.r1.tel[tel_id].trigger_time 
If this must be used as timestamp for retrieving calibration data, it should be indeed better an absolute time (or at least the time reference should be easily accessible)

thomasgas:
I would propose to add more docstrings to the functions since some are missing.

ctapipe/io/containers.py
FrancaCassol: Ok, done

FrancaCassol:
Hi, 
Could this PR be merged?

After that my plan is:
- update  ctapipe_io_nectarcam  so as to initialise the monitoring container 
- provide to git a NectarCam pedestal and flat-field data file (**where should be put it in order that git finds it**? I would like to use it also for the tests)
- Perform an other PR with  calibration tool examples and notebooks that run on the above real calibration data

- update ctapipe_io_lst taking out the monitoring containers that will then be in ctapipe
- update cta-lstchain taking out the general calibration code that will then be in ctapipe

FrancaCassol:
Hi @kosack, @MaxNoe @dneise,

are there other comments? Otherwise, could you please merge this PR so that I go ahead with the developments described above? 

Could you also please reply to the question concerning the data for testing the code? Where should I put them?

Thanks!

Issue #1059, opened 2019-04-25
DL1 data file layout discussion


kosack:
Following the recent discussion in the ASWG, I've created a small notebook that uses the ctapipe functionality to write out a sample DL1 data file that contains only the DL1.TEL.EVT.IMAGE  data (e.g. no cleaning or parameterization, which would be DL1.TEL.EVT.PARAM).

It also stores the instrument and simulation metadata, and some DL1.SUB.EVT data (the trigger pattern for now, but you could imagine this could also include an index map table).    We can use this issue to discuss the hierarchy. For now, I tried to keep it not too complex (no deep sub-datasets), and I just use the default ctapipe containers, so didn't add any extra information. 

For all tables, some extra custom metadata are stored, like the unit of each column and the ctapipe version. 

I also attached the MC event info to all camera tables (it's repetitive, but convenient). That could also be stored in a separate table.

* The data file can be found here (private to CTA members):
- v1: https://indico.cta-observatory.org/event/2265/contributions/19711/attachments/15269/19392/example_dl1.h5
- v3: https://indico.cta-observatory.org/event/2265/contributions/19711/attachments/15269/19424/example_dl1_v3.h5
- v4: https://indico.cta-observatory.org/event/2265/contributions/19711/attachments/15269/19444/example_dl1_v4.h5
- v5: https://indico.cta-observatory.org/event/2265/contributions/19711/attachments/15269/19471/example_dl1_v5.h5

* and the notebook I used is here:
https://gist.github.com/kosack/7cbbe848ab4840893810c4fa3d75be9a

* The output file looks like this:

![image](https://user-images.githubusercontent.com/11677812/57465462-f1d19200-727e-11e9-96f6-4ed11bc89afc.png)

kosack:
just noticed I'm missing the tel_id in the image tables, but that's easy to add (not so useful without it!)

kosack:
The easiest way to explore the file is with pytables probably (since the vector columns cause pandas to complain):

```py3
>>> import tables
>>> t = tables.open_file("example_dl1.h5")       
>>> t.root.dl1_event.tel_image_GCT_CHEC

/dl1_event/tel_image_GCT_CHEC (Table(7,)) 'Storage of DL0Container,DL1CameraContainer,MCEventContainer'
  description := {
  "dl1camera_image": Float64Col(shape=(1, 2048), dflt=0.0, pos=0),
[...]

PR #1068, opened 2019-05-09
Allow tables in arbitrarily deep  hdf5 groups

kosack:
This is a change to `HDF5TableWriter` to allow tables to be output to arbitrarily deep groups in an HDF5 file. 

The main change (other than some reformatting) is:
- the `group_name` attribute of the HDF5TableWriter is now considered the "base" group (from which all sub-tables will be under)
- `writer.write(table_name="x", containers=y)` works as before, but now you can add deeper nesting like `writer.write(table_name="subgroup/anothergroup/x", containers=y)` and *subgroup* and *anothergroup* will be created automatically if they don't exist. 
- the unit test `test_write_to_any_location` has been re-enabled (it was marked as *xfail* before)

A full example:

```py3
with HDF5TableWriter("output.h5", group_name="/dl1/tel") as writer:
    for event in source:
         ...
         writer.write("image/NectarCam", somedata) 
         writer.write("params", paramdata)
```
Would create 2 tables: `/dl1/tel/image/NectarCam` and `/dl1/tel/params`

This allow a  nicer structure, as shown  in #1059  (v5 and above)

### DIFF
--- a/ctapipe/io/hdf5tableio.py
+++ b/ctapipe/io/hdf5tableio.py
@@ -1,6 +1,7 @@
 """Implementations of TableWriter and -Reader for HDF5 files"""
 import enum
 from functools import partial
+from pathlib import PurePath
 
 import numpy as np
 import tables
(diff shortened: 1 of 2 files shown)


vuillaut:
Definitely something we want.

kosack:
I think this is ready to go - codacy complains a bit about some test variable names, but I think those can be ignored for now.

PR #1069, opened 2019-05-10
Fix broken telescope id string, Fixes #1058

### LINKED ISSUE #1058: TelescopeDescription string representation changed
Old behavior:
```py3
>>> str(sub.tel[410])
'SST-1m:Digicam'
>>> str(sub[tel][100])
'SST-ASTRI:AstriCam'
```

Current Behavior:
```py3
>>> str(sub.tel[410])
'1m:DigiCam'
>>> str(sub[tel][100])
'ASTRI:AstriCam'
```

Seems the telescope type is no longer prepended. Since the full telescope identifier is used for output table naming, this causes some problems since a variable name that starts with a number (like `1m`) is forbidden.  Could we go back to the old behavior? Any objections?

kosack:
- telescope names are now normalized to `TYPE_OPTICS_CAMERA` (usable as an output file or group name)
- `SubarrayDescription.get_tel_ids_for_type()` was broken due to the previous name change, this fixes it
- added a lot of unit tests to prevent regression
- added a function and property to map telescope *IDs* to telescope *indices* (useful for creating things like trigger patterns or packing telescope data)

### DIFF
--- a/ctapipe/instrument/telescope.py
+++ b/ctapipe/instrument/telescope.py
@@ -13,9 +13,9 @@ Todo:
   telescope :-))
 
 """
-from .optics import OpticsDescription
 from .camera import CameraGeometry
-from .guess import guess_telescope
+from .guess import UNKNOWN_TELESCOPE, guess_telescope
+from .optics import OpticsDescription
 
 
 class TelescopeDescription:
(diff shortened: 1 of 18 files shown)


ctapipe/instrument/telescope.py
```diff
@@ -85,18 +85,25 @@ def from_name(cls, optics_name, camera_name):
         camera = CameraGeometry.from_name(camera_name)
         optics = OpticsDescription.from_name(optics_name)
 
-        t = guess_telescope(camera.n_pixels, optics.equivalent_focal_length)
```
MaxNoe:
```
try:
    t = guess_telescope(camera.n_pixels, optics.equivalent_focal_length)
except ValueError:
    t = UNKNOWN_TELESCOPE

```

kosack: yes, that's better. Done.

PR #1070, opened 2019-05-10
Hipehdf5

devine73:
As announced in the ASWG meeting on Thursday 9th, the HDF5 reader for the HDF5 data format from the LAPP for R1 (direct conversion from simtel file).
To contact the author:
      pierre.aubert@lapp.in2p3.fr

### DIFF
--- a/ctapipe/io/simteleventsource.py
+++ b/ctapipe/io/simteleventsource.py
@@ -75,7 +75,7 @@ class SimTelEventSource(EventSource):
             self.input_url,
             allowed_telescopes=self.allowed_tels if self.allowed_tels else None,
             skip_calibration=self.skip_calibration_events,
-            zcat=not self.back_seekable,
+            #zcat=not self.back_seekable,
         )
         if self.back_seekable and self.is_stream:
             raise IOError('back seekable was required but not possible for inputfile')
--- a/ctapipe/io/__init__.py
+++ b/ctapipe/io/__init__.py
@@ -2,6 +2,7 @@ from .array import get_array_layout
 from .eventseeker import EventSeeker
 from .eventsource import EventSource, event_source
 from .hdf5tableio import HDF5TableReader, HDF5TableWriter
+from .hipehdf5eventsource import HiPeHDF5EventSource
 from .tableio import TableWriter, TableReader
 
 # import event sources to make them visible to EventSource.from_url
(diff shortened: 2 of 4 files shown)


ctapipe/io/simteleventsource.py
```diff
@@ -75,7 +75,7 @@ def __init__(self, config=None, parent=None, **kwargs):
             self.input_url,
             allowed_telescopes=self.allowed_tels if self.allowed_tels else None,
             skip_calibration=self.skip_calibration_events,
-            zcat=not self.back_seekable,
+            #zcat=not self.back_seekable,
```
vuillaut: why commenting this line?

MaxNoe: yes, this should not be changed. First: it's completely unrelated and second: why?

devine73: I commented this line to ensure the tests passed on my computer.

MaxNoe: You dont Change unrelated code to make tests pass. This is what tests are for!

kosack:
Event sources other than SimTelEventSource are now in separate repos to avoid complicating ctapipe with non-standard formats, could we put this in it's own repo as a plugin (similar to ctapipe_io_lst, ctapipe_io_magic, ctapipe_io_nectarcar, etc)?

kosack:
Also, it may be better to wait until we define a common DL1 format - we don't want to have more than one of them in the end! If this format has nice features, I would suggest please add comments to the issue #1059 or a new issue to explain what should be added to the common format.

MaxNoe:
Isn't this R0/DL0?

kosack:
> Isn't this R0/DL0?

Monte-Carlo doesn't really fit yet with the final data levels - I guess it's more R0 than R1/DL0 though.

ctapipe/io/simteleventsource.py
kosack: the tests for that work fine on my machine (and on travis), so it shouldn't be removed

MaxNoe: This hints to an outdated eventio version. Use `python setup.py test` or do `pip install -e .` again, then it should upgrade eventio.

devine73:
Ok, I will make a plugin. Do you want something like ctapipe_io_R0 or maybe ctapipe_io_MCR0 to express this is a Monte-Carlo data format ? Or other names...

kosack:
This code is moving to a plugin, so I'll close this PR for now

Issue #968, opened 2019-02-15
Camera frame transformations


watsonjj:
When plotting star positions on the camera (transforming from sky coordinates to the camera frame), the positions do not appear where one would expect for different telescope types:

LST
![image](https://user-images.githubusercontent.com/17825673/52850908-02a5b800-3115-11e9-91f3-3439b8fd3975.png)

CHEC
![image](https://user-images.githubusercontent.com/17825673/52850937-1bae6900-3115-11e9-81af-f2b0f61a0cfd.png)

I would expect to see a mirroring in the second image with respect the the first, due to the 2M telescope.

As I understand now (from talking to @ParsonsRD), the camera frame is not the actual coordinates you expect to have when physically looking at the face of the camera, but it is instead a "MCCameraFrame". This frame has already taken into account this reflection in the simulations, and therefore does not need to be handled differently in the analysis.

However, when handling observations from real camera data, this approach is not intuitive.

@ParsonsRD has suggested to include an additional camera frame which represents what you would see if you look at the face of the camera:

CameraFrame (New frame which represents what you see when you look at the camera face. Data read from real telescopes are associated with this frame)
    &#8595;
MCCameraFrame (Currently existing CameraFrame. Data read from simulations are associated with this frame)
    &#8595;
TelescopeFrame, etc.

watsonjj:
Code snippet for the above plots:

```python
import matplotlib.pyplot as plt
from astropy.time import Time
from astropy.coordinates import SkyCoord, EarthLocation
import astropy.units as u
from ctapipe.coordinates import CameraFrame, HorizonFrame
from ctapipe.visualization import CameraDisplay
from ctapipe.instrument import CameraGeometry


location = EarthLocation.of_site('Roque de los Muchachos')
obstime = Time('2018-11-01T04:00')

crab = SkyCoord.from_name("crab nebula")

horizon_frame = HorizonFrame(location=location, obstime=obstime)
pointing = crab.transform_to(horizon_frame)
camera_frame = CameraFrame(
    telescope_pointing=pointing,
    focal_length=u.Quantity(2.15, u.m),#28 for LST
    obstime=obstime,
    location=location,
)


# cam = CameraGeometry.from_name('LSTCam')
cam = CameraGeometry.from_name('CHEC')
fig, ax = plt.subplots()
display = CameraDisplay(cam, ax=ax)

ax.set_title(
[...]

PR #1085, opened 2019-06-06
Add EngineeringCameraFrame

MaxNoe:
Question: Do we need a case separation for 1 / 2 mirror telescopes?

### DIFF
--- a/ctapipe/visualization/mpl_camera.py
+++ b/ctapipe/visualization/mpl_camera.py
@@ -7,11 +7,14 @@ import logging
 
 import numpy as np
 from astropy import units as u
+from astropy.coordinates import SkyCoord
 from matplotlib import pyplot as plt
 from matplotlib.collections import PatchCollection
 from matplotlib.colors import Normalize, LogNorm, SymLogNorm
 from matplotlib.patches import Ellipse, RegularPolygon, Rectangle
 from numpy import sqrt
+
+from ..coordinates import CameraFrame, EngineeringCameraFrame
 
 __all__ = ['CameraDisplay']
 
@@ -56,6 +59,8 @@ class CameraDisplay:
         rescale the vmin/vmax values when the image changes.
         This is set to False if `set_limits_*` is called to explicity
         set data limits.
+    engineering_frame : bool (default False)
+        Transform camera coordinates into the engineering frame
 
     Notes
     -----
@@ -94,7 +99,8 @@ class CameraDisplay:
             cmap=None,
             allow_pick=False,
             autoupdate=True,
-            autoscale=True
+            autoscale=True,
+            engineering_frame=False,
     ):
         self.axes = ax if ax is not None else plt.gca()
         self.geom = geometry
@@ -117,25 +123,32 @@ class CameraDisplay:
         if not hasattr(self.geom, "mask"):
             self.geom.mask = np.ones_like(self.geom.pix_x.value, dtype=bool)
 
-        for xx, yy, aa in zip(
-                u.Quantity(self.geom.pix_x[self.geom.mask]).value,
-                u.Quantity(self.geom.pix_y[self.geom.mask]).value,
-                u.Quantity(np.array(self.geom.pix_area)[self.geom.mask]).value):
+        pix_x = self.geom.pix_x[self.geom.mask]
+        pix_y = self.geom.pix_y[self.geom.mask]
+        pix_area = self.geom.pix_area[self.geom.mask]
+        pix_rotation = self.geom.pix_rotation
+        if engineering_frame:
+            from_frame = CameraFrame()
+            coord = SkyCoord(x=pix_x, y=pix_y, frame=from_frame)
+            trans = coord.transform_to(EngineeringCameraFrame())
+            pix_x, pix_y = trans.x, trans.y
+            pix_rotation = 90 * u.deg - pix_rotation
 
+        for x, y, area in zip(pix_x.value, pix_y.value, pix_area.value):
             if self.geom.pix_type.startswith("hex"):
-                rr = sqrt(aa * 2 / 3 / sqrt(3)) + 2 * PIXEL_EPSILON
+                r = sqrt(area * 2 / 3 / sqrt(3)) + 2 * PIXEL_EPSILON
                 poly = RegularPolygon(
-                    (xx, yy), 6, radius=rr,
-                    orientation=self.geom.pix_rotation.rad,
+                    (x, y), 6, radius=r,
+                    orientation=pix_rotation.to_value(u.rad),
                     fill=True,
                 )
             else:
-                rr = sqrt(aa) + PIXEL_EPSILON
+                r = sqrt(area) + PIXEL_EPSILON
                 poly = Rectangle(
-                    (xx - rr / 2., yy - rr / 2.),
-                    width=rr,
-                    height=rr,
-                    angle=self.geom.pix_rotation.deg,
+                    (x - r / 2, y - r / 2),
+                    width=r,
+                    height=r,
+                    angle=pix_rotation.to_value(u.deg),
                     fill=True,
                 )
 
--- a/ctapipe/visualization/mpl_camera.py
+++ b/ctapipe/visualization/mpl_camera.py
@@ -7,7 +7,6 @@ import logging
 
 import numpy as np
 from astropy import units as u
-from astropy.coordinates import SkyCoord
 from matplotlib import pyplot as plt
 from matplotlib.collections import PatchCollection
 from matplotlib.colors import Normalize, LogNorm, SymLogNorm
--- a/ctapipe/visualization/mpl_camera.py
+++ b/ctapipe/visualization/mpl_camera.py
@@ -14,8 +14,6 @@ from matplotlib.colors import Normalize, LogNorm, SymLogNorm
 from matplotlib.patches import Ellipse, RegularPolygon, Rectangle
 from numpy import sqrt
 
-from ..coordinates import CameraFrame, EngineeringCameraFrame
-
 __all__ = ['CameraDisplay']
 
 logger = logging.getLogger(__name__)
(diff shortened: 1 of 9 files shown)


watsonjj:
This is connected to the proposal in #968, correct?

MaxNoe:
Yes, I will make a follow up PR introducing camera frame definition/selection to the fits files and maybe to the plotting.

The idea is to transform into the internal frame on reading and maybe back on display if wished.

MaxNoe:
camera_frames



```python
from ctapipe.instrument import CameraGeometry
from ctapipe.coordinates import EngineeringCameraFrame
import astropy.units as u


fig, axs = plt.subplots(2, 2, constrained_layout=True, figsize=(6, 3))

axs = axs.ravel()

model = Gaussian(0 * u.m, 0.1 * u.m, 0.1 * u.m, 0.03 * u.m, 0 * u.deg)

cam = CameraGeometry.from_name('FlashCam')
image, *_ = model.generate_image(cam, 1500)

CameraDisplay(cam, ax=axs[0], image=image)
CameraDisplay(
    cam,
    ax=axs[1],
    frame=EngineeringCameraFrame(),
    image=image,
)

cam = CameraGeometry.from_name('LSTCam')
image, *_ = model.generate_image(cam, 1500)

CameraDisplay(cam, ax=axs[2], image=image)
CameraDisplay(
    cam,
    ax=axs[3],
    frame=EngineeringCameraFrame(),
    image=image,
)

axs[0].set_title('CameraFrame')
axs[1].set_title('EngineeringCameraFrame')


plt.show()
```

kosack:
This is great! (been needed for a while due to the confusing Monte-Carlo system) 

Be sure to document whether the engineering frame is looking at the back of the camera or the front. Also, perhaps we should  draw a "North" arrow on the CameraDisplay, and even maybe some text like "Front" or "Back" (or a vector "into the plane" or "out-of the plane" symbol).  That would avoid a lot of confusion.

kosack:
possibly also related to #255

kosack:
and #875

MaxNoe:
The question if it's the back or front is important for dual mirror. This is not yet taken into account here!

So we should ask, what we want for this, I guess front?

watsonjj:
In CHEC, our (engineering) mapping is defined as if we are looking at the camera from the front

ctapipe/visualization/mpl_camera.py
```diff
@@ -117,25 +118,25 @@ def __init__(
...
-                u.Quantity(self.geom.pix_y[self.geom.mask]).value,
-                u.Quantity(np.array(self.geom.pix_area)[self.geom.mask]).value):
+        pix_x = self.geom.pix_x.value[self.geom.mask]
+        pix_y = self.geom.pix_y.value[self.geom.mask]
+        pix_area = self.geom.pix_area.value[self.geom.mask]
 
+        for x, y, area in zip(pix_x, pix_y, pix_area):
             if self.geom.pix_type.startswith("hex"):
-                rr = sqrt(aa * 2 / 3 / sqrt(3)) + 2 * PIXEL_EPSILON
+                r = sqrt(area * 2 / 3 / sqrt(3)) + 2 * PIXEL_EPSILON
```
kosack: these double names were just because single-character variable names were forbidden in the style guide since they are difficult to search for in a text editor. But I guess if we use modern IDEs with true syntax support (rather than search/replace), it's ok.

docs/ctapipe_api/coordinates/index.rst
```diff
@@ -35,6 +36,13 @@ The following special frames are defined for CTA:
 they can be transformed to and from any other `astropy.coordinates` frame, like
 `astropy.coordinates.AltAz` or `astropy.coordinates.ICRS` (RA/Dec)
 
+The two different coordinate frames are shown here:
+
+.. plot:: ../examples/plot_camera_frames.py
+    :include-source:
+
+The `CameraFrame` is used internally and is used by H.E.S.S. and `sim_telarray`,
+the `EngineeringCameraFrame` is used by `MAGIC` and `FACT`.
```
kosack: Actually, in HESS we also use the EngineeringFrame.  The HESS analysis framework (SASH) is independent of simtelarray and it's built-in analysis (read_hess).  I think the SMASH simulations also use the EngineeringFrame, but I'd have to check to be sure.  Might be better to say the CameraFrame comes from the sim_telarray conventions, which in turn come from the conventions in Corsika.

kosack:
> draw a "North" arrow on the CameraDisplay
Actually the more I think of it this is also dependent on the pointing direction of the telescope, so it's not so easy... I guess in the EngineeringFrame we assume the telescope is parked, so "Up" might be more useful?

MaxNoe:
I verified this for single mirror telescopes using FACT and pedestals of stars we took some time ago in 2015, the positions are queried from the yale bright star catalogue (V/50 in Vizier):

fact_star_pedestal

```python
from ctapipe.instrument import CameraGeometry
from ctapipe.coordinates import EngineeringCameraFrame, CameraFrame
import astropy.units as u
import numpy as np
from astroquery.vizier import Vizier
from astropy.coordinates import SkyCoord, AltAz
from fact.instrument.constants import LOCATION, FOCAL_LENGTH_MM
from astropy.time import Time

print('done')

vizier = Vizier(
    catalog='V/50',
    columns=['_RAJ2000', '_DEJ2000', 'Vmag', 'Name']
)
t = vizier.query_region('Aldebaran', radius=2.5 * u.deg, catalog='V/50')[0]
stars = SkyCoord(ra=t['_RAJ2000'], dec=t['_DEJ2000'], frame='icrs')

obstime = Time('2015-11-16T00:35:11')
altaz = AltAz(obstime=obstime, location=LOCATION)
pointing = SkyCoord.from_name('Aldebaran').transform_to(altaz)

camera_frame = CameraFrame(
    location=LOCATION,
    obstime=obstime,
    focal_length=FOCAL_LENGTH_MM * u.mm,
    telescope_pointing=pointing,
)
engineering_frame = EngineeringCameraFrame(
    location=LOCATION,
    obstime=obstime,
    focal_length=FOCAL_LENGTH_MM * u.mm,
    telescope_pointing=pointing,
)

stars_cam = stars.transform_to(camera_frame)
stars_eng = stars.transform_to(engineering_frame)

print(stars_cam.x)
print(stars_eng.x)


fig, axs = plt.subplots(1, 2, constrained_layout=True, figsize=(6, 3))

cam = CameraGeometry.from_name('FACT')
image = np.genfromtxt('./aldebaran_pedvar.txt')

d1 = CameraDisplay(cam, ax=axs[0], image=image, cmap='inferno')
d2 = CameraDisplay(
    cam.transform_to(engineering_frame),
    ax=axs[1],
    image=image,
    cmap='inferno',
)

axs[0].plot(stars_cam.x.value, stars_cam.y.value, 'wo', mfc='none', ms=25, mew=2)
axs[1].plot(stars_eng.x.value, stars_eng.y.value, 'wo', mfc='none', ms=25, mew=2)

d1.set_limits_minmax(0, 0.3 * image.max())
d2.set_limits_minmax(0, 0.3 * image.max())

axs[0].set_title('CameraFrame')
axs[1].set_title('EngineeringCameraFrame')

plt.show()
```

docs/ctapipe_api/coordinates/index.rst
MaxNoe: Okay, then I think de should also discuss if es really want to call it `CameraFrame` and `EngineeringCameraFrame` and which One de really want to use as internal discription.

watsonjj:
I have had a play around with this PR, along with star field data (Draco constellation) that was recently taken with CHEC. I achieved the following:

image

The notebook for this can be found at: 
https://github.com/watsonjj/CHECLabPySB/blob/master/CHECLabPySB/d190607_pointing/engineering.ipynb

CHEC's engineering camera frame is defined by looking at the camera from the secondary mirror. The top of the camera is pointed to the top of the telescope in park position. To convert from this frame to the ctapipe `CameraFrame`, one must rotate 90 degrees anti-clockwise.

I achieved the frame transformations with the following code:
```python
from ctapipe.coordinates.camera_frame import CameraFrame, CartesianRepresentation, AffineTransform, frame_transform_graph

CAMERA_TO_ENGINEERING2M_MATRIX = np.array([
    [0, 1, 0],
    [-1, 0, 0],
    [0, 0, 1]
])
ENGINEERING2M_FRAME_TRANSFORM_OFFSET = CartesianRepresentation(0, 0, 0, unit=u.m)

ENGINEERING2M_TO_CAMERA_MATRIX = np.array([
    [0, -1, 0],
    [1, 0, 0],
    [0, 0, 1]
])


class Engineering2MCameraFrame(CameraFrame):
    pass

@frame_transform_graph.transform(AffineTransform, CameraFrame, Engineering2MCameraFrame)
def camera_to_engineering2M(from_coord, to_frame):
    return CAMERA_TO_ENGINEERING2M_MATRIX, ENGINEERING2M_FRAME_TRANSFORM_OFFSET

@frame_transform_graph.transform(AffineTransform, Engineering2MCameraFrame, CameraFrame)
def engineering2M_to_camera(from_coord, to_frame):
    return ENGINEERING2M_TO_CAMERA_MATRIX, ENGINEERING2M_FRAME_TRANSFORM_OFFSET
```

As you can see, two different matrices are required from the transformation back and forth in this case.

My solution was to define a new camera frame `Engineering2MCameraFrame`...however it would be nicer to avoid having two different EngineeringCameraFrames for the two types of telescopes... any ideas?

MaxNoe:
That should be easy. Add a frame attribute `n_mirrors` and return the one or the other matrix dependent on it.

docs/ctapipe_api/coordinates/index.rst
kosack: I think as long as we document what each means, it's fine to keep it as is.  Engineering makes the most sense for a frame where the telescope is parked and where somebody might be working on the camera, to replace a module or something.

MaxNoe:
I added the dual mirror case, @watsonjj can you check if it now works with this PR out of the box using `EngineeringFrame(n_mirros=2)`?

watsonjj:
I can confirm the dual mirror case works:
https://github.com/watsonjj/CHECLabPySB/blob/master/CHECLabPySB/d190607_pointing/engineeringv2.ipynb

PR #1092, opened 2019-06-19
relax absolute version requirement on eventio==0.11.0

kosack:


kosack:
this is needed to build a new conda package for v0.6.2 that doesn't require eventio==0.11.0 (and this can read recent montecarlo)

MaxNoe:
I think this should be changed to `eventio >=0.11, <= 0.17.0` as that version made a breaking change for ctapipe, see https://github.com/cta-observatory/ctapipe/pull/966/files

kosack:
can 0.17.0 read all the latest files? The main reason for this was to allow 0.6.2 to read some Prod3 files that 0.11 couldn't read (I can't remember which ones it couldn't read and for what reason though).  Also, could you switch to semantic versioning, so we can tell from the version if the API breaks?  perhaps it will just be easier to release ctapipe-0.7, but that has API changes...

MaxNoe:
0.17.0 is the version with the breaking change.

0.16.2 is the latest version that is compatible to ctapipe 0.6.2 and implemented fixes for reading certain types of frankenstein files and there was support for different new containers added in between 0.11 and 0.16.2

MaxNoe:
>  Also, could you switch to semantic versioning, so we can tell from the version if the API breaks? 

We already do, but as the version was below 1.0, the breakage indicator is one to the right.

The problem here is, we made a breaking change and implemented new functionality.
And some people now want the new functionality without the breaking change.
This is however not easy without maintaining multiple release branches, which is a burden we do not want to have I think.

kosack:
> This is however not easy without maintaining multiple release branches, which is a burden we do not want to have I think.

Agreed for now at least. I don't have time to manage it.  So perhaps I just say wait until v0.7 and deal with the API change.

MaxNoe:
I think you misunderstood me. I think there is no reason not to release a ctapipe `0.6.3` with the dependency of `eventio >=0.11,<0.17`

MaxNoe:
But I think you just need to create a new branch from the `v0.6.2` release and call it `v0.6.x` or similar.

kosack:
> branch from the v0.6.2

I think github  created a branch automatically from v0.6.2 tag (now there is both a tag and branch with the same name, which is a bit confusing), so I just need to figure out how to deal with that.  Otherwise it's ok.  I think I might still close this, since the problem we had with reading data with eventio-0.11 seems not to be a large issue. I'll wait a bit to see...

MaxNoe:
I think it was not possible to read SCT data because of the number of pixels it used a different version of the camera organization type, whihc was not yet implemented.

kosack:
I'm closing this since it will be fixed in the next release.

Issue #962, opened 2019-02-11
About GainSelector class and gain_channel thresholds


misabelber:
Hello,
I've been trying to use the GainSelectorThresholds class from ctapipe.camera.calib and I found a couple of troubles that can make it a little inconvenient to use(or maybe I'm doing something worng, in that case I will be thankful to learn the proper way). 
The first thing is that as it is defined, it searches for samples in the waveforms over a certain threshold in PHE, but waveforms are given in dc counts, so if  you use the waveforms directly from the events (in simtel files, for example), it will give a wrong result.
 The default values  of the thresholds are read from ctapipe-extra/ctapipe_resources/gain_channel_thresholds.ecsv (and are in phe), and perhaps it would be better to have them in dc counts (or add a conversion factor for each camera definition). For example, for LSTCam 100 phe --> 4095 dc counts. 
The other issue is that the function pick_gain_channel searches for high gain waveform values purely larger than the threshold (>), but this will never happen, since samples will saturate at that exact value. The condition should be substitued by >=.
 Thank you very much!

maxnoe:
I think this issue is fixed since a long time, please reopen if this still needs to be addressed.

Issue #822, opened 2018-11-19
Why is image two dimensional?


MaxNoe:
the `DL1CameraContainer.image` is a 2d array, that can contain multiple images (e.g. for ASTRI).

Why is that? Shouldn't be `DL1` be always a single image?

watsonjj:
A dimension per channel (hi gain, low gain). Should become a 1D array when gain selection has been implemented (see #683).

watsonjj:
Solved by #1095

Issue #818, opened 2018-11-16
mapping for gain-channels?


dneise:
Would it be good to define the mapping for the gain channels in `waveforms`? 
I.e. something like `{'low-gain': 0, 'high-gain': 1}` ?

MaxNoe:
@kosack proposed using an enum here:

https://github.com/cta-observatory/ctapipe/pull/683/files#diff-722d37cc1260e3557284e9056a2d9f3cR18

watsonjj:
Solved by #1093

PR #1093, opened 2019-06-20
Refactoring of GainSelector

watsonjj:
This PR is for the refactoring and discussion of the existing (but mostly unused) GainSelector classes.

My next PR will be dedicated to the integration of this class in the calibration chain.

I have simplified the GainSelector class in a similar way as was recently performed for the ImageExtractor class in design. It acts as a low-level class which operates on the waveforms numpy array.

This PR picks up from #683, and includes the suggestion from #818. Other related issues include: #419, #962 and #822.

### DIFF
--- a/ctapipe/calib/camera/gainselection.py
+++ b/ctapipe/calib/camera/gainselection.py
@@ -15,6 +15,9 @@ __all__ = [
 
 
 class GainChannel(IntEnum):
+    """
+    Possible gain channels
+    """
     HIGH = 0
     LOW = 1
 
@@ -44,7 +47,7 @@ class GainSelector(Component):
         if waveforms.ndim == 2:  # Return if already gain selected
             return waveforms
         elif waveforms.ndim == 3:
-            n_channels, n_pixels, n_samples = waveforms.shape
+            n_channels, n_pixels, _ = waveforms.shape
             if n_channels == 1:  # Reduce if already single channel
                 return waveforms[0]
             else:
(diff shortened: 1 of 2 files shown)


ctapipe/calib/camera/gainselection.py
```diff
@@ -1,173 +1,110 @@
...
-        pass
+        if waveforms.ndim == 2:  # Return if already gain selected
+            return waveforms
+        elif waveforms.ndim == 3:
+            n_channels, n_pixels, _ = waveforms.shape
+            if n_channels == 1:  # Reduce if already single channel
+                return waveforms[0]
+            else:
+                pixel_channel = self.select_channel(waveforms)
+                return waveforms[pixel_channel, np.arange(n_pixels)]
```
thomasarmstrong: Just a question, why not just 'return waveforms[pixel_channel, :]?

watsonjj: Unfortunately is isn't as simple to index the array like that. `waveforms[pixel_channel, :]` returns an array of shape (n_pix, n_pix, n_samples).

watsonjj: (When pixel_channel is an array of n_pix itself)

thomasarmstrong: Ah ok

kosack:
I just remembered that it's also necessary to keep the mask of which channel was used as well as the gain-selected waveform. Perhaps that could be done in another PR.

vuillaut:
> I just remembered that it's also necessary to keep the mask of which channel was used as well as the gain-selected waveform. Perhaps that could be done in another PR.

+1 

@watsonjj Are you working on a fixing PR right now?
This PR broke lstchain with no easy way to fix it without the `pixel_channel` mask.

PR #1094, opened 2019-06-21
Restore returning selected channels in gain selection

watsonjj:


vuillaut:
That was quick.
Thank you.

ctapipe/calib/camera/gainselection.py
```diff
@@ -45,14 +45,17 @@ def __call__(self, waveforms):
             Shape: (n_pix, n_samples)
         """
         if waveforms.ndim == 2:  # Return if already gain selected
-            return waveforms
+            pixel_channel = None  # Provided by EventSource
```
kosack: is `pixel_channel` defined?

kosack: never mind - seems it was fixed as I typed that :)

vuillaut:
Ha didn't pay attention enough.
`pixel_channel` shouldn't be of type `bool` rather than `int`?

watsonjj:
> Ha didn't pay attention enough.
> `pixel_channel` shouldn't be of type `bool` rather than `int`?

I don't think so. Then it would be used as a mask instead of an index when indexing numpy arrays

vuillaut:
Ha yes thanks, it's not a mask indeed.

PR #1096, opened 2019-06-24
First version of the MARS tailcut (1st pass)

HealthyPear:
This is meant to be a first version of the tailcut used in CTA-MARS.
There are also many automatic PEP-8-compliant format changes due to _black_.

**For the moment:**
- it only implements the 1st pass,
- it is written in a way which is doesn't make easy to expand it later on.

I added also the associated unit test, with additional combinations.
This test is also back-compatible.

**What I plan to do next:**
- create a new _tailcut_double_bound_ method for a 2-boundary tailcut,
- call such function in place of this version of the mars 1st pass cleaning,
- the real final MARS cleaning should then be _tailcut_double_bound + (ctapipe's?) time-gradient + tailcut_double_bound_.

### DIFF
--- a/ctapipe/image/cleaning.py
+++ b/ctapipe/image/cleaning.py
@@ -87,7 +87,7 @@ def tailcuts_clean(
         )
 
 
-def mars_image_cleaning(
+def mars_cleaning_1st_pass(
     geom,
     image,
     picture_thresh=7,
@@ -133,9 +133,6 @@ def mars_image_cleaning(
 
     """
 
-    # For debug
-    # print("Original image: {}".format(image))
-
     pixels_from_tailcuts_clean = tailcuts_clean(
         geom,
         image,
@@ -143,27 +140,7 @@ def mars_image_cleaning(
         boundary_thresh[0],
         keep_isolated_pixels,
         min_number_picture_neighbors,
-    )  # this does checks for core pixels and first neighbors
-
-    # For debug
-    '''print(
-        "After tailcuts_clean [{},{}]: {}".format(
-            picture_thresh, boundary_thresh[1], pixels_from_tailcuts_clean
-        )
-    )'''
-
-    # ++++++++++++++++++++++++++++++++++++
-    # THIS PART IS NOT USED FOR THE MOMENT
-    # ++++++++++++++++++++++++++++++++++++
-    # Call dilate on this set of pixels and add all next neighbors,
-    # regardless of their value
-    # extended_image = dilate(geom, pixels_from_tailcuts_clean)
-    #
-    # if extended_image is None:
-    #     print("No extend image available!")
-    # else:
-    #     print("Extended image = {}".format(extended_image))
-    # ++++++++++++++++++++++++++++++++++++
+    )  # this selects any core pixel and first neighbors
 
     # At this point we don't know yet which ones should be kept.
     # In principle, the pixel thresholds should be hierarchical from core to
@@ -173,7 +150,7 @@ def mars_image_cleaning(
     # the mask we got from 'tailcuts_clean'.
 
     pixels_above_2nd_boundary = image >= boundary_thresh[1]
-    # print("pixels_above_2nd_boundary: {}".format(pixels_above_2nd_boundary))
+
     # and now it's the same as the last part of 'tailcuts_clean', but without
     # the core pixels, i.e. we start from the neighbors of the core pixels.
     pixels_with_previous_neighbors = geom.neighbor_matrix_sparse.dot(
--- a/ctapipe/image/cleaning.py
+++ b/ctapipe/image/cleaning.py
@@ -2,7 +2,7 @@
 Image Cleaning Algorithms (identification of noisy pixels)
 """
 
-__all__ = ["tailcuts_clean", "dilate", "mars_cleaning_1st_pass", "fact_cleaning"]
+__all__ = ["tailcuts_clean", "dilate", "mars_cleaning_1st_pass", "fact_image_cleaning"]
 
 import numpy as np
 from scipy.sparse.csgraph import connected_components
--- a/ctapipe/image/cleaning.py
+++ b/ctapipe/image/cleaning.py
@@ -91,7 +91,7 @@ def mars_cleaning_1st_pass(
     geom,
     image,
     picture_thresh=7,
-    boundary_thresh=[5, 5],
+    boundary_thresh=5,
     keep_isolated_pixels=False,
     min_number_picture_neighbors=0,
 ):
@@ -110,12 +110,12 @@ def mars_cleaning_1st_pass(
         Camera geometry information
     image: array
         pixel values
-    picture_thresh: float or array
+    picture_thresh: float
         threshold above which all pixels are retained
-    boundary_thresh: array
-        array of 2 thresholds, first the one above which pixels are retained if
-        they have a neighbor already above the picture_thresh, then the one
-        applied iteratively to the neighbor of the neighbor
+    boundary_thresh: float
+        threshold above which pixels are retained if
+        they have a neighbor already above the picture_thresh; it is then
+        reapplied iteratively to the neighbor of the neighbor
     keep_isolated_pixels: bool
         If True, pixels above the picture threshold will be included always,
         if not they are only included if a neighbor is in the picture or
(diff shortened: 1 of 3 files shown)


ctapipe/image/cleaning.py
```diff
@@ -67,16 +73,123 @@ def tailcuts_clean(geom, image, picture_thresh=7, boundary_thresh=5,
...
+    min_number_picture_neighbors: int
+        A picture pixel survives cleaning only if it has at least this number
+        of picture neighbors. This has no effect in case keep_isolated_pixels is True
+
+    Returns
+    -------
+    A boolean mask of *clean* pixels.  To get a zero-suppressed image and pixel
+    list, use `image[mask], geom.pix_id[mask]`, or to keep the same
+    image size and just set unclean pixels to 0 or similar, use
+    `image[~mask] = 0`
```
dneise: Imo `mask` does not very well express whether it contains the pixels to be discarded, or the pixels surviving the cleaning. The example is necessary to understand what is returned. I think `survivors`or `signal` might be better. This might be a comment for all cleaning functions.

ctapipe/image/cleaning.py
```diff
@@ -67,16 +73,123 @@ def tailcuts_clean(geom, image, picture_thresh=7, boundary_thresh=5,
...
+
+    """
+
+    # For debug
+    # print("Original image: {}".format(image))
```
dneise: Please remove such debug lines before merging.

ctapipe/image/cleaning.py
```diff
@@ -67,16 +73,123 @@ def tailcuts_clean(geom, image, picture_thresh=7, boundary_thresh=5,
...
+    min_number_picture_neighbors=0,
+):
+    """
+    Clean an image by selection pixels that pass a three-threshold tail-cuts
+    procedure.
+    All thresholds are defined with respect to the pedestal
+    dispersion. All pixels that have a signal higher than the core (picture)
+    threshold will be retained, along with all those above the boundary
+    threshold that are neighbors of a core pixel AND all those above
+    the boundary threshold that are neighbors of a neighbor of a core pixel.
```
dneise: In case any reference like a paper or proceeding exists, which you might have used in order to implement this, it might be nice to include the reference here for interested colleagues. 

In addition, in case you would rather refer to the original MAGIC implemenation (which is closed source, as far as I know) I believe it might still be helpful to include a reference here. Just in case someone is interested in comparing to the original C++ code.

ctapipe/image/cleaning.py
HealthyPear: Dear Dominik,

yes, I commented them, but not remove them, my bad!

ctapipe/image/cleaning.py
HealthyPear: Dear Dominik,

for this specific case this is simply orally-distributed information.
I don't know wich reference it should be used exactly, I will ask Abelardo.

Cheers,
Michele

ctapipe/image/cleaning.py
dneise: as I said: no big deal. Should just not be forgotten in the end.

ctapipe/image/cleaning.py
dneise: thats nice of you, thanks.

ctapipe/image/cleaning.py
```diff
@@ -2,15 +2,20 @@
 Image Cleaning Algorithms (identification of noisy pixels)
 """
 
-__all__ = ['tailcuts_clean', 'dilate']
+__all__ = ["tailcuts_clean", "dilate"]
```
kosack: just noticed that `__all__` doesn't include all the public functions in this module.  Please add `mars_cleaning_first_pass`, and while you are at it  `apply_time_delta_cleaning` and `fact_cleaning` to this list.

HealthyPear: OK!

MaxNoe: I think apply time delta cleaning was not supposed to be public, as it's only used by the FACT cleaning, right @LukasNickel ?

HealthyPear: I wish to be able to say that I didn't add it because I also thought of this, but honestly I forgot to add it. Good to know that it wasn't sure, then!

MaxNoe: Could you rename it with a leading underscore?

HealthyPear: Done!

LukasNickel: > I think apply time delta cleaning was not supposed to be public, as it's only used by the FACT cleaning, right @LukasNickel ?

Yeah, its just a helper function

kosack:
The build is failing because of some poorly formatted documentation in `fact_image_cleaning` (not something you added, but might as well fix it now):  This only fails now since it is now included in the docs due to the change to `__all__`.

```
Warning, treated as error:
/home/travis/build/cta-observatory/ctapipe/ctapipe/image/cleaning.py:docstring of ctapipe.image.cleaning.fact_image_cleaning:7:Unexpected indentation.
```

I think the problem is just the text before the *Parameters* section is not proper ReStructuredText in the NumpyDoc standard (see e.g. https://numpydoc.readthedocs.io/en/latest/example.html).  It should be changed to something like the following (though it would be better to add the references to bibliography.rst  with better identifier names and just reference them here).

```
Clean an image by selection pixels that pass the fact cleaning procedure.

Cleaning contains the following steps:
1. Find pixels containing more photons than a threshold t1
2. Remove pixels with less than N neighbors
3. Add neighbors of the remaining pixels that are  above a lower threshold t2
4. Remove pixels with less than N neighbors arriving within a given timeframe
5. Remove pixels with less than N neighbors
6. Remove pixels with less than N neighbors arriving within a given timeframe

References
----------
See  [1]_ and for implementation [2]_

.. [1]   On the hunt for photons: analysis of Crab Nebula data obtained
    by the first G-APD Cherenkov telescope, Thomas Fabian Temme
    http://dx.doi.org/10.17877/DE290R-17773

.. [2] https://github.com/fact-project/fact-tools
```

ctapipe/image/cleaning.py
```diff
@@ -93,6 +183,11 @@ def dilate(geom, mask):
     mask: ndarray
         input mask (array of booleans) to be dilated
     """
+    print("Dilate : input mask = {}".format(mask))
```
MaxNoe: Remove debug prints here

ctapipe/image/cleaning.py
```diff
@@ -67,16 +73,100 @@ def tailcuts_clean(geom, image, picture_thresh=7, boundary_thresh=5,
...
+        return (pixels_above_boundary & pixels_with_picture_neighbors) | (
+            pixels_in_picture & pixels_with_boundary_neighbors
+        )
+
+
+def mars_cleaning_1st_pass(
+    geom,
+    image,
+    picture_thresh=7,
+    boundary_thresh=[5, 5],
```
kosack: you probably shouldn't use a mutable type as a default here - it can lead to unexpected behavior since it's initialized at load time, and then can be modified by code later.

kosack: set it to `None` instead, and in the code check for non and set to `[5,5]` if you really need a default.

kosack: see for example https://docs.python-guide.org/writing/gotchas/#mutable-default-arguments

HealthyPear: I just realized that the version I pushed here is slightly older that that on _protopipe_: there I dropped the list, and instead I fixed both boundary thresholds to the same value which is anyway what MARS is doing.
So boundary_thresh should be actually just 5 (a default value like another).

HealthyPear: For the new methods I plan to develop to make things more modular, I will make one totally general

ctapipe/image/cleaning.py
HealthyPear: An update on this: Abelardo told that he doesn't have a reference at the moment, since what he did is the result of a inspiration from EventDisplay (in fact, MAGIC doesn't use double pass cleaning). I will contact the responsible from EventDisplay.

ctapipe/image/cleaning.py
HealthyPear: Done.

ctapipe/image/cleaning.py
HealthyPear: Done.

ctapipe/image/cleaning.py
MaxNoe: Use a `tuple`: `(5, 5)` that is immutable and you can still set both thresholds.

HealthyPear: Thanks for the advice! If it's ok for you, though, since the image cleaning implementation of CTA-MARS is indeed p-b-b there is no need to do this here. I can do it in a more general implementation of a double-boundary-tailcut which will be able to use for any purpose later on.

kosack:
@MaxNoe are you satisfied with the changes? Can we merge this?

PR #1101, opened 2019-07-09
Add uncertainties to TimingParemeters

MaxNoe:
fixes #1101

### DIFF
--- a/ctapipe/image/timing_parameters.py
+++ b/ctapipe/image/timing_parameters.py
@@ -3,7 +3,7 @@ Image timing-based shower image parametrization.
 """
 
 import numpy as np
-from numpy.polynomial.polynomial import polyfit, polyval
+from numpy.polynomial.polynomial import polyval
 from ctapipe.io.containers import TimingParametersContainer
 from .hillas import camera_to_shower_coordinates
 
@@ -50,7 +50,7 @@ def timing_parameters(geom, image, pulse_time, hillas_parameters):
         hillas_parameters.y,
         hillas_parameters.psi
     )
-    (intercept, slope), cov = polyfit(
+    (intercept, slope), cov = np.polyfit(
         longi.value, pulse_time, deg=1, w=np.sqrt(image), cov=True,
     )
     slope_err, intercept_err = np.sqrt(np.diag(cov))
--- a/ctapipe/image/timing_parameters.py
+++ b/ctapipe/image/timing_parameters.py
@@ -50,7 +50,7 @@ def timing_parameters(geom, image, pulse_time, hillas_parameters):
         hillas_parameters.y,
         hillas_parameters.psi
     )
-    (intercept, slope), cov = np.polyfit(
+    (slope, intercept), cov = np.polyfit(
         longi.value, pulse_time, deg=1, w=np.sqrt(image), cov=True,
     )
     slope_err, intercept_err = np.sqrt(np.diag(cov))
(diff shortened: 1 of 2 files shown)


kosack:
the  values for the `timing.slope` seem to have changed by a factor of 2. That seems strange - is there a difference between the two polyfit functions in what is returned?

ctapipe/image/timing_parameters.py
```diff
@@ -50,9 +50,10 @@ def timing_parameters(geom, image, pulse_time, hillas_parameters):
         hillas_parameters.y,
         hillas_parameters.psi
     )
-    intercept, slope = polyfit(
-        longi.value, pulse_time, deg=1, w=np.sqrt(image)
+    (intercept, slope), cov = np.polyfit(
```
watsonjj: `np.polyfit` returns (slope, intercept)
`numpy.polynomial.polynomial.polyfit` returns (intercept, slope)

So this is now incorrect.

`numpy.polynomial.polynomial.polyfit` is generally recommended over `np.polyfit`. Alternatively the `numpy.polynomial.polynomial.Polynomial` class could be used:
https://docs.scipy.org/doc/numpy-1.9.3/reference/generated/numpy.polynomial.polynomial.Polynomial.html

watsonjj: The recommendation for `numpy.polynomial.polynomial.polyfit` over `np.polyfit` can be found in the docs: https://docs.scipy.org/doc/numpy/reference/generated/numpy.polyfit.html

MaxNoe: Yes, but I found no way on getting an uncertainty from the polynomial module.

Also: I think `polynomial.polyfit` scales the data so that the resulting polynomial has support `[0, 1]`, which results in unphysical and not comparable values, right?

watsonjj: Appears to be in this PR that has been open for ~3 years: https://github.com/numpy/numpy/pull/7880

I have not heard of this scaling nor experienced comparable return values, could you point me to where this is described?

MaxNoe:
```
In [1]: import numpy as np

In [2]: from numpy.polynomial.polynomial import Polynomial, polyfit

In [3]: x = np.linspace(10, 20, 50)

In [4]: y = 5 * x + 2 + np.random.normal(0, 0.1, size=50)

In [5]: np.polyfit(x, y, deg=1)
Out[5]: array([5.00275528, 1.95542864])

In [6]: polyfit(x, y, deg=1)
Out[6]: array([1.95542864, 5.00275528])

In [7]: Polynomial.fit(x, y, deg=1).coef
Out[7]: array([76.99675779, 25.01377638])

In [8]: Polynomial.fit(x, y, deg=1, domain=[]).coef
Out[8]: array([1.95542864, 5.00275528])
```

MaxNoe: What I meant was the `Out[7]` result. That startled me a lot when I first tried `Polynomial.fit`

watsonjj: Ah right - didn't see your edit to specify you meant `Polynomial.fit`. It surprises me too! Honestly I've always stuck with `numpy.polynomial.polynomial.polyfit`.

kosack: the docs mention `Polynomial.fit(x, y, deg=1).convert().coef`

MaxNoe: `domain = []` seems to do the same directly, right?

kosack:
ANyhow, for now I suggest just sticking with np.polyfit, but fixing the order of the return (perhaps add a comment in the docstring that we could use the polynomial module in the future if it adds the ability to return a covariance matrix)

MaxNoe:
The return is fixed and the tests are passing

PR #1105, opened 2019-07-12
Add method to read bright star catalog and find bright stars in sky region

moritzhuetten:
added a new 'astro' class in `ctapipe.utils` with a `get_bright_stars()` method, which returns the Yale bright star catalog (read from ctapipe-extra) into an astropy table, or with arguments
`get_bright_stars(pointing = astropy.SkyCoord(), radius = astropy.Angle(), magnitude_cut = float)`
selects stars brighter than the chosen magnitude in a region of given radius around a sky position

### DIFF
--- a/ctapipe/utils/__init__.py
+++ b/ctapipe/utils/__init__.py
@@ -20,6 +20,7 @@ __all__ = [
     'get_table_dataset',
     'get_dataset_path',
     'find_in_path',
+    'get_bright_stars'
     'CutFlow',
     'PureCountingCut',
     'UndefinedCut',
--- a/ctapipe/utils/__init__.py
+++ b/ctapipe/utils/__init__.py
@@ -20,7 +20,7 @@ __all__ = [
     'get_table_dataset',
     'get_dataset_path',
     'find_in_path',
-    'get_bright_stars'
+    'get_bright_stars',
     'CutFlow',
     'PureCountingCut',
     'UndefinedCut',
--- a/ctapipe/utils/__init__.py
+++ b/ctapipe/utils/__init__.py
@@ -6,7 +6,7 @@ from .table_interpolator import TableInterpolator
 from .unstructured_interpolator import UnstructuredInterpolator
 from .datasets import (find_all_matching_datasets, get_table_dataset, get_dataset_path,
                        find_in_path)
-from .astro import (get_bright_stars)
+from .astro import get_bright_stars
 from .CutFlow import CutFlow, PureCountingCut, UndefinedCut
 
 
--- a/ctapipe/utils/__init__.py
+++ b/ctapipe/utils/__init__.py
@@ -6,6 +6,7 @@ from .table_interpolator import TableInterpolator
 from .unstructured_interpolator import UnstructuredInterpolator
 from .datasets import (find_all_matching_datasets, get_table_dataset, get_dataset_path,
                        find_in_path)
+from .astro import (get_bright_stars)
 from .CutFlow import CutFlow, PureCountingCut, UndefinedCut
 
 
--- a/ctapipe/utils/astro.py
+++ b/ctapipe/utils/astro.py
@@ -37,7 +37,7 @@ def get_bright_stars(pointing=None, radius=None, magnitude_cut=None):
     from astropy.table import Table
     from ctapipe.utils import get_table_dataset
 
-    catalog = get_table_dataset("yale_bright_star_catalog5", 
+    catalog = get_table_dataset("yale_bright_star_catalog5",
                                 role="bright star catalog")
 
     starpositions = SkyCoord(ra=Angle(catalog['RAJ2000'], unit=u.deg),
--- a/ctapipe/utils/astro.py
+++ b/ctapipe/utils/astro.py
@@ -6,17 +6,10 @@ import logging
 
 logger = logging.getLogger(__name__)
 
-try:
-    import ctapipe_resources
-except ImportError:
-    raise RuntimeError("Please install the 'ctapipe-extra' package, "
-                       "which contains the ctapipe_resources module "
-                       "needed by ctapipe. (conda install ctapipe-extra)")
-
 __all__ = ['get_bright_stars']
 
 
-def get_bright_stars(pointing=SkyCoord(ra=0. * u.rad, dec=0. * u.rad, frame='icrs'), 
+def get_bright_stars(pointing=SkyCoord(ra=0. * u.rad, dec=0. * u.rad, frame='icrs'),
                      radius=180. * u.deg, magnitude_cut=7.96):
     """
     Returns an astropy table containing star positions above a given magnitude within
@@ -35,14 +28,15 @@ def get_bright_stars(pointing=SkyCoord(ra=0. * u.rad, dec=0. * u.rad, frame='icr
     Returns
     -------
     Astropy table:
-       List of all stars after cuts with names, catalog numbers, magnitudes, 
+       List of all stars after cuts with names, catalog numbers, magnitudes,
        and coordinates
     """
     from astropy.io import fits
     from astropy.table import Table
     from ctapipe.utils import get_dataset_path
 
-    hdulist = fits.open(get_dataset_path("yale_bright_star_catalog5.fits.gz"))
+    catalog = get_dataset_path("yale_brigh2t_star_catalog5.fits.gz")
+    hdulist = fits.open(catalog)
     table = Table(hdulist[1].data)
 
     starpositions = SkyCoord(ra=Angle(table['RAJ2000'], unit=u.deg),
--- a/ctapipe/utils/astro.py
+++ b/ctapipe/utils/astro.py
@@ -42,7 +42,7 @@ def get_bright_stars(pointing=SkyCoord(ra=0. * u.rad, dec=0. * u.rad, frame='icr
     from astropy.table import Table
     from ctapipe.utils import get_dataset_path
 
-    hdulist = fits.open(get_dataset_path("brightstars.fits.gz"))
+    hdulist = fits.open(get_dataset_path("yale_bright_star_catalog5.fits.gz"))
     table = Table(hdulist[1].data)
 
     starpositions = SkyCoord(ra=Angle(table['RAJ2000'], unit=u.deg),
--- a/ctapipe/utils/astro.py
+++ b/ctapipe/utils/astro.py
@@ -1,4 +1,9 @@
 # Licensed under a 3-clause BSD style license - see LICENSE.rst
+"""
+This module is intended to contain astronomy-related helper tools which are
+not provided by external packages and/or to satisfy particular needs of
+usage within ctapipe.
+"""
 from astropy.coordinates import Angle
 from astropy.coordinates import SkyCoord
 from astropy import units as u
@@ -22,7 +27,7 @@ def get_bright_stars(pointing=None, radius=None, magnitude_cut=None):
     Parameters
     ----------
     pointing: astropy Skycoord
-       pointing direction in the sky
+       pointing direction in the sky (if none is given, full sky is returned)
     radius: astropy angular units
        Radius of the sky region around pointing position. Default: full sky
     magnitude_cut: float
--- a/ctapipe/utils/astro.py
+++ b/ctapipe/utils/astro.py
@@ -7,9 +7,6 @@ usage within ctapipe.
 from astropy.coordinates import Angle
 from astropy.coordinates import SkyCoord
 from astropy import units as u
-import logging
-
-logger = logging.getLogger(__name__)
 
 __all__ = ['get_bright_stars']
 
@@ -39,7 +36,6 @@ def get_bright_stars(pointing=None, radius=None, magnitude_cut=None):
        List of all stars after cuts with names, catalog numbers, magnitudes,
        and coordinates
     """
-    from astropy.table import Table
     from ctapipe.utils import get_table_dataset
 
     catalog = get_table_dataset("yale_bright_star_catalog5",
--- a/ctapipe/utils/astro.py
+++ b/ctapipe/utils/astro.py
@@ -50,7 +50,8 @@ def get_bright_stars(pointing=None, radius=None, magnitude_cut=None):
 
     if radius is not None:
         if pointing is None:
-            raise ValueError('Sky pointing, pointing=SkyCoord(), must be provided if radius is given.')
+            raise ValueError('Sky pointing, pointing=SkyCoord(), must be \
+                              provided if radius is given.')
         separations = starpositions.separation(pointing)
         table['separation'] = separations
         table = table[separations < radius]
--- a/ctapipe/utils/astro.py
+++ b/ctapipe/utils/astro.py
@@ -9,12 +9,15 @@ logger = logging.getLogger(__name__)
 __all__ = ['get_bright_stars']
 
 
-def get_bright_stars(pointing=SkyCoord(ra=0. * u.rad, dec=0. * u.rad, frame='icrs'),
-                     radius=180. * u.deg, magnitude_cut=7.96):
+def get_bright_stars(pointing=None, radius=None, magnitude_cut=None):
     """
     Returns an astropy table containing star positions above a given magnitude within
     a given radius around a position in the sky, using the Yale bright star catalog
-    which needs to be present in the ctapipe-extra package.
+    which needs to be present in the ctapipe-extra package. The included Yale bright
+    star catalog contains all 9096 stars, excluding the Nova objects present in the
+    original catalog from  Hoffleit & Jaschek (1991), 
+    http://adsabs.harvard.edu/abs/1991bsc..book.....H, and is complete down to 
+    magnitude ~6.5, while the faintest included star has mag=7.96.
 
     Parameters
     ----------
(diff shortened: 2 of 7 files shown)


ctapipe/utils/astro.py
```diff
@@ -0,0 +1,55 @@
...
+from astropy import units as u
+import logging
+
+logger = logging.getLogger(__name__)
+
+__all__ = ['get_bright_stars']
+
+
+def get_bright_stars(pointing=SkyCoord(ra=0. * u.rad, dec=0. * u.rad, frame='icrs'),
+                     radius=180. * u.deg, magnitude_cut=7.96):
```
kosack: it might be clearer to have magnitude_cut default to `None`, so it clear all stars are included, rather than a "magic" value (though you could still note in the docstring that the Yale catalog stops at magnitude 7.96).   Should also mention which catalog is used in the docstring, ideally with a citation and date.

ctapipe/utils/astro.py
```diff
@@ -0,0 +1,55 @@
...
+from astropy.coordinates import SkyCoord
+from astropy import units as u
+import logging
+
+logger = logging.getLogger(__name__)
+
+__all__ = ['get_bright_stars']
+
+
+def get_bright_stars(pointing=SkyCoord(ra=0. * u.rad, dec=0. * u.rad, frame='icrs'),
```
MaxNoe: I would not give a default for the pointing.

ctapipe/utils/astro.py
```diff
@@ -0,0 +1,55 @@
...
+    -------
+    Astropy table:
+       List of all stars after cuts with names, catalog numbers, magnitudes,
+       and coordinates
+    """
+    from astropy.io import fits
+    from astropy.table import Table
+    from ctapipe.utils import get_dataset_path
+
+    catalog = get_dataset_path("yale_brigh2t_star_catalog5.fits.gz")
```
MaxNoe: These three lines can just be `Table.read` I think.

ctapipe/utils/astro.py
```diff
@@ -0,0 +1,55 @@
...
+    hdulist = fits.open(catalog)
+    table = Table(hdulist[1].data)
+
+    starpositions = SkyCoord(ra=Angle(table['RAJ2000'], unit=u.deg),
+                             dec=Angle(table['DEJ2000'], unit=u.deg), frame='icrs')
```
MaxNoe: Add `copy=False` to make it faster and less memory intensive

ctapipe/utils/astro.py
MaxNoe: Maybe use a radius that is closer to the max. fov of CTA telescope's? Maybe 15 degrees? Agreed with @kosack on the magnitude cut.  use `None` there.

ctapipe/utils/astro.py
```diff
@@ -0,0 +1,55 @@
...
+    table['ra_dec'] = starpositions
+
+    if radius < 180. * u.deg:
```
MaxNoe: I think skycoords have a more efficient way to do this:

https://docs.astropy.org/en/stable/api/astropy.coordinates.SkyCoord.html#astropy.coordinates.SkyCoord.search_around_sky

ctapipe/utils/astro.py
```diff
@@ -0,0 +1,55 @@
...
+        separations = starpositions.separation(pointing)
+        table['separation'] = separations
+        table = table[separations < radius]
+    if magnitude_cut < 7.96:
```
MaxNoe: Maybe do the magnitude cut first, as it is faster and will significantly reduce the number of comparison needed for the radius cut.

ctapipe/utils/__init__.py
```diff
@@ -6,6 +6,7 @@
 from .unstructured_interpolator import UnstructuredInterpolator
 from .datasets import (find_all_matching_datasets, get_table_dataset, get_dataset_path,
                        find_in_path)
+from .astro import (get_bright_stars)
```
MaxNoe: unnecessary parentheses

ctapipe/utils/astro.py
moritzhuetten: changed and added in 96c2b77

ctapipe/utils/astro.py
moritzhuetten: corrected in 96c2b77

ctapipe/utils/astro.py
moritzhuetten: Thanks for pointing this out. Syntax shortened in 96c2b77

ctapipe/utils/astro.py
moritzhuetten: Done in 96c2b77

ctapipe/utils/astro.py
moritzhuetten: So the default is `None` now, which loads the whole catalog. Advantage of this standard I see is that loading the full sky (i.e., radius 180°), doesn't need a (standard) pointing to be defined. But please correct me if you see a more convenient standard.

ctapipe/utils/astro.py
```diff
@@ -23,32 +26,33 @@ def get_bright_stars(pointing=SkyCoord(ra=0. * u.rad, dec=0. * u.rad, frame='icr
...
+    table = Table.read(catalog)
 
     starpositions = SkyCoord(ra=Angle(table['RAJ2000'], unit=u.deg),
-                             dec=Angle(table['DEJ2000'], unit=u.deg), frame='icrs')
+                             dec=Angle(table['DEJ2000'], unit=u.deg), 
+                             frame='icrs', copy=False)
     table['ra_dec'] = starpositions
 
-    if radius < 180. * u.deg:
+    if magnitude_cut != None:
```
MaxNoe: Use `is not None` here

ctapipe/utils/astro.py
```diff
@@ -23,32 +26,33 @@ def get_bright_stars(pointing=SkyCoord(ra=0. * u.rad, dec=0. * u.rad, frame='icr
...
+        table = table[table['Vmag'] < magnitude_cut]
+        
+    if radius != None:
```
MaxNoe: same as above

ctapipe/utils/astro.py
MaxNoe: No, now the default value is `None` which is not a valid input for `pointing`.

No default is different than `None` as default. Make t a positional argument, just `get_bright_stars(pointing, ...)`

ctapipe/utils/astro.py
moritzhuetten: I am not sure, both how to implement it and if it improves much, unless you see a globally better way to implement this feature (as some other object instead of returning an astropy table), there are surely many ways to do it. First, `search_around_sky()` apparently requires an array of SkyCoord's, not a Skycoord object containing many pointings, as the current syntax loads the data (I couldn't find a solution to revert the order). Second, I fear to then produce the cut table also with still at least containing the star names and magnitudes also requires some dedicated lines... So I didn't see a solution on the short term, but if you do, please let me know!

ctapipe/utils/astro.py
moritzhuetten: Done in 96c2b77

ctapipe/utils/__init__.py
moritzhuetten: Done in 96c2b77

ctapipe/utils/astro.py
MaxNoe: Mmh okay, so this is a case where the `pointing = None` makes sense. 
Maybe we should split it into two methods, one which loads the catalog and one which searches around a position in the sky?

ctapipe/utils/astro.py
moritzhuetten: Done in f4139b6

ctapipe/utils/astro.py
moritzhuetten: Done in f4139b6

ctapipe/utils/astro.py
MaxNoe: Yeah, sorry. Even the error message mentions that:
```
ValueError: One of the inputs to search_around_sky is a scalar. search_around_sky is intended for use with array coordinates, not scalars.  Instead, use ``coord1.separation(coord2) < seplimit`` to find the coordinates near a scalar coordinate```

File: ctapipe/utils/astro.py (hunk 1 above)
Comment (moritzhuetten) [reply ctapipe/utils/astro.py]: Yes, I was also thinking about this, especially if the funcion is used in a loop. But then you could also make a more simple function loading just the catalog, and the user has to code the two lines to cut it her/himself. As it is now I thought it's a convenience feature to allow to load & select in region/magnitude in just one line...

File: ctapipe/utils/astro.py [hunk 10]
```diff
@@ -0,0 +1,59 @@
...
+    starpositions = SkyCoord(ra=Angle(table['RAJ2000'], unit=u.deg),
+                             dec=Angle(table['DEJ2000'], unit=u.deg),
+                             frame='icrs', copy=False)
+    table['ra_dec'] = starpositions
+
+    if magnitude_cut is not None:
+        table = table[table['Vmag'] < magnitude_cut]
+
+    if radius is not None:
+        separations = starpositions.separation(pointing)
```
MaxNoe: How helpfull is the error message if radius is provided but not pointing? I guess something along the lines of
`NoneType has no attribute something`?

Maybe explicitly throw an error if `radius is not None` but `if pointing is None`

moritzhuetten: Like this 7d57d25?

ctapipe/utils/__init__.py
```diff
@@ -6,6 +6,7 @@
 from .unstructured_interpolator import UnstructuredInterpolator
 from .datasets import (find_all_matching_datasets, get_table_dataset, get_dataset_path,
                        find_in_path)
+from .astro import get_bright_stars
```
MaxNoe: Add this method to `__all__`

ctapipe/utils/astro.py
MaxNoe: ok

MaxNoe:
Sorry, one more thing. I have a distaste for submodules called utils . They tend to acquire random stuff over time.

Is there a better place for this? E.g. the coordinates submodule?

moritzhuetten:
> Sorry, one more thing. I have a distaste for submodules called utils . They tend to acquire random stuff over time.
> 
> Is there a better place for this? E.g. the coordinates submodule?

Fine with me, if you prefer (or any other place). In any case, my idea for an 'astro' submodule was to gather there all astronomy-related stuff (which astropy or other modules) can not provide. For example, I discussed with @noahhdf last week whether potentially this could also a place to add some dark-patch-finder or selector (according to day, moon distance,...) or similar helpers. But that's another story, of course.

MaxNoe:
> dark-patch-finder or selector (according to day, moon distance,...) or similar helpers. But that's another story, of course.

I build something like this some time ago for FACT using the hipparcos star catalogue, but haven't used it in at least two years since e don't do ratescans where the shifter selects the spot anymore. 

https://github.com/fact-project/darkspot

kosack:
We can probably make another issue or PR to deprecate `ctapipe.utils` and move the things inside to new places. I agree "utils" is really a catch-all for unrelated things.

kosack:
I updated ctapipe-extra and released a new package v0.2.18 that should work with your tests. Let's see if it works.

ctapipe/utils/tests/test_astro.py
```diff
@@ -0,0 +1,12 @@
...
+from astropy import units as u
+
+def test_get_bright_stars():
+
+    pointing = SkyCoord(ra=83.275 * u.deg, dec=21.791 * u.deg, frame='icrs')
+
+    table = get_bright_stars(pointing, radius=5. * u.deg, magnitude_cut=3.5)
+
+    assert len(table) == 1
+    assert table[0]['Name'] == '123Zet Tau'
```
kosack: Just a question: is the cut of 5 degrees too large for this test? I assume you want to try to *only* get Zeta Tau.  You are not necessarily guaranteed to get the table back in the same order (even if that is the current way it's implemented, this might change, and shouldn't be a requirement), unless you say somewhere that the returns results are sorted by distance to the pointing?  Otherwise, i'd suggest a smaller radius so you are guaranteed to get only 1 result, so then table[0] must be Zeta Tau.

MaxNoe: Zeta Tauri should be the only star brighter than 3.5 in a 5 degree radius around crab. Also see the `assert len(table) == 1`

moritzhuetten:
> I updated ctapipe-extra and released a new package v0.2.18 that should work with your tests. Let's see if it works.

Hi Karl, thanks a lot!
however, when Travis runs it fails with
```
>       table = get_bright_stars(pointing, radius=2. * u.deg, magnitude_cut=3.5)
/home/travis/build/cta-observatory/ctapipe/ctapipe/utils/tests/test_astro.py:17: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
/home/travis/build/cta-observatory/ctapipe/ctapipe/utils/astro.py:40: in get_bright_stars
    catalog = get_dataset_path("yale_bright_star_catalog5.fits.gz")
/home/travis/build/cta-observatory/ctapipe/ctapipe/utils/datasets.py:143: in get_dataset_path
    return ctapipe_resources.get(filename)
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
resource_name = 'yale_bright_star_catalog5.fits.gz'
    def get(resource_name):
        """ get the filename for a resource """
        if not pkg_resources.resource_exists(__name__, resource_name):
            raise FileNotFoundError("Couldn't find resource: '{}'"
>                                   .format(resource_name))
E           FileNotFoundError: Couldn't find resource: 'yale_bright_star_catalog5.fits.gz'
```
Any idea why it doesn't find the file?

Thanks and cheers
Moritz

MaxNoe:
You need to increase the version number of ctapipe-extra in the `yml` environment files.

ctapipe/utils/tests/test_astro.py
```diff
@@ -0,0 +1,20 @@
+from ..astro import get_bright_stars
+from astropy.coordinates import SkyCoord
+from astropy import units as u
+
+try:
+    import ctapipe_resources
+except ImportError:
+    raise RuntimeError("Please install the 'ctapipe-extra' package, "
+                       "which contains the ctapipe_resources module "
+                       "needed by ctapipe. (conda install ctapipe-extra)")
```
MaxNoe: This should not be needed. A similar error is raised by `get_dataset_path` if the file is not found.

ctapipe/utils/tests/test_astro.py
kosack: ah, i missed the assert. It's fine then

ctapipe/utils/tests/test_astro.py
moritzhuetten: yes, sorry, I suspected. Was just a "test" to figure out why the test by Travis CI did not succeed. Just removing it again.

kosack:
yes, just update `environment.yml`, and `py3.6_env.yml`, `py3.7_env.yml`  and change the line with ctapipe-extra to:

```
 - ctapipe-extra=0.2.18
``` 

That should fix it.

ctapipe/utils/astro.py
```diff
@@ -0,0 +1,62 @@
...
+    -------
+    Astropy table:
+       List of all stars after cuts with names, catalog numbers, magnitudes,
+       and coordinates
+    """
+    from astropy.table import Table
+    from ctapipe.utils import get_dataset_path
+
+    catalog = get_dataset_path("yale_bright_star_catalog5.fits.gz")
+    table = Table.read(catalog)
```
kosack: Oh, another thing, you should change this to:
```py3
from ctapipe.utils import get_table_dataset
...

catalog = get_table_dataset("yale_bright_star_catalog5", role="bright star catalog")
```

That both returns an astropy Table automatically, and also registers the file with the Provenance system so we know that it was opened and used.

kosack:
Once you update the yml files it should be fine.

moritzhuetten:
> Once you update the yml files it should be fine.

And I am just updating the API doc...

moritzhuetten:
> And I am just updating the API doc...

Should be done now!

MaxNoe:
You forgot the two other yaml files (py3.6 and py3.7), unfortunately we need all three of them

moritzhuetten:
> You forgot the two other yaml files (py3.6 and py3.7), unfortunately we need all three of them

Added!

ctapipe/utils/__init__.py
MaxNoe: This is still missing!

ctapipe/utils/__init__.py
```diff
@@ -20,6 +20,7 @@
     'get_table_dataset',
     'get_dataset_path',
     'find_in_path',
+    'get_bright_stars'
```
MaxNoe: missing comma

ctapipe/utils/__init__.py
moritzhuetten: > This is still missing!

Now added

ctapipe/utils/__init__.py
moritzhuetten: corrected.

PR #1106, opened 2019-07-15
Restore apply_time_delta_cleaning as a public method

watsonjj:
`apply_time_delta_cleaning` is a useful function that we started utilising in the CHEC cleaning. I suggest it is restored as a public method.

### DIFF
--- a/ctapipe/image/cleaning.py
+++ b/ctapipe/image/cleaning.py
@@ -229,9 +229,10 @@ def number_of_islands(geom, mask):
 
 
 def apply_time_delta_cleaning(
-    geom, mask, arrival_times, min_number_neighbors, time_limit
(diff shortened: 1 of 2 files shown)


MaxNoe:
Are there any tests for this function?

ctapipe/image/cleaning.py
```diff
@@ -222,7 +228,7 @@ def number_of_islands(geom, mask):
     return num_islands, island_labels
 
 
-def _apply_time_delta_cleaning(
+def apply_time_delta_cleaning(
     geom, mask, arrival_times, min_number_neighbors, time_limit
 ):
     """ Remove all pixels from selection that have less than N
```
kosack: should perhaps change the docstring to "identify all pixels", since it doesn't remove anything.

Issue #1102, opened 2019-07-09
import ctapipe.image module slow because of numba functions


vuillaut:
Some modules take a long time to import.
e.g.:
```
In [1]: %time import ctapipe.image
CPU times: user 22.2 s, sys: 471 ms, total: 22.6 s
Wall time: 22.4 s
```

Is this normal? It seems unexpectedly long.
I checked on two different machines and get similar results.

I have trouble tracking down the reason of it, any help or explanation would be welcome.

MaxNoe:
This seems to be compilation in `ctapipe/image/extractor.py`.

If I remove `from .extractor import *` from `ctapipe/image/__init__.py` the import is done in 9.6 seconds.

vuillaut:
Couldn't the compilation be done during installation rather than import?

MaxNoe:
This is numba code, I don't think so

vuillaut:
Ha yes right.

MaxNoe:
The `neighbor_average_waveform_jit is the offender`

It compiles on import, I though numba would compile on first use???

MaxNoe:
This function is compiled for multiple input data types, If I just compile for one, the import time is reduced to 3 seconds.

For half of the signatures, there is a `TODO Deprecate`

kosack:
> It compiles on import, I though numba would compile on first use???

that used to be the behavior - maybe it changed?

kosack:
It's also strange that such a small function takes so long to compile. Perhaps it is compiling versions for many possible input types? That can be restricted, I think

MaxNoe:
It's already restricted to 8 different input combinations. 4 of them marked with `# TODO: deprecate`

kosack:
Perhaps we can restrict to one only (I think we can assume all waveforms will be the same type), so that might give some speedup.

kosack:
I think just adding `cache=True` to the decorator may solve it: 

```py3
@guvectorize(
    [
        (float64[:], int64, int64, int64, float64[:]),
        (float32[:], int64, int64, int64, float64[:]),
    ],
    '(s),(),(),()->()',
    nopython=True,
    cache=True,
)
def sum_samples_around_peak(waveforms, peak_index, width, shift, ret):

```

MaxNoe:
Could we then import it in `setup.py` so that it basically happens on install?

kosack:
Wouldn't that mean I would then need to build a package for all architectures again? I was hoping to not require that if possible

[...]

PR #1108, opened 2019-07-16
 Reduce numba signatures to improve import time

watsonjj:
Addresses #1102

```
In [1]: %time from ctapipe.image import extractor
CPU times: user 1.54 s, sys: 260 ms, total: 1.8 s
Wall time: 1.51 s
```

### DIFF
--- a/.mailmap
+++ b/.mailmap
@@ -41,5 +41,27 @@ Thomas Armstrong <thomas.armstrong@physics.ox.ac.uk> thomasarmstrong <thomas.arm
 Thomas Gasparetto <thomas.gasparetto@ts.infn.it> thomas gasparetto <thomas.gasparetto@ts.infn.it>
 
 Tarek Hassan <tarek.hassan@desy.de> tialis <goldstein2034@hotmail.com>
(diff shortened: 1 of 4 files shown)


MaxNoe:
This splits the function into to functions, right?

Are there no uses of this function with `ndim=3`? If so, for what do we need the `_3d` function?

kosack:
> Are there no uses of this function with ndim=3? If so, for what do we need the _3d function?

I guess the question is: do we still want to be able to integrate each gain channel separately in some cases (e.g. for checking linearity)? I guess it would be nice to have that, but could be added back later upon need as well.

watsonjj:
As stated in #1095, I wanted to put the 3d waveform functionality of the charge extractors through a deprecation cycle. Splitting `neighbor_average_waveform` into two functions achieves this more elegantly. And neither of the two `neighbor_average_waveform` functions are compiled on import - only on call.

watsonjj:
Also, `ctapipe.calib.camera.flatfield` and `ctapipe.calib.camera.pedestal` appear to import the extractors directly, and seem to somewhat do the calibration chain itself inside the classes, and therefore the tests still use the 3d waveform functionality of the extractors. I haven't had time to look into the code in detail, but I intend to investigate and refactor soon.

PR #1111, opened 2019-07-18
Add Fields to calibration containers

FrancaCassol:
This PR add two fields to the Monitoring containers that are necessary for the LST1 calibration.
Would be it possible to accept it fast? Otherwise I must double the containers in lstchain

Thanks very much

### DIFF
--- a/ctapipe/io/containers.py
+++ b/ctapipe/io/containers.py
@@ -683,8 +683,13 @@ class FlatFieldContainer(Container):
 
     charge_median_outliers = Field(
         None,
-        "Boolean np array of charge (median) outliers (n_chan, n_pix)"
+        "Boolean np array of charge median outliers (n_chan, n_pix)"
     )
+    charge_std_outliers = Field(
+        None,
+        "Boolean np array of charge std outliers (n_chan, n_pix)"
+    )
+
     time_median_outliers = Field(
         None,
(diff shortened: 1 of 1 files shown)


MaxNoe:
@FrancaCassol I think you own us a round of drinks at the next meeting, this is PR  #1111

FrancaCassol:
:-) 
only if you approve it ;-)

ctapipe/io/containers.py
```diff
@@ -763,6 +768,12 @@ class WaveformCalibrationContainer(Container):
         "np array of (digital count) to (photon electron) coefficients (n_chan, n_pix)"
     )
 
+    pedestal_per_sample = Field(
```
MaxNoe: Is this pedestal for each sample? Or pedestale averaged over samples?

If it is per sample, I would expect shape `(n_channel, n_pixel, n_samples)`

FrancaCassol: It is an averaged value but corresponding to one sample, to put it the same value for n_sample it seems to me overdone... also because generally this value is subtracted from the charge. I put the value for one sample because the signal integration window could be different from the pedestal estimation window, so this value will be simply multiplied by the signal window width...

MaxNoe: Ah ok, understood.

PR #1115, opened 2019-07-30
Add Monitoring container to MC data

FrancaCassol:
Waiting for a better solution concerning the handing of Monitoring data in ctapipe, I propose to change the DataContainer() in simteleventsource.py with an EventAndMonDataContainer() .
This should permit to test the calibration tools with MC data following the same procedure used for real data.

### DIFF
--- a/ctapipe/io/simteleventsource.py
+++ b/ctapipe/io/simteleventsource.py
@@ -1,7 +1,7 @@
 import warnings
 import numpy as np
 from ctapipe.io.eventsource import EventSource
-from ctapipe.io.containers import DataContainer
+from ctapipe.io.containers import EventAndMonDataContainer
 from astropy import units as u
 from astropy.coordinates import Angle
 from astropy.time import Time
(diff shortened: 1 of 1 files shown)


kosack:
Is the idea then to later fill the mon containers from simulations?

FrancaCassol:
Exactly.

FrancaCassol:
Hi @kosack,

could you please merge this PR?
Thanks

MaxNoe:
This only exchanges the class of this container in the SimtelEventSource.
It does not fill it. 

What is the reason, the the `mon` field is not just added to `DataContainer`?

FrancaCassol:
Hi @MaxNoe,

> This only exchanges the class of this container in the SimtelEventSource.
> It does not fill it.

the container  will be filled by the calibration scripts (as for the data).

> What is the reason, the the `mon` field is not just added to `DataContainer`?

I made this choice to separate the event and the monitoring level. As said, this
is surely not the final/better solution, but it works fine at least at R1 level.

PR #1119, opened 2019-08-26
`timing_parameters` cleaning mask

vuillaut:
It appears that implicitly masking pixels for image > 0 rather than explicitly asking for a cleaning mask tends to make the user (at least me) use the function in a bad way.
Explicitly requiring a cleaning mask should make the user more aware of the need to clean the image before using the function `timing_parameters`.

### DIFF
--- a/ctapipe/image/timing_parameters.py
+++ b/ctapipe/image/timing_parameters.py
@@ -36,19 +36,18 @@ def timing_parameters(geom, image, pulse_time, hillas_parameters, cleaning_mask=
     timing_parameters: TimingParametersContainer
     """
 
+    unit = geom.pix_x.unit
+
     if cleaning_mask is not None:
         image = image[cleaning_mask]
         geom = geom[cleaning_mask]
+        pulse_time = pulse_time[cleaning_mask]
 
     if (image < 0).any():
         raise ValueError("The non-masked pixels must verify signal >= 0")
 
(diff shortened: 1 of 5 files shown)


MaxNoe:
The original reasoning was, that people pass in `image[mask]`, `geom[mask]` like with the hillas calculation and that then the negative sizes should be ignored.

Maybe it's better to raise an error on negative values here.

But `cleaning_mask` should be optional, to enable the `image[mask], geom[mask]` thingy, right?

ctapipe/image/timing_parameters.py
```diff
@@ -27,21 +27,26 @@ def timing_parameters(geom, image, pulse_time, hillas_parameters):
...
+    cleaning_mask: optionnal, array, dtype=bool
+        The pixels that survived cleaning, e.g. tailcuts_clean
+        The non-masked pixels must verify signal > 0
 
     Returns
     -------
     timing_parameters: TimingParametersContainer
     """
 
+    cleaning_mask = np.ones(image.shape, dtype=bool) if cleaning_mask is None else cleaning_mask
```
MaxNoe: unnecessary and slow as it allocates a new array.
```python

if cleaning_mask is not None:
    image = image[cleaning_mask]
    geom = geom[cleaning_mask] 
```

vuillaut:
Sorry I made the changes too quickly without pytesting.

PR #1121, opened 2019-09-12
Added tailcuts data volume reducer

Hckjs:
Hey, it's my first PR at all, so hopefully i'm doing this right:

I've added a `TailCutsDataVolumeReducer`, which reduce a ndarray of waveforms (n_pix, m_samples) to a ndmasked_array like `np.ma.masked_array(waveforms, mask=mask)`.

The reduction has mainly 3 Steps:
1) Normal `tailcuts_clean` with picture and boundary thresholds
2) Adding iteratively all pixels with Signal S >= iteration threshold (variable argument)
    with `dilate` until no new pixels were added. It doesn't really take affect on images with a small 
    amount of pixels (like 5 or 6) with cherenkov signal, but you can get way more True-Positives for
    images with a higher amount of pixels (like >50) with cherenkov signal, without getting much
    more False-Positives.
3) Adding pixels at the end with `dilate` to get a bit more conservative (variable argument for how
    many times) and closing some gaps from step 2).

So there is just a `tailcuts_clean`, `dilate` and `numpy` dependency.

Just for the visualization an example on simulation data (well i was searching for a quite good one, but it will look similar to this):
The different colours are the different classificators (TP = True-Positive, FN = False-Negative and so on). On the left side its with normal `tailcuts_clean` and on the right side with step 2) in addition (same thresholds). Its without any dilation at the end.

camtctcmod

The test i've added is just a simple one:

It creates a ndarray of images (all the same images ) with 5 chosen pixels (see picture below), which are all connected to a line.
One core-pixel and two boundaries should be selected in step 1) with `tailcuts_clean`. Two more should be selected in step 2) with 2 iteration steps and also all the neighbors of those.
So it just puts up a self-made mask of those pixels and looks whether they are equal with the `TailCutsDataVolumeReducer` or not.

reducer_test


Hckjs:
Hey @vuillaut ,
Tanks for your response!
The `gain_selector` helped a bit to understand, but i thought, that i have to redefine `__call__` in the subclass `TailCutsDataVolumeReducer`, because in the parent class its an abstractmethod. In `gain_selector` its not. Or do i have a thinking error here?
So the actual reduction has to be implemented into the `__call__` in `DataVolumeReducer` with some abstractmethod (like `select_pixels` or something, as you mentioned) and then i have to redefine this `select_pixels` in my subclass `TailcutsDataVolumeReducer`?
Do I have to use traitlets for all the params `picture_thresh, boundary_thresh` etc.?

vuillaut:
> So the actual reduction has to be implemented into the `__call__` in `DataVolumeReducer` with some abstractmethod (like `select_pixels` or something, as you mentioned) and then i have to redefine this `select_pixels` in my subclass `TailcutsDataVolumeReducer`?

It is a suggested implementation that would be consistent with the `gain_selector` one.

> Do I have to use traitlets for all the params `picture_thresh, boundary_thresh` etc.?
Yes, I think it's the main reason why the method 
```
CameraCalibrator(data_volume_reducer=TailCutsDataVolumeReducer())
```
does not work.

kosack:
> Do I have to use traitlets for all the params picture_thresh, boundary_thresh etc.?

Yes, all user-definable options should be traitlets with the `config=True` metadata flag set, then they will become configurable.  I found also that this is not so well documented or friendly, but following the GainSelector as @vuillaut suggests is probably the easiest.    I'm going to check in a few improvements to the whole mechanism soon, with some better examples.   You might also check with @watsonjj since he implemented the CameraCalibrator and related classes, so he can check that it is implemented in the way he intended .

kosack:
If I understand correctly, in this method, you basically clean each *time slice* of the full shower "movie"  separately. . That means that the picture and boundary threshold are *not* the same as the ones used to clean the final integrated image later: the unit is "PE/ns" rather than integrated "PE", so the values used should be something like the standard picture/boundary thresholds divided by the number of samples in the integration window., correct?  

Another way to do this would be to first do a rough integration of the pixels, clean and dilate the final "1D" image, and then use that mask to keep the full waveform where needed.

LenkaTomankova:
Hi @kosack,

> Another way to do this would be to first do a rough integration of the pixels, clean and dilate the final "1D" image, and then use that mask to keep the full waveform where needed.

This is actually pretty much how it is being done right now. The image is calibrated and peak-integrated first, so the resulting dimension is just n_pix (and not (n_pix, m_samples)) with units of pe. Currently we pass thresholds as absolute pe-values. As R1 data will only be relatively calibrated, we will eventually have to derive relative thresholds on the fly, but that is the next step down the road.

MaxNoe:
> Currently we pass thresholds as absolute pe-values.

What is *currently* in this context? This certainly does not fit the implementation in this PR

Hckjs:
> What is _currently_ in this context? This certainly does not fit the implementation in this PR

I made a mistake there. I thought that with the dim (n_pixels, n_samples) in the dummy scheme was meant , that it should make the reduction for more than one image at once. Instead of the time bins for each pixel, i thought 'n_samples' was meant for 'images'. Thats why i made up the big loop at the beginning, I'm sorry, now i got it. I'll change it in the next comit, so that you have a 1D image (calibrated and peak-integrated) as input, then doing the reduction and have a mask as output.

Hckjs:
A little update here:

The Algorithm itself stays the same, but
1) The whole implementation should now looks like the `gain_selector`
2) You have now the waveforms (n_pix, n_sample) as input for the `TailCutsDataVolumeReducer` and get just a mask as output. (There is now a rough integration at the beginning with the `LocalPeakWindowSum`)
3) The `NullDataVolumeReducer` has also a mask as output.
4) I've added 1 line in the `CameraCalibrator` (r1 to DL0), which applies the mask onto the waveforms. (Instead of doing this in the VolumeReducer, to have masks as outputs from the VolumeReducers)
5) Added traitlets for the parameters.

I've made a small .ipynb as an example with two variants which both work with the implementation. In variant 1 you can simply use it as intended in the `CameraCalibrator` as argument and get the reduced image. In variant 2 you manually apply the mask onto the unreduced image to get the reduced image as well as the unreduced image and maybe do some comparative studies or just play a bit with it.

[Link for the .ipynb](https://github.com/Hckjs/Notebooks/blob/master/Example_script_TCDVR.ipynb)

ctapipe/image/reducer.py
```diff
@@ -42,5 +66,109 @@ class NullDataVolumeReducer(DataVolumeReducer):
...
+class TailCutsDataVolumeReducer(DataVolumeReducer):
+    """
+    Reduce the time integrated shower image in 3 Steps:
+
+    1) Select pixels with tailcuts_clean.
+    2) Add iteratively all pixels with Signal S >= boundary_thresh
+       with ctapipe module dilate until no new pixels were added.
+    3) Adding new pixels with dilate to get more conservative.
+    """
+    camera_geom = traits.Any(
```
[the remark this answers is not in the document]
watsonjj: Actually better yet, pass the subarray to `__init__` as is done for the `ImageExtractor` classes

ctapipe/calib/camera/calibrator.py
```diff
@@ -171,8 +171,9 @@ def _calibrate_dl0(self, event, telid):
         if self._check_r1_empty(waveforms):
             return
 
-        reduced_waveforms = self.data_volume_reducer(waveforms)
-        event.dl0.tel[telid].waveform = reduced_waveforms
+        reduced_waveforms_mask = self.data_volume_reducer(waveforms)
+        waveforms[~reduced_waveforms_mask] = 0
```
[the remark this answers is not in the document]
Hckjs: Got an  "AttributeError: 'NoneType' object has no attribute 'copy'" for the calibrator_test if i do it like this, so i just add 1 line
`waveforms_copy = waveforms.copy()`

Hckjs:
Ok so now the subarray is passed to the `__init__` like in `ImageExtractor`.

Is the supercharge
`super().__init__(config=config, parent=parent, **kwargs)`
necessary here? Had some problems to understand that.

Hopefully everything's ok, thanks for the comments!

Also updated the [Example_Script](https://github.com/Hckjs/Notebooks/blob/master/Example_script_TCDVR.ipynb)

kosack:
seems there is a problem with the master branch right now due to astropy 4.0 - once that's fixed (just wait a bit) we can merge this.

watsonjj:
From my understanding (especially from the recent ASWG call), this is basically ready to be merged, aside from some updates to ensure compatibility with recent PRs. Is this correct @Hckjs?

ctapipe/image/reducer.py
```diff
@@ -17,23 +22,91 @@ class DataVolumeReducer(Component):
     Base component for data volume reducers.
     """
 
-    @abstractmethod
```
kosack, line 20: I think now we should make `DataVolumeReducer` (and thus its children like TailcutsDataVolumeReducer)  inherit from `TelescopeComponent`, and use `TelescopeParameters` for all of its config options. That allows us to set different thresholds for each telescope, which I imagine is needed?  See the implementation of `ImageCleaner` or `ImageExtractor` for examples.

kosack: ah, I just saw @watsonjj 's comment to move that to another PR.  Ok I'll accept this then, and you should open an issue about updating it to use TelescopeComponent

kosack:
You might have to push a small change (for example run `black` on the code to clean up style) to trigger Travis, since we migrated from travis-ci.org to travis-ci.com.

ctapipe/image/reducer.py
Hckjs: All right, thank you! I'll will make another PR then for the Telescope Component!

Hckjs:
@watsonjj Yes i updated it now to the latest PR's. I also made a small change to the used `image_extractor` in the `VolumeReducer`:

- You can choose the used `image_extractor` by handing it in to the `__init__`.
- If you use the `TCDVR` by the `CameraCalibrator` you will always use the same `image_extractor` used in `CameraCalibrator`, except you choose another one.
- If `TCDVR` not used by `CameraCalibrator` and no `image_extractor` is choosen, the default one, which is used, is set to the `NeighborPeakWindowSum`.

ctapipe/image/reducer.py
MaxNoe: No need to make another PR, just update it here

Hckjs: ok then, i'll commit it here.

Hckjs: I set `keep_isolated_pixels=False` for the `tailcuts_clean` like in `ImageCleaner`, since there is no BoolTelescopeParameter isn't it?

Hckjs:
Everything should be included now

ctapipe/image/reducer.py
MaxNoe: Actually, you should just use an image cleaner component as a child component, right? That way, you would automatically expose the options of `ImageCleaner` even when those might change

MaxNoe: Also, adding the `BoolTelescopeParameter` should be trivial, have a look at `ctapipe/core/traits.py` , you should be able to just adapt the flaot and int versions.

ctapipe/image/reducer.py
```diff
@@ -42,5 +116,68 @@ class NullDataVolumeReducer(DataVolumeReducer):
...
     """
 
-    def __call__(self, waveforms):
-        return waveforms
+    def select_pixels(self, waveforms, telid=None, selected_gain_channel=None):
+        mask = waveforms != 0
+        return mask
+
+
+class TailCutsDataVolumeReducer(DataVolumeReducer):
```
MaxNoe, line 124: This shoudl probably just have a child component `TailCutsImageCleaner` like this, you'd inherit the configuration options automatically.

ctapipe/image/reducer.py
```diff
@@ -42,5 +116,68 @@ class NullDataVolumeReducer(DataVolumeReducer):
...
+        )
+
+        # 1) Step: TailcutCleaning at first
+        mask = tailcuts_clean(
+            geom=camera_geom,
+            image=charge,
+            picture_thresh=self.picture_thresh.tel[telid],
+            boundary_thresh=self.boundary_thresh.tel[telid],
+            min_number_picture_neighbors=self.min_number_picture_neighbors.tel[telid],
+            keep_isolated_pixels=False
```
MaxNoe, line 168: This is maybe not a good idea for a data volume reduction right?

We probably want to be conservative here, so not removing bright pixels even if they are isolated might be not a good idea.

ctapipe/image/reducer.py
```diff
@@ -42,5 +116,68 @@ class NullDataVolumeReducer(DataVolumeReducer):
...
+        )
+        pixels_above_boundary_thresh = charge >= self.boundary_thresh.tel[telid]
```
MaxNoe, line 170: Why are you adding pixels above boundary thresholds back here?
Shouldn't this be handled by the tail cuts cleaning already?

Hckjs: I use the `pixels_above_boundary_thresh`-mask here for selecting the pixels in the loop in line 177 (iteratively). The tail cuts cleaning itself would just select the boundaries of one iteration step, but i need also the 'outer' boundaries.

ctapipe/image/reducer.py
Hckjs: Ah yes you're right, i'll swap it to the `BoolTelescopeParaemter` now.

ctapipe/image/reducer.py
MaxNoe: See #1281

ctapipe/image/reducer.py
Hckjs: Alright, added the `TailCutsImageCleaner` as a subcomponent.

Hckjs:
I've also added the BoolTelParam `iteration_steps`:
If set to 'False', the iteration_steps in step 2) are skipped. So the whole reduction would just do a
normal `tailcuts_clean` with a defined amount of dilations at the end with `end_dilates`.

ctapipe/image/reducer.py
```diff
@@ -1,39 +1,113 @@
...
+        **kwargs
+    ):
+        """
+        Parameters
+        ----------
+        config: traitlets.loader.Config
+            Configuration specified by config file or cmdline arguments.
+            Used to set traitlet values.
+            Set to None if no configuration to pass.
+        tool: ctapipe.core.Tool or None
```
MaxNoe, line 39: Lying docs here. Tool is  not a keyword argument anymore.

ctapipe/image/reducer.py
```diff
@@ -1,39 +1,109 @@
...
         Call the relevant functions to perform data volume reduction on the
         waveforms.
 
         Parameters
         ----------
-        waveforms : ndarray
+        waveforms: ndarray
             Waveforms stored in a numpy array of shape
             (n_pix, n_samples).
+        telid: int
```
MaxNoe, line 66: This is called `tel_id` everywhere else I think, we should be consistent

ctapipe/image/reducer.py
```diff
@@ -42,5 +112,55 @@ class NullDataVolumeReducer(DataVolumeReducer):
...
+    1) Select pixels with tailcuts_clean.
+    2) Add iteratively all pixels with Signal S >= boundary_thresh
+       with ctapipe module dilate until no new pixels were added.
+    3) Adding new pixels with dilate to get more conservative.
+    """
+    end_dilates = IntTelescopeParameter(
+        default_value=1,
+        help="Number of how many times to dilate at the end."
+    ).tag(config=True)
+    iteration_steps = BoolTelescopeParameter(
```
MaxNoe, line 133: Iteration steps sound like an Int, to like a bool. 

Maybe call it `do_boundary_dilation`?

ctapipe/image/reducer.py
```diff
@@ -42,5 +112,55 @@ class NullDataVolumeReducer(DataVolumeReducer):
...
+class TailCutsDataVolumeReducer(DataVolumeReducer):
+    """
+    Reduce the time integrated shower image in 3 Steps:
+
+    1) Select pixels with tailcuts_clean.
+    2) Add iteratively all pixels with Signal S >= boundary_thresh
+       with ctapipe module dilate until no new pixels were added.
+    3) Adding new pixels with dilate to get more conservative.
+    """
+    end_dilates = IntTelescopeParameter(
```
MaxNoe, line 129: To make it explicit that this is an integer, let's call it `n_end_dilations`

ctapipe/image/reducer.py
Hckjs: At least in `extractor.py` and `calibrator.py` it's called telid, that's why i write it like this without `_`.
should i rewrite there too?

MaxNoe:
The codacy complaints are invalid (unable to import X)

watsonjj:
@vuillaut Your review is required as you requested changes at some point

Issue #1066, opened 2019-05-02
DL1 Data Processing Tool


kosack:
This is a "parent" issue to track some developments needed to implement proper DL1 data handling in ctapipe.  This is part of a larger issue to make a standard chain (See #977)

## Assumptions:
- DL1 data are in HDF5 format
- development should be synchronized to the *protopipe* development (the final Tool should replace the *protopipe* dl1 script, *dl1-data-handler*, and optionally later also the *lstchain* dl1 script,  if that is useful)

## Tasks:

- [x] refactor basic image calibration code (see #1026 )
- [x] Do gain-selection during  or after calibration and charge extraction (so that we have only a single DL1 image), (See #1026, *additional tasks*)
- [x] Decide on DL1 file format/layout: (discussion: #1059 )
- [x] Implement writing SubarrayDescription to an HDF5 file 
- [x] Implement creating a SubarrayDescription from an HDF5 file (written with prev step)
- [x] Implement standard ctapipe DL0 -> DL1  Tool (based on protopipe, lstchain, dl1-dumper, etc)
    - Metadata tables:
        - [x] simulation run configuration
        - [x] observation configuration 
    - Event tables:
        - [x] monte-carlo events table
        - [x] trigger table
        - [x] standard calibrated image tables  (easy, just needs calib)
        - [x] standard image parameter tables  (requires image processing chain and configuration thereof)
    - Monitoring tables (lower priority for now, but needed later)
        - [x] pointing correction table 
        - [x] calibration monitoring (see implementation in LSTCamEventSource for example)
- [ ] Implement a DL1EventSource that reads output from the DL1 tool described in prev step  so that a DL2 tool can also be created. 
- [x] Ensure the DL1 tool can be configured by a user (via config file and command-line) and that the input and output configuration is tracked and stored in the provenance  metadata (see #1052)
- [x] store full provenance  metadata in output file

kosack:
Just to note, I've made big progress on this and have a prototype working.  During its development, i found a number of small bugs or missing features in ctapipe, so I will first start (in the next week or so) to create pull requests for those changes before committing the full tool.

kosack:
I will close this, since it's all implemented (except for a DL1EventSource, but that can be a separate issue)

PR #1123, opened 2019-09-30
Improvements to Container classes

kosack:
This is the first of a set of PRs that are needed for #1066.  These are just some naming cleanups and new containers (not a full redesign of the event structure yet). It includes:

Deprecation of Fields
----------------------
new `DeprecatedField` class to mark fields that are going away soon

 Normalizing the names of  the Concentration and Leakage fields:
---------------------------------------------
 Writing `ConcentrationContainers` lead to columns named "concentration_concentration_cog", now removed the redundant prefix
- same for leakage parameters: use "leakage" prefix, and remove that prefix from the names.
- renamed `leakage1_pixel` to `pixels_width_1` (similar for `pixels_width_2`), and and `leakage1_intensity` to `intensity_width_1` to be clearer what they mean. These then get written with the prefix as `leakage_intensity_width_1`, for example

Added an ImageParametersContainer
--------------------------------------
Just for collecting all of the various parameter sets. May eventually be replaced with a less hard-coded version, but is useful now.

Added EventIndexContainer and TelEventIndexContainer
---------------------------------------------------------
Store things like `event_id`, `tel_id`, and `obs_id` in one central place, so you can add them to any event list easily.  Deprecated places where they were filled elsewhere.

Added a few new containers for later use:
- added a MorphologyContainer for morphology parameters (calculation will be added in separate PR)
- added a `SimulatedShowerDistribution` container, for storing the sim_telarray thrown shower histograms.

Updated SimTelEventSource
----------------------------
- Fill new index containers correctly
- use pathlib, allow filenames to include `~` or environment variables


ctapipe/io/containers.py
```diff
@@ -519,74 +576,73 @@ class MuonIntensityParameter(Container):
...
-    container_prefix = ''
 
-    leakage1_pixel = Field(
-        nan,
-        'Percentage of pixels after cleaning'
-        ' that are in camera border of width=1'
+    container_prefix = "leakage"
+
+    percent_1pix = Field(
+        nan, "Percentage of pixels after cleaning that are in camera border of width=1"
```
MaxNoe: This is a ratio, not a percentage, right? So 10 % would be 0.1, not 10.

kosack: good point. Should we change it to "fraction_1pix" or multiply by 100?

ctapipe/image/leakage.py
```diff
@@ -43,8 +43,8 @@ def leakage(geom, image, cleaning_mask):
     size = np.sum(image[cleaning_mask])
 
     return LeakageContainer(
-        leakage1_pixel=leakage_pixel1 / geom.n_pixels,
-        leakage2_pixel=leakage_pixel2 / geom.n_pixels,
-        leakage1_intensity=leakage_intensity1 / size,
-        leakage2_intensity=leakage_intensity2 / size,
+        fraction_1pix=leakage_pixel1 / geom.n_pixels,
```
MaxNoe: both are fractions, one is the fraction of pixels, the other the fraction of p.e. So maybe::
```
pixel_fraction_1
pixel_fraction_2
pe_fraction_1
pe_fraction_2
```
?

kosack: how about this?: 

- `leakage_fraction_intensity_1pix`
- `leakage_fraction_intensity_2pix`
- `leakage_fraction_pixels_1pix`
- `leakage_fraction_pixels_2pix`

Or we just define "leakage" as a fraction, and we can remove `_fraction` from all of them? (which brings us back to the first updated version, I think!)

MaxNoe: Maybe width instead of pix to make it clearer? Naming things....

leakage_pixels_width_1
leakage_pixels_width_2
leakage_intensity_width_1
leakage_intensity_width_2

I don't like the `1pix` `2pix` part.

kosack: yeah, I had also considered "neighbors", but that was a bit long... Width is fine.

kosack: though it's not clear the unit of width (not hillas width, for example)

kosack: I updated the docstring (which was basically useless before) to explain it better.

ctapipe/core/container.py
```diff
@@ -28,12 +28,18 @@ def __init__(self, default, description="", unit=None, ucd=None):
...
     def __repr__(self):
-        desc = f'{self.description}'
+        desc = f"{self.description}"
         if self.unit is not None:
-            desc += f' [{self.unit}]'
+            desc += f" [{self.unit}]"
         return desc
 
 
+class DeprecatedField(Field):
```
MaxNoe: This is not used anywhere?

MaxNoe: Shouldn't this raise a warning on something like `__getattr__`?

kosack: yes, I like that suggestion better. I'll change it.

MaxNoe: Maybe only on `__get__`, as otherwise we will be swamped by warnings when setting those values in the event sources etc.

Sou only getting them should raise this warning, right?

MaxNoe: Oh, and this automatically removes these fields from `items` and `fields`, so they will not be saved...

MaxNoe: Mmh. forget it, this does not really work...

kosack:
> The deprecated class does not really work in general unfortunately. 

Ah, yes, that is a problem - perhaps a simple class decorator may be better, but then you get a warning when the Container is constructed, not on access.

kosack:
Do you have a better suggestion? I could just go back to my original idea: A DeprecatedField class that is a subclass of Field and that raises a warning on access? (I just never got around to adding the warning)

MaxNoe:
No, unfortunately I couldn't get it to work.

ctapipe/core/container.py
kosack: Yeah, I think I"ll just revert to the original implementation, but add a warning string.

kosack:
I think we can figure out later how to warn in a nicer way (on use). For now,  I think it's fine just to mark the class as deprecated.

PR #1126, opened 2019-10-02
a few small random cleanups

kosack:
- properly import camera calibrator (avoid * import)
- add `number_of_islands` to `__all__` in `ctapipe.image.cleaning`
- install eventio with conda, not pip (so all packages are conda controlled)
- change pyflakes options to be in line with what Black recommends (so running `black` on a file gives no pyflakes warnings)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -84,3 +84,6 @@ target
 
 examples/notebooks/*.html
 examples/notebooks/*.png
+
+# Visual Studio Code config files
+.vscode
--- a/.gitignore
+++ b/.gitignore
@@ -85,5 +85,3 @@ target
 examples/notebooks/*.html
 examples/notebooks/*.png
 
-# Visual Studio Code config files
-.vscode
--- a/environment.yml
+++ b/environment.yml
@@ -36,6 +36,6 @@ dependencies:
   - wheel
   - xz
   - zlib
-  - eventio~=0.20.1
+  - eventio~=1
 
 
--- a/environment.yml
+++ b/environment.yml
@@ -36,7 +36,6 @@ dependencies:
   - wheel
   - xz
   - zlib
-  - pip:
-    - eventio~=0.20.1
+  - eventio~=0.20.1
 
 
--- a/environment.yml
+++ b/environment.yml
@@ -36,6 +36,7 @@ dependencies:
   - wheel
   - xz
   - zlib
+  - pip
   - pip:
     - eventio~=1.0
 
--- a/environment.yml
+++ b/environment.yml
@@ -36,6 +36,7 @@ dependencies:
   - wheel
   - xz
   - zlib
-  - eventio~=1
+  - pip:
+    - eventio~=1.0
 
 
--- a/setup.cfg
+++ b/setup.cfg
@@ -30,12 +30,9 @@ github_project = cta-observatory/ctapipe
 
 [flake8]
 exclude=
-  astropy_helpers,
-  ah_bootstrap.py,
-  ez_setup.py,
   build,
-  examples/brainstorm,
-  docs
+  docs,
+  .eggs
 max-line-length=90
 #ignore=W291,E303,W391,F403,F401,W503,W1202
 select = C,E,F,W,B,B950
--- a/setup.cfg
+++ b/setup.cfg
@@ -28,13 +28,7 @@ url = https://github.com/cta-observatory/ctapipe
 edit_on_github = False
 github_project = cta-observatory/ctapipe
 
-[pep8]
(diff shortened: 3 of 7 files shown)


environment.yml
```diff
@@ -36,7 +36,6 @@ dependencies:
   - wheel
   - xz
   - zlib
-  - pip:
-    - eventio~=0.20.1
+  - eventio~=0.20.1
```
MaxNoe: You probably want to go to 0.21 since we implemented new containers needed in current prod4 simulations

kosack: Ok, I'll update that.  We will eventually want to be able to read prod4 (or later prod5) even with old versions of ctapipe, so it would be nice to be able to set this to something that is future proof.  What pattern would you suggest to support any future version of eventio up to an API change?  Would just `>=0.20.1` work? or do you expect major changes at a certain version (so would also need `<=0.25` or something?) I would like to avoid the problems we had with ctapipe-0.6.2 where it was tied too tightly to eventio-0.11 and we couldn't read new data with it.

kosack: You might consider releasing the next version not as 0.21, but 1.0.0, so then you can use the first digit for API changes, and second for new features like new data block types, and third for bug fixes

MaxNoe: Ok,  just could release the 1.0

MaxNoe: I released v1.0: https://github.com/cta-observatory/pyeventio/releases

.gitignore
```diff
@@ -84,3 +84,6 @@ target
 
 examples/notebooks/*.html
 examples/notebooks/*.png
+
+# Visual Studio Code config files
+.vscode
```
MaxNoe: Just my opinion, but I think editor related stuff should go in the developers global ignore file, not in each project.

E.g. 

```
$ echo ".vscode" >> $HOME/.gitignore
$ git config --global core.excludesfile $HOME/.gitignore
```

This also holds for OS specific files like `DS_STORE` or `thumbs.db` or `desktop.ini`

setup.cfg
```diff
@@ -42,8 +42,11 @@ exclude=
   build,
   examples/brainstorm,
   docs
-max-line-length=90
-ignore=W291,E303,W391,F403,F401,W503,W1202
+max-line-length=80
+#ignore=W291,E303,W391,F403,F401,W503,W1202
+select = C,E,F,W,B,B950
+ignore = E501,W503,E203
```
MaxNoe: can you comment what these are?

.gitignore
kosack: yes, that's fine. we can remove it.

setup.cfg
kosack: these are the ones recommended by the developer of "black" (essentially matching black's internal rules, so if you blacken a file, you won't get flake8 warnings). The old ones were some rules I ignored in an ad-hoc fashion, but probably not for good reasons.

kosack: the rules can be found here: https://lintlyci.github.io/Flake8Rules/

setup.cfg
```diff
@@ -28,13 +28,7 @@ url = https://github.com/cta-observatory/ctapipe
...
-[pep8]
-exclude=astropy_helpers,ah_bootstrap.py,ez_setup.py,build,examples/brainstorm,docs
-max-line-length=90
-ignore=W291,E303,W391,F403,F401,W503,W1202,E402 
-
-
-[flake8]
+[pycodestyle]
 exclude=
   astropy_helpers,
```
MaxNoe: These files all got removed from the repo

kosack: done.

kosack:
at some point we should just run "black" on all files, to remove most of the style warnings.  It kind of messes up the blame history though.

kosack:
@MaxNoe can you upload eventio=1.0.0  to pypi? I'll make a conda package as well, but that will take some time as usual.

MaxNoe:
It should have been uploaded automatically after I made the release...

MaxNoe:
Ah dam, a missing comma in the classifier list prevented the upload to PyPi

kosack:
> It should have been uploaded automatically after I made the release...

Doesn't seem to be there (the install fails), and `pip search eventio` only shows the 0.21.4

MaxNoe:
@kosack https://github.com/cta-observatory/pyeventio/pull/195

If you approve, I can retag it so it will be uploaded to pypi

MaxNoe:
@kosack, you wanna keep eventio 1.0 via pip or do you change it to a  conda package?

kosack:
> @kosack, you wanna keep eventio 1.0 via pip or do you change it to a conda package?

We can keep it as pip for now - I haven't had time to update the conda package.

PR #1129, opened 2019-10-09
TelescopeParameter Trait

kosack:
This adds classes allow one to specify a single value for all telescopes, or specialize it by telescope type or id.  This is a needed part of #1066.  It works as follows:

```py3
    class SomeComponent(Component):
        tel_param = FloatTelescopeParameter()
        tel_param_int = IntTelescopeParameter()

    comp = SomeComponent()

    comp.tel_param = [
        ("type", "*", 1.0),
        ("type", "LST_LST_LSTCam", 16.0),
        ("type", "*NectarCam", 10.0), # wildcards allowed
        ("id", 16, 10.0),
    ]

    # single value allowed (converted to ("type","*",val) )
    comp.tel_param = 4.5


```

The rules are applied in-order (which is why they are a list, not a dict which does not maintain order in YAML for example). 

These options can also be specified in a config file, for example in JSON like this:
```json
 "TailcutsImageCleaner": {
        "boundary_threshold_pe": [
            ["type", "*", 5.0],
            ["type", "LST_LST_LSTCam", 3.0],
            ["type", "MST_MST_NectarCam", 4.0]
        ],
        "min_picture_neighbors": 2,
        "picture_threshold_pe":  [
                ["type", "*", 10.0],
                ["type", "LST_LST_LSTCam", 6.0],
                ["type", "MST_MST_NectarCam", 6.0],
                ["id", 12, 15.0]
        ]
    },
```

It also introduces another class called `TelescopeParameterResolver` which is constructed with a `TelescopeParameter` and a `SubarrayDescription`.  This class computes the mapping between the rules in the TelescopeParameter when it is constructed, and provides a `resolver.value_for_tel_id(id)` method that gives back the correct value for a given telescope ID.
This 2-step process is needed, since the subarray is not known at configuration time.

### DIFF
--- a/ctapipe/core/traits.py
+++ b/ctapipe/core/traits.py
@@ -138,7 +138,7 @@ class TelescopeParameter(List):
     form: `[(command, argument, value), ...]`.
 
     Command can be one of:
-    - 'default' match all telescopes,  Here argument is ignored.
+    - '*' match all telescopes,  Here argument is ignored.
     - 'type': argument is then a telescope type  string (e.g.
        `('type', 'SST_ASTRI_CHEC', 4.0)`,  to apply to all telescopes of that type
     - 'id':  argument is a specific telescope ID `['id', 89, 5.0]`)
@@ -151,7 +151,7 @@ class TelescopeParameter(List):
 
     .. code-block: python
     [
-        ('default', '', 5.0),                       # default
+        ('*', '', 5.0),                       # default
         ('type', 'LST_LST_LSTCam', 5.2,
         ('type', 'MST_MST_NectarCam', 4.0,
         ('type', 'MST_MST_FlashCam', 4.5,
--- a/ctapipe/core/traits.py
+++ b/ctapipe/core/traits.py
@@ -137,8 +137,8 @@ def has_traits(cls, ignore=("config", "parent")):
 
 class TelescopeParameter(List):
     """
-    Allow a parameter value to be specified as a simple float, or as a list of
-    patterns that match different telescopes.
+    Allow a parameter value to be specified as a simple value (of type *dtype*),
+    or as a list of patterns that match different telescopes.
     The patterns are given as a list of 3-tuples in in the
     form: `[(command, argument, value), ...]`.
 
--- a/ctapipe/core/traits.py
+++ b/ctapipe/core/traits.py
@@ -161,13 +161,13 @@ class TelescopeParameter(List):
 
     def __init__(self, dtype=float, **kwargs):
         super().__init__(**kwargs)
-        if type(dtype) is not type:
+        if not isinstance(dtype, type):
             raise ValueError("dtype should be a type")
         self._dtype = dtype
 
     def validate(self, obj, value):
         # support a single value for all (convert into a default value)
-        if type(value) == self._dtype:
+        if isinstance(value, self._dtype):
             value = [["*", "", value]]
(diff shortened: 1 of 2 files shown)


ctapipe/core/traits.py
```diff
@@ -124,3 +128,135 @@ def has_traits(cls, ignore=("config", "parent")):
     here.
     """
     return bool(set(cls.class_trait_names()) - set(ignore))
+
+
+class TelescopeParameter(List):
+    """
+    Allow a parameter value to be specified as a simple float, or as a list of
```
MaxNoe: Float only by default, so replace with `dtype`? Also `dtype` sounds like  numpy dtype would be supported.

ctapipe/core/traits.py
```diff
@@ -124,3 +128,135 @@ def has_traits(cls, ignore=("config", "parent")):
...
+    patterns that match different telescopes.
+    The patterns are given as a list of 3-tuples in in the
+    form: `[(command, argument, value), ...]`.
+
+    Command can be one of:
+    - '*' match all telescopes,  Here argument is ignored.
```
MaxNoe: Why give it then? Isn't it possible two just use a 2 tuple for `*`?

ctapipe/core/traits.py
```diff
@@ -124,3 +128,135 @@ def has_traits(cls, ignore=("config", "parent")):
...
+
+    Examples
+    --------
+
+    .. code-block: python
+    [
+        ('*', '', 5.0),                       # default
+        ('type', 'LST_LST_LSTCam', 5.2,
+        ('type', 'MST_MST_NectarCam', 4.0,
+        ('type', 'MST_MST_FlashCam', 4.5,
```
MaxNoe: missing parenthese here. Doctest?

ctapipe/core/traits.py
kosack: I decided to make it symmetric, so all commands are 2-tuples (makes it also easier to convert to a table, etc, later).

ctapipe/core/traits.py
kosack: fixed

ctapipe/core/traits.py
kosack: I could get rid of the "*" command altogether, and just replace it with either:
 `("type", "*", 2.5)`
or
`("id", "*", 2.5)`

Would that be better? So far only "*" or a type or id would be allowed, but we could later implement true globs, like `("type","LST*", 2.5)`

ctapipe/core/traits.py
kosack: generally the base class is not to be used, I added specializations like FloatTelescopeParameter and IntTelescopeParameter. Numpy dtypes should in principle work, so I think it's ok to keep the name. 

Yes, I will update the docstring

kosack:
Ok, now wildcard globs are allowed in the *type* argument, so you can match all MSTs regardless of camera with `("type", "MST*", 15.0)` for example.

kosack:
@MaxNoe are you happy with the latest changes? I'd like to merge this soon to be able to start to merge the DL1 writer parts which depend on it.  Or we can discuss it monday if not.

MaxNoe:
@kosack looks good to me

MaxNoe:
@kosack I'm will not be in Bologna btw.

PR #1131, opened 2019-10-14
Cleaning for biggest cluster

HealthyPear:
In an effort to minimize the use of external libraries in _protopipe_ (see issues cta-observatory/protopipe#5 and cta-observatory/protopipe#6), I implemented a function to clean just for the biggest cluster in an image.
I also implemented the associated unit test, which returned a positive result during my local tests.

In the definition of the function, there is a numpy array conversion which I propose to move into the recently added function _number_of_islands_.


ctapipe/image/cleaning.py
```diff
@@ -355,3 +355,43 @@ def fact_image_cleaning(
         geom, pixels_to_keep, arrival_times, min_number_neighbors, time_limit
     )
     return pixels_to_keep
+
+
+def biggest_island(geom, image, mask):
```
MaxNoe: I would recommend to make this smaller. Give this function just the labels and return a new mask with fewer `Trues`, thus this function gets smaller and is more general. Also  `num_islands` will be calculated twice if you also use it in the regular analysis.

So I would change this to:
```python
def largest_island(labels):
    return labels == np.argmax(np.bincount(labels[labels > 0]))  # no need for where
```

ctapipe/image/cleaning.py
```diff
@@ -355,3 +355,43 @@ def fact_image_cleaning(
...
+
+    selected_geom : `ctapipe.instrument.CameraGeometry`
+        Camera geometry information filtered for the biggest island
+    selected_image : array
+        Pixel values of the calibrated image filtered for the biggest island.
+
+    """
+    # Finds all islands from the cleaned pixels and label them
+    num_islands, labels = number_of_islands(geom, mask)
+    labels = labels.astype("int64")  # this is temporaneus: I think it's better
```
MaxNoe: Why? Wat is the dtype before?

HealthyPear: From my tests I could see that it was float64, so then when I used numpy.argmax, it would say that it was expecting int64 instead.

MaxNoe: Okay, then I agree that the Change should be made in num_islands directly

ctapipe/image/cleaning.py
HealthyPear: Thanks for this input.
Let me just understand better:

- I don't understand the "fewer Trues" part, for me the only Trues are always those related to the largest island and the mask reflects always the original geometry,

- I don't understand why _number_of_islands_, should be called twice in the regular analysis (do you mean _protopipe_?)

- you suggest that in the regular analysis I call first _number_of_islands_, then _largest_island_ and finally I select the geometry and image instead of doing everything inside _biggest_island_,

ctapipe/image/cleaning.py
MaxNoe: I checked, for me the dtype is int32, so no reason to change this. Where did you encounter floats?

ctapipe/image/cleaning.py
MaxNoe: * Yes, this was a bit confusing, I meant return a mask where only the pixels in the largest group are `True`
* number_of_islands is a useful paramter for G/H separation, that should be stored into DL1.2, so if you call it inside largest_island without being able to get the number of islands, you need to call it again to be able to store the value.
* Yes, exactly. Small, composable functions!

ctapipe/image/cleaning.py
HealthyPear: I tested it in a separate jupyter notebook.
I placed the first 4 prints inside _biggest_island_, just after the call to _number_of_islands_, in particular, the last one is,

`print(f"dtype of labels is {labels.dtype}")`

I then remove the line in which I forcibly convert the labels to int64.
 The output of the prints is the following:

```
There are 4 islands.
Island labels:
[1. 1. 0. 0. 2. 2. 2. 2. 0. 3. 3. 0. 0. 0. 4. 4. 4.]
dtype of labels is float64
```

Then I get the following error when trying to select the biggest islands' label,

```
<ipython-input-10-51e0d6b9b8e0> in <module>
     21 
     22 # Apply function to test
---> 23 biggest_cam, biggest_image = biggest_island(cam, image, clean_mask)
     24 
     25 # Check if the function recovers the truth

<ipython-input-9-b8099df578f5> in biggest_island(geom, image, mask)
     72 
     73     # Get the label of the biggest island and create a new mask
---> 74     biggest_label = np.argmax(np.bincount(labels[np.where(labels>0)]))
     75     print(f"The biggest one is #{biggest_label}.")
     76     biggest_mask = labels == (biggest_label)

<__array_function__ internals> in bincount(*args, **kwargs)

TypeError: Cannot cast array data from dtype('float64') to dtype('int64') according to the rule 'safe'
```

ctapipe/image/cleaning.py
HealthyPear: > * Yes, this was a bit confusing, I meant return a mask where only the pixels in the largest group are `True`

This was already happening in my function, am I wrong?  _biggest_mask_ is a mask with the same dimensions of the original camera, but only the pixels belonging to the biggest island are set to True (`biggest_mask = labels == biggest_label`).

> * number_of_islands is a useful paramter for G/H separation, that should be stored into DL1.2, so if you call it inside largest_island without being able to get the number of islands, you need to call it again to be able to store the value.

OK, I understand now.

> * Yes, exactly. Small, composable functions!

Yes, I agree.

kosack: > I don't understand why number_of_islands, should be called twice in the regular analysis (do you mean protopipe?)

For example, we will include the number of islands in the DL1 parameters (so we should avoid computing it twice)

ctapipe/image/cleaning.py
MaxNoe: That's strange. What are your input values? For it always returns int32

HealthyPear: The ones from the unit test:

```
    # Create a simple camera
    cam = CameraGeometry.make_rectangular(17, 1)
    cam.cam_id = "test"  # this overwrites the default -1 set by the function
    # Create a simple image made of ones and zero "photoelectrons"
    image = np.array([2, 1, 1, 1, 1, 2, 2, 1, 0, 2, 1, 1, 1, 0, 2, 2, 2])
    # Let's say that we clean with a simple tailut with the following parameters
    # p = 2, b = 1, min neighbors=0
    # This cleaning would find the following islands
    clean_mask = np.zeros(cam.n_pixels).astype("bool")
    clean_mask[[0, 1]] = 1
    clean_mask[[4, 5, 6, 7]] = 2  # this is the biggest
    clean_mask[[9, 10]] = 3
    clean_mask[[14, 15, 16]] = 4
```

Then I call the function and get the error,

`biggest_cam, biggest_image = biggest_island(cam, image, clean_mask)`

HealthyPear: Just to clear things up, the following is a minimal example:

```
cam = CameraGeometry.make_rectangular(5, 1)
image = np.array([0,0,1,1,1])
clean_mask = np.zeros(cam.n_pixels).astype("bool")
clean_mask[[2, 3, 4]] = 1 # the only and largest island
num_islands, labels = number_of_islands(cam, clean_mask)
print(f"There are {num_islands} islands.")
print(f"Islands labelled : {labels}")
print(f"dtype of labels is {labels.dtype}")
```

which returns the following output:

```
There are 1 islands.
Islands labelled : [0. 0. 1. 1. 1.]
dtype of labels is float64
```

HealthyPear: To be sure, I just replicated this result using only the conda version of ctapipe 0.7.

If it can be of some help, in this case when importing _number_of_islands_ I get the following NumPy warning,

```
/Users/michele/Applications/anaconda3/envs/protopipe/lib/python3.7/site-packages/corsikaio/subblocks/dtypes.py:20: FutureWarning: Passing (type, 1) or '1type' as a synonym of type is deprecated; in a future version of numpy, it will be understood as (type, (1,)) / '(1,)type'.
  return np.dtype(dict(**dt))
```

MaxNoe: Found it. Can you add, dtype='int32' to the call to np.zeros in num islands?

HealthyPear: Yes, in this way the problem is solved at the very beginning.

ctapipe/image/cleaning.py
```diff
@@ -356,3 +356,26 @@ def fact_image_cleaning(
...
+        Flattened array containing a list of labelled islands from a cleaned image.
+        Second returned value of the function 'number_of_islands'.
+
+    Returns
+    -------
+
+    islands_labels : array
+        A boolean mask created from the input labels and filtered for the largest island.
+
+    """
```
MaxNoe: See cta-observatory/protopipe/issue/15, check if there are even islands, e.g.:

```
valid = island_labels > 0
if np.count_nonzero(valid) == 1:
     return np.zeros(island_labels.shape, dtype='bool')
```
```

Comment (HealthyPear) [reply ctapipe/image/cleaning.py]: Ok, I understand what you mean.
Still, I would like to elaborate a bit on this.

`largest_island` depends on `number_of_islands`: in principle, there is no reason to call it if there are no islands.

Do you think it is more efficient to check for the existence of the islands inside `largest_island` or outside by using the already computed number by your function? In the end, both solutions use a conditional, am I right?

Also, there is no reason to call `largest_island` in case of only one island surviving the cleaning, since the original cleaning mask has already only one island.

The way I fixed the bug in protopipe follows this principle by using the ctapipe functions in a modular way: first I call `number_of_islands` and then I check how many there are and I call `largest_island` only if necessary.

Naturally, if we find a more efficient way to do it even better: I will call the ctapipe functions anyway once the new stable release is out.

Comment (MaxNoe) [reply ctapipe/image/cleaning.py]: I for myself doubt that it is even a good idea to calculate hillas parameters only on the largest island.

Do you have comparisons?

Comment (MaxNoe) [reply ctapipe/image/cleaning.py]: Anyways, this function should either return an empty mask if no island is there or it should raise a meaningfull error message. The error message from deep in numpy not indication the real problem is not helpful for someone using this function.

Comment (HealthyPear) [reply ctapipe/image/cleaning.py]: > I for myself doubt that it is even a good idea to calculate hillas parameters only on the largest island.
> 
> Do you have comparisons?

This is a very good question! The work we are doing now with CTA-MARS and in the near future should shed some light on this.

Comment (HealthyPear) [reply ctapipe/image/cleaning.py]: > Anyways, this function should either return an empty mask if no island is there or it should raise a meaningfull error message. The error message from deep in numpy not indication the real problem is not helpful for someone using this function.

Ok, I get it.
I'll will improve it to catch the case with no islands and the case with one should be automatic.

Thanks for the input!

Comment (vuillaut) [reply ctapipe/image/cleaning.py]: > I for myself doubt that it is even a good idea to calculate hillas parameters only on the largest island.
> 
> Do you have comparisons?

It should be for the direction reconstruction only.
Don't do it for the energy or classification.

Comment (HealthyPear) [reply ctapipe/image/cleaning.py]: In your implementation, why did you use a boolean instead of using directly the labels array?

Meaning, why didn't you just do the following?

```
if np.count_nonzero(island_labels) == 0:
    return np.zeros(island_labels.shape, dtype='bool')
```

HealthyPear: > > I for myself doubt that it is even a good idea to calculate hillas parameters only on the largest island.
> > Do you have comparisons?
> 
> It should be for the direction reconstruction only.
> Don't do it for the energy or classification.

Yes, this WOULD be the case for protopipe.
For energy estimation and g/h classification, we should use the content of the whole set of islands, but for the moment the cleaning is the same.
The reason is that the performance of these two cleanings has not been studied in model estimation with protopipe.
I hope that in the future this part will be studied in more detial, but for the moment this is what we have.

MaxNoe: > Meaning, why didn't you just do the following?

Did not think about it. Should be fine!

HealthyPear:
> LGTM

I had to go on Urban Dictionary....I feel so old.

MaxNoe:
> >    LGTM
> I had to go on Urban Dictionary....I feel so old.

I don't think that's an "age thing". More a "used to code review" thing

PR #1133, opened 2019-10-16
Use count_nonzero instead of sum for bool arrays

MaxNoe:
This came up on workshop I lectured at.

Kind of surprises me, but count_nonzero is much faster than sum for counting the True's in an array.

```
In [11]: mask = np.random.randint(0, 1, 2000).astype(bool)

In [12]: %timeit np.sum(mask)
7.7 µs ± 895 ns per loop (mean ± std. dev. of 7 runs, 100000 loops each)

In [13]: %timeit np.count_nonzero(mask)
1.05 µs ± 99.3 ns per loop (mean ± std. dev. of 7 runs, 1000000 loops each)
```

### DIFF
--- a/ctapipe/image/leakage.py
+++ b/ctapipe/image/leakage.py
@@ -34,8 +34,8 @@ def leakage(geom, image, cleaning_mask):
     mask1 = border1 & cleaning_mask
     mask2 = border2 & cleaning_mask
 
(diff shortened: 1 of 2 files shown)


MaxNoe:
Apparently `sum` casts to `long` for boolean input, which it why this is slower.

PR #1137, opened 2019-10-17
Also test ctapipe with a pure pip installation (no conda)

MaxNoe:
This tries adding travis runs that use normal python with pip instead of conda


MaxNoe:
I rebased this to provide better reviewable diffs for each commit and to remove my travis config adaption madness. I would recommend to not squash this PR as the commits touch several different fields but were required to make it run with the current ctapipe-extra master.

MaxNoe:
Okay, the last remaining issue is now that the tests are failing because the newest `LSTCam-003` has a different pixel order than `LSTCam`.

I think these should be named differently in any case, to make sure which one is from the simtel files and which is the real one.

What do you think @kosack, @cta-observatory/ctapipe-core-devs

MaxNoe:
I could also install an older version of ctapipe-extra for now and only do the travis config

MaxNoe:
Okay, finally this is goiing green.

I have two motivations for this
* conda somedays creates as much problems as it solves on other days, so by testing both on conda and on pure python/pip we can see if something broke on our end our on conda's
* I wanted this to have before uploading ctapipe to pypi so that we are sure the setup.py is complete, I found some missing requirements, especially for the tests and building the docs

This should be ready for merge now and should be squashed.

I removed the discussion about CameraGeometries / Versions and will open another PR with that as soon as this is merged here.

kosack:
are the changes to the example notebook in there for a reason?

MaxNoe:
> are the changes to the example notebook in there for a reason?

Yes, for some reason the version of the notebook converter installed with pip complained about the alternative text (it was a filename and it complaint the the file does not exist).

Issue #876, opened 2018-12-11
Move MC pointing information


kbruegge:
For now we move it to `event.pointing.tel`. This should already be a SkyCoordinate in the HorizonFrame.

MaxNoe:
Done a long time ago. Now cleaning up some remaining examples.

PR #1138, opened 2019-10-17
Source visualisation in camera

vuillaut:
Add function to overlay a point source in the camera 
```
display = CameraDisplay(geom, image)
fig = display.axes.get_figure()

event_pos = event_pos_in_camera(event, tel_id)
display.overlay_point_source(event_pos[0], event_pos[1])
```

image

### DIFF
--- a/ctapipe/coordinates/transforms.py
+++ b/ctapipe/coordinates/transforms.py
@@ -1,0 +1,36 @@
+from astropy.coordinates import SkyCoord, AltAz
+from ..coordinates import CameraFrame
+import astropy.units as u
+
+
+
+def event_pos_in_camera(event, tel_id, horizon_frame=AltAz()):
+    """
+    Return the position of the source in the camera frame
+
+    Parameters
+    ----------
+    event: `ctapipe.io.containers.DataContainer`
+    tel_id: int - id of the telescope
+    horizon_frame: `astropy.coordinates.builtin_frames.altaz.AltAz`
+
+    Returns
+    -------
+    (x, y) (float, float): position in the camera
+    """
+
+    focal = event.inst.subarray.tel[tel_id].optics.equivalent_focal_length
+
+    telescope_pointing = SkyCoord(alt=event.mc.tel[tel_id].altitude_raw * u.rad,
+                                  az=event.mc.tel[tel_id].azimuth_raw * u.rad,
+                                  frame=horizon_frame)
+
+    event_direction = SkyCoord(alt=event.mc.alt,
+                               az=event.mc.az,
+                               frame=horizon_frame)
+
+    camera_frame = CameraFrame(focal_length=focal,
+                               telescope_pointing=telescope_pointing)
+
+    camera_pos = event_direction.transform_to(camera_frame)
+    return camera_pos.x, camera_pos.y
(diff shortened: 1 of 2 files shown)


ctapipe/coordinates/transforms.py
```diff
@@ -0,0 +1,36 @@
+from astropy.coordinates import SkyCoord, AltAz
+from ..coordinates import CameraFrame
+import astropy.units as u
+
+
+
+def event_pos_in_camera(event, tel_id, horizon_frame=AltAz()):
+    """
```
kosack: this function name makes it seem like it's the position of an event in the camera, not the source position.  Is that really what you want ?  Also this will be *very* slow if you plot many events, since it constructs the camera frame each time. 

I think it may be better to store a constructed frame in either the CameraDisplay or the CameraGeometry so that it can be re-used.  Something like `display.frame`, then you can just update its' parameters as needed (if the pointing changes).    This also reminds me to clean up the place where pointings are stored, and to use SkyCoords for positions.

ctapipe/coordinates/transforms.py
```diff
@@ -0,0 +1,36 @@
...
+    -------
+    (x, y) (float, float): position in the camera
+    """
+
+    focal = event.inst.subarray.tel[tel_id].optics.equivalent_focal_length
+
+    telescope_pointing = SkyCoord(alt=event.mc.tel[tel_id].altitude_raw * u.rad,
+                                  az=event.mc.tel[tel_id].azimuth_raw * u.rad,
+                                  frame=horizon_frame)
+
```
kosack: this is the alt/az location of the *event*  in the sky not the telescope, right? I would think you want to use the telescope pointing.  In the case of diffuse simulations, this would change for each event and is not what you want.

MaxNoe: This should be the pointing (it's mc.tel[tel_id]... not mc.altitdude)

vuillaut: It is the telescope pointing.
Pointing should be in `event.pointing.tel[tel_id]` but this is not the case for MC and right now telescope pointing seems to be in `event.mc.tel[tel_id].altitude_raw`.
The event altitude is in `event.mc.alt`.

ctapipe/coordinates/transforms.py
MaxNoe: >  Also this will be very slow if you plot many events, since it constructs the camera frame each time.

Instantiating an `AltAz` frame takes < 1ms on my machine. Why do you think, this will slow down plotting?

MaxNoe: CameraFrame instantiation takes 200 µs

ctapipe/coordinates/transforms.py
kosack: let's fix the SimTelEventSource to put it in pointing then

MaxNoe: This is a long standing issue, see here: https://github.com/cta-observatory/ctapipe/issues/876

kosack:
What happens here if the camera is in the EngineeringCameraFrame? 
Perhaps we should attach a Frame as an attribute of a CameraDisplay (so it knows what frame it is in), and then any function like this would be something like:
```py3
def overlay_point_source(self, coordinate):
     coord_local = coordinate.transform_to(self.frame)
     ...
```

HealthyPear:
I think this PR would benefit a lot now from an update from the master branch

maxnoe:
Closing in favor of #2203

PR #1141, opened 2019-10-22
Pointing container

vuillaut:
Filling pointing container from MC data and using it in tests analysis
Then the pointing information is the corrected one.

### DIFF
--- a/ctapipe/io/hessioeventsource.py
+++ b/ctapipe/io/hessioeventsource.py
@@ -161,14 +161,23 @@ class HESSIOEventSource(EventSource):
                                                        get_azimuth_raw(tel_id))
                     data.mc.tel[tel_id].altitude_raw = (file.
                                                         get_altitude_raw(tel_id))
-                    data.mc.tel[tel_id].azimuth_cor = (file.
-                                                       get_azimuth_cor(tel_id))
-                    data.mc.tel[tel_id].altitude_cor = (file.
-                                                        get_altitude_cor(tel_id))
-                    data.pointing[tel_id].altitude = (data.mc.tel[tel_id].altitude_raw
-                                                      - data.mc.tel[tel_id].altitude_cor) * u.rad
-                    data.pointing[tel_id].azimuth = (data.mc.tel[tel_id].azimuth_raw
-                                                     - data.mc.tel[tel_id].azimuth_cor) * u.rad
+                    azimuth_cor = file.get_azimuth_cor(tel_id)
+                    altitude_cor = file.get_altitude_cor(tel_id)
+
+                    # hessioeventsource pass 0 if there is no altitude/azimuth correction
+                    if azimuth_cor == 0 and data.mc.tel[tel_id].azimuth_raw != 0:
+                        data.mc.tel[tel_id].azimuth_cor = np.nan
+                        data.pointing[tel_id].azimuth = u.Quantity(data.mc.tel[tel_id].azimuth_raw, u.rad)
+                    else:
+                        data.mc.tel[tel_id].azimuth_cor = azimuth_cor
+                        data.pointing[tel_id].azimuth = u.Quantity(azimuth_cor, u.rad)
+
+                    if altitude_cor == 0 and data.mc.tel[tel_id].altitude_raw != 0:
+                        data.mc.tel[tel_id].altitude_cor = np.nan
+                        data.pointing[tel_id].altitude = u.Quantity(data.mc.tel[tel_id].altitude_raw, u.rad)
+                    else:
+                        data.mc.tel[tel_id].altitude_cor = altitude_cor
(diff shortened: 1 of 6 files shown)


ctapipe/io/hessioeventsource.py
```diff
@@ -165,6 +165,11 @@ def _generator(self):
                                                        get_azimuth_cor(tel_id))
                     data.mc.tel[tel_id].altitude_cor = (file.
                                                         get_altitude_cor(tel_id))
+                    data.pointing[tel_id].altitude = (data.mc.tel[tel_id].altitude_raw
+                                                      - data.mc.tel[tel_id].altitude_cor) * u.rad
```
MaxNoe: Are you sure that `altitude_cor` is the the *correction for the altitude* and not the *corrected* altitude?

MaxNoe: Also, you should use `u.Quantity(..., u.rad)`, it's faster than multiplying the unit.

MaxNoe: See line 1018ff of `io_hess.h`:

```
struct hess_tracking_event_data_struct
{
   int tel_id;            ///< The telescope ID number (1 ... n)
   double azimuth_raw;    ///< Raw azimuth angle [radians from N->E].
   double altitude_raw;   ///< Raw altitude angle [radians].
   double azimuth_cor;    ///< Azimuth corrected for pointing errors.
   double altitude_cor;   ///< Azimuth corrected for pointing errors.
   int raw_known;         ///< Set if raw angles are known.
   int cor_known;         ///< Set if corrected angles are known.
};
```

MaxNoe: I remember that we set it to zero to be compatbible in the simteleventsource to be compatible with the hessioeventsource, but I think it would be better to use `NaN` if it is not available at all.

vuillaut: Ha. Thanks for the info.
I'll propose some corrections then.

MaxNoe: See that copy paste error in the comment ;-)

vuillaut:
@MaxNoe @kosack 
3 lines were > 100 characters...

ctapipe/io/hessioeventsource.py
```diff
@@ -161,10 +161,27 @@ def _generator(self):
...
-                                                       get_azimuth_cor(tel_id))
-                    data.mc.tel[tel_id].altitude_cor = (file.
-                                                        get_altitude_cor(tel_id))
+                    azimuth_cor = file.get_azimuth_cor(tel_id)
+                    altitude_cor = file.get_altitude_cor(tel_id)
+
+                    # hessioeventsource pass 0 if there is no altitude/azimuth correction
+                    if azimuth_cor == 0 and data.mc.tel[tel_id].azimuth_raw != 0:
+                        data.mc.tel[tel_id].azimuth_cor = np.nan
+                        data.pointing[tel_id].azimuth = \
```
kosack: Please use () rather than `\` for continuation.  Actually, perhaps just run "black" on this before committing - it tends to fix all style problems.  I think we should make that a policy (or even commit hook) anyhow, to avoid style changes.  The default settings.

MaxNoe: In the simteleventsour whe have `pointing = data.pointing[tel_id]` and `mc = `data.mc.tel[tel_id]` in the telescope loop so the lines get much shorter and more readable. That's something black won't do ;-)

kosack: yes, also simplifying the long lines is a better way to go! .  Even just defining `tel = data.mc.tel[tel_id]`, and `pointing = data.pointing[tel_id]`, and then using those  references would make this easier to follow.

vuillaut: Ok I cleaned the whole loop following @MaxNoe suggestion. Looks more clear.

ctapipe/io/hessioeventsource.py
```diff
@@ -130,57 +130,52 @@ def _generator(self):
...
-                                                           get_ref_step(tel_id))
-                    data.mc.tel[tel_id].time_slice = (file.
-                                                      get_time_slice(tel_id))
-                    data.mc.tel[tel_id].azimuth_raw = (file.
-                                                       get_azimuth_raw(tel_id))
-                    data.mc.tel[tel_id].altitude_raw = (file.
-                                                        get_altitude_raw(tel_id))
+                    mc.photo_electron_image = hessio_mc_npe
+                    mc.meta['refstep'] = (file.get_ref_step(tel_id))
+                    mc.time_slice = (file.get_time_slice(tel_id))
```
MaxNoe: unneccessary parenthese

vuillaut: fixed

Issue #1044, opened 2019-04-09
Waveform sample rate


watsonjj:
Currently the only place to obtain the sample rate is from inside the mc container (`event.mc.time_slice`). This information is required to correctly calculate the pulse time. 

Where should this parameter (and other similar waveform properties such as a reference pulse shape) be stored?

kosack:
I guess in principle, it should be inside the instrument description, so in `CameraGeometry`.  It's a part of how you define the camera.  It could go in a more technical camera class (that we will add later, with module info, etc), but I think it's so fundamental to how you treat a camera image sequence that it should go in `CameraGeometry`.

kosack:
I'd suggest adding an attribute `CameraGeometry.sampling_rate` that defines it in MHz, and that field should also end up in the table metadata when you use `CameraGeometry.to_table()`

kosack:
in addition (or instead of that), the waveforms in R0-DL0 could have units attached (divide by sampling_rate, and convert to ns), but one still needs to store the sampling rate somewhere for that to work.

watsonjj:
> in addition (or instead of that), the waveforms in R0-DL0 could have units attached (divide by sampling_rate, and convert to ns), but one still needs to store the sampling rate somewhere for that to work.

Wasn't this decided against in the core developers meeting? (#684)

I agree with the decision, due to the overhead it adds to the waveform processing. We should just correctly consider the sampling rate in the charge and time extraction.

kosack:
Ok, agreed, no units attached to the waveforms, but the sampling_rate in camera should still have units.

watsonjj:
Resolved by #1142 and #1227

PR #1142, opened 2019-10-23
Add sampling_rate to instruments description

watsonjj:


watsonjj:
At the moment this has the problem that the sampling rate isn't stored in the camgeom file from ctapipe-resources. Is there an easy way to add this information to the files?

MaxNoe:
Why should it be stored in a camgeom?

watsonjj:
Because the only place that somewhat makes sense to store it that exists at the moment is `CameraGeometry`. I agree that the sampling rate is not really part of the `CameraGeometry`...

watsonjj:
This was requested in #1044

ctapipe/instrument/camera.py
```diff
@@ -341,6 +344,7 @@ def from_table(cls, url_or_table, **kwargs):
             pix_y=tab['pix_y'].quantity,
             pix_area=tab['pix_area'].quantity,
             pix_type=tab.meta['PIX_TYPE'],
+            sampling_rate=u.Quantity(tab.meta["SAMP_RATE"], u.GHz),
```
kosack: these are stored in FITS, and that keyword is longer than 8 characters so should be "SAMPRATE" or "SAMPFREQ" perhaps

kosack: in the case that this info is missing, perhaps fall back to a default (until we update the input files), so things work as before: i.e.:
```py3
tab.meta.get("SAMP_RATE", u.Quantity(1,u.GHz))
```

kosack: or better yet, issue a warning that the sampling rate is unknown and defaulting to 1.0 GHz

kosack: once this works properly, we can just run `ctapipe-dump-instrument` on a file with all CTA cameras and use the outputs to update ctapipe-extra

ctapipe/instrument/camera.py
```diff
@@ -309,6 +312,7 @@ def to_table(self):
                                TAB_TYPE='ctapipe.instrument.CameraGeometry',
                                TAB_VER='1.0',
```
kosack: could you bump this to 1.1, since this changes the format slightly?

kosack:
> Maybe it's time to rename that class when we use it not only to describe geometry?

We had a small off-line discussion about that - I think it's ok for now if you consider the time-axis as part of the "Geometry", but it's true if later we start to add things that are more technical to it, we either need another class, or to rename it to just CameraDescription.

ctapipe/io/hessioeventsource.py
```diff
@@ -226,6 +226,7 @@ def _build_telescope_description(self, file, tel_id):
         num_tiles = file.get_mirror_number(tel_id)
         cam_rot = file.get_camera_rotation_angle(tel_id)
         num_mirrors = file.get_mirror_number(tel_id)
+        sampling_rate = u.Quantity(1 / file.get_time_slice(tel_id), u.GHz)
```
thomasarmstrong: Sorry, working on a temporary desk top, could you tell me what get_time_slice returns here? Is it the length of the readout? I just worry that if it is just the number of samples then it would only convert to GHz assuming that the sampling rate itself is 1GHz (i.e. does the function take into account the FADC_MHz and FADC_SUM_BINS parameters from sim_telarray)?

watsonjj: From pyhessio:
```
    def get_time_slice(self, telescope_id):
        """
        Returns Width of readout time slice (i.e. one sample) [ns].
         If telescope_id is not valid return 0.
        Parameters
        ----------
        telescope_id: int
            telescope's id
        Returns float
        """
        return self.lib.get_time_slice(telescope_id)
```

thomasarmstrong: OK, it should perform as expected then.

Issue #1127, opened 2019-10-04
Scale of the pulse time channel correct? 


TjarkMiener:
I was modifying the [DL1DH writer](https://github.com/cta-observatory/dl1-data-handler/tree/update_ctapipe) to make it compatible with the newest ctapipe release. As discussed in cta-observatory/dl1-data-handler#74, it seems that the scale of pulse time channel is not correct (not in nanoseconds). However, the charge channel looks good to me. I also tested for FlashCam and DigiCam, which are showing the same behaviour. So maybe I'm messing something up with the configurations of the `CameraCalibrater` or the `GainSelector`?

watsonjj:
The pulse time is currently extracted in units of the sampling rate. If the sampling rate is per ns, then the units are ns.

The method in `extract_pulse_time_around_peak` extracts the weighted average of the samples around the peak. This can result in extreme values for pixels with no signal. There is no limit applied to the return value. Perhaps we wish to constrain the value to be within the readout window. Or perhaps we want to consider a more robust method.

The time image should look logical after tailcut cleaning. Or the pulse time can be constrained in the analysis.

Finally, the pulse time does need to be converted to ns for cameras that don't sample per ns. I will create a PR for this.

MaxNoe:
> This can result in extreme values for pixels with no signal. 

Why? A weighted average should never produce values outside min / max of the samples. So how can this produce values outside of the readout window?

watsonjj:
> A weighted average should never produce values outside min / max of the samples

Sure it can, if the values being averaged seem to be decreasing or increasing (almost like they were on the edge of a pulse) then the average will be outside the range. If these values are not on the edge of the pulse, and are just a symptom of waveform noise/variations, then the resulting weighted average can be way outside the range.

```
import numpy as np
from matplotlib import pyplot as plt

sample_i = np.array([17, 18, 19, 20, 21, 22])
value = np.array([3.970581  ,  0.22058105,  1.470581  ,  3.0881958 , -3.8970947 , -4.779419])
plt.plot(sample_i, value)
print(np.average(sample_i, weights=value))
### -351.5269059715945
```

![image](https://user-images.githubusercontent.com/17825673/67369059-a4b95200-f578-11e9-95c5-c487df89ef12.png)

vuillaut:
[...]

PR #1143, opened 2019-10-23
Improve extract_pulse_time_around_peak to ignore negative samples

watsonjj:
Created to address #1127

### DIFF
--- a/ctapipe/image/tests/test_extractor.py
+++ b/ctapipe/image/tests/test_extractor.py
@@ -103,7 +103,7 @@ def test_neighbor_average_waveform(camera_waveforms):
     assert_allclose(average_wf[0, 48], 98.565743, rtol=1e-3)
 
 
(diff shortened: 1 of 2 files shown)


ctapipe/image/tests/test_extractor.py
```diff
@@ -112,6 +112,13 @@ def test_extract_pulse_time_around_peak(camera_waveforms):
     assert_allclose(pulse_time[0], 41.2, rtol=1e-3)
 
 
+def test_extract_pulse_time_around_peak_within_range(example_event):
+    telid = list(example_event.r0.tel)[0]
+    waveforms = example_event.r1.tel[telid].waveform
```
kosack: it may be nicer to define some  dummy waveforms here, where you know the real peak, so this doesn't change when we update the test event file. That way you can also check that you reconstruct the "real" peak

PR #1144, opened 2019-10-23
Remove 3-dimension extractors

watsonjj:
Remove the extractor functionality to handle the 3-dimensional waveforms.

These were to handle waveforms with a gain channel dimension. This functionality was marked as deprecated following the gain-selection PR.


kosack:
I guess in the case we need to analyze calibration runs with 2 gains, we can just do 2 loops, with 2 gainselectors and run the normal extractor

PR #1150, opened 2019-10-25
added DOI for release v0.7.0

kosack:


MaxNoe:
Did you do this by hand or did you setup automatic doi creation on tags?

vuillaut:
> Did you do this by hand or did you setup automatic doi creation on tags?

I think it does since ctapipe v0.7.0 got released automatically.

PR #1154, opened 2019-10-25
add Taubin fit

momorning:
Hello, I am working with Alison Mitchell at UZH this semester.

Adding a new muon ring fit algorithm, called Taubin fit. 
The fit was referred to: https://www.zeuthen.desy.de/students/2018/Summerstudents2018/Reports/Pillera.pdf (eqn 9).
I will be working on adding a test file to show how it works.

### DIFF
--- a/ctapipe/image/muon/fitting.py
+++ b/ctapipe/image/muon/fitting.py
@@ -487,8 +487,10 @@ def taubin_circle_fit(
     x_masked = x[mask]
     y_masked = y[mask]
 
-    taubin_r_initial = x.max() / 2
-    taubin_error = taubin_r_initial * 0.1
+    R = x.max()  # x.max() just happens to be identical with R in many cases.
+
+    taubin_r_initial = R / 2
+    taubin_error = R * 0.1
     xc = 0
     yc = 0
 
@@ -501,9 +503,9 @@ def taubin_circle_fit(
         error_xc=taubin_error,
         error_yc=taubin_error,
         error_r=taubin_error,
-        limit_xc=(x.min(), x.max()),
-        limit_yc=(y.min(), y.max()),
-        limit_r=(0, 2*x.max()),
+        limit_xc=(-2*R, 2*R),
+        limit_yc=(-2*R, 2*R),
+        limit_r=(0, R),
         pedantic=False
     )
     fit.migrad()
--- a/ctapipe/image/muon/fitting.py
+++ b/ctapipe/image/muon/fitting.py
@@ -455,12 +455,7 @@ def _impact_parameter_chisq(params, phi, hist, mirror_radius):
     return np.sum((hist - scale * theory)**2)
 
 
-
-def taubin_circle_fit(
-    x,
-    y,
-    mask
-):
+def taubin_circle_fit(x, y, mask):
     """
     reference : Barcelona_Muons_TPA_final.pdf (slide 6)
 
--- a/ctapipe/image/muon/fitting.py
+++ b/ctapipe/image/muon/fitting.py
@@ -24,7 +24,7 @@ def cherenkov_integral(lambda1, lambda2):
     return 1 / lambda1 - 1 / lambda2
 
 
-def kundu_chaudhuri_circle_fit(x, y, weights, mask=None):
+def kundu_chaudhuri_circle_fit(x, y, weights):
     """
     Fast and reliable analytical circle fitting method previously used
     in the H.E.S.S. experiment for muon identification
@@ -41,8 +41,6 @@ def kundu_chaudhuri_circle_fit(x, y, weights, mask=None):
         weights of the points
 
     """
-    if mask is not None:
-        weights = weights * mask
 
     weights_sum = np.sum(weights)
     mean_x = np.sum(x * weights) / weights_sum
@@ -457,13 +455,9 @@ def _impact_parameter_chisq(params, phi, hist, mirror_radius):
     return np.sum((hist - scale * theory)**2)
 
 
-def taubin_circle_fit(x, y, weights, mask):
+def taubin_circle_fit(x, y, mask):
     """
     reference : Barcelona_Muons_TPA_final.pdf (slide 6)
-
-    Note:
-        weights parameter is ignored, but _circle_fit functions should
-        all provide the same interface
 
     Parameters
     ----------
--- a/ctapipe/image/muon/fitting.py
+++ b/ctapipe/image/muon/fitting.py
@@ -8,14 +8,14 @@ from astropy.units import Quantity
 from ctapipe.utils.quantities import all_to_value
 
 __all__ = [
-    'kundu_chaudhuri_circle_fit',
-    'taubin_circle_fit',
-    'psf_likelihood_fit',
-    'impact_parameter_chisq_fit',
-    'mirror_integration_distance',
-    'expected_pixel_light_content',
-    'radial_light_intensity',
-    'efficiency_fit',
+    "kundu_chaudhuri_circle_fit",
+    "taubin_circle_fit",
+    "psf_likelihood_fit",
+    "impact_parameter_chisq_fit",
+    "mirror_integration_distance",
+    "expected_pixel_light_content",
+    "radial_light_intensity",
+    "efficiency_fit",
 ]
 
 
@@ -52,15 +52,15 @@ def kundu_chaudhuri_circle_fit(x, y, weights):
     b1 = np.sum(weights * (x - mean_x) * y)
     b2 = np.sum(weights * (y - mean_y) * y)
 
-    c1 = 0.5 * np.sum(weights * (x - mean_x) * (x**2 + y**2))
-    c2 = 0.5 * np.sum(weights * (y - mean_y) * (x**2 + y**2))
+    c1 = 0.5 * np.sum(weights * (x - mean_x) * (x ** 2 + y ** 2))
+    c2 = 0.5 * np.sum(weights * (y - mean_y) * (x ** 2 + y ** 2))
 
     center_x = (b2 * c1 - b1 * c2) / (a1 * b2 - a2 * b1)
     center_y = (a2 * c1 - a1 * c2) / (a2 * b1 - a1 * b2)
 
-    radius = np.sqrt(np.sum(
-        weights * ((center_x - x)**2 + (center_y - y)**2)
-    ) / weights_sum)
+    radius = np.sqrt(
+        np.sum(weights * ((center_x - x) ** 2 + (center_y - y) ** 2)) / weights_sum
+    )
 
     return radius, center_x, center_y
 
@@ -83,11 +83,12 @@ def _psf_neg_log_likelihood(params, x, y, weights):
     This will usually be x and y coordinates and pe charges of camera pixels
     """
     radius, center_x, center_y, sigma = params
-    pixel_distance = np.sqrt((center_x - x)**2 + (center_y - y)**2)
+    pixel_distance = np.sqrt((center_x - x) ** 2 + (center_y - y) ** 2)
 
     return np.sum(
-        (np.log(sigma) + 0.5 * ((pixel_distance - radius) / sigma)**2) * weights
+        (np.log(sigma) + 0.5 * ((pixel_distance - radius) / sigma) ** 2) * weights
     )
+
 
 def psf_likelihood_fit(x, y, weights):
     """
@@ -119,7 +120,7 @@ def psf_likelihood_fit(x, y, weights):
         standard deviation of the gaussian profile (indictor for the ring width)
     """
     unit = x.unit
-    x, y= all_to_value(x, y, unit=unit)
+    x, y = all_to_value(x, y, unit=unit)
 
     start_r, start_x, start_y = kundu_chaudhuri_circle_fit(x, y, weights)
 
@@ -127,12 +128,12 @@ def psf_likelihood_fit(x, y, weights):
         _psf_neg_log_likelihood,
         x0=(start_r, start_x, start_y, 5e-3),
         args=(x, y, weights),
-        method='L-BFGS-B',
+        method="L-BFGS-B",
         bounds=[
-            (0, None),      # radius should be positive
+            (0, None),  # radius should be positive
             (None, None),
             (None, None),
-            (0, None),      # std should be positive
+            (0, None),  # std should be positive
         ],
     )
 
@@ -143,14 +144,7 @@ def psf_likelihood_fit(x, y, weights):
 
 
 def impact_parameter_chisq_fit(
-        pixel_x,
-        pixel_y,
-        weights,
-        radius,
-        center_x,
-        center_y,
-        mirror_radius,
-        bins=30,
+    pixel_x, pixel_y, weights, radius, center_x, center_y, mirror_radius, bins=30,
 ):
     """
     Impact parameter calculation for a ring fit before.
@@ -185,11 +179,11 @@ def impact_parameter_chisq_fit(
         _impact_parameter_chisq,
         x0=(mirror_radius / 2, bin_centers[np.argmax(hist)], 1),
         args=(bin_centers, hist, mirror_radius),
-        method='L-BFGS-B',
+        method="L-BFGS-B",
         bounds=[
-            (0, None),         # impact parameter should be positive
-            (-np.pi, np.pi),   # orientation angle should be in -pi to pi
-            (0, None),         # scale should be positive
+            (0, None),  # impact parameter should be positive
+            (-np.pi, np.pi),  # orientation angle should be in -pi to pi
+            (0, None),  # scale should be positive
         ],
     )
 
@@ -223,14 +217,11 @@ def mirror_integration_distance(phi, phi_max, impact_parameter, mirror_radius):
     """
     phi = phi - phi_max
     ratio = impact_parameter / mirror_radius
-    radicant = 1 - ratio**2 * np.sin(phi)**2
+    radicant = 1 - ratio ** 2 * np.sin(phi) ** 2
 
     if impact_parameter > mirror_radius:
         distance = np.empty_like(phi)
-        mask = np.logical_and(
-            phi < np.arcsin(1 / ratio),
-            phi > -np.arcsin(1 / ratio)
-        )
+        mask = np.logical_and(phi < np.arcsin(1 / ratio), phi > -np.arcsin(1 / ratio))
         distance[np.logical_not(mask)] = 0
         distance[mask] = 2 * mirror_radius * np.sqrt(radicant[mask])
     else:
@@ -240,14 +231,14 @@ def mirror_integration_distance(phi, phi_max, impact_parameter, mirror_radius):
 
 
 def radial_light_intensity(
-        phi,
-        phi_max,
-        cherenkov_angle,
-        impact_parameter,
-        pixel_fov,
-        mirror_radius,
-        lambda1=300e-9,
-        lambda2=900e-9,
+    phi,
+    phi_max,
+    cherenkov_angle,
+    impact_parameter,
+    pixel_fov,
+    mirror_radius,
+    lambda1=300e-9,
+    lambda2=900e-9,
 ):
     """
     Amount of photons per azimuthal angle phi on the muon ring as given in
@@ -277,29 +268,31 @@ def radial_light_intensity(
     """
 
     return (
-        0.5 * const.fine_structure *
-        cherenkov_integral(lambda1, lambda2) *
-        pixel_fov / cherenkov_angle *
-        np.sin(2 * cherenkov_angle) *
-        mirror_integration_distance(phi, phi_max, impact_parameter, mirror_radius)
+        0.5
+        * const.fine_structure
+        * cherenkov_integral(lambda1, lambda2)
+        * pixel_fov
+        / cherenkov_angle
+        * np.sin(2 * cherenkov_angle)
+        * mirror_integration_distance(phi, phi_max, impact_parameter, mirror_radius)
     )
 
 
 def expected_pixel_light_content(
-        pixel_x,
-        pixel_y,
-        center_x,
-        center_y,
-        phi_max,
-        cherenkov_angle,
-        impact_parameter,
-        sigma_psf,
-        pixel_fov,
-        pixel_diameter,
-        mirror_radius,
-        focal_length,
-        lambda1=300e-9,
-        lambda2=900e-9,
+    pixel_x,
+    pixel_y,
+    center_x,
+    center_y,
+    phi_max,
+    cherenkov_angle,
+    impact_parameter,
+    sigma_psf,
+    pixel_fov,
+    pixel_diameter,
+    mirror_radius,
+    focal_length,
+    lambda1=300e-9,
+    lambda2=900e-9,
 ):
     """
     Calculate the expected light content of each pixel for a muon ring with the
@@ -342,13 +335,18 @@ def expected_pixel_light_content(
         number of photons for each pixel given in pixel_x, pixel_y
     """
     phi = np.arctan2(pixel_y - center_y, pixel_x - center_x)
-    pixel_r = np.sqrt((pixel_x - center_x)**2 + (pixel_y - center_y)**2)
+    pixel_r = np.sqrt((pixel_x - center_x) ** 2 + (pixel_y - center_y) ** 2)
     ring_radius = cherenkov_angle * focal_length
 
     light = radial_light_intensity(
-        phi, phi_max,
-        cherenkov_angle, impact_parameter,
-        pixel_fov, mirror_radius, lambda1, lambda2
+        phi,
+        phi_max,
+        cherenkov_angle,
+        impact_parameter,
+        pixel_fov,
+        mirror_radius,
+        lambda1,
+        lambda2,
     )
 
     result = light * pixel_diameter * norm.pdf(pixel_r, ring_radius, sigma_psf)
@@ -356,15 +354,15 @@ def expected_pixel_light_content(
 
 
 def efficiency_fit(
-        pe_charge,
-        pixel_x,
-        pixel_y,
-        pixel_fov,
-        pixel_diameter,
-        mirror_radius,
-        focal_length,
-        lambda1=300e-9,
-        lambda2=900e-9,
+    pe_charge,
+    pixel_x,
+    pixel_y,
+    pixel_fov,
+    pixel_diameter,
+    mirror_radius,
+    focal_length,
+    lambda1=300e-9,
+    lambda2=900e-9,
 ):
     """
     Estimate optical efficiency for a muon ring using method of [mitchell15]_.
@@ -452,7 +450,7 @@ def _impact_parameter_chisq(params, phi, hist, mirror_radius):
     imp_par, phi_max, scale = params
     theory = mirror_integration_distance(phi, phi_max, imp_par, mirror_radius)
 
-    return np.sum((hist - scale * theory)**2)
+    return np.sum((hist - scale * theory) ** 2)
 
 
 def taubin_circle_fit(x, y, mask):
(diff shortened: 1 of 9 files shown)


dneise:
Hello, thanks for the first review @MaxNoe  and @kosack.
You were totally right, there was no state, so we turned `fit()` into a classmethod. I was wondering if we should try to get the interface `TaubinFitter.fit()` as close to `ChaudhuriKunduRingFitter.fit()` as possible. 

At the moment the problem is, that `TaubinFitter.fit()` would need some starting values, which basically only depend on the Camera type. So I assume they could be fed in via a config file... not sure though.

what do you think.

MaxNoe:
I'm confused. The kundu chauduri is a simple function: https://github.com/cta-observatory/ctapipe/blob/f34624a8bdb5ef0a6fd85ec7c3d3ff490d666543/ctapipe/image/muon/fitting.py#L23

MaxNoe:
I would also say that starting values are normally guessable. E.g. a weighted average for center x, y (or even just 0, 0) and the most common value (1.2 deg * focal length / fov) for the radius

dneise:
Ah ... I think we were confused as well. You are looking at `image/muon/fitting.py` we were looking at `image/muon/muon_ring_finder.py` ... 

Thanks for the hint, we'll have a closer look and come back later (after the weekend :-D)

kosack:
There is a lot of refactoring to do in the Muon module, so i wouldn't worry too much about its current design - let's try to simplify it as much as possible. 

 I would suggest doing what we did in other modules:  define everything as either a simple function, or as a class with `__call__(..)` set such that it acts like a function (if state and configuration needs to be stored between calls). 

That way both the ChaudhuriKundu and Taubin fits are identical.   The final design should support user configuration like the EventSource and CameraCalibrator classes, but if the algorithms are really simple, they don't need to be subclasses, we can just have a "MuonReconstructor" or whatever that calls the right function depending on the user's choice (e.g. an attribute called `method`)

AMWMitchell:
Hi - does anyone know why "fitting.py" exists ? There is a version of the Chaudhuri-Kundu fit in there as a function, but it is never used. 
We tried to mirror the implementation of the ChaudhuriKunduRingFitter in muon_ring_finder.py 
--> This is the version that is *actually* used by muon_reco_function. 
What changes are needed to the Taubin fit? 
(Leaving future refactoring of the muon code aside for now.)

We should move the Taubin fit test from test_muon_fitting into test_muon_ring_finder.

MaxNoe:
I implemented the kundu chauduri function in fitting.py before the other classes were there. 

I don't know why code review did not spot the duplication when the other things were merged.

AMWMitchell:
Just checked the history and it seems both versions were created in October 2016. 
The issue remains that the version in "fitting.py" (a bad name) was never used in the muon calibration chain. 
Not sure if we've done a sanity check that the two implementations give the same results & open question on which to retain + how to structure the ring fitting (function / class).

For this pull request - what needs to change to the Taubin Fit ?

MaxNoe:
Like we said, just implement it as a simple function for now!

kosack:
> Just checked the history and it seems both versions were created in October 2016.
The issue remains that the version in "fitting.py" (a bad name) was never used in the muon calibration chain.

Can we just remove the non-used one? That sounds like a confusion that will cause problems soon!

kosack:
(and sorry @momorning that this is becoming a discussion on re-factoring the muon module, but I think this PR can be a catalyst for that! - we can divide the work though)

dneise:
Maybe we could quickly move the muon refactoring discussion into an issue...

AMWMitchell:
Weirdly, I'm not sure that any of the functions in fitting.py are used. Many seem to be duplicated in e.g. muon_integrator.py (which is called by the muon chain) but structured differently.

Doesn't mean they are wrong, just we should work out what we want to keep. 
A bigger discussion than this PR, but depending which version we prefer determines how to implement the TaubinFit - as a class like in muon_ring_finder or a function as in fitting ?

kosack:
I think the question is just if it needs configuration or state between calls, if yes, it should be a class, if not, just a function (which we can wrap in a class later if we need to).  For example, if those fits don't need any special state stored, you could define both as a function, but then just modify your MuonRingFinder class to call the right function depending on a user-configuration variable  called "fit_method" or something (which is just a traits.CaselessStrEnum matching "taubin" or "chaudhury-kundu").

kosack:
You should think about how a user would use it at the high level. I'd guess something like:
```py3
muonfit = MuonFitter(fit_method="taubin")
result = muonfit(x,y,weights,times)
```

and you want the user to be able to specify this in a config file, like:
```
'MuonFitter': { 
    'fit_method': 'taubin'
} 
```
which could be achieved with something like:

```py3

def taubin(x, y, w, t):
   ... 

def chaudhury_kundu(x, y, w, t)
   ...


class MuonFitter(Component):
     fit_method = traits.CaselessStrEnum(
        ['taubin', 'chaudhury-kundu'], 
        default_value='taubin'
     ).tag(config=True)

    def __init__(self,  config=None, parent=None, **kwargs):
        super().__init__(config, parent, **kwargs)
        fit_map = {'taubin': taubin, 'chaudhury-kundu': chaudhury_kundu}
        self.__call__ = fit_map[self.fit_method]
```

momorning:
Code restructured, specify fit_method and tel_type when calling RingFitter
Example usage: 
```muonfit = MuonRingFitter(fit_method="taubin", tel_type = 'MST_MST_FlashCam')
muon_ring_parameters = muonfit.fit(x[mask],y[mask],img)```

File: ctapipe/image/muon/muon_ring_finder.py [hunk 1]
```diff
@@ -1,28 +1,139 @@
...
-from ctapipe.image.muon.ring_fitter import RingFitter
+from ctapipe.core import Component
 from ctapipe.io.containers import MuonRingParameter
+import traitlets as traits
+from iminuit import Minuit
+
+__all__ = ['MuonRingFitter']
+
+
+class MuonRingFitter(Component):
```
MaxNoe: I think this should not contain the functions for taubin and chaudhuri fit. Define them as simple  functions outside the class that only get the inputs they need. Then call those functions here.

That makes it easier to test the single components and enhances readability.

ctapipe/image/muon/muon_ring_finder.py
```diff
@@ -1,28 +1,139 @@
...
+    """Different ring fit algorithms for muon rings
+    """
+    fit_method = traits.CaselessStrEnum(
+        ['taubin', 'chaudhuri_kundu'],
+        default_value='chaudhuri_kundu'
+    ).tag(config=True)
+
+    tel_type = traits.CaselessStrEnum(
```
MaxNoe: I don't think we should hard code a list of tel-types here. If you require information about the telescopes, the `__call__` method should get the `subarray` and from there you should extract the information you need.
If it's only a few things you need, like the focal length, you should just give that as input.

ctapipe/image/muon/muon_ring_finder.py
```diff
@@ -1,28 +1,139 @@
...
+        ['LST_LST_LSTCam', 'MST_MST_NectarCam', 'MST_MST_FlashCam', 'MST_SCT_SCTCam',
+         'SST_1M_DigiCam', 'SST_GCT_CHEC', 'SST_ASTRI_ASTRICam', 'SST_ASTRI_CHEC'],
+        default_value='LST_LST_LSTCam'
+    ).tag(config=True)
+
+    def __init__(self, config=None, parent=None, **kwargs):
+        super().__init__(config, parent, **kwargs)
+
+    def fitFormula(self, xc, yc, r):
```
MaxNoe: Place this outside and called it something like `taubin_loss_function`.

ctapipe/image/muon/muon_ring_finder.py
```diff
@@ -1,28 +1,139 @@
...
+        '''taubin fit formula
+        reference : Barcelona_Muons_TPA_final.pdf (slide 6)
+        '''
+        upper_term = sum(
```
MaxNoe: using python's sum is slow on numpy array, fastest is to use the `.sum()` method.

So do: `upper_term = ((x - xc)**2 + (y - yc)**2 - r**2).sum()`

ctapipe/image/muon/muon_ring_finder.py
```diff
@@ -1,28 +1,139 @@
...
+                    (np.array(self.y) - yc) ** 2
+            )
+        )
+
+        return np.abs(upper_term) / np.abs(lower_term)
+
+    def initialization(self):
+        """ initialization for taubin_fit with the
+        initial default values, errors, and limits
+        """
```
MaxNoe: Again, this should not hard code any starting values for a telescope. Where do those come from? 
You can calculate the most common r from the focal length and field of view for example, which you can get from the instrument description.

ctapipe/image/muon/muon_ring_finder.py
```diff
@@ -69,3 +180,16 @@ def fit(self, x: u.deg, y: u.deg, weight, times=None):
         output.ring_fit_method = "ChaudhuriKundu"
 
         return output
+
+    def fit(self, x, y, img):
```
MaxNoe: this should be `__call__`. It  should not not save `x` and `y` into the component. That state tracking is not necessary.

kosack:
It may be good to split thjis into 2 PRs: 
1. define the taubin_fit function (as was started)
2. refactor the MuonReconstruction classes (the later discussion)

That will be more simple than doing it all at once, perhaps (but it's up to @momorning and @AMWMitchell  )

AMWMitchell:
OK - we'll refactor the muon reconstruction code in a separate PR. 
For this one, changes to MuonRingFitter should be enough - @MaxNoe can you point us to an example usage of the TelescopeParameter traitlet in order to access the instrument description?
We should be able to calculate reasonable limits and starting values from the focal length and the field of view.

ctapipe/image/muon/muon_ring_finder.py
```diff
@@ -16,73 +15,65 @@ class MuonRingFitter(Component):
...
-        taubin_error = [0.003, 0.005, 0.005, 0.005, 0.006, 0.004, 0.007, 0.004]
-        taubin_limit = [(-0.0365, 0.0365), (-0.0640, 0.0640), (-0.0625, 0.0625), (-0.0646, 0.0646),
-                        (-0.0719, 0.0719), (-0.0462, 0.0462), (-0.0848, 0.0848)]
-        taubin_params = {'Name': names, 'rInitial': taubin_r_initial,
-                         'error': taubin_error, 'limit': taubin_limit}
-        taubin_dictIndex = taubin_params['Name'].index(self.tel_type)
-        radius = taubin_params['rInitial'][taubin_dictIndex]
-        error = taubin_params['error'][taubin_dictIndex]
-        limit = taubin_params['limit'][taubin_dictIndex]
+        focal_length = self.teldes.optics.equivalent_focal_length / u.m
```
MaxNoe: use `to_value(u.m)` to get a pure number expressed in meter from an astropy quantity

MaxNoe: This avoids problems, if focal length is not in m, but e.g. in mm:
```
In [2]: import astropy.units as u
In [3]: f = 5000 * u.mm

In [4]: f =  f / u.m

In [5]: 0.5 / f
Out[5]: <Quantity 0.0001 m / mm>
```

dneise:
Ok, so the next problem you can see, when you click on the details next to "Travis CI" and then choose e.g. Python 3.7:

Or just click this link:
https://travis-ci.org/cta-observatory/ctapipe/jobs/611793323?utm_medium=notification&utm_source=github_status

Then scroll down to the first failing test, you see that all tests in `ctapipe/image/muon/tests/test_muon_ring_finder.py` are failing at the moment ... 

When you scroll down further you can find out why.

dneise:
The test failures are reported at line 1470 of that website, this direct link should bring you there.

https://travis-ci.org/cta-observatory/ctapipe/jobs/611793323#L1470

MaxNoe:
You can also run the tests locally using `python setup.py test`

ctapipe/image/muon/tests/test_muon_ring_finder.py
```diff
@@ -1,113 +1,42 @@
...
-    )
-    assert np.isclose(
-        result.ring_center_x.to_value(u.rad), center_x / focal_length,
-        rtol=0.05
-    )
-    assert np.isclose(
-        result.ring_center_y.to_value(u.rad), center_y / focal_length,
-        rtol=0.05,
-    )
+    assert np.isclose(xc_fit * flashcam_focal_length / u.m / u.rad, center_xs / u.m, 1e-1)
```
MaxNoe: you can use astropys unit aware version of isclose: `u.isclose`, also, like before: use `.to_value` to convert quantities to plain numbers.

MaxNoe: Currently all results (e.g. the hillas parameters) are in camera coordinates, this looks like the xc_fit is in CameraCoordiantes but expressed in degrees using a custom conversion. This should not be the case.

ctapipe/image/muon/muon_ring_finder.py
kosack: If you need the FOV, i think the most clear way would be to use either the outer bounding circle of the camera, or perhaps an inner one avoiding the border (for odd-shaped cameras that may be better).  Something like:
```py3
cam = subarray.tel[x].camera
r_outer = np.hypot(cam.pix_x, cam.pix_y).max()
fov = np.pi * r_outer**2
```
or
```py3
m = cam.get_border_pixel_mask(2)   # move in 2 border layers
r_inner = np.hypot(cam[m].pix_x, cam[m].pix_y).max()
fov = np.pi * r_inner**2
```

THen you can convert that to degrees by using the focal length (`subarray.tel[x].optics.equivalent_focal_length`)

ctapipe/image/muon/tests/test_muon_ring_finder.py
```diff
@@ -33,9 +33,9 @@ def test_chaudhuri_kundu_fitter():
     yc_fit = muon_ring_parameters.ring_center_y
     r_fit = muon_ring_parameters.ring_radius
 
-    assert np.isclose(xc_fit * flashcam_focal_length / u.m / u.rad, center_xs / u.m, 1e-1)
-    assert np.isclose(yc_fit * flashcam_focal_length / u.m / u.rad, center_ys / u.m, 1e-1)
-    assert np.isclose(r_fit * flashcam_focal_length / u.m / u.rad, ring_radius / u.m, 1e-1)
+    assert u.isclose((xc_fit * focal_length).to_value(u.m*u.rad), center_xs.to_value(u.m), 1e-1)
+    assert u.isclose((yc_fit * focal_length).to_value(u.m*u.rad), center_ys.to_value(u.m), 1e-1)
+    assert u.isclose((r_fit * focal_length).to_value(u.m*u.rad), ring_radius.to_value(u.m), 1e-1)
```
MaxNoe: if you use `u.isclose` you do not need to strip the units! Also, comparing something as equal that has different units looks strange, what is going on here?

momorning: It is true that it could be as simple as that when I use meter for all of the inputs and outputs. Does that look okay?

MaxNoe: In the end, it should probably look just like this: `u.isclose(r_fit, ring_radius, 1e-1)`

momorning: yes exactly

dneise:
I think we still need a test for `fit_method="taubin"`

ctapipe/image/muon/tests/test_muon_ring_finder.py
dneise: I think this can be resolved.

momorning:
> I think we still need a test for `fit_method="taubin"`

I did not put a test for Taubin because it requires telescope description  (`event.inst.subarray.tel[telid]`) and I was not sure how I should put that with random fake muon generation. Do you have any suggestions?

dneise:
Ah yes sure. If you look into `test_muon_ring_finder.py` line 18. You do there:
`geom = CameraGeometry.from_name("FlashCam")`

you can also create a telescope description in a similarly easy fashion:

```
from ctapipe.instrument.telescope import TelescopeDescription
teldes = TelescopeDescription.from_name('MST', 'FlashCam')
```

ctapipe/image/muon/muon_ring_finder.py
```diff
@@ -69,3 +170,16 @@ def fit(self, x: u.deg, y: u.deg, weight, times=None):
         output.ring_fit_method = "ChaudhuriKundu"
 
         return output
+
+    def fit(self, x, y, img):
+        """allows any fit to be called in form of
+            MuonRingFitter(fit_method = "name of the fit")
+        """
+        self.x = x
```
MaxNoe: There is no reason, x and y should be stored to the instance here. It's not necessary to add this statefulness.

Just hand it over to the fit methods, which then can be free functions.

dneise: fixed that

ctapipe/image/muon/muon_ring_finder.py
dneise: thx, done.

ctapipe/image/muon/muon_ring_finder.py
dneise: exactly .. did it sir

ctapipe/image/muon/muon_ring_finder.py
dneise: I think regarding the initial values, we should enable the MuonRingFinder to configure the initial values of the respective method it uses via traitlets. At least that would be nice. 

But honestly, even nicer would be if the intial values would not matter too much, then we can fix them and just not bother. I'd like to study the behaviour of the fit from the initial values a little ... and see how much they matter. What do you think?

dneise: We did not really solve the issue with the hardcoded stuff in `muon_reco_functions.py` yet, but I think in this PR we also did not make it worse. so I go ahead and resolve this comment.

dneise:
Okay, please have a look again ... we think this is okay-ish

Soon we will add a hough circle fit and then look at the muon Tool again with respect to the configuration and hardcoded limits and so on.

ctapipe/image/muon/fitting.py
```diff
@@ -83,6 +87,16 @@ def _psf_neg_log_likelihood(params, x, y, weights):
         (np.log(sigma) + 0.5 * ((pixel_distance - radius) / sigma)**2) * weights
     )
 
+def strip_unit_savely(x, y):
```
MaxNoe: This converts to si base units and returns the values in those units. I don't think we want this, right?

How about:

```
x = u.Quantitity(x, copy=False)
x, unit = v.value, x.unit
y = u.Quantity(y, copy=False).to_value(unit)

return x, y, unit
```
This will 
* not copy the data
* keep the unit of x
* make sure y is convertible to x
return the values of x y and the unit
```

Comment (dneise) [reply ctapipe/image/muon/fitting.py]: nice, thanks

Comment (MaxNoe) [reply ctapipe/image/muon/fitting.py]: Also *safely*

Comment (dneise) [reply ctapipe/image/muon/fitting.py]: How do you like that?
I added arbitrary number of args, since for the hough fit I've already seen we need this function with 3 arguments

Comment (dneise) [reply ctapipe/image/muon/fitting.py]: is the name okay? it does not strip ... is rather splits ... hmm

Comment (MaxNoe) [reply ctapipe/image/muon/fitting.py]: Also thought about using `*args` 

Something like: `all_to_value(*args, unit=unit)`?

And then you can do stuff like `all_to_value(x, y, r, unit=x.unit)` or `all_to_value(x, y, r, unit=u.m)`?

Comment (dneise) [reply ctapipe/image/muon/fitting.py]: hmm thats more explicit. ... so yes.

Comment (dneise) [reply ctapipe/image/muon/fitting.py]: should it still return the unit then? I guess no.

Comment (MaxNoe) [reply ctapipe/image/muon/fitting.py]: Don't think it should.

Comment (dneise) [reply ctapipe/image/muon/fitting.py]: done .. and I assume resolved.

File: ctapipe/image/muon/fitting.py [hunk 12]
```diff
@@ -83,6 +87,23 @@ def _psf_neg_log_likelihood(params, x, y, weights):
...
+def all_to_value(*args, unit):
+    '''strips unit if all args are convertible to the same unit.
+
+    - does not copy the data
+    - the unit returned, will be the one of the 1st arg
+    - makes sure all args are convertible to the same unit
+    - return the values of all args and the unit
+    - Raise a meaningful error in case the args are not of a convertible unit.
+
+    Returns: (*args_without_unit, unit)
```
MaxNoe: No, just *args without unit.

ctapipe/image/muon/fitting.py
```diff
@@ -83,6 +87,23 @@ def _psf_neg_log_likelihood(params, x, y, weights):
         (np.log(sigma) + 0.5 * ((pixel_distance - radius) / sigma)**2) * weights
     )
 
+def all_to_value(*args, unit):
```
MaxNoe: Nice, that's a really general function. Move it somewhere more appropriate?

dneise: yes ... do you have an idea where?

ctapipe/image/muon/fitting.py
dneise: damn docs are not my friend ...

ctapipe/image/muon/fitting.py
dneise: I assume moving functions to better places can easily done in the future. This PR should maybe not depend on it.

dneise: moved it to utils

dneise:
Hello, hello :-D @MaxNoe and @kosack 
If you're still unhappy with this, don't hesitate to shout it out ... I'll fix it .. no opposition from my side to be expected here ... just tell me, if you have the time.

ctapipe/image/muon/muon_ring_finder.py
```diff
@@ -1,71 +1,50 @@
...
-        output.ring_radius = radius  # *u.deg
-        output.ring_phi = np.arctan(centre_y / centre_x)
-        output.ring_inclination = np.sqrt(centre_x ** 2. + centre_y ** 2.)
-        # output.meta.ring_fit_method = "ChaudhuriKundu"
-        output.ring_fit_method = "ChaudhuriKundu"
+        output.ring_center_x = center_x
+        output.ring_center_y = center_y
+        output.ring_radius = radius
+        output.ring_phi = np.arctan2(center_y, center_x)
+        output.ring_inclination = np.sqrt(center_x ** 2. + center_y ** 2.)
```
MaxNoe: This is not really an inclinitation, right? 

It still is, because how lengths relate to angles but I find still confusing.

Maybe better ring_center_r, ring_center_phi?

ctapipe/image/muon/muon_ring_finder.py
```diff
@@ -1,71 +1,50 @@
...
+    weights = weights * mask
+    return kundu_chaudhuri_circle_fit(x, y, weights)
+
+def taubin(x, y, weights, mask):
+    '''taubin_circle_fit with x, y, weights, mask interface'''
+    return taubin_circle_fit(x, y, mask)
+
+FIT_METHODS = [kundu_chaudhuri, taubin]
+FIT_METHOD_BY_NAME = {m.__name__: m for m in FIT_METHODS}
+FIT_METHOD_NAMES = list(FIT_METHOD_BY_NAME.keys())
```
MaxNoe: Do we really need those 3? The dictionary should suffice, right?

ctapipe/image/muon/fitting.py
```diff
@@ -453,3 +453,84 @@ def _impact_parameter_chisq(params, phi, hist, mirror_radius):
...
+    Parameters
+    ----------
+    x: array-like or astropy quantity
+        x coordinates of the points
+    y: array-like or astropy quantity
+        y coordinates of the points
+    mask: array-like boolean
+        true for pixels surviving the cleaning
+    """
+    orinal_unit = x.unit
```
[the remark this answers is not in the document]
dneise: oh ! :-D ... thx

ctapipe/image/muon/fitting.py
```diff
@@ -453,3 +453,84 @@ def _impact_parameter_chisq(params, phi, hist, mirror_radius):
...
+        error_yc=taubin_error,
+        error_r=taubin_error,
+        limit_xc=(-2*R, 2*R),
+        limit_yc=(-2*R, 2*R),
+        limit_r=(0, R),
+        pedantic=False
+    )
+    fit.migrad()
+
+    radius = fit.values['r'] * orinal_unit
```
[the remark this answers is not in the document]
dneise: okay thanks ... I'll try to fix it in all places then...

ctapipe/image/muon/muon_ring_finder.py
dneise: Should I really modify this container? It was not introduced in this PR .. and honestly ... modifying a container always makes me uneasy .. who knows in how many places it is being used already.

I'd like another opinion on this.

ctapipe/image/muon/muon_ring_finder.py
dneise: sure .. I can put the list into the dict comprehension .. and call `list(dict.keys())` twice in lines 31,32 ... I do not think it is nicer ... do you?

dneise: well .. you had not made the comment if you didn't.. fixing it.

ctapipe/image/muon/muon_ring_finder.py
MaxNoe: But you modified the contents of the container to not have unit degree anymore (lying docs now maybe even?).

ctapipe/image/muon/muon_ring_finder.py
dneise: done

ctapipe/image/muon/fitting.py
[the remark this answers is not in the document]
dneise: done

ctapipe/image/muon/fitting.py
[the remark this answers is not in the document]
dneise: done

ctapipe/image/muon/muon_ring_finder.py
MaxNoe: Just seemed like triple bookeeping to me, did not think about the changes further down

ctapipe/image/muon/muon_ring_finder.py
dneise: no you are right, it is not an inclination

MaxNoe: It is, the center point of the ring is the inclination of the muon to the optical axis of the telescope.

dneise: ah btw. these two: `ring_phi` and `ring_inclination` only depend on center x and y ... so why do we store them? I though defining these as properties of the muon container would be nice, what do you think?

MaxNoe: Probably the other way around is more useful. Ring r and phi are more interesting that x and y. Maybe ist just makes sense to keep both for now.

ctapipe/image/muon/muon_ring_finder.py
dneise: done

ctapipe/image/muon/muon_ring_finder.py
dneise: I moved these into properties and renamed them in `containers.py` ... which is strictly speaking not part of this PR .. what why not.

ctapipe/io/containers.py
```diff
@@ -527,8 +527,19 @@ class MuonRingParameter(Container):
     ring_center_x = Field(0.0, "centre (x) of the fitted muon ring")
     ring_center_y = Field(0.0, "centre (y) of the fitted muon ring")
     ring_radius = Field(0.0, "radius of the fitted muon ring")
-    ring_phi = Field(0.0, "Orientation of fitted ring")
-    ring_inclination = Field(0.0, "Inclination of fitted ring")
+
+    @property
+    @functools.lru_cache()
```
MaxNoe: Does this return the cached value if I change the attributes?

MaxNoe: What does:

```
c = MuonRingParameter(...)
print(c.muon_ring_phi)
c.ring_center_x = 5 * u.m
print(c.muon_ring_phi)
```
return?

dneise: Ah, I see. ...

dneise: so maybe caching here is not so good. invalidating the cache every time an attribute is set, seems overly complex.

dneise: also strictly speaking caching is optimizing for run time, and I did not show, that run time is an issue here, which is premature optimization.

dneise: Of course ... in the previous version .. when somebody changed `ring_center_x` it was not guaranteed that they would also modify `ring_center_r` and `ring_center_phi` ... so having the latter two as properties seems still right, in my opinion.

Do you agree?

MaxNoe: Yes, but this would also effect other containers, so I would propose to handle that in another PR

dneise: okay .. so I just renamed the container fields as you wished. (force pushed)

ctapipe/utils/quantities.py
```diff
@@ -0,0 +1,18 @@
+'''utils to deal with astropy.units.Quantity'''
+from astropy.units import Quantity
+
+def all_to_value(*args, unit):
+    '''converts all args to value if convertible to the same unit.
+
+    - does not copy the data
+    - the unit returned, will be the one of the 1st arg
+    - makes sure all args are convertible to the same unit
+    - return the values of all args and the unit
```
MaxNoe: this does not return the unit

dneise:
I think we should not merge until some other core dev looked at this, maybe also @AMWMitchell  wants to give a statement?

dneise:
Hm tests are failing for the conda build .. .but I don't get why

dneise:
Ah wait, it is also failing for the master .. so this is unrelated.

AMWMitchell:
@dneise The restructuring looks good to me, but it is probably best reviewed (and accepted for merging) by @MaxNoe and @kosack

dneise:
Okay I re-ran the tests, they passed now. \o/ So maybe this can now be merged and then #1199 could be reviewed in more details, since the diff should be nice and small then? @kosack ?

dneise:
I modified the tests to be `pytest.parametrized` ... just saw this existed .. and think it applies here.
@MaxNoe and @kosack please re-review if you have the time.

dneise:
@kosack can we merge this one? The hough stuff is critical, I see ... but this one here should be fine, no?

Issue #703, opened 2018-04-10
Different telescopes and camera calibration


watsonjj:
Current state of camera calibration:
At each calibration step there is a loop over telescopes. The same calibration is applied to each telescope.

As raised in #702, it will be almost always required that each telescope type uses its own extractor. This is currently not supported in the current way we do things, and it also raised a few questions in my mind that I need help addressing:

1. Are the DL0 files still planned to be stored per telescope?
2. Currently the event containers have `tel` sub-containers. Are these planned to be kept?
3. We need some way of defining the defaults for certain Components (e.g. the default `ChargeExtractor` to use) based on the telescope. Perhaps some sort of "set-up" Component that supplies the user with the correct classes based on the telescope.
4. Should the `ctapipe.calib.camera` Components be restructured to only apply to a single telescope at a time. This would make a lot more sense in the pipeline flow. As the name suggests, these components are camera specific, and their configuration will change depending on the telescope being investigated. A wrapper class (`CameraCalibratorGroup`) could be created which loops through each telescope, and applies the provided calibrators for that telescope.
5. What is the planned approach to combine the data together for each telescope when the telescopes aren't all in a single file like they are for the sim_telarray output.

Thinking about this has made me almost wonder if instead of the `tel` sub-container,  it would be more intuitive to handle a telescope container at the highest level, and beneath it are the "event" containers for each telescope. But I don't know if this fits with the design planned for ctapipe, and if its too late to consider such a change.

But irregardless, it would be nice to discuss the questions I have. It might also be nice to have a call to discuss big topics like these instead of the github issues page.

watsonjj:
The ability to set the extractor traits per telescope was added in #1155

The ability to choose a different charge extraction method per telescope is still unavailable (briefly discussed at Bologna meeting, #1148, won't be implemented for now)

The loop over telescopes in CameraCalibrator is still be discussed in #1165, so for now this issue will remain open.

maxnoe:
Implemented in #1661

Issue #702, opened 2018-04-10
Different integration windows per camera type


kpfrang:
Hi,

is there a way in `CameraCalibrator` to specify different widths and shifts for the trace integration window per camera type?

To my understanding, the DL1 container is filled for all telescopes with the values specified in `WindowIntegrator`. That implies that for all telescopes the same values will be used.

watsonjj:
There are a few ways it can be done. If you wish to configure via the command line or a config file, then you can have a look at examples/calibration_pipeline.py (line 145).

If you wish to set the values directly in the script then one way to do this is to define the individual calibration classes yourself instead of using the `CameraCalibrator` convenience class. You can find an example of this being done in examples/display_integrator.py. For example:

```python
        kwargs = dict(config=self.config, tool=self)

        eventsource = EventSourceFactory.produce(**kwargs)

        self.extractor = ChargeExtractorFactory.produce(
           **kwargs, window_width=12, window_shift=4
        )

        self.r1 = CameraR1CalibratorFactory.produce(
            eventsource=eventsource, **kwargs
        )

        self.dl0 = CameraDL0Reducer(**kwargs)

        self.dl1 = CameraDL1Calibrator(extractor=self.extractor, **kwargs)
```

watsonjj:
However, to properly address you question, applying a different charge_extraction per telescope is not something that has been implemented yet... and needs a discussion on how to best implement it...

Perhaps the extractor argument to `CameraDL1Calibrator` could either be a `ChargeExtractor`, or a list of `ChargeExtractors`, where it is assumed each index in the list corresponds to the next telescope

kpfrang:
Okay, thanks for the clarification.

> Perhaps the extractor argument to CameraDL1Calibrator could either be a ChargeExtractor, or a list of ChargeExtractors, where it is assumed each index in the list corresponds to the next telescope

Something like this was exactly what I wondered whether it already exists. I guess it would be nice to have something like this as it would make it easier to play around with the numbers.

MaxNoe:
> Perhaps the extractor argument to CameraDL1Calibrator could either be a ChargeExtractor, or a list of ChargeExtractors, where it is assumed each index in the list corresponds to the next telescop

Maybe `dict[camera-type] -> ChargeExtractor` would be the most useful

watsonjj:
> Maybe dict[camera-type] -> ChargeExtractor would be the most useful

[...]

Issue #1148, opened 2019-10-24
configuration: is what we have sufficient? Can we simplify it? (with full example: ctapipe-stage1-process)


watsonjj:
- May need something like TelescopeParameterResolver for instances of algorithm classes (so you can use a different image cleaner per telescope, for example)

watsonjj:
The recently accepted configuration approach (#1129) works internally to a component using the Traitlets. A single component can then handle the configuration of multiple telescopes.

An alternative to this is to have a Component instance per telescope. This has a few benefits:
- Keeps the internals of the component simple - no need for a special traitlet
- Removes the telescope loop from components such as CameraCalibrator. Instead the list of CameraCalibrators will be looped over
- Provides an easy solution to #703 & #702 
- This ability to configure the Component to handle multiple telescope types will be unnecessary when each file contains only one telescope

A dict of Components could be created by a configuration/telescope-loop class. The telescope loop is taken outside of Components, and the corresponding Component is obtained from the class for the current telescope. This idea was also suggested in #702.

watsonjj:
A similar topic was raised in #888

watsonjj:
Also in #851

watsonjj:
Decision for the time being:

- Recent configuration changes are good enough for what we need for now - typically you only wish to change traitlet values (such as cleaning thresholds), and not use different algorithms for different telescopes.

PR #1155, opened 2019-10-25
Update ImageExtractors to utilise TelescopeParameters

watsonjj:
Allows per-telescope configuration of ImageExtractors as discussed in #703 & #702

### DIFF
--- a/ctapipe/calib/camera/calibrator.py
+++ b/ctapipe/calib/camera/calibrator.py
@@ -95,6 +95,8 @@ class CameraCalibrator(Component):
         image_extractor : ctapipe.image.extractor.ImageExtractor
             The ImageExtractor to use. If None, then NeighborPeakWindowSum
             will be used by default.
+        subarray: ctapipe.instrument.SubarrayDescription
+            Description of the subarray
         kwargs
         """
         super().__init__(config=config, parent=parent, **kwargs)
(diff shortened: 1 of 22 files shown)


watsonjj:
My initial impression of these TelescopeParameters is that they are a bit cumbersome to use, for two primary reasons:

1. As subarray isn't available until after the event loop has started, the initialisation of TelescopeParameterResolver has to be delayed until the first call.
2. It's a shame that the resolving is a separate class to the TelescopeParameter itself.

These could both be solved if we had access to a subarray database before the event loop started, and if telid was guaranteed to correctly reference the corresponding telescope in the subarray database.

Alternatively, the per-telescope traitlet configuration (`TelescopeParameter`) could be abandoned, and instead adopt the approach suggested in #1148 (Component instance per telescope configuration). However this would likely not be compatible with the Traitlets configuration system.

kosack:
> would it not be useful to have the  read the subarray data from the first event if it if not defined

The end goal is to remove the subarray info from the event altogether, so that the event structure is really only data that changes per-event.  Everything that needs it should have it explicitly passed in, though right now we often construct Components before we load the subarray info, so we have to think of the right order in which to do it.  This is a good testbed for that

thomasarmstrong:
Ok, that makes sense to me.

Issue #722, opened 2018-04-24
Container Cleanup


watsonjj:
I would like to use this issue to discuss some clean-ups to the Container classes that I would propose:

### inst
What is the plans for the inst container? Although it is currently marked as deprecated, it is the main accessor for instrument information in the pipeline.

### event.r0.tel[*].num_samples
Is this Field necessary? The same information can be obtained from the waveform shape.

### tels_with_data
Is this Field necessary? The same information can be obtained from, for example, `event.r0.tels.keys()`

### event_id and obs_id
Many of the containers (`r0`, `r1`, `dl0`, and even `MuonIntensityParameter`) contain these Fields, but it is a duplication of information. As it is an "event" container (it is updated per event), wouldn't this information be better located at the top level (`event.event_id`, or event perhaps `event.id`) instead of replicated in each sub-container 

### Methods inside containers
Containers such as `NectarCAMContainer` and `SST1MContainer` contain the methods to fill them (from the file) inside the container. This approach was fine to get a version of the EventSources into ctapipe, but I personally dislike approach. In my mind, it makes sense (and is easier to follow in the code) if all the reading from the file and filling of containers occurred in the EventSource class, and the containers.py was strictly kept to defining the containers and fields.

If anyone has some other cleanup suggestions to the containers, they are welcome in this thread.

dneise:
I think this is a very good idea. 

Regarding the methods in containers, I think you are right, it was not a good idea to have these methods there.

dneise:
```
class DL0CameraContainer(Container):
    trigger_type = Field(0o0, "camera trigger type")
```

Often the Field default value is `None`, here is it "octal zero" ... I assume this implies trigger types in ctapipe are typically expressed in base 8... Do I understand this correctly?

dneise:
I never understood what the difference between `observation` and `event` was, e.g. in:
```
class R1Container(Container):
    obs_id = Field(-1, "observation ID")
    event_id = Field(-1, "event id number")
```

Now looking through the entire `containers.py` file I found:
```
class MuonRingParameter(Container):
    obs_id = Field(0, "run identification number")
    event_id = Field(0, "event identification number")
```
which seems to explain what the difference is. Is this naming scheme:

 * observation <-> run
 * event <-> event 

still open for discussion? 

[...]

PR #1157, opened 2019-10-26
Enable access to subarray information from file before event loop

watsonjj:
In #1155 I encountered another situation where it would be useful to have the subarray information available immediately after initialising the eventsource (before starting the event loop).

This appears to be a fairly straightforward change, which could help keep a lot of other code tidy.

This idea was originally suggested in #722

### DIFF
--- a/ctapipe/io/eventsource.py
+++ b/ctapipe/io/eventsource.py
@@ -190,6 +190,18 @@ class EventSource(Component):
         """
         return False
 
+    @property
+    @abstractmethod
+    def subarray(self):
+        """
+        Obtain the subarray from the file
+
+        Returns
+        -------
+        ctapipe.instrument.SubarrayDecription
+
+        """
+
     @abstractmethod
     def _generator(self):
         """
--- a/ctapipe/io/eventsource.py
+++ b/ctapipe/io/eventsource.py
@@ -194,7 +194,7 @@ class EventSource(Component):
     @abstractmethod
     def subarray(self):
         """
-        Obtain the subarray from the file
+        Obtain the subarray from the EventSource
 
         Returns
         -------
--- a/ctapipe/io/hessioeventsource.py
+++ b/ctapipe/io/hessioeventsource.py
@@ -51,6 +51,12 @@ class HESSIOEventSource(EventSource):
         '''This class should never be chosen in event_source()'''
         return False
 
+    @property
+    def subarray(self):
+        with self.pyhessio.open_hessio(self.input_url) as file:
+            next(file.move_to_next_event())
+            return self._build_subarray_info(file)
+
     def __exit__(self, exc_type, exc_val, exc_tb):
         HESSIOEventSource._count -= 1
         self.pyhessio.close_file()
--- a/ctapipe/io/simteleventsource.py
+++ b/ctapipe/io/simteleventsource.py
@@ -159,6 +159,10 @@ class SimTelEventSource(EventSource):
     def is_compatible(file_path):
         return is_eventio(Path(file_path).expanduser())
 
+    @property
+    def subarray(self):
+        return self._subarray_info
+
     def _generator(self):
         if self.file_.tell() > self.start_pos:
             self.file_._next_header_pos = 0
--- a/ctapipe/io/tests/test_event_source.py
+++ b/ctapipe/io/tests/test_event_source.py
@@ -20,6 +20,10 @@ class DummyReader(EventSource):
     def is_compatible(file_path):
         return False
 
+    @property
+    def subarray(self):
+        return None
+
 
 def test_can_be_implemented():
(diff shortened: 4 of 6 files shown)


watsonjj:
@kosack In #722 you suggested in addition to attaching the `SubarrayDescription` to the `EventSource`, to also:
- attach `MCHeaderContainer` to the `EventSource`
- mark `event.inst` as deprecated

Should I include that in this PR?

kosack:
yes, this will be solved with the major restructuring of the eventsource/instrument that I proposed a while ago. I should really write up a more detailed description. 

Related are:
- #1041 (discussion on separating mon, svc, inst data from event)
-  #747 (where to put monitoring and service data)

kosack:
I think #1041 is the most relevant: see the proposals there for how to separate it.  My preference would be to have it be an attribute that can be accessed from an EventSource (header info), or have it returned in the event source loop (less ideal).

ctapipe/io/eventsource.py
```diff
@@ -190,6 +190,18 @@ def is_stream(self):
         """
         return False
 
+    @property
+    @abstractmethod
+    def subarray(self):
```
MaxNoe: I don't know if that's a good idea. Are we we sure, subarray info can come from an event source? 
I don't think this is the case for e.g. LST real data event source

watsonjj: In the current scheme the event source must provide it in one way or another (to attach it to the event container), even if the information is not located in the file. So I don't think this change should be a problem.

MaxNoe: ah ok

kosack: yeah, the name EventSource may be confusing, but that's the idea - it provides the events + any other data associated with them (it already does this in fact! Just mixes the two concepts, so I don't see a problem)

watsonjj:
So the inst container is already deprecated.

On second thoughts, I would like to avoid doing the same for the MCHeader in this PR. It isn't as simple to disentangle from the event building as the SubarrayDescription as its being updated per event at the moment...

Can this PR be accepted?

kosack:
if we want to follow the current data model I presented in Bologna, we should really separate four things:
1. metadata headers:
   - **simulation** (what is in the MCHeaderContainer + other simulation header info, like MCThownShowerDistribution, etc)
   - **observation** (right now we don't have this in sim data, but we will need it)
   -  **instrument**: at least the SubarrayDescription, other things?
1. event data 
1. monitoring data (things that change far slower than events, but still change)
1. service data (things that do not change during an observation, but may be needed for analysis)

As discussed a bit in  #1041, at least *some* of this should be loaded automatically by the EventSource (at least event + the header info, perhaps not the monitoring or service by default since those can be loaded later on demand). 

So if you want to implement all that here, let's decide on what it should look like.  i.e.:

```py3
source = EventSource(input_url=x)   
```
already here, the input can be a single file (events), or in the case of most prototypes and probably all real data, a set of files: either we assume there is one "index" file that contains links to others, like in gammapy, or we accept a directory here, or assume there is always a single file and we use the features of e.g. HDF5 to generate a file that contains linked datasets.  How does LSTChain or NectarChain handle this?

Then there is the use of source:

```py3
source.instrument.peek()  # would be nice to be able to do this before the event loop
print(source.simulation.mcheader)
print(source.simulation.thrown_shower_distribution) # in SimTel at end of file only, so not easy to do. In the HDF5 DL1, this could work though
print(source.observation) # implement only when we define what is in a Observation
```
That means in SimTelEventSource having perhaps to do some trickery to make sure it's loaded, but shouldn't be hard. 

Defining classes that need subarray info, either at construction time, or call time?
```
calibrate =  CameraCalibrator(..., subarray=source.instrument)?
```

The event loop itself:
```py3
for event in source:
      event = calibrate(event)
      event = parameterize(event) 
      # or 
      event = parameterize(event, subarray=source.instrument)
```

Which also brings up: is there anything else in instrument than the subarray?

kosack:
Ok, lets leave the big .changes for another PR. I'll move my previous comment to the issue

ctapipe/io/hessioeventsource.py
```diff
@@ -51,6 +51,12 @@ def is_compatible(file_path):
         '''This class should never be chosen in event_source()'''
         return False
 
+    @property
+    def subarray(self):
+        with self.pyhessio.open_hessio(self.input_url) as file:
```
kosack: this of course will fail if the file is a stream. Do we want to break that idea? I think you can change the logic to still work with streams.

watsonjj: Its done this way because of the limitation of the hessio library. Do we have to account for streams with hessio? 

pyeventio allows the SubarrayDescription to be generated before starting the event loop.

PR #1158, opened 2019-10-26
Add resolving to the TelescopeParameter

watsonjj:
This PR attaches the resolving of a `TelescopeParameter` to the parameter itself, removing the need for `TelescopeParameterResolver`.

This is achieved by making TelescopeParameter a traitlet which returns a `TelescopeParameterList` instead of a `list`. 

`TelescopeParameterList` is a subclass of `list` with methods `attach_subarray` and `resolve`, the same functionality that was previously found in `TelescopeParameterResolver`.

This is how a TelescopeParameter was utilised before:
```python
class SomeComponent(Component):
   tel_param1 = IntTelescopeParameter(
      default_value=[("type", "*", 10), ("type", "LST*", 100)]
   )

def __init__(self, subarray):
   self._tel_param1_resolver = TelescopeParameterResolver(subarray, self.tel_param1)

def __call__(self, telid):
   param1 = self._tel_param1_resolver.value_for_tel_id(telid)
```

With this PR, a Component can utilise a `TelescopeParameter` as below:
```python
class SomeComponent(Component):
   tel_param1 = IntTelescopeParameter(
      default_value=[("type", "*", 10), ("type", "LST*", 100)]
   )

def __init__(self, subarray):
   self.tel_param1.attach_subarray(subarray)

def __call__(self, telid):
   param1 = self.tel_param1.resolve(telid)
```

In the future, when the database is defined, it may be possible to have the `TelescopeParameterList` to retrieve the `SubarrayDescription` itself, removing the need to attach it manually.

### DIFF
--- a/ctapipe/core/traits.py
+++ b/ctapipe/core/traits.py
@@ -42,10 +42,10 @@ __all__ = [
     "enum_trait",
     "classes_with_traits",
     "has_traits",
+    "TelescopeParameterList",
     "TelescopeParameter",
     "FloatTelescopeParameter",
     "IntTelescopeParameter",
-    "TelescopeParameterResolver",
 ]
 
 import logging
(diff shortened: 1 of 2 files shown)


kosack:
This is definitely a lot cleaner/nicer than the 2-step implementation.  Perhaps it would be even nicer to remove the `resolve()` function and replace it with just `__get__()`, so that you access the value with `myparam[15]` (where 15 is a tel_id)?

MaxNoe:
:+1: for `__getitem__`

ctapipe/core/traits.py
```diff
@@ -138,6 +138,65 @@ def has_traits(cls, ignore=("config", "parent")):
     return bool(set(cls.class_trait_names()) - set(ignore))
 
 
+class TelescopeParameterList(list):
```
MaxNoe: Do we need to inherit from list here? I think this adds many unwanted interfaces, right? Shouldn't we just store the initial values in a member list instance?

watsonjj: The previous traitlet was a list, so this PR doesn't change anything about unwanted interfaces. But you might be right that we would prefer defining our own class for this purpose.

MaxNoe: But on the otherside, with just inheriting from list it works with parsing config files?

MaxNoe: Probably good as it is, but I'm always a bit hesitant about inheriting from builtins.

kosack: Sorry, I just noticed this after it was merged - yes it does seem strange. Also, it's not a list but a dict, right? You give it a telescope id and it gives you a value? (the tel_ids are not necessarily contiguous) Perhaps we need to make another PR.

Also, what is the `klass` attribute of TelescopeParameter?

MaxNoe: It's a list in the sense that what is in the config file is a list of these 3 tuples `(selector, selection, value)`. So probably by inheriting from `list`, everything just works concerning loading / saving config files.

watsonjj:
```python
class SomeComponent(Component):
   tel_param1 = IntTelescopeParameter(
      default_value=[("type", "*", 10), ("type", "LST*", 100)]
   )

def __init__(self, subarray):
   self.tel_param1.attach_subarray(subarray)

def __call__(self, telid):
   print(self.tel_param1)
   param1 = self.tel_param1[telid]
   print(param1)

```

In this example the first print statement will output `[("type", "*", 10), ("type", "LST*", 100)]`, which is a TelescopeParameterList (subclass of list).
The second print will output `10` (unless its an LST in which case it will output `100`. We have overloaded the `__getitem__` method of a normal list.

kosack: I think my confusion was that in the original implementation, the traitlet just checks that the user has created a list of "patterns" (triplets), but then the Resolver converted that into  a *dict* of value per telescope id.  But now there are 3 steps: `TelescopeParameter` → `TelescopeParameterList` (replacing the list of tuples used in the original representation of the pattern) → dict (by using __getitem__), is that right?  

So it seems ok, just maybe needs better explanation since it's complicated.

watsonjj: Yeah I think the confusing part is that the `__getitem__` of `TelescopeParameterList` operates more like a dict instead of a list, but this was done for conveniance.

kosack: It's recommended to never subclass `list` in python, so maybe making it an attribute as @MaxNoe said is better.

watsonjj: > `TelescopeParameterList` (replacing the list of tuples used in the original representation of the pattern)

`TelescopeParameterList` **IS** the list of tuples.

PR #1160, opened 2019-10-28
Add application of DL1 charge calibration to CameraCalibrator

watsonjj:
This PR adds the ability to apply the DL1 charge calibration that has been attached to the event in the DL1CameraCalibrationContainer (`event.mon.tel[telid].calibration.dl1`)

### DIFF
--- a/ctapipe/calib/camera/calibrator.py
+++ b/ctapipe/calib/camera/calibrator.py
@@ -227,14 +227,10 @@ class CameraCalibrator(Component):
             charge = charge * correction
 
         # Calibrate extracted charge
-        try:
-            pedestal = event.mon.tel[telid].calibration.dl1.pedestal_offset
-            absolute = event.mon.tel[telid].calibration.dl1.absolute_factor
-            relative = event.mon.tel[telid].calibration.dl1.relative_factor
-            charge = (charge - pedestal) * relative / absolute
-        except AttributeError:
-            pass
-            # TODO: Switch to EventAndMonDataContainer for mon
+        pedestal = event.calibration.tel[telid].dl1.pedestal_offset
+        absolute = event.calibration.tel[telid].dl1.absolute_factor
+        relative = event.calibration.tel[telid].dl1.relative_factor
+        charge = (charge - pedestal) * relative / absolute
 
         event.dl1.tel[telid].image = charge
         event.dl1.tel[telid].pulse_time = pulse_time
--- a/ctapipe/calib/camera/calibrator.py
+++ b/ctapipe/calib/camera/calibrator.py
@@ -228,9 +228,9 @@ class CameraCalibrator(Component):
 
         # Calibrate extracted charge
         try:
-            pedestal = event.mon.tel[telid].dl1.pedestal_offset
-            absolute = event.mon.tel[telid].dl1.absolute_factor
-            relative = event.mon.tel[telid].dl1.relative_factor
+            pedestal = event.mon.tel[telid].calibration.dl1.pedestal_offset
+            absolute = event.mon.tel[telid].calibration.dl1.absolute_factor
+            relative = event.mon.tel[telid].calibration.dl1.relative_factor
             charge = (charge - pedestal) * relative / absolute
         except AttributeError:
             pass
--- a/ctapipe/calib/camera/calibrator.py
+++ b/ctapipe/calib/camera/calibrator.py
@@ -211,7 +211,7 @@ class CameraCalibrator(Component):
             #   - Read into dl1 container directly?
             #   - Don't do anything if dl1 container already filled
             #   - Update on SST review decision
-            corrected_charge = waveforms[..., 0]
+            charge = waveforms[..., 0]
             pulse_time = np.zeros(n_pixels)
         else:
             # TODO: pass camera to ImageExtractor.__init__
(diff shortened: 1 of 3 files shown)


ctapipe/calib/camera/calibrator.py
```diff
@@ -218,18 +218,27 @@ def _calibrate_dl1(self, event, telid):
...
             correction = self._get_correction(event, telid)
             corrected_charge = charge * correction
 
-        event.dl1.tel[telid].image = corrected_charge
+        # Calibrate extracted charge
+        try:
+            pedestal = event.mon.tel[telid].dl1.pedestal
+            absolute = event.mon.tel[telid].dl1.absolute
+            relative = event.mon.tel[telid].dl1.relative
+            calibrated_charge = (corrected_charge - pedestal) * relative / absolute
```
MaxNoe: Why have relative and absolute and not a single constant here?

watsonjj: Traditionally in HESS the calibration coefficient is split into two. I think this is due to the separate techniques used to derive each:
- gain: extracted from SPE
- flat-fielding: extracted from uniform bright illumination to correct for the relative Quantum Efficiencies between PMTs

If each parameter is derived from a separate source (separate dataset, separate script), then perhaps its logical to keep this distinction. But I suppose that is a question to the Camera Calibration coordinator...

kosack: I assume *absolute* is the DC/PE measurement, which is computed once every so often (and loaded from a Service table), whereas the relative is the FlatFielding correction, which is computed either on-the-fly, or in a previous pass over the same event data (looking at flasher events).  So the two are computed in different places.   Perhaps naming them less generally would help this?

kosack: Note that you can still for example do reconstruction, without the absolute gain correction, relying only on the relative correction, but then it makes comparing different cameras more complicated (since any cuts on intensity will be different for each camera).   In principle you can also get away with no relative correction (flatfield) if you use a model analysis or something that doesn't rely on the image having a uniform response, but then computing things like Hillas parameters is annoying.

watsonjj: The "general naming" was on purpose. 

The different equivalent names for the absolute calibration are DC2PE and gain. But these can be misnomers when calibrating SiPMs. As the QE (referred in SiPMs as PDE) also changes with the voltage across the photosensor, it is more appropriate to calibrate from DC (or mV) to photons, to ensure the same illumination response is present in each SST.

The other name for relative as you said is flat-fielding. I am not opposed as much against that name. Just "absolute" and "relative" seemed more descriptive about what they are calculated against.

kosack: Ok agreed, general names are better (but still not clear if they are scales or shifts)

ctapipe/io/containers.py
```diff
@@ -126,21 +126,40 @@ class DL1CameraContainer(Container):
...
+    relative = Field(
+        1,
+        "Coefficients for the relative correction between pixels to achieve a "
+        "uniform charge response (post absolute calibration) from a "
+        "uniform illumination."
+    )
+    time = Field(
+        0,
+        "Coefficients for the timing correction before charge extraction "
+        "for each pixel"
```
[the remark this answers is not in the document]
watsonjj: I think its whats known as a time flatfielding. Its the additive offset to offset the waveforms by before you extract the charge. I can add "additive" to the description

watsonjj:
I've added "_factor" to multiplicative coefficients, and "offset" to additive coefficients

ctapipe/io/containers.py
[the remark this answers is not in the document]
kosack: or perhaps `time_shift`

kosack:
what happens currently with monte-carlo data? Are these calibration monitoring containers filled with the values from the MC file?

Also, be clear that "monitoring" is not event-wise, it's time-wise, so if these are intended to be interpolated to an event time (in the case that things like the pedestals), then i'm not sure the monitoring value is the one you want to use.

watsonjj:
> what happens currently with monte-carlo data? Are these calibration monitoring containers filled with the values from the MC file?

No. The calibration coefficients depend on the charge extraction technique, and the calibration approach (e.g. converting to photons)

> Also, be clear that "monitoring" is not event-wise, it's time-wise, so if these are intended to be interpolated to an event time (in the case that things like the pedestals), then i'm not sure the monitoring value is the one you want to use.

The values used here are intended to be the ones that you apply for an event. It may be that these are loaded once per observation, and don't change. Or it may be that they are interpolated from a table depending on the event time (and could be basically the same for multiple events). Or they may be continuously updated from an on-the-fly calibration. Let me know if this is not the correct place for such coefficients...

watsonjj:
These values are intended to be filled by a `Calculator` class, which I will implement in a future PR (plus adapt Franca's Calculator classes)

kosack:
> The values used here are intended to be the ones that you apply for an event. It may be that these are loaded once per observation, and don't change. Or it may be that they are interpolated from a table depending on the event time (and could be basically the same for multiple events). Or they may be continuously updated from an on-the-fly calibration. Let me know if this is not the correct place for such coefficients...

If these are the instantaneous values for the event, not the "monitored values" once per second or more, then they really should be in the event, not monitoring container (I know we currently mix the two, but that will be separated at some point). Instantaneous values should be in the "DL1/Event/Telescope" data model (or R1 depending on when they are computed) and are derived generally from monitoring data.  Perhaps for now, we just put them in `event.calib.tel[x]` to remind ourselves that they are the per-event quantities, so that when we remove all MON data from the event structure, we don't confuse the two?

kosack:
I think I need to write down a complete proposal for the full internal and external data model at some point, but that will require some time (should be done and updated before a 1.0 release though!). For now, let's try to put things somewhere that makes it less painful to move them around later.  Probably just not calling anything "Monitoring" now is best, since we so far do not really compute any true monitoring quantities (except perhaps the outputs from Franca's pedestal tool, if they are not per-event)

watsonjj:
> Really, it's only pedestals I'm worried about confusing. Maybe also flatfield, but anything that doesn't change during an observation (DC/PE, etc) ,should eventually leave the event container completely.

@kosack Does it make sense that it leaves the event container completely? It's important to know the correct value for the current event. As you have said in #1165:

> Therefore what is in the event are instantaneous values of those quantities, interpolated to the current event time. Other COmponents that use them do not need to know how they were interpolated (or even if they were computed on-the-fly like with pedestals, perhaps), just that the value in the Container is the right one for this event. Therefore they don't need to distinguish between data coming from a "service" or from an event stream.

watsonjj:
The dl1 calibration coefficients that are applicable for the current event (and should be used in the CameraCalibrator) are now located at `event.calibration.tel[telid].dl1`

kosack:
> @kosack Does it make sense that it leaves the event container completely? It's important to know the correct value for the current event. As you have said in #1165:

I think we keep the instantaneous values in the event.   (just don't call them "mon")

watsonjj:
> I think we keep the instantaneous values in the event. (just don't call them "mon")

Yep done

PR #1162, opened 2019-10-29
Remove inheritance on list for TelescopeParameterList

watsonjj:
Created to address the conversation which occurred in #1158 after merging.

### DIFF
--- a/ctapipe/core/traits.py
+++ b/ctapipe/core/traits.py
@@ -42,7 +42,7 @@ __all__ = [
     "enum_trait",
     "classes_with_traits",
     "has_traits",
-    "TelescopeParameterList",
+    "TelescopeParameterLookup",
     "TelescopeParameter",
     "FloatTelescopeParameter",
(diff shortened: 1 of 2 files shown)


ctapipe/core/traits.py
```diff
@@ -138,9 +138,20 @@ def has_traits(cls, ignore=("config", "parent")):
...
+        Handles the lookup of corresponding configuration value from a list of
+        tuples for a telid.
+
+        Parameters
+        ----------
+        telecope_parameter_list : list
+            List of tuples in the form `[(command, argument, value), ...]`
+        """
+        if telecope_parameter_list is None:
+            telecope_parameter_list = []
```
vuillaut: Why passing `None` as the default value for `telescope_parameter_list` and not directly `[ ]`?

watsonjj: Because a list is mutable:
https://docs.python-guide.org/writing/gotchas/#mutable-default-arguments

Issue #445, opened 2017-06-27
Typed Items


maxnoe:
I think it would be very usefull for i/o, provenance and sane defaults to introduce typed `Items` for containers.

Containers would then be very similiar to ORM Models, as for example in `peewee`. One could even imaging a direct translation for outputting containers into databases.

Possible non-exhaustive list:

* FloatItem
* IntegerItem
* MapItem
* ListItem
* ArrayItem
* StringItem
* DateTimeItem

This would make it clear, to which column types in FITS or HDF5 or ... one has to convert those entities.

What do you think?

kosack:
I'm happy with that. I originally considered it, but then just went with assuming the type of the default value (which of course may be confusing). This is basically the same design as any typical ORM, like SQLAlchemy.   It would be good to have the following metadata in each Item:
 
* type (by the item class)
* shape (all should support array operations, e.g. an IntItem could have a default shape of 1, a scalar, but in principle could be an array)
* unit 
* UCD (universal content descriptor, from the VO)
* description

I already support this in the HDFTableWriter, but it infers the type from the current value of the Container, not from the Items (which is error-prone), so this would be a better way to do it.

MaxNoe:
See #501

maxnoe:
@kosack implemented type checking for the fields in `validate` in #1163

PR #1163, opened 2019-10-30
DL1 writer tool (ctapipe-stage1-process)

kosack:
Overview:
-----------
This is a PR to start a discussion, containing a fully working DL1 writer tool that writes a proposed standard DL1 HDF5 format. This is currently all contained in `ctapipe/tools/stage1.py`.  

This PR is provided so we can discuss the general workflow, usage,  and data format produced, but please see the caveats below. 

Caveats:
---------
Currently there are a lot of extra features included in the `stage1.py` file, that were needed to get everything to work properly: 
- a configurable `ImageCleaner` class
- a configurable `DataChecker` class (for performing cuts and tracking statistics)
- a function to write core metadata
- some new `Container` classes, and a few algorithms to fill them
- some of the functions inside the `Stage1Process` Tool could be also moved to common places (e.g. functions to write the SubarrayDescription in HDF5 format, could go in a file like `ctapipe/io/hdf5.py`)

These all should eventually move to standard places in ctapipe, but for now are included so that they can **slowly be replaced by separate PRs** that put them where needed. I suggest we avoid discussion/review of the *implementations* of those pieces until their subsequent PRs, but the overall idea of them can be discussed here if needed. 

No unit tests are currently there for these features (will come with the PRs mentioned).

Ideally once we are ready to merge this PR, it will be simplified down to a bare minimum, and the rest will be already parts of ctapipe. 

Usage:
--------
If you want to test running this, you need to check out this branch, run `make develop` (so that the new executable `ctapipe-stage1-process` is generated), then run:

```
> ctapipe-stage1-process --help 
```
And follow instructions. 

You can either specify a lot of things on the command-line, or better yet use a config file similar to this:
```json
{
    "Stage1Process": {
        "config_file": "",
        "output_filename": "lapalma_proton_small.h5",
        "overwrite": true,
        "write_images": true,
        "image_extractor_type": "NeighborPeakWindowSum",
        "image_cleaner_type": "TailcutsImageCleaner"
    },
    "EventSource": {
        "allowed_tels": [],
        "input_url": "~/Data/CTA/Prod3/LaPalmaRefSim/proton_20deg_180deg_run18___cta-prod3-demo-2147m-LaPalma-baseline.simtel.gz",
        "skip_calibration_events": true
    },
    "TailcutsImageCleaner": {
        "boundary_threshold_pe": [
            ["type","*", 5.0],
            ["type", "LST*", 3.0],
            ["type", "MST*", 4.0]
        ],
        "min_picture_neighbors":[
            ["type","*",2]
        ],
        "picture_threshold_pe":  [
                ["type", "*", 10.0],
                ["type", "LST_LST_LSTCam", 6.0],
                ["type", "MST_MST_NectarCam", 6.0],
                ["id", 12, 15.0]
        ]
    },
    "ImageDataChecker": {
        "selection_functions": {
            "enough_pixels": "lambda im: np.count_nonzero(im) > 2",
            "enough_charge": "lambda im: im.sum() > 100"
        }
    },
    "ThresholdGainSelector": {
        "threshold": 4000
    }
}
```
Currently the config-file overrides anything on the command line (this is how `traitlets.config` works, but it is not ideal), so it is best to leave out anything you would want to change on the command-line from the config file. 

Output:
--------
The output can be viewed using `vitables` GUI, or accessed using the `tables` (PyTables) python module (for the DL1 data), or the Astropy `Table` or Pandas for the configuration data. 

In Vitables it looks like this (if you enable the `--write-images` option to get the DL1a data as well as the parameters):
image

Example access to the image data:
-----------------------------------
```py3
import tables
f = tables.open_file("lapalma_proton_small.h5")
ims_lst =  f.root.dl1.event.telescope.images.tel_001

images = ims_lst.col("image")
images_mc = ims_lst.col("mc_photo_electron_image") 

# plot residuals between reconstructed image and monte-carlo charge:
plt.pcolormesh(images - images_mc)
```
image

Here, the x-axis is *pixel_id* and the y axis is *Event count*.

```py3
plt.hist( (images - images_mc).ravel() , bins=100)
```
image


bregeon:
just a quick question to understand better the context, how much different is "ctapipe-stage1-process" from "protopipe write_dl1.py"

vuillaut:
Related to #1165 
In the R0/R1/DL0/DL1 containers, we map telescopes per `tel_id`. Why not keep the same in the DL1 file structure and not merge them per `telescope_type` as it is done now?
The file will be produced per telescope anyway so it does make sense to have one table per `tel_id`, no?

kosack:
> just a quick question to understand better the context, how much different is "ctapipe-stage1-process" from "protopipe write_dl1.py"

It will replace it in the next major protopipe release

kosack:
@vuillaut  Though I suppose you are right, in real data, we'd split by tel_id, not type, so perhaps it is easier to do that. It's an easy change, but it then makes making a plot of e.g. some quantitiy for each telescope type more annoying, since you have to merge all telescope tables first.   But this way, merging is also not too hard, since a single telescope would write a table called LST_LST_LSTCam, with `tel_id=1`

kosack:
Maybe I'll make an option to split by telescope type vs telescope id. For training or benchmarking, type is most useful, but for real data, id is better... (but then we have 2 variants of the data model... not sure that is nice either). Perhaps just an API function for easily converting between the two would be ok (e.g. something that merges all the per-telescope tables into a single per-type table on request)

watsonjj:
> Maybe I'll make an option to split by telescope type vs telescope id. For training or benchmarking, type is most useful, but for real data, id is better... (but then we have 2 variants of the data model... not sure that is nice either).

I think this would be a bad approach, for exactly as you say.

Would storing a table per telescope id be that inefficient, considering we generally have a fixed number of telescopes?

Alternatively just a high level method in the DL1Reader to obtain the table for a telid might be sufficient

kosack:
> Would storing a table per telescope id be that inefficient, considering we generally have a fixed number of telescopes?

There is overhead for each table (metadata, etc), but not sure how much that is.  If you think of this in database terms, nobody would ever store multiple tables with the same schema since that's hard to manage, they would just invent a new index. But here, since merging would be easer (just linking the telescope data sets into a single file), it may be the right way to go.

MaxNoe:
A third possibility would be to use hdf5 variable length arrays in just a single table.

From the data point of view, I like this the most.

I don't know, what the performance impact will be.

kosack:
By the way, just to try to explain future plans: the idea for this Tool would be eventually to be structured like this:
image


Where the ImageProcessor is a Component that includes an ImageCleaner and ImageParameterizer (such that you could also have multiple ImageProcessors, if you want, like in the current  ProtoPipe where one uses island cleaning and the other not).

Currently the Monitoring data stream is not needed, since we are working only with Monte-Carlo, but it will be needed for real data. 

It's not quite structured in such a modular way so far, but getting there.

kosack:
> A third possibility would be to use hdf5 variable length arrays in just a single table.

We tried that a while back, and the performance was pretty bad - you lose a lot of the HPC features. It also makes it hard to read anything with pandas, etc.

kosack:
> Alternatively just a high level method in the DL1Reader to obtain the table for a telid might be sufficient

Yes, that's what I meant by an API function.  I think it may be the right way to go, but it means you can't 
 just use Pandas or PyTables in an easy way - you have to go through another API first, which is somewhat annoying.

MaxNoe:
> We tried that a while back, and the performance was pretty bad - you lose a lot of the HPC features. It also makes it hard to read anything with pandas, etc.

Reading arrays with pandas is always bad, does not matter whether same shape per row or different length.

This  is what I tried: 

```python
import tables
import numpy as np
from tables import IsDescription, UInt64Col, UInt16Col


class DL1(IsDescription):
    array_event_id = UInt64Col()
    telescope_id = UInt16Col()


telescope_events = [
    (1, 1, np.random.normal(size=1440)),
    (1, 2, np.random.normal(size=1039)),
    (1, 3, np.random.normal(size=1855)),
    (2, 3, np.random.normal(size=1855)),
    (2, 4, np.random.normal(size=1855)),
    (3, 1, np.random.normal(size=1440)),
] 


with tables.open_file('test.hdf5', mode='w') as f:
    f.create_group('/', 'dl1')
    table = f.create_table('/dl1', 'telescope_events', DL1)

    images = f.create_vlarray('/dl1', 'images', tables.Float32Atom())

    for event_id, tel_id, img in telescope_events:
        row = table.row
        row['array_event_id'] = event_id
        row['telescope_id'] = tel_id
        row.append()
        images.append(img)
```

MaxNoe:
This has the advantage, that you can read the data that has the same shape for all telescope nicely using pandas and for the images, you get lists of numpy arrays, which seems to be not that bad compared to multiple tables.

MaxNoe:
For me, reading the variable length images is a little faster than reading from three same shape tables:

```
In [1]: import h5py

In [2]: f = h5py.File('./vlarray.hdf5')

In [3]: %timeit images = f['dl1/images'][:]
3.02 ms ± 170 µs per loop (mean ± std. dev. of 7 runs, 100 loops each)

In [4]: f = h5py.File('tables.hdf5')

In [5]: %timeit images = [f[f'dl1/type_{i}']['image'][:] for i in range(3)]
3.91 ms ± 13 µs per loop (mean ± std. dev. of 7 runs, 100 loops each)
```

The variable length array file is slightly larger, I don't know why.

Here is the code:
```python

import tables
import numpy as np
from tables import IsDescription, UInt64Col, UInt16Col, Float32Col

# create some pseudo data
np.random.seed(0)

num_pixels = [1855, 1039, 2048]
telescope_types = [0] * 4 + [1] * 10 + [2] * 20
telescope_ids = np.arange(len(telescope_types))
n_events = 100


telescope_events = []
for event_id in range(n_events):
    n_telescopes = min(2 + np.random.poisson(10), len(telescope_ids))
    triggered = np.random.choice(telescope_ids, n_telescopes, replace=False)

    for tel_id in triggered:
        size = num_pixels[telescope_types[tel_id]]
        telescope_events.append((
            event_id, tel_id, np.random.normal(size=size)
        ))


# write out as one table with metadata and one variable length array column
class DL1(IsDescription):
    array_event_id = UInt64Col()
    telescope_id = UInt16Col()


with tables.open_file('vlarray.hdf5', mode='w') as f:
    f.create_group('/', 'dl1')
    table = f.create_table('/dl1', 'telescope_events', DL1)

    images = f.create_vlarray('/dl1', 'images', tables.Float32Atom())

    for event_id, tel_id, img in telescope_events:
        row = table.row
        row['array_event_id'] = event_id
        row['telescope_id'] = tel_id
        row.append()
        images.append(img)


# write out as same shape tables for each telescope type
def description(num_pixel):
    class DL1(IsDescription):
        array_event_id = UInt64Col()
        telescope_id = UInt16Col()
        image = Float32Col(shape=(num_pixel))

    return DL1


with tables.open_file('tables.hdf5', mode='w') as f:
    f.create_group('/', 'dl1')
    tables = {}

    for event_id, tel_id, img in telescope_events:
        tel_type = telescope_types[tel_id]
        if tel_type not in tables:
            desc = description(num_pixels[tel_type])
            tables[tel_type] = f.create_table('/dl1', f'type_{tel_type}', desc)

        table = tables[tel_type]
        row = table.row
        row['array_event_id'] = event_id
        row['telescope_id'] = tel_id
        row['image'] = img
        row.append()
```

vuillaut:
> > Would storing a table per telescope id be that inefficient, considering we generally have a fixed number of telescopes?
> 
> There is overhead for each table (metadata, etc), but not sure how much that is. If you think of this in database terms, nobody would ever store multiple tables with the same schema since that's hard to manage, they would just invent a new index. But here, since merging would be easer (just linking the telescope data sets into a single file), it may be the right way to go.

Trying to compare the advantages and drawbacks of both approaches, here is what I can come up with:
* Per `tel_type`
    - Advantages:
        - easy plotting and benchmarks across same telescope type
        - easy training for one telescope type
    - Drawback:
         - different data model between R1/DL0 and DL1
         - when processing tables from R1/DL0 (per `tel_id`), we won't be able to do that in parallel per table since that would be different processes writing in the same output.
* Per `tel_id`
    - Advantages:
         - consistent data model from R1 to DL1
         - efficient analysis and writing 
    - Drawback:
         - a bit more annoying benchmark and training per telescope type. 
         - more metadata

In the case of per `tel_id`, getting a table of all `tel_type` is not that difficult, if you have a table of `tel_id` per `tel_type`, that is two lines instead of one:
```
tel_ids = table_of_tel_ids[tel_type]
pd.concat([pd.read_hdf(df(filename, key=f'path_to_{tel_id}') for tel_id in tel_ids])
```
(something similar with `pytables`)


Of course, I like loading all my parameters at once, but I am not sure this outweigh the drawbacks.

vuillaut:
> > A third possibility would be to use hdf5 variable length arrays in just a single table.
> 
> We tried that a while back, and the performance was pretty bad - you lose a lot of the HPC features. It also makes it hard to read anything with pandas, etc.

There is also the compression issue. It seems varlen arrays are not (well) compressible.

MaxNoe:
> That might well be compression.

I did not enable any compression for either.

kosack:
> The variable length array file is slightly larger, I don't know why.

I think there is some extra header info you have to store with each chunk for var-length arrays (to know how much of the array is stored in that chunk), so there is overhead.

kosack:
I don't think we should worry about supporting reading image data into pandas, only parameter data (for images we should use either pytables or astropy tables, which natively support multidimensional columns).

kosack:
Here is also another way one can specify a config file (autogenerated from the tool itself, in python format rather than JSON). (I generated this with `tool.generate_sample_config()`). Note that there are a few places where we should clean up the docstrings (like EventSource, where some should move to user-docs). This format is perhaps nicer than the JSON variant, since it contains comments, etc. 

```py
# Configuration file for ctapipe-stage1-process.

#------------------------------------------------------------------------------
# Application(SingletonConfigurable) configuration
#------------------------------------------------------------------------------

## This is an application.

## The date format used by logging formatters for %(asctime)s
#c.Application.log_datefmt = '%Y-%m-%d %H:%M:%S'

## The Logging format template
#c.Application.log_format = '[%(name)s]%(highlevel)s %(message)s'

## Set the log level by value or name.
#c.Application.log_level = 30

#------------------------------------------------------------------------------
# Tool(Application) configuration
#------------------------------------------------------------------------------

## This is an application.

## name of a configuration file with parameters to load in addition to command-
#  line parameters
#c.Tool.config_file = ''

## The Logging format template
#c.Tool.log_format = '%(levelname)s [%(name)s] (%(module)s/%(funcName)s): %(message)s'

#------------------------------------------------------------------------------
# Stage1Process(Tool) configuration
#------------------------------------------------------------------------------

## compression level, 0=None, 9=maximum
#c.Stage1Process.compression_level = 5

## compressor algorithm to use.
#c.Stage1Process.compression_type = 'zlib'

## GainSelector to use.
#c.Stage1Process.gain_selector_type = 'ThresholdGainSelector'

## ImageCleaner to use.
#c.Stage1Process.image_cleaner_type = 'TailcutsImageCleaner'

## Method to use to turn a waveform into a single charge value
#c.Stage1Process.image_extractor_type = 'NeighborPeakWindowSum'

## DL1 output filename
#c.Stage1Process.output_filename = 'events.dl1.h5'

## overwrite output file if it exists
#c.Stage1Process.overwrite = False

## show progress bar during processing
#c.Stage1Process.progress_bar = False

## Store DL1/Event/Image data in output
#c.Stage1Process.write_images = False

## Generate PyTables index datasets for all tables that contain an event_id or
#  tel_id. These speed up in-kernal pytables operations,but add some overhead to
#  the file. They can also be generated and attached after the file is written
#c.Stage1Process.write_index_tables = False

## Compute and store image parameters
#c.Stage1Process.write_parameters = True

#------------------------------------------------------------------------------
# EventSource(Component) configuration
#------------------------------------------------------------------------------

## Parent class for EventFileReaders of different sources.
#
#  A new EventFileReader should be created for each type of event file read into
#  ctapipe, e.g. sim_telarray files are read by the `SimTelEventSource`.
#
#  EventFileReader provides a common high-level interface for accessing event
#  information from different data sources (simulation or different camera file
#  formats). Creating an EventFileReader for a new file format ensures that data
#  can be accessed in a common way, irregardless of the file format.
#
#  EventFileReader itself is an abstract class. To use an EventFileReader you
#  must use a subclass that is relevant for the file format you are reading (for
#  example you must use `ctapipe.io.SimTelEventSource` to read a hessio format
#  file). Alternatively you can use `event_source()` to automatically select the
#  correct EventFileReader subclass for the file format you wish to read.
#
#  To create an instance of an EventFileReader you must pass the traitlet
#  configuration (containing the input_url) and the `ctapipe.core.tool.Tool`.
#  Therefore from inside a Tool you would do:
#
#  >>> event_source = EventSource(self.config, self)
#
#  An example of how to use `ctapipe.core.tool.Tool` and `event_source()` can be
#  found in ctapipe/tools/display_dl1.py.
#
#  However if you are not inside a Tool, you can still create an instance and
#  supply an input_url via:
#
#  >>> event_source = EventSource( input_url="/path/to/file")
#
#  To loop through the events in a file:
#
#  >>> event_source = EventSource( input_url="/path/to/file") >>> for event in
#  event_source: >>>    print(event.count)
#
#  **NOTE**: Every time a new loop is started through the event_source, it
#  restarts from the first event.
#
#  Alternatively one can use EventFileReader in a `with` statement to ensure the
#  correct cleanups are performed when you are finished with the event_source:
#
#  >>> with EventSource( input_url="/path/to/file") as event_source: >>>    for
#  event in event_source: >>>       print(event.count)
#
#  **NOTE**: The "event" that is returned from the generator is a pointer. Any
#  operation that progresses that instance of the generator further will change
#  the data pointed to by "event". If you wish to ensure a particular event is
#  kept, you should perform a `event_copy = copy.deepcopy(event)`.
#
#  Attributes
#  ----------
#  input_url : str
#      Path to the input event file.
#  max_events : int
#      Maximum number of events to loop through in generator
#  metadata : dict
#      A dictionary containing the metadata of the file. This could include:
#      * is_simulation (bool indicating if the file contains simulated events)
#      * Telescope:Camera names (list if file contains multiple)
#      * Information in the file header
#      * Observation ID

## list of allowed tel_ids, others will be ignored. If left empty, all telescopes
#  in the input stream will be included
#c.EventSource.allowed_tels = set()

## Path to the input file containing events.
#c.EventSource.input_url = ''

## Maximum number of events that will be read from the file
#c.EventSource.max_events = None

#------------------------------------------------------------------------------
# DataChecker(Component) configuration
#------------------------------------------------------------------------------

## Manages a set of selection criteria that operate on the same type of input.
#  Each time it is called, it returns a boolean array of whether or not each
#  criterion passed.

## dict of '<cut name>' : lambda function in string format to accept (select) a
#  given data value.  E.g. `{'mycut': 'lambda x: x > 3'}`
#c.DataChecker.selection_functions = {}

#------------------------------------------------------------------------------
# ImageDataChecker(DataChecker) configuration
#------------------------------------------------------------------------------

## for configuring image-wise data checks

## dict of '<cut name>' : lambda function in string format to accept (select) a
#  given data value.  E.g. `{'mycut': 'lambda x: x > 3'}`
#c.ImageDataChecker.selection_functions = {'enough_pixels': 'lambda im: np.count_nonzero(im) > 2', 'enough_charge_pe': 'lambda im: im.sum() > 80'}

#------------------------------------------------------------------------------
# TailcutsImageCleaner(ImageCleaner) configuration
#------------------------------------------------------------------------------

## Apply Image Cleaning to a set of telescope images

## second-level threshold in photoelectrons
#c.TailcutsImageCleaner.boundary_threshold_pe = [('type', '*', 5.0)]

## Minimum number of neighbors above threshold to consider
#c.TailcutsImageCleaner.min_picture_neighbors = [('type', '*', 2)]

## top-level threshold in photoelectrons
#c.TailcutsImageCleaner.picture_threshold_pe = [('type', '*', 10.0)]

#------------------------------------------------------------------------------
# FixedWindowSum(ImageExtractor) configuration
#------------------------------------------------------------------------------

## Extractor that sums within a fixed window defined by the user.

## Define the start position for the integration window
#c.FixedWindowSum.window_start = 0

## Define the width of the integration window
#c.FixedWindowSum.window_width = 7

#------------------------------------------------------------------------------
# GlobalPeakWindowSum(ImageExtractor) configuration
#------------------------------------------------------------------------------

## Extractor which sums in a window about the peak from the global average
#  waveform.

## Define the shift of the integration window from the peak_index (peak_index -
#  shift)
#c.GlobalPeakWindowSum.window_shift = 3

## Define the width of the integration window
#c.GlobalPeakWindowSum.window_width = 7

#------------------------------------------------------------------------------
# LocalPeakWindowSum(ImageExtractor) configuration
#------------------------------------------------------------------------------

## Extractor which sums in a window about the peak in each pixel's waveform.

## Define the shift of the integration window from the peak_index (peak_index -
#  shift)
#c.LocalPeakWindowSum.window_shift = 3

## Define the width of the integration window
#c.LocalPeakWindowSum.window_width = 7

#------------------------------------------------------------------------------
# NeighborPeakWindowSum(ImageExtractor) configuration
#------------------------------------------------------------------------------

## Extractor which sums in a window about the peak defined by the wavefroms in
#  neighboring pixels.

## Weight of the local pixel (0: peak from neighbors only, 1: local pixel counts
#  as much as any neighbor)
#c.NeighborPeakWindowSum.lwt = 0

## Define the shift of the integration window from the peak_index (peak_index -
#  shift)
#c.NeighborPeakWindowSum.window_shift = 3

## Define the width of the integration window
#c.NeighborPeakWindowSum.window_width = 7

#------------------------------------------------------------------------------
# BaselineSubtractedNeighborPeakWindowSum(NeighborPeakWindowSum) configuration
#------------------------------------------------------------------------------

## Extractor that first subtracts the baseline before summing in a window about
#  the peak defined by the wavefroms in neighboring pixels.

## End sample for baseline estimation
#c.BaselineSubtractedNeighborPeakWindowSum.baseline_end = 10

## Start sample for baseline estimation
#c.BaselineSubtractedNeighborPeakWindowSum.baseline_start = 0

#------------------------------------------------------------------------------
# ManualGainSelector(GainSelector) configuration
#------------------------------------------------------------------------------

## Manually choose a gain channel.

## Which gain channel to retain
#c.ManualGainSelector.channel = 'HIGH'

#------------------------------------------------------------------------------
# ThresholdGainSelector(GainSelector) configuration
#------------------------------------------------------------------------------

## Select gain channel according to a maximum threshold value.

## Threshold value in waveform sample units. If a waveform contains a sample
#  above this threshold, use the low gain c
```

kosack:
The default is now to split by tel_id, as @vuillaut suggests (I retained the ability to do the old behavior as well, but it's a hidden option).

kosack:
I found a possible problem with splitting by telescope_id... I just tried this on a 567-telescope Prod3b file and when the file is closed, I get over 600 warning messages similar to this, and the output file is corrupt (as if some of the datasets were not completely written).  So there may be a problem writing this many data sets at once.  When using the `Stage1Process.split_datasets_by=tel_type`, it does not happen. 

```
Exception ignored in: <function Node.__del__ at 0x7fb1c0483c80>
  File "/Users/kkosack/anaconda3/envs/cta-dev/lib/python3.7/site-packages/tables/node.py", line 321, in __del__
    self._f_close()
  File "/Users/kkosack/anaconda3/envs/cta-dev/lib/python3.7/site-packages/tables/table.py", line 2966, in _f_close
    self.flush()
  File "/Users/kkosack/anaconda3/envs/cta-dev/lib/python3.7/site-packages/tables/table.py", line 2900, in flush
    self.row._flush_buffered_rows()
  File "tables/tableextension.pyx", line 1331, in tables.tableextension.Row._flush_buffered_rows
  File "tables/tableextension.pyx", line 746, in tables.tableextension.Row.table.__get__
  File "/Users/kkosack/anaconda3/envs/cta-dev/lib/python3.7/site-packages/tables/file.py", line 2159, in _check_open
    raise ClosedFileError("the file object is closed")
```

Anyone know if there is some limitation in the pytables library, or hdf5 itself? I suspect interleaved writing of many datasets to the same file could be problematic...   

This happens both for the full file (a few thousand events) or when restricting it to 10 events, so the problem seems to be the number of datasets.  Perhaps this is more reason to perhaps use the telescope_Type splitting, and just rely on sorting by event_id to do the merge.

vuillaut:
> I found a possible problem with splitting by telescope_id... I just tried this on a 567-telescope Prod3b file and when the file is closed, I get over 600 warning messages similar to this, and the output file is corrupt (as if some of the datasets were not completely written). So there may be a problem writing this many data sets at once. When using the `Stage1Process.split_datasets_by=tel_type`, it does not happen.
> 
> ```
> Exception ignored in: <function Node.__del__ at 0x7fb1c0483c80>
> Traceback (most recent call last):
>   File "/Users/kkosack/anaconda3/envs/cta-dev/lib/python3.7/site-packages/tables/node.py", line 321, in __del__
>     self._f_close()
>   File "/Users/kkosack/anaconda3/envs/cta-dev/lib/python3.7/site-packages/tables/table.py", line 2966, in _f_close
>     self.flush()
>   File "/Users/kkosack/anaconda3/envs/cta-dev/lib/python3.7/site-packages/tables/table.py", line 2900, in flush
>     self.row._flush_buffered_rows()
>   File "tables/tableextension.pyx", line 1331, in tables.tableextension.Row._flush_buffered_rows
>   File "tables/tableextension.pyx", line 746, in tables.tableextension.Row.table.__get__
>   File "/Users/kkosack/anaconda3/envs/cta-dev/lib/python3.7/site-packages/tables/file.py", line 2159, in _check_open
>     raise ClosedFileError("the file object is closed")
> ```
> 
> Anyone know if there is some limitation in the pytables library, or hdf5 itself? I suspect interleaved writing of many datasets to the same file could be problematic...
> 
> This happens both for the full file (a few thousand events) or when restricting it to 10 events, so the problem seems to be the number of datasets. Perhaps this is more reason to perhaps use the telescope_Type splitting, and just rely on sorting by event_id to do the merge.

Damn.
It does not seem to be an HDF5 limitation though:
See e.g. https://stackoverflow.com/questions/35321093/limit-on-number-of-hdf5-datasets

kosack:
It could be something to do with the default chunk size:
It says in the pytables documentation: (https://www.pytables.org/usersguide/optimization.html):

> But if your datasets are small (say, up to 10 MB) or your number of nodes is contained (up to 1000), you > should not worry about that as the default parameters in PyTables are already tuned for those sizes (although you may want to adjust them further anyway). At any rate, reading this chapter will help you in > your life with PyTables.

here, the number of nodes is well over 1000 (due to the parameters and images datasets being split)

devine73:
I everyone,
I had this warnings also on the simtel -> hdf5 converter we developed.
Tables complains because the number of nodes opened at the same time is greater than the default value of tables.parameters.NODE_CACHE_SLOTS.
To avoid that we did some : 
     tables.parameters.NODE_CACHE_SLOTS = max(tables.parameters.NODE_CACHE_SLOTS, 3*nbTel + 20)
In this case, hdf5 is still performant event if is not the default usage.
Cheers,
Pierre

kosack:
@devine73 Thanks! That solved it.

labsaha:
@kosack  Any solution to  this issue on RA & DEC [https://github.com/cta-observatory/cta-lstchain/issues/270](url)?

kosack:
Even though this PR is not in a perfect state, it works, and I suggest we merge it into master since the code has no influence on the rest of ctapipe (it's a self-contained script).  

That will allow us to do format testing using it, and subsequent PRs will re-factor this into Components with proper unit tests as mentioned above. Otherwise, it is currently blocking us from proceeding.

ctapipe/tools/stage1.py
```diff
@@ -0,0 +1,868 @@
...
+    max = Field(np.nan, "value of pixel with maximum intensity")
+    min = Field(np.nan, "value of pixel with minimum intensity")
+    mean = Field(np.nan, "mean intensity")
+    std = Field(np.nan, "standard deviation of intensity")
+
+
+def intensity_statistics(image) -> IntensityContainer:
+    """ compute intensity statistics of an image  """
+    return IntensityContainer(
+        max=image.max(), min=image[image > 0].min(), mean=image.mean(), std=image.std()
```
MaxNoe: This > 0 is a bit surprising to me. shouldn't it be everywhere or nowhere?

So better call `image_statistics(image[image > 0])` instead and don't apply any masking here?
also include higher moments?

bregeon: To make the comment more explicit, 'intensity_statistics' is called on clean images, so to get the min, one clearly has to get zeroes out of the way... then one has to decide whether to keep them or no for the mean and std.
From the point of view of statistics, I agree with Max, better call `image_statistics(image[image > 0])`

ctapipe/tools/stage1.py
```diff
@@ -0,0 +1,868 @@
...
+    MorphologyContainer:
+        parameters related to the morphology
+    """
+
+    num_islands, island_labels = number_of_islands(geom=geom, mask=image_mask)
+
+    return MorphologyContainer(num_pixels=image_mask.sum(), num_islands=num_islands)
+
+
+class IntensityContainer(Container):
```
moralejo: The use of Intensity both for the charge of individual pixels and for the total image charge seems prone to confusion: "mean intensity" or "standard deviation of intensity" can refer to pixels within an image(like here), or to images within a run.

ctapipe/tools/stage1.py
kosack: yes agreed, it's clearer. I don't like hidden cuts

kosack:
Re the comments from @bregeon, yes most of these would be fixed in a refactoring. It's clear there is work to be done to make this nicer from the standpoint of a library function, but not sure it's necessary for testing.  The main point here, is this is a quick and dirty script to do some prototyping, but it will nearly all disappear after refactoring into library functions. 

> - many small classes, methods and functions lack basic documentation

yes, these will be documents as each function is refactored into the main library 

> - for MC the subarray is written in the DL0/R1 file, what about real DL0 data?

not sure I understand - this is only DL1, we are not defining DL0 here. I would guess it would also being a file, or at least a reference to a database entry. DL0 is another story altogether.  ONce we have some standardized subarrays, however, we might be able to drop the full subarray here, and just have an index key to a standard one. Again, that's more for a data model discussion. 

> - there is no place holder for data reduction: real data will be reduced (to DL0) the information will have to be propagated to DL1, and we will want to simulate data reduction on MC and that may happen in that stage1 tool.

You mean data volume reduction? That is not covered so far in this DL1 format or prototype.  THe CTA requirements are that we don't do DVR until after the first year of operations or so, so I'm not too worried - there will be many DL1 model changes in between. Proabbly dont' need to worry about it here, and just assume if we do some preliminary DVR, we'll just write 0s (might need to add a mask somewhere for it though). 

> - I was not able to figure out what happens if the output file exists but overwrite is set to false, what is the error message?

You get a message from the pytables library - it will not overwrite by default, and can only add datasets, not modify existing ones. 

> - I would like information about the behavior of the writer and the full stage1 script in case there is an issue when the writer is called, is there a return code? (helpful when run in production)

so far it's in the provenance info if it failed, but there's a PR open (for ctapipe in general) #1214 on that subject.  the Tool class will handle that. 

> - I am wondering about a potential security issue with the setup of the selector, there seems to be a bold eval of a string (what If my selector is rm -rf / ?):
`name: eval(func_str) for name, func_str in selection_functions.items()`

Yes, I thought of that, but not sure there is a better way. Anyhow, that is a discussion for  #1207 ,  not this PR (once that's merged, it will be removed from here)

> - why is zlib the default compression and not zstd?

Don't remember - I think it was just that zstd doesn't work in HDFView or maybe it was vitables? We can choose a better one once a study is done. 

> - only DL1/event is added to Provenance, what about DL1/image?

Not sure I understand - the images are inside DL1/Event, right?

> - I would recommend to speficify data types in the Field instead of letting the code find it (cf. comment in `Stage1ProcessorTool`).

see #1224 and all related PRs. Not really a question for this PR, but agreed. 

> - I would log a debug message when no histograms are written but the config files asks for it.

yes, good idea. At least a warning. In a side note, I found there is a way to capture all warnings in the logger, so that will probably be added to the Tool base class eventually. 

> -   the `image>0` cut in the intensity container should be fixed, as already reported.

agreed

bregeon:
@kosack Thank you very much for your quick answers, I am fine with all of them, and clearly there were a few misunderstanding on my side. Thanks for the pointers to other PRs.
I confirm I can approve that PR once the 'image>0' is moved to:
`params.intensity = intensity_statistics(image=clean_image)`

moralejo:
> You can either specify a lot of things on the command-line, or better yet use a config file similar to this:
>
> {
>     "Stage1Process": {
>         "config_file": "",
>         "output_filename": "lapalma_proton_small.h5",
>         "overwrite": true,
>         "write_images": true,
>         "image_extractor_type": "NeighborPeakWindowSum",
>         "image_cleaner_type": "TailcutsImageCleaner"
>     },

The name should be "Stage1ProcessorTool" (it is correct in examples/stage1_config.json).  
Not really belonging to this PR, but: in general, it would be good that if an unknown key is found in the config file, an error message is shown, and the program stops at that point - otherwise it may be complicated to realize what is wrong if the program fails elsewhere due to the missing configurations.

ctapipe/tools/stage1.py
```diff
@@ -0,0 +1,868 @@
...
+                    tel_index,
+                    params.hillas,
+                    params.timing,
+                    params.leakage,
+                    params.concentration,
+                    params.morphology,
+                    params.intensity,
+                ]
+
+                # currently the HDF5TableWriter has problems if the first event
```
moralejo: Is there then any problem for specifying them in the fields straightaway? Otherwise, could we for the time being not use nan but some conventional default error values? Having in the output the events for which the parameterisation failed may be useful. Another "ugly" alternative: we could just make sure that the first event has no nans, but keep all subsequent ones.

kosack: Acutally, that was the original intent (to write all events) ,but due to how things work currently, it's not possible - at some point we changed many of the "Default values" in the containers to either None or np.nan, which makes it impossible to guess the data type if the info is not filled in (and is nan is not supported by pytables).  Since the schema is generated by the info in the first event, it fails if the first event is not parameterizable.  In #1224  I suggested a fix for that, but I think for now we have to wait for that fix, and then in the future, we can store all events.    It's not really clear what is a good "junk" value for all parameters, however, particularly for integers.

MaxNoe: This ja a hot topic currently nearly everywhere in data science. Newer file formats and libraries start supporting a general missing value indicator.

E.g. pandas since 1.0 and Apache arrow

ctapipe/tools/stage1.py
```diff
@@ -0,0 +1,868 @@
...
+    return MorphologyContainer(num_pixels=image_mask.sum(), num_islands=num_islands)
+
+
+class IntensityContainer(Container):
+    """ Store statistics on the intensity distribution of images """
+
+    max = Field(np.nan, "value of pixel with maximum intensity")
+    min = Field(np.nan, "value of pixel with minimum intensity")
+    mean = Field(np.nan, "mean intensity")
+    std = Field(np.nan, "standard deviation of intensity")
```
rlopezcoto: is it ok to inizialize all these fields with `np.nan`? If none of the images pass the filters, will the writing of `nan` values not throw any errors?

ctapipe/tools/stage1.py
```diff
@@ -0,0 +1,868 @@
...
+    
+    Or use an external configuration file, where you can specify all options:
+    > ctapipe-stage1-process --config stage1_config.json --progress 
+    
+    The config file should be in JSON or python format (see traitlets docs). For an 
+    example, see ctapipe/examples/stage1_config.json in the main code repo. 
+    """
+
+    output_filename = Unicode(
+        help="DL1 output filename", default_value="events.dl1.h5"
```
rlopezcoto: pep8: `help="` -> `help = "`
Same below

ctapipe/tools/stage1.py
```diff
@@ -0,0 +1,868 @@
...
+        + classes_with_traits(ImageExtractor)
+        + classes_with_traits(GainSelector)
+    )
+
+    def setup(self):
+
+        # prepare output path:
+
+        output_path = Path(expandvars(self.output_filename)).expanduser()
+        if output_path.exists() and self.overwrite:
```
rlopezcoto: what about if `output_path.exists() and not self.overwrite`

ctapipe/tools/stage1.py
```diff
@@ -0,0 +1,868 @@
...
+        "CTA INSTRUMENT SUBTYPE": "",
+        "CTA INSTRUMENT VERSION": "",
+        "CTA INSTRUMENT ID": str(subarray),
+    }
+
+    for key, value in metadata.items():
+        writer._h5file.root._v_attrs[key] = value
+
+
+def morphology(geom, image_mask) -> MorphologyContainer:
```
rlopezcoto: Is the calculation of number of islands usually known as `Morphology`? It may be, but I do not seem to recall hearing it before...

ctapipe/tools/stage1.py
MaxNoe: Nan is the only safe default. Everything else might be a valid value for some of these fields or another.

We are long planning the introduction of static typing for fields, see #445 #501 #1224 

I'm planning to work on this as soon as I defended my thesis (april)

ctapipe/tools/stage1.py
MaxNoe: No, pep8 says no space for around = for keyword arguments.

ctapipe/tools/stage1.py
MaxNoe: me neither...

But Morphology is probably meant as kind of a catch all container for non-hillas image related parameters, so islands fits here

ctapipe/tools/stage1.py
bregeon: Karl answered this above:

> You get a message from the pytables library - it will not overwrite by default, and can only add datasets, not modify existing ones.

ctapipe/tools/stage1.py
bregeon: I agree we can imagine more than just number of islands to be added to this "morphology" container at some point. I quite like the name actually!

ctapipe/tools/stage1.py
kosack: we could check and throw a better error message though (the message from pytables is not always clear)

ctapipe/tools/stage1.py
kosack: yes, what @MaxNoe said is right - I couldn't think of a better name, and we could add other image-morphology parameters here later. Maybe Topology is better?

bregeon: Topology also came to my mind as well, but I'd say we can change that name if we wish according to its "final" content at a later point...

bregeon:
It seems to me we had enough reviews of this PR :)

Here is a summary of comments for which there is no issue opened yet:
* AM: The use of Intensity both for the charge of individual pixels and for the total image charge seems prone to confusion.
* JB: Data Volume Reduction is not covered by current format
* JB, RLC: Error message when output file exists but not overwritten
* JB: "eval(func_str)" not ideal from the security point of view
* JB: switch to zstd compression at some point
* JB: log warning when no histogram is written but this was asked
* RLC: Morphology or Topology, think about a better name?

All these comments are very minor, can be fixed later and do not prevent this PR from being approved.

@kosack would you have time this week to just fix the `image>0` cut so that we can approve and merge this PR?

MaxNoe:
> "eval(func_str)" not ideal from the security point of view

This would be fairly easy to fix by giving it a dictionary of stuff it is allowed to use. E.g. as a first step,

```python
eval_globals = {
    'u': u,       # astropy units
    'np': np,   # numpy,
    # more?
}
eval(func_str, eval_globals)
```
this would prevent stuff like `subprocess.run('rm -rf *')`

kosack:
> This would be fairly easy to fix by giving it a dictionary of stuff it is allowed to use. E.g. as a first step,

 I referenced that comment in #1207 (once that's merged it replaces what is done in this PR)

kosack:
Currently waiting on these to be accepted to replace or fix functionality in this PR:
- [ ] #1168 (command-line bug fix)
- [ ] #1207 (event selector)
- [ ] #1221 CTA reference metadata

ctapipe/tools/stage1.py
```diff
@@ -0,0 +1,767 @@
...
+        image of pixels surviving cleaning (True=survives)
+    Returns
+    -------
+    MorphologyContainer:
+        parameters related to the morphology
+    """
+
+    num_islands, island_labels = number_of_islands(geom=geom, mask=image_mask)
+
+    return MorphologyContainer(num_pixels=image_mask.sum(), num_islands=num_islands)
```
MaxNoe, line 128: `np.count_nonzero`

ctapipe/tools/stage1.py
MaxNoe: And you can calculate the std of the std of the pixels for each run and then the std of the std of the std over multiple runs. 

This data is stored in `/dl1/event/telescope/parameters` which makes it pretty clear, doesn't it?

ctapipe/tools/stage1.py
MaxNoe: Just call it on the cleaned image or `image[cleaning]`

ctapipe/tools/stage1.py
```diff
@@ -0,0 +1,767 @@
...
+
+class ExtendedImageParametersContainer(ImageParametersContainer):
+    """
+    Extra parameters to add to the ImageParametersContainer
+
+    TODO: should eventually just move to ImageParametersContainer.
+    """
+
+    intensity = Field(IntensityContainer(), "intensity statistics")
+    mc_intensity = Field(IntensityContainer(), "MC intensity statistics")
```
MaxNoe, line 155: This is an mc value and should probably not go into a container mixed with non-mc data.

ctapipe/tools/stage1.py
```diff
@@ -0,0 +1,767 @@
...
+
+
+class ExtendedImageParametersContainer(ImageParametersContainer):
+    """
+    Extra parameters to add to the ImageParametersContainer
+
+    TODO: should eventually just move to ImageParametersContainer.
+    """
+
+    intensity = Field(IntensityContainer(), "intensity statistics")
```
MaxNoe, line 154: This could go into the morphology container

ctapipe/tools/stage1.py
```diff
@@ -0,0 +1,767 @@
...
+    TODO: update MCCameraEventContainer and DL1TelescopeContainer; remove this
+    """
+
+    container_prefix = ""
+
+    mc_photo_electron_image = Field(
+        None, "Monte-carlo image of photo electrons on the camera plane, without noise"
+    )
+
+    image_mask = Field(None, "Boolean array of pixels, True=used in parameterization")
```
MaxNoe, line 171: Isn't this already part of the DL1 container?

ctapipe/tools/stage1.py
```diff
@@ -0,0 +1,767 @@
...
+    split_datasets_by = CaselessStrEnum(
+        values=["tel_id", "tel_type"],
+        default_value="tel_id",
+        help="Splitting level for the parameters and images datasets",
+    ).tag(config=True)
+
+    compression_type = CaselessStrEnum(
+        values=["blosc:zstd", "zlib"],
+        help="compressor algorithm to use. ",
+        default_value="zlib",
```
MaxNoe, line 265: Why use the slower, less compressing method as default? zstd was supported by pytables for quite some time now and does not require extra dependencies.

ctapipe/tools/stage1.py
```diff
@@ -0,0 +1,767 @@
...
+    def _write_simulation_histograms(self, writer: HDF5TableWriter):
+        """ Write the distribution of thrown showers
+
+        Notes
+        -----
+        - this only runs if this is a simulation file. The current implementation is
+          a bit of a hack and implies we should improve SimTelEventSource to read this
+          info.
+        - Currently the histograms are at the end of the simtel file, so if max_events
+          is set to non-zero, the end of the file may not be read, and this no
```
MaxNoe, line 426: typo in this → thus

ctapipe/tools/stage1.py
```diff
@@ -0,0 +1,767 @@
...
+
+        if type(self.event_source) is not SimTelEventSource:
+            return
+
+        hists = self.event_source.file_.histograms
+        if hists is not None:
+            hist_container = SimulatedShowerDistribution()
+            hist_container.prefix = ""
+            for hist in hists:
+                if hist["id"] == 6:
```
MaxNoe, line 464: Why only the 6th histogram? At least simulated and triggered events would be interesting, right?

ctapipe/tools/stage1.py
```diff
@@ -0,0 +1,767 @@
...
+                    params.concentration,
+                    params.morphology,
+                    params.intensity,
+                ]
+
+                # currently the HDF5TableWriter has problems if the first event
+                # has NaN as values, since it can't infer the data types.
+                # that implies we need to specify them in the Fields, rather than
+                # infer from first event, perhaps.  For now we skip them.
+                parameters_were_computed = (
```
MaxNoe, line 688: Why do this with this check and not with the config option?

ctapipe/tools/stage1.py
```diff
@@ -0,0 +1,767 @@
...
+    def start(self):
+
+        with HDF5TableWriter(
+            self.output_filename, mode="a", add_prefix=True, filters=self._hdf5_filters
+        ) as writer:
+
+            if self.write_parameters is False:
+                # don't need to write out the image mask if no parameters are computed,
+                # since we don't do image cleaning in that case.
+                writer.exclude("/dl1/event/telescope/images", "image_mask")
```
MaxNoe, line 741: Writing DL1a with cleaning applied might be interesting for the deep learning people

vuillaut: quite the contrary, the image without cleaning should always be written in DL1a

MaxNoe: Yes of course, I meant an additional cleaning masked, not the image with pixels set to 0.

kosack: You can just multiply the mask by the image, so I think it's ok for deep-learning (unless they cannot apply some math to inputs first?).

ctapipe/tools/stage1.py
kosack: this is simulated events, needed for the computation of effective areas.  We could of course store others, if there is a use case for them.  Even this can now be re-computed from the data (since we keep triggered and simulated event info).

ctapipe/tools/stage1.py
```diff
@@ -0,0 +1,767 @@
...
+            params.timing = timing_parameters(
+                geom=tel.camera.geometry,
+                image=clean_image,
+                pulse_time=data.pulse_time,
+                hillas_parameters=params.hillas,
+            )
+            params.leakage = leakage(
+                geom=tel.camera.geometry, image=data.image, cleaning_mask=mask
+            )
+            params.concentration = concentration(
```
MaxNoe, line 573: I would suggest to only pass in `geom[mask]` and `image[mask]` here. concentration is quite slow if you do this for all pixels. (needs to rotate them into the shower coordinates and sort by brightness)

ctapipe/tools/stage1.py
MaxNoe: I meant that it is probably nice for them to include the cleaning mask even if no dl1b is written

ctapipe/tools/stage1.py
```diff
@@ -0,0 +1,766 @@
...
+
+            event.mc.prefix = "mc"
+            event.trig.prefix = ""
+            event_index.event_id = event.index.event_id
+            event_index.obs_id = event.index.obs_id
+            tel_index.event_id = event.index.event_id
+            tel_index.obs_id = event.index.obs_id
+            self._cur_obs_id = event.index.obs_id
+
+            # On the first event, we now have a subarray loaded, and other info, so
```
MaxNoe, line 610: subarray is loaded before the first event!

ctapipe/tools/stage1.py
```diff
@@ -0,0 +1,766 @@
...
+        is_initialized = False
+        self.event_source.subarray.info(printer=self.log.debug)
+
+        for event in tqdm(
+            self.event_source,
+            desc=self.event_source.__class__.__name__,
+            total=self.event_source.max_events,
+            unit="ev",
+            disable=not self.progress_bar,
+        ):
```
MaxNoe, line 596: break out the body of the loop into a function here?

ctapipe/tools/stage1.py
MaxNoe: > and is nan is not supported by pytables?

What do you mean with that? pytables is perfectly able to store and read nans and will correctly deduce a type of float for nans:

```
In [10]: f = tables.open_file('/tmp/test.hdf5', 'w')                                                                                          

In [11]: f.create_array('/', 'all_nan', np.full((2, 1855), np.nan))                                                                           
Out[11]: 
/all_nan (Array(2, 1855)) ''
  atom := Float64Atom(shape=(), dflt=0.0)
  maindim := 0
  flavor := 'numpy'
  byteorder := 'little'
  chunkshape := None
```

ctapipe/tools/stage1.py
MaxNoe: Better explicitly raise an error here

ctapipe/tools/stage1.py
```diff
@@ -0,0 +1,766 @@
...
+            )
+
+            container.bins_core_dist = xbins * u.m
+            container.bins_energy = 10 ** ybins * u.TeV
+            container.histogram = eventio_hist["data"]
+            container.meta["hist_title"] = eventio_hist["title"]
+            container.meta["x_label"] = "Log10 E (TeV)"
+            container.meta["y_label"] = "3D Core Distance (m)"
+
+        if type(self.event_source) is not SimTelEventSource:
```
MaxNoe, line 455: `not isinstance(self.event_source, SimTelEventSource)` ?

ctapipe/tools/stage1.py
MaxNoe: What is the error here, if nan is encountered? Why can we not write an event with nan in the hillas container?

ctapipe/tools/stage1.py
kosack: once #1286 is there, will replace with `.is_simulated`

ctapipe/tools/stage1.py
kosack: remnant from before that was implemented. Will remove

MaxNoe: ALready fixed it in my PR

ctapipe/tools/stage1.py
MaxNoe: The PR is open https://github.com/cta-observatory/ctapipe/pull/1313

ctapipe/tools/stage1.py
MaxNoe: https://github.com/kosack/ctapipe/pull/8

kosack:
Final things needed:
- [ ] dl1/event/telescope needs a time column
- [ ] dl1/monitoring/pointing needs to be there. Suggest 1 row only for MCs, since pointing is fixed, but in general this would be a table of measurement_time, parameters that one could use to interpolate to the event_time.  Perhaps 2 rows, with the same info but the start time of the mc run, and end time of the mc run (to avoid any confusion with extrapolation).

MaxNoe:
@kosack I also think that the tel_type_id and event_type fields are not really in the right place at `telescope_event_index`.
 Telescope event index should only be `obst_id, event_id, tel_id`.

Event type and time could be in an `TelescopeTriggerContainer` and the `tel_type_id` belongs into the instrument configuration, this is not needed as event-wise column I think.

kosack:
> Event type and time could be in an TelescopeTriggerContainer and the tel_type_id belongs into the instrument configuration, this is not needed as event-wise column I think.

It was there as a way to index the parameters if you write a single telescope/parameters table (which is how it was originally implemented when this PR was opened). It may no longer be necessary - so far we only split by tel_id or tel_type, so the type is now redundant. Perhaps we should just remove it.  In that sense it was an "index", just not a sequential one.

ctapipe/tools/stage1.py
```diff
@@ -0,0 +1,731 @@
...
+# define the version of the DL1 data model written here. This should be updated
+# when necessary:
+# - increase the major number if there is a breaking change to the model
+#   (meaning readers need to update scripts)
+# - increase the minor number if new columns or datasets are added
+# - increase the patch number if there is a small bugfix to the model.
+DL1_DATA_MODEL_VERSION = "v1.0.0"
+
+
+def write_reference_metadata_headers(output_path, obs_id, subarray, writer):
```
MaxNoe, line 63: output_path is unused in this function

ctapipe/tools/stage1.py
```diff
@@ -0,0 +1,731 @@
...
+
+    write_parameters = Bool(
+        help="Compute and store image parameters", default_value=True
+    ).tag(config=True)
+
+    compression_level = Int(
+        help="compression level, 0=None, 9=maximum", default_value=5, min=0, max=9
+    ).tag(config=True)
+
+    split_datasets_by = CaselessStrEnum(
```
MaxNoe, line 171: Do we want an option to not split the image parameter level?

ctapipe/tools/stage1.py
```diff
@@ -0,0 +1,731 @@
...
+            "Overwrite output file if it exists",
+        ),
+        "progress": (
+            {"Stage1ProcessorTool": {"progress_bar": True}},
+            "show a progress bar during event processing",
+        ),
+    }
+
+    classes = List(
+        [EventSource, CameraCalibrator, ImageQualityQuery]
```
MaxNoe, line 244: I think this should be `classes_with_traits(EventSource)` since one needs to configure the event source class one is actually using. Setting options for the `EventClass` does not work for subclasses.

ctapipe/tools/stage1.py
```diff
@@ -0,0 +1,731 @@
...
+        if self.write_parameters is False and self.write_images is False:
+            raise ToolConfigurationError(
+                "The options 'write_parameters' and 'write_images' are "
+                "both set to False. No output will be generated in that case. "
+                "Please enable one or both of these options."
+            )
+
+        # setup components:
+
+        self.gain_selector = self.add_component(
```
MaxNoe, line 277: Why must this be explicitly created instead of being able to configure this for the event source?

ctapipe/tools/stage1.py
```diff
@@ -0,0 +1,731 @@
...
+
+        Note that this function should be run first
+        """
+        self.log.debug("Writing simulation configuration")
+
+        class ExtraMCInfo(Container):
+            container_prefix = ""
+            obs_id = Field(0, "MC Run Identifier")
+
+        extramc = ExtraMCInfo()
```
MaxNoe, line 334: give obs_id as kwarg to container creation

ctapipe/tools/stage1.py
```diff
@@ -0,0 +1,731 @@
...
+
+        for event in tqdm(
+            self.event_source,
+            desc=self.event_source.__class__.__name__,
+            total=self.event_source.max_events,
+            unit="ev",
+            disable=not self.progress_bar,
+        ):
+
+            if not is_initialized:
```
MaxNoe, line 531: This can now be moved out of the event loop since #1313 is merged

ctapipe/tools/stage1.py
```diff
@@ -0,0 +1,731 @@
...
+            )
+            writer.write(
+                table_name="dl1/event/subarray/trigger",
+                containers=[event.index, event.trig],
+            )
+            # write the telescope tables
+            self._write_telescope_event(writer, event)
+
+        if is_initialized is False:
+            raise ValueError(f"No events found in file: {self.event_source.input_url}")
```
MaxNoe, line 556: Do we really want to raise on empty runs? I don't know how this is for cta mc productins, but at least in FACT we had some Proton simulation runs where no event triggered. Why would you raise on successfully processing an empty file? THat's valid information, no?

kosack: Perhaps a warning would be better. It's there becasue if you for example specify a list of telescopes [1,2,3,4], but there are no events that trigger those telescopes, there should be a good error message (I added it since I ran into that problem and was confused why I had no output).   Also rewording it to "no events found in the file for the telescope ids [xxxx]" would be clearer.

kosack: basically it's to explain to the user that they misconfigured the telescope list or something and excluded all events.

ctapipe/tools/stage1.py
kosack: This is the same pattern we use for the CameraCalibrator - the user passes in their own class.  It could be moved to the event source, of course. I think it was only because the SimTelEventSource only recently had the R1 calibration steps added.  ANyhow, let's leave that for a future refactoring, it's not really relevant here.

ctapipe/tools/stage1.py
kosack: yes, done.

ctapipe/tools/stage1.py
kosack: it was originally like that, but lots of people complained.  However, I'm happy to add it back. Then we definitely do need the tel_type field in the index container (there was a discussion to remove it)

ctapipe/tools/stage1.py
kosack: fixed

ctapipe/tools/stage1.py
MaxNoe: I don't see why we need that. This info can be joined in by telescope_id and via the subarray description.

MaxNoe: The point of the index for me is to enable such joining. It should be the minimal primary key. These files should be pretty normalized, shouldn't they?

kosack:
Ok, this still has one bug introduced by one of the past commits recently:

It seems to no longer reads the  input file! If I add `--input=events.simtel.gz`, I now get:
```
INFO [Stage1ProcessorTool] (tool/run): Starting: ctapipe-stage1-process
ERROR [Stage1ProcessorTool] (tool/run): EventSource: No input_url was specified.  Use --help for more info
```

I suspect this is due to the change from changing the list of `classes` from ` EventSource` to `classes_with_traits(EventSource)`.

kosack:
Also another bug: `self._cur_obs_id` disappeared at some point, but it's needed to write the headers

MaxNoe:
One thing: is trigger information really dl1 and not service or monitoring? This seems pretty independent of datalevels

kosack:
> One thing: is trigger information really dl1 and not service or monitoring? This seems pretty independent of datalevels

There is not service or monitoring outside of data levels.   Trigger is also clearly event data, not monitoring or service, as it comes once per event. 

It may however be an exact copy of the dl0/event/trigger data (which is not yet defined).  I have no idea if there would be some algorithms or calibrations applied, but if not it's just a copy.  It's needed to analyze Dl1 files, so must be there.

MaxNoe:
> It may however be an exact copy of the dl0/event/trigger data (which is not yet defined). I have no idea if there would be some algorithms or calibrations applied, but if not it's just a copy. It's needed to analyze Dl1 files, so must be there.

Oh, that sounds really strange to me. Yes it's event wise, but it's not related to the actual data. So I would have it expected it to be outside this tree. Why is there no monitoring / service info outside the dlX tree?

It seems that many things could better belong to an out-of-tree section than under a dlX. 

And than you could require `monitoring/trigger/event/telescope/` to be present in DL1 files and `monitoring/trigger/event/subarray` in DL2 files.

But that's probably a discussion not for this here.

kosack:
> So I would have it expected it to be outside this tree. Why is there no monitoring / service info outside the dlX tree?

Monitoring data also has data levels, as does service data. See the top-level data model document

kosack:
> And than you could require monitoring/trigger/event/telescope/ 

I think your definition of Monitoring is not the same as the CTA one.  It's data taken during an observation (or during processing) at a somewhat regular rate, say roughly 1 Hz, i.e. not connected to the events.  For example, you could have dl1/monitoring/pedestal for the pedestals you compute every few seconds.  dl1/monitoring/trigger would be stats of the triggers made ever few seconds. It might not belong in these files, since not really needed, but could be computed elsewhere.  Again what we keep calling a "dl1" file is not really correct: it contains a subset of DL1 info, but there can be other DL1 files (dl1/monitoring files for all instruments or weather for example).

kosack:
> Oh, that sounds really strange to me. Yes it's event wise, but it's not related to the actual data levels.

it definitely is - there will be processing and calibration applied to all data generated at the DL0 level to produce DL1 info, so it may not be identical to the DL0 variant.  For example, there could be timing corrections. So far this is not defined.

MaxNoe:
> It may however be an exact copy of the dl0/event/trigger data (which is not yet defined). I have no idea if there would be some algorithms or calibrations applied, but if not it's just a copy. It's needed to analyze Dl1 files, so must be there.

Ah, I see.

The thing that bugs me a bit is, that the changes from dl0 to dl1 are pretty independent here, right? A refinement of pointing position is independent of the image parameterization.

So wouldn't it be more practical to have the hierarchy as :

`category/datalevel` instead of `datalevel/category`, especially since the datalevels and the transformations in between are defined for each category seperately?

kosack:
> category/datalevel instead of datalevel/category, especially since the datalevels and the transformations in between are defined for each category seperately?

I think it makes more sense they way we have it - you would rarely if ever mix data from different data levels, but you would often mix event/monitoring/service into a single file.   Within a data level, some data is independent, but most isn't.   Most of dl1/monitoring is generated from the events or simultaneously with them  (or will be when implemented).  There are some exceptions, but that is the main use case.

ctapipe/tools/stage1.py
MaxNoe: done

MaxNoe:
@kosack I'm working on a fix for the extractor issue

ctapipe/tools/stage1.py
```diff
@@ -0,0 +1,750 @@
...
+
+        tel = self.event_source.subarray.tel[tel_id]
+        geometry = tel.camera.geometry
+        image_selected = image[signal_pixels]
+
+        # check if image can be parameterized:
+        image_criteria = self.check_image(image_selected)
+        self.log.debug(
+            "image_criteria: %s",
+            list(zip(self.check_image.criteria_names[1:], image_criteria)),
```
vuillaut, line 454: would it be interesting to write the criteria status event per event as columns in the DL1 parameters table? overkill?

vuillaut:
There is also an important discrepancy between true and reco parameters. Do you have an idea why?

image

ctapipe/tools/stage1.py
```diff
@@ -0,0 +1,750 @@
...
+            site="Other",  # need a way to detect site...
+            class_="Subarray",
+            type_="unknown",
+            version="unknown",
+            id_=subarray.name,
+        ),
+    )
+
+    # convert all values to strings, since hdf5 can't handle Times, etc.:
+    # TODO: add activity_stop_time?
```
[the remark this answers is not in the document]
MaxNoe: no, done

ctapipe/tools/stage1.py
```diff
@@ -0,0 +1,750 @@
...
+    description = __doc__ + f" This currently writes {DL1_DATA_MODEL_VERSION} DL1 data"
+    examples = """
+    To process data with all default values:
+    > ctapipe-stage1-process --input events.simtel.gz --output events.dl1.h5 --progress
+
+    Or use an external configuration file, where you can specify all options:
+    > ctapipe-stage1-process --config stage1_config.json --progress
+
+    The config file should be in JSON or python format (see traitlets docs). For an
+    example, see ctapipe/examples/stage1_config.json in the main code repo.
```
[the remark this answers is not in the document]
MaxNoe: What do you mean? Exporting the used config? That is supported by the `Tool` itself, isn't it?

ctapipe/tools/stage1.py
```diff
@@ -0,0 +1,750 @@
...
+
+        Returns
+        -------
+        np.ndarray, ImageParametersContainer:
+            cleaning mask, parameters
+        """
+
+        tel = self.event_source.subarray.tel[tel_id]
+        geometry = tel.camera.geometry
+        image_selected = image[signal_pixels]
```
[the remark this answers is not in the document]
MaxNoe: Improvement in what? I suspect this will harm g/h separation. Let's try to different analysis approaches after this as a baseline is there.

MaxNoe:
This looks like a combination of noise and cleaning. 
Looking through some of the images, I noticed that the showers are much larger at sub-cleaning-threshold values (long tails with 1-2 pe per pixel)

That fits with most parameters here:

* width /length gets smaller, the effect for width is stronger
* intensity gets smaller (more pronounced for small showers, where a larger proportion of the light is in dim  pixels)
* Concentration gets larger (more light in less pixels)
* concentration core gets smaller (because the core is smaller as per 1 and 2

We could apply the same cleaning to the true image before calculating the true parameters, but that would be problematic when using a cleaning method that relies on time, as we don't have peak time for the true image right now.

kosack:
Yes, you probably have to plot those in bins of total intensity to better understand the distributions. The true_hillas parameters are just the ones computed from the noise-free image, so at low intensity, you're dominated by noise. Since this is a power-law spectrum, the plots you make are mostly dominated by low-energy/low-intensity images.   I don't think this is a problem with the code itself.

ctapipe/tools/stage1.py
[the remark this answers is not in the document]
vuillaut: I mean having a way to dump the exhaustive config file without having to download the (non-exhaustive?) example from the repository. It's a proposal of improvement, nothing blocking of course.

ctapipe/tools/stage1.py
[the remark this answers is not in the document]
vuillaut: You are right, I think we looked only at angular and energy resolution but we should go up to sensitivity. We can go on this way.

ctapipe/tools/stage1.py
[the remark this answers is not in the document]
vuillaut: > We could apply the same cleaning to the true image before calculating the true parameters, but that would be problematic when using a cleaning method that relies on time, as we don't have peak time for the true image right now.

That would be a solution indeed but we could also argue that the signal extraction (and thus the cleaning) is part of the benchmark.

vuillaut:
> We could apply the same cleaning to the true image before calculating the true parameters, but that would be problematic when using a cleaning method that relies on time, as we don't have peak time for the true image right now.

That would be a solution indeed but we could also argue that the signal extraction (and thus the cleaning) is part of the benchmark.

vuillaut:
Ok. I will approve since I don't have any major/blocking feedback, just a couple of minor of minor ones that can (or not) be addressed here or later, as your pref.
Maybe just remove the TODO if there is nothing TODO anymore ;-)
@MaxNoe and @kosack congrats for the massive work, the DL1 looks really great

MaxNoe:
Woohoo

Issue #1166, opened 2019-10-31
GainSelection should be applied to R0 waveform, not R1


kosack:
Right now the `CameraCalibrator` applies the GainSelector (in `calibrated_dl0()`) and it applies it to the R1 waveforms (which are in "rough photoelectrons" after the R1 calibration).   This means that the threshold is calibration dependent (e.g. it's something like 7.0 photoelectrons/ns for an LST).  In reality, this should be applied in the step from R0 → R1, where the threshold is in Digital Counts and thus the threshold for switching should be (2^12-1) = 4095 for a 12-bit ADC and assuming the front-end amplifiers are linear in the whole digitization range. 

I suggest we either:
- remove the gain selector from CameraCalibrator and move it to SimTelEventSource (not ideal, since then you'd have to configure it, but that is currently where the R1 calibration happens)
- just change CameraCalibrator to apply it to the r0 waveform (but remember in the future, this would not be available for real data)

That way we don't have to "calibrate" the gain threshold for each camera each time the DC/PE or pedestals change.

HealthyPear:
This is for NectarCam, using 50 events from R0 waveforms.

![image](https://user-images.githubusercontent.com/17836610/67963812-82ba7200-fbff-11e9-89b2-80c5766a259c.png)


This is for NectarCam, using 50 events from R1 waveforms.

![image](https://user-images.githubusercontent.com/17836610/67963583-23f4f880-fbff-11e9-974a-3ada138cc4ad.png)

watsonjj:
Should the gain selection be applied after the mc pedestal subtraction (but before the mc.dc_to_pe is applied)?

https://github.com/cta-observatory/ctapipe/blob/5b81bb67358bed6bf406e9c0ed48721fd8cb3f00/ctapipe/io/simteleventsource.py#L282

vuillaut:
> Right now the `CameraCalibrator` applies the GainSelector (in `calibrated_dl0()`) and it applies it to the R1 waveforms (which are in "rough photoelectrons" after the R1 calibration). This means that the threshold is calibration dependent (e.g. it's something like 7.0 photoelectrons/ns for an LST). In reality, this should be applied in the step from R0 → R1, where the threshold is in Digital Counts and thus the threshold for switching should be (2^12-1) = 4095 for a 12-bit ADC and assuming the front-end amplifiers are linear in the whole digitization range.
> 
> I suggest we either:
> 
[...]

PR #1167, opened 2019-11-04
Move gain selection from CameraCalibrator to EventSource

watsonjj:
#1166 highlighted some issues with the current order for gain selection

The first option in #1166 (perform the gain selection in the `SimtelEventSource`) made most sense to me as:
- R1 waveforms are supposed to be single channel (as the correct gain channel is decided at the camera server, before data volume reduction is performed and the resulting DL0 waveforms are stored).
- EventSources are meant to provide the single-channel DL0 (R1 for the moment, during prototyping), however the waveforms provided by the simtel `EventSources` have both gain channels.

This is accomplished by passing a GainSelector to the `SimtelEventSource` when initialising. By default, a `SimtelEventSource` will use a `ThresholdGainSelector` (which has a default threshold of 1000).

R1 waveforms are consequently now single channel (2 dimensions: n_pixels, n_samples) as is intended in the data model.

R0 waveforms are still available from the `SimtelEventSource` and contain both gain channels.

### DIFF
--- a/ctapipe/io/hessioeventsource.py
+++ b/ctapipe/io/hessioeventsource.py
@@ -40,8 +40,7 @@ class HESSIOEventSource(EventSource):
             Passes the correct logger to the component.
             Set to None if no Tool to pass.
         gain_selector : ctapipe.calib.camera.gainselection.GainSelector
-            The GainSelector to use. If None, then ManualGainSelector will be
-            used, which by default selects the high/first gain channel.
+            The GainSelector to use. If None, then ThresholdGainSelector will be used.
         kwargs
         """
         super().__init__(config=config, parent=parent, **kwargs)
--- a/ctapipe/io/hessioeventsource.py
+++ b/ctapipe/io/hessioeventsource.py
@@ -2,6 +2,7 @@ from astropy import units as u
 from astropy.coordinates import Angle
 from astropy.time import Time
 from ctapipe.io.eventsource import EventSource
+from ctapipe.io.simteleventsource import apply_simtel_r1_calibration
 from ctapipe.io.containers import DataContainer
 from ctapipe.instrument import (
     TelescopeDescription,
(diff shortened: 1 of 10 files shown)


kosack:
I think it would also make sense to set the default threshold to 4095 (or at least something like 4000), which is closer to the correct dynamic range for all current cameras, that have 12-bit ADCs (and presumably linear gains in that range).   Even if we should really tailor each to the specific camera, the default is what most people will naively use and should be as close to correct as possible.

kosack:
Also, i think it's better to use a ThresholdGainSelector by default in the EventSource.  That way for a quick example event loop, we don't have to pass in a GainSelector, just stick with the default that gives you what you should expect.

watsonjj:
> I think it would also make sense to set the default threshold to 4095 (or at least something like 4000), which is closer to the correct dynamic range for all current cameras, that have 12-bit ADCs (and presumably linear gains in that range). Even if we should really tailor each to the specific camera, the default is what most people will naively use and should be as close to correct as possible.

Are there any linearity effects which mean you would want to switch gain way before hitting the saturation (and therefore have a threshold a bit lower than 4095)?

watsonjj:
> Also, i think it's better to use a ThresholdGainSelector by default in the EventSource. That way for a quick example event loop, we don't have to pass in a GainSelector, just stick with the default that gives you what you should expect.

This is what I do in this PR, isn't it?

kosack:
> This is what I do in this PR, isn't it?

Sorry, I misread it - yes this is good.

ctapipe/io/hessioeventsource.py
```diff
@@ -11,22 +12,38 @@
...
+        config : traitlets.loader.Config
+            Configuration specified by config file or cmdline arguments.
+            Used to set traitlet values.
+            Set to None if no configuration to pass.
+        tool : ctapipe.core.Tool
+            Tool executable that is calling this component.
+            Passes the correct logger to the component.
+            Set to None if no Tool to pass.
+        gain_selector : ctapipe.calib.camera.gainselection.GainSelector
+            The GainSelector to use. If None, then ManualGainSelector will be
```
MaxNoe: Lying docs here. The code uses `ThresholdGainSelector`

PR #1168, opened 2019-11-06
ensure command-line args have higher precedence than config options

kosack:
Just a small hack to re-load the command-line args after a config file is loaded, so that any option on the command-line overrides those in the config file (which is what a user would expect).

For example, if `"enable_feature": true` is in a config file:

old behavior:
```
some-tool --config "config.json" --enable_feature=False
```
Would result in `enable_feature=True` (instead of false, as the user might expect)

Caveat: order on the command-line doesn't matter (no way to do that with current traitlets.config implementation), so `some-tool --enable_feature=False --config="config.json"` has the same result. 

There are a lot more problems with the current config system, but at least this makes it a bit less painful.


ctapipe/core/tool.py
```diff
@@ -137,6 +137,7 @@ def initialize(self, argv=None):
             except Exception as err:
                 raise ToolConfigurationError(f"Couldn't read config file: {err}")
         self.log.info(f"ctapipe version {self.version_string}")
+        self.parse_command_line(argv)  # command-line takes precedence
```
MaxNoe: From traitlets code here:
https://github.com/ipython/traitlets/blob/master/traitlets/config/application.py

We see that the parsed command line options are stored in `self.cli_config`:
https://github.com/ipython/traitlets/blob/32ced29b0ab06914ff92e9bdcd8806536633d701/traitlets/config/application.py#L704-L705

Instead of parsing again, we could just call `self.update_config(self.cli_config)` again, that probably avoids the parsing step and still does what we want.

kosack: Just re-parsing is pretty fast, and probably avoids it breaking everything if they change the internals of traitlets.config (which is annoyingly very poorly documented). But i'm ok either way.

vuillaut: I like the `update_config` better (more self-explanatory imo) but with soft preference so good either-way.

kosack: ok changed to use update_config

ctapipe/core/tests/test_tool.py
```diff
@@ -76,5 +81,33 @@ class MyTool(Tool):
...
+            self.sub = self.add_component(SubComponent(parent=self))
+
+    config = Config(
+        {"MyTool": {"userparam": 12.0}, "SubComponent": {"component_param": 15.0}}
+    )
+
+    tool = MyTool(config=config)  # sets component_param to 15.0
+
+    tool.run(argv=["--component_param", "20.0"])
+    assert tool.sub.component_param == 20.0
```
MaxNoe: Maybe also test that a config option that was not given on the cli but as `Config` has the correct value and not the default?

kosack: that's covered by other tests already, but could add it again in case somehow the command-line breaks the config functionality. .

MaxNoe: I thought that maybe the update might reset stuff to the defaults if not given in the cli. but if that's not the case or tested elsewhere, great

kosack: I'll add it just in case. traitlets.config is so poorly documented, that who knows.

PR #1170, opened 2019-11-12
ImPACT improvements to interpolations

ParsonsRD:
This pull request is an improvement to the ImPACT interpolation classes to allow it to also interpolate efficiently between zenith and azimuth angles. This however will not break functionality when using older versions of the templates that do not have this information.

Also support is added for a new (experimental) template type created by the TemplateFitter code, which contains an uncertainty on each template.


ctapipe/reco/ImPACT.py
```diff
@@ -93,20 +95,24 @@ class ImPACTReconstructor(Reconstructor):
...
 
         # First we create a dictionary of image template interpolators
         # for each telescope type
         self.root_dir = root_dir
         self.priors = prior
         self.minimiser_name = minimiser
 
-        self.file_names = {"CHEC": ["GCT_05deg_ada.template.gz",
-                                    "GCT_05deg_time.template.gz"],
+        self.file_names = {"CHEC": ["GCT_05deg_ada_ex.template.gz",
```
kosack: it may be a good idea to make these Component configuration variables, or perhaps a configurable "pattern" like 

```py3
   template_pattern = "{cam_type}_05deg_{template_type}"
```
Where `template_type` is one of ["ada", "time", ...] 

Also what is the meaning of  "ada" and "ada_ex" ? could that be written more clearly in full words?

ctapipe/reco/ImPACT.py
```diff
@@ -161,12 +169,17 @@ def initialise_templates(self, tel_type):
 
             self.prediction[tel_type[t]] = \
```
kosack: avoid `\` for continuation lines, use parentheses. I would recommend running "black" on this file in fact.

ctapipe/reco/ImPACT.py
```diff
@@ -161,12 +169,17 @@ def initialise_templates(self, tel_type):
...
                 TemplateNetworkInterpolator(self.root_dir + "/" +
-                                            self.file_names[tel_type[t]][0])
+                                            self.file_names[tel_type[t]][0],
```
kosack: use pathlib here for filename generation.  If root_dir is a `Path`, you can just do: `rootdir / self.file_names[tel_type[t]][0]`.  That way it will not break on systems like Windows (on which surprisingly ctapipe works fine so far!)

ctapipe/utils/unstructured_interpolator.py
```diff
@@ -28,23 +28,25 @@ class UnstructuredInterpolator:
...
+            Interpolation grid points
+        values: ndarray
+            Interpolation values
         function_name: str
             Name of class member function to call in the case we are interpolating
             between class predictions, for numpy arrays leave blank
         """
-
-        self.keys = np.array(list(interpolation_points.keys()))
+        print("bounds",bounds)
```
kosack: use the Component logging system here: `self.log.debug("bounds=%s", bounds)` , not `print()`

ctapipe/reco/ImPACT.py
ParsonsRD: Yes, I fully agree, this is a holdover from when the camera names kept changing. I will try to make this all a bit clearer.

ctapipe/utils/unstructured_interpolator.py
ParsonsRD: This print statement was accidentally left in anyway. I'll just remove it.

ctapipe/reco/ImPACT.py
ParsonsRD: Done.

ctapipe/reco/ImPACT.py
MaxNoe: How do you know? Related to https://github.com/cta-observatory/pyeventio/issues/109#issuecomment-492659009? :rofl:

kosack: My office mate is running it on windows! I was pretty surprised that it all worked (even pyeventio), if installed via pip.

kosack: You also need to register any files you open with the Provenance system (if not using `get_dataset_path`, which does it automatically).  To do that, just do:
```py3
from ctapipe.core import Provenance
PROV = Provenance()
...
PROV.add_input_file(filename, role="ImPACT Template file")
```

ctapipe/reco/ImPACT.py
MaxNoe: This could now be a TelescopeParameter

PR #1174, opened 2019-11-14
Test with 3.8

MaxNoe:


kosack:
I wonder if we still need to maintain the env files for 3.6/3.7/3.8? or can we just go back to a single one?

MaxNoe:
I think the reason we need this hasn't change. Prevent conda from chaning the python version in the update step.

watsonjj:
FYI a new Numba version has been released to fully support 3.8
https://numba.pydata.org/numba-doc/latest/release-notes.html#version-0-47-0-jan-2-2020

MaxNoe:
I just triggered a rebuilt. On the pure pip install it failed due to the error in the setup of ctapipe-extra and on conda it failed as not all packages were yet available for 3.8.

We need to fix the situation with ctapipe-extra, either we put the simtel version of the nectar cam back, or we switch to the download system or touch a lot of tests. 

I would opt for quickly adding the simtel NectarCam back and release a new version of ctapipe-extra fast, then look into the download stuff.

kosack:
> We need to fix the situation with ctapipe-extra, either we put the simtel version of the nectar cam back, or we switch to the download system or touch a lot of tests.

can you explain more about what the problem is? Or open an issue on ctapipe-extra?

MaxNoe:
> can you explain more about what the problem is? Or open an issue on ctapipe-extra?

A recent PR removed the `NectarCam.camgeom.fits.gz` which was the simtel version and added the real world config for the NectarCam. 
This file was used extensively in the ctapipe unit tests.

kosack:
> A recent PR removed the NectarCam.camgeom.fits.gz which was the simtel version and added the real 
> world config for the NectarCam.
> This file is used extensively in the ctapipe unit tests.

Ah, yes it would be good to have it - or just change the name of which camera is being used in the tests (to LST or whatever NectarCam has been renamed to).  I think none of the tests rely on the camera definition/pixel-ordering being the same as the MC one, just that a camera definition exists.

MaxNoe:
Ok, builds now. 3.8 with conda does not work due to still not all packages being available on conda for 3.8 or version conflicts.

As we have only 5 parallel builds anyways, using 6 configs would have doubled the build time.

watsonjj:
Would there be some way of continually testing for the 3.8 conda version, to see when it becomes available, without it being required in the PR checks or impacting the "build passing" badge? Perhaps a separate travis test file, and a unique badge? Or perhaps another PR like this, and we see when it manages to pass.

watsonjj:
> As we have only 5 parallel builds anyways, using 6 configs would have doubled the build time.

How will we approach this in the future when the 3.8 conda version is available?

MaxNoe:
At the current pace of conda publishing 3.8 modules, at that point we can probably drop python 3.6 ;)

MaxNoe:
@kosack merge?

PR #1176, opened 2019-11-18
Use descriptor protocol to mark fields as deprecated

MaxNoe:
Found the solution to the descriptor protocol problem watching this talk: https://www.youtube.com/watch?v=UANN2Eu6ZnM&feature=youtu.be

### DIFF
--- a/ctapipe/core/__init__.py
+++ b/ctapipe/core/__init__.py
@@ -4,7 +4,7 @@ Core functionality of ctapipe
 """
 
 from .component import Component, non_abstract_children
-from .container import Container, Field, DeprecatedField, Map
+from .container import Container, Field, Deprecated, Map
 from .provenance import Provenance, get_module_version
 from .tool import Tool, ToolConfigurationError
 
@@ -13,7 +13,7 @@ __all__ = [
     'Container',
     'Tool',
     'Field',
-    'DeprecatedField',
+    'Deprecated',
     'Map',
     'Provenance',
     'ToolConfigurationError',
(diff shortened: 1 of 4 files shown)


kosack:
cool! that sounds like just what we need.

kosack:
Looks like there is still a reference to `DeprecatedField` in `__init__.py`

MaxNoe:
There is a remaining issue here, the default is not set.

dneise:
right and the test did not expect the warning.

MaxNoe:
Also issuing a custom warning would be nice `class DeprecatedFieldWarning(Warning)` (Do not inherit from `DeprecationWarning` as these are filtered by default.

dneise:
Ah I get it .. the default is not set, since the Field is already initialized, when it is handed over to Deprecated ...

dneise:
This is too deep for me, I've got no idea how to fix this.

MaxNoe:
The defaults are set in the `__init__` of container where it loops over all the `fields`. But `fields` now only contain the non-deprecated fields.

dneise:
just saying, I feel unable to do this ...

MaxNoe:
To be included in the introduction of typed fields in the next version

watsonjj:
Related PR/issue?

MaxNoe:
Non yet. This was just a refactoring I had planned.

MaxNoe:
Getting rid of the deprecations is probably better than improving how we mark things as deprecated.

Issue #1165, opened 2019-10-30
Discussion on restructuring Event Container structure


kosack:
Since it's come up a few times that things in the current Event container structure are a bit haphazard, let's start a discussion on what should move.  For reference, here is what it is now:

- `event_type`  *Event type*
 - `index`    *event indexing information*
    - `event_id`  *event identifier*
    - `obs_id`  *observation identifier*
 - `r0`    *Raw Data*
    - `obs_id`  *observation ID*
    - `event_id`  *event id number*
    - `tels_with_data`  *list of telescopes with data*
    - `tel[X]`  *map of tel_id to R0CameraContainer*
       - `trigger_time`  *Telescope trigger time, start of waveform readout,*
       - `trigger_type`  *camera's event trigger type if applicable*
       - `num_trig_pix`  *Number of trigger groups (sectors) listed*
       - `trig_pix_id`  *pixels involved in the camera trigger*
       - `waveform`  *numpy array containing ADC samples(n_channels, n_p*
 - `r1`    *R1 Calibrated Data*
    - `obs_id`  *observation ID*
    - `event_id`  *event id number*
    - `tels_with_data`  *list of telescopes with data*
    - `tel[X]`  *map of tel_id to R1CameraContainer*
       - `trigger_time`  *Telescope trigger time, start of waveform readout*
       - `trigger_type`  *camera trigger type*
       - `waveform`  *numpy array containing a set of images, one per AD*
       - `selected_gain_channel`  *Numpy array containing the gain channel chosen for*
 - `dl0`    *DL0 Data Volume Reduced Data*
    - `obs_id`  *observation ID*
    - `event_id`  *event id number*
    - `tels_with_data`  *list of telescopes with data*
    - `tel[X]`  *map of tel_id to DL0CameraContainer*
       - `trigger_time`  *Telescope trigger time, start of waveform readout*
       - `trigger_type`  *camera trigger type*
       - `waveform`  *numpy array containing data volume reduced p.e. sa*
 - `dl1`    *DL1 Calibrated image*
    - `tel[X]`  *map of tel_id to DL1CameraContainer*
       - `image`  *Numpy array of camera image, after waveform extrac*
       - `pulse_time`  *Numpy array containing position of the pulse as de*
 - `dl2`    *Reconstructed Shower Information*
    - `shower[X]`  *Map of algorithm name to shower info*
       - `alt`  *reconstructed altitude [deg]*
[...]

PR #1179, opened 2019-11-21
Include ImageParameterContainer in DataContainer

nbiederbeck:
ImageParameters like HillasParameters are *event-wise per-telescope* needed for reconstruction algorithms and thus need to be accessible in a DataContainer

```python
event = DataContainer()
event.imageparameters[0].hillas  # etc.
```

### DIFF
--- a/ctapipe/io/containers.py
+++ b/ctapipe/io/containers.py
@@ -142,23 +142,23 @@ class DL1CameraCalibrationContainer(Container):
     pedestal_offset = Field(
         0,
         "Additive coefficients for the pedestal calibration of extracted charge "
-        "for each pixel"
+        "for each pixel",
     )
     absolute_factor = Field(
         1,
         "Multiplicative coefficients for the absolute calibration of extracted charge into "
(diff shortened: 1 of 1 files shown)


watsonjj:
The imageparameters should be considered DL1 level. There was some discussion about how to include them in #722, where a DL1b container was suggested. But this discussion was a while ago. @kosack have you made some decision on the container location for hillas parameters in the DL1Writer tool developements?

kosack:
I'm going to include this in the next update of the DL1 writer - I'm still working out a good scheme to store this in the event data structure, so I'd prefer to wait until that is done.    The ctapipe-stage1-process (DL1 writer) tool will be the reference implementation and also define the internal data structure

kosack:
(and yes it should be at least in `event.dl1.tel[x]` somwhere (I'll probably divide that into images (dl1a) and parameters (dl1b) as discussed in the ASWG F2F meeting)

kosack:
This discussion should move to: #1165

PR #1191, opened 2019-11-29
Cam display with degrees

dneise:
I played a little bit around, in order to get a CameraDisplay in the TelescopeFrame.

test code: (mainly from lst_analysis_bootcamp_2018)
```python
from ctapipe.calib import CameraCalibrator
from ctapipe.coordinates import TelescopeFrame
from ctapipe.io import event_source
from ctapipe.utils.datasets import get_dataset_path
%matplotlib inline


input_url = get_dataset_path('gamma_test_large.simtel.gz')
source = event_source(input_url, max_events=49)

for event in source:
    pass

calibrator = CameraCalibrator(subarray=source.subarray)
calibrator(event)
tel_id = 4

camera = event.inst.subarray.tel[tel_id].camera
dl1 = event.dl1.tel[tel_id]


fig, (a1, a2) = plt.subplots(1, 2, figsize=(12, 8))

d1 = CameraDisplay(camera, ax=a1)
d1.image = dl1.image
# d1.add_colorbar()
d1.axes.grid()

# -----------------------------------------
# this is the part which is "new"
# -----------------------------------------
camera.equivalent_focal_length = event.inst.subarray.tel[tel_id].optics.equivalent_focal_length
d2 = CameraDisplay(camera.transform_to(TelescopeFrame()), ax=a2)
d2.image = dl1.image
# d2.add_colorbar()
d2.axes.grid()
```

As you can see, I allowed the `CameraGeometry` to store the focal_length as well, in order to be able to use it in its `transform_to()` method. So far so good ... but the in the result the Pixel angles are messed up, and I have no idea where this comes from.

image


ctapipe/instrument/camera.py
```diff
@@ -202,7 +246,7 @@ def _calc_pixel_area(pix_x, pix_y, pix_type):
         Note this will not work on cameras with varying pixel sizes.
         """
 
-        dist = np.min(np.sqrt((pix_x - pix_x[0])**2 + (pix_y - pix_y[0])**2))
+        dist = np.min(np.sqrt((pix_x[1:] - pix_x[0])**2 + (pix_y[1:] - pix_y[0])**2))
```
dneise: Ah btw. I think this was a bug ... I have no idea why this worked ... I think this line tries to find the distance between the pixels, simply by calculating the distance from pixel zero to all other pixels and then finding the minimum.

The result should be the typical pixel pitch, unless pixel zero happens to sit in a special location outside of the normal grid. However, if one does not exclude pixel zero from the others .. the result should always be zero, since pixel zeros distance from itself is zero, and that should be the minimum, no? 

So I have no idea how this line worked ...

MaxNoe:
@kosack I have fixes for these in the two PRs in your fork.

One thing: is trigger information really dl1 and not service or monitoring? This seems pretty independent of datalevels

HealthyPear:
Hello!

I need to do something very similar in _protopipe_, even though in my case it is focused on the parametrization, rather than in the visualization (which is of course needed).

From what I can see it is MUCH easier than [what I tried sometime ago](https://indico.cta-observatory.org/event/2656/contributions/23534/attachments/17259/22828/test_camera_shrinking.ipynb).
There, I was applying the coordinate transformations to the pixels coordinates on a copy of the camera geometry, which was quite slow and messy.

I didn't know there was a `transform_to` method inside `CameraGeometry` itself, and what you did, by just adding the focal length as an attribute to it, seems to do the job quite faster (even though I didn't tested it yet on parameter distributions).

The only additional thing that we should do (but this is out of scope for this PR) is to extend the input capabilities of `ctapipe.reco.HillasReconstructor` (currently it expects always parameters from the `CameraFrame`).
In general, I think we should always work in the `TelescopeFrame` and go to the `HorizonFrame` with the direction reconstruction.

Should we re-focus our attention of this PR?
I would like to reproduce (temporarely) this in _protopipe_ to get some parameter distributions in degrees to compare against _CTA-MARS_.
I see that there are some conflicts which I guess are only due to the fact that in the meanwhile _ctapipe_ got upgraded.

MaxNoe:
> Should we re-focus our attention of this PR?

No, what you are talking about and what is done here are completely different, yet related things.

I agree with you, that the HillasReconstructor should start from TelescopeFrame (and that probably also means calculating Hillas Parameters in the telescope frame.

This is however a large, long-standing issue and not something this PR should cover.

kosack:
> I agree with you, that the HillasReconstructor should start from TelescopeFrame (and that probably also means calculating Hillas Parameters in the telescope frame.

Just to note,  that is already being done in protopipe, and the code should move soon to the ctapipe version once tested.

kosack:
closed, replaced with #1407

PR #1199, opened 2019-12-04
Add Hough fit

dneise:
@momorning and me add the hough fit here, based on the implementation by Laurits Tani from here:
https://github.com/Laurits7/circlehough

This PR is supposed to be looked at after #1154  is merged, since the diff will then appear to be very nice.


dneise:
having to fix the installation of Laurits [circlehough](https://github.com/Laurits7/circlehough/releases) dependency  ... but gotta run.

MaxNoe:
It's a very small package developed for a master's thesis basically providing a single function.

I don't say this lightly, but I think I would opt for just including the code in ctapipe

dneise:
> I don't say this lightly, but I think I would opt for just including the code in ctapipe

I agree with Max here.

dneise:
How is our policy regarding Cython? 
circlehough provides both, the `.pyx` file so the user can generate and compile the sources locally as well as the pregenerated `.c` file so the user does not need to have cython installed.

dneise:
We could also ask Laurits how he feels about letting cta-obs maintain that package for him?

kosack:
It would be a bit annoying to add cython code now that we removed it elsewhere (it makes packaging so much easier if we don't have to compile it on all systems), so I guess a separate package is best. 

Is there a reason not the use the Hough transform in SciKit-Image?

kosack:
Also looking at his code, it would probably work fine as a Numba function (would need to be re-written to remove the cython stuff though).

dneise:
I think they never compared it to the hough fit in scikit.image. What they did is, they used CircleModel from skimage and wrapped it into RanSac ... and this performed worse than their own implementation.

The reason was basically limited resolution IIRC, normally in a hough trafo, a point either contributes to a bin in the accumulator or it does not. They introduced a triangular weight, so that points further away from a ring do contribute but not so much. This improved their accuracy .. especially the accuracy for R a lot.

Screenshot from 2019-12-11 16-42-38

"ringM" is the RanSac fit to the skimage Ringmodel, "hough" is their own implementation. This was done on toy MCs IIRC not on Corsika MCs.

kosack:
I still have the problem that I cannot build a ctapipe conda package if this is included (unless I also maintain a package for `circlehough`, which I don't really want to do). 

I forsee 3 options:
- convince the circlehough maintainer to add it to conda-forge, so there is a package with proper dependency tracking
- add the full source code to ctapipe (as a sub-package, not as a separate package, so that the ctapipe package it self contains it)
- making it optional (so in a default conda install of ctapipe, it would not be available, but if the user manually installed it with pip, the functionality would work).

maxnoe:
I am closing this PR, since it is very old, has conflicts and relies on an external dependency we are not keen to include.

If you think this feature is still worth pursuing, please feel free to re-open (or open a new pull request) fixing the conflicts and addressing the review comments.

The best way forward would probably be to implement the hough trafo in ctapipe itself using numba.

PR #1201, opened 2019-12-06
configurable image cleaners

kosack:
This PR creates some configurable `ImageCleaners`  that are just `Component` wrappers for the existing algorithms (similar to what was in #1163 and intended to replace those.   The parameters of each can be configured differently for each telescope in the array (or by type), since they use `TelescopeParameters` traits.

 It also creates a `TelescopeComponent` base class to help with attaching a SubarrayDescription to `Components` that need them. 

The basic usage inside a Tool is:
```py3
clean = self.add_component(
    ImageCleaner.from_name(cleaning_method, subarray=subarray, parent=self)
)

...

clean_mask = clean(tel_id, image, arrival_times)
```

### DIFF
--- a/ctapipe/core/__init__.py
+++ b/ctapipe/core/__init__.py
@@ -3,20 +3,21 @@
 Core functionality of ctapipe
 """
 
-from .component import Component, non_abstract_children
+from .component import Component, TelescopeComponent, non_abstract_children
 from .container import Container, Field, DeprecatedField, Map
 from .provenance import Provenance, get_module_version
 from .tool import Tool, ToolConfigurationError
 
 __all__ = [
-    'Component',
-    'Container',
-    'Tool',
-    'Field',
-    'DeprecatedField',
-    'Map',
-    'Provenance',
-    'ToolConfigurationError',
-    'non_abstract_children',
-    'get_module_version',
+    "Component",
+    "TelescopeComponent",
+    "Container",
+    "Tool",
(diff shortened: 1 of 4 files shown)


dneise:
Could we add one more test, which ensures that the `ImageCleaner` really always needs a subarray, like this?
```python
@pytest.mark.parametrize(
    "method", ["TailcutsImageCleaner", "MARSImageCleaner", "FACTImageCleaner"]
)
def test_image_cleaner_needs_subarray(method):
    with pytest.raises(TypeError):
        ImageCleaner.from_name(method)
```

Or is that actually more of a test for `TelescopeComponent` .. at the moment `TelescopeComponent` seems not to be tested at all.

dneise:
c.f. #1203 I though it would be nice to test all subclasses automatically, without having to list them all manually.

kosack:
> c.f. #1203 I though it would be nice to test all subclasses automatically, without having to list them all manually.

Of course, we still need to add some configs for each subclass, so it doesn't save too much time, but I do like that it prevents accidentally missing one. 

> Could we add one more test, which ensures that the ImageCleaner really always needs a subarray, like this?

Yes, I'll add some more. I'm definitely worried about the case where the user forgets the subarray and expects the config file to be followed, but in reality it is using defaults. Though I   guess that should be tested for in the `TelescopeComponent` tests (which I just notice I forgot)

PR #1206, opened 2019-12-13
Add guessing MAGIC telescope

adonini:
Fix to read MAGICCam from MC files. For the issue look at cta-observatory/protopipe#36. Solution to this is adding the missing MAGIC infos in `guess.py`.

### DIFF
--- a/ctapipe/instrument/guess.py
+++ b/ctapipe/instrument/guess.py
@@ -24,6 +24,8 @@ TELESCOPE_NAMES = {
     GuessingKey(1855, 28.0): GuessingResult("LST", "LST", "LSTCam", 1),
     GuessingKey(11328, 5.59): GuessingResult("MST", "SCT", "SCTCam", 1),
     # Non-CTA Telescopes
+    GuessingKey(1039, 16.97): GuessingResult("LST", "MAGIC", "MAGICCam", 1),
+    GuessingKey(1039, 0.0): GuessingResult('LST', 'MAGIC', 'MAGICCamMars', 1),
     GuessingKey(960, 15.0): GuessingResult("MST", "HESS-I", "HESS-I", 1),
     GuessingKey(2048, 36.0): GuessingResult("LST", "HESS-II", "HESS-II", 1),
     GuessingKey(1440, 4.998): GuessingResult("SST", "FACT", "FACT", 1),
--- a/ctapipe/instrument/guess.py
+++ b/ctapipe/instrument/guess.py
@@ -25,7 +25,6 @@ TELESCOPE_NAMES = {
     GuessingKey(11328, 5.59): GuessingResult("MST", "SCT", "SCTCam", 1),
     # Non-CTA Telescopes
     GuessingKey(1039, 16.97): GuessingResult("LST", "MAGIC", "MAGICCam", 1),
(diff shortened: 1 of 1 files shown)


ctapipe/instrument/guess.py
```diff
@@ -24,6 +24,8 @@
     GuessingKey(1855, 28.0): GuessingResult("LST", "LST", "LSTCam", 1),
     GuessingKey(11328, 5.59): GuessingResult("MST", "SCT", "SCTCam", 1),
     # Non-CTA Telescopes
+    GuessingKey(1039, 16.97): GuessingResult("LST", "MAGIC", "MAGICCam", 1),
+    GuessingKey(1039, 0.0): GuessingResult('LST', 'MAGIC', 'MAGICCamMars', 1),
```
MaxNoe: Could you explain why a focal length of 0 should return another Camera instance? And my impression is, that `MAGICCamMars` has the pixel coordinates in the `EngineeringCameraFrame` aka MAGIC camera coordinates which is not yet supported as input for camera geometries and thus coordinate transformations will be wrong.

adonini: Yeah sorry my mistake putting a 0 focal lenght and not the same as the other cam, anyway the right camera detected from the MC is MAGICCam, so I can just eliminate the support to the `MAGICCamMars` if you say it's not needed. I just saw it was implemented too and then I added it.

MaxNoe: There cannot be duplicate keys. The combination of pixels / focal_length have to be unique, otherwise one will not be reachable.

MaxNoe: If that should arise, we would need a third field in guessing key so telescopes with the same pixel number / focal length could be differentiated. However, where do you need the guessing?

MaxNoe: EventSources nowadays shoud directly produce instrument info, so now guessing is necessary.

adonini: ok so better include the `MAGICCam` only that is the "real" one from what I understand. I was trying to analyze MC that have both LST and MAGIC telescopes, and I got `UNKNOWN` camera for MAGIC. Looking at the code the issue was with `event_source` that guess the telescope.

adonini: I tried to run the code again with the addd of the line `GuessingKey(1039, 16.97): GuessingResult("LST", "MAGIC", "MAGICCam", 1)` and it was working. No `UNKNOWN` camera anymore

kosack: Should you remove MAGICCamMARS then? Or is it somehow a different camear?

MaxNoe: No, it is just in the wrong coordinate frame.

kosack: so it should be removed anyhow - there should only be one "MagicCam". The guess method just helps define the name of a camera, not how it constructed, which is up to the EventSource implementation to do correctly and in the correct `Frame`.

adonini:
So maybe it’s better to remove MagicCamMars to avoid any confusion?

ctapipe/instrument/guess.py
adonini: So it’s better to remove MagicCamMars to avoid any confusion?

MaxNoe: yes

PR #1207, opened 2019-12-16
Add a Selector class for keeping track of quality cuts

kosack:
This replaces the older `utils.CutFlow` with a  new Component class that can be configured from a config file.  It is a class that can apply a series of quality criteria to an object, and keep track of totals (e.g. for cuts that happen per-event, before data are written to an output table) - it doesn't reject events, rather it just keeps track of which criteria pass.   The totals can also be written to an output file, as an `astropy.Table`.

It's used like so:
```py3

criteria = [
    ('criteria_1',  'lambda x: (x>2.0) and (x < 100)') ,
    ('another_check', 'lambda x: x < 3'), 
]
select = QualityQuery(quality_criteria=criteria) # can also be configured from a config file

# assuming the input is always the same type, we just call select(input) for each item:
for ii in range(100):
    result = query(ii) # result is an array of bools for each criterion
    if all(result == True):
         # do stuff

print(select.to_table()) # shows the total results as a table, with both individual and cumulative product sums.
```

Table output (from the unit test)
```
In [11]: query.to_table(functions=True)
Out[11]:
<Table length=4>
      criteria       counts cumulative_counts        func
       str20         int64        int64             str17
-------------------- ------ ----------------- -----------------
               TOTAL      4                 4    lambda x: True
         high_enough      3                 3   lambda x: x > 3
a_value_not_too_high      3                 2 lambda x: x < 100
            smallish      2                 1  lambda x: x < 10
```

You can also for example store the results in an output table with HDF5TableWriter, as a column of flags, for each event. 
Note for example, you could pass it a Container or something if you wanted and have criteria like: 

```
[
    ('good_min_width': 'lambda hillas: hillas.width > 0.2'),
    ('good_min_length': 'lambda hillas: hillas.length >1.0)'
]
```

An example can be found in #1163 , where it's used to select good images, configured as follows in a JSON file:. 
```json
  "ImageQualityQuery": {
        "quality_criteria": [
            ["enough_pixels", "lambda im: np.count_nonzero(im) > 2"],
            ["enough_charge", "lambda im: im.sum() > 100"]
        ]
    }
```

## Caveats:

- ~~The name 'Selector' may be confusing since there already exists a `GainSelector` that is a different thing altogether (suggestions for a better name, or keep this one? I had originally called it 'Checker')~~ → Renamed to QualityQuery
- ~~Currently the critieria are stored as a `dict`, but there is a strong assumption on order (for the cumulative product), so perhaps this is not a good idea if the data will be written to e.g.  a YAML file in the future, where dicts don't preserve order (in pyyaml at least).  Perhaps a list of tuples would be better? or a list of dicts?~~

### DIFF
--- a/ctapipe/core/selector.py
+++ b/ctapipe/core/selector.py
@@ -6,10 +6,18 @@ __all__ = ["Selector", "SelectionFunctionError"]
 
 from collections.abc import Callable
 
-import numpy as np
+import astropy.units as u  # for use in selection functions
+import numpy as np  # for use in selection functions
 
 from .component import Component
 from .traits import Dict
+
+# the following are what are allowed to be used
+# in selection functions (passed to eval())
+ALLOWED_GLOBALS = {
+    "u": u,  # astropy units
+    "np": np,  # numpy
+}
 
 
 class SelectionFunctionError(TypeError):
@@ -28,7 +36,8 @@ class Selector(Component):
     selection_functions = Dict(
         help=(
             "dict of '<cut name>' : lambda function in string format to accept ("
-            "select) a given data value.  E.g. `{'mycut': 'lambda x: x > 3'}` "
+            "select) a given data value.  E.g. `{'mycut': 'lambda x: x > 3'}`. You may "
+            "use `numpy` as `np` and `astropy.units` as `u`, but no other modules."
         )
     ).tag(config=True)
 
@@ -43,7 +52,8 @@ class Selector(Component):
 
         try:  # generate real functions from the selection function strings
             self._selectors = {
-                name: eval(func_str) for name, func_str in selection_functions.items()
+                name: eval(func_str, ALLOWED_GLOBALS)
+                for name, func_str in selection_functions.items()
             }
         except NameError as err:
             raise SelectionFunctionError(
--- a/ctapipe/core/selector.py
+++ b/ctapipe/core/selector.py
@@ -11,8 +11,10 @@ import numpy as np
 from .component import Component
 from .traits import Dict
 
+
 class SelectionFunctionError(TypeError):
     pass
+
 
 class Selector(Component):
     """
@@ -39,18 +41,20 @@ class Selector(Component):
 
         self.selection_functions = selection_functions  # update
 
-        try:# generate real functions from the selection function strings
+        try:  # generate real functions from the selection function strings
             self._selectors = {
                 name: eval(func_str) for name, func_str in selection_functions.items()
             }
         except NameError as err:
-            raise SelectionFunctionError("Couldn't evaluate one of the selection "
-                                         "function strings")
+            raise SelectionFunctionError(
+                f"Couldn't evaluate one of the selection function strings: {err}"
+            )
 
         for name, func in self._selectors.items():
             if not isinstance(func, Callable):
-                raise SelectionFunctionError(f"Selection criterion '{name}' is not a function")
-
+                raise SelectionFunctionError(
+                    f"Selection criterion '{name}' is not a function"
+                )
 
         # arrays for recording overall statistics
         self._counts = np.zeros(len(self._selectors), dtype=np.int)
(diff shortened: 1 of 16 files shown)


ctapipe/core/selector.py
```diff
@@ -2,11 +2,17 @@
 Data Quality selection
 """
 
-__all__ = ["Selector"]
+__all__ = ["Selector", "SelectionFunctionError"]
+
+from collections import Callable
```
MaxNoe: `from collections.abc`, importing directly from collections raises a warning in 3.7 and an error in 3.8

kosack: done

kosack:
to mitigate security issue of using eval here, see https://github.com/cta-observatory/ctapipe/pull/1163#issuecomment-597003077

ctapipe/utils/CutFlow.py
```diff
@@ -25,6 +26,11 @@ def __init__(self, name="CutFlow"):
         """
         self.cuts = OrderedDict()
         self.name = name
+        warn(
```
MaxNoe: I think we talked about deprecation cycles in ctapipe somewhere and that we would do them until the first 1.0 release.

kosack: cutflow is currently used heavily in Protopipe, so I will keep it around just that code is migrated. I guess that will be quick

MaxNoe:
@kosack the cutflow tests fail now

ctapipe/core/selector.py
```diff
@@ -0,0 +1,126 @@
...
+
+        # add a selection to count all entries and make it the first one
+        selection_functions = {"TOTAL": "lambda x: True"}
+        selection_functions.update(self.selection_functions)
+
+        self.selection_functions = selection_functions  # update
+
+        try:  # generate real functions from the selection function strings
+            self._selectors = {
+                name: eval(func_str) for name, func_str in selection_functions.items()
```
MaxNoe: add `eval(func_str, {}, {'np': np, 'u': u}) here, to make only numpy and astropy units available for func_str here.

MaxNoe: `import astropy.units as u` of course.

And a test showing that it fails with something like `subprocess.run(['echo', "about to erase 5 years of mc production"])` fails.

kosack:
I've now made the security changes requested, and also switched to using a list of tuples  of (criteria_description, selection_function) for configuration.  

One further simplification could be to remove the criteria_description, so you just have the function string, and no more detail.  Any preference whether or not we keep the descriptive name or not?

watsonjj:
> The name 'Selector' may be confusing since there already exists a GainSelector that is a different thing altogether (suggestions for a better name, or keep this one? I had originally called it 'Checker')

This reminds me of [pandas.DataFrame.query](https://pandas.pydata.org/pandas-docs/stable/reference/api/pandas.DataFrame.query.html). So how about the class being called `Query`? It does not seem like this class is doing any selecting itself...it is instead just querying the data.

kosack:
> This reminds me of pandas.DataFrame.query. So how about the class being called Query? It does not seem like this class is doing any selecting itself...it is instead just querying the data.

That's not a bad idea - yes, it does not drop anything, that's still up to the user, it just returns an array of which criteria passed and keeps a running total. Originally I think I called it `DataChecker`, but didn't really like that term (but maybe it was ok).  Query may be too generic a term since there will be other queries, potentially like for databases.  Perhaps `QualityQuery`?  The purpose is to evaluate a set of criteria  on the same object (event, images, etc). 

 It's only useful for event-wise analysis, of course, anything operating on all events at once like a `DataFrame` would just use `frame.query()` as you suggest.   

 In a standard analysis there would be one of these for:
- checking if images in an event should be parameterized (just to avoid errors, or to reduce DL1 data size, but perhaps it's not a good idea to make such cuts and store all events?) 
- checking if telescopes in an event should be used in reconstruction
- checking if an event should be reconstructed at all

If we switch to "chunked" analysis, where we write out DL1 tables and reconstruct them whole, then this is better done another way. So it's just a stop-gap until then and could eventually be dropped.

PR #1209, opened 2019-12-18
added bad pixel info field to DL1CameraContainer

moritzhuetten:
We are using this additional field for MAGIC data processing to store (multidimensional) bad pixel information

### DIFF
--- a/ctapipe/io/containers.py
+++ b/ctapipe/io/containers.py
@@ -126,6 +126,10 @@ class DL1CameraContainer(Container):
         "the extractor."
         "Shape: (n_pixel, n_samples)",
     )
+    badpixels = Field(
+        None, 
+        "Numpy (nd-)array of bools indicating (a specific type of) bad pixels"
+    )
 
 
 class DL1Container(Container):


moritzhuetten:
As @kosack commented for #1210, also this info might be better not saved event-wise, but more seldom (once a run/file, for example).

FrancaCassol:
Hi,
this can be added to the PixelStatusContainer, which is filled with monitoring data and not necessarily updated at each event. For the moment we use several status arrays, one for each type, but this evolve if other solutions are better


```
class MonitoringCameraContainer(Container):
    """
    Container for camera monitoring data
    """

    flatfield = Field(FlatFieldContainer(), "Data from flat-field event distributions")
    pedestal = Field(PedestalContainer(), "Data from pedestal event distributions")
    pixel_status = Field(
        PixelStatusContainer(), "Container for masks with pixel status"
    )
    calibration = Field(
        WaveformCalibrationContainer(), "Container for calibration coefficients"
    )
```

and
```
class PixelStatusContainer(Container):
    """
    Container for pixel status information
    It contains masks obtained by several data analysis steps
    At r0/r1 level only the hardware_mask is initialized
    """

    hardware_failing_pixels = Field(
        None,
        "Boolean np array (True = failing pixel) from the hardware pixel status data ("
        "n_chan, n_pix)",
    )

    pedestal_failing_pixels = Field(
        None,
        "Boolean np array (True = failing pixel) from the pedestal data analysis ("
        "n_chan, n_pix)",
    )

    flatfield_failing_pixels = Field(
        None,
        "Boolean np array (True = failing pixel) from the flat-field data analysis ("
        "n_chan, n_pix)",
    )

```

moritzhuetten:
Hi, nice, I didn't know that a corresponding container structure also existed, so I guess we can close this pull request. However, related to the simultaneous #1210, it is not yet clear to me how to give a time stamp to the monitored quantity. (Or, it would be good to have a timestamp possibility also here, when failing pixel information is updated also within a run, which is the case in my use case).

watsonjj:
As this seems to already exist in `PixelStatusContainer` I will close this

PR #1210, opened 2019-12-18
added WeatherContainer for event-based outside atmospheric conditions

moritzhuetten:
We use this approach in processing MAGIC data to read outside air temperature, pressure, and humidity for each event, to later properly compute the (ra,dec) coordinates with Astropy according to

``
AltAzframe = SkyCoord(az=event.pointing[1].azimuth, alt=event.pointing[1].altitude, 
                 obstime=event.trig.gps_time,
                 location=EarthLocation(...), frame='altaz', pressure=event.weather.air_pressure, 
                 temperature=event.weather.air_temperature,
                 relative_humidity=event.weather.air_humidity
                 )
``

### DIFF
--- a/ctapipe/io/containers.py
+++ b/ctapipe/io/containers.py
@@ -51,6 +51,7 @@ __all__ = [
     "TelEventIndexContainer",
     "ImageParametersContainer",
     "SimulatedShowerDistribution",
+    "WeatherContainer",
 ]
 
 
@@ -127,6 +128,11 @@ class DL1CameraContainer(Container):
         "Shape: (n_pixel, n_samples)",
     )
 
+class WeatherContainer(Container):
+    """Storage of event-wise weather information (MAGIC implementation)."""
+    air_temperature = Field(None, "Outside air temperature")
+    air_pressure = Field(None, "Outside air pressure")
+    air_humidity = Field(None, "Outside air humidity")
 
 class DL1Container(Container):
     """ DL1 Calibrated Camera Images and associated data"""
@@ -540,6 +546,8 @@ class DataContainer(Container):
         EventCalibrationContainer(),
         "Container for calibration coefficients for the current event"
     )
+    weather = Field(Map(WeatherContainer),
+                    'Outside weather conditions at the observatory')
 
 
 class SST1MDataContainer(DataContainer):


kosack:
Surely these shouldn't be per-event quantities, right? Rather Monitoring quantities, measured once per second at most.  

This is a good time to define a general MonitoringContainer that contains at least a single timestamp (or a start_time, end_time range?) , so that we can manage all Monitoring quantities in the same way.

ctapipe/io/containers.py
```diff
@@ -127,6 +128,11 @@ class DL1CameraContainer(Container):
         "Shape: (n_pixel, n_samples)",
     )
 
+class WeatherContainer(Container):
+    """Storage of event-wise weather information (MAGIC implementation)."""
+    air_temperature = Field(None, "Outside air temperature")
+    air_pressure = Field(None, "Outside air pressure")
+    air_humidity = Field(None, "Outside air humidity")
```
kosack: need a `unit=x` field in these

ctapipe/io/containers.py
```diff
@@ -540,6 +546,8 @@ class DataContainer(Container):
         EventCalibrationContainer(),
         "Container for calibration coefficients for the current event"
     )
+    weather = Field(Map(WeatherContainer),
```
kosack: why is this a map?

kosack: I think just a `Field(WeatherConainer(), "weather information")` would be fine

moritzhuetten: Oh, I see, the Map() is thought to be used for repeatedly (for several telescopes) used sub-containers, which isn't necessary here. So no reason to use "Map", I assume.

moritzhuetten:
> Surely these shouldn't be per-event quantities, right? Rather Monitoring quantities, measured once per second at most.
> 
> This is a good time to define a general MonitoringContainer that contains at least a single timestamp (or a start_time, end_time range?) , so that we can manage all Monitoring quantities in the same way.

Sure, as there (seemingly) hasn't been such a feature, so far, we have been saving the info event-wise, but of course, such a _Monitoring quantities_ feature, with some method for interpolating to the time of an event, would be even better. Do you have a specific design in mind?

kosack:
> Do you have a specific design in mind?

I guess just having a simple common base class that defines the time columns would be enough for now, than any monitoring Container would derive from it and only define the quantities it needs.  That way we have a common way to store a monitoring file (or to read one).  Might want to check that there is isn't already a standard in ACADA, but i'm sure it can't be too complex.  

something like:
```py3
class MonitoringContainer(Container):
    time = Field(None, "Measurement time, as a TAI timestamp in unix format", unit=u.s)

class WeatherMonitoringContainer(MonitoringContainer):
    atmospheric_pressure = Field(np.nan *u.bar, "atmospheric pressure", unit=u.bar)
    ...
```

Or we could store it as an astropy `Time` object (but that's currently not supported by the `io.TableWriter/TableReader`, though it could be added pretty easily).

moritzhuetten:
Sure. Actually, there already is a `MonitoringContainer(Container)` class in `containers.py` (but without time field), do mean that? And do I understand right, that then all fields have an extra dimension in time? But what to do if different monitored quantities are taken at different times? Yes, in my case (MAGIC) the monitoring quantities are recorded for a fixed moment in time/timestamp, not interval. (See also @FrancaCassol's comment to my simultaneous #1209)

GernotMaier:
Not an expert on pointing, so maybe this is a stupid question:  does giving pressure and temperature to astropy mean that it applies correction for atmospheric refraction to the altaz - to - radec conversion?

If yes, how are the differences between air showers and star taken into account?
- air showers are light sources at ~10 km distances, stars at infinity
- air showers are mostly blue, while star are more redish

moritzhuetten:
> Not an expert on pointing, so maybe this is a stupid question: does giving pressure and temperature to astropy mean that it applies correction for atmospheric refraction to the altaz - to - radec conversion?
> 
> If yes, how are the differences between air showers and star taken into account?
> 
>     * air showers are light sources at ~10 km distances, stars at infinity
> 
>     * air showers are mostly blue, while star are more redish

Yes, this data is meant to be used (below others) for a precise altaz - to - radec conversion from starguider positions. Indeed, as far as I know, in current IACT analyses/Cherenkov light simulations this effect is not taken into account for the Cherenkov light and may add to the systematics of the primary particle angular reconstruction. W.r.t. starlight, the effect is usually tiny (less than one arcmin for zenith angles smaller than 60 degrees), but may become large at very large zenith angles. We also discussed this a while ago shortly with @moralejo .

moritzhuetten:
We close this PR for now and work on a a more efficient implementation using some MonitoringContainer as suggested by @kosack . Thanks for the suggestions so far!

moritzhuetten:
> > Not an expert on pointing, so maybe this is a stupid question: does giving pressure and temperature to astropy mean that it applies correction for atmospheric refraction to the altaz - to - radec conversion?
> > If yes, how are the differences between air showers and star taken into account?
> > ```
> > * air showers are light sources at ~10 km distances, stars at infinity
> > 
> > * air showers are mostly blue, while star are more redish
> > ```
> 
> Yes, this data is meant to be used (below others) for a precise altaz - to - radec conversion from starguider positions. Indeed, as far as I know, in current IACT analyses/Cherenkov light simulations this effect is not taken into account for the Cherenkov light and may add to the systematics of the primary particle angular reconstruction. W.r.t. starlight, the effect is usually tiny (less than one arcmin for zenith angles smaller than 60 degrees), but may become large at very large zenith angles. We also discussed this a while ago shortly with @moralejo .

UPDATE: In Corsika, Cherenkov light refraction is in fact taken into account with the `raybnd()` function.

MaxNoe:
@moritzhuetten Under which conditions?

IACT atmo with CERWLEN?

PR #1212, opened 2020-01-15
Fixes for astropy 4.x

kosack:
The master build is currently broken since the update to astropy-4

in astropy 4, and the latest NumPy, numpy functions *retain* units correctly
whereas before they were stripped off. Functions that relied on them disappearing
then break.

In `geometry_converter_hex`: 
* fixed instances of unit problems
* Also first fixed, and then, removed the `@jit` (originally it did not ever compile in no_python mode), however now when compiled with `@njit`, it seems it introduces some rounding error that breaks the hex converter (probably means the algorithm is unstable, but that's for another day!)

### DIFF
--- a/ctapipe/image/geometry_converter_hex.py
+++ b/ctapipe/image/geometry_converter_hex.py
@@ -9,14 +9,10 @@ from ctapipe.instrument import CameraGeometry
 
 logger = logging.getLogger(__name__)
 
-__all__ = [
-    "convert_geometry_hex1d_to_rect2d",
-    "convert_geometry_rect2d_back_to_hexe1d"
-]
+__all__ = ["convert_geometry_hex1d_to_rect2d", "convert_geometry_rect2d_back_to_hexe1d"]
 
 
-def unskew_hex_pixel_grid(pix_x, pix_y, cam_angle=0 * u.deg,
-                          base_angle=60 * u.deg):
+def unskew_hex_pixel_grid(pix_x, pix_y, cam_angle=0 * u.deg, base_angle=60 * u.deg):
     r"""transform the pixel coordinates of a hexagonal image into an
     orthogonal image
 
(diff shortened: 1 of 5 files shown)


ctapipe/image/geometry_converter_hex.py
```diff
@@ -240,7 +238,6 @@ def reskew_hex_pixel_from_orthogonal_edges(x_edges, y_edges, square_mask):
     return unrot_x, unrot_y
 
 
-@jit
 def get_orthogonal_grid_edges(pix_x, pix_y, scale_aspect=True):
```
kosack: This function is now fully jit-able with `@njit`, but when enabled, the number of bins at the end is off by 1, which seems to be to be due to a slight difference in rounding error somewhere (didn't have time to debug). For now, I've disabled the jit, but clearly this is something that will pop up at some point.

ctapipe/visualization/mpl_array.py
```diff
@@ -149,7 +149,7 @@ def set_vector_uv(self, u, v, c=None, **kwargs):
         if self._quiver is None:
             coords = self.tel_coords
             self._quiver = self.axes.quiver(
-                coords.x, coords.y,
```
MaxNoe: Better use a specific unit here? `to_value(u.m)`?

Or is these really unit agnostiv

kosack: It should be agnostic, but probably isn't. I've chanted it to meters, since that's assumed elsewhere (e.g. in "length").  The problem seems to have been the UV coordinates, though, so i'm fixing that now. They are in m, but quiver doesn't work with unit quantities.

kosack: It was driving me crazy debugging this, until I just realized that the variable "u" is overwriting the package `import astropy.units as u`...

kosack: Fixed it now, and now everywhere assumes or converts to meters.

MaxNoe: > It was driving me crazy debugging this, until I just realized that the variable "u" is overwriting the package import astropy.units as u.

F**K

PR #1213, opened 2020-01-17
bump astropy version requirement to allow v4.0

kosack:
just changes setup.py to allow astropy 3.x or 4.x series.  Right now, if you run "python setup.py develop" after installing astropy 4.x, it gets downgraded to 3.x without this.

### DIFF
--- a/ci/install.sh
+++ b/ci/install.sh
@@ -27,5 +27,6 @@ if [[ "$CONDA" == "true" ]]; then
 	travis_wait 20 conda env update -n cta-dev --file py${PYTHON_VERSION}_env.yaml
 	conda activate cta-dev
 else
-	pip install -e .[all]
+	pip install -U pip
+	pip install -U -e .[all]
 fi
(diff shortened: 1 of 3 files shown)


kosack:
Well, I thought this would be simple, but it seems #1212 didn't fix all astropy 4 problems (only those with the new NumPy - in fact it was downgrading to astropy 3). 

Now there are some more unit problems in the Muon code...

kosack:
strangely, on my machine, with astropy 4.0 and numpy 0.17.2, I cannot reproduce the errors seen in Travis.

MaxNoe:
> angle_ring = np.linspace(0, 2 * math.pi, 360.)

This is just an error, the 360 must be a plain integer

MaxNoe:
I think we found a pip bug here, opened https://github.com/pypa/pip/issues/7609

kosack:
that explains why the pip versions failed (as astropy 4 got installed), whereas the conda ones didn't orignally

MaxNoe:
Actually a known limitation of pips dependency resolution...

PR #1215, opened 2020-01-28
2-Pass waveform integration à-la-CTA/MARS

### LINKED ISSUE #602: Implement 2-pass trace integration
We should have an implementation of the 2-pass trace-integration (using timing info) that is described e.g. in *Holder et al, 2006*.  The result is a higher signal-to-noise image, and the ability to apply lower cleaning thresholds in subsequent steps. 

The basic procedure is:

1. use a wide (20ns) trace integration
2. apply a 1st stage image cleaning (tailcuts)
3. calculate the time gradient of the resulting pixels
4. use the time-gradient to predict the position of the pulses to better accuracy 
5. re-integrate with a 10ns window using the information from step 4

This is not so far different from the `NeighborPeakIntegrator` with some extra time information, and would be useful for comparing our results with MARS and EventDisplay.

HealthyPear:
Fixes #602 

**Prerequisites:**

- [x] Merge PR #1131 - continuous integration will fail until this is done.

**Description:**

- _ctapipe_-based version of the calibration method used in _CTA/MARS_ and described also [here](https://forge.in2p3.fr/projects/step-by-step-reference-mars-analysis/wiki#Single-pixel-spectrum).
- triggered from development of _protopipe_ and comparison against _CTA/MARS_ (in particular DL1, see 
cta-observatory/protopipe#31 and cta-observatory/protopipe#37 for more information)

**Caveats:**

1. There is at least one cut inside this image extractor, between the 1st and 2nd pass (n_pixels > 3); in _ctapipe_ we shouldn't do this, aka any image extractor should always return charge and pulse times; what I am testing in _protopipe_ and I propose to do also here, is to add a 3rd return argument (at this point to any image extractor) named `status,` which acts like a "quality label", so later the user can select events on the basis of image extraction quality for more complex image extractors that allow internal selection (like this one),
2. This triggers another problem: in the current protopipe/CTA-MARS what is compared and enters in the benchmark of the charge resolution is the 1st pass, whereas what enters in the final image cleaning is the 2nd pass; this image extractor should then export both passes, but right now in ctapipe only 1 charge and 1 pulse time per image is exported by the call to the calibration method
3. To be as much as close to CTA/MARS as possible, the `ctapipe.image.timing_parameters` should be modified in order to use a robust fitting algorithm (like `scipy.stats.siegelslopes`), though I couldn't find an easily importable solution which treated also the fit errors.


ctapipe/image/extractor.py
```diff
@@ -209,15 +214,38 @@ def subtract_baseline(waveforms, baseline_start, baseline_end):
...
+    window_width : int
+        Size of the smoothing kernel.
+
+    Returns
+    -------
+    sum : array_like
+        Array containing the sums for each of the kernel positions.
+        Shape: (n_samples - window_width - 1)
+
+    Dr. Michele Peresano, 2019
```
kosack: I'd suggest removing your name here, and instead adding it to the `CODEOWERS` file at the top level (we decided against putting names in the comments).  You also should also update `AUTHORS` and `.mailmap` with your information

kosack: However, if there is a reference for the MARS analysis, it would be nice to add it also to `docs/references.rst` and add a citation link.  Note that it should not be in the `Returns` section (sphinx won't like that), but in either the description or `See Also` section (see https://numpydoc.readthedocs.io/)

HealthyPear: I'll try to see what could be the right citation

ctapipe/image/extractor.py
```diff
@@ -386,15 +409,239 @@ class BaselineSubtractedNeighborPeakWindowSum(NeighborPeakWindowSum):
...
+        # WARNING: these values should be read from a configuration file
+        # à-la-protopipe and they will depend in principle also on camera type.
+        subarray = self.subarray
+        if subarray.tel[telid].type == "LST":
+            core_th = 6  # (not yet optimized)
+        if subarray.tel[telid].type == "MST":
+            core_th = 8  # (not yet optimized)
+        if subarray.tel[telid].type == "SST":
+            core_th = 4  # (not yet optimized)
+
```
kosack: This should just be a `TelescopeParameter` in the class, so that it can be configured from the config file. 
 
at the top of the class, and you can even set these values as the default:
```py3
core_threshold = FloatTelescopeParameter(
    default_value=[('type','*',6.0), ('type','LST*',6.0), ('type','MST*',8.0), ('type','SST*',4.0)],
    help="Picture threshold for internal tail-cuts pass"
)
```
Then here you can just use:
 ```py3
core_th = self.core_threshold.tel[telid]
```

and it will use the correct threshold

Comment (HealthyPear) [reply ctapipe/image/extractor.py]: Thanks, done!

File: ctapipe/image/extractor.py (hunk 1 above)
Comment (GernotMaier) [reply ctapipe/image/extractor.py]: If I understood it correctly is this the two-pass integration method described in section 4 of https://ui.adsabs.harvard.edu/abs/2006APh....25..391H/abstract ? 

If this is correct, then it maybe shouldn't be called CTA/mars analysis :-)

Comment (HealthyPear) [reply ctapipe/image/extractor.py]: Hi Gernot, I apologize for the late reply.

I don't know if the version of the algorithm I coded here comes directly from the reference that you provide (in any case, from what you say it seems a necessary reference to be added to the documentation so I will do it).
I just tried with as much fidelity as possible to translate in _ctapipe_-language the version that is used inside CTA/MARS.
That is the reason for the name of the PR (if this is what you refer to).
I also explicited in the class documentation that `"[...] This is in particular the CTA-MARS version."`.

For me there is no problem in changing names, perhaps we can ask @moralejo for a confirmation on this.

Comment (moralejo) [reply ctapipe/image/extractor.py]: Hi, the Eventdisplay reference is indeed to be included:  I implemented this in MARS for the CTA analysis (it is not used in MAGIC) to get closer to Eventdisplay. Still, in the spirit of having fully independent analysis codes, I just took the idea, "make a time fit with significant pixels, then do a second round of pulse integration on the rest of the pixels using the times 'predicted' by the fit". The implementation was done in a completely independent manner, and it is bound to differ in the details. I do not think it makes much sense to copy every detail of the Mars version for protopipe - as you @HealthyPear realized, some things (like ROOT's "robust" fit details) would take quite some effort and are probably not essential.

Comment (HealthyPear) [reply ctapipe/image/extractor.py]: > Hi, the Eventdisplay reference is indeed to be included: I implemented this in MARS for the CTA analysis (it is not used in MAGIC) to get closer to Eventdisplay. Still, in the spirit of having fully independent analysis codes, I just took the idea, "make a time fit with significant pixels, then do a second round of pulse integration on the rest of the pixels using the times 'predicted' by the fit". The implementation was done in a completely independent manner, and it is bound to differ in the details. 

I will then add the Eventdisplay one; do you have any suggestion for the reference on the CTA/MARS side? Otherwise as I stated above I specified that this is the version used there.

Comment (HealthyPear) [reply ctapipe/image/extractor.py]: > I'd suggest removing your name here, 

Done

> and instead adding it to the `CODEOWERS` file at the top level (we decided against putting names in the comments). 

Done - I hope is the correct syntax

> You also should also update `AUTHORS` and `.mailmap` with your information

I have already done this sometime ago, do I have to change or update something?

Comment (moralejo) [reply ctapipe/image/extractor.py]: > 
> I will then add the Eventdisplay one; do you have any suggestion for the reference on the CTA/MARS side? Meanwhile, I took the liberty to add the URL of the wiki page as a citation.

Unfortunately we have no public reference for that, since as I said it is not a method we use in the MAGIC analysis. Wiki page is ok for internal reference.

Comment (kosack) [reply ctapipe/image/extractor.py]: possibly the ICRC proceeding by Maxim Shayduk - I think that's the origin of the method.

File: ctapipe/image/extractor.py [hunk 3]
```diff
@@ -10,21 +10,26 @@
...
 from abc import abstractmethod
 from functools import lru_cache
 import numpy as np
 from traitlets import Int
-from ctapipe.core.traits import IntTelescopeParameter
+from ctapipe.core.traits import IntTelescopeParameter, FloatTelescopeParameter
 from ctapipe.core import Component
 from numba import njit, prange, guvectorize, float64, float32, int64
 
+from ctapipe.image.cleaning import number_of_islands, largest_island, tailcuts_clean
```
MaxNoe, line 29: use relative imports here.

MaxNoe:
@HealthyPear #1267 is merged, if you merge or rebase vs master, this should work now.

HealthyPear:
> It seems no unit test ever runs the part of the code that start in "step 5" in the comments (i.e. when there are more than 3 pixels). So that means you need to either add a new unit test, or modify the existing one to test that scenario.
> 
> This is the code in the _else_ clause of the block:
> 
> ```python
> if (np.count_nonzero(image_2) < 3) or len(nonCore_pixels_ids) == 0:
> ...
> else:
>   #  [this is so far not tested]
> ```

Ah yes...this happens because the simulated waveform is too bright (this is also why I had to add  ` or len(nonCore_pixels_ids) == 0`).

I will modify the unit-test to make one dim enough to trigger that part

HealthyPear:
Meanwhile:

does anyone have any comment about the **caveats** I pinpointed in the description?
In particular points **1)** and **2)** - if nothing changes in that regard, in order to make things work for the current benchmarks I will have to keep the calibration part of the code in _protopipe_ a bit tweaked.

kosack:
For the caveats:
* adding a status return value is ok, though then all ImageExtractors need one to be consistent (so @watsonjj should comment)
* The plotting of the first pass is only needed for some very specialized benchmark plots, so i think you could just implement a user option to disable the second pass (returning only the first), and then you can just enable that option when you want to make that particular plot, and turn it off otherwise. 
* For the robust fit, I agree, and I think that's a general problem for the timing fit - can you open an issue on that specifically? At least it should use siegelslopes, but maybe use another method to estimate the error.

MaxNoe:
> adding a status return value is ok, though then all ImageExtractors need one to be consistent (so @watsonjj should comment)

I disagree. We should not start with return codes. Use exceptions.

kosack:
> I disagree. We should not start with return codes. Use exceptions.

It's not an exception. This is really status info - it tells which pass of the algorithm was used for the final integration.  In this method, there are 2 passes, and if the second cannot be run due to not enough information to use the timing, it returns the first pass result.  It's useful to be able to keep track of that.  THe final science results are still the same: you get an image each time, but in some images, it couldn't refine the integration windows. 

The alternative is just to throw away this info, which is also ok, but might be harder to debug or benchmark.

watsonjj:
> 2. This triggers another problem: in the current protopipe/CTA-MARS what is compared and enters in the benchmark of the charge resolution is the 1st pass, whereas what enters in the final image cleaning is the 2nd pass; this image extractor should then export both passes, but right now in ctapipe only 1 charge and 1 pulse time per image is exported by the call to the calibration method

I suggest splitting __call__ into two member functions, which execute each pass respectively. Then __call__ should call these functions. This way in your code which compares protopipe and CTA-MARS, you can call the first pass function directly. I don't think we should change the __call__ to return multiple charge and time arrays.

watsonjj:
> To be more clear, right now __call__() returns two images: the intensity and peak time. @HealthyPear is suggesting to just have it return (intensity, peak_time, status), where status has some extra info about how the algorithm ran on a per-image basis. If we implement #1243 , ImageExtractors will also return other images (pulse_width, rise_time, fall_time, etc), and could also require more status info.

This is not a bad idea. Though I would be more explicit and call the variable `quality`, not `status`.

watsonjj:
> 3. To be as much as close to CTA/MARS as possible, the ctapipe.image.timing_parameters should be modified in order to use a robust fitting algorithm (like scipy.stats.siegelslopes), though I couldn't find an easily importable solution which treated also the fit errors.

Open an issue for this so it can be discussed properly. Without knowing more about the CTA-MARS approach, or the flaws in the current linear regression approach, I would be reluctant to use something more complicated.

watsonjj:
> > To be more clear, right now call() returns two images: the intensity and peak time. @HealthyPear is suggesting to just have it return (intensity, peak_time, status), where status has some extra info about how the algorithm ran on a per-image basis. If we implement #1243 , ImageExtractors will also return other images (pulse_width, rise_time, fall_time, etc), and could also require more status info.
> 
> This is not a bad idea. Though I would be more explicit and call the variable quality, not status.

Perhaps lets add this in a second PR, as it changes ImageExtractors on the whole

ctapipe/image/extractor.py
```diff
@@ -227,8 +235,31 @@ def integration_correction(
...
+        Array containing the sums for each of the kernel positions.
+        Shape: max (n_samples - (window_width - 1))
+
+    """
+    sums = np.convolve(waveform, np.ones(width, dtype=int), "valid")
+    return sums
+
+
 class ImageExtractor(Component):
-    def __init__(self, subarray, config=None, parent=None, **kwargs):
```
watsonjj, line 231: Remove this change, subarray should be a required parameter. I presume this was an accident in merging the recent PRs

HealthyPear:
ok lots of discussion here! Thanks for you input

ctapipe/image/tests/test_extractor.py
```diff
@@ -44,15 +45,36 @@ def subarray():
...
+    waveform_model = WaveformModel.from_camera_readout(readout)
+    waveform = waveform_model.get_waveform(charge, time, n_samples)
+
+    selected_gain_channel = np.zeros(charge.size, dtype=np.int)
+
+    return waveform, subarray, telid, selected_gain_channel, charge, time
+
+
 @pytest.fixture(scope="module")
 def toymodel(subarray):
```
watsonjj, line 79: call `get_test_toymodel` in `toymodel` to remove duplicated code

watsonjj:
There are some codacy complaints you should fix

HealthyPear:
> I think if you fix a few Codacy complaints + implement @watsonjj's suggestion to split the `__call__()` function into two sub-methods, it will be ready.
> 
> For **call**, something like:
> 
> ```python
> def __call__(self, waveforms, telid, selected_gain_channel):
>      charge1, pulse_time1 = self.apply_pass_1(waveforms, telid, selected_gain_channel)
>      self.apply_pass_2(waveforms, telid, selected_gain_channel, charge1, pulse_time1)
> ```
> 
> That way, you can just call apply_pass_1 for doing the benchmarks.

Yes, it looks like a neat solution....

Also in between the two, there is the preliminary image cleaning from which the actual return values are obtained by checking the result.

PR #1218, opened 2020-02-05
Fix TelescopeParameter config,  bug #1216

### LINKED ISSUE #1216: Tool.get_current_config() is broken for TelescopeContainers
This used to work before the refactoring, so probably isn't too hard to fix.  All TelescopeParameters just return "nan" when queried right now, so you get something like the following when you call get_current_config(), rather than the correct values:

```json
        "TailcutsImageCleaner": {
            "boundary_threshold_pe": null,
            "min_picture_neighbors": null,
            "picture_threshold_pe": null
         },
```

kosack:
This PR will attempt to fix #1216.   Starting with adding a unit test to check for it, and then a solution.

### DIFF
--- a/ctapipe/core/traits.py
+++ b/ctapipe/core/traits.py
@@ -159,7 +159,11 @@ class TelescopePatternList(UserList):
         else:
             raise RuntimeError("No TelescopeParameterLookup was registered")
 
-    def attach_subarray(self, subarray):
+    def attach_subarray(self, subarray: "ctapipe.instrument.SubarrayDescription"):
+        """
+        Register a SubarrayDescription so that the user-specified values can be
+        looked up by tel_id. This must be done before using the `.tel[x]` property
+        """
         self._subarray = subarray
         self._lookup.attach_subarray(subarray)
 
--- a/ctapipe/core/traits.py
+++ b/ctapipe/core/traits.py
@@ -149,7 +149,7 @@ class TelescopeParameterLookup:
         telecope_parameter_list : list
             List of tuples in the form `[(command, argument, value), ...]`
         """
-        self._telecope_parameter_list = telecope_parameter_list
+        self._telescope_parameter_list = telecope_parameter_list
         self._value_for_tel_id = None
         self._subarray = None
         self._subarray_global_value = None
@@ -170,7 +170,7 @@ class TelescopeParameterLookup:
         """
         self._subarray = subarray
         self._value_for_tel_id = {}
-        for command, arg, value in self._telecope_parameter_list:
+        for command, arg, value in self._telescope_parameter_list:
             if command == "type":
                 matched_tel_types = [
                     str(t) for t in subarray.telescope_types
@@ -216,13 +216,13 @@ class TelescopeParameterLookup:
             )
 
     def append(self, *args, **kwargs):
-        self._telecope_parameter_list.append(*args, **kwargs)
+        self._telescope_parameter_list.append(*args, **kwargs)
 
     def __iter__(self, *args, **kwargs):
-        return self._telecope_parameter_list.__iter__(*args, **kwargs)
+        return self._telescope_parameter_list.__iter__(*args, **kwargs)
 
     def __len__(self, *args, **kwargs):
-        return self._telecope_parameter_list.__len__(*args, **kwargs)
+        return self._telescope_parameter_list.__len__(*args, **kwargs)
 
 
 class TelescopeParameter(List):
--- a/ctapipe/core/traits.py
+++ b/ctapipe/core/traits.py
@@ -1,6 +1,7 @@
 import os
 from fnmatch import fnmatch
 from typing import Optional
+from collections import UserList
 from traitlets import (
     Bool,
     CaselessStrEnum,
@@ -42,7 +43,6 @@ __all__ = [
     "enum_trait",
     "classes_with_traits",
     "has_traits",
-    "TelescopeParameterLookup",
     "TelescopeParameter",
     "FloatTelescopeParameter",
     "IntTelescopeParameter",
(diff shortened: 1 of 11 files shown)


kosack:
The following "works", but is really a hack (and probably breaks some other functionality of traitlets, like the ability to write a default config file).  The Hack is just to test for the value being a TelescopeParameter and treating it specially in `Component.get_current_config()`.  It has to be done by string comparison, since importing TelescopeParameter there causes an import loop.   

In Component:
```py3
    def get_current_config(self):
        """ return the current configuration as a dict (e.g. the values
        of all traits, even if they were not set during configuration)
        """
        config = {}
        for key, value in self.traits(config=True).items():
            if type(value).__name__.endswith("TelescopeParameter"):
                config[key] = value.get(self)._telescope_parameter_list
            else:
                config[key] = value.get(self)

        return {self.__class__.__name__: config}
```

I don't think I like this solution, so doing it properly by keeping `TelescopeParameter` as something that just stores a list of tuples (where it's internal "klass" is just list), and separating the TelescopeParameterLookup object entirely is probably best (syncing them using `observe()` perhaps as mentioned in #1216), but that will take some thought

kosack:
The solution was to 
- introduced a new class TelescopePatternList, which is what a TelescopeParameter trait now stored (it's "klass")
- This new class is just a list (UserList), but also contains a reference to a TelescopeParameterLookup (which is basically unchanged from before,)
- to access the original pattern list one does: `component.tel_param`
- to access the "lookup table" value per telescope, one now does: `component.tel_param.get(tel_id`) This is an API change from before, but unavoidable (other than overloading __getitem__, which is then confusing since a list should still work as a list)
- added a handler for this in the json dumping part of the Provenence system

This *introduces a small API change* (Though this feature was still new and unstable): you now need to use `parameter.get(tel_id)` to get the lookup value, as `parameter[2]` would give you the second item in the list of patterns (which is expected behavior, I think)

kosack:
The tests should hopefully work once #1219 is merged (unrelated bug due to latest pandas)

MaxNoe:
Is the name `get` required by traitlets or could we give it a more speaking name like `get_for_tel_id`?

kosack:
> Is the name get required by traitlets or could we give it a more speaking name like get_for_tel_id?

No, the class is a fancy list, not a traitlet, so there is no special naming convention required. I just wanted to avoid the past implementation that used __getitem__, which is confusing if a list doesn't do what you expect.   We could use another name (just chose `get`, since in the original implementation it was `value_for_telescope()` but people found that wordy).  I do prefer something more clear, like `telparam.for_tel_id(6)`

In the current implementation, you can also do `telparam.lookup[tel_id']` (accessing the underlying TelescopeParameterLookup directly), so perhaps we can just get rid of the function if this is clearer?

MaxNoe:
I like `for_tel_id`, or renaming `lookup` to `tel` -> `param.tel[tel_id]`, error could be raised by making it a property:
```
@property
def tel(self):
    if self._lookup is not None:
         return ..
    raise ValueError('No subarray attached')
````

**Comment by kosack** [comment]
I like the tel[tel_id] way best I think.  I'll try that.

File: ctapipe/core/traits.py [hunk 1]
```diff
@@ -138,22 +139,51 @@ def has_traits(cls, ignore=("config", "parent")):
...
+class TelescopePatternList(UserList):
+    """
+    Representation for a list of telescope pattern tuples. This is a helper class
+    used  by the Trait TelescopeParameter as its value type
+    """
+
+    def __init__(self, *args):
+        super().__init__(*args)
+        self._lookup = None
+        self._old_getitem = self.__getitem__
```
MaxNoe: What is this for? `__getitem__` is not overriden.

ctapipe/core/traits.py
```diff
@@ -138,22 +139,51 @@ def has_traits(cls, ignore=("config", "parent")):
...
+        self._subarray = None
```
MaxNoe: Why is the subarray stored here and in lookup?

ctapipe/core/traits.py
```diff
@@ -270,38 +297,40 @@ def __init__(self, dtype=float, default_value=None, **kwargs):
...
+                    raise TraitError(f"Value should be a {self._dtype}")
+                if not isinstance(command, str):
+                    raise TraitError("command must be a string")
+                if command not in ["type", "id"]:
+                    raise TraitError("command must be one of: '*', 'type', 'id'")
+                if command == "type":
+                    if not isinstance(arg, str):
+                        raise TraitError("'type' argument should be a string")
+                if command == "id":
+                    arg = int(arg)
```
MaxNoe: All other raise `TraitError`, but this will raise `ValueError` if the id arg is not convertible to integer

ctapipe/core/traits.py
```diff
@@ -138,22 +139,51 @@ def has_traits(cls, ignore=("config", "parent")):
...
         tuples for a telid.
 
         Parameters
         ----------
-        telecope_parameter_list : list
+        telescope_parameter_list : list
             List of tuples in the form `[(command, argument, value), ...]`
         """
-        self._telecope_parameter_list = telecope_parameter_list
+        self._telescope_parameter_list = telescope_parameter_list
```
MaxNoe: `TelescopePatternList` holds a reference to its `TelescopeParameterLookup` and the other way around, so these objects will never be garbage collected.

Give the Lookup a `weakref` or avoid the mutual reference?

ctapipe/core/traits.py
kosack: forgot to remove that when I did the last change - I'll remove it (the old implementation overrode `__getitem__`)

ctapipe/core/traits.py
kosack: That was for allowing the subarray to be re-applied automatically on change, but that works without it now. Will remove.

ctapipe/core/traits.py
```diff
@@ -270,38 +297,40 @@ def __init__(self, dtype=float, default_value=None, **kwargs):
...
-
-        # Convert to TelescopeParameterLookup
-        normalized_value = TelescopeParameterLookup(normalized_value)
-
-        # Validate with super method
-        super().validate(obj, normalized_value)
+        normalized_value = TelescopePatternList(None)
+        if isinstance(value, self._dtype):
+            value = [("type", "*", value)]
+        if isinstance(value, UserList) or isinstance(value, list):
```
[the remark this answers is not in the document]
kosack: done

ctapipe/core/traits.py
kosack: This is perhaps leftover from the original implementation - not sure we need to keep the telescope_parameter_list... let me check.  Otherwise I'll use a WeakRef

kosack: seems it's only really used to get the global values in the case that `attach_subarray()` has not been run, but it would require quite some refactoring to make that work without it, so I think I'll just make it a weakref and be done (hopefully this will all be replaced with a better config system soon anyhow)

kosack: I decided just to do a deepcopy() of the TelescopeList inside TelescopeLookup. That should work, I think.

kosack: seems that what was making the "retain subarray" between changes functionality work, which is somewhat complex. I'll see if I can fix it, but I think it will require some large refactoring from what was implemented before. Perhaps that's not a necessary feature?  @watsonjj - you added that feature - is that really important?

watsonjj: I don't remember if anything depends on this behaviour. It just seemed unintuitive that the subarray is forgotten when you change the values of the traitlet.

watsonjj: > so these objects will never be garbage collected.

Is that really true? From what I can find, they won't be freed immediately, as they will never reach a reference count of zero. But the garbage collector itself will periodically free circularly-referenced objects.

MaxNoe: Ah, yes. In fact, the whole point of the gc is circular references, as everything else is handled by ref counting.

kosack: I'm not sure that the functionality broke because of the circular ref.  I think it's just because now if any change is made, a new TelescopeList is generated by validate(), and thus the link to the TelescopeParametersLookup is lost.  Strangely, it did work after I first refactored, so I must have changed something recently. Anyhow, I'll see if I can get it working again.

kosack:
I think this works now.  The style warnings from Codacy are all things that can be ignored

PR #1220, opened 2020-02-06
Correction for optical aberrations

HealthyPear:
**Description:**

Triggered by the comparison between _protopipe_ and _CTA-MARS_ (see cta-observatory/protopipe#37), this pull request implements,

- a function `barrel_distortion` the apply the homonym distortion effect (points move towards the center of the reference frame proportionally to their distance to it),
- a function `apply_coma_correction` to apply the homonym correction via such distortion,
- a modification of the coordinate transformations between `CameraFrame` and `TelescopeFrame` to include such correction.
- new unit-tests `test_barrel_distortion` and `test_apply_coma_correction`
- a modification to `test_roundtrip_camera_horizon` to incorporate this correction in the process.

**Caveats:**

- This is a DRAFT pull request because (I have mis-clicked the button) I am experiencing a `UnitConversionError` in `test_roundtrip_camera_horizon`
- Only pixel coordinates are affected, while their surface areas remain unchanged. This is not needed for the moment, since this quantity in not used in the image parametrization process, but it could be relevant for events with few pixels above the cleaning thresholds (low-energy events everywhere in the camera or high-energy events strongly clipped).

### DIFF
--- a/ctapipe/coordinates/camera_frame.py
+++ b/ctapipe/coordinates/camera_frame.py
@@ -19,35 +19,98 @@ from .telescope_frame import TelescopeFrame
 from .representation import PlanarRepresentation
 
 
+def barrel_distortion(x, y, alpha=0):
+    """Apply a barrel distortion to a set of coordinates.
+
+    Parameters
+    ----------
+    x : array_like
+        Cartesian coordinates along the 'x' axis.
+        Shape: (n_pixels)
+    y : array_like
+        Cartesian coordinates along the 'y' axis.
+        Shape: (n_pixels)
+    alpha : float
+        Correction factor for which negative values mean shrinking.
+
+
+    Returns
+    -------
+    xp : array_like
+        Cartesian coordinates along the 'x' axis corrected for barrel distortion.
+        Shape: (n_pixels)
+    yp : array_like
+        Cartesian coordinates along the 'y' axis corrected for barrel distortion.
+        Shape: (n_pixels)
+
+    M.Peresano, K.Kosack, 2020
+    """
+    r = np.hypot(x, y).value  # pixel-wise radii
+    xp = x / (1.0 - alpha * r)
+    yp = y / (1.0 - alpha * r)
+    return xp, yp
+
+
+def apply_coma_correction(pix_x, pix_y, alpha=0):
+    """Apply correction for Coma aberration to a set of pixel coordinates.
+
+    This aberration shifts the focused light further away from the center of
+    the camera. The pixels must be moved back before parametrizing to image.
+    The radial shift is proportional to the distance from the center.
+
+    Parameters
+    ----------
+    x : array_like
+        Cartesian coordinates of the pixels along the 'x' axis.
+        Shape: (n_pixels)
+    y : array_like
+        Cartesian coordinates of the pixels along the 'y' axis.
+        Shape: (n_pixels)
+    alpha : float
+        Correction factor equivalent to (1-f) where f is the ratio between
+        effective and nominal focal length of the telescope
+        (https://www.mpi-hd.mpg.de/hfm/CTA/MC/Prod3/Config/PSF/flen.pdf)
+        In this way alpha > 0 means positive distortion.
+
+    Returns
+    -------
+    x_coma : array_like
+        Cartesian coordinates of the pixels along the 'x' axis corrected for
+        Coma aberration.
+        Shape: (n_pixels)
+    y_coma : array_like
+        Cartesian coordinates of the pixels along the 'y' axis corrected for
+        Coma aberration.
+        Shape: (n_pixels)
+
+    M.Peresano, K.Kosack, 2020
+    """
+    # Take the original pixel-coordinates and correct them
+    x_coma, y_coma = barrel_distortion(pix_x, pix_y, alpha=alpha)
+    return x_coma, y_coma
+
+
 class MirrorAttribute(Attribute):
-    '''A frame Attribute that can only store the integers 1 and 2'''
+    """A frame Attribute that can only store the integers 1 and 2"""
 
     def convert_input(self, value):
-        '''make sure input is 1 or 2'''
+        """make sure input is 1 or 2"""
         if value in (1, 2):
             return value, False
 
-        raise ValueError('Only 1 or 2 mirrors supported')
+        raise ValueError("Only 1 or 2 mirrors supported")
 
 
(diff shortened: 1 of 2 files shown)


ctapipe/coordinates/camera_frame.py
```diff
@@ -19,35 +19,98 @@
...
+        Shape: (n_pixels)
+    y_coma : array_like
+        Cartesian coordinates of the pixels along the 'y' axis corrected for
+        Coma aberration.
+        Shape: (n_pixels)
+
+    M.Peresano, K.Kosack, 2020
+    """
+    # Take the original pixel-coordinates and correct them
+    x_coma, y_coma = barrel_distortion(pix_x.value, pix_y.value, alpha=alpha) * u.m
```
MaxNoe: This will be wrong if pix_x or pix_y is not in m. Also, why not just use barrel_distortion with units?

ctapipe/coordinates/camera_frame.py
```diff
@@ -19,35 +19,98 @@
 from .representation import PlanarRepresentation
 
 
+def barrel_distortion(x, y, alpha=0):
```
MaxNoe: If my reading is correct, this implements barrel distortion for alpha > 0 and pincushion distortion for alpha < 0

MaxNoe: Also: alpha * r is not unit less. Should it be something like `alpha * r / focal_length`?

MaxNoe:
Also, just applying with the negative alpha does not invert the transformation, as

x' = x / (1 - α r) ⇒ x = x' · (1 - α r) ≠ x' / (1 + α r)

ctapipe/coordinates/camera_frame.py
kosack: Perhaps just define alpha as having units of inverse length, then it should all work in any unit system.

kosack:
While we're modifying the Camera→Telescope frame transform, perhaps it would be a good idea now to also add a translation parameter? It would only be one extra option (or 2 really, for x and y translation)

In this coordinate transform we should do the following things:
* Apply optics transform:
  * apply focal length (done)
  * apply anamorphic corrections (done in this PR):
* apply astrometric corrections: 
  * rotation (done)
  * translation (missing) 

If translation is added, then we have all that is necessary to do astrometric (pointing) corrections and properly model the mirror's effect (At least for Davies-Cotton-like telescopes - not sure if more are needed for Schwarzschild-Couder)

ctapipe/coordinates/camera_frame.py
HealthyPear: Is it something different from meters for some cameras?
I will check again the units, as I wrote in the description I have exactly that problem when running the roundtrip test, perhaps is all connected.

kosack:
> Also, just applying with the negative alpha does not invert the transformation, as

Good point - I think this implies the unit test isn't doing it's job  - there is one that does the full round-trip, but so far it doesn't add this new correction, so it must be updated to include all parameters of the transform.

HealthyPear:
> > Also, just applying with the negative alpha does not invert the transformation, as
> 
> Good point - I think this implies the unit test isn't doing it's job - there is one that does the full round-trip, but so far it doesn't add this new correction, so it must be updated to include all parameters of the transform.

Yes, I am still working on `test_roundtrip.py`, there I have set `coma_aberration` as an attribute of the `CameraFrame` object.

ctapipe/coordinates/camera_frame.py
MaxNoe: It does not matter if we currently have cameras not in meters, this kind of thing is an anti-pattern. Functions should that expect unit input should work regardless which units are passed in if they have the correct dimension.

This function does nothing but strip units, call the other function and add a unit back on.
The whole function should not be necessary!

HealthyPear: I divided the 2 functions because perhaps one want to use barrel distortion in some other cases then Coma correction.
I am working on the units.

MaxNoe: But barrel distortion also can do pincussion distortion (if alpha has the other sign). And why not just call the distortion functions in the transform.

Also, please don't intermix optical terms. Coma is this:
https://en.wikipedia.org/wiki/Coma_(optics)

This is not directly related to distortion.

MaxNoe: * Coma is variation of magnification over the entrance pupil

* Distortion is variation of magnification over the angle of incidence

MaxNoe:
@HealthyPear please see also this comment I made in the camera transformation, I think is is directly what you want to implement here:

```
    # this assumes an equidistant mapping function of the telescope optics
    # or a small angle approximation of most other mapping functions
    # this could be replaced by actually defining the mapping function
    # as an Attribute of `CameraFrame` that maps f(theta, focal_length) -> r,
    # where theta is the angle to the optical axis and r is the distance
    # to the camera center in the focal plane
```

HealthyPear:
> @HealthyPear please see also this comment I made in the camera transformation, I think is is directly what you want to implement here:
> 
> ```
>     # this assumes an equidistant mapping function of the telescope optics
>     # or a small angle approximation of most other mapping functions
>     # this could be replaced by actually defining the mapping function
>     # as an Attribute of `CameraFrame` that maps f(theta, focal_length) -> r,
>     # where theta is the angle to the optical axis and r is the distance
>     # to the camera center in the focal plane
> ```

Hi Max,
sorry for the delay.

I actually discovered that what is done in CTA-MARS is much more simple than what I was tryng to do. I misunderstood and over-complicated the matter.
Basically what is done there is just a scaling transformation - but still, it is reported that ,

> in order to correct for the average effect of optical aberrations, that in some telescopes shift the focused light spot "outwards" (i.e. further away from the center than it would be for perfect optics), we modify the pixels coordinates before proceeding to image parametrization:, we move all pixels towards the center, with the shift being proportional to the distance from the center of the camera. That is, we "shrink" the camera by a given factor

So this is still made for optical aberrations: I get the difference with Coma, I misunderstood - it should be only because there is a nominal focal length (which is the one extracted from the simtel files) and an effective focal length (which is in the document I report in the function documentation and that gives rise to the scale factor).

I already rewrote the thing, but I was thinking together with Karl to open a separate PR for that (in which I just add a "scaling" section to the transformation functions) and to check what we can save from this one (but there seem to be problems anyway so I don't know)

PR #1221, opened 2020-02-12
CTA Reference Metadata handling

kosack:
This PR adds some helper classes that allows one to set and validate the required metadata keywords as specified in the CTA Top level Data Model document.   The classes just collect the info, and provide a `to_dict()` method that generates a flat list of *keyword=value* pairs in the correct format. This makes it easy to attach these to the header of any output file.  It works like this:

```py3
  from ctapipe.io import metadata as meta

  reference = meta.Reference(
        contact=meta.Contact(
            name="Somebody", email="a@b.com", organization="CTA Consortium"
        ),
        product=meta.Product(
            description="An Amazing Product",
            creation_time="2020-10-11 15:23:31",
            data_category="S",
            data_level="DL1",
            data_association="Subarray",
            data_model_name="Unofficial DL1",
            data_model_version="1.0",
            data_model_url="http://google.com",
            format="hdf5",
        ),
        process=meta.Process(type_="Simulation", subtype="Prod3b", id_=423442,),
        activity=meta.Activity.from_provenance(prov_activity.provenance),
        instrument=meta.Instrument(
            site="CTA-North",
            class_="array",
            type_="Layout H1B",
            version="1.0",
            id_="threshold",
        ),
    )

    # write the metadata

    table = Table(dict(x=[1, 2, 3], y=[15.2, 15.2, 14.5]))
    table.meta = reference.to_dict()
    for file_name in [tmp_path / "test.fits", tmp_path / "test.ecsv"]:
        table.write(file_name)
```

Output of `reference.to_dict()`:
```
OrderedDict([('CTA REFERENCE VERSION', '1'),
             ('CTA CONTACT EMAIL', 'a@b.com'),
             ('CTA CONTACT NAME', 'Somebody'),
             ('CTA CONTACT ORGANIZATION', 'CTA Consortium'),
             ('CTA PRODUCT ID', '122c824b-1349-462c-b503-be1d09c4cff5'),
             ('CTA PRODUCT CREATION TIME', '2020-10-11 15:23:31.000'),
             ('CTA PRODUCT DATA ASSOCIATION', 'Subarray'),
             ('CTA PRODUCT DATA CATEGORY', 'S'),
             ('CTA PRODUCT DATA LEVEL', 'DL1'),
             ('CTA PRODUCT DATA MODEL NAME', 'Unofficial DL1'),
             ('CTA PRODUCT DATA MODEL URL', 'http://google.com'),
             ('CTA PRODUCT DATA MODEL VERSION', '1.0'),
             ('CTA PRODUCT DESCRIPTION', 'An Amazing Product'),
             ('CTA PRODUCT FORMAT', 'hdf5'),
             ('CTA PROCESS ID', 423442),
             ('CTA PROCESS TYPE', 'Simulation'),
             ('CTA PROCESS SUBTYPE', 'Prod3b'),
             ('CTA ACTIVITY ID', '76a2f594-0748-4693-b277-7820f0fd7718'),
             ('CTA ACTIVITY TYPE', 'software'),
             ('CTA ACTIVITY NAME', 'test'),
             ('CTA ACTIVITY SOFTWARE NAME', 'ctapipe'),
             ('CTA ACTIVITY SOFTWARE VERSION', '0.7.0.post135+gite8ff9d5'),
             ('CTA ACTIVITY START TIME', '2020-02-12 14:09:12.606'),
             ('CTA INSTRUMENT CLASS', 'array'),
             ('CTA INSTRUMENT ID', 'threshold'),
             ('CTA INSTRUMENT TYPE', 'Layout H1B'),
             ('CTA INSTRUMENT SITE', 'CTA-North'),
             ('CTA INSTRUMENT SUBTYPE', 'unspecified'),
             ('CTA INSTRUMENT VERSION', '1.0')])
```

Eventually this can be merged with the Provenance system, so that when one registers an output file, this same information is requested (but I'll leave that for a future refactoring, since the Provenance system needs a lot of other changes)

It also adds a `DateTime` trait in `ctapipe.core.traits` that takes anything that can be convertable into an `astropy.time.Time` and renders it as an ISO time string.


MaxNoe:
Can you link the document?

kosack:
> Can you link the document?

https://forge.in2p3.fr/boards/313/topics/2651

MaxNoe:
@kosack  I also meant in the docstrings so people know what lies behind this.

ctapipe/io/metadata.py
```diff
@@ -0,0 +1,188 @@
...
+            name=activity["activity_name"],
+            _type="software",
+            _id=activity["activity_uuid"],
+            start=activity["start"]["time_utc"],
+            software_name="ctapipe",
+            software_version=activity["system"]["ctapipe_version"],
+        )
+
+    name = Unicode()
+    _type = Unicode("software")
```
MaxNoe: Are these private? Or is the underscore just to avoid overriding python builtins? If the latter, I would put the underscore at the back to not make people think these are somehow private.

MaxNoe: Also overriding builtins in a class namespace is not really an issue. It will only shadow the builtin in the class block, not even in the methods.

ctapipe/io/metadata.py
```diff
@@ -0,0 +1,188 @@
...
+
+    @default("start_time")
+    def default_time(self):
+        return Time.now().iso
+
+
+class Instrument(HasTraits):
+    """ Instrumental Context """
+
+    site = Enum(["CTA-North", "CTA-South", "Other"], "Other")
```
MaxNoe: Does this mean if I use ctapipe for another observatory, ie magic, I'm forced to use `Other` instead of 'MAGIC'?

kosack: Yes.  We could add others, though this is "CTA Metadata" the keywords start with the word "CTA" and are defined only for CTA, so it is not intended to be a generic scheme. In this case "site" means which part of CTA, not which observatory site in general (which I guess would be *Roque de los Muchachos* for CTA-N).  However, I do need to add SDMC, since some files will be produced off-site.  Anyhow, this may evolve as the CTA data model evolves, but for now I just want to support what is documented.

kosack: For files generated for other telescopes, they can define their own other metadata keywords. Probably we should have a top-level one, such as the one defined by OGIP for FITS that tells which instrument you are even talking about( `TELESCOP="CTA"` then imples use the CTA Reference Headers, while `TELESCOP="MAGIC"` would have other metadata)

kosack:
> @kosack I also meant in the docstrings so people know what lies behind this.

We still have no Document Management System for CTA (it's coming I'm told, but don't know when), so no way to link this document in a permanent way. I'll add it to the bibliography though.

ctapipe/io/metadata.py
kosack: But since we do have some files in ctapipe-extra for other telescopes, and I would like to label them with this scheme, I've added MAGIC, HESS, VERITAS, FACT, and Whipple (even though those are not really "sites"). Note that the real site will come in the Observation Configuration metadata, which is not yet defined (that gives the site name "Paranal", "Roque de los Muchachos", etc, as well as the Earth Location in lon, lat, altitude, etc).  This is just to associate a data product with a part of CTA at the highest-level.

ctapipe/io/metadata.py
kosack: ok will do (underscore after).  Yes, it was to avoid pylint from complaining about the builtin overrides, but it's true I could just ignore them, but it could lead to some confusion.

kosack:
Can we merge this? It's not meant to be a perfect solution, just something quick and dirty so we can write the appropriate headers to all output files.

kosack:
> There are a lot of warnings in the travis log of this nature:

Those are just because the CTA headers are space-separated when flattened into a single list (to be more FITS compatible, and also since HDF5 doesn't support hierarchical attributes), which leads to warnings that they are not valid python names. It doesn't really matter, just means that you have:

```py3
reference_metadata['CTA REFERENCE VERSION'] 
```
 instead of

```py3
reference_metadata.CTA_REFERENCE_VERSION
```

In a hierarchical representation, it would be `reference_metadata.CTA.REFERENCE.VERSION`

Perhaps we could just turn off that warning in pytables somehow.  Or we suggest to change the CTAO specification (comments are already over though) to allow an underscore, but then that makes the problem if an attribute in the tree has a real underscore, it can't be differentiated from the separator.

kosack:
Perhaps we could just put the write function a decorator that filters out that warning (https://docs.python.org/3/library/warnings.html)

MaxNoe:
> to be more FITS compatible, and also since HDF5 doesn't support hierarchical attributes), 

It actually supports numpy structured arrays, which is kinda cool. Don't know how langauge agnostic that is though.

kosack:
> numpy structured arrays

as attributes? Anyhow, may be overkill - I'll just try filtering the warnings.

MaxNoe:
> as attributes? Anyhow, may be overkill - I'll just try filtering the warnings.

yep

```
In [9]: with h5py.File('test.hdf5', 'w') as f: 
   ...:     f.attrs['test_dict'] = np.array([(0, 'foo')], dtype=[('x', int), ('name', 'S3')]) 
   ...:                                                                                                                                       

In [10]:                                                                                                                                      

In [10]: with h5py.File('test.hdf5', 'r') as f: 
    ...:     print(f.attrs['test_dict']) 
    ...:      
    ...:                                                                                                                                      
[(0, b'foo')]
```

Although I agree that we probably shouldn't use that

kosack:
> Although I agree that we probably shouldn't use that

interesting though. I'd have to look in the structure of the file to see what HDF5 object it translates that into!

MaxNoe:
http://davis.lbl.gov/Manuals/HDF5-1.8.7/UG/13_Attributes.html

Aparrently it can be any supported hdf5 data type, so this example will be an hdf5 array of a compound of integer and char

ctapipe/io/metadata.py
```diff
@@ -0,0 +1,222 @@
...
+    "Contact",
+    "Process",
+    "Product",
+    "Activity",
+    "Instrument",
+    "write_to_hdf5",
+]
+
+
+class Contact(HasTraits):
```
MaxNoe: Don't these all have to be config=True so one can set it in a config file?

ctapipe/io/metadata.py
```diff
@@ -0,0 +1,222 @@
...
+    name = Unicode("unknown")
+    email = Unicode("unknown")
+    organization = Unicode("unknown")
+
+
+class Product(HasTraits):
+    """Data product information"""
+
+    description = Unicode("unknown")
+    creation_time = Unicode()
```
MaxNoe: Time traitlet?

ctapipe/io/metadata.py
kosack: these aren't configuration options, I'm use using traitlets to do the validation.

ctapipe/io/metadata.py
kosack: is there one? We could create one, but for this it may be overkill. This is just to be able to write header keywords to a file.  Would be nice to validate that it is in an astropy.Time format though, but that can come later.

ctapipe/io/metadata.py
MaxNoe: Why shouldn't the analyser be able to set his or her info in the config?

kosack: I could just have the user give a dict of keywords and values, but wanted to at least do some checking to ensure they use the right values

ctapipe/io/metadata.py
MaxNoe: No, but we need one anyway probably at some point, right?

ctapipe/io/metadata.py
kosack: I suppose they could, but that wasn't the point of this for now. We could add that later, or you just make it a config option of Tool, and it's used when building this dictionary.  So far it inherits only from traitlets.HasTraits, not traitlets.config, so doesn't have configuration functionality. I even considered not using traitlets at all, but they are convenient.

bregeon:
> > @kosack I also meant in the docstrings so people know what lies behind this.
> 
> We still have no Document Management System for CTA (it's coming I'm told, but don't know when), so no way to link this document in a permanent way. I'll add it to the bibliography though.

The document has been put on the forge DMSF:
https://forge.in2p3.fr/projects/computing-info/dmsf?folder_id=1946

Although it's probably not the final place for CTAO, I would say there is hope that documents in the forge DMSF will be eventually transferred properly.

ctapipe/io/metadata.py
kosack: done

ctapipe/core/traits.py
```diff
@@ -48,13 +49,28 @@
...
 logger = logging.getLogger(__name__)
 
 
+class DateTime(TraitType):
+    """ A trait representing a point in Time, as understood by `astropy.time`"""
+    def validate(self, obj, value):
+        """ try to parse and return an ISO time string """
+        try:
+            the_time = Time(value)
+            return the_time.iso
```
MaxNoe, line 66: Why return the string and not the time instance? Now if one wants to use it, one has to convert it back to a time again.

There might also be a confusion with datetime of the standardlib.

kosack: I'll change it to Timestamp.

kosack: In the end I used `AstroTime` since it's too easy to have import problems with other `Time` objects, and this way it's clearer it's an Astropy Time.

MaxNoe: Not really. AStropy internally converts into two jd floats in a specific scale. It's only the default representation that sticks.

You could set it consistently independent of input format by using:

```python
>>> t  = Time('2020-01-20 20:00:00')
>>> print(t)
2020-01-20 20:00:00.000
>>> t.format = 'isot'
>>> print(t)
2020-01-20T20:00:00.000
```

kosack: ah, good - I'll set it to ISO then, and that should work

kosack:
> I still like the idea of setting these values through the config though.
> How are they supposed to be set now?

This is a lower-level helper function, really, just to start testing using these headers.  I think eventually, I will add some of them as config values to all `Tools`, and then the tool can automatically generate the full list from the config + provenance info by calling these functions internally.   That will happen as the main pipeline stage tools evolve, but first I just want to get something that works.

kosack:
currently there is still a problem here when writing the metadata  FITS header cards introduced by storing the times as `Time` objects instead of strings, since FITS doesn't understand astropy Times:

```
ctapipe/io/tests/test_metadata.py::test_construct_and_write_metadata
  /Users/kkosack/anaconda3/envs/cta-dev/lib/python3.7/site-packages/astropy/io/fits/convenience.py:589: AstropyUserWarning: Attribute `CTA ACTIVITY START TIME` of type <class 'astropy.time.core.Time'> cannot be added to FITS Header - skipping
    f"added to FITS Header - skipping", AstropyUserWarning)
```

So I could convert all values to strings in the `_to_dict()` function, or I have to treat the time vales specially

MaxNoe:
I would prefer to do that in the to fits method. Other data formats might use other formats for storing dates (e.g. hdf5).

There is also an open issue about the ability to directly set astropy times as values in FITS headers:
https://github.com/astropy/astropy/issues/10073

PR #1222, opened 2020-02-12
Camera scale transformation using effective focal length

HealthyPear:
**Description:**

Triggered by the comparison between _protopipe_ and _CTA-MARS_ (see cta-observatory/protopipe#37), this pull request modifies both `telescope_to_camera` and `camera_to_telescope` functions in ctapipe.coordinates.camera_frame.

The modification makes possible to use the _effective_ focal length instead of the _nominal_ one in the last step of each transformation. The values for the effective focal length will be added to `ctapipe-extra` and should be read only once at the beginning of the analysis chain.

This PR overwrites #1220 from which the distortion functions could be saved, if necessary.

The intended effect is described in the CTA-MARS step-by-step analysis [wiki-page](https://forge.in2p3.fr/projects/step-by-step-reference-mars-analysis/wiki) as follows,

> in order to correct for the average effect of optical aberrations, that in some telescopes shift the focused light spot "outwards" (i.e. further away from the center than it would be for perfect optics), we modify the pixels coordinates before proceeding to image parametrization:, we move all pixels towards the center, with the shift being proportional to the distance from the center of the camera. That is, we "shrink" the camera by a given factor

such factor is the (inverse) ratio between effective and nominal focal lengths as reported [here](https://www.mpi-hd.mpg.de/hfm/CTA/MC/Prod3/Config/PSF/flen.pdf)

**Caveats:**

- Only pixel coordinates are affected, while their surface areas remain unchanged. This is not needed for the moment, since this quantity in not used in the image parametrization process.
Could it be relevant for events with few pixels above the cleaning thresholds (low-energy events everywhere in the camera or high-energy events strongly clipped)?.

### DIFF
--- a/ctapipe/coordinates/camera_frame.py
+++ b/ctapipe/coordinates/camera_frame.py
@@ -20,34 +20,26 @@ from .representation import PlanarRepresentation
 
 
 class MirrorAttribute(Attribute):
-    '''A frame Attribute that can only store the integers 1 and 2'''
+    """A frame Attribute that can only store the integers 1 and 2"""
 
     def convert_input(self, value):
-        '''make sure input is 1 or 2'''
+        """make sure input is 1 or 2"""
         if value in (1, 2):
             return value, False
 
-        raise ValueError('Only 1 or 2 mirrors supported')
+        raise ValueError("Only 1 or 2 mirrors supported")
 
 
 # Go from SimTel / HESS to MAGIC/FACT/Engineering frame and back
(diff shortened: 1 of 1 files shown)


HealthyPear:
It appears that this change is not needed inside the trasformations.
The reason is that this is simply a value substitution between nominal and equivalente focal length.

This value substitution can be done externally by the pipeline itself, before calling HillasReconstructor.

I will close both this and PR #1220 and open an issue on protopipe instead.

Issue #855, opened 2018-11-29
Make generalized pixel groups in CameraGeometry


kosack:
```py3
neighbor_matrix = geom.group['neighbors'].matrix
patch_matrix = geom.group['patches'].matrix
patch_matrix_where = geom.group['patches'].matrix_where
patch_matrix_sparse = geom.group['patches'].matrix_sparse
```

This implies something like a PixelMapper class that has an input and output
dimension (e.g. for neighbors, it's pixel_id -> pixel_id, where as for
patches it's pixel_id -> patch_id)

kosack:
this is a repeat of #803 , so I will close it

MaxNoe:
I found that one could use shapely to combine patches using ids and then get objects one can plot nicely with matplotlib:

```
import matplotlib.pyplot as plt
from matplotlib.patches import RegularPolygon, PathPatch
from matplotlib.collections import PatchCollection
from matplotlib.path import Path
import numpy as np
from fact.instrument.camera import get_pixel_dataframe
from shapely.geometry import MultiPolygon, Polygon as ShapelyPolygon
from shapely.ops import cascaded_union


df = get_pixel_dataframe()

fig, axs = plt.subplots(1, 3, figsize=(15, 5))

for ax in axs:
    ax.set_xlim(-200, 200)
    ax.set_ylim(-200, 200)
    ax.set_aspect(1)

pixels = PatchCollection([
    RegularPolygon(xy=(x, y), numVertices=6, radius=9.55 / np.sqrt(3))
    for x, y in zip(df['x'], df['y'])
])
verts = np.array([p.vertices for p in pixels.get_paths()])

for ax, key in zip(axs, ['CHID', 'bias_patch_id', 'trigger_patch_id']):
    ax.set_title(key)
    patches = []
    for patch_id, g in df.groupby(key):

        polygons = [ShapelyPolygon(verts[i]) for i in g['CHID']]
        boundary = cascaded_union(polygons)

        if isinstance(boundary, MultiPolygon):
            paths = [
                Path(np.array(polygon.exterior.coords))
                for polygon in boundary
            ]
            path = Path.make_compound_path(*paths)
        else:
            path = Path(boundary.exterior.coords)

        patch = PathPatch(path)
        patches.append(patch)

    col = PatchCollection(patches)
    col.set_array(np.arange(len(patches)))
    col.set_edgecolor('k')
    col.set_linewidth(1)

    ax.add_collection(col)

plt.show()
```

[...]

PR #1227, opened 2020-03-11
Move pulse shape attributes from mc container to CameraGeometry

watsonjj:
The pulse shape is a property of the camera, and should be found in the camera description within the subarray container. This provides a common location for the pulse shape for both MC and real data, and makes the pulse shape more readily available for charge extraction algorithms.

### DIFF
--- a/ctapipe/io/simteleventsource.py
+++ b/ctapipe/io/simteleventsource.py
@@ -48,7 +48,7 @@ def build_camera_geometry(cam_settings, pixel_settings, telescope):
         pix_rotation=pix_rotation,
         cam_rotation=-Angle(cam_settings['cam_rot'], u.rad),
         apply_derotation=True,
-        reference_pulse_shape=pixel_settings['ref_shape'].astype('float64'),
+        reference_pulse_shape=pixel_settings['ref_shape'].astype('float64', copy=False),
         reference_pulse_step=u.Quantity(pixel_settings['ref_step'], u.ns),
     )
 
--- a/ctapipe/io/simteleventsource.py
+++ b/ctapipe/io/simteleventsource.py
@@ -48,6 +48,8 @@ def build_camera_geometry(cam_settings, pixel_settings, telescope):
         pix_rotation=pix_rotation,
         cam_rotation=-Angle(cam_settings['cam_rot'], u.rad),
         apply_derotation=True,
+        reference_pulse_shape=pixel_settings['ref_shape'].astype('float64'),
+        reference_pulse_step=u.Quantity(pixel_settings['ref_step'], u.ns),
     )
 
     return camera
@@ -317,20 +319,15 @@ class SimTelEventSource(EventSource):
             tracking_positions = array_event['tracking_positions']
             for tel_id, telescope_event in telescope_events.items():
                 tel_index = self.file_.header['tel_id'].tolist().index(tel_id)
-                telescope_description = self.file_.telescope_descriptions[tel_id]
 
                 adc_samples = telescope_event.get('adc_samples')
                 if adc_samples is None:
                     adc_samples = telescope_event['adc_sums'][:, :, np.newaxis]
                 _, n_pixels, n_samples = adc_samples.shape
-                pixel_settings = telescope_description['pixel_settings']
 
                 mc = data.mc.tel[tel_id]
                 mc.dc_to_pe = array_event['laser_calibrations'][tel_id]['calib']
                 mc.pedestal = array_event['camera_monitorings'][tel_id]['pedestal']
-                mc.reference_pulse_shape = pixel_settings['ref_shape'].astype('float64')
-                mc.meta['refstep'] = float(pixel_settings['ref_step'])
-                mc.time_slice = float(pixel_settings['time_slice'])
                 mc.photo_electron_image = (
                     array_event
                     .get('photoelectrons', {})
--- a/ctapipe/calib/camera/calibrator.py
+++ b/ctapipe/calib/camera/calibrator.py
@@ -6,6 +6,7 @@ calibration and image extraction, as well as supporting algorithms.
 import warnings
 
 import numpy as np
+from astropy import units as u
 
 from ctapipe.core import Component
 from ctapipe.image.extractor import NeighborPeakWindowSum
@@ -113,6 +114,7 @@ class CameraCalibrator(Component):
         kwargs
         """
         super().__init__(config=config, parent=parent, **kwargs)
+        self.subarray = subarray
 
         self._r1_empty_warn = False
         self._dl0_empty_warn = False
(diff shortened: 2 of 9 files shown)


ctapipe/io/simteleventsource.py
```diff
@@ -48,6 +48,8 @@ def build_camera_geometry(cam_settings, pixel_settings, telescope):
         pix_rotation=pix_rotation,
         cam_rotation=-Angle(cam_settings['cam_rot'], u.rad),
         apply_derotation=True,
+        reference_pulse_shape=pixel_settings['ref_shape'].astype('float64'),
+        reference_pulse_step=u.Quantity(pixel_settings['ref_step'], u.ns),
```
MaxNoe: Add `copy=False`?

watsonjj:
> Two cents not concerning merging:
> 
> * Rename `CameraGeometry` to `Camera` or `CameraDescription` as this is not only about camera geometry anymore. Another way would be to have a `Camera` class with members `geometry`, and `electronics` if we want it to be more hierarchical?

I agree. I will address it in a future PR.

> * Is 1 and 1 ns a good default? why not make it clear that it is not set by using `None` / `nan` and have to the code that needs it decide how to handle absent values.

Good suggestion. I will fix in this PR.

kosack:
This sounds fine to me, but it brings up some questions on what is a calibration and where does calibration data go.  If this is really just part of the description of a camera, and not the calibration of that camera, then yes, it should go in CameraGeometry. If it is calibration, it might need to go elsewhere.  

I agree CameraGeometry should then be renamed later - note that the original naming was really to have only things related to image geometry in there, and have another class for specifics of the hardware, like sampling rates, trigger patches, etc.  So CameraDescription woudl contain a CameraGeometry and a CameraHardware (or something like that)   But that can be revisited, and perhaps it is better to  just have a single CameraDescription class.

kosack:
The problem right now is that it cannot be written to a file, which is a problem - when we write a DL1 file, for example, we use `CameraGeometry.to_table()` to store the camera, and in your current implementation, this no longer works.  

One quick fix would be to perhaps allow 2 tables to be written, like in SubarrayDescription, introducing a "kind" option. So if you do `geom.to_table(kind="reference")`, you get a table with 2 columns "reference_pulse_value", and "reference_pulse_time"  that could be stored along with the pixel table (which is the default, `kind='geometry'`).  Alternately, we do what I mentioned above, and introduce another class with its own `to_table()` option.

MaxNoe:
I don't understand why it is currently impossible to write the pulse shape.

Each camera geometry gets its own file, so the variable length array is not a problem since only one entry is in the file, right?

kosack:
The current file is a table with each row representing a pixel

kosack:
The idea of the instrument module was that it was easily mappable to a set of database tables in the future, so making it store things that don't map to a relational database might be problematic.

MaxNoe:
> The idea of the instrument module was that it was easily mappable to a set of database tables in the future, so making it store things that don't map to a relational database might be problematic.

Most database today have very general support for those kinds of things, e.g. by allowing very powerfull `JSON` type columns.

watsonjj:
Why do we use astropy tables here? This problem could be easily avoided if we could just store arbitrary arrays to a hdf5 file (for example). When we move to a database, wouldn't each array be a single record (not each element of the array), so then this file would be similar in the retrieval of parameters?

watsonjj:
> This sounds fine to me, but it brings up some questions on what is a calibration and where does calibration data go. If this is really just part of the description of a camera, and not the calibration of that camera, then yes, it should go in CameraGeometry. If it is calibration, it might need to go elsewhere.

I think it belongs as in the description of the camera. Its also an input to the sim_telarray simulation. I don't know of any cameras that need to regularly update their pulse shape description.

MaxNoe:
> When we move to a database, wouldn't each array be a single record (not each element of the array), so then this file would be similar in the retrieval of parameters?

Yes. there are several options for storing arrays in a database row. 
Binary BLOB  columns (encoded somehow (protobuf, avro, paquet, msgpack, bson) or just plain data), JSON columns and probably more.

I don't think this has anything to do with the problem here.

If we wanted to prepare for a database system of the configuration, it would probably best to store this in an `sqlite` database for now. Adaption to a production db would then be rather trivial, just change the connection url in sqlalchemy or whatever db abstraction layer we choose.

watsonjj:
> If we wanted to prepare for a database system of the configuration, it would probably best to store this in an sqlite database for now. Adaption to a production db would then be rather trivial, just change the connection url in sqlalchemy or whatever db abstraction layer we choose.

And would you also propose to replace the camgeom fits files in ctapipe-extra with this?

MaxNoe:
> And would you also propose to replace the camgeom fits files in ctapipe-extra with this?

I'm not actually proposing to do this. I only said that that would be the logical step if we wanted to try databases. 

Having a file representation and a database representation is probably better than reducing to what sqlite can support.

It does not support arrays, for example, while psql does:
https://www.postgresql.org/docs/current/arrays.html

watsonjj:
> The problem right now is that it cannot be written to a file, which is a problem - when we write a DL1 file, for example, we use CameraGeometry.to_table() to store the camera, and in your current implementation, this no longer works since you would then lose this information.

Do you currently store the MC info to the DL1 file? If so, how did you previously handle this variable length array?

watsonjj:
Can we revisit the storage of the reference pulse in the table for another time? Its an important point, but as the charge is extracted by the time the DL1 file is written, it is unlikely that any algorithm down the chain will need knowledge of the pulse shape, so I feel this is acceptable for the time being.

kosack:
> @kosack Do you currently store the MC info to the DL1 file? If so, how did you previously handle this variable length array?

I currently store them exactly the same way as in FITS format - I just use astropy.table to convert to a HDF5 dataset.  It's the main reason for using Astropy.table - you can easily manipulate the data, units and other metadata are preserved,  and no special io libraray is needed to read them. They can be stored in FITS, ECSV (text), HDF5, etc, with no change to the code. 

I'd not suggest to do any implementation in SQLIte or anything right now - the point is this info will be stored by CTAO, and it's not up to ctapipe to do it, but I wanted to make it easy to write things in simple-to-read formats for now, so that we don't need extra code to understand them. Astropy.table is great for that.  

Anyhow, I don't see any problem just storing this pulse info in its own table, right? That's whats already done for the TelescopeDescription, SUbarrayDescription, etc.   Each is defined by multiple tables.   Rather than hacking our own solution and inventing a new file format, just a simple table with 2 columns time, and reference_value is enough, and makes it really easy for anyone to look at or change the data without ctapipe.

watsonjj:
> I just use astropy.table to convert to a HDF5 dataset.

I'm confused. The `MCCameraEventContainer` currently has the fields `photo_electron_image` with size (n_pixels) and `reference_pulse_shape` with size (n_ref_pulse_samples). If you are already able to store both of these into an astropy table, then we should be able to do the same here, right?

watsonjj:
> Currently the reference pulse shape is not stored in DL1, since it's not needed after that point.

Just seen your edit. In that case no functionality is lost by this PR, and we can perform the refactoring of `CameraGeometry` in a separate PR.

kosack:
>I'm confused. The MCCameraEventContainer currently has the fields photo_electron_image with size >(n_pixels) and reference_pulse_shape with size (n_ref_pulse_samples). If you are already able to store both of these into an astropy table, then we should be able to do the same here, right?

There's no problem writing it, it's just that CameraGeometry is a table with one row per pixel, and that information is not per-pixel, so you need another table.

MaxNoe:
Oh, right.

So CameraGeometry table should probably stay the same and we can now think about a sensical fits layout.

So we can have a table for a geometry, a table for pulse shape, a table for pedestal coefficients, whatever in a camera description fits file.

kosack:
exactly

MaxNoe:
That would be great and would also solve things like #855 I think.

watsonjj:
Can we approve this PR? I will perform the refactoring of the CameraGeometry class in a separate PR

PR #1233, opened 2020-03-13
Improvement of integration_correction testing, implementation, and docstring

watsonjj:
The integration_correction function corrects an integration of a pulse for the part of the pulse that is outside the integration window. If a noise-less unit pulse is being integrated, then the `integration * correction` should always equal 1, irregardless of the integration window width and shift.

This PR improves the integration_correction method:
- Tests now check that the values returned are correct as per the intention of the method
- Integration windows that extend beyond the reference pulse are now correctly corrected
- The implementation is cleaned up a bit, and the variable names are more clear
- The docstring is improved to more accurately describe the purpose of the method

### DIFF
--- a/ctapipe/calib/camera/calibrator.py
+++ b/ctapipe/calib/camera/calibrator.py
@@ -15,7 +15,7 @@ __all__ = ["CameraCalibrator"]
 
 
 def integration_correction(
-    n_chan, pulse_shape, refstep, time_slice, window_width, window_shift
+    reference_pulse_shape, reference_pulse_step, sampled_step, window_width, window_shift
 ):
     """
     Obtain the integration correction for the window specified.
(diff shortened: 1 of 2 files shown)


ctapipe/calib/camera/calibrator.py
```diff
@@ -15,62 +15,59 @@
...
-        Numpy array containing the pulse shape for each channel.
-    refstep : int
+    reference_pulse_shape : ndarray
+        Numpy array containing the pulse shape for each gain channel
+    reference_pulse_step : float
         The step in time for each sample of the reference pulse shape
-    time_slice : int
-        The step in time for each sample of the waveforms
+    sampled_step : int
+        The step in time for each sample of the sampled waveforms
```
kosack: This could be documented better- I'm not sure I understand what it is from the description. It sounds like it should be the width of the time bins for the waveform, but that would not be an integer necessarily.  Does it have a unit like ns?  Maybe calling it sample_width?

watsonjj: Thanks, done

ctapipe/calib/camera/calibrator.py
```diff
@@ -35,13 +35,14 @@ def integration_correction(
     reference_pulse_shape : ndarray
         Numpy array containing the pulse shape for each gain channel
     reference_pulse_step : float
-        The step in time for each sample of the reference pulse shape
-    sampled_step : int
-        The step in time for each sample of the sampled waveforms
+        The step in time for each sample of the reference pulse shape in ns
+    sample_width : float
```
kosack: one more pedantic comment, if *not* using astropy quantities here, you should name it `sample_width_ns`.  Otherwise, looks good.

PR #1234, opened 2020-03-13
Remove extract_pulse_time_around_peak

watsonjj:
In #1146 `sum_samples_around_peak` and `extract_pulse_around_peak` were combined into `extract_around_peak`, but I forgot to delete `extract_pulse_around_peak` (which is no longer used anywhere). This PR removes `extract_pulse_around_peak`.


kosack:
might need to pull in latest master so that travis stops complaining.

PR #1235, opened 2020-03-18
Fix comparison warning

MaxNoe:
works but relies on an cpython implementation detail


MaxNoe:
Ok, what behaviour was intended here? This now fails because it compares an array.

Before, `somearray is not 0` would enter the true branch.

kosack:
might check with @ParsonsRD if in doubt

MaxNoe:
@watsonjj but the change fails because it now compares 0 to a numpy array.

So I need some information how to convert this check.

PR #1239, opened 2020-03-20
Replace np.power with power operator, fixes #1028

### LINKED ISSUE #1028: numpy.power(x, 2) --> x**2
I happened to run some profiling on simple numpy functions:

```python
In [43]: x = np.random.rand(10)
In [44]: %timeit np.power(x,2)
1.94 µs ± 5.53 ns per loop (mean ± std. dev. of 7 runs, 1000000 loops each)
In [45]: %timeit x**2
1.05 µs ± 11.7 ns per loop (mean ± std. dev. of 7 runs, 1000000 loops each)

In [46]: x = np.random.rand(100)
In [47]: %timeit np.power(x,2)
2.62 µs ± 13.7 ns per loop (mean ± std. dev. of 7 runs, 100000 loops each)
In [48]: %timeit x**2
1.11 µs ± 7.62 ns per loop (mean ± std. dev. of 7 runs, 1000000 loops each)

In [49]: x = np.random.rand(1000)
In [50]: %timeit np.power(x,2)
9.97 µs ± 255 ns per loop (mean ± std. dev. of 7 runs, 100000 loops each)
In [51]: %timeit x**2
2.2 µs ± 9.53 ns per loop (mean ± std. dev. of 7 runs, 100000 loops each)

In [52]: x = np.random.rand(10000)
In [53]: %timeit np.power(x,2)
76.1 µs ± 383 ns per loop (mean ± std. dev. of 7 runs, 10000 loops each)
In [54]: %timeit x**2
4.77 µs ± 21.8 ns per loop (mean ± std. dev. of 7 runs, 100000 loops each)
 ```

So the better-looking `array**2` instead of `np.power(array, 2)` is from 2 to 30 times faster...we should change it in all the code.

This is valid only for `array**2`, whereas for `array**3` or the two methods have the same performance.

MaxNoe:


kosack:
This made me notice that we still  have Charge Resolution calculation stuff in ctapipe - that has all (or will be) moved to notebooks that operate on DL1 files, so not needed here anymore. We could drop that code.

watsonjj:
> This made me notice that we still have Charge Resolution calculation stuff in ctapipe - that has all (or will be) moved to notebooks that operate on DL1 files, so not needed here anymore. We could drop that code.

Yes, I was waiting on the DL1 writer...

PR #1240, opened 2020-03-20
Fix hillas width 0, fixes #772

### LINKED ISSUE #772: Corrupted weighting in HillasReconstructor for small images

MaxNoe:
By rounding values very close to zero to exact zero, nan values are avoided.

These happend due to floating point imprecisions resulting in covariances like `-8.47032947e-22`,
and then sqrt returns nan.

### DIFF
--- a/ctapipe/image/hillas.py
+++ b/ctapipe/image/hillas.py
@@ -11,6 +11,9 @@ from astropy.units import Quantity
 from ..io.containers import HillasParametersContainer
 
 
+HILLAS_ATOL = np.finfo(np.float64).eps
+
+
 __all__ = [
     'hillas_parameters',
     'HillasParameterizationError',
@@ -138,7 +141,7 @@ def hillas_parameters(geom, image):
     eig_vals, eig_vecs = np.linalg.eigh(cov)
 
     # round eig_vals to get rid of nans when eig val is something like -8.47032947e-22
-    near_zero = np.isclose(eig_vals, 0)
+    near_zero = np.isclose(eig_vals, 0, atol=HILLAS_ATOL)
     eig_vals[near_zero] = 0
 
     # width and length are eigen values of the PCA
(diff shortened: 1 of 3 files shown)


MaxNoe:
Ah, dam. It only build the second commit on travis. I wanted to show that the first one fails and the fix makes it pass.

MaxNoe:
Fixes  #772

MaxNoe:
Sorry @kosack, one more time. We do indeed want to test for exact equality here and I also fixed the random seed.

vuillaut:
Hi @MaxNoe 
Is event weighting handled correctly when `width=0` ?

MaxNoe:
> Hi @MaxNoe
> Is event weighting handled correctly when width=0 ?

I think the hillas reconstructor raises if any telescope events have width=0 since https://github.com/cta-observatory/ctapipe/pull/994

vuillaut:
> > Hi @MaxNoe
> > Is event weighting handled correctly when width=0 ?
> 
> I think the hillas reconstructor raises if any telescope events have width=0 since #994

Indeed.
Looks good to me then.

PR #1242, opened 2020-03-23
Improvements to ImageExtractor wrt integration_correction and sampling rate

watsonjj:
The integration_correction is a powerful tool. It removes the dependance of the resulting extracted charge on the size of the integration window used. This means that any other calibration is integration window independent!

However, there are two problems in the way we currently use it:
1. As it is currently attached to the CameraCalibrator, it is the same method applied for any ImageExtractor. For the currently defined extractors this has been somewhat alright, as they are all integration-window based. But for more advanced extractors (e.g. digital filter), this will not be the correct correction to apply
2. We allow it to be skipped if there is no reference_pulse_shape defined. This means that real camera data might not benefit from this function if the developer has not specified a reference pulse shape, and then their calibration is integration window dependant (which may be accidentally overlooked)

I have addressed both of these points in this PR, by:
- Adding a `_calculate_correction` method to the ImageExtractors, so one can define the appropriate way to calculate the correction for each particular extractor. For a noise-less unit pulse the resulting charge from an ImageExtractor should always equal 1, ensuring the correct units for the resulting charge - this is the goal of the correction factor.
- Removing the ability to skip over the correction if there is no reference pulse defined.
- Moving integration_correction method to extractor.py, as this is a more appropriate place for it now.

Also included in this PR is the correct application of the sampling rate to the extracted charge and pulse time. So they are now in units of "waveform unit * nanosecond" and "nanosecond" respectively. 

These changes have allowed me to update the tests for ImageExtractor so that they no longer expect an otherwise meaningless value for the charge, but instead the charge should equal the total charge (i.e. the true charge) specified in generating the waveform.

There is no change to the API of the calibration in this PR.

### DIFF
--- a/ctapipe/calib/camera/calibrator.py
+++ b/ctapipe/calib/camera/calibrator.py
@@ -4,75 +4,12 @@ calibration and image extraction, as well as supporting algorithms.
 """
 
 import warnings
-
 import numpy as np
-from astropy import units as u
-
 from ctapipe.core import Component
 from ctapipe.image.extractor import NeighborPeakWindowSum
 from ctapipe.image.reducer import NullDataVolumeReducer
 
 __all__ = ["CameraCalibrator"]
-
-
-def integration_correction(
-    reference_pulse_shape, reference_pulse_step, sample_width_ns,
-    window_width, window_shift
-):
-    """
-    Obtain the correction for the integration window specified.
-
-    For any integration window applied to a noise-less unit pulse, the
-    correction (returned by this function) multiplied by the integration
-    result should equal 1.
-
-    This correction therefore corrects for the Cherenkov signal that may be
-    outside the integration window, and removes any dependence of the resulting
-    image on the window_width and window_shift parameters. However, the width
-    and shift of the window should still be optimised for the pulse finding and
-    to minimise the noise included in the integration.
-
-    Parameters
-    ----------
-    reference_pulse_shape : ndarray
-        Numpy array containing the pulse shape for each gain channel
-    reference_pulse_step : float
-        The step in time for each sample of the reference pulse shape in ns
-    sample_width_ns : float
-        The width of the waveform sample time bin in ns
-    window_width : int
-        Width of the integration window (in units of n_samples)
-    window_shift : int
-        Shift to before the peak for the start of the integration window
-        (in units of n_samples)
-
-    Returns
-    -------
-    correction : ndarray
-        Value of the integration correction for each gain channel
-    """
-    n_channels = len(reference_pulse_shape)
-    correction = np.ones(n_channels, dtype=np.float)
-    for ichannel, pulse_shape in enumerate(reference_pulse_shape):
-        pulse_max_sample = pulse_shape.size * reference_pulse_step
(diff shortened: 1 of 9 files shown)


watsonjj:
This PR will also solve #1044

kosack:
It would be useful to also add a config option to disable this correction - @HealthyPear  had to hack that in ProtoPipe in order to cross-check with MARS (which does no integration correction). So just a boolean `enable_integration_correction` would be great, perhaps in the base Extractor class, such that _compute_correction() is not called if is is disabled.

watsonjj:
Yeah I had considered this. Honestly I'd prefer to avoid adding this option...the integration correction should be an integral part of the charge extraction, and having that option could lead to misuse...

With regards to the cross-checks - the integration correction should only introduce a scale factor difference. Unless I am mistaken, the comparisons done by @HealthyPear remove account for  possible bias due to different calibrations, and so should be unaffected by the integration_correction. Is that correct @HealthyPear?

watsonjj:
Besides, with the integration correction, the extracted charge from simtelarray should match 1-to-1 with the true charge, with no additional calibration needed. So it is technically the correct way to process the MC (especially for calculating the charge resolution).

This is a plot of the extracted charge distribution for different window widths and shifts, versus the true charge for **no integration correction applied**:
image

This is the same plot **with integration correction applied**:
image

HealthyPear:
Hi @watsonjj !

Yes I agree with you.

It is true that I had to do this hack into _protopipe_ as @kosack mentioned, but:
- it is no the "worst" hack; there are bigger things that I had to overwrite and they will disappear almost completely with the next release of _ctapipe_,
- the fact that I had to disable the integration correction is only temporary, since I did it just to minimize any possible source of mistake in the comparison; as you correctly state, applying it will just change the (average) bias in the charge resolution, for which we correct when assessing the calibration benchmarks (i.e. the calibration notebook in protopipe).

I agree that this correction should be always there.
If we want to implement this option anyway, a possible solution would be to set it true by default and let the the documentation will do the rest...

moralejo:
A couple of (late) comments on this:

> * Adding a `_calculate_correction` method to the ImageExtractors, so one can define the appropriate way to calculate the correction for each particular extractor. For a noise-less unit pulse the resulting charge from an ImageExtractor should always equal 1, ensuring the correct units for the resulting charge - this is the goal of the correction factor.

It does not seem to me that the goal is the right one. Real or simulated pulses are composed of photons with a time structure. Take a single p.e. pulse shape (like LST's low-gain one) which has some "ringing" with undershoots after the main peak. For the single p.e. shape, if you integrate a number of samples which don't include any undershoot, the correction factor would be <1. But for a real pulse (be it a calibration pulse, or a pulse from a shower), with a time structure, the main peaks of the photon pulses will fill the undershoots of earlier photons, so the mean or typical pulse will have a pretty different shape, and the correction will be wrong.

> * Removing the ability to skip over the correction if there is no reference pulse defined.

I actually think it would be nice, at least for testing purposes, if the pulse correction could be skipped altogether in some easy way, through a line in the configuration file.

PR #1244, opened 2020-03-30
Waveform toy model

watsonjj:
`obtain_time_image` creates a simple timing image for a Cherenkov shower toy model

`WaveformModel` creates a waveform based on an array of charges (per pixel) and an array of time (per pixel). It does this by convoluting the reference pulse shape with an upsampled waveform, with the charges placed in the samples corresponding to the pulse timing.

These classes can be used in the future for testing image extraction techniques.

### DIFF
--- a/ctapipe/image/tests/test_toy.py
+++ b/ctapipe/image/tests/test_toy.py
@@ -21,7 +21,7 @@ def test_intensity():
     # make a toymodel shower model
     model = Gaussian(x=x, y=y, width=width, length=length, psi=psi)
 
-    image, signal, _ = model.generate_image(
+    _, signal, _ = model.generate_image(
         geom, intensity=intensity, nsb_level_pe=5,
     )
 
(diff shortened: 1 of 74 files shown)


watsonjj:
Here is a waveform movie to demonstrate the class, corresponding to the event I generate in the tests:
movie

kosack:
By the way, I'd recommend running `black` on all code before merging. I'll start enforcing that soon. That way we are sure that the formatting is standard everywhere.

ctapipe/image/tests/test_toy.py
```diff
@@ -58,7 +58,7 @@ def test_skewed():
     model = SkewedGaussian(
         x=x, y=y, width=width, length=length, psi=psi, skewness=skewness
     )
-    image, signal, _ = model.generate_image(geom, intensity=intensity, nsb_level_pe=5,)
+    _, _, _ = model.generate_image(geom, intensity=intensity, nsb_level_pe=5,)
```
MaxNoe, line 61: then just don't save anything in variables at all :wink:

watsonjj: ah...yeh...

Issue #1159, opened 2019-10-28
Proposal: require "black" code formatter as pre-commit hook


kosack:
To avoid style change problems, I suggest we require devs to either manually run "black" (or in their editor of choice), or install it as a pre-commit hook (see e.g. https://ljvmiranda921.github.io/notebook/2018/06/21/precommits-using-black-and-flake8/)

Black Documentation: https://black.readthedocs.io/en/stable/index.html

For bits of code where black really does a worse job (are there any?), I think it can be disabled explicitly similar to yapf's `# yapf: disable`

Edit: actually it's `# fmt: off` and end with `# fmt: on`.

watsonjj:
Do we also adopt the 88 character line length?

kosack:
It's ok with me - I already have the tests only check for 90 I think, even though we recommend 80.  The idea of 88 is just that you should normally stick to 80, but it's ok to go over a bit in cases where it makes sense.  The alternative is to specify it in a config file, so black uses 80.

vuillaut:
To be considered:
https://github.com/marketplace/actions/black-code-formatter

kosack:
How does that work? Does it modify the code when you push? Wouldn't that create conflicts? Or does it just warn that the user didn't run black?

kosack:
Ah I see it just uses the --check option, so might be a good idea ! I guess it's like a harsher version of what codacy is already doing with its pep8 checking, but also applying to actual formatting.

MaxNoe:
done

PR #1246, opened 2020-04-01
Add pyproject toml

MaxNoe:
Reference for pyproject.toml and why we should have one:
https://www.python.org/dev/peps/pep-0518/
https://www.python.org/dev/peps/pep-0517/
https://snarky.ca/what-the-heck-is-pyproject-toml/

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,0 +1,3 @@
+[build-system]
+requires = ["setuptools >= 40.6.0", "wheel"]
+build-backend = "setuptools.build_meta"
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,3 +1,21 @@
 [build-system]
 requires = ["setuptools >= 40.6.0", "wheel"]
 build-backend = "setuptools.build_meta"
+
+[tool.black]
+line-length = 88
+target-version = ['py36', 'py37', 'py38']
+include = '\.pyi?$'
+exclude = '''
+/(
+    \.eggs
+  | \.git
+  | \.hg
+  | \.mypy_cache
+  | \.tox
+  | \.venv
+  | _build
+  | buck-out
+  | build
+)/
+'''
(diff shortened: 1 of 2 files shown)


kosack:
Should we put any other config in there while we're at it? Like `black` configs? (https://github.com/psf/black/blob/master/pyproject.toml) - the defaults are pretty much correct, but it might be nice to specify only the python version we support so it uses modern features when possible, and also to exclude files it shoudn't format when running in recursive mode.  Soon I would like to install it as a pre-commit hook, after we reformat all code, so we fix all small style problems. If the config is already there, that would help.

kosack:
I guess this can eventually replace setup.py, but it seems like the features are not all there yet (installing in developer mode for example). I wonder if conda will support it eventually? Right now, I build the conda package in a pretty automated fashion, reading the necessary info from setup.py, but that will break I guess if we go full PEP 517

MaxNoe:
There is no "full" pep 517. You need a build system. You can now specify that build system in the pyproject.toml.

But you still have to choose `setuptools`, or `poetry` or something else.
We could think about switching to poetry. But I think that's another matter.

MaxNoe:
I added the black config

kosack:
> There is no "full" pep 517. You need a build system. You can now specify that build system in the pyproject.toml.

What I meant was this quote here from the article above: 

> But there's a bonus if you use a pyproject.toml file with a setup.cfg configuration for setuptools: you don't need a setup.py file anymore! 

Right now we have a mix of setup.py, setup.cfg, and pyproject.toml.  But I guess that's fine for now.

MaxNoe:
Ah ok. But we probably want to keep the automatic versioning, right? So that needs a script, not just a config.

watsonjj:
Does this address #1159?

MaxNoe:
@watsonjj now, it just adds a config, it does not run black by itself

Issue #1224, opened 2020-02-28
Containers, Fields and units


vuillaut:
Hi.

I always assumed (but discover this is only at file reading) that the unit in the Containers enforced the setting of compatible units.
I mean that I expected
```
from ctapipe.io import HDF5TableReader, HDF5TableWriter
from ctapipe.io.containers import Container, Field
import astropy.units as u

class MyContainer(Container):
  x = Field(None, unit=u.m)

c = MyContainer()
c.x = 1 * u.TeV
```
to throw an error.

And that I would get
```
c.x = 1 * u.cm
c.x 
Out[1]: <Quantity 0.01 m>
```
Is this something we would like to implement ?

I'd like to have your feedback before trying something.

MaxNoe:
Related #445, #501

vuillaut:
> Related #445, #501

I should have made a search before asking  🤦‍♂️
Thanks.

kosack:
Just to note here, even though it's closed, I think validation on setting a value is probably too costly for performance, but I always intended to implement a .validate() method of Containers that could be called once, e.g. during the schema generation in HDF5TableWriter, just to check that things are ok.  Presumably, since containers are filled for every event, one doesn't need to check every time that something has the right data type and unit. 

In any case, this is not yet implemented, and I've run into problems because it is missing.  E.g. in the current demo of writing the DL1 data, I have the problem that since we use the default value to define the data type, but the default is often np.nan or None, we lose that information, and if the container is not filled correctly on the first event, the schema is incorrect.  So this should be fixed.

PR #1248, opened 2020-04-02
Replace muon container defaults with nan, remove redundant fields

MaxNoe:
The muon containers had a lot of reduntant fields (obs id, event id, input file) that should be handled by the respective containers or by the provenance system.

Also, they were initialized to valid values, not nan.

### DIFF
--- a/ctapipe/io/containers.py
+++ b/ctapipe/io/containers.py
@@ -558,7 +558,6 @@ class MuonRingParameter(Container):
         np.full((3, 3), nan), "covariance matrix of the muon ring fit"
     )
     ring_containment = Field(nan, "containment of the ring inside the camera")
-    ring_fit_method = Field(None, "fitting method used for the muon ring")
 
 
 class MuonIntensityParameter(Container):
@@ -581,7 +580,6 @@ class MuonIntensityParameter(Container):
     prediction = Field(None, "image prediction")
     mask = Field(None, "image pixel mask")
     optical_efficiency_muon = Field(nan, "optical efficiency muon")
-    intensity_fit_method = Field("", "intensity fit method")
 
 
 class HillasParametersContainer(Container):
--- a/ctapipe/io/containers.py
+++ b/ctapipe/io/containers.py
@@ -569,14 +569,14 @@ class MuonIntensityParameter(Container):
     ring_width = Field(nan, "width of the muon ring in degrees")
     ring_time_width = Field(nan, "duration of the ring image sequence")
     impact_parameter = Field(
-        nan, "distance of muon impact position from centre of mirror"
+        nan, "distance of muon impact position from center of mirror"
     )
     impact_parameter_chi2 = Field(nan, "impact parameter chi squared")
     intensity_cov_matrix = Field(nan, "covariance matrix of intensity")
     impact_parameter_pos_x = Field(nan, "impact parameter x position")
     impact_parameter_pos_y = Field(nan, "impact parameter y position")
-    cog_x = Field(nan, "Centre of Gravity x")
-    cog_y = Field(nan, "Centre of Gravity y")
+    cog_x = Field(nan, "Center of Gravity x")
+    cog_y = Field(nan, "Center of Gravity y")
     prediction = Field(None, "image prediction")
     mask = Field(None, "image pixel mask")
     optical_efficiency_muon = Field(nan, "optical efficiency muon")
@@ -593,7 +593,7 @@ class HillasParametersContainer(Container):
     phi = Field(nan, "polar coordinate of centroid", unit=u.deg)
 
     length = Field(nan, "standard deviation along the major-axis")
-    width = Field(nan, "stndardad spread along the minor-axis")
+    width = Field(nan, "standard spread along the minor-axis")
     psi = Field(nan, "rotation angle of ellipse", unit=u.deg)
 
     skewness = Field(nan, "measure of the asymmetry")
--- a/ctapipe/io/containers.py
+++ b/ctapipe/io/containers.py
@@ -315,52 +315,52 @@ class MCHeaderContainer(Container):
             "[0]=R.A., [1]=Declination in mode 1."
         ),
     )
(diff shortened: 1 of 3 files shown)


kosack:
For the initialization to np.nan, that still causes some big problems when writing the data, mainly that the TableWriter cannot deduce the correct type, and then subsequent events fail if the nans are not filled in on the first event.  For now we have to skip all initial events that contain nan values.  Probably that should be fixed by #1224 though.

MaxNoe:
> The main problem is that setting all to nan now makes it really hard to tell what data type they were supposed to be. 

floats!

I have entered -1 for integers that are normally positive. The hard part is the marker for signed integers.

ctapipe/io/containers.py
```diff
@@ -542,120 +542,46 @@ class SST1MDataContainer(DataContainer):
...
+    )
+    ring_center_distance = Field(
+        nan * u.deg, "Distance of ring center from camera center", unit=u.deg
+    )
+    ring_chi2_fit = Field(nan, "chisquare of the muon ring fit", unit=u.deg)
+    ring_cov_matrix = Field(
+        np.full((3, 3), nan), "covariance matrix of the muon ring fit"
+    )
+    ring_containment = Field(nan, "containment of the ring inside the camera")
+    ring_fit_method = Field(None, "fitting method used for the muon ring")
```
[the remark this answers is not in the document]
kosack: though now that I think about it - does this change per event? It could just be configuration info, and dropped here.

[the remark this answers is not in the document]
MaxNoe: Maybe they wanted to add more than one? But that could be done with two prefixes now.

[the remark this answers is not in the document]
kosack: yeah, you can always add a column using Pandas later when merging two datasets. It's likely not needed now.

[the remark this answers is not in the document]
MaxNoe: At the only place it was used, it inserted the config value of the Component, so I removed it.

MaxNoe:
Let's merge this after #1245 is merged

ctapipe/io/containers.py
```diff
@@ -668,8 +592,8 @@ class HillasParametersContainer(Container):
     r = Field(nan, "radial coordinate of centroid")
     phi = Field(nan, "polar coordinate of centroid", unit=u.deg)
 
-    length = Field(nan, "RMS spread along the major-axis")
-    width = Field(nan, "RMS spread along the minor-axis")
+    length = Field(nan, "standard deviation along the major-axis")
+    width = Field(nan, "stndardad spread along the minor-axis")
```
rlopezcoto, line 596: typo

ctapipe/io/containers.py
```diff
@@ -542,120 +542,44 @@ class SST1MDataContainer(DataContainer):
...
-    ring_completeness = Field(0.0, "fraction of ring present")
-    ring_pix_completeness = Field(0.0, "fraction of pixels present in the ring")
-    ring_num_pixel = Field(0, "number of pixels in the ring image")
-    ring_size = Field(0.0, "size of the ring in pe")
-    off_ring_size = Field(0.0, "image size outside of ring in pe")
-    ring_width = Field(0.0, "width of the muon ring in degrees")
-    ring_time_width = Field(0.0, "duration of the ring image sequence")
+    ring_completeness = Field(nan, "fraction of ring present")
+    ring_pix_completeness = Field(nan, "fraction of pixels present in the ring")
+    ring_num_pixel = Field(-1, "number of pixels in the ring image")
```
rlopezcoto, line 566: why not `nan` as well?

ctapipe/io/containers.py
```diff
@@ -542,120 +542,44 @@ class SST1MDataContainer(DataContainer):
...
-    prediction = Field([], "image prediction")
-    mask = Field([], "image pixel mask")
-    optical_efficiency_muon = Field(0.0, "optical efficiency muon")
-    intensity_fit_method = Field("", "intensity fit method")
-    inputfile = Field("", "input file")
+    impact_parameter_chi2 = Field(nan, "impact parameter chi squared")
+    intensity_cov_matrix = Field(nan, "covariance matrix of intensity")
+    impact_parameter_pos_x = Field(nan, "impact parameter x position")
+    impact_parameter_pos_y = Field(nan, "impact parameter y position")
+    cog_x = Field(nan, "Centre of Gravity x")
```
rlopezcoto, line 578: changing `COG_x` -> `cog_x` may affect some of the scripts that are using these containers, but ok.

ctapipe/io/containers.py
kosack: there is no nan for integers

MaxNoe: nan implies float or double, this is a positive integer so -1 as missing value marker makes more sense

ctapipe/io/containers.py
MaxNoe: The other stuff I changed will do the same, this is a breaking change, yes

kosack: that will be the case in all of the container cleanups that will happen soon - so far we use only lower case elsewhere, so I would avoid doing something different here.

MaxNoe: By the way, are these fields the same has in the hillas parameters? surely looks the same

rlopezcoto: `COG_x` (or `cog_x`) is likely calculated in an "uncleaned" image whose pixel_ids have been limited by their distance to the fitted ring. It is the same as `ring_size`. Since they are not calculated the same, I would still keep them separated, maybe calling them `ring_cog_x/y`?

MaxNoe: this will be dealt with container prefixes, a thing inside a container named `MuonRing` should not be called something like `ring_`

MaxNoe:
okay, that didn't create conflicts with #1245 , I'm slightly surprised, but let's go then @rlopezcoto

PR #1251, opened 2020-04-04
Fix input url handling in display_dl1

MaxNoe:
The event source would always take the gamma simtel file and not the file given on the command line.

This allows it to get the `input_url` traitlet also from config or parent, not only from the keyword argument.

### DIFF
--- a/ctapipe/tools/display_dl1.py
+++ b/ctapipe/tools/display_dl1.py
@@ -173,15 +173,14 @@ class DisplayDL1Calib(Tool):
 
     def __init__(self, **kwargs):
         super().__init__(**kwargs)
+        self.config.EventSource.input_url = get_dataset_path('gamma_test_large.simtel.gz')
         self.eventsource = None
         self.calibrator = None
         self.plotter = None
 
     def setup(self):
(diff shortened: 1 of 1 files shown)


ctapipe/tools/display_dl1.py
```diff
@@ -173,15 +173,14 @@ class DisplayDL1Calib(Tool):
 
     def __init__(self, **kwargs):
         super().__init__(**kwargs)
+        self.config.EventSource.input_url = get_dataset_path('gamma_test_large.simtel.gz')
```
kosack, line 176: Why not just use the default value, rather than setting the config explicitly here?

MaxNoe: Where could I set the default? I don't want to set it globally for `EventSource` but just in this script

kosack: Ah yes.  It's not possible without changing it in the EventSource. I had a similar problem in the DL1 tool for defining new defaults... This works, but it is a bit confusing as it looks like a configuration is being forced, even though it's overwritten later by the command-line.

PR #1253, opened 2020-04-04
Muon refactoring

MaxNoe:
* Move `MuonRingFitter` into an appropriately called file
* Remove the unused abstract baseclass that was there before
* Remove the `ring_` prefix from the fields in the `MuonRingParameterContainer`


MaxNoe:
The diff is somewhat misleading. I removed the old muon fitter (which was probably intendend as a base class but was actually not used anywhere) and moved the strangely named `muon_ring_finder` into that `ring_fitter` file.

MaxNoe:
@kosack could you review?

PR #1254, opened 2020-04-06
Remove unused and untested muon fitting code

MaxNoe:
First step in refactoring of the muon code.

This removed a bunch of unused and untested functions, added by myself four years ago.

Maybe I'll add some of these back, but then by integrating it with the other code and with tests added.

So for now, let's just purge this.


MaxNoe:
At least not in the LST muon code

MaxNoe:
Which I'm going to refactor as well along with this...

ctapipe/image/muon/fitting.py
```diff
@@ -65,394 +51,6 @@ def kundu_chaudhuri_circle_fit(x, y, weights):
...
-        mirror_radius,
-        lambda1,
-        lambda2,
-    )
-
-    result = light * pixel_diameter * norm.pdf(pixel_r, ring_radius, sigma_psf)
-    return result
-
-
-def efficiency_fit(
```
rlopezcoto, line 356: isn't this `efficiency_fit` the one used to calculate the `muon_efficiency` given as the output of the muon code?

MaxNoe: No, this is a different, in tested function doing the same thing.

kosack:
looks like you need to remove `vacanti94` from `bibliography.rst`. It's somewhat annoying that you need to use all references, or they cause an error, but I don't see a way to disable that. 

```Warning, treated as error:
/home/travis/build/cta-observatory/ctapipe/docs/bibliography.rst:13:Citation [vacanti94] is not referenced.
```

MaxNoe:
@kosack could you re-approve

PR #1256, opened 2020-04-06
Calculate ring containment analytically

MaxNoe:
This replaces a brute force scanning approach by an analytical calculation based on the "intersection of two circles"


moralejo:
Both the old and new approach assume a circular camera. Perhaps one could check actual pixels' positions (but this requires to go back to a scanning method I am afraid).

PR #1261, opened 2020-04-07
Muon intensity improvements

MaxNoe:
Starting to work on making the muon code work with the rest of ctapipe, e.g. instrument descriptions, components, etc.

* [X] Remove unit handling from likelihood functions
* [x] Define Components
* [x]  Add tests


ctapipe/image/muon/intensity_fitter.py
```diff
@@ -1,37 +1,525 @@
...
+
+    max_lambda = FloatTelescopeParameter(
+        help="Minimum wavelength for Cherenkov light in m", default_value=600e-9,
+    ).tag(config=True)
+
+    oversampling = IntTelescopeParameter(
+        help='Oversampling for the line integration', default_value=3
+    ).tag(config=True)
+
+    def fit(
```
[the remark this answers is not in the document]
MaxNoe: will do

ctapipe/image/muon/intensity_fitter.py
```diff
@@ -1,37 +1,525 @@
...
+    center_y,
+    radius,
+    ring_width,
+    pixel_x,
+    pixel_y,
+    pixel_diameter,
+    oversampling=3,
+    min_lambda=300e-9,
+    max_lambda=600e-9
+):
```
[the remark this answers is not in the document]
MaxNoe: good idea

ctapipe/image/muon/intensity_fitter.py
```diff
@@ -1,37 +1,525 @@
...
+    -------
+    ndarray: likelihood for each pixel
+
     """
 
-    def fit(self, x, y, charge, center_x, center_y, radius, times=None):
+    sq = 1 / np.sqrt(2 * np.pi * (ped**2 + pred * (1 + spe_width**2)))
+    diff = (image - pred)**2
+    denom = 2 * (ped**2 + pred * (1 + spe_width**2))
+    expo = np.exp(-diff / denom) + 1e-16  # add small epsilon to avoid nans
```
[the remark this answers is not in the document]
MaxNoe: I will refactor this likelihood a bit more. This needlessly computes first exp and then log. I should be able to directly write down the log likehood. 

Also I saw that this is basically copy paste from `ctapipe/image/pixel_likelihoods.py`

MaxNoe:
What's still missing compared to the hard coded values in `analyze_muon_event` before is the hole radii for the telescopes and also a nicer way to access the mirror radius than taking guessing it from the area would be nice.

kosack:
> What's still missing compared to the hard coded values in analyze_muon_event before is the hole radii for the telescopes and also a nicer way to access the mirror radius than taking guessing it from the area would be nice.

Is that info anywhere in the MCs? We could add it to the InstrumentDescription.  If not in the MCs, we could use an external table (better than hard-coded).  Eventually this will be solved by: #738, so we can have all configs for each "Prod" in a single place

MaxNoe:
> Is that info anywhere in the MCs?

Not to my knowledge in any of the simtel objects. The optics description is missing much info. I asked Konrad some time ago about a new object for that (individual mirror shapes and positions) I don't remember if he ever respondend.

ctapipe/image/muon/intensity_fitter.py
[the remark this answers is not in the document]
MaxNoe: I want to make this an exercise for my new master student (improvement of the likelihood calculation). So I would leave this untouched for now in this PR. Ok?

MaxNoe:
I will add the hole radii as a telescope parameter for now with the defaults as they were defined before

ctapipe/tools/muon_reconstruction.py
```diff
@@ -5,121 +5,176 @@
...
+        return cam_coords.transform_to(TelescopeFrame())
 
     def finish(self):
-        Provenance().add_output_file(self.outfile,
-                                     role='dl1.tel.evt.muon')
-        self.writer.close()
+        if self.outfile:
+            Provenance().add_output_file(
+                self.outfile,
+                role='dl1.tel.evt.muon',
```
kosack, line 169: This naming has now changed, and since this isn't in a DL1 file, maybe just set the role to "telescope muon analysis parameters". It's just some text to recall why the file was produced semantically, so so far there is no dictionary for what to put there.

ctapipe/tools/muon_reconstruction.py
```diff
@@ -5,121 +5,176 @@
...
+        ))
+        self.cleaning = self.add_component(
+            TailcutsImageCleaner(
+                parent=self,
+                subarray=self.source.subarray,
+            )
+        )
+        if self.outfile:
+            self.writer = self.add_component(HDF5TableWriter(
+                self.outfile, "muons", add_prefix=True
```
kosack, line 76: we could consider using the same hierarchy as the DL1 files and put the tables in the group `"dl1/event/telescope/parameters/muons"` instead, that might make merging files easier later (but pathnames inside a bit more complex!) That could come later though.

ctapipe/tools/muon_reconstruction.py
```diff
@@ -5,121 +5,176 @@
...
-                        ring_params = muon_evt['MuonRingParams'][idx]
-                        cam_id = str(event.inst.subarray.tel[tel_id].camera)
-                        self.num_muons_found[cam_id] += 1
-                        self.log.debug("INTENSITY: %s", intens_params)
-                        self.log.debug("RING: %s", ring_params)
-                        self.writer.write(table_name=cam_id,
-                                          containers=[intens_params,
-                                                      ring_params])
+        if self.outfile:
+            self.writer.write(f'tel_{tel_id}', [event_index, ring, result])
```
kosack, line 153: use `f'tel_{tel_id:03d}` so that the sort order is correct (tel_001, tel_002, ..tel_015, ... tel_152)

kosack:
Right now, with the vector columns added, I can no longer use `pd.read_hdf()`. Any idea of a way to make that work? 

```py3
>>> df = pd.read_hdf("muons.hdf5", "/muons/tel_1")

ValueError: Must pass 2-d input
```

ctapipe/tools/muon_reconstruction.py
```diff
@@ -5,121 +5,176 @@
...
+
+    classes = [
+        CameraCalibrator, EventSource, MuonRingFitter, MuonIntensityFitter,
+    ]
 
-    aliases = t.Dict({
+    aliases = {
         'input': 'EventSource.input_url',
-        'outfile': 'MuonDisplayerTool.outfile',
-        'display': 'MuonDisplayerTool.display',
```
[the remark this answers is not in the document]
MaxNoe: Yes, working on plots right now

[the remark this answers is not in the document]
MaxNoe: soirry, I'll put back the draft. I'm not finished here

ctapipe/image/muon/intensity_fitter.py
```diff
@@ -1,37 +1,528 @@
...
+
+        Returns
+        -------
+        MuonIntensityParameters
         """
-        pass
+        telescope = self.subarray.tel[tel_id]
+        if telescope.optics.num_mirrors != 1:
+            raise NotImplementedError(
+                f'Currently only single telescopes are supported in {self.__class__.__name__}'
```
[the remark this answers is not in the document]
MaxNoe: Good catch

ctapipe/tools/muon_reconstruction.py
```diff
@@ -5,121 +5,176 @@
...
+
+    def process_telescope_event(self, event_index, tel_id, dl1):
+        event_id = event_index.event_id
+
+        self.log.debug(f'Processing event {event_id}, telescope {tel_id}')
+        image = dl1.image
+        clean_mask = self.cleaning(tel_id, image)
+
+        if np.count_nonzero(clean_mask) <= self.min_pixels.tel[tel_id]:
+            self.log.info(
```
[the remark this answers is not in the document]
MaxNoe: agree

ctapipe/tools/muon_reconstruction.py
```diff
@@ -5,121 +5,176 @@
...
+                f'It {i}: r={ring.ring_radius:.2f}'
+                f', x={ring.ring_center_x:.2f}'
+                f', y={ring.ring_center_y:.2f}'
+            )
+
+            dist = np.sqrt((x - ring.ring_center_x)**2 + (y - ring.ring_center_y)**2)
+            mask = np.abs(dist - ring.ring_radius) / ring.ring_radius < 0.4
+
+        if np.count_nonzero(mask) <= self.min_pixels.tel[tel_id]:
+            self.log.info(
```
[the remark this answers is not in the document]
MaxNoe: agree

ctapipe/image/muon/intensity_fitter.py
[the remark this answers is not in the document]
kosack: Yes, that's fine

ctapipe/tools/muon_reconstruction.py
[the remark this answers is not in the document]
kosack: ah, ok - I had a request for review :) I'll hold off.

MaxNoe:
@kosack yes, that's a bit unfortunate with pandas, even if I specify column names of the 1d columns it  does not work.

This worked for me:
```
with h5py.File('muons.hdf5', 'r') as f:
    muons = f['muons/tel_1']
    df = pd.DataFrame({c: muons[c][:] for c in muons.dtype.names if muons[c].ndim == 1})"
```

kosack:
It seems to only be the 3D cov matrix column causing the problem, though.  So that could be dropped, or put in a separate table of the same length. It's probably only needed for debugging anyhow, so by default you could exclude that column, but have an option to enable it.

MaxNoe:
@kosack yes, but let's merge #1253 first, it also touches container fields of the muon analysis

ctapipe/tools/muon_reconstruction.py
MaxNoe: How would the path for mc information look like? ANd is there somewhere I can read it up?

kosack: There is discussion in #1066  and #1059 .  Right now the MC parameters are in `/dl1/event/subarray/mcshower` but since it's not really specific to DL1, we may move it elsewhere and probably rename to `true` instead of `mc`

MaxNoe: make an extra level for simulation values? `/sim/event/subarray/` `/sim/event/telescope/` ?

ctapipe/tools/muon_reconstruction.py
```diff
@@ -5,121 +5,177 @@
...
+
+        if tel_id not in self.pixels_in_tel_frame:
+            self.pixels_in_tel_frame[tel_id] = self.pixel_to_telescope_frame(tel_id)
+
+        pixel_coords = self.pixels_in_tel_frame[tel_id]
+        x = pixel_coords.delta_az
+        y = pixel_coords.delta_alt
+
+        mask = clean_mask
+        for i in range(3):
```
kosack, line 117: Is there a reason this is repeated 3 times? If so, please add a comment to explain it (otherwise it will be removed).

MaxNoe: What do you mean it will be removed?

It's an iterative fit procedure (the mask is updated with the result from the fit before)

MaxNoe: but yes, I can add a comment.

ctapipe/tools/muon_reconstruction.py
kosack: yes that was one of the other suggestions

ctapipe/tools/muon_reconstruction.py
kosack: I mean, if I say that,  would assume it's a mistake and remove it. Just adding a comment that it runs the fit multiple times to iterate is enough, I think.  It's not obvious that the loop does something different each time.

kosack: Might also comment why 3 times is a good enough number of iterations

MaxNoe: I did not evaluate this. This comes from allison's and ruben's codes. But looking at the iterations for some muon events, it does not change a lot between 2 and three. Better would be some kind of convergence criterium with max number of iterations. But that's is also something that could be addressed in another PR after this has a reasoable structure.

This PR is more about transforming the mess in `muon_reco_functions` into a nice tool, not improving the algorithms themselves.

ctapipe/tools/muon_reconstruction.py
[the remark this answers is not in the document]
MaxNoe: fixed the command line options, putting of plots for later

MaxNoe:
@cta-observatory/ctapipe-core-devs this is ready for review I'd say.

I'll postpone working on the algorithms and muon related image features with my master's student in more PRs. I'm currently processing three sets of LST MCs that will show how this works and abelardo @1261 can compare with his results on these datasets.

ctapipe/tools/muon_reconstruction.py
```diff
@@ -5,121 +5,223 @@
...
-    aliases = t.Dict({
+    aliases = {
+        'i': 'EventSource.input_url',
         'input': 'EventSource.input_url',
-        'outfile': 'MuonDisplayerTool.outfile',
-        'display': 'MuonDisplayerTool.display',
+        'o': 'MuonAnalysis.output',
+        'output': 'MuonAnalysis.output',
         'max_events': 'EventSource.max_events',
         'allowed_tels': 'EventSource.allowed_tels',
```
kosack, line 52: In other tools, I've used "max-events" and "allowed-tels" for these (following usual command-line arg conventions of not using underscores). But that's not critical. In the future we should unify all tool arguments, though.

ctapipe/tools/muon_reconstruction.py
```diff
@@ -5,121 +5,223 @@
...
+        )
 
-                self.log.info(
-                    "Event Number: %d, found %s muons",
-                    numev, dict(self.num_muons_found)
-                )
+    def get_fov(self, tel_id):
+        '''Guesstimate fov radius for telescope with id `tel_id`'''
+        # memoize fov calculation
+        if tel_id not in self.field_of_view:
```
kosack, line 192: Is there a reason not to just use the `@functools.lru_cache` decorator instead of an explicit memoization?  (same for other places where you cache the result).

ctapipe/tools/muon_reconstruction.py
```diff
@@ -5,121 +5,229 @@
...
+        for event in tqdm(self.source, desc='Processing events: '):
+            self.process_array_event(event)
+
+    def process_array_event(self, event):
+        self.calib(event)
+
+        for tel_id, dl1 in event.dl1.tel.items():
+            self.process_telescope_event(event.index, tel_id, dl1)
+
+        self.writer.write('sim/event/subarray', [event.index, event.mc])
```
kosack, line 104: should be `sim/event/subarray/shower` I think, since subarray is a group in the DL1 format, not a table.  But this can be unified later.

ctapipe/tools/muon_reconstruction.py
```diff
@@ -5,121 +5,229 @@
 The resulting output can be read e.g. using for example
 `pandas.read_hdf(filename, 'muons/LSTCam')`
```
kosack, line 6: Update this example or remove it (data has moved)

ctapipe/tools/muon_reconstruction.py
MaxNoe: ah yes, forgot the last name

MaxNoe:
@kosack I addressed your comments.

@moralejo @rlopezcoto I processed the 3 sets of muon simulations you provided:

Is that what you'd expect? I only made a cut of at least 100 Pixels after cleaning, no other checks on the rings.

efficiency
impact
radius
width

ctapipe/tools/muon_reconstruction.py
MaxNoe: I thought lru cache does not work for methods. Did I remember that wrong?

kosack: ah, yes that is a problem. Still, making it a decorator (needs to be a custom one like mentioned here https://stackoverflow.com/questions/33672412/python-functools-lru-cache-with-class-methods-release-object, might still be clearer.

moralejo:
> 
> @moralejo @rlopezcoto I processed the 3 sets of muon simulations you provided:
> 
> Is that what you'd expect? I only made a cut of at least 100 Pixels after cleaning, no other checks on the rings.
> 
> efficiency
> width

Hi,

apologies for slow reaction (due to Easter break). The muon efficiency should in principle scale with the simulated efficiency. Seems close, but not exact. However the current calculated efficiency, though roughly ok, has some not yet understood (at least by me) dependency with the fitted muon width. Might have to do with the approximations in the likelihood. 

Is this done with GlobalPeakWindowSum? That is to be preferred, because it is an unbiased extractor that does not "pick" NSB on empty pixels. With a biased extractor, in contrast, you will have a contribution from NSB-only pixels that will worsen the correlation between reconstructed efficiency and muon efficiency.

Also, all these comparisons are better made with rings in which ring_containment is 1 (say, >0.99), that is, rings not clipped by the FoV, because if a ring is clipped, the impact parameter determination, and hence the efficiency determination, are more uncertain.
Finally, the number of events seem to change a lot, which cannot be due to trigger, and so it must be due to ring selection.  I expect that difference will be much less noticeable once only fully contained rings are used.

MaxNoe:
> Finally, the number of events seem to change a lot, which cannot be due to trigger, and so it must be due to ring selection. I expect that difference will be much less noticeable once only fully contained rings are used.

Why shouldn't be this due to trigger or cleaning levels? These simulated datasets are essentially halving the amount of light. I would be very suprised if this had no impact on trigger rates / rates after image cleaning.

Also have a look at the impact distribution. You basically loose the high impact events as these are probably to dim  no to survive the cleaning.

I processed this with default settings (did not check what they are).

MaxNoe:
All these rings are contained @moralejo, I think these were simulated like this (max angle to optical axis)

rlopezcoto:
Hi, sorry that I also step into the conversation late. A couple of comments:
- Yes, muons were simulated so that they are contained in the dish
- it is also strange that the total number of events is so different... maybe the files are not complete?

MaxNoe:
> it is also strange that the total number of events is so different... maybe the files are not complete?

I processed the exact same number of array events (50k)

MaxNoe:
@kosack what is strange is that the TailCutsImageCleaner config values do not appear in `--help-all`. Why is that? I used `add_component`

MaxNoe:
@moralejo @rlopezcoto seems the cleaning levels are on the high side (10, 5) and that the image cleaning component does not allow to set the keep isolated pixels option

moralejo:
> Hi, sorry that I also step into the conversation late. A couple of comments:
> 
> * Yes, muons were simulated so that they are contained in the dish

You mean they all intersect the dish? Ok, but the algorithm does not know that, and if the ring ends sharply at the FoV edge (as some do, if they land close to the mirror edge) the likelihood lacks the info on the (lack of) signal on (non-existent) beyond-FoV pixels, and the fit is necessarily less precise. Besides, on real data we of course will need such a cut.

MaxNoe:
> You mean they all intersect the dish? 

That's two different things. Angle to optical axis makes containment, impact parameter makes uniformity / completeness. 

But I think these simulations were calculated in a way, that both things are so that only nice rings are produced.

MaxNoe:
Core range: 0 – 9.6 m
Zenith: 0°
diffuse: True
viewcone: 0.9°

So basically every muon hits the dish with low impact and so that it is contained in the camera.

moralejo:
> 
> Why shouldn't be this due to trigger or cleaning levels? These simulated datasets are essentially halving the amount of light. 

Rings are very bright compared to the trigger thresholds. Would be funny, and dangerous, if trigger alone has such a large effect, at least on the nice, quite complete rings we want to use for the calibration.

> I would be very suprised if this had no impact on trigger rates / rates after image cleaning.
> 

Cleaning is a more likely culprit, because you request a rather large number of pixels after it, and a factor two in light is a lot. You will make cleaned rings "shorter" when halving the light, and that will indeed kill more efficiently the larger impact parameters, consistent with your observation.

> Also have a look at the impact distribution. You basically loose the high impact events as these are probably to dim no to survive the cleaning.

MaxNoe:
> because you request a rather large number of pixels after it,

The current version in lstchain event requests 150 if I'm not mistaken.

moralejo:
> > You mean they all intersect the dish?
> 
> That's two different things. Angle to optical axis makes containment, impact parameter makes uniformity / completeness.
> 

Angle to optical axis determines ring center. Given that, how _complete_ the ring is, is given by impact parameter. If it hits at dish center, it is as complete as it can be. If it hits on the edge, it is less than half ring, even for a muon parallel to the telescope axis - in that case, the ring is incomplete, but _contained_, i.e. all pixels that lie along the ring exist, even though some or many of them have no signal. The fact that the algorithm has access to the pixels without signal makes the impact parameter determination more precise.
If, on the contrary, the ring is clipped, i.e. camera-edge ring pixels have large signal (due to the event's combination of direction & impact) the fit is more unreliable. For this MC, this will probably have little effect, because of the production limits, but there are no limits on real data.



What about signal extractor in your test? Are they done with GlobalPeakWindowSum?

MaxNoe:
> Angle to optical axis determines ring center. Given that, how complete the ring is, is given by impact parameter.

That's what I wrote, isn't it?

> What about signal extractor in your test? Are they done with GlobalPeakWindowSum?

No, they were done with the default extractor which is `NeighborPeakWindowSum`.

I added the option to choose a different image extractor in the latest commit, set the default to be `GlobalPeakWindowSum` and I am currently reprocessing the data.

> but there are no limits on real data.

Sure, but this script is, like all other's currently in ctapipe, not suited to process any real data due to the lack of handling calibration and so on.
As soon as the DL1 writer is finalized, it will probably makes sense to transform this script to read DL1 data and then it can also be expandend to do classification into muon rings / non muon rings.
Currently, this tool does not deal with selecting muon rings (aside from the `min_pixels` on the ring and the fact that the ring fit must have been successful)

MaxNoe:
@moralejo @rlopezcoto 

I reprocessed with the `GlobalPeakWindowSum` with default settings. Results did not change much.

Here is a scatter plot with linear regressions for width vs efficiency:

efficiency_vs_width

MaxNoe:
When looking at the standard deviation of the pixels not on the muon ring, it has std dev around 1.6, not 1.1. I'll reprocess with pedestal set to 1.1. Maybe this fixes the width depdendency (more width means more noise and when the pedestal value is wrong, this could explain the width dependency)

MaxNoe:
@kosack @moralejo so we just found out, that the file named OptivalEff0.5 actually was simulated with 0.6. So the results fit nicely (at least relative to each other):

efficiency

kosack:
> that the file named OptivalEff0.5 actually was simulated with 0.6. So the results fit nicely (at least relative to each other):

Just out of curiosity, is that in the MC info anywhere? Would be nice to not rely on filenames, which can be wrong.

rlopezcoto:
> Just out of curiosity, is that in the MC info anywhere? Would be nice to not rely on filenames, which can be wrong.

Yes, @MaxNoe wrote [this script](https://github.com/cta-observatory/pyeventio/pull/197) to read that. In any case, it was a problem of parsing by hand the optical efficiency, so it was also quickly spotted in the input scripts...

MaxNoe:
> Just out of curiosity, is that in the MC info anywhere? Would be nice to not rely on filenames, which can be wrong.
Unfortunately it's just in the config, not in any of the data fields.

ctapipe/tools/muon_reconstruction.py
MaxNoe: or wait for the time we use  python >= 3.8: https://docs.python.org/3/library/functools.html#functools.cached_property

moralejo:
> @kosack @moralejo so we just found out, that the file named OptivalEff0.5 actually was simulated with 0.6. So the results fit nicely (at least relative to each other):
> 
That is great, thanks!  

As for the dependency with width: using an unbiased extractor the result of the fit should not depend strongly on the assumed std dev of the pedestal. That would affect the Chi2 of the fit, but I do not think it would bias the best-fit width in any direction.

I think the dependency has to do with how the "predicted" light yield in a pixel is calculated. The prediction gives an amount of light per "azimuth" around the muon trajectory, which translates in amount of light per arc length. For a given pixel we multiply that prediction times the angle subtended by the pixel's whole width from the ring center, and then there is a factor which scales it according to where the pixel is (along the radius) relative to the ring, with a gaussian model of the ring radial profile (the std dev of which is the width in the fit). This is clearly an approximation - one would need to make instead a proper calculation of the "intersection" between the ring an the pixel edges.

MaxNoe:
> As for the dependency with width: using an unbiased extractor the result of the fit should not depend strongly on the assumed std dev of the pedestal. That would affect the Chi2 of the fit, but I do not think it would bias the best-fit width in any direction.

This is what I found, the results have not changed after I gave a higher pedestal.

watsonjj:
@moralejo @rlopezcoto Are you happy for this to be merged?

PR #1262, opened 2020-04-07
add option to use effective focal length, fixes #1223

### LINKED ISSUE #1223: Effective Focal Length
Currently in the transform from `CameraFrame` (meters on the camera plane) to `TelescopeFrame` (degrees on the sky, before refraction), we use the nominal `focal_length` as read from the SimTelarray files (actually we use what we call the "equivalent focal length", which is the nominal focal length taking into account the 1 or 2-mirror optics).

However, the correct focal length to use is the *effective focal length* measured using an optical point-source ray-traced from a distance.  The difference is something like 2% for MSTs, and 4% for LSTs (see table from Konrad Bernloehr). I believe using this focal length in the coordinate transform is equivalent to the "optical aberration correction" used in MARS (@moralejo is that correct? - there it uses the effective/nominal ratio to shrink the camera, but that's the same as just using the effective focal length in the coordinate transform)

Here is how I suggest how to proceed (let me know if there are better ideas):
1. produce a lookup table (just a text file in ECSV forma) for the effective focal length for each optics type, and put it in `ctapipe-extra` (so it can be modified, or replaced, for new MCs, or better measurements). It should not be hard-coded
2. either:
   1. in the `SimTelEventSource` check for this file, and replace `equivalent_focal_length` per telescope with the effective one if it exists, requiring no further change to the coordinate transform code.  
   1. add a field `effective_focal_length` to `OpticsDescription` (and fill it in SimTelEventSource), and then do the check when we do the coordinate transform of which one to pass to the `CameraFrame`.

kosack:
also ran "black" on the code (so ignore style changes)

The relevant code is just a new option `focal_length_choice` which can be (so far) "nominal" (default) or "effective".

Currently.  If the effective focal length is chosen and is not in the file, an exception is raised (to avoid getting something the user didn't' ask for). We could also change this to automatically cvall back to Nominal, but that might be error prone.   

A future enhancement could be to add another option "user", and also a config option called "user_focal_length" that would allow the user to set a focal length as a lookup table as a TelescopeParameter (need to think a bit how that would work, since the subarray is not yet constructed, however).


kosack:
see #1223

kosack:
Still need to add a test, using the example file in the eventio test resources: `pyeventio/tests/resources/prod4_pixelsettings_v3.gz`

MaxNoe:
Does switching the default make sense? If the effective focal length is the better one to use, shouldn't it be the default?

kosack:
> Does switching the default make sense? If the effective focal length is the better one to use, shouldn't it be the default?

The default up to now was to use the nominal, since  the effective focal length doesn't exist in most simulation files (for CTA, only prod4 and above, and it requires a special simulation run to add that information). So I think it's better to stick with the solution that works everywhere (nominal) as we do now.

PR #1264, opened 2020-04-07
Move container definition into image module

MaxNoe:
This solves the circular import Problem @HealthyPear is facing in #1215 .

As image imported these containers from io, it imported the whole simetelevent source which imports the calibration stack which imports something from image.

### DIFF
--- a/ctapipe/image/__init__.py
+++ b/ctapipe/image/__init__.py
@@ -6,4 +6,5 @@ from .reducer import *
 from .muon import *
 from .geometry_converter import *
 from .leakage import *
-from .concentration import concentration
+from .concentration import concentration, ConcentrationContainer
+from .containers import ImageParametersContainer, MorphologyContainer
--- a/ctapipe/image/concentration.py
+++ b/ctapipe/image/concentration.py
@@ -1,6 +1,22 @@
 import numpy as np
-from ..io.containers import ConcentrationContainer
+from numpy import nan
+
 from .hillas import camera_to_shower_coordinates
+from ..core import Container, Field
+
+
+class ConcentrationContainer(Container):
+    """
+    Concentrations are ratios between light amount
+    in certain areas of the image and the full image.
+    """
+
+    container_prefix = "concentration"
+    cog = Field(
+        nan, "Percentage of photo-electrons in the three pixels closest to the cog"
+    )
+    core = Field(nan, "Percentage of photo-electrons inside the hillas ellipse")
+    pixel = Field(nan, "Percentage of photo-electrons in the brightest pixel")
 
 
 def concentration(geom, image, hillas_parameters):
--- a/ctapipe/image/containers.py
+++ b/ctapipe/image/containers.py
@@ -1,0 +1,27 @@
+from numpy import nan
+from ..core import Container, Field
+from .hillas import HillasParametersContainer
+from .leakage import LeakageContainer
+from .timing_parameters import TimingParametersContainer
+from .concentration import ConcentrationContainer
+
+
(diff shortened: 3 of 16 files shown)


kosack:
with this cause (or solve) any circular imports?

MaxNoe:
@kosack solves, see new description.

kosack:
Ok, though now `reco` will depend on `image`, but that may have already been the case.

MaxNoe:
This is the chain:
```
  File "<string>", line 1, in <module>
  File "/home/maxnoe/Uni/CTA/ctapipe/ctapipe/image/__init__.py", line 1, in <module>
    from .hillas import *
  File "/home/maxnoe/Uni/CTA/ctapipe/ctapipe/image/hillas.py", line 11, in <module>
    from ..io.containers import HillasParametersContainer
  File "/home/maxnoe/Uni/CTA/ctapipe/ctapipe/io/__init__.py", line 8, in <module>
    from .simteleventsource import SimTelEventSource
  File "/home/maxnoe/Uni/CTA/ctapipe/ctapipe/io/simteleventsource.py", line 18, in <module>
    from ctapipe.calib.camera.gainselection import ThresholdGainSelector
  File "/home/maxnoe/Uni/CTA/ctapipe/ctapipe/calib/__init__.py", line 5, in <module>
    from .camera import *
  File "/home/maxnoe/Uni/CTA/ctapipe/ctapipe/calib/camera/__init__.py", line 6, in <module>
    from .calibrator import CameraCalibrator
  File "/home/maxnoe/Uni/CTA/ctapipe/ctapipe/calib/camera/calibrator.py", line 9, in <module>
    from ctapipe.image.extractor import NeighborPeakWindowSum
  File "/home/maxnoe/Uni/CTA/ctapipe/ctapipe/image/extractor.py", line 30, in <module>
    from ctapipe.image.timing_parameters import timing_parameters
  File "/home/maxnoe/Uni/CTA/ctapipe/ctapipe/image/timing_parameters.py", line 8, in <module>
    from .hillas import camera_to_shower_coordinates
```

MaxNoe:
@kosack I like the fact that they are defined where they are used / emitted

MaxNoe:
Ah dam, I introduced another circular import.

kosack:
> I like the fact that they are defined where they are used / emitted

I do to, but often  you don't need the algorithm to use the data, so if they are grouped, you entangle them

kosack:
You can check with something like:
```bash
pydeps --show-cycles --max-bacon=2 --cluster --keep-target-cluster -T png ctapipe
```

MaxNoe:
ctapipe

MaxNoe:
@kosack that does not really seem to work, as I back import the containers in io but it does not show a dependency from io to image

kosack:
> Or does it only show cycles?

I think with those options, it only shows cycles. If you show everything it's huge.

HealthyPear:
The errors in the log now seem to be different, they don't show anymore a circular import: there seems to be a problem with notebooks and the stub file from `HillasParametersContainer`

MaxNoe:
Ok, so I now moved all image related containers into the image submodule and removed them from io.

The only reason to remove them from io is that the docs create a warning, which is treated as error on travis. This is not good I think.

Also, does this create problems with the table reader? If the containers are not all in io/containers anymore?

Maybe better to just move the container definions into their own submodule? `ctapipe.containers`?

HealthyPear:
> Maybe better to just move the container definions into their own submodule? `ctapipe.containers`?

I would do this, it seems a cleaner approach

PR #1265, opened 2020-04-07
Black pre-commit hook and initial run on all files

kosack:
For discussion... This PR does 2 things:
1.  runs `black`  (https://github.com/psf/black) on all files, fixing most style errors and making the code style uniform
2. adds black as a GIT pre-commit hook (using the `pre-commit` package)

If the developer runs `pre-commit install`, afterward all commits will be automatically run through black to ensure common style.  

Although this initial commit will cause some confusion in the blame history (though it can be filtered out with `git blame --ignore-rev <rev>`), it will make future commits much easier to track, since slightly different developer styles will be unified. 

Any code that should not be re-formatted (where the formatting was specifically done by hand for some reason), can be protected with `# fmt: off` and `# fmt: on` comments.


MaxNoe:
If it would only use single quotes :see_no_evil:

kosack:
> If only it would use single quotes

Well, they are explicitly preferred in docstrings by PEP 257, but for other cases, I agree single quotes can be easier to type (on qwerty keyboards at least, not on others) . But with a pre-commit hook, it doesn't really matter :-) You can type what you like.

kosack:
I actually have an emacs package installed that chooses the correct quote based on context when I type a single quote, and it works pretty well. Nothing  like that for PyCharm, which I use more often however.

kosack:
similar move from gammapy: https://github.com/gammapy/gammapy/pull/1764

MaxNoe:
I fully agree it would be a good thing, but maybe we'll do this as last commit before the next release? So everyone can get a clean start or start over from the new release?
This would be hell for all open PRs. 

So let's go through the currently open PRs, see what is in shape for the next release and what needs more work and get the next release prepared quickly.

kosack:
Yes, I agree that this can be delayed until the next release

watsonjj:
> So let's go through the currently open PRs, see what is in shape for the next release and what needs more work and get the next release prepared quickly.

Should we create a project/milestone to organise what we want in the next release? Also perhaps delete the old ones that are now invalid

MaxNoe:
@watsonjj I already added a 0.8 milestone

kosack:
> Should we create a project/milestone to organise what we want in the next release? Also perhaps delete the old ones that are now invalid

Let's use a project board instead of a milestone - it's much more flexible

watsonjj:
Is it time to close this PR and open a fresh one with black applied?

kosack:
replaced by #1345

PR #1269, opened 2020-04-08
Ignore unmentioned references

MaxNoe:
These might in turn not actually be unmentioned, but inside functions not part of `__all__`.

So I think it's best to just filter these warnings.

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -40,17 +40,19 @@ needs_sphinx = '1.5'
 # Add any Sphinx extension module names here, as strings. They can be
 # extensions coming with Sphinx (named 'sphinx.ext.*') or your custom
 # ones.
-extensions = ['sphinx.ext.autodoc',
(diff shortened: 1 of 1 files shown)


watsonjj:
Is this in response to https://github.com/cta-observatory/ctapipe/pull/1254#issuecomment-610349573? It wasn't clear

kosack:
This might prevent typos in references from being checked though? But it would be nice to be able to have citations without an explicit reference.

MaxNoe:
> This might prevent typos in references from being checked though? But it would be nice to be able to have citations without an explicit reference.

Let me test this.

MaxNoe:
> This might prevent typos in references from being checked though? But it would be nice to be able to have citations without an explicit reference.

The problem is it even complains about citations as not being referenced when they are referenced by a method not included in `__all__`. And this is clearly not what we want. We also want to use references in internal methods.

PR #1276, opened 2020-04-13
Fix bug in ImageExtractor and WaveformModel concerning waveform sampling rate

watsonjj:
When checking the calibration for simtelarray files with different sampling rates, the result shows that the sampling rate is not currently correctly accounted for:

image

This result is what led me to create the WaveformModel in #1244, so I could properly diagnose this problem. However, in using the WaveformModel, I discovered I was not correctly integrating the charge when sampling the waveform from the convolved readout. As a result, the samples were not in p.e. as they should be.

Once this bug was fixed, and the tests for ImageExtractor were refactored to use WaveformModel (and cleaned up), I found the bug in ImageExtractor occurred in `extract_around_peak`. As the samples from simtelarray (and now WaveformModel, with this bugfix) are already in photoelectrons, they should not be divided by the sampling rate. The sampling rate is already accounted for during the waveform sampling... Instead one should just sum the waveform samples.

Once this change was made, the correct behaviour can be seen in the extracted charge for different sampling rates:

image

### DIFF
--- a/ctapipe/image/extractor.py
+++ b/ctapipe/image/extractor.py
@@ -129,7 +129,7 @@ def neighbor_average_waveform(waveforms, neighbors, lwt):
     """
     n_neighbors = neighbors.shape[0]
     sum_ = waveforms * lwt
-    n = np.zeros(waveforms.shape, dtype=np.int32)
+    n = np.zeros(waveforms.shape, dtype=np.int32) + lwt
     for i in prange(n_neighbors):
         pixel = neighbors[i, 0]
         neighbor = neighbors[i, 1]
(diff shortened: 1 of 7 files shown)


watsonjj:
Remaining codacy complaint is numba related

ctapipe/image/extractor.py
```diff
@@ -130,7 +129,7 @@ def neighbor_average_waveform(waveforms, neighbors, lwt):
     """
     n_neighbors = neighbors.shape[0]
     sum_ = waveforms * lwt
-    n = np.zeros(waveforms.shape, dtype=np.int32)
+    n = np.zeros(waveforms.shape, dtype=np.int32) + lwt
```
MaxNoe, line 132: `np.full(waveforms.shape, lwt, dtype=np.int32)` avoids the intermediate array.

PR #1283, opened 2020-04-15
Update documentation for CameraCalibrator and ImageExtractor

watsonjj:
Update the documentation to match the changes made during #1026

### DIFF
--- a/docs/ctapipe_api/calib/index_camera.rst
+++ b/docs/ctapipe_api/calib/index_camera.rst
@@ -21,8 +21,9 @@ CameraCalibrator
 
 The primary class in this module is the `CameraCalibrator`. This class handles
 two data level transition stages for the event:
-    * R1 -> DL0 (:ref:`image_reducers`)
-    * DL0 -> DL1 (:ref:`image_charge_extractors`)
+
+* R1 -> DL0 (:ref:`image_reducers`)
+* DL0 -> DL1 (:ref:`image_charge_extractors`)
 
 The class takes a ctapipe event container, and fills the
 next data level containers with the calibrated information.
--- a/docs/ctapipe_api/calib/index_camera.rst
+++ b/docs/ctapipe_api/calib/index_camera.rst
@@ -1,19 +1,23 @@
 .. _calib_camera:
 
-==================
-Camera calibration
-==================
+##################
+Camera Calibration
+##################
 
 .. currentmodule:: ctapipe.calib.camera
 
+
+************
 Introduction
-============
+************
 
 This module contains all the methods and classes that are used to calibrate the
 CTA Cameras (MC, prototypes and final camera calibration algorithms).
 
+
+****************
 CameraCalibrator
-==========
+****************
 
 The primary class in this module is the `CameraCalibrator`. This class handles
 two data level transition stages for the event:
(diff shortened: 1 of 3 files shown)


docs/ctapipe_api/calib/index_camera.rst
```diff
@@ -9,20 +9,19 @@ Camera calibration
...
-next data level container with the calibrated information.
-If the container is empty for the source data level then that calibration is
-skipped and the target data level container is unchanged (for example
-performing the calibration in r1.py on data read into the R1 container will
-leave the container unchanged, as there is no data in the R0 container)
+This module contains all the methods and classes that are used to calibrate the
+CTA Cameras (MC, prototypes and final camera calibration algorithms).
+
+CameraCalibrator
+==========
```
kosack, line 16: I think this underline is what is stopping the doc build (just needs to be the right length).  Also, it's at the same heading level (`===`) as the title "Charge Extraction".  If it's supposed to be 1-level lower, use `---` (see here https://documentation-style-guide-sphinx.readthedocs.io/en/latest/style-guide.html)

docs/ctapipe_api/image/extractor.rst
```diff
@@ -1,10 +1,75 @@
...
+
+
+***********************
+Units and Normalisation
+***********************
+
+An `ImageExtractor` should preserve the units of the waveform samples when
+extracting the charge. I.e. if the samples are already calibrated into
+photoelectrons, then the charge reported by the `ImageExtractor` should be in
+photoelectrons. The simple `ImageExtractors` (e.g. `LocalPeakWindowSum`)
```
MaxNoe, line 65: This is a very general comment but it seems to apply here.

How can samples be photo electrons? The conversion from the sum or integral of the waveforms is always dependent on the method used and its settings.

There needs to be a conversion from waveform integral / sum or whatever the extractor calculates (could also just use the maxium or width or whatever) to  photo electrons.

Shouldn't we just stop stalking about a waveform being in units of photo electrons?
Or at least have a well defined definition of what that means?

watsonjj: In the SimtelEventSource, the dc_to_pe factor is applied to the samples.

If the waveform from simtelarray only contained an X p.e. signal, no noise or NSB, then the sum of all the samples would equal X. Therefore technically the waveform samples are in units of photoelectrons.

In the case of a window summing a subset of the signal pulse, the integration_correction will scale the summed samples by the amount of signal not included in the sum, to ensure the same answer is still returned. Therefore ensuring the units of the charge are still in p.e.

I accept that the same may not be true from a different EventSource - the units may not be in units of photoelectrons. In this case, whatever units the waveform samples are provided in, then the resulting extracted charge will have the same units. A further calibration may be needed to convert the charge to photoelectrons (which is possible in the CameraCalibrator). 

I have tried to convey this in the documentation, giving the simtelarray -> photoelectrons case as an example.

watsonjj: There is no integral over time performed in the charge extraction. The "sampling" of the waveform has already performed this integral of the continuous readout into the waveform samples.

kosack: I suppose it's not really true for FlashADCs (they do not integrate, but really sample, and FlashCam has a non-linear amplifier, so it's not PE).

PR #1284, opened 2020-04-18
Path improvements

MaxNoe:
This fixes a few issues concerning path traitlets and arguments

* traits.Path now uses pathlib.Path internally

* Path supports urls of the form `file://`, this is to make the `input_url` argument of the eventsource really an url (only file supported for now, but in the future we could think about allowing `http` to get a file directly from the web)

* Changed a lot of tool traitlets from Unicode to Path where appropriate


MaxNoe:
@kosack @watsonjj I fixed the codacy issues

PR #1285, opened 2020-04-20
Store filters and use for new datasets in HDFWriter

MaxNoe:
This allows passing kwargs to `tables.File.create_table` in `HDFTableWriter.write` to be able to set compression per table, not just globally per file.

This came up in https://github.com/cta-observatory/cta-lstchain/pull/353.

### DIFF
--- a/ctapipe/io/hdf5tableio.py
+++ b/ctapipe/io/hdf5tableio.py
@@ -32,6 +32,13 @@ PYTABLES_TYPE_MAP = {
 }
 
 
+DEFAULT_FILTERS = tables.Filters(
+    complevel=5,           # compression medium, tradeoff between speed and compression
+    complib='blosc:zstd',  # use modern zstd algorithm
+    fletcher32=True,       # add checksums to data chunks
+)
+
+
 class HDF5TableWriter(TableWriter):
     """
     A very basic table writer that can take a container (or more than one)
@@ -74,11 +81,11 @@ class HDF5TableWriter(TableWriter):
         'a' if you want to append data to the file
     root_uep : str
         root location of the `group_name`
+    filters: pytables.Filters
+        A set of filters (compression settings) to be used for
+        all datasets created by this writer.
     kwargs:
         any other arguments that will be passed through to `pytables.open()`.
-        e.g. to set the compression level to 7 pass : `filters=tables.Filters(
-        complevel=7)`
-
     """
 
     def __init__(
@@ -88,6 +95,7 @@ class HDF5TableWriter(TableWriter):
         add_prefix=False,
         mode="w",
         root_uep="/",
+        filters=DEFAULT_FILTERS,
         **kwargs,
     ):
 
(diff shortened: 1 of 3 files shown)


vuillaut:
Ha nice, I was looking into this also.

MaxNoe:
@vuillaut 

>  If the table already exists and some kwargs are passed, the kwargs won't be taken into consideration without the user noticing. Is this correct?

Yes, but I don't really see a way around it as warning if the table already exists would be raised for each subsequent write in an event loop.

One could check if the kwargs given are compatible with what is already in the file, but that would mean adding checks for specific kwargs, at this is not easy to achieve in general.

MaxNoe:
I fixed the codacy issue, but codacy did not update

vuillaut:
> @vuillaut
> 
> > If the table already exists and some kwargs are passed, the kwargs won't be taken into consideration without the user noticing. Is this correct?
> 
> Yes, but I don't really see a way around it as warning if the table already exists would be raised for each subsequent write in an event loop.
> 
> One could check if the kwargs given are compatible with what is already in the file, but that would mean adding checks for specific kwargs, at this is not easy to achieve in general.

Would it be possible to update the kwargs of the table (effectively rewriting the existing table)?

MaxNoe:
> Would it be possible to update the kwargs of the table (effectively rewriting the existing table)?

I'm a bit reluctant as this could be a potentially very expensive operation (imagine appending a single row to a huge table).

This should be done explicitly I think.

ctapipe/io/hdf5tableio.py
```diff
@@ -270,12 +271,17 @@ def write(self, table_name, containers):
             name of table to write to
         containers: `ctapipe.core.Container` or `Iterable[ctapipe.core.Container]`
             container to write
+        **kwargs
+            If the table is not yet created, these kwargs are passed
+            to ``create_table`` of the hdf5 file.
```
vuillaut, line 276: Maybe then explicitly state "else the schema of the existing table is used"

kosack:
I want to try this with the DL1 writer first - I had originally implemented it this way, I think , but ran into problems writing other data like the instrument config, etc, if the filters changed (and also reading the data in some HDF5 viewers).  For that reason, I forced all datasets to use the same filters. But it may have been a mistake.  

Can you explain the use case? Why not just enable compression everywhere?

MaxNoe:
> Can you explain the use case? Why not just enable compression everywhere?

`astropy.tables` uses `h5py`, not `tables`. The compressions they support out of the box are not guaranteed to be the same.

In the case for lstchain, the small instrument definition was written first with `astropy.tables`,
so the file was created by h5py and thus not set up with the `tables.Filters` used for the writer.

kosack:
> In the case for lstchain, the small instrument definition was written first with astropy.tables,
so the file was created by h5py and thus not set up with the tables.Filters used for the writer.

I do the same in the stage1 tool, and it seems to work. I think i did have to think a bit about the order of opening the files, if I recall, but I didn't run into any big problem.

kosack:
I just want to understand why this is absolutely needed: 

I wanted to avoid passing any implementation-specific options to `write()`, so that if we implement another backend (FITSTableWriter, etc), things work without modification to anything other than the constructor.

MaxNoe:
> I do the same in the stage1 tool, and it seems to work. I think i did have to think a bit about the order of opening the files, if I recall, but I didn't run into any big problem.

@rlopezcoto and @FrancaCassol encountered problems with file locking on macOS when the hdfwriter was already open and then using `astropy.tables` to save the instrument config. We couldn't find a way to write the instrument config after opening the writer (so the filters are set correctly).

I wasn't able to reproduce the problem on linux (also worked on travis).

MaxNoe:
@kosack we could also store the filters on the writer and use them later. That would be better than giving it to the write method, right?

MaxNoe:
@kosack @vuillaut I changed the code to get the filters in `__init__`, store them, and apply to each new created dataset and don't rely on the default filters in the file.

kosack:
By the way, the problem with having an HDF5TableWriter open and then trying to write with Astropy can just be worked around by writing the configuration tables from astropy *after* the writer is closed. I'd guess it's not a great idea to have two libraries (pytables and h5py) writing at the same time to the same open file anyhow, just in case there are some conflicts in how they operate even if they use the same underlying c library.

PR #1294, opened 2020-04-27
Remove deprecated instrument from event container

MaxNoe:
Removed the `inst` container from the event.

Added subarray attributes / arguments where appropriate.

The only thing failing currently is the muon code that get's removed in #1261, so as soon as that is merged, everything here should be passing.


watsonjj:
Thanks for doing this Max! I'll approve as soon as tests pass

MaxNoe:
Ok, this should be passing now.

PR #1296, opened 2020-04-27
Use pointing container in examples and docs

MaxNoe:
Some examples were still using the values from `event.mc` and not `event.pointing`.

Also, interestingly, `event.pointing` is missing the `.tel`. Should we add that to make it consistent?


kosack:
> Also, interestingly, event.pointing is missing the .tel. Should we add that to make it consistent?

Yes, I noticed that as well - might as well make the hierarchy more consistent (even though it then introduces another class with basically nothing in it, but that can be fixed if we clean up the full structure as in #1165 )

PR #1298, opened 2020-04-27
Toy source

MaxNoe:
Fix broken and untested toy source.

This also adds a method to guess the camera radius:

chec
flashcam
lst

### DIFF
--- a/ctapipe/io/toymodel.py
+++ b/ctapipe/io/toymodel.py
@@ -5,95 +5,132 @@ Create a toymodel event stream of array events
 import logging
 
 import numpy as np
-from ctapipe.image import toymodel
-from scipy.stats import norm
 import astropy.units as u
 
-from .containers import DataContainer
+from ..containers import (
+    DataContainer, DL1CameraContainer, EventIndexContainer,
+)
+from ..core import traits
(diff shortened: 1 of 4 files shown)


ctapipe/io/toymodel.py
```diff
@@ -5,95 +5,132 @@
...
-        container.count = event_id
-
-        # handle single-telescope case (ignore others:
-        if single_tel:
-            if single_tel not in container.r0.tels_with_data:
+class ToyEventSource(EventSource, TelescopeComponent):
+
+    trigger_probability = traits.FloatTelescopeParameter(
+        default_value=0.5, help='Probability that the telescope has an event',
+    )
```
kosack, line 25: I think all of these need the usual `.tag(config=True)` for them to be config parameters that work properly.

MaxNoe: Ah, of course. That's really annoying

kosack: yes, it's one of the reasons I want to get rid of that config system. I had a hand-made version that does a large fraction of the functionality with only traitlets, and not traitlets.config, but there are still some nice things in there that are hard to reproduce.

PR #1299, opened 2020-04-28
Rename coordinates of NominalFrame and TelescopeFrame, fixes #1260

### LINKED ISSUE #1260: Misleading names of coordinates in Telescope and Nominal Frame
@moralejo made a good point that `delta_alt` and `delta_az` imply that `delta_alt = altaz1.alt - altaz2.alt`.

Which is not the cases, as these are in a rotated coordinate frame and this holds only for the center point.

We should come up with more appropriate names or just go back to x and y and then make sure that x and y are aligned with the axes in the CameraFrame.

Opinions?

MaxNoe:
I went with `fov_lon` and `fov_lat`, as this is close to what is used by the open gamma ray astro data formats.


MaxNoe:
Btw. we could also remove the requirement that pointing or origin is in `AltAz` and also allow `ICRS`, than we would have a telescope/nominal frame aligned with the equatorial system, as it is also forseen in the gamma-astro-data format.

ctapipe/coordinates/nominal_frame.py
```diff
@@ -50,8 +50,8 @@ class NominalFrame(BaseCoordinateFrame):
     '''
     frame_specific_representation_info = {
         UnitSphericalRepresentation: [
-            RepresentationMapping('lon', 'delta_az'),
-            RepresentationMapping('lat', 'delta_alt'),
+            RepresentationMapping('lon', 'fov_lon'),
+            RepresentationMapping('lat', 'fov_lat'),
```
rlopezcoto, line 54: what is the reasoning behind `fov_lon` and `fov_lat`?

rlopezcoto: ok, just found the discussion in #1260

PR #1301, opened 2020-04-28
Container Improvements and Cleanup for DL1

kosack:
This is just a small first pass of cleaning up the containers *relating to  DL1*.  I don't intend this to be the full refactoring of all containers yet, I'm just doing it in small stages.    So far, this just removes the need for defining new Containers to make a DL1 dumper (see #1066).  For future proposed changes see #1165 and #1224 

## Container Changes:
- `DL1CameraContainer`:
  - renamed `pulse_time` → `peak_time` in anticipation of other pulse times (like `rise_time`, etc)
  - added `selected_gain_channel` (also to DL0 container), since this is needed
  - added image `parameters` as a sub-container
  - added `image_mask` to store the cleaning mask (still need to think what to do in case of non-mask based denoising, but that can come later)
- Created an `MCDL1CameraContainer` that extends this container to add:
  - `true_image` (photo electrons from MC without noise)
  - `true_parameters` (set of parameters computed from the true_image)
  - note this might not need a new container, it could just be an MC version of the `DL1CameraContainer` with the prefix set to *"true_"*, though that contains some fields that are maybe not available in the MCs
- `EventIndexContainer`:
  - removed deprecated fields for event_id, obs_id, and event_type elsewhere
  - make sure everything uses it
  - added event_type and made it an Enum (was a string, which doesn't store well)

## Features
- `Fields` now have optional `dtype`, `ndim`, and `allow_none` metadata attributes
 - `Field` and `Container` now have `.validate()` methods that check the values against the given metadata, if specified

## things that are not so nice still:
- `MCDL1CameraContainer` and `MCCameraEventContainer` should be merged / cleaned up
   - right now `MCEventContainer` is shower info (could rename `MCShowerContainer` instead)
   - the telescope-wise MC info in `MCCameraEventContainer`  is not really event-wise info: it contains pointing info (that could be removed), and calibration coefficients that do not change per event.  For this reason I do not want to write this container to DL1, since it adds a lot of "images" that do not change.  I think the calibration info should move either to  be under the the MCHeaderContainer (assuming it does not change during an observation file), or otherwise it could be migrated to the real Calibration containers. 
- I'd like to drop "MC" jargon and replace it with "Simulation" for containers, and "true_X" for parameters if possible. 

There are lots of other general issues, but those can be discussed in the issues linked above. 

download-1

for reference, here is the MCEventContainer and MCCameraEventContainer(in the latter,  I think nearly everything could be moved elsewhere, and the former renamed to "SimulatedShowerContainer")
download-2


MaxNoe:
> I'd like to drop "MC" jargon and replace it with "Simulation" for containers, and "true_X" for parameters if possible.

Whole heartedly agree

ctapipe/containers.py
```diff
@@ -47,16 +47,25 @@
...
     "SimulatedShowerDistribution",
+    "EventType",
 ]
 
 
+class EventType(enum.Enum):
+    """ note these will evolve once CTA has standardized them """
+
+    data = 0  # standard physics event
+    calibration = 1  # event from calibration system
```
MaxNoe, line 58: Can you directly add values for `Dark`, `Pedestal` and `LaserCalibration`? these are in simtel and are available in the new eventio version

kosack: yes, that would be great.  For old data, it's still just called "calibration" right?  Are the new emitted by pyeventio also upper-case like "LaserCalibration"? 


I'm also trying to figure out if we can understand if there are CTA-approved integers to use for these.

MaxNoe: There is an integer type now.

MaxNoe: And this works for all data that has those events, not only for newly processed data. This ID is there since a long time, iot what just not passed on by the `eventio.SimTelFile`.

MaxNoe: @param type  Dark (0), pedestal (1), in-lid LED (2) or laser/LED (3+) data.

kosack: So that's a second integer? E.g. a sub-type of the event_Type?

kosack: Right now I get it as `array_event["type"]` in SimTelEventSource, and that seems to return a string. Has that changed?

MaxNoe: No, but there is an additional `array_event['calibration_type']` for calibration events

kosack: we could for now use 10+calibration_type for the event_type of calibration events, so then you get 0=data, 10,11,12... , rather than adding another field, though another field is also ok

MaxNoe: You could simply enumerate the types and we can fill it appropriately in the source.

kosack:
I've now gone a bit further and fixed #1224 by adding Field and Container validation.  You can now also optionally specify `ndims` and `dtype` of Fields, and they can be checked.  So far this is not used yet in the `TableWriter` but could be.

ctapipe/containers.py
MaxNoe: I think if we wanted to support the types known to simtel, this should be:

```
data = 0
dark = 1
pedestal = 2
internal_led = 3
external_led_or_laser = 4  # not sure about this one, but I guess external LED / Laser is used for flat fielding? So maybe just call it flatfielding?

Comment (kosack) [reply ctapipe/containers.py:58]: I'll map these to the ones in the latest (internal so far) document from CTA, where there are something like 10 event/trigger types defined. They are a bit oddly numbered, so may change, but at least it's closer to what will be the standard.

Comment (kosack) [reply ctapipe/containers.py:58]: can you tell me what are the possible values of `array_event['calibration_type']`?

Comment (MaxNoe) [reply ctapipe/containers.py:58]: I gave them above:
from io_hess.h:
```
@param type Dark (0), pedestal (1), in-lid LED (2) or laser/LED (3+) data.
```

Comment (kosack) [reply ctapipe/containers.py:58]: Ah, I didn't realize that was the actual values. Thanks

Comment (MaxNoe) [reply ctapipe/containers.py:58]: Don't know though what the `3+` is supposed to mean.

File: ctapipe/core/container.py [hunk 2]
```diff
@@ -18,33 +24,111 @@ class Field:
         Help text associated with the item
     unit: `astropy.units.Quantity`
         unit to convert to when writing output, or None for no conversion
-    ucd: str
+    ucd: str or astropy.unit
```
MaxNoe, line 27: ucd should not be a unit, right?

ctapipe/core/container.py
```diff
@@ -18,33 +24,111 @@ class Field:
         Help text associated with the item
     unit: `astropy.units.Quantity`
```
MaxNoe, line 25: unit should not be a quantity

kosack: good point.What's a bit annoying is that:
```py3
unit = u.m
isinstance(unit, u.Unit) # → False
type(unit) # →  astropy.units.core.IrreducibleUnit
```
I'm not sure the best way to specify the type of an astropy  unit... Perhaps `Union[u.core.CompoundUnit, u.core.IrriducibleUnit, str]`, but that's a bit ugly

MaxNoe: it's `u.UnitBase`, see: https://github.com/astropy/astropy/issues/6517

kosack: I guess `u.core.UnitBase` works

kosack: Eventually would be nice to use something like `mypy`, but that's going to need a lot of work since these types hints are a bit lazy

Issue #1019, opened 2019-03-20
Take pixel size from simtel, do not guess it


maxnoe:
Right now, we guess the pixel sizes from the given area.

Actually they are in the simtel file, we should just take those values.

```
In [1]: from eventio import SimTelFile                                                         

In [2]: f = SimTelFile('../data/chec.simtel.gz')                                               

In [3]: cam = f.telescope_descriptions[1]['camera_settings']['pixel_size']
                                  
```

dneise:
just for my curiosity, is it clear what size means? Like inner circle radius?

dneise:
But I agree we should avoid guessing.

MaxNoe:
> just for my curiosity, is it clear what size means? Like inner circle radius?

According to the comment it is the diameter flat side to flat side for hexagons and edge size for square.

maxnoe:
As discussed in the PR, we are not really guessing but *calculating* from well defined inputs to well-defined outputs. So no reason to store duplicated info.

PR #1302, opened 2020-04-28
Add pix_size to CameraGeometry, read in simtel eventsource

MaxNoe:
- [ ] IO
- [ ] Use it where appropriate

### DIFF
--- a/ctapipe/instrument/camera/geometry.py
+++ b/ctapipe/instrument/camera/geometry.py
@@ -13,7 +13,7 @@ from scipy.spatial import cKDTree as KDTree
 from scipy.sparse import lil_matrix, csr_matrix
 import warnings
 
-from ctapipe.utils import get_table_dataset, find_all_matching_datasets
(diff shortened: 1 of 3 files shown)


kosack:
Can you explain why this is needed?  It's easy to calculate from the pix_area and pix_type

MaxNoe:
I just went through the open issues: #1019 

Why not set it directly from the simtelfile?

kosack:
> Why not set it directly from the simtelfile?

We only use the pix_size for plotting, so it was just to try to limit the number of things to keep track of in a camera definition file.   Also, the size of the pixel as a diameter has little meaning in reality, since there are Winston cones which means the sensitive area of a "pixel" is a hexagon (for hex cameras), and can be larger than the size of the phototube. For a hexagon, what is the use of defining a diameter? An area gives you the same information, and more accurately.

MaxNoe:
ok, I see it's redundant and area more what's needed

PR #1303, opened 2020-04-28
Add array pointing to pointing, use .tel map

MaxNoe:
* As discussed in #1296 , adds `.tel` accessor for the telescope pointings
* Also adds the array pointing


MaxNoe:
@kosack @vuillaut had to merge, sorry, please review again

PR #1308, opened 2020-05-06
Hdf5 column order

MaxNoe:
This makes sure columns are written in the order they appear in the containers written, not sorted alphabetically

### DIFF
--- a/ctapipe/io/hdf5tableio.py
+++ b/ctapipe/io/hdf5tableio.py
@@ -144,6 +144,7 @@ class HDF5TableWriter(TableWriter):
         meta = {}  # any extra meta-data generated here (like units, etc)
 
         # create pytables schema description for the given container
+        pos = 0
         for container in containers:
             for col_name, value in container.items(add_prefix=self.add_prefix):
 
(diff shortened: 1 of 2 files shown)


ctapipe/io/hdf5tableio.py
```diff
@@ -440,7 +454,7 @@ def read(self, table_name: str, container: Container):
 
 
 def tr_convert_and_strip_unit(quantity, unit):
-    return quantity.to(unit).value
+    return quantity.to_value(unit)
```
vuillaut, line 457: This function has no added value anymore, wouldn't it be better to just replace `tr_convert_and_strip_unit` everywhere by `quantity.to_value(unit)`?

MaxNoe: No, that's not possible. This function is used for the column transforms. 
And that must be callable as `func(col)`. Or you need to introduce a special case for quantities. But I'd rather have a very simple function than a special case in the application of the transforms.

MaxNoe: You could make it a lambda like for the case where no unit is given, but hey, it doesn't harm anyone. And if you inspect the column transforms, it is readable as opposed to `lambda at <hex adress>`

Issue #1097, opened 2019-06-25
How redirect to a file the logging messages of a tool?


FrancaCassol:
Hi,

I would like to redirect to a log file the logging messages of a ctapipe tool (avoiding the unix redirection), which is the best way of doing it? 

Thanks

MaxNoe:
add something like 

```
import logging

log = logging.getLogger('ctapipe')
logfile = logging.FileHandler('/path/to/togfile')
log.addHandler(logfile)
```

to your script.

kosack:
probably we should put an option to do that in the `Tool` base class as well, so that all tools can do it.

kosack:
I haven't tried it but you might have to remove the default StreamHandler if you don't want stdout *as well as the file* in @MaxNoe's example above (assuming addHandler() doesn't replace the old one by default).  Otherwise, you get it both to stdout and to the file you specified.

kosack:
If Max's suggestion works, can we close this?

FrancaCassol:
Hi @kosack,

Yes, it works but I do not menage to eliminate the stdout. I did this but it seems not to work, do you have other suggestions?

```
            log.removeHandler(logging.StreamHandler(sys.stdout))
            log.removeHandler(logging.StreamHandler(sys.stderr))

```
Also the output to file has some problem with the fonts:

[1;32mINFO[0m [CalibrationHDF5Writer] (tool/initialize): ctapipe version 0.6.2.post161+gitc154718
[1;32mINFO[0m [CalibrationHDF5Writer] (tool/run): Starting: CalibrationHDF5Writer
[1;32mINFO[0m [CalibrationHDF5Writer.FlasherFlatFieldCalculator] (flatfield/__init__): extractor <ctapipe.image.extractor.LocalPeakWindowSum object at 0x7f4dbc180f60>

MaxNoe:
This is not a Problem with fonts, this includes color escape sequences.
We should set a formatter that does not include them.

For the moment you can look into the files with `less -R` which will show the colors correctly.

The code you showed cannot work, because you are trying to remove a freshly created handler.
I managed to silence the logging output do console by using:

```
h = logging.StreamHandler()
logging.addHandler(h)
h.close()
```

FrancaCassol:
Clever idea, but it doesn't work for me :-( 
Inside my tool I do:
```
    def setup(self):
        kwargs = dict(parent=self)

        if self.log_file!='None':
[...]

PR #1310, opened 2020-05-07
Add --quiet/-q option and --logfile

nbiederbeck:
from #1097: 
> We could add a `-q, --quiet` option to silence the StreamHandler and a `--logfile` option for writing a logfile to every tool.

For now this draft PR only adds the `-q/--quiet` option, which sets `logLevel = 60`, so nothing gets logged.


MaxNoe:
Actually, when writing that issue, I didn't realize that traitlets.Application already gives us already a `--log_level` option. 

I also think quiet should only apply to the stream handler, not all handlers (you probably want full file log even if you use `-q`)

nbiederbeck:
Yes this is a question I just stumbled upon:

if you use `--logfile` do you want to have terminal output or not?

MaxNoe:
> if you use --logfile do you want to have terminal output or not? 

By default: yes. But I would say `-q` and `--logfile` is a pretty natural combination.

nbiederbeck:
I could not change the colored output in the logfile. For me it seems that the `log_format` gets overwritten somewhere (internally?) with the `ColoredFormatter` format, thus using `self.log_format` yields the colored format

kosack:
You could make "--quiet" only set the log level for the console logger, and not the file logger.  Or in more general, we could add a --file-log-level option, so you have more control (the uyser can then set `--log-level=WARNING --file-log-level=DEBUG`

That gives the ability to set both independently.  In principle that can be achieved with a logging config file (As supported in the standard library), but not sure we properly support that yet - with one of those, you can have fine-grained control of all loggers, even specific module-level ones., so can for example turn the debug logging on only for hillas parameterization, or something.

kosack:
One question: what is the advantage of this in the current implementation over using `--log-level=WARNING`? (or higher?).  I would guess we should never go above WARNING, otherwise important info might be lost.

MaxNoe:
@kosack I think when I proposed `-q` and `-v` in the issue, I didn't know about `--log-level`.

@nbiederbeck Could you update this, to implement a general way to configure logging, using e.g. the logging configuration files as described here:

https://docs.python.org/3/howto/logging.html#configuring-logging

nbiederbeck:
> You could make "--quiet" only set the log level for the console logger, and not the file logger

This is what happens, it set the "default" logger loglevel to 60, so it logs nothing. Afterwards I add the FileHandler if necessary with given loglevel.
This way `-q` is short for `--log-level=60`.

nbiederbeck:
> https://docs.python.org/3/howto/logging.html#configuring-logging

I'll look into that

dneise:
> Colored logfiles are okay with less -R and tail, but annoying in text editors

+1

MaxNoe:
Yes, I think we need to revisit this color logger in general, it prevented me from changing the format of the level.

kosack:
> Yes, I think we need to revisit this color logger in general, it prevented me from changing the format of the level.

There are fancier systems like "rich" (https://github.com/willmcgugan/rich), but I think we can just fix the color logger we use if there is a problem. It's nice to be able to identify warnings visually, and if logged to a file, we don't use it anyhow.

nbiederbeck:
I added `--log-level` and  `--log-file-level`, which both take the level numbers and names (eg. 10/DEBUG)

If supplied `--log-file` the console output is untouched and additional logging to file happens

ctapipe/core/tool.py
```diff
@@ -109,21 +109,33 @@ def main():
             "command-line parameters"
         ),
     ).tag(config=True)
-    log_format = Unicode(
-        "%(levelname)s [%(name)s] (%(module)s/%(funcName)s): %(message)s",
-        help="The Logging format template",
-    ).tag(config=True)
 
-    _log_formatter_cls = ColoredFormatter
+    log_level = Enum(
```
MaxNoe, line 113: I think this is already there in the base class, no?

nbiederbeck: ah, yes it is

ctapipe/core/tool.py
```diff
@@ -138,11 +150,22 @@ def initialize(self, argv=None):
...
         self.update_config(self.cli_config)
 
+        log_config = create_log_config(
+            name=self.name,
+            log_level=self.log_level,
+            log_file=self.log_file,
+            log_file_level=self.log_file_level,
+        )
+
+        logging.config.dictConfig(log_config)
```
MaxNoe, line 164: can we add a json or yaml file for the log config and only add the default config if none was given?

ctapipe/core/logging.py
```diff
@@ -1,39 +1,110 @@
-""" helpers for better logging """
+"""Helpers for better logging."""
 
 import logging
 
+DEFAULT_LOGGING_FORMAT = "%(levelname)s [%(name)s] (%(module)s/%(funcName)s): %(message)s"
```
MaxNoe, line 5: Can we add `%(asctime)s` as the first here?

MaxNoe: Also a dot between module and func name would make more sense

ctapipe/core/tool.py
nbiederbeck: to use with e.g. `--log-config`? I initially decided against this, because the logger options can be set with `--config`

MaxNoe: Can they? Like configuring each logger? So can I set `ctapipe.io` to DEBUG while the rest is `WARNING`? That's possible with the logging config files.

nbiederbeck: Ah I understand now the direction. Yes I think that's not possible with the config file and would be easier/better using logconfig files

nbiederbeck:
Setup logging via file with `--log-config`, and a configuration schema in yaml based on https://docs.python.org/3/library/logging.config.html#logging-config-dictschema

*or*

Use default logging and potentially set `--log-level`, `--log-file`, `--log-file-level`

nbiederbeck:
I added a simple test for the setting of the log level.

There is a pytest fixture [`caplog`](https://docs.pytest.org/en/latest/logging.html#caplog-fixture) for capturing logging output, which would be really nice for test.

I can't get this to work, maybe someone has experience with it?

ctapipe/core/logging.py
```diff
@@ -1,39 +1,118 @@
...
-    Application logger functionality from LevelFormatter (in application.py)
-    """
-    highlevel_limit = logging.WARN
-    highlevel_format = " %(levelname)s |"
+
+class PlainFormatter(logging.Formatter):
+    """Custom logging.Formatter used for file logging."""
+
+
+class FancyFormatter(logging.Formatter):
```
maxnoe, line 14: Imho, Colored is better than Fancy.

ctapipe/core/tool.py
```diff
@@ -109,21 +109,38 @@ def main():
...
+        default_value="INFO",
+        help="Logging Level for Console Logging",
+    ).tag(config=True)
+    log_file = Path(
+        default_value=None, exists=None, directory_ok=False,
+        help="Filename for the log",
+    ).tag(config=True)
+    log_file_level = Enum(
+        values=Application.log_level.values,
+        default_value="DEBUG",
```
maxnoe, line 128: I would not put DEBUG here. Info should be fine as default. 

Probably `WARNING` for `log_level` and `INFO` for `log_file_level` would be a sane default.

ctapipe/core/tool.py
```diff
@@ -109,21 +109,38 @@ def main():
...
     def __init__(self, **kwargs):
         # make sure there are some default aliases in all Tools:
-        if self.aliases:
-            self.aliases["log-level"] = "Application.log_level"
-            self.aliases["config"] = "Tool.config_file"
+        aliases = {
+            "config": "Tool.config_file",
+            "log-config": "Tool.log_config",
+            "log-level": "Tool.log_level",
+            "log-file": "Tool.log_file",
```
maxnoe, line 138: I think this could just be `-l`, `--log`

ctapipe/core/tool.py
```diff
@@ -138,11 +155,28 @@ def initialize(self, argv=None):
...
         self.update_config(self.cli_config)
 
+        cfg = update_logging_config(
+            config=DEFAULT_LOGGING,
+            log_level=self.log_level,
+            log_file=self.log_file,
+            log_file_level=self.log_file_level,
+        )
+
+        if self.log_config is not None:
```
maxnoe, line 169: Why is this not part of `update_logging_config`?

ctapipe/tools/camdemo.py
```diff
@@ -18,12 +18,12 @@
 
 
 class CameraDemo(Tool):
-    name = "ctapipe-camdemo"
+    name = "ctapipe.camdemo"
```
maxnoe, line 21: Why this change?

nbiederbeck: I hope my memory does not let me down, but this was needed for more setting different levels in the config (-file?). eg
one "global" setting for all of `ctapipe`, a different  one for `ctapipe.io` .. this way `ctapipe.camdemo` would get the gloabl ctapipe setting

maxnoe: But this name is what the tool is called on the command line, there should be no dot here.

We should find a different solution. Even if this was the correct solution, we would need to change that in every tool.

maxnoe: Also, when the components are correctly setup, everything should start from this top level tool.

nbiederbeck:
@maxnoe I think I adressed all your points.

 I marked this PR as draft because I still get failing tests and thought travis would not check drafts

nbiederbeck:
Failing tests: 

* `test_stage_1`
* `test_muon_reconstruction`

both fail here:
```python
Caught unexpected exception: unsupported operand type(s) for +: 'PosixPath' and 'str'

File "/home/travis/build/cta-observatory/ctapipe/ctapipe/io/hdf5tableio.py", line 118, in __init__
    self.log.debug("h5file: %s", self._h5file)
```

where `self._h5file` is the file opened by `tables`. I checked, and this can be substituted by `filename`. I don't get why this happens only in this PR and not eg. on main branch.

Should I include this fix?

maxnoe:
This is a bug in hdf5 I fixed some time ago: https://github.com/PyTables/PyTables/pull/809

Does travis somehow install an old version of pytables?

nbiederbeck:
To save one click, this is what travis does 🤷‍♂️:
travis

maxnoe:
@nbiederbeck sorry, the latest release is from October *last* year. So the fix is not yet included.

maxnoe:
I have no idea why this only errors in the conda build then.

maxnoe:
Can you add a `pip freeze` as last command to the install section? That will be helpful for any future debugging as well.

maxnoe:
Ah, wait. I guess it is because you changed the log level of something to debug? Before it was probably info, so the message probably was never actually rendered.

nbiederbeck:
so ..

* pytables version is in all 4 travis runs `3.6.1`
* I'll add `pip freeze`, this would have made my search faster
* maybe it's because of a different logging level now, but I wonder why this did'nt pop up earlier (somewhere..) I'll look into it

maxnoe:
No it doesn't, this still weirds me out. Also, at some point this must have come up, otherwise I wouldn't have opened the PR at pytables.

nbiederbeck:
changing the level of the `root` logger "fixes" this problem

nbiederbeck:
## Quick overview of the new functionality in this PR

For tools:
* Add a default `ctapipe` logger with level `WARN`
* Add `-q` to not log anything onto command line
* Change the `ctapipe` logger level with `--log-level`
* Log to a file with `--log-file [--log-file-level]` , or `--log` or `-l` (regardless of `--log-level` or `-q`)
* Configure logging in-depth with our config files (`--config`)
    * Here we support loggers from other libraries, see example below

```python
{
    "CameraDemo": {
        "log_config": {
            "loggers": {
                "matplotlib": {"level": "DEBUG", "handlers": ["console"]},
                "ctapipe.utils": {"level": "ERROR"},
            }
        }
    }
}
```

maxnoe (approved):

kosack (approved):

PR #1313, opened 2020-05-07
Add more properties to EventSource, implement for SimTelEventSource, fixes #1286 

### LINKED ISSUE #1286: Make `is_simulated` a property of the event source
I think this is important information that should not be hidden in `metadata` and be required to be implemented for all event sources.

Opinions?

MaxNoe:
For all EventSources:
* obs_id
* is_simulation
* data_levels

For SimTelEventSource:
* mc_header


ctapipe/io/simteleventsource.py
```diff
@@ -205,6 +206,22 @@ def __exit__(self, exc_type, exc_val, exc_tb):
     def close(self):
         self.file_.close()
 
+    @property
+    def is_simulation(self):
+        return True
+
+    @property
+    def datalevels(self):
+        return ('R0', 'R1', 'DL0')
```
watsonjj, line 215: Instead of generic strings, would it be worth to have these as fixed enums?

watsonjj: Then a developer has a reference location to specify the correct levels instead of guessing, and adds a place for documenting the levels in the context of ctapipe

MaxNoe: Good Idea

MaxNoe:
@watsonjj I added the enum and removed two stump sources that were used nowhere (the `FitsSource` even had

```
def __iter__(self):
    # return containers
    pass
```

ctapipe/io/datalevels.py
```diff
@@ -0,0 +1,13 @@
...
+
+
+class DataLevel(Enum):
+    '''Enum of the different Data Levels'''
+    R0 = auto()
+    R1 = auto()
+    R2 = auto()
+    DL0 = auto()
+    DL1a = auto()
+    DL1b = auto()
```
kosack, line 11: we don't really have DL1a/b anymore. I'd just drop them for now (we realized it really depends on the use case, like training, etc, which parameters or images are filled in).  Just call it all DL1 for now until that is settled.

ctapipe/io/eventsource.py
```diff
@@ -200,6 +196,40 @@ def subarray(self):
...
+
+        Returns
+        -------
+        tuple[ctapipe.io.DataLevel]
+        """
+
+    @property
+    @abstractmethod
+    def obs_id(self):
+        """
```
kosack, line 225: Eventually this should be a full ObservationConfiguration but since that is not well defined yet, it's ok.   That way, you get the InstrumentConfiguration (SubarrayDescription) and the ObservationConfiguration as "metadata" attached to the source.   The ObservationConfiguration would contain other parameters describing the observation mode, etc (like the mcheader for sims, which is.a sort of SimulatedObservationConfiguration), though of course they are all  keyed by the obs_id, so this is fine. 

In fact, we could make that generic: e.g. define that all EventSources provide an ObservationConfiguration (ObsConfigContainer?), and we just make MCHeader a sub-class of this, adding parameters specific to MCs.

kosack: Note that you already get the obs_id inside the MCHeader (for simulations)

MaxNoe: Yes, this is where I get it in the SimTelEventSource

ctapipe/io/datalevels.py
MaxNoe: But for an event source it is a very big difference if it provides images or parameters. That was the idea behind DLa / DLb here.

Should I call it `DL1_images` ?

MaxNoe:
@kosack @watsonjj , sorry the merge accidentally removed a needed import, tests should be passing now

PR #1316, opened 2020-05-07
Calculate muon image parameters

MaxNoe:


ctapipe/instrument/camera/geometry.py
```diff
@@ -209,25 +209,50 @@ def __getitem__(self, slice_):
...
-    def _calc_pixel_area(pix_x, pix_y, pix_type):
+    @classmethod
+    def guess_pixel_area(cls, pix_x, pix_y, pix_type):
         """ recalculate pixel area based on the pixel type and layout
 
         Note this will not work on cameras with varying pixel sizes.
         """
 
-        dist = np.min(np.sqrt((pix_x[1:] - pix_x[0])**2 + (pix_y[1:] - pix_y[0])**2))
+        dist = cls.guess_pixel_width(pix_x, pix_y)
```
rlopezcoto, line 219: is this `dist` the radius (?) of the pixel? maybe it could be renamed to something more meaningful

MaxNoe: Sorry, it's explained in the other methods docstring. It is edge to edge distance for hexagons, edge width for square and diameter for circle

MaxNoe:
Had to resolve conflicts, please reapprove

ctapipe/image/muon/features.py
```diff
@@ -55,15 +52,14 @@ def photon_ratio_inside_ring(
         width of the ring
     """
 
-    total = np.sum(weights)
-
     pixel_r = np.sqrt((center_x - pixel_x)**2 + (center_y - pixel_y)**2)
     mask = np.logical_and(
         pixel_r >= radius - 0.5 * width,
         pixel_r <= radius + 0.5 * width
     )
```
moralejo, line 59: Why such a narrow range?

ctapipe/image/muon/features.py
```diff
@@ -55,15 +52,14 @@ def photon_ratio_inside_ring(
...
 
-    inside = np.sum(weights[mask])
+    inside = weights[mask].sum()
+    total = weights.sum()
```
moralejo, line 62: Is this with no pixel selection at all? It will be a very noisy quantity. If the goal is to test for possible shower contamination, it is better to take a wider area around the ring (say, 20% of the ring radius or so) and check for pixels above some threshold (or simply post-cleaning) which are outside of that area.

ctapipe/tools/muon_reconstruction.py
```diff
@@ -174,13 +194,7 @@ def process_telescope_event(self, event_index, tel_id, dl1):
...
-        # add ring containment, not filled in fit
-        ring.containment = ring_containment(
-            ring.radius,
-            fov_radius,
-            ring.center_x,
-            ring.center_y,
-        )
+        parameters = self.calculate_muon_parameters(tel_id, image, clean_mask, ring)
 
         # intensity_fitter does not support a mask yet, set ignored pixels to 0
```
moralejo, line 199: This will not work exactly as desired, the pixels will enter the fit all the same, even with 0 values. MuonIntensityFitter should support masks

ctapipe/image/muon/features.py
MaxNoe: I think you are confusing something here.

going from -0.5 width to 0.5 width is the very definition of a ring centered at radius `r` with the corresponding width.

ctapipe/image/muon/features.py
MaxNoe: This function is agnostic of that, it only does what you see here. You can pass x,y, weights of anything, either a full image or only the pixels after cleaning.

ctapipe/tools/muon_reconstruction.py
MaxNoe: Yes, agreed. Setting them to zero will bias the computation.

ctapipe/image/muon/features.py
moralejo: Ah, ok, you are right, here you pass x, y so no problem, they can be masked as well. But the comment holds for the higher level call of MuonIntensityFitter, right?

MaxNoe: Yes, this one should support a mask.

ctapipe/image/muon/features.py
moralejo: The width is "half-width", I think. So it should be (-width, width). But there will be some light (~32% assuming gaussian) beyond that distance, so if the goal is to compare "ring light" with (non-ring) shower light perhaps a larger range would be better.

MaxNoe: In the tool, the width to be taken into account here is currently customizable in terms of the pixel diameter. The default is `width = 1.5 * pixel diameter`

moralejo: ok, I was assuming it is the fitted ring width. It is confusing that that the fitted one is 'half-width',  while this one is full width. But I agree it is not a problem of this function. One should just choose a 'width' value to call it depending on the goal, e.g. catching all of the muon light or not.

MaxNoe: I think we should probably not call the standard deviation of the gaussian profile `width`. It is not a width in any usual sense.

PR #1318, opened 2020-05-07
Update tests with new ctapipe-extra, remove hessio remnants

MaxNoe:
* Also gets rid of the duplicated environments by using sed to replace the python version before installing


environment.yml
```diff
@@ -37,8 +38,7 @@ dependencies:
   - wheel
   - xz
   - zlib
-  - pip
   - pip:
     - eventio~=1.0
-
-
+    - zstandard
```
watsonjj, line 43: What is this addition for?

MaxNoe: To read simtel files compressed using the zstd compression

MaxNoe: This was used already for prod4 and will i think be used for prod5 as well. Its much faster than gzip at a similar compression rate.

watsonjj: Isn't this already required by pyeventio then?

MaxNoe: It's an optional dependency of pyeventio, so we need to install ot explicitly

MaxNoe: Otherwise you will get a `You need zstandard to read this file` error

kosack, line 43: Move this out of pip install, it has a conda package and it's even the base anaconda distribution (not even conda-forge)

MaxNoe: Ah, nice. Sure

MaxNoe:
@kosack @watsonjj pls review again

PR #1324, opened 2020-05-13
Rename muon containers to make naming more consistent

MaxNoe:
The muon containers were named pretty inconsistently to the other containers


MaxNoe:
Sorry, overlooked the column name in the tests.

PR #1325, opened 2020-05-13
First version of monitoring data storage

maxnoe:
Here I started implementing a monitoring data storage making use of sortedcontainers to efficiently lookup data for observations.

Keys are aribiraty as long as they are hashable and implement comparison and subraction operations.

### DIFF
--- a/ctapipe/monitoring/__init__.py
+++ b/ctapipe/monitoring/__init__.py
@@ -1,0 +1,6 @@
+from .data import MonitoringData
+
+
+__all__ = [
+    'MonitoringData',
+]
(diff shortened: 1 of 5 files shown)


maxnoe:
Closing this, as we will probably use fits files for monitoring data storage in cta and usage with those would be more straight forward using astropy tables. 

ctapipe_io_lst also has code interpolating pointing tables and finding closest entry that matches a timestamp here:
https://github.com/cta-observatory/ctapipe_io_lst/blob/v0.19.0/src/ctapipe_io_lst/pointing.py

PR #1329, opened 2020-05-14
use float32 for images, and validate

kosack:
- Changes Containers for images to specify
they should be float32
- call validate() when making schema in HDF5TableWriter

So far doesn't change the computation of the images. Need to ensure ImageExtractor creates float32s.

### DIFF
--- a/ctapipe/image/extractor.py
+++ b/ctapipe/image/extractor.py
@@ -91,19 +91,31 @@ def extract_around_peak(
     n_samples = waveforms.size
     start = peak_index - shift
     end = start + width
-    sum_[0] = 0
-    time_num = 0
-    time_den = 0
-    for isample in prange(start, end):
-        if 0 <= isample < n_samples:
-            sum_[0] += waveforms[isample]
-            if waveforms[isample] > 0:
-                time_num += waveforms[isample] * isample
-                time_den += waveforms[isample]
-    peak_time[0] = time_num / time_den if time_den > 0 else peak_index
 
+    # whole range invalid
+    if end <= 0:
+        peak_time[0] = peak_index / sampling_rate_ghz
+        sum_[0] = 0
+        return
+
+    # reduce to valid range
+    start = max(0, start)
+    end = min(end, n_samples)
+
+    i_sum = float64(0.0)
+    time_num = float64(0.0)
+    time_den = float64(0.0)
+
+    for isample in prange(start, end):
+        i_sum += waveforms[isample]
+        if waveforms[isample] > 0:
+            time_num += waveforms[isample] * isample
+            time_den += waveforms[isample]
+
+    peak_time[0] = time_num / time_den if time_den > 0 else peak_index
     # Convert to units of ns
     peak_time[0] /= sampling_rate_ghz
+    sum_[0] = i_sum
 
 
 @njit(parallel=True)
--- a/ctapipe/image/extractor.py
+++ b/ctapipe/image/extractor.py
@@ -40,7 +40,7 @@ from .hillas import hillas_parameters, camera_to_shower_coordinates
     nopython=True,
 )
 def extract_around_peak(
-        waveforms, peak_index, width, shift, sampling_rate_ghz, sum_, peak_time
+    waveforms, peak_index, width, shift, sampling_rate_ghz, sum_, peak_time
 ):
     """
     This function performs the following operations:
@@ -91,12 +91,6 @@ def extract_around_peak(
     n_samples = waveforms.size
     start = peak_index - shift
     end = start + width
-
-    # whole range invalid
-    if end <= 0:
-        peak_time[0] = peak_index / sampling_rate_ghz
-        sum_[0] = 0
-        return
 
     # reduce to valid range
     start = max(0, start)
--- a/ctapipe/image/extractor.py
+++ b/ctapipe/image/extractor.py
@@ -33,8 +33,8 @@ from .hillas import hillas_parameters, camera_to_shower_coordinates
 
 @guvectorize(
     [
-        (float64[:], int64, int64, int64, float64, float64[:], float64[:]),
-        (float32[:], int64, int64, int64, float64, float64[:], float64[:]),
+        (float64[:], int64, int64, int64, float64, float32[:], float32[:]),
+        (float32[:], int64, int64, int64, float64, float32[:], float32[:]),
     ],
     "(s),(),(),(),()->(),()",
     nopython=True,
--- a/ctapipe/io/hdf5tableio.py
+++ b/ctapipe/io/hdf5tableio.py
@@ -148,6 +148,9 @@ class HDF5TableWriter(TableWriter):
         # create pytables schema description for the given container
         pos = 0
         for container in containers:
+
+            container.validate()  # ensure the data are complete
+
             for col_name, value in container.items(add_prefix=self.add_prefix):
 
                 typename = ""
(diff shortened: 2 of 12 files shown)


ctapipe/io/hdf5tableio.py
```diff
@@ -148,6 +148,9 @@ class Schema(tables.IsDescription):
         # create pytables schema description for the given container
         pos = 0
         for container in containers:
+
+            container.validate()  # ensure the data are complete
```
vuillaut, line 152: `validate` checks that every `Field` in the container is of the expected type?

MaxNoe: See #1301

MaxNoe:
@kosack I added a test testing for all containers defined in `ctapipe/containers.py` if the default can be written using the hdf5 writer, this turned up some more missing units / bad defaults.

I also changed three time ranges to two fields min / max to avoid needless non-scalar fields.

kosack:
Thanks, this looks good. I'll just update the ImageExtractor then to produce the right data type

kosack:
I think to fix that it's just to change `extract_around_peak`, which currently returns float64s.  @watsonjj is that enough?  

An alternative would be to down-sample it afterward, but that is slower and probably not a good idea.  If we rely on any operations being on 64-bit versions, then the answer would change when we re-read DL1 data stored in 32-bits and re-compute things, which is not a good design.

ctapipe/image/extractor.py
```diff
@@ -33,8 +33,8 @@
 
 @guvectorize(
     [
-        (float64[:], int64, int64, int64, float64, float64[:], float64[:]),
-        (float32[:], int64, int64, int64, float64, float64[:], float64[:]),
+        (float64[:], int64, int64, int64, float64, float32[:], float32[:]),
+        (float32[:], int64, int64, int64, float64, float32[:], float32[:]),
```
MaxNoe, line 37: Are you sure this actually changes the output dtype? Isn't it inferred and this just precompiles these two signatures? Not sure.

kosack: I'm not sure - I would have thought if they are specified, it uses only them, but it may also infer... I'll add a test

kosack: seems to work in the test at least

MaxNoe: Ok, thatn probably good.

HealthyPear:
> I think to fix that it's just to change `extract_around_peak`, which currently returns float64s to return float32s instead. @watsonjj is that enough?
> 
> An alternative would be to down-sample it afterward, but that is slower and probably not a good idea. If we rely on any operations being on 64-bit versions, then the answer would change when we re-read DL1 data stored in 32-bits and re-compute things, which is not a good design.

In TwoPassWindowSum I had to force the conversion (line 903), is this is still needed? I was getting an error if I didn't convert to 32

kosack:
> In TwoPassWindowSum I had to force the conversion (line 903), is this is still needed? I was getting an error if I didn't convert to 32

Do you recall what error you were getting? I don't see anything that would have forbidden 64-bits,  in ctapipe at least (if it was in prototype, you have your own IO system, and may have defined it as 32 bits)

MaxNoe:
@kosack I had to require an rtol of only `1e-4` to make that test pass with float32. 

I guess 1e-4 on waveforms is still plenty, but are you sure we don't want to use float64 internally?

HealthyPear:
> > In TwoPassWindowSum I had to force the conversion (line 903), is this is still needed? I was getting an error if I didn't convert to 32
> 
> Do you recall what error you were getting? I don't see anything that would have forbidden 64-bits, in ctapipe at least (if it was in prototype, you have your own IO system, and may have defined it as 32 bits)

Ah yes, it is possible that was a ctapipe 0.7 issue....probably that conversion is not needed anymore.

As soon as we merge this, I will try take it out to see if nothing happens (as I expect)

kosack:
> I guess 1e-4 on waveforms is still plenty, but are you sure we don't want to use float64 internally?

Perhaps, or else we need to think about changing operations to minimize increasing floating point error, which normally you can ignore in 64 bits, but not in 32.

kosack:
The worry I have is that if we store a 64-bit version in the Container then if we run the analysis without writing a DL1 file, we would have all operations in 64bits, but if we write and read, it woudl change to 32 bits, and thus would give different results.  So no matter what, we should write 32-bits to the container in memory, so further algorithms don't expect a higher bit depth

MaxNoe:
I'll try to improve the numerical stability of the extractor.

MaxNoe:
When using `numpy.sum` in `extract_around_peak`, the relative error is one order of magnitude less (numpy.sum does uses a middle ground implementation between floating point accuracy and speed instead of a plain loop sum).

This increases runtime from 430 µs to 450 µs in the test I did. This is acceptable I think, right?

Should I push this?

kosack:
> Should I push this?

Sounds good, I say go ahead.

watsonjj:
> When using `numpy.sum` in `extract_around_peak`, the relative error is one order of magnitude less (numpy.sum does uses a middle ground implementation between floating point accuracy and speed instead of a plain loop sum).
> 
> This increases runtime from 430 µs to 450 µs in the test I did. This is acceptable I think, right?
> 
> Should I push this?

I'm confused, what does using `numpy.sum` solve? I specifically didn't use it to correctly sum samples when the window is at the edge of the waveform, and also to not include negative values in the pulse_time calculation

kosack:
It's just that now that we've switched to 32-bit images, we have to worry more about numerical rounding error, so the way things are summed is important.

kosack:
you can see now, in 32bit, the tests fail, since the difference after error is around 1e-3, and the tolerance is 1e-7:

```
>       np.testing.assert_allclose(event.dl1.tel[telid].image, y.sum(1))
E       AssertionError:
E       Not equal to tolerance rtol=1e-07, atol=0
E
E       Mismatched elements: 1613 / 2048 (78.8%)
E       Max absolute difference: 0.0012207
E       Max relative difference: 6.832399e-05
E        x: array([402.1624 , 694.62103, 981.598  , ..., 525.88116, 796.07776,
E              855.5047 ], dtype=float32)
E        y: array([402.16254, 694.6211 , 981.59845, ..., 525.881  , 796.07776,
E              855.50446], dtype=float32)
```

watsonjj:
I see. I'm surprised at the difference!

kosack:
> I see. I'm surprised at the difference!

It was so much more difficult back in the days of 32-bit (or even 16-bit!) computers and low-ram! using doubles used to have a large speed and memory impact, so we had to use 32-bit floats for everything, and always think about numerical stability.

MaxNoe:
@watsonjj numpy sum is only used for the image sum, not the weighted average.

The numpy sum uses an algorithm optimized as a tradeoff between speed and minimizing floating point error, it reduced rounding errors.

kosack:
see e.g. https://en.wikipedia.org/wiki/Kahan_summation_algorithm or https://code.activestate.com/recipes/393090/ 

I guess fsum is slow, and we don't really need that level of precision, so either stick with low-precision (assuming we don't add too many more sample), or use np.sum, which I assume uses something like the kahan algorithm

MaxNoe:
So what should we do here?

I would say this is the last blocker for 0.8. Should we revert the numpy sum and increase the rtol or leave it like it is now?

@kosack @watsonjj 

Would be nice to get it out today

watsonjj:
I think an rtol of 1e-4 is still fine for our purpose. There was no particular reason I previously used 1e-7

MaxNoe:
We could also use numba.double to make the accumulator a double bevor assigning to the 32bit array

MaxNoe:
@watsonjj why does the loop go over all samples and then do branching to check if `0 < i < n_samples`? 

I replaced this with min /  max before the loop and it's now twice as fast.

PR #1334, opened 2020-05-15
Use setuptools_scm for versioning, fixes #1333 

### LINKED ISSUE #1333: Replace custom version with setuptools_scm
There is an extension for setuptools, which provides the functionality we need and which 
integrates nicely with the build system requirements handled by pyproject.toml.

Astropy is are also using this:
https://github.com/astropy/astropy/blob/master/pyproject.toml

maxnoe:


kosack:
please add a PR description of what this solves

kosack:
could you merge in the master and fix conflicts? I'll take a look at this soon

kosack (dismissed):

kosack (approved):

maxnoe:
@kosack I recheck with the last astropy issues/PRs and implemented what they do, which makes sense I think.
So it behaves now like this:

* If installed locally, we use setuptools_scm via `_dev_version.py` This file is excluded from distribution (sdist / wheel), so will not be included in a pypi upload. Not excluding setuptools scm for dists lead to some weird issues for astropy, so this is avoided now.

* When building the dists, a file `_version.py` with just the version information is created. This is used for version information in releases

* When neither is available, a clear warning is given and version = 0.0.0 is set. The only scenario this happens in should be if people try to install ctapipe through a release tarball autogenerated by github. We should probably also add a warning at the bottom of each github release, that these tar balls should not be used to install ctapipe.

maxnoe:
Ah, we could also think about uploading the sdist using the github releases plugin for pypi, so we have a functioning tarball on github (but just using pypi / conda is also fine I think)

kosack (approved):

watsonjj (approved):

Issue #1337, opened 2020-05-18
wrong photo_electron_image when IGNORE_TELESCOPEs option was used running sim_telarray


yrenier:
I have corsika files with 7 telescopes and used IGNORE_TELESCOPEs=1,2,7 to keep only the LSTs (telescopes 3, 4, 5 and 6) during the sim_telarray simulations.
using ctapipe.io.event_source() on the resulting simtel file results with wrong data in event.mc.tel[tel_id].photo_electron_image.
this is due to the folowing existing code:
```
tel_index = self.file_.header['tel_id'].tolist().index(tel_id)
...
mc.photo_electron_image = (
                    array_event
                    .get('photoelectrons', {})
                    .get(tel_index, {})
                    .get('photoelectrons', np.zeros(n_pixels, dtype='float32'))
                )
```
in my case I have 'self.file_.header['tel_id'].tolist() = [3,4,5,6]' so  tel_index=1 looking for tel_id=4.
However array_event['photoelectrons'].keys() is always a subset of [2, 3, 4, 5] so the the code does not fill any info for telescopes 3 and 4 while get the info from the wrong telescopes otherwise.

The solution seems to be to use the following code instead:
```
mc.photo_electron_image = (
                    array_event
                    .get('photoelectrons', {})
                    .get(tel_id - 1, {})
                    .get('photoelectrons', np.zeros(n_pixels, dtype='float32'))
                )
```

yrenier:
The code above is for ctapipe 0.7, as the naming changed in master I will prepare a PR

maxnoe:
Pr with fix was merged.

PR #1338, opened 2020-05-18
fix bug when ignore_telescopes was used in sim_telarray

yrenier:
Issue reported here #1337

### DIFF
--- a/ctapipe/io/simteleventsource.py
+++ b/ctapipe/io/simteleventsource.py
@@ -370,8 +370,6 @@ class SimTelEventSource(EventSource):
             telescope_events = array_event["telescope_events"]
             tracking_positions = array_event["tracking_positions"]
             for tel_id, telescope_event in telescope_events.items():
-                tel_index = self.file_.header["tel_id"].tolist().index(tel_id)
-
                 adc_samples = telescope_event.get("adc_samples")
                 if adc_samples is None:
                     adc_samples = telescope_event["adc_sums"][:, :, np.newaxis]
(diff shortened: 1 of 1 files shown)


ctapipe/io/simteleventsource.py
```diff
@@ -382,7 +380,7 @@ def _generate_events(self):
                 mc.pedestal = array_event["camera_monitorings"][tel_id]["pedestal"]
                 mc.true_image = (
                     array_event.get("photoelectrons", {})
-                    .get(tel_index, {})
+                    .get(tel_id - 1, {})
```
MaxNoe, line 383: I think this should be just tel_id, why did you add the `-1`?

yrenier: for my case array_event['photoelectrons'] has keys which are from [2, 3, 4, 5] whereas file_.header['tel_id'].tolist() = [3,4,5,6].
From that observation the -1 is needed and I guess the difference comes from different indexes conventions.

MaxNoe: Yes, I can reproduce this. But this shouldn't be the case. I asked Konrad about this. Only the 1208 objects with the photo electrons seem to have this id off by one error.

watsonjj: Relevant previous PR: #943

watsonjj: I reiterate the need to have a test file with the photoelectron image inside, to test this behaviour and ensure it isn't broken

MaxNoe: We have that now. It's the `lst_prod3_...` file. But that does not have ignored telescopes. So tel_index == tel_id - 1.


Here we have another case. What is used here as `tel_id` in the photo electron block seems to be the **initial** telescope index before telescopes where ignored.

It would be interesting what's in there when telescopes are simulated with non-consecutive ids and then some are ignored.

PR #1339, opened 2020-05-18
fix bug where image was 64-bit in  case where no extraction is needed

kosack:


kosack:
we're missing a test for  telescopes with no extraction needed.

PR #1343, opened 2020-05-19
Make sure unit is really a unit, use fits format

MaxNoe:
Fix for the strange unit in the metadata as seen in @kosack  talk:

Screenshot_20200519_172143

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -10,6 +10,7 @@ matrix:
   include:
     - name: conda python 3.6
       os: linux
+      dist: bionic
       language: generic
       env:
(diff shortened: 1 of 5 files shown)


kosack:
yeah, I noticed that during the talk! clearly a but strange

kosack:
perhaps "vounit" format should be used. It's similar to FITS, but a but more modernized

MaxNoe:
Changed to vounit format

MaxNoe:
Ubuntu 18.04, Codename bionic beaver

PR #1345, opened 2020-05-20
reformat with black

kosack:
as in #1265 , added back in a pre-commit hook for black, and reformatted all code. 

Though this means devs have less flexibility, it makes future commits and diffs much easier to read since the formatting will be the same (most people already are using black in their editor, so code gets reformatted automatically anyhow).  This should be the recommended practice


MaxNoe:
Why this python based pre-commit setup?
Wouldn't it be enough to add black to the environment and dev dependencies and  add

```
black .
```
to `.git/hooks/pre-commit`?

kosack:
> Why this python based pre-commit setup?

I just followed what they said in the black docs. Not sure exactly why it's better or worse.

kosack:
ah, now I remember why I used pre-commit.  I couldn't figure out how to make a pre-commit hook part of the package.  This way, you specify what hooks are there, and they get installed.  Things inside `.git` are not committable, right? Is there a better way to do it for all developers?

kosack:
perhaps we put them in a hooks/ subdir of the repo, and use this: https://git-scm.com/docs/git-config#Documentation/git-config.txt-corehooksPath ?

MaxNoe:
@kosack seems fine then.

PR #1348, opened 2020-05-20
last fixes for release of v0.8

kosack:
- remove pip installs from environment.yml, which are now conda packages (for Linux and MacOS)

### DIFF
--- a/environment.yml
+++ b/environment.yml
@@ -42,5 +42,5 @@ dependencies:
   - xz
   - zlib
   - zstandard
-  - eventio>=1.0
+  - eventio>=1.1.1
   - ctapipe-extra>=0.3
--- a/environment.yml
+++ b/environment.yml
@@ -42,6 +42,5 @@ dependencies:
   - xz
   - zlib
   - zstandard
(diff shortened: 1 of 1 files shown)


kosack:
any other last-minute cleanups?

MaxNoe:
see #1349

environment.yml
```diff
@@ -42,6 +42,5 @@ dependencies:
   - xz
   - zlib
   - zstandard
-  - pip:
-    - eventio~=1.0
-    - https://github.com/cta-observatory/ctapipe-extra/archive/v0.3.0.tar.gz
+  - eventio>=1.0
```
MaxNoe, line 45: we require at least  1.1.1 for the calibration event type. I thought I updated this also here (in setup.py it's correct)

MaxNoe: change to `>=1.1.1,<2`

kosack: done.  For the conda package, all specs are taken from setup.py, so if a user installs it via conda, it will be all ok.

PR #1349, opened 2020-05-20
Use precomputed mean/std to compute skewness and kurtosis

MaxNoe:
master:
```
334 µs ± 1.48 µs per loop (mean ± std. dev. of 7 runs, 1000 loops each)
```

this:
```
55.7 µs ± 549 ns per loop (mean ± std. dev. of 7 runs, 10000 loops each)
```

### DIFF
--- a/ctapipe/image/statistics.py
+++ b/ctapipe/image/statistics.py
@@ -28,8 +28,12 @@ def skewness(data, mean=None, std=None):
     skewness: float
         computed skewness
     """
-    mean = mean or np.mean(data)
-    std = std or np.std(data)
+    if mean is None:
+        mean = np.mean(data)
+
+    if std is None:
+        std = np.std(data)
+
     return np.mean(((data - mean) / std) ** 3)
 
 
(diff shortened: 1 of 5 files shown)


ctapipe/image/statistics.py
```diff
@@ -1,17 +1,77 @@
...
 
 from ..containers import StatisticsContainer
 
 
+def skewness(data, mean=None, std=None):
+    '''Calculate skewnewss (normalized third central moment)
+    with allowing precomputed mean and std.
+
+    With precomputed mean and std, this is ~10x faster than scipy.stats.skew
+    for our use case (1D arrays woth ~100-1000 elements)
```
dneise, line 11: typo in with

ctapipe/image/statistics.py
```diff
@@ -1,17 +1,77 @@
...
+    std = std or np.std(data)
+    return np.average(((data - mean) / std)**3)
+
+
+def kurtosis(data, mean=None, std=None, fisher=True):
+    '''Calculate kurtosis (normalized fourth central moment)
+    with allowing precomputed mean and std.
+
+    With precomputed mean and std, this is ~10x faster than scipy.stats.skew
+    for our use case (1D arrays woth ~100-1000 elements)
```
dneise, line 37: same typo again

watsonjj:
These are simple mathematical operations and has couple of temporary arrays....you could check what further performance you get with `njit`, and maybe with `(fastmath=True)`

MaxNoe:
I tried with njit and the improvement if any is in the noise.

MaxNoe:
Added more tests and njit, since it gave a 10 % improvement. Pls. review again.

MaxNoe:
@kosack done

MaxNoe:
pytest cov seems to be unable to work with the njit functions. This is unfortunate but also not really surprising in hindsight.

This is way diff coverage is so low

PR #1350, opened 2020-05-20
Improve performance of concentration by 2x by stripping units internally

MaxNoe:


kosack:
I think we can leave other speed improvements for the next minor release - there's probably a lot we can do

MaxNoe:
Agreed. These were just so simple that I had to make them :wink:

PR #1351, opened 2020-05-21
use robust fit to get timing slope

### LINKED ISSUE #1275: use a robust fitting in timing_parameters
Right now the computation, using `np.polyfit()` of the timing axis in timing_parameters can be quite biased if there are outliers included (some noise included after cleaning, which happens relatively often).  It was suggested here  https://github.com/cta-observatory/ctapipe/issues/602#issuecomment-350823139 that we should use a robust fitting routine (e.g. one that is not strongly affected by outliers), and this is already implenmented in MARS and EventDisplay. 

Here is a comparison of the results of fitting a linear function  with polyfit to several robust methods included in SciPy:

```py3
    N = 15
    x = np.arange(N)
    y = 3*x+2 + np.random.uniform(5.0,size=N)-2.5
    
    # add some outliers
    y[outlier_pos] = 3*x[4]+2 + outlier_level
    
    pslope, pintercept = np.polyfit(x,y,1)
    tslope, tintercept, err_slope_low, err_slope_high= stats.theilslopes(y,x=x)
    sslope, sintercept, = stats.siegelslopes(y,x=x, method='hierarchical')
    ssslope, ssintercept, = stats.siegelslopes(y,x=x, method='separate')
```

image

I suggest we replace `polyfit`  with `scipy.stats.siegelslopes` to get a better fit.  The only problem is that these functions do not return error estimates, so those must be computed separately. 

This is also mentioned in #1215

kosack:
fixes #1275 

This just re-computes the slope with `scipy.stats.siegelslopes()` to get a result that is less affected by outliers (which is known to improve the TwoPassWindowSum image extraction method)

It still uses polyfit to estimate errors, since siegelslopes does not return a cov matrix.

### DIFF
--- a/ctapipe/image/timing.py
+++ b/ctapipe/image/timing.py
@@ -9,6 +9,8 @@ from ..containers import TimingParametersContainer
 from .hillas import camera_to_shower_coordinates
 from ..utils.quantities import all_to_value
 
+from scipy.stats import siegelslopes
+
 
 __all__ = ["timing_parameters"]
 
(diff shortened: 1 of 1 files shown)


HealthyPear:
From a pure analytical point of view I am not sure how to treat the mixing of algorithms for calculating the expectation value and it's error bar (assuming that is analytically correct, I have never seen it).

When I added this robust fit to that image extractor I kept out the error calculation for this reason (and also because it's not clear to me where it should propagate afterwards in the pipeline).

kosack:
> From a pure analytical point of view I am not sure how to treat the mixing of algorithms for calculating the expectation value and it's error bar (assuming that is analytically correct, I have never seen it).

It's not strictly correct, but better than nothing.  So far the errors are not used anywhere, but they could just be interesting for quality cuts (e.g. to identify really bad outliers)

kosack:
@HealthyPear I guess this is important to merge before the v0.8 release, though, since otherwise you cannot do your comparisons with this version.

HealthyPear:
Aaahh yes sorry I must have confused with protopipe!
I thought I added this together with the image extractor PR...instead I did it only in protopipe...

PR #1354, opened 2020-05-22
Miscelleanous performance improvements

MaxNoe:
## Quantity vs. Angle

```
In [4]: %timeit Angle(5, u.rad)                                                                                                               
10.1 µs ± 63.7 ns per loop (mean ± std. dev. of 7 runs, 100000 loops each)

In [5]: %timeit u.Quantity(5, u.rad)                                                                                                          
3.46 µs ± 28.7 ns per loop (mean ± std. dev. of 7 runs, 100000 loops each)
```

## Reimplement number_of_islands

From
```
318 µs ± 11.7 µs per loop (mean ± std. dev. of 7 runs, 1 loop each)
```
To
```
13.8 µs ± 15.4 µs per loop (mean ± std. dev. of 7 runs, 1 loop each)
```

## Timing parameters using sigma clipping instead of siegelslopes:

1855 pixels:

From
```
153 ms ± 382 µs per loop (mean ± std. dev. of 7 runs, 10 loops each)
```
To
```
311 µs ± 88.9 µs per loop (mean ± std. dev. of 7 runs, 1 loop each)
```

100 Pixels:
From 
```
4.35 ms ± 5.39 µs per loop (mean ± std. dev. of 7 runs, 100 loops each)
```
To
```
229 µs ± 102 µs per loop (mean ± std. dev. of 7 runs, 1 loop each)
```

## Overall

On the LST Muon Mono run:

From
```
10000/10000 [02:29<00:00, 66.99ev/s]
```
To
```
10000/10000 [01:28<00:00, 113.16ev/s]
```

### DIFF
--- a/ctapipe/image/timing.py
+++ b/ctapipe/image/timing.py
@@ -15,7 +15,7 @@ __all__ = ["timing_parameters"]
 
 
 @njit(cache=True)
-def linear_regression(x, y):
+def linear_regression(x, y, weights):
     """
     njit version of a least squares linear regression
 
@@ -40,14 +40,15 @@ def linear_regression(x, y):
     X = np.empty((len(x), 2))
     X[:, 0] = x
     X[:, 1] = 1
+    W = np.diag(weights)
 
-    cov = np.linalg.inv(X.T @ X)
-    params = cov @ X.T @ y
+    cov = np.linalg.inv(X.T @ W @ X)
+    params = cov @ X.T @ W @ y
     return params[0][0], params[1][0], cov
 
 
 @njit(cache=True)
-def sigma_clipping_linreg(x, y, kappa=2.5, n_iter=3):
+def sigma_clipping_linreg(x, y, weights, kappa=2.5, n_iter=3):
     """
     Linear regression with sigma clipping.
     Iteratively perform a linear regression, excluding points
@@ -75,7 +76,7 @@ def sigma_clipping_linreg(x, y, kappa=2.5, n_iter=3):
         covariance matrix of the slope and intercept
     """
 
-    a, b, cov = linear_regression(x, y)
+    a, b, cov = linear_regression(x, y, weights=weights)
     mask = np.ones(len(x), dtype=np.bool_)
 
     for i in range(n_iter):
@@ -83,7 +84,7 @@ def sigma_clipping_linreg(x, y, kappa=2.5, n_iter=3):
         sigma = np.std(delta)
         mask &= np.abs(delta) < (kappa * sigma)
 
-        a, b, cov = linear_regression(x[mask], y[mask])
+        a, b, cov = linear_regression(x[mask], y[mask], weights=weights[mask])
 
     return a, b, cov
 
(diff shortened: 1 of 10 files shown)


MaxNoe:
Okay, I am no bumping my head against this since an hour. Sometimes numba just get's stuck on `set.add`, I suspect a numba bug, but I cannot reproduces this outside of the num islands code

MaxNoe:
For my contrived test case, the sigma clipping and the siegelslopes are very similar:
```
import numpy as np
from scipy.stats import siegelslopes
from ctapipe.image.timing import linear_regression, sigma_clipping_linreg


np.random.seed(0)

x = np.random.uniform(-0.2, 0.2, 50)
x.sort()
y = 10 * x + 5 + np.random.normal(0, 0.2, 50)

y[-3] = 10
y[-5] = 12
y[-15] = 7


xp = np.array([x.min(), x.max()])

plt.plot(x, y, 'o')
plt.plot(xp, 10 * xp + 5, '-', label='truth')


a, b, cov = linear_regression(x, y)
plt.plot(xp, a * xp + b, '-', label='linear regression')

a, b, cov = sigma_clipping_linreg(x, y)
plt.plot(xp, a * xp + b, '-', label='sigma clipping')

a, b = siegelslopes(y, x)
plt.plot(xp, a * xp + b, '-', label='siegel slopes')


plt.legend()
plt.savefig('robust.png', dpi=300)

```

robust

ctapipe/image/timing.py
```diff
@@ -76,12 +76,13 @@ def sigma_clipping_linreg(x, y, kappa=3, n_iter=3):
     """
 
     a, b, cov = linear_regression(x, y)
+    mask = np.ones(len(x), dtype=np.bool_)
```
kosack, line 79: can use `np.ones_like(x, dtype=np.bool_)` here

MaxNoe: ones_like was a bit slower and we only support 1d here anyways.

kosack:
numba does support `np.linalg.lstsq` (https://numba.pydata.org/numba-doc/dev/developer/autogen_numpy_listing.html?highlight=lstsq#numpy.linalg.lstsq),  so you might use that rather than your own implementation (to avoid inverse problems), or use SVD or something

MaxNoe:
But numpy.linalg.lstsq does not support weights, does it?

kosack:
https://stackoverflow.com/questions/27128688/how-to-use-least-squares-with-weight-matrix

ctapipe/image/timing.py
```diff
@@ -4,17 +4,91 @@
...
+    cov: 2x2 ndarray
+        covariance matrix of the slope and intercept
+    """
+    y = y.reshape((-1, 1))
+    X = np.empty((len(x), 2))
+    X[:, 0] = x
+    X[:, 1] = 1
+    W = np.diag(weights)
+
+    cov = np.linalg.inv(X.T @ W @ X)
```
kosack, line 45: I suggest using `np.linalg.pinv()` here instead (pseudo-inverse, using SVD) That will never fail,  and is more stable.

MaxNoe:
Closed as it was getting to large and mixed changes affecting the results with changes only speeding things up. I will open follow-op PRs for the single changes.

PR #1356, opened 2020-06-03
improve TwoPassWindowSum extractor

kosack:
Due to the fact that each pixel might have a different value for `window_shift`, the original implementation was very slow (5-10 seconds/event), since it computed the integration correction for each event and each pixel.  However, really there is only need to compute it 3 times:  once for the usual case where the pulse is inside the readout window, and once each for the edge cases where it is before or after the window (where the window size and shift is smaller like in MARS).

This PR just minimally refactors the existing code to avoid repeated calls to `_compute_correction()` by memoization and treating the edge cases separately.    It results in a large speed increase when using the TwoPass method, though there is still work to be done (and some will be solved by merging in #1354, which improves the speed of timing_parameters, which is used internally here)

**Update**: this also fixes a bug where the integration windows were smaller than intended by 1 sample.

### DIFF
--- a/ctapipe/image/extractor.py
+++ b/ctapipe/image/extractor.py
@@ -26,6 +26,7 @@ from ctapipe.core.traits import IntTelescopeParameter, FloatTelescopeParameter
 from ctapipe.core import TelescopeComponent
 from numba import njit, prange, guvectorize, float64, float32, int64
 from scipy.ndimage.filters import convolve1d
+from typing import Tuple
 
 from . import number_of_islands, largest_island, tailcuts_clean
 from .timing import timing_parameters
@@ -624,6 +625,10 @@ class TwoPassWindowSum(ImageExtractor):
         help="only run the first pass of the extractor, for debugging purposes",
     ).tag(config=True)
 
+    peak_finding_window_width = IntTelescopeParameter(
+        default_value=3, help="width of sliding window used to do peak detection"
+    ).tag(config=True)
+
     @lru_cache(maxsize=4096)
     def _calculate_correction(self, telid, width, shift):
         """Obtain the correction for the integration window specified for each
@@ -660,7 +665,9 @@ class TwoPassWindowSum(ImageExtractor):
             shift,
         )
 
-    def _apply_first_pass(self, waveforms, telid):
+    def _apply_first_pass(
+        self, waveforms, telid
+    ) -> Tuple[np.ndarray, np.ndarray, np.ndarray]:
         """
         Execute step 1.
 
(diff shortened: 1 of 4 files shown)


kosack:
@HealthyPear I noticed in the comments when I was refactoring this that you always talk about a window of 5 pixels wide, but the window width is set to 4.  Is that a mistake?

HealthyPear:
> @HealthyPear I noticed in the comments when I was refactoring this that you always talk about a window of 5 pixels wide, but the window width is set to 4. Is that a mistake? Or is that the sliding window for peak detection, not the integration window?

It should be correct because the width is added to the peak index, so e.g. when the peak falls before the readout window I set the predicted peak position to the sample with index 0 and then using a width of 4 you get the window [0,1,2,3,4] which is long 5 samples

kosack:
Ah yes, In `extract_around_peak()`:

```py3
    start = peak_index - shift
    end = start + width
```

But that is quite confusing, and sounds like a bug (in extract_around_peak).  @watsonjj  is this the expected behavior? It means when the user asks for a window width of 4, it actually integrates over 5 samples. 

e.g. if start is at sample index 3, and width is 4, end is at sample index 7, which means the window goes from [3,7] inclusive and is 5 samples wide.

kosack:
Ok, after discussing with @HealthyPear , I've fixed the window widths: they should have been 5, not 4.

HealthyPear:
Ok just to summarize: all widths in this image extractor should be increased by 1 !

ctapipe/image/extractor.py
```diff
@@ -752,25 +715,23 @@ def _apply_first_pass(self, waveforms, telid, selected_gain_channel):
         # Since we have to add 1 sample on each side, window_shift will always
         # be (-)1, while window_width will always be window1_width + 1
         # so we the final 5-samples window will be 1+3+1
-        window_widths = np.full_like(startWindows, width + 1)
-        window_shifts = np.full_like(startWindows, 1)
+        window_width = width + 2
```
MaxNoe, line 718: Before it was width +1, now width + 2?

ctapipe/image/extractor.py
```diff
@@ -916,29 +873,43 @@ def _apply_second_pass(
...
         # window along each waveform
 
         # Now the definition of peak_index is really the peak.
-        # We have to add 2 samples each side, so the shist will always
+        # We have to add 2 samples each side, so the shift will always
         # be (-)2, while width will always end 4 samples to the right.
         # This "always" refers to a 5-samples window of course
-        window_widths = np.full_like(predicted_peaks, 4, dtype=np.int64)
-        window_shifts = np.full_like(predicted_peaks, 2, dtype=np.int64)
+        window_width_default = 5
```
MaxNoe, line 885: before 4, now 5

PR #1357, opened 2020-06-04
Dl1 event source

LukasNickel:
**Changes**
This provides a DL1EventSource, so that we can read dl1 files. Using this should be pretty much equivalent to the SimtelEventSource (or any other event source). 

Besides that, merged files are supported by changing `source.obs_id` to `source.obs_ids`. This is a list with one entry for single runs. This is a breaking change as `source.obs_id` is removed for the base class as well.

**Noteable things:**
- Tests construct a dl1 file by calling the stage-1 tool on a simtel file.
- Parameters and images can both be read. True dl1 information is included with the container changes from #1495. The unit tests do not cover this case as the test file is missing simulated images.
- Pointings are matched based on observation time. At some point it would be desirable to match based on obs_id, but that information is not present in the files right now. (#1484) Currently a bisect search is performed to find the closest observation time which might not work for merged simulations with different pointings (as performed for the IRF construction to increase statistics)

**Demo:**
image


MaxNoe:
For a test file, you could just call the stage-1 process on a test simtel file in a fixture

ctapipe/io/dl1eventsource.py
```diff
@@ -0,0 +1,389 @@
...
+            config=config,
+            parent=parent,
+            **kwargs
+        )
+
+        self.file_ = tables.open_file(input_url)
+        self._subarray_info = self._prepare_subarray_info(
+            self.file_.root.configuration.instrument
+        )
+        self._mc_header = self._parse_mc_header(
```
MaxNoe, line 47: This should be optional, we shoudl support observed dl1 files out of the box here

MaxNoe: so something like `if 'simulation' in self.file_.root.configuration`

LukasNickel:
> For a test file, you could just call the stage-1 process on a test simtel file in a fixture

Yeah. But im not sure if that file covers everything we want. In my case the pointing table is empty for example

MaxNoe:
Wait what? For which file is the pointing table empty?

ctapipe/io/dl1eventsource.py
LukasNickel: True. Could just let the parsing function return None if no mcheader table is included?

LukasNickel:
> Wait what? For which file is the pointing table empty?
> 
> That should not be the case for any input file!

The file I processed for testing. The table has 0 rows.
ctapipe-extra/gamma_test.simtel.gz with the stage1 tool

ctapipe/io/dl1eventsource.py
MaxNoe: Not as long as you directly give it the `simulation` group here, this will raise an AttributeError or Keyerror before you enter the function

LukasNickel: Same for the event wise information

MaxNoe:
No for me:

Screenshot_20200604_123617

MaxNoe:
Anyhow, this is a prod2 file, we should use a more recent test file. 
E.g. the ones provided by Johan.

LukasNickel:
> Not for me, i see the expected one row per telescope:
> 
> Screenshot_20200604_123617

My bad. I was using an old test file. That one was not properly written.
This file includes it, so I can implement the pointing.

ctapipe/io/dl1eventsource.py
```diff
@@ -0,0 +1,389 @@
...
+
+    def close(self):
+        self.file_.close()
+
+    @staticmethod
+    def is_compatible(file_path):
+        """
+        Implementation needed!
+        """
+        return False
```
MaxNoe, line 62: The first 8 bytes of all HDF5 files are `b'\x89HDF\r\n\x1a\n'`, which is an easy way of make sure it's hdf5.
Then you could check for the metadata we write in stage1.

MaxNoe: https://www.loc.gov/preservation/digital/formats/fdd/fdd000229.shtml

MaxNoe: So the first step would be:

```
with open(file_path, 'rb') as f:
     magic_number = f.read(8)
if magic_number != b'\x89HDF\r\n\x1a\n':
    return False
```

ctapipe/io/dl1eventsource.py
```diff
@@ -0,0 +1,389 @@
...
+        """
+        return True
+
+    @property
+    def subarray(self):
+        return self._subarray_info
+
+    @property
+    def datalevels(self):
+        return (DataLevel.DL1_IMAGES, DataLevel.DL1_PARAMETERS)
```
MaxNoe, line 77: Should be depending on what is actually in the file

ctapipe/io/dl1eventsource.py
```diff
@@ -0,0 +1,389 @@
...
+        return set(self.file_.root.dl1.event.subarray.trigger.col("obs_id"))
+
+    @property
+    def mc_header(self):
+        return self._mc_header
+
+    def _generator(self):
+        try:
+            yield from self._generate_events()
+        except EOFError:
```
MaxNoe, line 90: This should not really happen with hdf5 files, that's copy paste from the simtel source I assume

ctapipe/io/dl1eventsource.py
```diff
@@ -0,0 +1,389 @@
...
+        self.file_.root.configuration.simulation.run
+        """
+        # We just assume there is only one row?
+        # Thats kind of what the SimTelEventSource does
+        # If mixed configurations are in one file, it gets problematic anyway, right? -> subarray layout?
+        header = header_table.row
+        return MCHeaderContainer(
+            corsika_version=header["corsika_version"],
+            simtel_version=header["simtel_version"],
+            energy_range_min=header["energy_range_min"] * u.TeV,
```
MaxNoe, line 169: always use `u.Quantity(value, unit)` instead of multiplication

MaxNoe: If value is an array, also add `copy=False`

ctapipe/io/dl1eventsource.py
```diff
@@ -0,0 +1,389 @@
...
+        data.meta["input_url"] = self.input_url
+        data.meta["max_events"] = self.max_events  # Does this have any effect?
+        data.mcheader = self._mc_header
+
+        # loop array events
+        for counter, array_event in enumerate(self.file_.root.dl1.event.subarray.trigger):
+            # this should be done in a nicer way to not re-allocate the
+            # data each time (right now it's just deleted and garbage
+            # collected)
+            data.r0.tel.clear()
```
MaxNoe, line 220: this should not be necessary in every iteration

ctapipe/io/dl1eventsource.py
```diff
@@ -0,0 +1,389 @@
...
+
+            data.count = counter
+
+            obs_id = array_event['obs_id']
+            event_id = array_event['event_id']
+            time = array_event['time']
+            data.index.obs_id = obs_id
+            data.index.event_id = event_id
+
+            # where returns an iterator. Is there a way to directly get the row?
```
MaxNoe, line 236: yes, calling `next` on the iterator

ctapipe/io/dl1eventsource.py
```diff
@@ -0,0 +1,389 @@
...
+        and
+        self.file_.root.configuration.instrument.subarray.layout
+        tables.
+
+        Returns
+        -------
+        SubarrayDescription :
+            instrumental information
+        """
+        available_optics = instrument_description.telescope.optics.iterrows()
```
kosack, line 110: You can make this whole function much simpler if you use astropy table, which is how it was intended to be used.  ll Instrument classes have a `from_table()` method to support creating a full SubarrayDescription without having to parse anything. 

For example: 

```py3
cam_table = Table.read(file, "/instrument/telescope/camera/LSTCam")
geom = CameraGeometry.from_table(cam_table)
```
In that way, all the headers are also read correctly, and units applied, etc.   I had started making a `SubarrayDescription.from_hdf5()` method to do all that, but got sidetracked at some point. I can check if I still have the code somewhere, or I can prepare something for you.

ctapipe/io/dl1eventsource.py
```diff
@@ -0,0 +1,389 @@
...
+            for telescope_event in telescope_events:
+                obs_id = telescope_event['obs_id']
+                event_id = telescope_event['event_id']
+                teltrigger_time = telescope_event['telescopetrigger_time']
+                tel_id = telescope_event['tel_id']
+
+                self._fill_telescope_pointing(data, teltrigger_time)
+
+                # bc of indexing in the table (1->001, 12->012, 123->123)
+                index_id = ("000" + str(tel_id))[-3:]
```
MaxNoe, line 269: use `f'{tel_id:003d}'` here

ctapipe/io/dl1eventsource.py
kosack: It means mixing a bit pytables and astropy.table, but that's also how we wrote it (and astropy.table adds a special metadata dataset that is handled automatically by this function, and includes astropy units, in a JSON blob)

ctapipe/io/dl1eventsource.py
kosack: also you don't need to do this at all.  There is a `ctapipe.io.TableReader` that turns a table back into the relevant objects, row by row.  It  should not be done manually.  see:
*  https://cta-observatory.github.io/ctapipe/api/ctapipe.io.HDF5TableReader.html?highlight=hdf5tablereader#ctapipe.io.HDF5TableReader 
* or the tutorial here:
https://cta-observatory.github.io/ctapipe/examples/table_writer_reader.html#Reading-one-row-at-a-time:

ctapipe/io/dl1eventsource.py
```diff
@@ -0,0 +1,389 @@
...
+
+            # where returns an iterator. Is there a way to directly get the row?
+            mc_info_iterator = self.file_.root.simulation.event.subarray.shower.where(
+                f"(obs_id=={obs_id})&(event_id=={event_id})"
+            )
+            mc_info = validate_single_result_query(mc_info_iterator)
+            if mc_info:
+                data.mc = MCEventContainer(
+                    energy=mc_info['true_energy'],
+                    alt=mc_info['true_alt'],
```
kosack, line 244: also here, do not do this manually, use a HDF5TableReader.

kosack:
I notice in the HDF5TableReader docstring, we mention:

 Todo:
 - add ability to synchronize reading of multiple tables on a key
 - add ability (also with TableWriter) to read a row into n containers at
       once, assuming no naming conflicts (so we can add e.g. event_id)

at least the second one may be needed (or else it may already work to call read() multiple times per container if more than one container is serialized into the same row). I'm also not sure if we correctly implemented the optional Container prefix

ctapipe/io/dl1eventsource.py
LukasNickel: Sounds good!
I'll try it out.

LukasNickel:
> I notice in the HDF5TableReader docstring, we mention:
> 
> Todo:
> 
>     * add ability to synchronize reading of multiple tables on a key
> 
>     * add ability (also with TableWriter) to read a row into n containers at
>       once, assuming no naming conflicts (so we can add e.g. event_id)
> 
> 
> at least the second one may be needed (or else it may already work to call read() multiple times per container if more than one container is serialized into the same row). I'm also not sure if we correctly implemented the optional Container prefix

There is no way to read a specific row of a table with the HDF5TableReader, right?
This would be needed to read the telescope events, that match to the observed shower.
Or is that what you mean with synchronized reading?

ctapipe/io/dl1eventsource.py
LukasNickel: Ill have a look into that.
Right now it seems that the missing `from_table` methods are in the way, so I will do that first

kosack:
> There is no way to read a specific row of a table with the HDF5TableReader, right?
This would be needed to read the telescope events, that match to the observed shower.
Or is that what you mean with synchronized reading?

Not currently, but it would not be too hard to add.  But I think it shoudl not be needed - what I meant by synchronized reading is that the events are sorted by event_id already, so all you have to do is seek forward until you get the event you want.  Similar to what I showed in the example above . I don't think you need random access (even if you can with HDF5).    Basically, it would looke like this:

1. read trigger entry for the next event, which tells you what telescopes have data in that event
2. from each telescope in the event, read one row from the corresponding telescope table, after which you have read all the event data
3. move to step 1

So all you have to do is keep track of the current row in each table (which the HDF5TableReader already does).  

For exception cases where there is for some reason an extra event inserted somewhere that breaks the synchronization (not sure how that could happen), you could also verify that the event_ids match, and if not seek forward until they do, or implement some recovery procedure. But that can come later.    If we cannot assume the rows are sorted in all tables, we could use the built-in pytables index on the event_id field to ensure we loop through them in order in the HDF5TableWriter (but I think that is not necessary).

ctapipe/io/dl1eventsource.py
LukasNickel: Sounds good!

Since the OpticsDescription has no `from_table` method yet, I am looking into that right now. The CameraGeometries each have their own table, whereas the OpticDescriptions are all stored in one table. So one would need to implement a `from_row` function instead.
Am I missing something?

LukasNickel:
Sounds reasonable to assume.
In my head the tables were all arbitrarily sorted and of uncorrelated length.
Didn't think the synchronization was that easy!

kosack:
> Sounds reasonable to assume.
In my head the tables were all arbitrarily sorted and of uncorrelated length.
Didn't think the synchronization was that easy!

Hopefully it should be! But if we run into any problems, we can just enable the `--write-index-tables` option by default and then that allows efficient searching by event_id and tel_id, and we could modify the TableReader to support that. I think it should not be necessary though (I hope).

ctapipe/io/dl1eventsource.py
kosack: Yes, I guess it would work pretty much like the OpticsDescription.from_name() function. In fact we could just modify that to  allow one to pass in a table - right now it reads from the default table in ctapipe-extra as a string, but could also accept a table directly. 

Something like:
```py3
optics_table=Table.read(outfile, "/instrument/telescope/optics")

od = OpticsDescription.from_name(
    name="SST-GCT", 
    optics_table=optics_table)
)
```

LukasNickel: Thats even better!

kosack: see #1358

ctapipe/io/dl1eventsource.py
```diff
@@ -0,0 +1,389 @@
...
+        # Will there be one of "e_fl" or "fl" in the columns?
+        # This will only work if "e_fl" is available
+        optic_descriptions = {}
+        for optic in available_optics:
+            optic_description = OpticsDescription(
+                name=optic['name'].decode(),
+                num_mirrors=optic['num_mirrors'],
+                equivalent_focal_length=u.Quantity(
+                    optic['equivalent_focal_length'],
+                    u.m,
```
vuillaut, line 124: The unit could be read from the table instead of assumed

ctapipe/io/dl1eventsource.py
```diff
@@ -0,0 +1,389 @@
...
+            )
+            optic_descriptions[optic['description'].decode()] = optic_description
+
+        tel_positions = {}
+        tel_descriptions = {}
+        for telescope in available_telescopes:
+            tel_positions[telescope['tel_id']] = (
+                telescope['pos_x'],
+                telescope['pos_y'],
+                telescope['pos_z'],
```
vuillaut, line 140: the positions would need an unit too
```
np.array([telescope['pos_x'],
                telescope['pos_y'],
                telescope['pos_z']]) * available_telescopes['pos_x'].unit
```

LukasNickel: Good catches, but I don't think there is much point in reviewing the code right now. There is a lot, that is going to change once the container IO works the way we want it to.

ctapipe/io/dl1eventsource.py
```diff
@@ -0,0 +1,424 @@
...
+        if params and images:
+            return (DataLevel.DL1_IMAGES, DataLevel.DL1_PARAMETERS)
+        elif params:
+            return (DataLevel.DL1_PARAMETERS,)
+        elif images:
+            return (DataLevel.DL1_IMAGES,)
+
+    @property
+    def obs_id(self):
+        return set(self.file_.root.dl1.event.subarray.trigger.col("obs_id"))
```
MaxNoe, line 121: This shouldn't be a set, should it? Or it should be a set every where.

ctapipe/io/dl1eventsource.py
```diff
@@ -0,0 +1,424 @@
...
+
+    @property
+    def mc_headers(self):
+        return self._mc_headers
+
+    @property
+    def obs_ids(self):
+        return self.mc_headers.keys()
```
MaxNoe, line 129: `obs_ids` should also work for observed data, this is not mc specific

ctapipe/io/dl1eventsource.py
```diff
@@ -0,0 +1,424 @@
...
+
+    def _generator(self):
+        yield from self._generate_events()
+
+    def _prepare_subarray_info(self):
```
MaxNoe, line 134: This could now use the methods added here #1405

ctapipe/io/dl1eventsource.py
LukasNickel: Not sure if I understand you correctly. The table includes the obs_id multiple times. For files with only a single obs_id you can just take the first entry, but for merged files this wont work. Or do you just want to cast it back to a list or tuple?

ctapipe/io/dl1eventsource.py
LukasNickel: Listed as ToDo. I wanted to wait until that Pr is merged

ctapipe/io/dl1eventsource.py
MaxNoe: I think we have not yet discussed the possibility of having multiple obs ids in the same DL1 file. The current API clearly expects a single int `obs_id`. 

I see we might want to be able to deal with merged files, but I think `obs_id` should be the one of the current event? 

Not sure. Having a set here is unexpected to me.

ctapipe/io/dl1eventsource.py
LukasNickel: This might actually be an artifact. I dont see how this would be used over the obs_id property

ctapipe/io/dl1eventsource.py
LukasNickel: I see the issue.
But for a single event we have `event.index.obs_id` or not?

kosack: The DL1 format is indeed designed to allow multiple obs_ids in the same file (hence why the simulation config is a table, not header info).  However, the idea was that at least now, you process single runs, and then we would create a "ctapipe-dl1-merge" tool that could join files.  That at least allows the logic to be a bit simpler.    However, I guess that reading the DL1 files will need eventually to support merged files.

To answer the last question: Each event should have a single obs_id, yes. It's the obs_id of the original un-merged file (so the simulation id for MC files) That allows one to uniquely identify an event with (obs_id, event_id).

LukasNickel: > Therefore having the obs_id property of a DL1EventSource is either:
> 
>     1. not needed, as you always can get the obsid from the event (event.index.obs_id)
>     2. could just be a `set` or `list` 

Actually the base class lists the property `obs_id` as an integer and the name suggests one as well imo.
Currently (with 45a00b4)  there are two properties in the dl1 source: `obs_ids` returning a list and `obs_id` returning the single entry if there is only one id. But that feels a bit hacky as well.

kosack: Probably we should change it in the base class as well. In principle we can always have multiple merged runs. It should always return a tuple or list, even if a single obs_id, and perhaps the property should be `source.obs_ids` just like `source.datalevels` (which also is a tuple)

LukasNickel: Agreed.

How does this relate to simtel files? Are there any merged files and would these list both obs ids in the file header?

kosack: now that we have now, but in principle we could have merged simtel files (we decided not to support them so far in SimTelEventSource, since there were no examples).  I think there exists a tool for simtelarray that can merge them, however.

LukasNickel: So we just make `obs_id` a tuple/list of one entry for the `SimtelEventSource` and just dont support it for now?

kosack (changes requested):
## Review from @kosack 

### General
*  Looks good! Once a few things are fixed, this will be very useful!
* Please merge the latest changes from the upstream master so the changes to read/write the SubarrayDescription (#1405) don't appear here
* for the future, not this PR perhaps: might be nice to add a round-trip test, where we write out DL1 data and read it back in. That could wait until I've refactored the DL1 tool into components, making it much easier. 

### Warnings / log messages

When reading a file, we get a lot of warnings like the following (many hundreds of them, if the file has many telescopes).  This even overwhelms the Jupyter Lab io system (it gives an error of too high a rate of log messages).  If it's hard to suppress them, perhaps we could just not emit them at the WARNING level, but in DEBUG or something? (so when the logger is set to INFO, you don't get spammed with these)
```
Table /dl1/event/telescope/parameters/tel_014 has column leakage_intensity_width_2 that is not in container MorphologyContainer. It will be skipped.
Table /dl1/event/telescope/parameters/tel_014 has column concentration_cog that is not in container MorphologyContainer. It will be skipped.
Table /dl1/event/telescope/parameters/tel_014 has column concentration_core that is not in container MorphologyContainer. It will be skipped.
Table /dl1/event/telescope/parameters/tel_014 has column concentration_pixel that is not in container MorphologyContainer. It will be skipped.
Table /dl1/event/telescope/parameters/tel_014 has column intensity_max that is not in container MorphologyContainer. It will be skipped.
Table /dl1/event/telescope/parameters/tel_014 has column intensity_min that is not in container MorphologyContainer. It will be skipped.
Table /dl1/event/telescope/parameters/tel_014 has column intensity_mean that is not in container MorphologyContainer. It will be skipped.
Table /dl1/event/telescope/parameters/tel_014 has column intensity_std that is not in container MorphologyContainer. It will be skipped.
Table /dl1/event/telescope/parameters/tel_014 has column intensity_skewness that is not in container MorphologyContainer. It will be skipped.
Table /dl1/event/telescope/parameters/tel_014 has column intensity_kurtosis that is not in container MorphologyContainer. It will be skipped.
Table /dl1/event/telescope/parameters/tel_014 has column peak_time_max that is not in container MorphologyContainer. It will be skipped.
Table /dl1/event/telescope/parameters/tel_014 has column peak_time_min that is not in container MorphologyContainer. It will be skipped.
Table /dl1/event/telescope/parameters/tel_014 has column peak_time_mean that is not in container MorphologyContainer. It will be skipped.
Table /dl1/event/telescope/parameters/tel_014 has column peak_time_std that is not in container MorphologyContainer. It will be skipped.
Table /dl1/event/telescope/parameters/tel_014 has column peak_time_skewness that is not in container MorphologyContainer. It will be skipped.
Table /dl1/event/telescope/parameters/tel_014 has column peak_time_kurtosis that is not in container MorphologyContainer. It will be skipped.
```
For each table.  I think these are expected so shouldn't be warnings: they are just the columns of other containers that are in a table that serialized multiple containers.  Is there a way we can suppress them and only show ones that are really not in *any* container?

### leaking filehandles

After opening a DL1 file and looping over events and then exiting, I notice the HDF5 is opened at least  2 to 40+ times (perhaps dep on number of telescopes or events?), and more if the source is looped over more than onece (without re-constructing it).  Is that expected?. I guess it should only retain one filehandle (this may be a problem with HDF5TableReader - perhaps we can allow it to pass a handle, rather than open a file?)  
```
Closing remaining open Closing remaining open files:/Users/kkosack/Projects/CTA/Working/ctapipe/examples/gammas2.dl1.h5...done/Users/kkosack/Projects/CTA/Working/ctapipe/examples/gammas2.dl1.h5...done
/Users/kkosack/Projects/CTA/Working/ctapipe/examples/gammas2.dl1.h5...done/Users/kkosack/Projects/CTA/Working/ctapipe/examples/gammas2.dl1.h5...done
/Users/kkosack/Projects/CTA/Working/ctapipe/examples/gammas2.dl1.h5...done/Users/kkosack/Projects/CTA/Working/ctapipe/examples/gammas2.dl1.h5...done
/Users/kkosack/Projects/CTA/Working/ctapipe/examples/gammas2.dl1.h5...done/Users/kkosack/Projects/CTA/Working/ctapipe/examples/gammas2.dl1.h5...done
/Users/kkosack/Projects/CTA/Working/ctapipe/examples/gammas2.dl1.h5...done/Users/kkosack/Projects/CTA/Working/ctapipe/examples/gammas2.dl1.h5...done/Users/kkosack/Projects/CTA/Working/ctapipe/examples/gammas2.dl1.h5...done
/Users/kkosack/Projects/CTA/Working/ctapipe/examples/gammas2.dl1.h5...done/Users/kkosack/Projects/CTA/Working/ctapipe/examples/gammas2.dl1.h5...done
/Users/kkosack/Projects/CTA/Working/ctapipe/examples/gammas2.dl1.h5...done/Users/kkosack/Projects/CTA/Working/ctapipe/examples/gammas2.dl1.h5...done/Users/kkosack/Projects/CTA/Working/ctapipe/examples/gammas2.dl1.h5...done
/Users/kkosack/Projects/CTA/Working/ctapipe/examples/gammas2.dl1.h5...done/Users/kkosack/Projects/CTA/Working/ctapipe/examples/gammas2.dl1.h5...done
Users/kkosack/Projects/CTA/Working/ctapipe/examples/gammas2.dl1.h5...done/Users/kkosack/Projects/CTA/Working/ctapipe/examples/gammas2.dl1.h5...done/
Users/kkosack/Projects/CTA/Working/ctapipe/examples/gammas2.dl1.h5...done/Users/kkosack/Projects/CTA/Working/ctapipe/examples/gammas2.dl1.h5...done
/Users/kkosack/Projects/CTA/Working/ctapipe/examples/gammas2.dl1.h5...done/Users/kkosack/Projects/CTA/Working/ctapipe/examples/gammas2.dl1.h5...done
/Users/kkosack/Projects/CTA/Working/ctapipe/examples/gammas2.dl1.h5...done/Users/kkosack/Projects/CTA/Working/ctapipe/examples/gammas2.dl1.h5...done
/Users/kkosack/Projects/CTA/Working/ctapipe/examples/gammas2.dl1.h5...done/Users/kkosack/Projects/CTA/Working/ctapipe/examples/gammas2.dl1.h5...done
/Users/kkosack/Projects/CTA/Working/ctapipe/examples/gammas2.dl1.h5...done/Users/kkosack/Projects/CTA/Working/ctapipe/examples/gammas2.dl1.h5...done/
Users/kkosack/Projects/CTA/Working/ctapipe/examples/gammas2.dl1.h5...done/Users/kkosack/Projects/CTA/Working/ctapipe/examples/gammas2.dl1.h5...done/Users/kkosack/Projects/CTA/Working/ctapipe/examples/gammas2.dl1.h5...done/Users/kkosack/Projects/CTA/Working/ctapipe/examples/gammas2.dl1.h5...done
/Users/kkosack/Projects/CTA/Working/ctapipe/examples/gammas2.dl1.h5...done
/Users/kkosack/Projects/CTA/Working/ctapipe/examples/gammas2.dl1.h5...done
/Users/kkosack/Projects/CTA/Working/ctapipe/examples/gammas2.dl1.h5...done/Users/kkosack/Projects/CTA/Working/ctapipe/examples/gammas2.dl1.h5...done
/Users/kkosack/Projects/CTA/Working/ctapipe/examples/gammas2.dl1.h5...done/Users/kkosack/Projects/CTA/Working/ctapipe/examples/gammas2.dl1.h5...done
/Users/kkosack/Projects/CTA/Working/ctapipe/examples/gammas2.dl1.h5...done
/Users/kkosack/Projects/CTA/Working/ctapipe/examples/gammas2.dl1.h5...done
```

LukasNickel:
> ## Review from @kosack
> ### General
> 
>     * Please merge the latest changes from the upstream master so the changes to read/write the SubarrayDescription (#1405) don't appear here

will do

> 
> 
> ### Warnings / log messages
> 
> When reading a file, we get a lot of warnings like the following (many hundreds of them, if the file has many telescopes). This even overwhelms the Jupyter Lab io system (it gives an error of too high a rate of log messages). If it's hard to suppress them, perhaps we could just not emit them at the WARNING level, but in DEBUG or something? (so when the logger is set to INFO, you don't get spammed with these) 
> For each table. I think these are expected so shouldn't be warnings: they are just the columns of other containers that are in a table that serialized multiple containers. Is there a way we can suppress them and only show ones that are really not in _any_ container?

Oh right, I wanted to set them to debug at some point.
But its probably more useful to check for all the expected columns at once.


> ### leaking filehandles
> 
> After opening a DL1 file and looping over events and then exiting, I notice the HDF5 is opened at least 2 to 40+ times (perhaps dep on number of telescopes or events?), and more if the source is looped over more than onece (without re-constructing it). Is that expected?. I guess it should only retain one filehandle (this may be a problem with HDF5TableReader - perhaps we can allow it to pass a handle, rather than open a file?)

Havent noticed these yet. The TableReader indeed expects a path and opens the file. We could open the file in the EventSource and only pass the handle. Is that what you mean? Or adapt the reader to work with either a file handle or a filename to keep the old behaviour intact. Maybe we would also want to change the Writer then to keep the simmetry?

ctapipe/io/dl1eventsource.py
```diff
@@ -0,0 +1,321 @@
...
+    MorphologyContainer,
+    MCHeaderContainer,
+    MCEventContainer,
+    PeakTimeStatisticsContainer,
+    TimingParametersContainer,
+    TriggerContainer,
+)
+
+
+class DL1EventSource(EventSource):
```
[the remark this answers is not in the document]
LukasNickel: Thanks for the reminder!

kosack:
>  We could open the file in the EventSource and only pass the handle. Is that what you mean? Or adapt the reader to work with either a file handle or a filename to keep the old behaviour intact. Maybe we would also want to change the Writer then to keep the simmetry?

Yes, I think the second options is best: allow a filehandle to be passed, but also accept a filename. And yes, probably doing the same for the HDF5TableWriter would be a good idea, even if generally you don't need to construct more than one of them  as it handles all tables at once).

kosack:
One more minor feature request: can you implement `__len__()` in DL1EventSource to return the number of events in the file? That just makes it possible to have a progress bar when processing the events (e.g. `for event in  tqdm(source):`)

ctapipe/io/hdf5tableio.py
```diff
@@ -424,6 +427,14 @@ def _map_table_to_containers(self, table_name, containers, prefixes):
             for key in tab.attrs._f_list():
                 container.meta[key] = tab.attrs[key]
 
+        # check if the table has additional columns not present in any container
+        for colname in tab.colnames:
+            if colname not in self._cols_to_read[table_name]:
+                self.log.debug(
+                    f"Table {table_name} contains column {colname}"
```
kosack, line 434: need a space after `{colname} `

kosack (approved):

ctapipe/io/dl1eventsource.py
```diff
@@ -0,0 +1,350 @@
...
+                        continue
+                    image_row = next(image_iterators[f"tel_{tel:03d}"])
+                    dl1.image = image_row["image"]
+                    dl1.peak_time = image_row["peak_time"]
+                    dl1.image_mask = image_row["image_mask"]
+                    # ToDo: Find a place in the data container, where the true images can go #1368
+
+                if DataLevel.DL1_PARAMETERS in self.datalevels:
+                    if f"tel_{tel:03d}" not in param_readers.keys():
+                        print("miss", tel)
```
MaxNoe, line 291: print statement

MaxNoe: Should be logging or raising a warning

LukasNickel: Leftovers from debugging. This affects the case where the dl1 file was constructed with only a subset of telescopes (but the trigger table still contains all triggered telescopes).
We don't really log much in the event sources, but it probably doesn't hurt to start somewhere.

LukasNickel:
What functionality is this breaking by the way? Besides the `event_source.obs_id` everything should still be functional, its just new stuff being added if I remember correctly.

MaxNoe:
> Besides the event_source.obs_id everything should still be functional, its just new stuff being added if I remember correctly.

Exactly that. 

An abstract property of `EventSource` was renamed, so this will break all current `EventSource` implementations.

kosack:
was `eventsource.obs_id` ever used before? I don't see anywhere in ctapipe where is relied upon, but perhaps in other io plugins it is used.  In any case, the docstring of the property in `EventSource` should be updated to be `List[int]`

LukasNickel:
> was `eventsource.obs_id` ever used before? I don't see anywhere in ctapipe where is relied upon, but perhaps in other io plugins it is used. In any case, the docstring of the property in `EventSource` should be updated to be `List[int]`

Rarely. The `stage-1` tool accesses it and unit tests check it. External code could use it thought

MaxNoe:
I think we should probably just remove it. 

I introduced it as convenience method, but it ios certainly not general enough to force every event source to implement it.

MaxNoe (changes requested):
The DL1 source should check the `DL1_DATAMODEL_VERSION` of the file and make sure it is supported.

ctapipe/io/dl1eventsource.py
```diff
@@ -105,6 +105,8 @@ def is_compatible(file_path):
                 return False
             if metadata["CTA PRODUCT DESCRIPTION"] != "DL1 Data Product":
```
MaxNoe, line 106: This and the previous check can be simplified to `metadata.get('CTA PRODUCT DESCRIPTION') != "DL1 Data Product"

ctapipe/io/tests/test_dl1eventsource.py
```diff
@@ -18,6 +19,11 @@ def dl1_file():
     return f"{d.name}/testfile.dl1.h5"
 
 
+def test_is_compatible(dl1_file):
```
MaxNoe, line 22: Maybe also test a simtel file to see that `DL1EventSource.is_compatible(simtel_file) == False`

MaxNoe (approved):
Approved, but we should not merge this before 0.8.1 due to the breaking change

MaxNoe:
Checking the current code, there are some places where we rely on `EventSource.obs_id` and where there is no clear way of replacing that, e.g. when writing the histograms in the stage1 tool.

So I would propose the following:

Rename this to `current_obs_id` and document that this should be the `obs_id` of the last event that was yielded by the event source or the first event if no event has yet been yielded.

MaxNoe (changes requested):
Checking the current code, there are some places where we rely on EventSource.obs_id and where there is no clear way of replacing that, e.g. when writing the histograms in the stage1 tool.

So I would propose the following:

Rename this to current_obs_id and document that this should be the obs_id of the last event that was yielded by the event source or the first event if no event has yet been yielded.

LukasNickel:
> Checking the current code, there are some places where we rely on EventSource.obs_id and where there is no clear way of replacing that, e.g. when writing the histograms in the stage1 tool.

If it is only the stage-1 tool, we can also work around it there, e.g. by taking the last events obs_id.
This only matters for simtel files, which should not contain multiple obs_ids anyway, right?

MaxNoe:
Yes, right.

But this probably means that the stage 1 tool is only tested on a file without the histograms, otherwise the test should have failed here.

Can you fix that? So make the stage1 tool get the obs_id from the events and run the tests on a file with histograms?

The lst prod3 file should have the histograms.

kosack:
> Can you fix that? So make the stage1 tool get the obs_id from the events and run the tests on a file with histograms?

yes, you should be able to just use `event.index.obs_id` and not access it from the subarray.

ctapipe/io/dl1eventsource.py
```diff
@@ -0,0 +1,364 @@
...
+        Fill the telescope pointing information of a given event
+        """
+        # Same comments as to _fill_array_pointing apply
+        for tel in data.trigger.tel.keys():
+            if self.allowed_tels and tel not in self.allowed_tels:
+                continue
+            tel_pointing_table = self.file_.root.dl1.monitoring.telescope.pointing[
+                f"tel_{tel:03d}"
+            ]
+            closest_time = np.argmin(
```
maxnoe, line 354: I don't think we should use this approach here.

It will be very expensive to do this for each telescope event.

Maybe better check if there are different pointings and only search for the closest one if there are multiple pointings?

Also, there are data structures that are much better suited to finding something "closest" than calling argmin on the difference.

See for example what I started implementing in #1325

LukasNickel: > Maybe better check if there are different pointings and only search for the closest one if there are multiple pointings?

:+1: 
 
> Also, there are data structures that are much better suited to finding something "closest" than calling argmin on the difference.
> 
> See for example what I started implementing in #1325

I will need to have a look into that.

kosack:
I get an error reading the example DL1 file you provided: 
```
> /Users/kkosack/Projects/CTA/Working/ctapipe/ctapipe/io/dl1eventsource.py(285)_generate_events()
    283                 if self.allowed_tels and i["tel_id"] not in self.allowed_tels:
    284                     continue
--> 285                 data.trigger.tel[i["tel_id"]].time = i["time"]
    286
    287             self._fill_array_pointing(data, array_pointing_finder)

KeyError: 'time'
```

Is that due to a change in the data model? If so we need to check the version number (this file has the CTA PRODUCT DATA MODEL VERSION attribute set to `v1.0.0`

LukasNickel:
> I get an error reading the example DL1 file you provided:
> 
> ```
> > /Users/kkosack/Projects/CTA/Working/ctapipe/ctapipe/io/dl1eventsource.py(285)_generate_events()
>     283                 if self.allowed_tels and i["tel_id"] not in self.allowed_tels:
>     284                     continue
> --> 285                 data.trigger.tel[i["tel_id"]].time = i["time"]
>     286
>     287             self._fill_array_pointing(data, array_pointing_finder)
> 
> KeyError: 'time'
> ```
> 
> Is that due to a change in the data model or a bug fix? If so we need to check the version number (this file has the CTA PRODUCT DATA MODEL VERSION attribute set to `v1.0.0` or at least use `i.get("time", np.nan)` or something

`telescopetrigger_time` got changed to `time` (not sure where).
I thought I just had some very old version as the code was left untouched for a while and this occured after merging the master.
Not sure where that was changed, I will cover both cases.
Using `nan` will screw up the pointing code as that relies on the trigger times until we have `obs_id` to match pointings and observations.

LukasNickel:
This removed the prefix of the `TelescopeTriggerContainers`, changing the key under which the times are saved.
That's probably it.

https://github.com/cta-observatory/ctapipe/commit/0226213540d51f7f5ee07363941366fd12fc0a31

kosack:
> Do we want the prefix or not? Either way, both versions should have different version numbers

The prefix isn't really necessary unless we mix several containers into one table, so I think it's ok to leave it blank for that container.

LukasNickel:
Since we now have two versions of dl1 data and a third one proposed in #1484, we should probably save some test dl1 files somewhere instead of generating a new one with the stage-1 tool everytime.

Is adding files to ctapipe-extra still the way to go?

LukasNickel:
Does anyone know why the python 3.8 build is failing?

kosack:
seems to be missing the file '/tmp/tmprid7j6n8/testfile.dl1.h5', which is generated by the `dl1_file` fixture...  Have you tried re-starting the job? Perhaps it's just a filesystem problem

LukasNickel:
> Does anyone know why the python 3.8 build is failing?

Solved after merging #1501

kosack (dismissed):

LukasNickel:
Needed to add some commits to adapt to the changes on the master.
Would be great if you could review again.

kosack (approved):

ctapipe/io/dl1eventsource.py
```diff
@@ -0,0 +1,427 @@
...
+        images = "images" in self.file_.root.dl1.event.telescope
+        if params and images:
+            return (DataLevel.DL1_IMAGES, DataLevel.DL1_PARAMETERS)
+        elif params:
+            return (DataLevel.DL1_PARAMETERS,)
+        elif images:
+            return (DataLevel.DL1_IMAGES,)
+
+    @property
+    def obs_ids(self):
```
maxnoe, line 157: Cache this? `@lazy_property` from astropy

maxnoe (approved):

ctapipe/io/dl1eventsource.py
maxnoe: Can you add that in a new PR so we don't have to do the approve dance again? @LukasNickel

ctapipe/io/dl1eventsource.py
```diff
@@ -0,0 +1,427 @@
...
+
+        for counter, array_event in enumerate(events):
+            data.dl1.tel.clear()
+            data.simulation.tel.clear()
+            data.pointing.tel.clear()
+            data.trigger.tel.clear()
+
+            data.count = counter
+            data.trigger, data.index = next(events)
+            data.trigger.tels_with_trigger = (
```
maxnoe, line 283: Add this to the subarray? We have a method id list → mask, so it would only make sense to also implement the other way around there.

maxnoe (approved):

ctapipe/io/dl1eventsource.py
LukasNickel: You mean adding a new, general method to the subarray class and use it here?

ctapipe/io/dl1eventsource.py
LukasNickel: sure

ctapipe/io/dl1eventsource.py
maxnoe: yes

PR #1358, opened 2020-06-04
support user table in OpticsDescription.from_name

kosack:
This just allows one to provide a user-specified table to `OpticsDescription.from_name()`, rather than the default (which is a string filename from which to fetch the dataset). 

now you can do:
```py3
table = Table.read("data.dl1.h5", "/configuration/instrument/telescope/optics")
optics = OpticsDescription.from_name("SST-GCT", optics_table=table)
```

useful for reading from an HDF5 file, like in  #1357


LukasNickel:
Screenshot_2020-06-08_10-47-57

Unfortunately this doesn't seem to be quite enough to read the tables written by the stage1 tool.
It looks like some column names (and information inside) have changed since the default table was created.
- tel_description -> description 
- tel_type -> type
- mirror_type -> num_mirrors

This could be a separate issue of course.

kosack:
> Unfortunately this doesn't seem to be quite enough to read the tables written by the stage1 tool.
It looks like some column names (and information inside) have changed since the default table was created.

Strange! Not sure when that happened.  Yes, that will cause some problems. But i can just re-generate the default tables and update ctapipe-extra.  Not sure why it doesn't fail a unit test though

kosack:
Yes, it seems to be a bug - the subarray.to_table() method was not updated after the optics.from_name() method was updated.  So it's inconsistent, and unfortunately means the DL1 files are broken as well! Will have to make a bug-fix release soon.   I'll fix it here

kosack:
Yeah, it's really broken: the `OpticsDescription.from_name` function seems to convert from the old format rather than define it correctly. It still looks for mirror type ="DC" and things like that.

MaxNoe:
Is this superseeded by this: #1405 ?  Or do we still need this?

MaxNoe (approved):

MaxNoe:
Ok, rereading, this seems to be nice to have but the issue is still, that `OpticsTable.from_name` is broken?

LukasNickel:
Yeah, #1405 made it possible to construct the `SubarrayDescription` directly from a hdf file, so that the `from_name` method is no longer needed for that. The method itself is still broken though to my knowledge.

MaxNoe:
> The method itself is still broken though to my knowledge.

Broken for reading the new table but still working for the one in `ctapipe-extra`, right?

LukasNickel:
Yes. The stage-1 tool writes in the new format, the `from_name` method assumes the old one from `ctapipe-extra`.
So its all good as long as you don't give it a table in the new format.

maxnoe (dismissed):

vuillaut (approved):

maxnoe (approved):

vuillaut (approved):

PR #1364, opened 2020-06-11
Add support to read from prefixed columns

### LINKED ISSUE #1360: HDF5TableReader does not support prefixed columns
The stage1 tool collects all parameters from the camera images in one ImageParametersContainer.
While writing, the containers get expanded and prefixes get added to avoid collisions. So (x,y,...) from the HillasParametersContainer maps to (hillas_x, hillas_y, ...) and so forth.
See 
https://github.com/cta-observatory/ctapipe/blob/8c02fb772ec54c200b69721baf9ef20f590cd749/ctapipe/tools/stage1.py#L606-L609
 and 
https://github.com/cta-observatory/ctapipe/blob/8c02fb772ec54c200b69721baf9ef20f590cd749/ctapipe/tools/stage1.py#L716

The HDF5TableReader on the other hand tries to match the column names directly to the container, see https://github.com/cta-observatory/ctapipe/blob/8c02fb772ec54c200b69721baf9ef20f590cd749/ctapipe/io/hdf5tableio.py#L399-L402
Besides not being able to directly read into the ImageParametersContainer(?), the matching also fails when trying to read into the individual containers.

The easiest fix would probably be to add a check for prefixes to the read() Method of the HDF5TableReader.

LukasNickel:
This fixes #1360 in order to work towards a DL1EventSource.

- The `HDF5TableReader.read()` method now has an additional, optional argument `prefix`
- Following the discussion in #1360 three use cases are supported:
    - `prefix=False`: No prefix. This is the default, because per default the writer does not add a prefix as well
    - `prefix=True`: Using the container prefix
    - `prefix="my_prefix"`: Using a custom prefix

This enables the reading of the image parameter containers, that are written by the stage1 tool, but support for reading into multiple containers at once is still missing.

### DIFF
--- a/ctapipe/io/hdf5tableio.py
+++ b/ctapipe/io/hdf5tableio.py
@@ -161,7 +161,7 @@ class HDF5TableWriter(TableWriter):
                     continue
 
                 if col_name in Schema.columns:
-                    self.log.warning(f"Found duplicated column {col_name}, skipping")
+                    self.log.warning("Found duplicated column %s, skipping", col_name)
                     continue
 
                 # apply any user-defined transforms first
@@ -208,10 +208,10 @@ class HDF5TableWriter(TableWriter):
 
                 else:
                     self.log.warning(
-                        f"Column {col_name} of"
-                        f" container {container.__class__.__name__}"
-                        f" in table {table_name}"
-                        " not writable, skipping"
+                        "Column %s of container %s in table %s not writable, skipping",
+                        col_name,
+                        container.__class__.__name__,
+                        table_name
                     )
                     continue
 
@@ -280,8 +280,9 @@ class HDF5TableWriter(TableWriter):
                     row[colname] = value
                 except Exception:
                     self.log.error(
-                        f'Error writing col "{colname}" of'
-                        f' container "{container.__class__.__name__}"'
+                        'Error writing col "%s" of container "%s"',
+                        colname,
+                        container.__class__.__name__
                     )
                     raise
         row.append()
--- a/ctapipe/io/hdf5tableio.py
+++ b/ctapipe/io/hdf5tableio.py
@@ -364,10 +364,10 @@ class HDF5TableReader(TableReader):
 
         self._h5file.close()
 
-    def _setup_table(self, table_name, container):
+    def _setup_table(self, table_name, container, prefix):
         tab = self._h5file.get_node(table_name)
         self._tables[table_name] = tab
-        self._map_table_to_container(table_name, container)
+        self._map_table_to_container(table_name, container, prefix)
         self._map_transforms_from_table_header(table_name)
         return tab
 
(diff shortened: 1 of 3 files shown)


LukasNickel:
Does anyone know why codacy is still complaining?

MaxNoe:
> Does anyone know why codacy is still complaining?

Yes, you changed the signature only of the hdf table reader, not of the base class. Methods should have the same signatures for all base classes.

ctapipe/io/tableio.py
```diff
@@ -168,7 +168,7 @@ def _apply_col_transform(self, table_name, col_name, value):
         return value
 
     @abstractmethod
-    def read(self, table_name: str, container: Container):
+    def read(self, table_name: str, container: Container, prefix=False):
```
LukasNickel, line 171: Isn't this the base method codacy complains about?

MaxNoe: Yes, now I'm confused. The signatures are the same

kosack: it makes no sense.  I think I've had it with codacy - it gives to many strange false positives

kosack: I think the problem is there are two codacy's running: this one comes from Dominik Niese (his username is shown), and it is configured for python2, whereas the one from my account uses python3.  I have no idea how to stop the other one, or to choose which one reports.  It basically complains on any line that is not python2 compatible. Probably we should just kill it and use something else

ctapipe/io/hdf5tableio.py
```diff
@@ -161,7 +161,7 @@ class Schema(tables.IsDescription):
                     continue
 
                 if col_name in Schema.columns:
-                    self.log.warning(f"Found duplicated column {col_name}, skipping")
+                    self.log.warning("Found duplicated column %s, skipping", col_name)
```
MaxNoe, line 164: I like the f-strings better here, it is what we use everywhere else

ctapipe/io/hdf5tableio.py
```diff
@@ -208,10 +208,10 @@ def transform(enum_value):
 
                 else:
                     self.log.warning(
-                        f"Column {col_name} of"
-                        f" container {container.__class__.__name__}"
-                        f" in table {table_name}"
-                        " not writable, skipping"
+                        "Column %s of container %s in table %s not writable, skipping",
```
MaxNoe, line 211: same as above

ctapipe/io/hdf5tableio.py
kosack: I think the only reason to ever use the old %s syntax is for lines of code that are called very often, in which case python does some optimization for logs and warnings where it  doesn't evaluate the string if the log level is lower or warning is disabled (so saves a few cpu cycles).   Here, I agree f-strings are just more readable.

LukasNickel: Yeah, I thought there was some benefit for logging and linters usually warn about that.
I love f-strings though, so I'm happy to revert that.

LukasNickel:
I am pretty sure codacy stopped complaining the last time I visited this.
Can we get this merged or is there anything left to do?

Issue #1368, opened 2020-06-19
Missing container fields for the DL1EventSource


LukasNickel:
While working on the DL1EventSource, I figured that there is some information, that we might want to provide, that is not currently supported by the DataContainer. Namely:

- The DL1Container is missing a `tels_with_data`field, that is present in the lower data levels.
- There is no field for the MCDL1Containers. These are written out by the stage1 tool and include the true images and true parameters (but no true peak time or true image mask). Going with the same structure as with the measured dl1 information, we might want to add a MCDL1Container, that maps the telescope ids, and add a field mc_dl1 to the DataContainer.

Is there any reason to proceed in a different way? Per shower and per run mc information is already stored in the DataContainer anyway.

MaxNoe:
> The DL1Container is missing a tels_with_datafield, that is present in the lower data levels.

I think we said some time ago that these are redundant with the keys of the telescope maps

LukasNickel:
Agreed, but that reasoning applies to lower levels as well, no?

kosack:
There is still a lot of container structure to clean up, so I'm not surprised.  However, for the tels_with_data, outside the Trigger table I think Max is right that that is really redundant at this point.  So you only need to read back into the trigger table, and ensure the tel[] maps have the right data. 

For the MCDL1 parameters, the peak time was added only recently, so we don't write it out (only the true image so far). We do write the MCDL1Containers, that column is only missing when it is not filled.

kosack:
Actually, I just noticed that event/subarray/tels_with_triggger is never written out... it used to work, so I think it's a bug in the stage1.py script.

kosack:
removing the exclude line (670 above) fixes it:
![image](https://user-images.githubusercontent.com/11677812/85127972-b864a780-b230-11ea-9400-3b95c04c3aea.png)

MaxNoe:
You should fix the exclude method to not only do full matches or have a `regex=True` argument

LukasNickel:
Do we want to include code, that reads the "broken" dl1 files or just assume files have been written after this bug fix to the stage1 tool? Or alternatively "fix" the old files manually?

MaxNoe:
As there is no way to recover that information from the files, we should not attempt it.

LukasNickel:
You can perform queries on the telescope tables to look for the array event ids. Assuming that no data was discarded between the trigger and the dl1 calculation that should be the missing information.

kosack:
[...]

PR #1369, opened 2020-06-19
fix exclusion pattern that accidentally removed tels_with_trigger

kosack:
the `dl1/event/subarray/trigger` table was missing the tels_with_data column, which should be a bit pattern.  This was only because of a regexp being set too widely (excluding both the tel and tels_with_trigger column, rather than just the tel column)

### DIFF
--- a/ctapipe/io/tableio.py
+++ b/ctapipe/io/tableio.py
@@ -47,7 +47,7 @@ class TableWriter(Component, metaclass=ABCMeta):
 
     def _is_column_excluded(self, table_name, col_name):
         for pattern in self._exclusions[table_name]:
-            if pattern.match(col_name):
+            if pattern.fullmatch(col_name):
                 return True
         return False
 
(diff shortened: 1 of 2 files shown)


kosack:
see #1368

kosack:
Since the black-reformatting changed a few more lines, the change in question is just this:

```py3
        writer.exclude("dl1/event/subarray/trigger", "tel")
        writer.exclude("dl1/monitoring/subarray/pointing", "tel")
```
to:
```py3
        writer.exclude("dl1/event/subarray/trigger", "^tel$")
        writer.exclude("dl1/monitoring/subarray/pointing", "^tel$")
```

kosack:
> I am very much in favor of not changing the call here but the function to not perform regex matching by default or at least do fullmatch as opposed to just call match or even introducing a regex=False default option to the .exclude method

Agree. I switched to using `fullmatch()` (which I didn't know existed!). That is much better behavior.  If the user wants to exclude many columns by pattern, it's still easy, e.g. by using a regexp like `tel.*`

Issue #1274, opened 2020-04-13
Keras for shower reconstruction


capaths:
Hello, I'm from a team working on a shower reconstruction technique based on neural networks using Keras. We think it'd be nice to be able to use Keras models to create a shower reconstructor.

Currently we have a model that uses as input the calibrated data of an event from prod3b  using a custom [Sequence](https://www.tensorflow.org/api_docs/python/tf/keras/utils/Sequence) ([here](https://github.com/sborquez/gerumo/blob/master/gerumo/data/generator.py)) that allows the neural network to have a custom input and output. Maybe using a Sequence inside a new Reconstructor sub-class could help to have a common input and output in the `predict` method of that class for different reconstructors based on Keras.

MaxNoe:
Related: #1091

I would be reluctant to add one of the deep learning libraries as a dependency.
But as far as I know, all of them support exporting to onnx (at least tf and torch).

So what would you think about using onnx as serialization format?

capaths:
Sorry for the late reply. I've been testing the conversion of our models used for shower reconstruction to the onnx format using [keras2onnx](https://github.com/onnx/keras-onnx) and it seems to be working fine, but I had to make some adjustments to the code (nothing that changes the model architecture though).

ONNX as the serialization format seems like a good idea, as it could allow models from different frameworks, not only Keras, to be loaded and, compared to Tensorflow for example, doesn't seem too heavy to have as a dependency.

A while ago, we started to develop, as a contribution to ctapipe, a "Keras Reconstructor" for the usage of one or more models in a shower reconstruction. We'll be using this reconstructor as a base to an "ONNX Reconstructor" (or maybe "Deep Learning Reconstructor"?) and create a pull request when it's ready (in a couple of days).

PR #1371, opened 2020-06-26
Onnx shower reconstruction

capaths:
## Deep Learning Shower Reconstructor (#1274)

This pull request adds an reconstructor that allows one or more ONNX models to be used for a shower reconstruction. This reconstructor is an abstract class, that's because different algorithms will use different methods to combine the results from the different models and different inputs.

### New dependency

- **onnxruntime**: To load and run inferencing sessions on ONNX models.

### New classes

* **ONNXModel**: Loads a model from a given path to make predictions using the `predict` method, that allows multiple inputs to be used, given ordered or named arguments. The `predict` method returns a list with the model outputs.

* **DeepLearningReconstructor(reco.Reconstructor)**: An abstract Reconstructor class that loads a dictionary of models paths, one model for each kind of camera (FlashCam, LSTCam, ASTRICam, etc.). Then makes a prediction with the corresponding model for each observation made from the event. Once all the predictions have been made, the reconstructor will take all the outputs and merge them into one ``ReconstructedShowerContainer``.

#### Abstract methods

The abstract methods and properties defined for the DeepLearningReconstructor are:

- `supported_cameras` (property): Supported camera types for this algorithm (FlashCam, LSTCam, ASTRICam, etc.).
- `_to_input(self, event, tel_id, cam_name)`: Transforms an event observation (given by the ``tel_id``) into the input that the model will use.
- `_reconstruct(self, models_outputs)`: Receives all the predictions from the different telescopes and uses them to create a new ``ReconstructedShowerContainer`` with the reconstructed data. This value will be returned by the `predict()` method.

### Tests

For testing purposes I added in the `reco` tests directory another directory with three models, each one for different camera types (FlashCam, ASTRICam and LSTCam) with two inputs and outputs to test which are used in the DeepLearningReconstructor's unit tests.


maxnoe (changes requested):
A good starting point I think, but I am not sure yet where to go with this.

It seems *too* abstract to be useful. 

As a side note: we don't put test data inside this repository, you should remove the test data and also make sure they are not in the git history.

Test data is kept in `ctapipe-extra` and downloaded from a server automatically when running tests (using `get_dataset_path`)

TjarkMiener:
@capaths @maxnoe Is this PR still in progress and active? Can I help finalising/testing it? I have onnx models (mono and/or stereo full event reconstruction) at hand that give reasonable performances.

maxnoe:
@TjarkMiener we are currently finalizing the Initial version of the classical machine learning module.

I think we should probably do a work meeting discussing requirements and approaches for including deep learning application into ctapipe.

We should probably introduce a plugin system like for the eventsources also for reconstructions at several points in the pipeline for different kinds of models (predicting from waveforms, from dli images plus peak times, models using all telescopes at once, mono predictions that are later combined, ....).

There are a lot of possibilities and we should discuss what to support in a first iteration.

I am mostly for having an initial implementation to discuss but in this case I think we should devote some time to planning how to best integrate and what to support directly in ctapipe and what better lives in plugins.

kosack:
Shall we close this now that it is possible to have Reconstructor plugins that are external to ctapipe?

maxnoe:
Yes, closing this. We discussed in the datapipe meeting that we are doing this for now via the plugin system.

Anyone wanting to contribute and facing issues with the current reconstructor API, please join the discussion in #2291

Issue #1367, opened 2020-06-17
HDF5TableReader can not setup multiple containers for one table


LukasNickel:
(Following up the discussion in #1357)
This is already mentioned in the docstring of the HDF5TableReader, but will probably need to be solved in order to get the DL1EventSource working (without hacks).

Currently the `read()` method takes only one container and proceeds to then map the table columns to the container columns. After that, `_cols_to_read` contains only the columns that are in the table and match the container fields. This is tied to the table however, a second call to `read()` with another container will not work, because `_cols_to_read` still contains the columns matching the first container and does not get updated.

Example (requiring #1364 in order to read the dl1 table):
![Screenshot_2020-06-17_12-37-44](https://user-images.githubusercontent.com/15281090/84888938-583afd80-b098-11ea-8d0e-14b3536d8b95.png)

This means, that right now the only ways to read from the parameters table in a dl1 file would be:
- Manually reset `_cols_to_read` every time you want to iterate over any of the generators
- Setup multiple readers, one per container class

If we still want that 1 to 1 mapping of columns to fields in order to avoid the overhead of testing each column every time, we could maybe put an additional layer to the `_cols_to_read` dictionary. Right now it is constructed as `{table_name: [list_of_columns]}`. It could be expanded to `{table_name: {container_name/prefix name: [list_of_columns]}}` or smth similar.

Or we do not support multiple calls to `read()` on the same table at all and require a list of all containers to be handed on the single call. Potentially also in a second function like `read_multiple()` to make sure `read()` still works as expected everywhere.

Part of the question probably is whether we want multiple containers in one table, but without prefixes. The parameters table contains prefixes everywhere, but there might be reasons to write multiple containers with overlapping columns (like event _id ) leading to naming conflicts.

kosack:
This comes from the fact that we refactored TableWriter to support multiple containers, but didn't fix TableReader in the same way.  I think the best thing would be just to do the same as in TableWriter, and accept a list, and loop over it when building the scheme.

kosack:
Another way I had originally thought to implement this in both Reader and Writer classes was to just make a CompoundContainer class that looks like a single container, but contains a list of them (then the logic is in there, not in the Reader/Writer), but I'm not sure that ends up being simpler.

LukasNickel:
[...]

PR #1374, opened 2020-07-07
Add support to read multiple containers from one table at once

LukasNickel:
This allows to provide a list of containers to the `HDF5TableReader.read()` function to directly read a table into multiple containers without having to set up a new reader for each container.
The main reason for this are the dl1 parameters tables in the dl1 files. 

This makes use of the fact, that a container has two prefixes: `container.prefix` and `container.container_prefix` to avoid the case of an empty `container.prefix`. This might affect the edge case where multiple containers of the same kind (e.g for different cleaning levels) are written into one table. I am not sure if that is something we ever want to do though. 

Related discussion: #1367

### DIFF
--- a/ctapipe/io/hdf5tableio.py
+++ b/ctapipe/io/hdf5tableio.py
@@ -446,8 +446,10 @@ class HDF5TableReader(TableReader):
             of containers.
         """
 
+        return_iterable = True
         if isinstance(containers, Container):
             containers = (containers, )
+            return_iterable = False
 
         if prefixes is False:
             prefixes = ["" for container in containers]
@@ -480,10 +482,10 @@ class HDF5TableReader(TableReader):
                     container[fieldname] = self._apply_col_transform(
                         table_name, colname, row[colname]
                     )
-            if len(containers) == 1:
-                yield containers[0]
-            else:
+            if return_iterable:
                 yield containers
+            else:
+                yield containers[0]
             row_count += 1
 
 
--- a/ctapipe/io/hdf5tableio.py
+++ b/ctapipe/io/hdf5tableio.py
@@ -402,7 +402,9 @@ class HDF5TableReader(TableReader):
                 else:
                     colname_without_prefix = colname
                 if colname_without_prefix in container.fields:
-                    self._cols_to_read[table_name][container.container_prefix].append(colname)
+                    self._cols_to_read[table_name][container.container_prefix].append(
+                        colname
+                    )
                 else:
                     self.log.warning(
                         f"Table {table_name} has column {colname_without_prefix} that is not in "
@@ -426,7 +428,7 @@ class HDF5TableReader(TableReader):
             for key in tab.attrs._f_list():
                 container.meta[key] = tab.attrs[key]
 
-    def read(self, table_name, containers, prefix=False):
+    def read(self, table_name, containers, prefixes=False):
         """
         Returns a generator that reads the next row from the table into the
         given container. The generator returns the same container. Note that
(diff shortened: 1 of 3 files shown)


LukasNickel:
Codacy seems to be another false positive

MaxNoe:
You should not use `container_prefix`, this is just the class variable that provides the default for the actual `prefix`.

This is needed due to how `__slots__` work with class variables.

LukasNickel:
> You should not use `container_prefix`, this is just the class variable that provides the default for the actual `prefix`.
> 
> This is needed due to how `__slots__` work with class variables. Maybe we shoud rename it to `default_prefix`

How would you differentiate the containers then? All of them could be provided with an empty prefix.

MaxNoe:
> How would you differentiate the containers then? All of them could be provided with an empty prefix.
Although that might be a rather obscure use case.

Why do you need to? A list of containers should be sufficient to read the columns in order or am I missing something?

LukasNickel:
> Why do you need to? A list of containers should be sufficient to read the columns in order or am I missing something?
> That would also solve the issue with the same container type saved twice.

That works as long as you only call `read`once. You can however call it multiple times on the same table but with different container types and the columns to read are saved in between. We can debate if that is really needed, but it is something that is possible with the current code.

kosack:
I guess it should work like this: let's say you have 2 sets of Hillas Parameters, from a loose and a tight cleaning, and you have them both written with prefixes of `hillas_loose` and `hillas_tight` (so column names like `hillas_loose_length`)
```py3
hillas_loose = HillasParametersContainer()
hillas_loose.prefix = "hillas_loose"
hillas_tight = HillasParametersContainer()
hillas_tight.prefix = "hillas_tight"

 with HDF5TableReader(f.name) as reader:
            generator = reader.read(
                "/dl1/params",
                [hillas_loose, hillas_tight],
                prefixes=True,
            )
            hillas_l, hillas_t = next(generator)

```

LukasNickel:
That should work by just switching from `container_prefix` to the actual `prefix`.
I dont recall what exactly prevented me from doing it this way. Right now I can only think of a case where the user might expect
```
hillas = HillasParametersContainer()
hillas.prefix = ""
leakage = LeakageContainer()
leakage.prefix = ""

 with HDF5TableReader(f.name) as reader:
        generator = reader.read(
            "/dl1/params",
            [hillas, leakage],
            prefixes=True,
        )
        hillas_, leakage_ = next(generator)
```
to work on a table, that was written with prefixes disabled. 
This would just work with `prefixes=False` (which is the default anyway) however.

MaxNoe:
> to work on a table, that was written with prefixes disabled.
This would just work with prefixes=False (which is the default anyway) however.

Why would that not work? The column names are unique

LukasNickel:
Not with the current implementation of the table<->container mapping, that requires an unique identifier for each container.

Thinking about it, I dont really see any major disadvantages in going back to a simpler mapping and matching the columns in the 'container filling' loop instead of the setup loop though.

MaxNoe:
> Not with the current implementation of the table<->container mapping, that requires an unique identifier for each container.

I think this is an implementation detail that is not part of the API, so we can just change that to implement the new use case.

LukasNickel:
Totally. API wise this should be fine.

LukasNickel:
> You just have to look at the `.keys` of the containers.

I think that is actually the main problem with this implementation. I was looping over the selected table columns and checked if they are in the container fields when its much easier to do it the other way around. That probably avoids most of the issues.

ctapipe/io/hdf5tableio.py
```diff
@@ -436,45 +436,54 @@ def read(self, table_name: str, container: Container, prefix=False):
...
+                    if prefix:
+                        colname = f"{prefix}_{fieldname}"
+                    else:
+                        colname = fieldname
+                    if colname not in self._cols_to_read[table_name]:
+                        continue
+                    container[fieldname] = self._apply_col_transform(
+                        table_name, colname, row[colname]
+                    )
+            if len(containers) == 1:
```
MaxNoe, line 483: This might be confusing for someone who passes in a list with one entry who then also expects a list of one entry back.

MaxNoe: Maybe just set a member variable storing if a single container was passed in or an iterable of containers

PR #1375, opened 2020-07-09
add __len__ to SimTelEventSource

nbiederbeck:
Add the python magic function `__len__` which is accessibly via `len()` and is e.g. used as a total indicator in `tqdm`.
For this I take the number of showers as stored in the `MCHeaderContainer`.

I mainly have the `ctapipe-stage1-process` script in mind.

### DIFF
--- a/ctapipe/io/simteleventsource.py
+++ b/ctapipe/io/simteleventsource.py
@@ -218,6 +218,9 @@ class SimTelEventSource(EventSource):
     def close(self):
         self.file_.close()
 
+    def __len__(self):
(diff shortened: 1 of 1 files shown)


MaxNoe:
No, this is just wrong.

A `SimTelEventSource` loops over array events. The number of showers has nothing to do with that (only triggered events are produced, the same shower is possibly reused a number of times).

MaxNoe:
`__len__` of this `EventSource` is simply unknown until you completely looped through this, there is no way around.

PR #1385, opened 2020-07-17
Make gain_selector a trait of SimTelEventSource, fixes eventsource plugins

MaxNoe:
The dl1 tool had a gain_selector member / trait that it always passed to the `EventSource`, although this is an implementation detail of the `SimTelEventSource`, not a part of the public API of `EventSource`.

This fixes it by making the `gain_selector` a trait of the `SimTelEventSource`, thus bringing the API of the SimTelEventSource back in line with the base class.

### DIFF
--- a/ctapipe/io/simteleventsource.py
+++ b/ctapipe/io/simteleventsource.py
@@ -11,9 +11,9 @@ from eventio.simtel.simtelfile import SimTelFile
 from traitlets import observe
 from io import BufferedReader
 
-from ..calib.camera.gainselection import ThresholdGainSelector
+from ..calib.camera.gainselection import GainSelector
 from ..containers import EventAndMonDataContainer, EventType
-from ..core.traits import Bool, CaselessStrEnum
+from ..core.traits import Bool, CaselessStrEnum, create_class_enum_trait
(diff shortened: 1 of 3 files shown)


watsonjj:
Maybe this is nicer. However, it does remove the ability to configure the `GainSelector` in a script before it is passed to the `SimtelEventSource`.

MaxNoe:
> Maybe this is nicer. However, it does remove the ability to configure the GainSelector in a script before it is passed to the SimtelEventSource.

No, you can configure it by setting the proper configuration through the traitlets system. This is how it should be done.

It's not about being nicer, it's about having the correct public API, so the EventSources are interchangable

MaxNoe:
Actually, this goes a bit deep, but let's leave it here for that.

This pattern of passing in pre-configured components to other components is happening more often than just here, e.g. see #1392 .

This prevents configuring these components through the config system, which makes the command line tools much less powerfull than they could (and probably were meant to) be.

I'll open other PRs for the rest

MaxNoe:
The big issue here was, that we introduced an incompatible API for the `SimTelEventSource` compared to the one of `EventSource`, that should not happen.

PR #1387, opened 2020-07-17
Make sure event source provides R1 or DL0, fix datalevels of SimTelEventSource

MaxNoe:
This adds a check to the stage1 process that the selected event source does provide DL0 data, which is the needed starting point for this tool.

In the future, as soon as a DL1_IMAGES event source is ready, this could also be accepted, but would need code to skip the calibration and image extraction part to only calculate parameters.

### DIFF
--- a/ctapipe/io/simteleventsource.py
+++ b/ctapipe/io/simteleventsource.py
@@ -224,7 +224,7 @@ class SimTelEventSource(EventSource):
 
     @property
     def datalevels(self):
-        return (DataLevel.R0, DataLevel.R1, DataLevel.DL0)
+        return (DataLevel.R0, DataLevel.R1)
 
     @property
     def obs_id(self):
@@ -388,7 +388,7 @@ class SimTelEventSource(EventSource):
                 )
 
                 self._fill_event_pointing(
-                    data.pointing.tel[tel_id], mc, tracking_positions[tel_id],
+                    data.pointing.tel[tel_id], mc, tracking_positions[tel_id]
                 )
 
                 r0 = data.r0.tel[tel_id]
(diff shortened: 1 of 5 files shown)


watsonjj:
While I generally agree with this change, it is important to note that this would mean different data volume reduction algorithms (which are applied by the Camera Calibrator, R1 -> DL0) cannot by explored in this tool.

Actually, as it currently stands, the `SimtelEventSource` does not fill DL0. It fills up to R1.

MaxNoe:
@watsonjj I just realized this as well and fixed both. The EventSource no requires any of `R1` or `DL0` present and I removed the wrong `DL0` from `SimTelEventSource.datalevels`

MaxNoe:
@kosack @vuillaut Could we merge this? Or shall I open another PR only containing the fix for matplotlib so that at least other PRs can have passing travis results?

ctapipe/io/simteleventsource.py
```diff
@@ -224,7 +224,7 @@ def is_simulation(self):
 
     @property
     def datalevels(self):
-        return (DataLevel.R0, DataLevel.R1, DataLevel.DL0)
+        return (DataLevel.R0, DataLevel.R1)
```
vuillaut, line 227: does this mean `SimTelEventSource` is not supposed to provide  DL0 ever?

vuillaut: I am a bit lost with this line and the title of the PR

MaxNoe: DL0 was included here by mistake, the `SimTelEventSource` does not currently provide it.

I don't know what it is supposed to provide, and this PR is not about that.

Concerning the title: Currently the DL1 tool needs either DL0 or R1 to be present, so I added a check for this.

PR #1389, opened 2020-07-20
Mask for muons, fixes #1373

### LINKED ISSUE #1373: Passing a mask to MuonIntensityFitter?
The calculations to estimate the light in a muon ring are better done on a subset of pixels, usually selected up to a distance relative to the fitted ring. This used to be possible in 0.7 with the now removed MuonLineIntegrate, but its replacement MuonIntensityFitter is missing this feature, as far as I can see.

https://github.com/cta-observatory/ctapipe/blob/262b6aa887a33907901d7b0cd948b9c937ec60b9/ctapipe/image/muon/intensity_fitter.py#L443

moralejo:
Solves https://github.com/cta-observatory/ctapipe/issues/1373


MaxNoe:
Maybe you could also directly add, that the `ctapipe/tools/muon_reconstruction.py` uses this?

ctapipe/image/muon/intensity_fitter.py
```diff
@@ -467,7 +472,7 @@ class MuonIntensityFitter(TelescopeComponent):
     ).tag(config=True)
 
     def __call__(
-        self, tel_id, center_x, center_y, radius, image, pedestal,
+        self, tel_id, center_x, center_y, radius, image, pedestal, mask
```
MaxNoe, line 475: This makes `mask` a required argument. This function is used in the muon reconstruction tool, so this breaks it (and this is also why the tests fail).

Adding a `mask=None` default and making it optional would be best, I think.

MaxNoe:
The other failing tests are due to a matplotlib update and are fixed in #1387.

MaxNoe:
@moralejo if you merge the master again, the tests should pass

moralejo:
> Maybe you could also directly add, that the `ctapipe/tools/muon_reconstruction.py` uses this?

What input does that expect? simtel file? It is not specified in the help.

MaxNoe:
@moralejo It uses the generic `EventSource`, so any event source providing r1/dl0 should work. 

I'll also add the check to this tool that was added to the stage1 process.

PR #1390, opened 2020-07-20
Calculate log likelihood analytically in muon intensity fit

MaxNoe:
This is a large simplification of the pixel likelihood used to fit muons, as it is first simplified analytically which should enhance both speed and numerical precision.


kosack (dismissed):

kosack:
I think this just needs some small merge-conflicts to be resolved and we can merge it .

kosack (approved):

vuillaut (approved):

PR #1392, opened 2020-07-21
Make data volume reducer configurable for CameraCalibrator

maxnoe:
This overhauls how we treat the config of these objects. With the added traits, it should no be possible to completely configure these through the config system, so e.g. for running the dl1 tool.

It was previously impossible to set the image extractor of the camera calibrator through the config system, it had to be repeated as a traitlet for the Tool passing it down.

It was completely impossible to set the data volume reducer.

We might need to give the same treatment to other components.

### DIFF
--- a/ctapipe/calib/camera/calibrator.py
+++ b/ctapipe/calib/camera/calibrator.py
@@ -6,7 +6,7 @@ calibration and image extraction, as well as supporting algorithms.
 import warnings
 import numpy as np
 from ctapipe.core import Component
-from ctapipe.image.extractor import NeighborPeakWindowSum
+from ctapipe.image.extractor import ImageExtractor, NeighborPeakWindowSum
 from ctapipe.image.reducer import DataVolumeReducer
 from ctapipe.core.traits import create_class_enum_trait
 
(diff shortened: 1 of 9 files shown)


kosack:
This looks like a nice improvement.  Will you update  any of the Tools that use CameraCalibrators as well? I think it's mainly the Stage1ProcessorTool, where  there is a bit of repetition since it defines e.g. `image_extractor_type` externally.

maxnoe:
Yes, in fact I already did but did not yet push.

watsonjj:
Is there a danger that someone may be confused why their traitlet has no effect, because the `image_extractor` or `data_volume_reducer ` is explicitly passed as an argument to the calibrator? Would be nice if there was a warning if both were attempted, but I suppose that could be complicated to implement.

maxnoe:
I think this is exactly what you would expect in a config system.

Options given closer to where they are used override defaults from further away.

* User config overrides default config
* commandline config overrides user config
* directly given kwargs override commandline

watsonjj (approved):

kosack (approved):

PR #1398, opened 2020-07-22
Use sparse matrix approach for neighbor window sum

MaxNoe:
Use same neighbor approach like number_of_islands

This results in a speedup from:

```
166.10 ± 66.14 µs
```
to
```
69.12 ± 4.01 µs
```

using:

```py3
import numpy as np
from ctapipe.instrument import TelescopeDescription, SubarrayDescription
import astropy.units as u
from ctapipe.image.extractor import NeighborPeakWindowSum
from timeit import repeat


tel = TelescopeDescription.from_name('LST', 'LSTCam')
geom = tel.camera.geometry
subarray = SubarrayDescription(
    'perf',
    tel_positions={1: [0, 0, 0] * u.m},
    tel_descriptions={1: tel},
)


extractor = NeighborPeakWindowSum(subarray)

waveforms = np.random.normal(size=geom.n_pixels).astype('float32')
selected_gain_channels = np.zeros(geom.n_pixels, dtype=int)


# number_of_islands(geom, mask)
code = '''
extractor(waveforms, 1, selected_gain_channels)
'''

exec(code)

N = 200
times = np.array(repeat(code, number=N, repeat=10, globals=globals()))
print(f'{np.mean(times / N) * 1e6:.2f} ± {np.std(times / N) * 1e6:.2f} µs')
```

### DIFF
--- a/ctapipe/image/extractor.py
+++ b/ctapipe/image/extractor.py
@@ -124,11 +124,12 @@ def neighbor_average_waveform(waveforms, neighbors_indices, neighbors_indptr, lw
     waveforms : ndarray
         Waveforms stored in a numpy array.
         Shape: (n_pix, n_samples)
-    neighbors : ndarray
-        2D array where each row is [pixel index, one neighbor of that pixel].
-        Changes per telescope.
-        Can be obtained from
-        `ctapipe.instrument.CameraGeometry.neighbor_matrix_where`.
+    neighbors_indices : ndarray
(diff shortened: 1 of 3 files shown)


MaxNoe:
Side point: numba jitted functions do not contribute to coverage, we might want to create one instance of the tests with `NUMBA_DISABLE_JIT=1` for the coverage report.

Issue #1272, opened 2020-04-10
Improve pixel likelihood calculation in muon and image/pixel_likelihoods


MaxNoe:
The log likelihood functions first calculate the likelihood and then take the log numerically.

The formulas look like the log-likelihood could be written down directly.

MaxNoe:
The likelihood in the muon intensityfitter was improved in #1390

Issue #1266, opened 2020-04-07
Curiously little references 


maxnoe:
For a rather large scientific codebase, the number of references cited in docs/bibliography.rst is rather small.

I would urge everyone who authored code here to check if relevant references are given in the docstrings of respective methods.

@cta-observatory/ctapipe-core-devs

MaxNoe:
Maybe we should add https://sphinxcontrib-bibtex.readthedocs.io/en/latest/quickstart.html to make it easier to cite stuff?

kosack:
yes that would be great. It would also make it easier to cite reference in papers, since one could just import the bib file.

PR #1403, opened 2020-07-28
Simplify pixel log-likelihood.

nbiederbeck:
in the attempt to work on  #1272:

* simplify the log-likelihood analytically
* discard constants and factors which are not needed in minimization

but now the function `ctapipe.image.pixel_likelihood.poisson_likelihood_gaussian` is basically the same as `ctapipe.image.muon.intensity_fitter.negative_log_likelihood` in #1390 and duplicate code is not good.

Maybe do smth like `ctapipe.image.pixel_likelihood.negative_log_likelihood` and use this in the Muon and ImPACT fitting?

### DIFF
--- a/ctapipe/image/pixel_likelihood.py
+++ b/ctapipe/image/pixel_likelihood.py
@@ -54,7 +54,7 @@ class PixelLikelihoodError(RuntimeError):
 def poisson_likelihood_gaussian(image, prediction, spe_width, pedestal):
     """Calculate negative log likelihood for every pixel.
 
-    Gaussian approximation from de Naurois, p. 22 (between (24) and (25)).
+    Gaussian approximation from [denaurois2009]_, p. 22 (between (24) and (25)).
 
     Simplification:
 
@@ -80,12 +80,6 @@ def poisson_likelihood_gaussian(image, prediction, spe_width, pedestal):
 
         - \\ln{P} = \\ln{θ} + \\frac{(s - μ)^2}{θ}
 
-
-    References
-    ----------
-    - de Naurois, Rolland
-      'A high performance likelihood reconstruction of gamma-rays
-      for Imaging Atmospheric Cherenkov Telescopes' (2009)
 
     Parameters
     ----------
--- a/ctapipe/image/pixel_likelihood.py
+++ b/ctapipe/image/pixel_likelihood.py
@@ -9,7 +9,7 @@ over a number of potential contributing photoelectrons (which is slow).
 At high signal this simplifies to a gaussian approximation.
 
 The full and gaussian approximations are implemented, in addition to a general purpose
-implementation, which tries to intellegently switch 
+implementation, which tries to intellegently switch
 between the two. Speed tests are below:
 
 neg_log_likelihood_approx(image, prediction, spe, ped)
@@ -129,7 +129,7 @@ def neg_log_likelihood_numeric(
 
     likelihood = epsilon
 
-    ns = np.arange(*poisson(np.max(prediction)).ppf(confidence),)
+    ns = np.arange(*poisson(np.max(prediction)).ppf(confidence))
 
     ns = ns[ns >= 0]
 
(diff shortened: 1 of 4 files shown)


MaxNoe:
Yes, just use this function in the moon code

ctapipe/image/pixel_likelihood.py
```diff
@@ -51,42 +51,62 @@ class PixelLikelihoodError(RuntimeError):
...
+    .. math::
+
+        - \\ln{P} = \\ln{θ} + \\frac{(s - μ)^2}{θ}
+
+
+    References
+    ----------
+    - de Naurois, Rolland
+      'A high performance likelihood reconstruction of gamma-rays
+      for Imaging Atmospheric Cherenkov Telescopes' (2009)
```
kosack, line 88: could move this to `references.rst`.  See also #1266

MaxNoe:
Could you also apply the same treatment to the other functions and then use the right one in the muon code?

nbiederbeck:
> Could you also apply the same treatment to the other functions and then use the right one in the muon code?

yes

ctapipe/image/pixel_likelihood.py
nbiederbeck: done

kosack (dismissed):

MaxNoe:
@nbiederbeck Tests are failing due to an import error. Somewhere, probably it still imports one of the older names.

ctapipe/image/pixel_likelihood.py
```diff
@@ -177,13 +177,15 @@ def neg_log_likelihood(
     approx_mask = prediction > prediction_safety
 
     neg_log_l = 0
-    neg_log_l += neg_log_likelihood_approx(
-        image[approx_mask], prediction[approx_mask], spe_width, pedestal,
-    )
+    if np.sum(approx_mask) > 0:
```
MaxNoe, line 180: Use `np.count_nonzero` to check how many items are True in a mask. If you just need a single one, use `np.any`

ctapipe/image/pixel_likelihood.py
```diff
@@ -51,214 +46,152 @@ class PixelLikelihoodError(RuntimeError):
...
+    neg_log_l = 0
+    if np.sum(approx_mask) > 0:
+        neg_log_l += neg_log_likelihood_approx(
+            image[approx_mask], prediction[approx_mask], spe_width, pedestal,
         )
-    if np.any(gaus_pix):
-        like[gaus_pix] = poisson_likelihood_gaussian(
-            image[gaus_pix], prediction[gaus_pix], spe_width, ped
+
+    if np.sum(~approx_mask) > 0:
```
MaxNoe, line 185: This is `not np.any(...)`, right?

nbiederbeck: `not np.all(...)` I guess .. will check if there are speed differences between this one and `np.any(~...)`

nbiederbeck: well, there is a difference: 
```python
In [24]: mask = np.random.randint(0, 2, size=2000).astype('bool')

In [25]: %timeit np.any(~mask)
4.28 µs ± 22.8 ns per loop (mean ± std. dev. of 7 runs, 100000 loops each)

In [26]: %timeit not np.all(mask)
3.23 µs ± 19.3 ns per loop (mean ± std. dev. of 7 runs, 100000 loops each)
```

but am inclined to keep `np.any(~approx_mask)` due to lower mental overhead

MaxNoe: I dont see why having the tilde as unary binary not inside is clearer than not all

nbiederbeck: I added the `if` clauses to not give an empty array into the function. I think `any -> not empty (regarding Trues)` is clear. But  `not all -> not empty (regarding Falses)` seems harder than `any (of the other) -> not empty`. Maybe it's just me, I have trouble putting this into written words

MaxNoe: This is "hot code" which will be called often by a minimizer, if there is not a huge difference in readability, we should choose the faster code.

kosack (changes requested):
coverage.io mentions a few lines that are not tested in any unit test (see the warnings in the code view), but once those are fixed, I think this is ok.

kosack (approved):

maxnoe (approved):

PR #1404, opened 2020-07-31
TailCutsDataVolumeReducer now instantiates TailCutsImageCleaner, fixes #1399

### LINKED ISSUE #1399: Abstract DataVolumeReducer instantiates TailCutsImageCleaner
In it's `__init__` method, the `DataVolumeReducer` instantiates a `TailCutsImageCleaner`, which should belong only to the `TailCutsDataVolumeReducer`, not the base class.

Hckjs:
#1399

### DIFF
--- a/ctapipe/image/reducer.py
+++ b/ctapipe/image/reducer.py
@@ -5,8 +5,8 @@ from abc import abstractmethod
 import numpy as np
 from ctapipe.image import TailcutsImageCleaner
 from ctapipe.core import TelescopeComponent
-from ctapipe.core.traits import IntTelescopeParameter, BoolTelescopeParameter
-from ctapipe.image.extractor import NeighborPeakWindowSum
+from ctapipe.core.traits import IntTelescopeParameter, BoolTelescopeParameter, Unicode
+from ctapipe.image.extractor import ImageExtractor
 from ctapipe.image.cleaning import dilate
 
 __all__ = [
@@ -108,7 +108,10 @@ class TailCutsDataVolumeReducer(DataVolumeReducer):
        with ctapipe module dilate until no new pixels were added.
     3) Adding new pixels with dilate to get more conservative.
     """
-
+    image_extractor_type = Unicode(
+        default_value="NeighborPeakWindowSum", help="Name of the image_extractor"
+        "to be used"
+    ).tag(config=True)
     n_end_dilates = IntTelescopeParameter(
         default_value=1, help="Number of how many times to dilate at the end."
     ).tag(config=True)
@@ -118,14 +121,7 @@ class TailCutsDataVolumeReducer(DataVolumeReducer):
         "normal TailcutCleaning is used.",
     ).tag(config=True)
 
-    def __init__(
-        self,
-        subarray,
-        image_extractor=None,
-        config=None,
-        parent=None,
-        **kwargs
-    ):
+    def __init__(self, subarray, config=None, parent=None, **kwargs):
         """
         Parameters
         ----------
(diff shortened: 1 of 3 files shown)


ctapipe/image/reducer.py
```diff
@@ -131,6 +118,36 @@ class TailCutsDataVolumeReducer(DataVolumeReducer):
         "normal TailcutCleaning is used.",
     ).tag(config=True)
 
+    def __init__(
```
MaxNoe, line 121: This should have the same signature as the base class to make it configurable from outside the of scripts through the config system.

So I'd suggest going the same route as with the gain selector. So create a traitlet `image_extractor_type` and then use `ImageExtractor.from_name` to create it in the `__init__`

Hckjs: The image_extractor type is now not shown in the docs for `TailCutsDataVolumeReducer`, is it normal? Don't know why.

MaxNoe: What do you mean with "in the docs"?

Hckjs: in the docs/_build/html/api after make docs

MaxNoe: You need to add a docstring to the class with a section `Attributes` for that.

kosack:
I see a potential problem with using TailCutsImageCleaner - in fact it's a problem with traitlets.config, not with your code.  The configuration system only allows configuration by class, not by instance, so that means the TailCutsImageCleaner you use in this step would have to have the same configuration as the one used to clean the final images (assuming both the R1→DL0 and DL0→DL1 transforms are done in the same tool).  That could be an issue. One way to solve it would be just to create a subclass, like:

```python3
class DVRTailCutsImageCleaner(TailcutsImageCleaner):
    pass
```

then the cuts would be configured in a section called "DVRTailCutsImageCleaner", while "TailcutsImageCleaner" would be used for the DL1 cleaning. 

Or just use the `tailcuts_clean()` low level function internally, and make the thresholds explicitly a configuration of TailCutsDataVolumeReducer (similar to what was done in TwoPassImageCleaner)

MaxNoe:
@kosack

That is not true. 

This will work perfectly fine because of the hierarchy. 

In the dl1 tool, you would configure:

`CameraCalibrator.TailCutsVolumeReducer.TailCutsImageCleaner`

for this one and

`TailCutsImageCleaner`,

Assuming that CameraCalibrator and the normal TailcutsImageCleaner afe direct children of the tool, like it is now in the stage1 process.

kosack:
> In the dl1 tool, you would configure:
> CameraCalibrator.TailCutsVolumeReducer.TailCutsImageCleaner
> for this one and

I wasn't aware that that works in config files.  If so then great!  you can ignore my comment.

MaxNoe:
Yes, json would look like this:

```json
{
    "CameraCalibrator": {
        "data_volume_reducer_type": "TailCutsDataVolumeReducer",
        "TailCutsDataVolumeReducer": {
            "TailCutsImageCleaner": {
                "picture_threshold": 10,
            }
        }
    }
}
```

MaxNoe (approved):

MaxNoe (changes requested):
You should change this so you have a `image_extractor_type` traitlet so it has the same signature as the base class for `__init__`.

kosack (approved):

ctapipe/image/reducer.py
```diff
@@ -131,6 +131,28 @@ class TailCutsDataVolumeReducer(DataVolumeReducer):
...
+        kwargs
+        """
+        super().__init__(config=config, parent=parent, subarray=subarray, **kwargs)
+
+        self.cleaner = TailcutsImageCleaner(parent=self, subarray=self.subarray)
+
+        self.image_extractor = ImageExtractor.from_name(
+            self.image_extractor_type,
+            subarray=self.subarray,
+            config=self.config
```
MaxNoe, line 153: this should be `parent=self`, not `config=self.config` to correctly build the hierarchy.

Here you would give the config for the DataVolumeReducer as the config for the ImageExtractor, which is wrong.

MaxNoe: Just as above with the cleaner

MaxNoe (approved):

ctapipe/image/reducer.py
kosack: wouldn't `parent=self`  take care of this? Usually you do not have to pass the config manually

kosack (approved):

PR #1405, opened 2020-08-04
Read and write subarray to hdf methods

MaxNoe:
Tests currently fail because of camera.readout.sample_width_ns being converted from float32 to float64 somewhere and then failing the comparison

### DIFF
--- a/ctapipe/instrument/subarray.py
+++ b/ctapipe/instrument/subarray.py
@@ -458,7 +458,7 @@ class SubarrayDescription:
             for row in layout
         }
 
-        positions = np.column_stack([layout[f"pos_{c}"] for c in "xyz"])
+        positions = np.column_stack([layout[f"pos_{c}"].quantity for c in "xyz"])
 
         with tables.open_file(path, mode="r") as f:
             name = f.root.configuration.instrument.subarray._v_attrs.name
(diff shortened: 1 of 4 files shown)


kosack:
> Tests currently fail because of camera.readout.sample_width_ns being converted from float32 to float64 somewhere and then failing the comparison

That's a metadata attribute in the table, which in the current astropy.table implementation is written as a new HDF5 dataset with text strings in JSON format (so it goes from a float32 → JSON text → float64).  In the old implementation (pre astropy-3.0), it stored the metadata in HDF5 attributes, which was much nicer, but I guess they stopped doing that due to some incompatibility (though you can still read the old way).

MaxNoe:
Okay, should be fixed now.

kosack:
Perhaps it would be good now to remove the duplicate code in `tools/stage1.py`?

MaxNoe:
@kosack, yes will do

MaxNoe:
done

ctapipe/instrument/subarray.py
```diff
@@ -335,3 +339,134 @@ def get_tel_ids_for_type(self, tel_type):
...
+                    append=True,
+                    serialize_meta=serialize_meta,
+                )
+
+        with tables.open_file(output_path, mode="r+") as f:
+            f.root.configuration.instrument.subarray._v_attrs.name = self.name
+
+    @classmethod
+    def from_hdf(cls, path):
+        layout = Table.read(path, path="/configuration/instrument/subarray/layout")
```
LukasNickel, line 413: This looks a bit weird with the file path being named `path` as well.

LukasNickel (approved):

PR #1407, opened 2020-08-17
Fix CameraGeometry coordinate conversion for spherical frames

kosack:
As shown in #1191, it was not easy to transform a `CameraGeometry` from the `CameraFrame` to `TelescopeFrame`, or indeed other spherical frames. 

This PR does several related things:
1. modifies `SimTelEventSource` to properly construct CameraGeometries in the SubarrayDescription that have frames initialized to `CameraFrame(focal_length=flen)`
2. adds a unit test to demonstrate the correct way to convert a CameraGeometry from `CameraFrame` to `TelescopeFrame`
3. fixes `CameraGeometry.transform_to()` to work with non-cartesian frames

~~Note that the fix in part 3 introduces a bit of a "code smell": the attributes of the transformed coordinates — e.g. `(x, y)`, `(lon, lat)`, `(fov_lon, fov_lat)`—are chosen using an if/elif chain.  Perhaps there is a more generic way to get the first and second coordinates of different Frame representations?~~

edit: code-smell removed,  also added unit test to go from Spherical to Cartesian camera representations

### DIFF
--- a/ctapipe/instrument/camera/geometry.py
+++ b/ctapipe/instrument/camera/geometry.py
@@ -5,12 +5,14 @@ Utilities for reading or working with Camera geometry files
 import logging
 
 import numpy as np
+from typing import TypeVar
 from astropy import units as u
 from astropy.coordinates import Angle, SkyCoord
 from astropy.table import Table
 from astropy.utils import lazyproperty
 from scipy.spatial import cKDTree as KDTree
 from scipy.sparse import lil_matrix, csr_matrix
+from astropy.coordinates import BaseCoordinateFrame
 import warnings
 
 from ctapipe.utils import get_table_dataset
@@ -21,6 +23,7 @@ from ctapipe.coordinates import CameraFrame
 __all__ = ["CameraGeometry", "UnknownPixelShapeWarning"]
 
 logger = logging.getLogger(__name__)
+CG = TypeVar("CG", bound="CameraGeometry")  # for forward-referencing type hints
 
 
 class CameraGeometry:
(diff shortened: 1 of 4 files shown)


kosack:
Seems to still have an issue with pixel rotation:
image
(Left: CameraFrame, right: TelescopeFrame)

kosack:
better now,
image

Anyone know why the EngineeringCameraFrame class name is "FrameMeta", and not "EngineeringCameraFrame"?

ctapipe/instrument/camera/geometry.py
```diff
@@ -169,18 +180,41 @@ def transform_to(self, frame):
...
+            trans_y = trans.lat
+        else:
+            raise RuntimeError(
+                f"Unable to transform to frame {frame}, "
+                "which has unknown representation"
+            )
+
+        rot = np.arctan2(uv_y[0], uv_y[1])
+        det = np.linalg.det([uv_x.value, uv_y.value])
+        print(f"DEBUG: det={det}, rot={rot.to('deg')}")
```
MaxNoe, line 207: debug print

MaxNoe:
> Perhaps there is a more generic way to get the first and second coordinates of different Frame representations?

```

`In [31]: EngineeringCameraFrame().get_representation_component_names()                                                                        
Out[31]: OrderedDict([('x', 'x'), ('y', 'y')])

In [32]: AltAz().get_representation_component_names()                                                                                         
Out[32]: OrderedDict([('az', 'lon'), ('alt', 'lat'), ('distance', 'distance')])

In [33]: CameraFrame().get_representation_component_names()                                                                                   
Out[33]: OrderedDict([('x', 'x'), ('y', 'y')])

In [34]: TelescopeFrame().get_representation_component_names()                                                                                
Out[34]: OrderedDict([('fov_lon', 'lon'), ('fov_lat', 'lat')])``

**Comment by kosack** [comment]
round trip from Camera → Telescope → Camera  now works
image

**Comment by kosack** [comment]
@MaxNoe thanks, that should do it. I'll update to get rid of the case statement.

**Comment by MaxNoe** [comment]
I cannot reproduce your `FrameMeta` name:

```
❯ python -c 'from ctapipe.coordinates import EngineeringCameraFrame; print(EngineeringCameraFrame.__name__)'
EngineeringCameraFrame
```

kosack:
> I cannot reproduce your FrameMeta name:

was just a typo of mine in my test file (forgot a `()`, which surprisingly still worked)
image

MaxNoe:
> which surprisingly still worked

Not really suprisingly, since every python object is an instance of a class, everything has `__class__.__name__`. For classes that's normally `type`, if no metaclass is specified.

kosack:
> Not really suprisingly, since every python object is an instance of a class, everything has __class__.__name__. For classes that's normally type, if no metaclass is specified.

I'm not surprised `__class__.__name__` worked, but rather that the coordinate transform worked when passing the class rather than in instance. But I guess it's only because that frame has no attributes

MaxNoe:
Ah, yes.  If you pass a class it is default constructed.  There is actually an open discussion to deprecate this on Astropy happening right now.

kosack:
I think this is ready now. Any other comments?

MaxNoe (approved):

PR #1409, opened 2020-08-26
cosmetic change to CameraDisplay

kosack:
Just a small visual update:

- put Frame name in small text in lower-righthand corner
- use correct axes labels from frame

image


MaxNoe (approved):

MaxNoe:
Maybe we want to make the frame name optional? Or just add as a method?

LukasNickel (approved):

HealthyPear (approved):

PR #1419, opened 2020-09-15
Use lts regression instead of siegel slopes for timing parameters

MaxNoe:
Speedup ca. 5x.

This should be the same algorithm as used by EventDisplay and Mars.

Unfortunately, I could not find a good implementation for python, so I implemented it myself after the reference given in the bibliography.rst

### DIFF
--- a/ctapipe/fitting.py
+++ b/ctapipe/fitting.py
@@ -1,0 +1,150 @@
+from numba import njit
+import numpy as np
+
+
+@njit
+def design_matrix(x):
+    """
+    Build the design matrix for linear regression for
+    a given array of x-values.
+
+    This creates a (N, 2) matrix, where the first column are the
+    x values and the second contains all 1.
+    """
+    X = np.empty((len(x), 2), dtype=x.dtype)
+    X[:, 0] = x
+    X[:, 1] = 1
+
+    return X
+
+
+@njit
+def linear_regression(X, y):
+    """
+    Analytical linear regression
+
+    Arguments
+    ---------
+    X: np.array
+        Design matrix of shape (N, 2), as created by ``~ctapipe.fitting.design_matrix``
+    y: np.array
+        y values
+    """
+    return np.linalg.inv(X.T @ X) @ X.T @ y
+
+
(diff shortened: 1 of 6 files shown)


MaxNoe:
Here is an example of the result:

lts_regression

MaxNoe:
The tests fail because they tested for exact results of the old algorithm, it is not surprising that the values changed now that another algorithm is used.

kosack:
The main test that is failing here is the two-pass window sum integration (which is the algorithm used in MARS), which uses the timing slope to recompute the integration windows. 

In the test it makes a toy model for the "true" result and tries to reconstruct it, but now gets a worse result than before. Do we understand why? In this simple case, if the timing slope is correct, the answer should be very close to the truth value. So if we increase the tolerance, it would be good to be careful that it is correct to do so.  @HealthyPear how did you choose the tolerance for this test (rtol=0.07) ? Is it overly conservative?

HealthyPear:
> @HealthyPear how did you choose the tolerance for this test (rtol=0.07) ? Is it overly conservative?

No real reason: actually I even made it stricter than the tests just above, which use 0.1 ....

MaxNoe:
I relaxed the requirement to 0.1 like in the other tests, test should be passing now.

It is also unsurprising that this algorithms performs slightly worse in the absence of outliers, since it uses fewer points, so the variance will be higher.

kosack:
> It is also unsurprising that this algorithms performs slightly worse in the absence of outliers, since it uses fewer points, so the variance will be higher.

Is that going to be a problem for small images where there are only 3-4 points to fit?

kosack (approved):

MaxNoe:
> Is that going to be a problem for small images where there are only 3-4 points to fit?

How well are all the other algorithms handling events with 3-4 pixels in the shower? 

I don't expect this fitting to be any worse than other parts of the pipeline.
The fits also depends on the shower axis estimation, which will not be really reliable on 3-4 pixels in the first place.

HealthyPear (approved):

PR #1426, opened 2020-09-16
Use indices for camera / optics lookup in `from_hdf` / `to_hdf`

maxnoe:
This removes the reliance on the guessed names for the matching of cameras / optics to the telescopes


MaxNoe:
This also raises the point that the DL1 data format needs a formal specification, not just the implementation here.

And that would probably would also rule out the reliance on the astropy special datasets for table metadata.

kosack:
Can you explain this better? What is the effect on the output if you don't use names anymore? What is the problem with using names? The "guessed" names are of course just an artifact of the fact that simtel format files do not contain that metadata, but of course in the final CTA format, it will not be "guessed", but named officially.

MaxNoe:
Names are nowhere enforced to be unique, it is simply not checked and created the first problems with the MAGIC simulations, where both MAGIC telescopes have the same `str(telescope_description)` but are different in mirror tiles and mirror area.

Relying on indices is just all around more safe, since these indices are based on the full information stored in the telescope description (via `__hash__` and creating a set operations).

This makes it intrinsically more general and removes reliance on guessing now or the reliance on external information later.

It *just works* for any situation of different names and even for unknown telescopes.

Imagine someone using ctapipe for their own array not included in the guessing keys. The indices approach will just work fine but the name based approach will crash because all telescopes are "UNKNOWN_UNKNOWN_UNKNOWN".

Like I said, I would remove all real dependence on these guessing keys or external references matching telescope ids to something else and just rely on well-defined, self-describing formats.

Using indices and an index table joining these together is the normal approach for this.

MaxNoe:
Maybe we should even make an option that disables the guessing and checks that reconstruction still works as expected and nothing is relying on it.

MaxNoe:
Ah, I forgot to point out, that this is needed to correctly read back simulations of the LST + MAGIC subarray as currently simulated. I will open an issue about that and backreference here.

ctapipe/instrument/subarray.py
```diff
@@ -392,17 +398,17 @@ def to_hdf(self, output_path, overwrite=False):
             serialize_meta=serialize_meta,
             overwrite=overwrite,
         )
-        for camera in self.camera_types:
+        for i, camera in enumerate(self.camera_types):
             camera.geometry.to_table().write(
                 output_path,
-                path=f"/configuration/instrument/telescope/camera/geometry_{camera}",
+                path=f"/configuration/instrument/telescope/camera/geometry_{i}",
                 append=True,
```
kosack, line 405: probably best to give a format spec here like `{i:02d}` just in case there are more than 10 geometries.

kosack (changes requested):

ctapipe/instrument/subarray.py
maxnoe: Is it really important that in the improbable case of more than 10 telescopes the entries are not natural sorted?

maxnoe: Using a leading zero breaks backwards compatibility or would make providing backwards compatibility  harder, since `:02d` is not a valid format string for `str` values used before.

kosack (changes requested):
Everything looks fine except one possible issue:  

when the option `DL1Writer.split_datasets_by=tel_type` is enabled, we still get datasets named by the tel description string ("SST_ASTRI_CHEC").  Should we also switch that to be a stringified version of the tel_index? I think otherwise the same problem of UNKNOWN types will occur  , so perhaps switch to something like `teltype_{tel_type_index}`.   Note that so far there is only an optics index and camera index, so we'd need one more index to map optics+camera → telescope (or just use something like `optics{optics_index}_camera{camera_index}` as the dataset name)

maxnoe:
@kosack I don't really think that's a critical problem, since this is only an option and not used by default. It is not even a problem at all for DL1b and only a problem for DL1a if the cameras have different pixel numbers since then the images do not have the same shape.

maxnoe:
@kosack I think this is ready now. Is your issue about the `DL1DataWriter` really connected to this change here? Maybe another PR for that would be better.

kosack:
> @kosack I don't really think that's a critical problem, since this is only an option and not used by default. It is not even a problem at all for DL1b and only a problem for DL1a if the cameras have different pixel numbers since then the images do not have the same shape.

Ok, we can solve that in the future if it comes up as a problem

kosack (approved):

ctapipe/io/simteleventsource.py
```diff
@@ -300,15 +298,12 @@ def prepare_subarray_info(self, telescope_descriptions, header):
...
-            camera = self._camera_cache.get(telescope.camera_name)
-            if camera is None:
-                camera = build_camera(
-                    cam_settings,
-                    pixel_settings,
-                    telescope,
-                    frame=CameraFrame(focal_length=optics.equivalent_focal_length),
-                )
-                self._camera_cache[telescope.camera_name] = camera
+            camera = camera = build_camera(
```
kosack, line 301: `camera=` seems to be repeated

kosack (changes requested):

ctapipe/io/simteleventsource.py
maxnoe: fixed

maxnoe:
> Ok, we can solve that in the future if it comes up as a problem

We could solve that by storing the pixel number in the camera name, something like "UNKNOWN-1039-PIXEL-CAMERA"

ctapipe/instrument/subarray.py
```diff
@@ -458,66 +463,75 @@ def to_hdf(self, output_path, overwrite=False):
...
+        layout = QTable.read(path, path="/configuration/instrument/subarray/layout")
 
         cameras = {}
-        for name in set(layout["camera_type"]):
+
+        # backwards compatibility for older tables, index is the name
+        if "camera_index" not in layout.colnames:
+            layout["camera_index"] = layout["camera_type"]
+
+        for idx in sorted(set(layout["camera_index"])):
```
LukasNickel, line 474: Is the `sorted` necessary? Everything is addresed by the index anyway

maxnoe: Index can be the camera name if in "backwards" compatibility mode.

maxnoe: right

maxnoe: copy paste ;)

ctapipe/io/__init__.py
```diff
@@ -4,7 +4,7 @@
 from .tableio import TableWriter, TableReader
 from .datalevels import DataLevel
 from .astropy_helpers import read_table
-from .datawriter import DataWriter
+from .datawriter import DataWriter, DL1_DATA_MODEL_VERSION
```
LukasNickel, line 7: Broken by the code checker commit

maxnoe:
@kosack, this is how it looks if we add the mirror area to the optics name and the pixel number to camera name:

For this, I removed the LST from the guessing keys:
```
In [4]: [str(t) for t in s.subarray.telescope_types]
Out[4]: 
['SST_ASTRI_ASTRICam',
 'UNKNOWN_UNKNOWN-387M2_UNKNOWN-1855',
 'MST_MST_FlashCam']

``

**Review by LukasNickel — APPROVED** [review]

**Comment by kosack** [comment]
I tested this with a recent HESS simulation, which is essentially a FlashCam based LST + 4 MSTs with something similar to NectarCam on them.  The result is two cameras with the same name (UNKNOWN_UNKNOWN_UNKNOWN) , but everything does seem to work ok, except that the `subarray.info()` function fails to separate the two telescope types, since they have the same name.  
```
In [6]: sub.info()
Subarray : MonteCarloArray
Num Tels : 5
Footprint: 0.02 km2

          Type          Count Tel IDs
----------------------- ----- -------
UNKNOWN_UNKNOWN_UNKNOWN     5 1-5
```
image
image

maxnoe:
@kosack done, but no tests for this yet. (As for anything apperently connected to this guessing).

Can we add a small test file with "unknown" telescopes to the data server?

kosack:
thanks, that works better.  Let me see if I can find a small sim from HESS or similar with telescopes that have no GuessingKeys .  Alternately, we can just clear the list of GuessingKeys in a test and use one of the existing files.

maxnoe:
> Also, I suppose we could have the telescope type (LST/MST/SST) just generated from the mirror size rather than using UNKNOWN. Just define an acceptable range for each.

Done, the limits are 8 m / 16 m diameter.

kosack:
Great. Works well with HESS too.

kosack (approved):

HealthyPear (approved):

PR #1427, opened 2020-09-16
add function to get astropy table from ctapipe hdf5 file

kosack:
this just provides: 

```py
from ctapipe.io.astropy_helpers import h5_table_to_astropy
t = h5_table_to_astropy("gammas2.dl1.h5", "/dl1/event/telescope/parameters/tel_001")
t
```

```
<QTable length=50>
obs_id event_id tel_id  hillas_intensity        hillas_x             hillas_y       ... peak_time_min peak_time_mean peak_time_std   peak_time_skewness    peak_time_kurtosis
                                                   m                    m           ...
int32   int64   int16       float64             float64              float64        ...    float32       float32        float32           float64               float64
------ -------- ------ ------------------ -------------------- -------------------- ... ------------- -------------- ------------- --------------------- ---------------------
   101     1503      1  151.8677215576172  0.18290526905116353   0.3103048900205065 ...      8.831734      10.528524      1.492059    0.7524924874305725   -0.8947949409484863
   101     1506      1  195.8453986644745   0.1657280817609988 -0.15308384378334355 ...      9.131325       9.969362     0.4076317   -0.9799400568008423 -0.015481710433959961
   101     2900      1  984.3376245498657   0.6704040242695486  0.47259416766985674 ...      8.647298      11.999297     1.7776399 -0.010056410916149616   -0.8698732852935791
```

With correct units re-attached, and any other user metadata still recorded:

```py3
t.meta
```
```
{'CTAPIPE_VERSION': '0.8.0.post22+gitea2cf40'}
```

It also works correctly for tables with vector columns, like images:
```python
In [5]: t = h5_table_to_astropy("gammas2.dl1.h5", "/dl1/event/telescope/images/tel_001")

In [6]: t
Out[6]:
<QTable length=50>
obs_id event_id tel_id        image [1855]           peak_time [1855]    image_mask [1855]
int32   int64   int16           float32                  float32                bool
------ -------- ------ -------------------------- ---------------------- -----------------
   101     1503      1   1.1831577 .. -0.47390354      26.475227 .. 29.0    False .. False
   101     1506      1  -0.14622755 .. 0.47994235  27.13752 .. 17.578224    False .. False
   101     2900      1   1.8219908 .. -0.11884633        3.840397 .. 1.0    False .. False
   101     4007      1   -0.53255767 .. 1.7376837      25.0 .. 2.1106777    False .. False
   101     7007      1   -0.9513076 .. -2.6859846       22.21572 .. 19.0    False .. False
   101    10002      1    -0.56497747 .. 0.975543 11.490385 .. 17.094572    False .. False
   101    13306      1    -1.2410553 .. 2.0778234        6.0 .. 2.206301    False .. False
```

### DIFF
--- a/ctapipe/io/astropy_helpers.py
+++ b/ctapipe/io/astropy_helpers.py
@@ -1,4 +1,7 @@
 #!/usr/bin/env python3
+"""
+Functions to help adapt internal ctapipe data to astropy formats and conventions
+"""
 
 from pathlib import Path
 
@@ -44,7 +47,7 @@ def h5_table_to_astropy(h5file, path) -> QTable:
 
     other_attrs = {}
     column_units = {}  # mapping of colname to unit
-    for attr in table.attrs._f_list():
+    for attr in table.attrs._f_list():  # pylint: disable=W0212
         if attr.endswith("_UNIT"):
             colname = attr[:-5]
             column_units[colname] = table.attrs[attr]
--- a/ctapipe/io/astropy_helpers.py
+++ b/ctapipe/io/astropy_helpers.py
@@ -1,37 +1,47 @@
 #!/usr/bin/env python3
 
+from pathlib import Path
+
 import tables
-from astropy.table import Table
+from astropy.table import QTable
 from astropy.units import Unit
+
 
 __all__ = ["h5_table_to_astropy"]
 
 
-def h5_table_to_astropy(h5file, path) -> Table:
-    """Get a table from a ctapipe-format HDF5 table as an `astropy.table.Table`
+def h5_table_to_astropy(h5file, path) -> QTable:
+    """Get a table from a ctapipe-format HDF5 table as an `astropy.table.QTable`
     object, retaining units. This uses the same unit storage convention as
     defined by the `HDF5TableWriter`, namely that the units are in attributes
-    named by `<column name>_UNIT` that are parsible by `astropy.units`.
+    named by `<column name>_UNIT` that are parsible by `astropy.units`. Columns
+    that were Enums will remain as integers.
 
     Parameters
     ----------
-    h5file: Union[str, tables.file.File]
-       input filename or PyTables file handle
+    h5file: Union[str, Path, tables.file.File]
+        input filename or PyTables file handle
     path: str
-       path to table in the file
+        path to table in the file
+
+    Returns
+    -------
+    astropy.table.QTable:
+        table in Astropy Format
 
     """
 
     should_close_file = False
-    if not isinstance(h5file, tables.file.File) and isinstance(h5file, str):
+    if not isinstance(h5file, tables.file.File) and isinstance(h5file, (str, Path)):
         h5file = tables.open_file(h5file)
         should_close_file = True
     else:
         raise ValueError(
-            f"expected a string or PyTables filehandle for argument 'h5file', got {h5file}"
+            f"expected a string, Path, or PyTables "
+            f"filehandle for argument 'h5file', got {h5file}"
         )
 
-    table: tables.table.Table = h5file.get_node(path)
+    table = h5file.get_node(path)
 
     column_units = {}  # mapping of colname to unit
     for attr in table.attrs._f_list():
@@ -39,7 +49,7 @@ def h5_table_to_astropy(h5file, path) -> Table:
             colname = attr[:-5]
             column_units[colname] = table.attrs[attr]
 
-    astropy_table = Table(table[:])
+    astropy_table = QTable(table[:])
 
     for column, unit in column_units.items():
         astropy_table[column].unit = Unit(unit)
--- a/ctapipe/io/astropy_helpers.py
+++ b/ctapipe/io/astropy_helpers.py
@@ -6,7 +6,6 @@ import tables
 from astropy.table import QTable
 from astropy.units import Unit
 
-
 __all__ = ["h5_table_to_astropy"]
 
 
@@ -43,13 +42,16 @@ def h5_table_to_astropy(h5file, path) -> QTable:
 
     table = h5file.get_node(path)
 
+    other_attrs = {}
     column_units = {}  # mapping of colname to unit
     for attr in table.attrs._f_list():
         if attr.endswith("_UNIT"):
             colname = attr[:-5]
             column_units[colname] = table.attrs[attr]
+        else:
+            other_attrs[attr] = table.attrs[attr]
 
-    astropy_table = QTable(table[:])
+    astropy_table = QTable(table[:], meta=other_attrs)
 
     for column, unit in column_units.items():
         astropy_table[column].unit = Unit(unit)
--- a/ctapipe/io/astropy_helpers.py
+++ b/ctapipe/io/astropy_helpers.py
@@ -34,9 +34,11 @@ def h5_table_to_astropy(h5file, path) -> QTable:
     """
 
     should_close_file = False
-    if not isinstance(h5file, tables.file.File) and isinstance(h5file, (str, Path)):
+    if isinstance(h5file, (str, Path)):
         h5file = tables.open_file(h5file)
         should_close_file = True
+    elif isinstance(h5file, tables.file.File):
+        pass
     else:
         raise ValueError(
             f"expected a string, Path, or PyTables "
--- a/ctapipe/io/astropy_helpers.py
+++ b/ctapipe/io/astropy_helpers.py
@@ -1,0 +1,50 @@
+#!/usr/bin/env python3
+
+import tables
+from astropy.table import Table
+from astropy.units import Unit
+
+__all__ = ["h5_table_to_astropy"]
+
+
+def h5_table_to_astropy(h5file, path) -> Table:
+    """Get a table from a ctapipe-format HDF5 table as an `astropy.table.Table`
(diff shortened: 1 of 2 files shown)


ctapipe/io/astropy_helpers.py
```diff
@@ -0,0 +1,50 @@
+#!/usr/bin/env python3
+
+import tables
+from astropy.table import Table
+from astropy.units import Unit
+
+__all__ = ["h5_table_to_astropy"]
+
+
+def h5_table_to_astropy(h5file, path) -> Table:
```
MaxNoe, line 10: Maybe use `QTable`? In the end the point of this function is to provide access to the units. Maybe this could also be an argument. `table_class=QTable`

MaxNoe: The difference is that accessing columns or rows or single elements yields simple numbers for `Table` but quantities for `QTable`:

ctapipe/io/astropy_helpers.py
```diff
@@ -0,0 +1,50 @@
...
+    ----------
+    h5file: Union[str, tables.file.File]
+       input filename or PyTables file handle
+    path: str
+       path to table in the file
+
+    """
+
+    should_close_file = False
+    if not isinstance(h5file, tables.file.File) and isinstance(h5file, str):
```
MaxNoe, line 26: Probabably would be good to also allow `pathlib.Path`

ctapipe/io/astropy_helpers.py
```diff
@@ -0,0 +1,50 @@
...
+        h5file = tables.open_file(h5file)
+        should_close_file = True
```
MaxNoe, line 28: This is always True?

ctapipe/io/astropy_helpers.py
```diff
@@ -0,0 +1,50 @@
...
+    else:
+        raise ValueError(
+            f"expected a string or PyTables filehandle for argument 'h5file', got {h5file}"
+        )
+
+    table: tables.table.Table = h5file.get_node(path)
```
MaxNoe, line 34: I really don't understand these kinds of type hints. What are they for? 

type hints only make sense in API definitions, right?

ctapipe/io/astropy_helpers.py
kosack: I'm not sure Qtables are very standard or used anywhere, units still work in normal Tables.

```
table['energy'].quantity  # → returns a Quantity array for that column
table['energy'].to("GeV") # returns a Column in GeV
```

MaxNoe: The difference is mainly for iterating over the rows:
```
for row in table:
    print(row['energy')  # float for Table, Quantity for QTable
```

ctapipe/io/astropy_helpers.py
kosack: it's not really needed, just allows my editor to do tab-completion without running the code. In principle it would also raise a mypy warning.  I will remove it, since not critical

kosack:
I do not handle Enum columns in this right now, but I think that is probably ok, as I'm not sure how you would put them in a Table (perhaps as an Object column, which then would be hard to write out again)

ctapipe/io/astropy_helpers.py
MaxNoe: Ah, helps the editor, right.

ctapipe/io/astropy_helpers.py
kosack: Ok, will change to QTable

MaxNoe:
I think leaving them untouched is good. You will still be able to do things like:

```
table['event_type'] == EventType.SUBARRAY.value 
```

and faster than with an object column.

ctapipe/io/astropy_helpers.py
```diff
@@ -0,0 +1,50 @@
...
+
+    column_units = {}  # mapping of colname to unit
+    for attr in table.attrs._f_list():
+        if attr.endswith("_UNIT"):
+            colname = attr[:-5]
+            column_units[colname] = table.attrs[attr]
+
+    astropy_table = Table(table[:])
```
MaxNoe, line 42: You can give the table constructor a `units=` dictionary mapping column names to units

ctapipe/io/astropy_helpers.py
```diff
@@ -0,0 +1,50 @@
...
+        should_close_file = True
+    else:
+        raise ValueError(
+            f"expected a string or PyTables filehandle for argument 'h5file', got {h5file}"
+        )
+
+    table: tables.table.Table = h5file.get_node(path)
+
+    column_units = {}  # mapping of colname to unit
+    for attr in table.attrs._f_list():
```
MaxNoe, line 37: I would maybe check the other way around. So go over all columns and check if the expected field is there.

This would also not rely on private parts:

```
units = {col: table.attrs.get('f{col}_UNIT'} for col in table.dtype.names}
```
and then pass this to the `Table` constructor

ctapipe/io/astropy_helpers.py
kosack: doesn't seem to work for QTables

ctapipe/io/astropy_helpers.py
MaxNoe: Sorry, it has no get method

ctapipe/io/astropy_helpers.py
MaxNoe: Works for me:

```
In [3]: QTable({'x': [1, 2, 3]},  units={'x': 'm'})
Out[3]: 
<QTable length=3>
   x   
   m   
float64
-------
    1.0
    2.0
    3.0
```

kosack: what version are you using? In my case it fails

MaxNoe: 4.0.1.post1

kosack: really? Cutting and pasting your example fails for me. 
```
In [6]: astropy.__version__
Out[6]: '4.0.1.post1'
```
```
In [2]: from astropy.table import QTable

In [3]: QTable({'x': [1, 2, 3]},  units={'x': 'm'})
<ipython-input-3-43bbf31be07a> in <module>
----> 1 QTable({'x': [1, 2, 3]},  units={'x': 'm'})

~/anaconda3/envs/cta-dev/lib/python3.7/site-packages/astropy/table/table.py in __init__(self, data, masked, names, dtype, meta, copy, rows, copy_indices, **kwargs)
    505         elif kwargs:
    506             raise TypeError('__init__() got unexpected keyword argument {!r}'
--> 507                             .format(list(kwargs.keys())[0]))
    508
    509         if (isinstance(data, np.ndarray) and

TypeError: __init__() got unexpected keyword argument 'units'
```

ctapipe/io/astropy_helpers.py
kosack: > also not rely on private parts

those are not really private, see the pytables docs. They reserve `_f_` as public functions to avoid conflicting with the fact that you can access columns and attributes by `x.attr_name`

kosack:
One thing is that this function is really quite similar to `ctapipe.utils.get_table_dataset()`, so perhaps we should encourage the use of that function and simply call this one from inside it if an HDF5 generated by ctapipe is read?  It would just mean adding a "path" argument to that function.

ctapipe/io/astropy_helpers.py
MaxNoe: Ah, sorry. Got confused with my envs. I had the dev version of astropy installed into my user site so the output of `conda list astropy` showed the wrong version.

kosack:
I think I will wait to clean up ctapipe.utils.datasets. I just tried implementing it, and found too many things I want to change.  So let's just keep this PR simple.

ctapipe/io/tests/test_astropy_helpers.py
```diff
@@ -0,0 +1,41 @@
...
+    assert table["energy"].unit == u.TeV
+    assert "CTAPIPE_VERSION" in table.meta
+
+    # test using a string
+    table = h5_table_to_astropy(str(filename), "/events")
+
+    # test using a file handle
+    handle = tables.open_file(filename)
+    table = h5_table_to_astropy(handle, "/events")
+    handle.close()
```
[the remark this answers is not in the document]
kosack: can you do that if there is a conditional? (e.g. I don't want to close it if there was a handle passed in)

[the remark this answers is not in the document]
MaxNoe: This is in the test and you open the handle directly above.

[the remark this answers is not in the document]
kosack: ah, yes, I will use the context manager there. Wasn't sure pytables.open_file supported it

vuillaut (approved):

kosack:
is there any reason we can't merge this? Would be very helpful for benchmarking efforts.

maxnoe (approved):

maxnoe:
If we are strict about version numbers this would mean bumping the next release to 0.10, but it's not a breaking change, just new funcitonality, so I'll merge and we can think about what version number the next release shall be.

PR #1429, opened 2020-09-18
Add config options to disable integration correction, fixes #1418, fixes #1376

### LINKED ISSUE #1376: How to switch off charge integration corrections?
Hi,
there are cases in which one do not want to apply the pulse shape integration corrections to the integrated charges . With the present code, are there better methods than patching the reference pulse shape?

### LINKED ISSUE #1418: Make integration correction configurable
A couple of (late) comments on this:

> * Adding a `_calculate_correction` method to the ImageExtractors, so one can define the appropriate way to calculate the correction for each particular extractor. For a noise-less unit pulse the resulting charge from an ImageExtractor should always equal 1, ensuring the correct units for the resulting charge - this is the goal of the correction factor.

It does not seem to me that the goal is the right one. Real or simulated pulses are composed of photons with a time structure. Take a single p.e. pulse shape (like LST's low-gain one) which has some "ringing" with undershoots after the main peak. For the single p.e. shape, if you integrate a number of samples which don't include any undershoot, the correction factor would be <1. But for a real pulse (be it a calibration pulse, or a pulse from a shower), with a time structure, the main peaks of the photon pulses will fill the undershoots of earlier photons, so the mean or typical pulse will have a pretty different shape, and the correction will be wrong.

> * Removing the ability to skip over the correction if there is no reference pulse defined.

I actually think it would be nice, at least for testing purposes, if the pulse correction could be skipped altogether in some easy way, through a line in the configuration file.

_Originally posted by @moralejo in https://github.com/cta-observatory/ctapipe/pull/1242#issuecomment-692174072_

MaxNoe:


ctapipe/image/extractor.py
```diff
@@ -347,6 +351,11 @@ class FixedWindowSum(ImageExtractor):
         "(peak_index - shift)",
     ).tag(config=True)
 
+    apply_integration_correction = BoolTelescopeParameter(
+        default_value=True,
+        help="Weither or not to apply the integration window correction",
```
moralejo, line 356: Typo: Weither

ctapipe/image/extractor.py
```diff
@@ -397,12 +407,18 @@ class GlobalPeakWindowSum(ImageExtractor):
...
+
     window_shift = IntTelescopeParameter(
         default_value=3,
         help="Define the shift of the integration window from the peak_index "
         "(peak_index - shift)",
     ).tag(config=True)
 
+    apply_integration_correction = BoolTelescopeParameter(
+        default_value=True,
+        help="Weither or not to apply the integration window correction",
```
moralejo, line 419: typo

ctapipe/image/extractor.py
```diff
@@ -454,12 +471,18 @@ class LocalPeakWindowSum(ImageExtractor):
...
+
     window_shift = IntTelescopeParameter(
         default_value=3,
         help="Define the shift of the integration window"
         "from the peak_index (peak_index - shift)",
     ).tag(config=True)
 
+    apply_integration_correction = BoolTelescopeParameter(
+        default_value=True,
+        help="Weither or not to apply the integration window correction",
```
moralejo, line 483: ditto

ctapipe/image/extractor.py
```diff
@@ -511,17 +535,24 @@ class NeighborPeakWindowSum(ImageExtractor):
...
+
     lwt = IntTelescopeParameter(
         default_value=0,
         help="Weight of the local pixel (0: peak from neighbors only, "
         "1: local pixel counts as much as any neighbor)",
     ).tag(config=True)
 
+    apply_integration_correction = BoolTelescopeParameter(
+        default_value=True,
+        help="Weither or not to apply the integration window correction",
```
moralejo, line 553: typo

ctapipe/image/extractor.py
```diff
@@ -647,6 +679,11 @@ class TwoPassWindowSum(ImageExtractor):
         default_value=3, help="width of sliding window used to do peak detection"
     ).tag(config=True)
 
+    apply_integration_correction = BoolTelescopeParameter(
+        default_value=True,
+        help="Weither or not to apply the integration window correction",
```
moralejo, line 684: typo

moralejo (approved):

ctapipe/image/extractor.py
watsonjj, line 356: typo "Weither" -> "Whether"
Simply "Apply the integration window correction" would suffice

FrancaCassol (commented):
Approved after typo correction

MaxNoe:
Typo fixed

watsonjj (approved):

FrancaCassol (approved):

Issue #1421, opened 2020-09-16
Make output path of provenance.log configurable


maxnoe:
This is currently hard coded to `provenance.log` in `cwd` which makes it impossible to run the stage1 tool in a read-only directory.

kosack:
probably there needs to be a much more general discussion of how we store outputs. Ideally this file would be "attached" somehow to all output files produced by a Tool.  Though since the output files contain an activity_id and a data_product_id, that may be sufficient  

At minimum I'd suggest we make the filename and the output of provenance configurable.  

Note that the provenance system will likely be overhauled soon (there have been several meetings, and ongoing prototype work  to support IVOA prov,LogProv, etc.)

Bultako:
Could we think of having this path and filename as a property of the Tool that can be modified? (with a default value `provenance.log`) In this case we take the risk of ending up with different provenance files for different tools, instead of a single one, which frankly I do not see as a problem now..

MaxNoe:
Yes, this would be the implementation

MaxNoe:
Also, since it will be needed by everything in cta, provenance should probably be handled its own, self contained library with minimal dependencies, so it can be used by every other package as dependency

PR #1432, opened 2020-09-29
Make output path of provenance.log configurable 

Bultako:
This PR intends to solve issue #1421 
It adds `provenance_dir` as a configurable trait of a Tool.

### DIFF
--- a/ctapipe/core/tests/test_tool.py
+++ b/ctapipe/core/tests/test_tool.py
@@ -47,7 +47,7 @@ def test_provenance_dir():
         userparam = Float(5.0, help="parameter").tag(config=True)
 
     tool = MyTool()
-    assert str(tool.provenance_dir) == os.getcwd()
+    assert str(tool.provenance_log) == os.path.join(os.getcwd(), "application.provenance.log")
 
 
 def test_export_config_to_yaml():
--- a/ctapipe/core/tests/test_tool.py
+++ b/ctapipe/core/tests/test_tool.py
@@ -1,3 +1,4 @@
+import os
 import pytest
 from traitlets import Float, TraitError, List, Dict
 from traitlets.config import Config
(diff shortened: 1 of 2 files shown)


kosack (dismissed):

MaxNoe (changes requested):
At the moment, this is a single file.

So why add an output directory with a hard-coded filename and not an option for the filename itself?

Please change to a `provenance_log =Path(...)` trait for the output filename

kosack:
good point - perhaps by default set the filename to the name of the tool (tools should have a .name attribute), so you get something like "./ctapipe-something-tool.prov.log", and then let the user specify any path/name.

Bultako:
I've added `prov_filename` attribute as described by @kosack, so to have independent `prov_dir` and  `prov_filename` attributes for Tools with default values `cwd` and `tool.name.prov.log`

MaxNoe:
I think you misunderstood what we had in mind. 

The directory trait is unnecessary, since an argument for the file can be a relative or absolute path to any directory.
So this trait should be removed.

Then a trait `provenance_log` with type `ctapipe.core.traits.Path` should be added.
@kosack proposed that this trait should have a default value containing the name of the tool.

I think this can be achieved using a default method like this:

```
from traitlets import default
from ctapipe.core.traits import Path

class Tool():
    ...
    provenance_log = Path()
    
    @default('provenance_log')
    def _default_provenance_log(self):
        return self.name + '.provenance.log'
     
```

MaxNoe (approved):
LGTM, thank you!

Bultako:
Thanks @MaxNoe 
Indeed, I had completely misunderstood the point, I've made the suggested changes, having a single `provenance_log` attribute of type `Path()`

It now comes to my mind the use case when `provenance_log` value is set to a non-existing directory path, do we raise an error or create that path? in the latter case, should we implement a method in the `Path()` trait to create that non-existing folder? or we do use the [Pathlib.Path.mkdir()](https://docs.python.org/3/library/pathlib.html#pathlib.Path.mkdir) method instead?

MaxNoe:
Creating the directory could be nice, yes.

MaxNoe (approved):

kosack (approved):

MaxNoe:
Somehow the status update from travis was not received, I restarted the build

PR #1439, opened 2020-10-02
Return nan for non-invertible matrices

maxnoe:


MaxNoe:
@HealthyPear could you try the file that failed with this PR?

kosack (approved):

MaxNoe:
@HealthyPear tested this on a large simtel file that crashed before which now worked fine.

HealthyPear (approved):

kosack:
I think this is ok now, but we should check that in all places it is used, we ensure that NaN is an ok result. I guess the only critical place will be in the second pass of the TwoPassWindowSum, otherwise this info is not used, only written to the disk.

MaxNoe:
Ok, let's merge this and then check how algorithms using this one handle the nan cases.

PR #1442, opened 2020-10-05
Fix loss of performance in TwoPassWindowSum

HealthyPear:
Added

- check for the (likely very rare) case in which the timing fit results in an undetermined timing slope,
- new section in image extractors docs page with 1st pass image example for this extractor

## Preview

Nevermind the requirement, it doesn't have the same units...


ctapipe/image/extractor.py
```diff
@@ -935,7 +925,7 @@ def _apply_second_pass(
         # We have to add 2 samples each side, so the shift will always
         # be (-)2, while width will always end 4 samples to the right.
         # This "always" refers to a 5-samples window of course
-        window_width_default = 5
+        window_width_default = 5  # +1 since extract_around_peak stops at end
```
maxnoe, line 928: THis comment and the one above don't really make sense together and I think they can be simplified to "use an integration window of width 5 central around the peak."

maxnoe (dismissed):

ctapipe/image/extractor.py
```diff
@@ -743,38 +743,27 @@ def _apply_first_pass(
...
 
         # For each pixel, we slide a 3-samples window through the
-        # waveform without touching the extremes (so later we can increase it
-        # to 5), summing each time the ADC counts contained within it.
+        # waveform summing each time the ADC counts contained within it.
 
-        # 'width' could be configurable in a generalized version
-        # Right now this image extractor is optimized for LSTCam and NectarCam
         width = self.peak_finding_window_width.tel[telid]
         sums = convolve1d(waveforms, np.ones(width), axis=1, mode="nearest")
```
kosack, line 749: did you try this with `method='mirror'`, which I think is more similar to what is done in CTA-MARS?  That might fix the slight differences, if they are due to edge cases (literally pulses at the edge of the readout window).  Though it's hard to see how that would be intensity dependent.

kosack: looking at the previous implementation in 0.8, I think pulses that were in the edge bins were simply ignored, however

HealthyPear: yes this was indeed the case, because in CTAMARS the 3-samples peak search window never reaches the edges, but it stops before the last sample(s) of the readout window, allowing to increase the window as 1+3+1 (do I remember correctly @moralejo ?)

maxnoe: So change to `peak_index = argmax([:, 1:-1], axis=1)` to avoid the edges?

moralejo: > yes this was indeed the case, because in CTAMARS the 3-samples peak search window never reaches the edges, but it stops before the last sample(s) of the readout window, allowing to increase the window as 1+3+1 (do I remember correctly @moralejo ?)

Yes, that is how it is done.

maxnoe (approved):

maxnoe:
With the removal of the option to set the window width, this is now a breaking change

PR #1444, opened 2020-10-06
Remove need for `add_subcomponent` include subcomponents in Component.get_current_config

maxnoe:
While reviewing a PR in lstchain, I realized that calling `add_component` is not really necessary, we can just go through all the members and check which are components.

Also, components did not include the config of their subcomponents, only `Tools` did that.

This fixes both. `add_component` is now a no-op that only raises a deprecation warning, it has been removed from all ctapipe tools and `get_current_config` returns all subcomponent's config also for components.

I also tested that roundtrip is possible with `tool = Tool(config=Config(other_tool.get_current_config()))` so that we can be sure we can recreate a component with the output from `get_current_config`


kosack:
Nice! I never liked the `add_component()` method, but couldn't think of a better way to associate *instances* of configurable classes with their parent (traitlets.config associates the Class with it's Parent, but not instances). Using introspection is indeed cleaner and hides the implementation from the user, which is good. 

 What happens if the user does not store a reference to the Component in the Tool itself?. For example, there is nothing stopping people from constructing and using a Component in the `start()` method without setting `self._something=MyNewComponent()`, rather than in `setup()`, and then the config for that component would be lost if it's not a member of the tool instance. Perhaps just documenting things better (e.g. explicitly say "do all construction in setup()"), but then nobody reads docs...  I guess that was still an issue before, however, so perhaps shouldn't worry about it.

maxnoe:
Yes, I think this is an independent issue. I think nowhere `add_component` was used without assigning the result to `self`.

I don't think we can do anything about the other issue besides catching that in review.

kosack:
I think the issue before was just if somebody forgot to use `add_component()`, the info was lost, but it was at least clear why, since you never associated it with the parent.  In this case, it's a bit hidden: you forget to store a reference, and then the config disappears, which is a bit unexpected.    I like the cleaner look, but not sure if this will lead to confusion later.  Of course, there are many parts of the config system that are not really nice, and maybe somebody in the future will have time to replace it entirely.

maxnoe:
This is a classic of "why the hell did it work before".

maxnoe:
Also: I did not change anything with respect to **setting** configuration, only with respect to **getting** the configuration.

Why did this test break? Why did it work before when it was not putting in the right alias?

kosack (changes requested):
anyhow, despite my criticisms, I still like this better than what was there before, so I'll accept it if you add more info to the documentation and examples, specifically mentioning that one *must* store a reference to all Components in their parent Tool or Component.  So for example;

```py3
sub = SubComponent(parent=self) 
```
is forbidden, but
```py3
self.sub = SubComponent(parent=self)
```
is ok.  

Side note: since all Components and Tools do know their parent, maybe there is a way to validate this? Check that if `parent` is not None that the parent class contains a reference to self? Could be a `.validate()` method like in the `Field` class.

kosack:
> Why did this test break? Why did it work before when it was not putting in the right alias?

the way traitlets.config associates parent names is super complex... I wanted to try to reproduce it at one point in get_current_config() so the parameters get associated with the correct super-class (e.g. that it is `Tool.log_level` not `MyNewTool.log_level` as is done in the --help), but gave up. Perhaps it's some special case in that logic.

maxnoe:
But I did not change anything related to that code, did I?

kosack:
> But I did not change anything related to that code, did I?

Not that I see... Perhaps it worked by magic?

maxnoe:
I removed the `add_subcomponent` in the test.

maxnoe:
Okay, I figured it out, but it is really strange. After the meeting I will go back to working  on a new config system.

I made to changes to the test: I changed the config hierarchy:

Works Before:

```
aliases = {"component_param", "SubComponentParam.component_param"}
config  = {
    "MyTool": {"userparam": 12.0},
    "SubComponent": {"component_param": 15.0},
}
```

Does not work, intermediate
```
config  = {
    "MyTool": {
        "userparam": 12.0,
        "SubComponent": {"component_param": 15.0},
     }
}
```

Works, After:

```
aliases = {"component_param", "MyTool.SubComponentParam.component_param"}
config  = {
    "MyTool": {
        "userparam": 12.0,
        "SubComponent": {"component_param": 15.0},
     }
}
```

I don't think

maxnoe:
Ok, but `--help-all` does not work with aliases that have the full hierarchy...

maxnoe:
This seems to be a bug in traitlets.config.

Minimal example:

```
from traitlets.config import Application, Configurable
from traitlets import Int
import sys


class Test(Configurable):
    val = Int(default_value=5).tag(config=True)


class MyApp(Application):
    name = 'test'

    aliases = {'val': 'MyApp.Test.val'}

    def start(self):
        self.test = Test(parent=self)
        print(self.test.val)


if __name__ == "__main__":
    MyApp.launch_instance(sys.argv)
```

The alias works fine, but help-all raises an error

maxnoe:
I opened an issue: https://github.com/ipython/traitlets/issues/631

Let's see

ctapipe/tools/display_dl1.py
```diff
@@ -178,14 +178,11 @@ def __init__(self, **kwargs):
...
+        subarray = self.eventsource.subarray
 
-        self.calibrator = self.add_component(
-            CameraCalibrator(parent=self, subarray=self.eventsource.subarray)
-        )
-        self.plotter = self.add_component(
-            ImagePlotter(subarray=self.eventsource.subarray, parent=self)
-        )
+        self.calibrator = CameraCalibrator(parent=self, subarray=subarray)
+        self.plotter = ImagePlotter(subarray=subarray, parent=self)
```
dneise, line 185: I love this work.
But here I would keep the order or parameters consistent.

maxnoe: Ah yes, I'll fix that

maxnoe: done

maxnoe:
I  have no idea yet why how setting the config and the setting of the alias interact in this way, I'll put it on the long list of things we don't like about traitlets.

But I think this is good to go now and an improvement over the current state.

So please review again

kosack:
the idea of `traitlets.config` is really nice... the implementation is a huge mess, despite being used by big projects like Jupyter/IPython. I guess we will have to roll our own in the end.

kosack (approved):

kosack:
This is an API change right? The anywhere that uses Tools will have to be updated, so maybe we should not include this in 0.9.1

maxnoe:
@kosack I kept the add_componente function as a no-op that raises a deprecation warning, so it's not an API change

kosack (approved):

watsonjj:
> @kosack I kept the add_componente function as a no-op that raises a deprecation warning, so it's not an API change

I don't see this deprecation warning

maxnoe:
Because it didnt make it for the 0.9.1 release, I know completely removed it instead of deprecating

watsonjj (approved):

Issue #1414, opened 2020-09-10
Support multiple input files, appending or mergin  the output file for the stage1 tool


maxnoe:
During todays aswg meeting it came up that the prod5 proton production SimTel files are rather small and if processed one-by-one this would result in very many, very small DL1 files (~250.000).

It probably would be good to be able to run the stage1 on multiple simtel files and produce a single output or to be able to append to an existing output file to achieve the same by running the tool multiple times.

bregeon:
The reasons for this to happen is two fold:
1. jobs can run on farms usually from 12 to 24 hours, so that fixes the number of showers per job, hence the DL0 file size
2. DL1 file size is much smaller than DL0 file size, so you want some merge/reduce to happen in the number of files so that the analysis is then more efficient

Indeed, I think that part of the workflow will stay like that: either you have directly multiple input files and one output, or a way to append to a DL1 file or you have a merging tool.

bregeon:
To give one example, for Prod5, twe have 112000 DL0 files per site.

MaxNoe:
For now, a tool to merge the output of the DL1 stage seems sufficient. 

@Hckjs  is working on that

PR #1447, opened 2020-10-08
Dl1 merge tool

Hckjs:
First draft Version of a merge tool for DL1 files created by the stage1 tool. It merges all DL1 .h5 files with matching subarrays. It works but needs some changes regarding the way it merges the tables. It's a bit messy but i'm working on that.

**Some to do's**
- It tries to merge all DL1 .h5 files in the given directory, but fails if there is something else in this directory. Wildcards on a given file-scheme also doesn't work.

- The paths of the tables to be merged are handed in by a prearranged list, that could be done somehow automatically. It just merges all the tables in the lists for now.

refers to #1414


ctapipe/tools/dl1_merge.py
```diff
@@ -0,0 +1,211 @@
...
+        # Creates table for image statistics and adds the entries together.
+        # This does not append rows to the existing table
+        if table_path_image_statistics not in self.output_file:
+            head, tail = os.path.split(os.path.split(table_path_image_statistics)[0])
+            target_group = self.output_file.create_group(head, tail, createparents=True)
+            file.copy_node(table_path_image_statistics, newparent=target_group)
+            file.copy_node('/dl1/service/image_statistics.__table_column_meta__',
+                           newparent=target_group)
+        elif table_path_image_statistics in file:
+            for row in range(3):
```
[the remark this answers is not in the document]
maxnoe: assigning the table to a local variable like this might make it more readable:

```
table  = self.output_file.root[table_path_image_statistics]
```
```

**Comment by LukasNickel** [comment]
Nice!

Some things that I remember running into when I was trying this a while ago:
- On some files I had mismatching dtypes (#1411), which kinda broke the whole thing
- Does the `copy_node` function preserve the filters? If not it might be worth to add them to save some space
- For most tables the procedure is always the same: Append or copy if not existing. Thats why I just iterated over all nodes and blacklisted the configuration ones (that are supposed to not change)  for appending.
- Regarding the file patterns: You can use `glob` or `pathlib` to match. It's not a whole lot of work, but probably very nice when using the tool.

**Comment by kosack** [comment]
Great! This will be very useful.  This brings up some naming questions: originally I intended to have the tools  named `ctapipe-stage1-process` and `ctapipe-stage1-merge` (it's why it wasn't called just `ctapipe-stage1`.  If we want to go for naming by data level, that is also ok, but we should be consistent. Perhaps that can be another issue.  If the DL2 files are in a pretty similar format, of course this tool could be made to be more generic: it could merge DL1+DL1, DL2+DL2, or even join DL1+DL2 into a single file (in which case just calling it `ctapipe-merge` could be the cleanest)

File: ctapipe/tools/dl1_merge.py [hunk 2]
```diff
@@ -0,0 +1,211 @@
...
+                     (self.output_file.root[table_path_image_statistics].cols.
+                      cumulative_counts[row]),
+                      file.root[table_path_image_statistics].cols.cumulative_counts[row])
+
+    def merge_tables(self, file):
+        # 1) Create groups and copy tables if not in output_file
+        # 2) Append to table if it already exists in output_file
+
+        # For single tables in a group
+        for table_path_base in table_path_list_base:
```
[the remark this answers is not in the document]
maxnoe: Perhaps we really need to check the file first, before copying any data from it, as we don't want to append half of the tables of a broken file.

ctapipe/tools/dl1_merge.py
```diff
@@ -0,0 +1,211 @@
...
+
+            current_file_opened = tables.open_file(current_file, mode='r')
+            self.add_image_statistics(current_file_opened)
+            self.merge_tables(current_file_opened)
+            current_file_opened.close()
+
+            merged_files_counter += 1
+
+        self.log.info(f"{merged_files_counter} out of {len_input_path_list} Files "
+                       "had been merged!")
```
maxnoe, line 199: typo: have

ctapipe/tools/dl1_merge.py
```diff
@@ -0,0 +1,211 @@
...
+
+    def finish(self):
+        pass
+
+
+def main():
+    tool = DL1MergeTool()
```
maxnoe, line 206: I just learned that `Tools` have the `launch_instance` method.

So you can do: `DL1MergeTool.launche_instance()`

maxnoe (changes requested):
Thanks a lot!

THis will be very helpful.

Some comments.

ctapipe/tools/dl1_merge.py
```diff
@@ -0,0 +1,211 @@
...
+
+
+class DL1MergeTool(Tool):
+    name = "ctapipe-merge-dl1"
+    description = "Merges DL1-files from the stage1-process tool"
+    examples = """
+    To merge DL1-files created by the stage1-process-tool from a specific directory:
+    > ctapipe-merge-dl1 --input=/path/to/directory --ouput=/path/to/output_file.h5 --progress
+    """
+    input_dir = Path(help="input dl1-directory",
```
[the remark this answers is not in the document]
kosack: agreed.  you can then  just use `Path.glob(pattern)` to get the list of files, and by default have the pattern be all DL1 files.

ctapipe/tools/dl1_merge.py
```diff
@@ -0,0 +1,211 @@
...
+        # create output file with subarray from first file
+        self.first_subarray = SubarrayDescription.from_hdf(self.input_path_list[0])
+        self.first_subarray.to_hdf(self.output_path)
+        self.output_file = tables.open_file(self.output_path, mode='a')
+
+    def add_image_statistics(self, file):
+        # Creates table for image statistics and adds the entries together.
+        # This does not append rows to the existing table
+        if table_path_image_statistics not in self.output_file:
+            head, tail = os.path.split(os.path.split(table_path_image_statistics)[0])
```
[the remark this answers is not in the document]
kosack: I suggest also to use `pathlib` instead of os.path everywhere.  It's much cleaner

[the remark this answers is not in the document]
maxnoe: This is a path inside a hdf5 file, so I think just using string operation is better.

ctapipe/tools/dl1_merge.py
kosack: have you tried it? I just get an error about two instances. Not sure how it is supposed to work

maxnoe: Did you try this in `ipython`? Because I think it conflicts because ipython itself is already an instance of `Application` and only one can live per process.

maxnoe: This works:

```
❯ python
Python 3.7.7 (default, Mar 26 2020, 15:48:22) 
[GCC 7.3.0] :: Anaconda, Inc. on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> from ctapipe.core import Tool
>>> Tool.launch_instance()
INFO [Tool] (tool/initialize): ctapipe version 0.9.0.post9+gite02948a
```

Hckjs:
> Nice!
> 
> Some things that I remember running into when I was trying this a while ago:

Thanks for all the comments! That will help a lot.
> 
>     * Does the `copy_node` function preserve the filters? If not it might be worth to add them to save some space
What do you mean by the 'filters' ?
> 
>     * For most tables the procedure is always the same: Append or copy if not existing. Thats why I just iterated over all nodes and blacklisted the configuration ones (that are supposed to not change)  for appending.

Thats a good idea, i was already thinking about doing it somehow this way.

>     * Regarding the file patterns: You can use `glob` or `pathlib` to match. It's not a whole lot of work, but probably very nice when using the tool.

Thanks!

maxnoe:
Filters refer to compression options. 

> Additional keyword arguments may be passed to customize the copying process. For instance, title and filters may be changed, user attributes may be or may not be copied, data may be sub-sampled, stats may be collected, etc. See the documentation for the particular node type.

This sounds like by default, they are not changed but could be if additional kwargs are used.

maxnoe:
The defaults all look good, user attributes are copied and the filters are preserved.

ctapipe/tools/dl1_merge.py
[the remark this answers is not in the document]
Hckjs: I switched it now, so you can just use specific files as input or using a file pattern as `--pattern='pattern*.h5'` . If you don't give any files nor pattern, it takes all .h5 files in current directory (default file_pattern='*.h5*)

ctapipe/tools/dl1_merge.py
```diff
@@ -82,17 +98,24 @@ def setup(self):
...
+
         if self.skip_parameters is True and self.skip_images is True:
             self.log.warning('Skip-parameters and skip-images are both'
                              'set to True')
 
-        self.input_path_list = sorted(self.input_dir.iterdir())
+        args = self.parser.parse_args(self.extra_args)
+        self.input_files = args.input_files
+        if not self.input_files:
+            self.input_files = sorted(Path('.').glob(self.file_pattern))
```
maxnoe, line 110: I think you should still keep input dir. And do something like:

```
if self.input_dir is not None:
    self.input_files.extend(sorted(...)))
```

ctapipe/tools/dl1_merge.py
```diff
@@ -0,0 +1,234 @@
...
+
+class MergeTool(Tool):
+    name = "ctapipe-merge"
+    description = "Merges DL1-files from the stage1-process tool"
+    examples = """
+    To merge DL1-files created by the stage1-process-tool from current directory:
+
+    > ctapipe-merge file1.h5 file2.h5 file3.h5 --ouput=/path/output_file.h5 --progress
+
+    If you want a specific file pattern as input files use --patter='pattern.*.dl1.h5', e.g.:
```
kosack, line 43: Just a typo: should be `--pattern` not `--patter`

ctapipe/tools/dl1_merge.py
```diff
@@ -0,0 +1,211 @@
...
+    examples = """
+    To merge DL1-files created by the stage1-process-tool from a specific directory:
+    > ctapipe-merge-dl1 --input=/path/to/directory --ouput=/path/to/output_file.h5 --progress
+    """
+    input_dir = Path(help="input dl1-directory",
+                     exists=None, directory_ok=True, file_ok=False).tag(config=True)
+    output_path = Path(help="Merged-DL1 output filename").tag(config=True)
+    skip_images = Bool(help="Skip DL1/Event/Image data in output",
+                       default_value=False).tag(config=True)
+    skip_parameters = Bool(help="Skip image parameters",
```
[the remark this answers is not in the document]
Hckjs: Think so too, it's just a little bit of code more, maybe someone uses it. I would leave it there if its ok.

ctapipe/tools/dl1_merge.py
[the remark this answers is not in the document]
Hckjs: I needed it twice, because there was the node at the end of this path, so i had to split it once to get the path from the node as 'head' and twice to create the group, because i need the end from 'head' of first split. But i changed it anyway

ctapipe/tools/dl1_merge.py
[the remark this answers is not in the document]
maxnoe: ACtually we came up with a use case for this. As soon as we merge the dl1 tool with the muon analysis, you might only want to merge the muon parameters, not all parameters.

ctapipe/tools/dl1_merge.py
Hckjs: If i switch it to
```
def main():
    MergeTool.launch_instance()
```
it initializes and starts correctly, but doesn't do anything. Am i missing something?

Hckjs:
**New Update:**
I've added the Block/Allowlists and a `check_file`-function. There are now mainly two allowlists (there are more, but the others are just for correct looping). 

- In `Allowlist_base` is every node except the ones made by the SubarrayDescription. If any of these is missing in the File (except you make specific constellations with the traits, like skip-images or something), you will get a warning and the whole file will be skipped.

- In `Allowlist_unessential` are nodes, which doesn't have to be in the merging file. That means that you get a warning if one of these is missing, but the rest of the file would still get merged. I've already put the following into it:
```
allowlist_unessential = ['/simulation/service/shower_distribution',
                         '/simulation/event/telescope/images',
                         '/simulation/event/telescope/parameters']
```

maxnoe:
So, better name for `allowlist_base` would maybe be `required_nodes` and for `allowlist_unessestial` `optional_nodes`?

ctapipe/tools/dl1_merge.py
```diff
@@ -34,15 +34,15 @@
                   '/simulation/event/telescope/parameters',
                   '/simulation/event/telescope/images',
                   '/simulation/service/shower_distribution']
+optional_nodes = ['/simulation/service/shower_distribution',
+                  '/simulation/event/telescope/images',
```
maxnoe, line 38: This is duplicated to the below elements in `allow_list` simu. Is that needed?

maxnoe (changes requested):
This looks good now, I think.

The only thing missing would be provenance information.

So at the very least you need to add `Provenance().add_input()` for all input files and `Provenance().add_output()` to the output files.

ctapipe/tools/dl1_merge.py
Hckjs: I need the `allowlist_simu` for checking the file. When its real data and doesn't have the `simulation`-group, it should not give out a warning (so i can skip these groups in checking). `/simulation/event/subarray/shower` is also just in `allowlist_simu`. And i was thinking that its more clearly this way and you can just add groups for future purpose.

maxnoe:
I ran the current status on this branch on the output of the stage 1 tool from ctapipe 0.8 and it worked nicely (with `--skip-images --skip-simu-images`):

```
-rw-rw---- 1 mnoethe app 7.3G Oct 20 19:08 gamma_diffuse_dl1b.h5
-rw-rw---- 1 mnoethe app  12G Oct 20 20:16 gamma_dl1b.hdf5
-rw-rw---- 1 mnoethe app 7.9G Oct 20 21:49 proton_dl1b.hdf5
```

These were 5099, 1026 and 4973 files respectively.

A small number of files was not included, e.g. for the protons:

```

$ ctapipe-merge --output /net/cta-tank/POOL/projects/cta/Prod3-LaPalma-Benchmark/stage-1/0.8.0_fixed/proton_dl1b.hdf5 /net/cta-tank/POOL/projects/cta/Prod3-LaPalma-Benchmark/stage-1/0.8.0_fixed/protons/proton_20deg_180deg_run*.dl1.h5 --progress --overwrite --skip-images --skip-simu-images                                                                                                                                         
INFO [MergeTool] (tool/initialize): ctapipe version 0.9.0.post16+gitd76f1b0INFO [MergeTool] (tool/run): Starting: ctapipe-mergeMerging:   0%|                                                                                                    | 0/4973 [00:00<?, ?Files/s]INFO [MergeTool] (dl1_merge/start): Merging simulation-filesMerging:  41%|███████████████████████████████████▍                                                   | 2023/4973 [57:39<1:24:11,  1.71s/Files]WARNING [MergeTool] (dl1_merge/check_file): /dl1/service/image_statistics is not in /net/cta-tank/POOL/projects/cta/Prod3-LaPalma-Benchmark/stage-1/0.8.0_fixed/protons/proton_20deg_180deg_run2867___cta-prod3-demo-2147m-LaPalma-baseline.dl1.h5. Skip fileWARNING [MergeTool] (dl1_merge/check_file): /dl1/service/image_statistics.__table_column_meta__ is not in /net/cta-tank/POOL/projects/cta/Prod3-LaPalma-Benchmark/stage-1/0.8.0_fixed/protons/proton_20deg_180deg_run2867___cta-prod3-demo-2147m-LaPalma-baseline.dl1.h5. Skip fileWARNING [MergeTool] (dl1_merge/check_file): /simulation/service/shower_distribution is not in /net/cta-tank/POOL/projects/cta/Prod3-LaPalma-Benchmark/stage-1/0.8.0_fixed/protons/proton_20deg_180deg_run2867___cta-prod3-demo-2147m-LaPalma-baseline.dl1.h5. Continue with merging file. This table will be incomplete or empty
Merging:  46%|███████████████████████████████████████                                              | 2288/4973 [1:05:11<1:16:13,  1.70s/Files]WARNING [MergeTool] (dl1_merge/check_file): /dl1/service/image_statistics is not in /net/cta-tank/POOL/projects/cta/Prod3-LaPalma-Benchmark/stage-1/0.8.0_fixed/protons/proton_20deg_180deg_run3108___cta-prod3-demo-2147m-LaPalma-baseline.dl1.h5. Skip file
WARNING [MergeTool] (dl1_merge/check_file): /dl1/service/image_statistics.__table_column_meta__ is not in /net/cta-tank/POOL/projects/cta/Prod3-LaPalma-Benchmark/stage-1/0.8.0_fixed/protons/proton_20deg_180deg_run3108___cta-prod3-demo-2147m-LaPalma-baseline.dl1.h5. Skip file
WARNING [MergeTool] (dl1_merge/check_file): /simulation/service/shower_distribution is not in /net/cta-tank/POOL/projects/cta/Prod3-LaPalma-Benchmark/stage-1/0.8.0_fixed/protons/proton_20deg_180deg_run3108___cta-prod3-demo-2147m-LaPalma-baseline.dl1.h5. Continue with merging file. This table will be incomplete or empty
Merging:  70%|████████████████████████████████████████████████████████████▋                          | 3466/4973 [1:39:11<39:23,  1.57s/Files]WARNING [MergeTool] (dl1_merge/check_file): /simulation/service/shower_distribution is not in /net/cta-tank/POOL/projects/cta/Prod3-LaPalma-Benchmark/stage-1/0.8.0_fixed/protons/proton_20deg_180deg_run4213___cta-prod3-demo-2147m-LaPalma-baseline.dl1.h5. Continue with merging file. This table will be incomplete or empty
Merging:  71%|█████████████████████████████████████████████████████████████▍                         | 3509/4973 [1:40:22<38:28,  1.58s/Files]WARNING [MergeTool] (dl1_merge/check_file): /simulation/service/shower_distribution is not in /net/cta-tank/POOL/projects/cta/Prod3-LaPalma-Benchmark/stage-1/0.8.0_fixed/protons/proton_20deg_180deg_run4253___cta-prod3-demo-2147m-LaPalma-baseline.dl1.h5. Continue with merging file. This table will be incomplete or empty
Merging:  88%|████████████████████████████████████████████████████████████████████████████▍          | 4366/4973 [2:04:52<17:23,  1.72s/Files]WARNING [MergeTool] (dl1_merge/check_file): /dl1/service/image_statistics is not in /net/cta-tank/POOL/projects/cta/Prod3-LaPalma-Benchmark/stage-1/0.8.0_fixed/protons/proton_20deg_180deg_run5036___cta-prod3-demo-2147m-LaPalma-baseline.dl1.h5. Skip file
WARNING [MergeTool] (dl1_merge/check_file): /dl1/service/image_statistics.__table_column_meta__ is not in /net/cta-tank/POOL/projects/cta/Prod3-LaPalma-Benchmark/stage-1/0.8.0_fixed/protons/proton_20deg_180deg_run5036___cta-prod3-demo-2147m-LaPalma-baseline.dl1.h5. Skip file
WARNING [MergeTool] (dl1_merge/check_file): /simulation/service/shower_distribution is not in /net/cta-tank/POOL/projects/cta/Prod3-LaPalma-Benchmark/stage-1/0.8.0_fixed/protons/proton_20deg_180deg_run5036___cta-prod3-demo-2147m-LaPalma-baseline.dl1.h5. Continue with merging file. This table will be incomplete or empty
Merging: 100%|███████████████████████████████████████████████████████████████████████████████████████| 4973/4973 [2:21:57<00:00,  1.71s/Files]
INFO [MergeTool] (dl1_merge/start): 4970 out of 4973 files has been merged!
INFO [MergeTool] (tool/run): Finished: ctapipe-merge
INFO [MergeTool] (tool/run): Output: /net/cta-tank/POOL/projects/cta/Prod3-LaPalma-Benchmark/stage-1/0.8.0_fixed/proton_dl1b.hdf5
Closing remaining open files:/net/cta-tank/POOL/projects/cta/Prod3-LaPalma-Benchmark/stage-1/0.8.0_fixed/proton_dl1b.hdf5...done
```

@LukasNickel could you check why these 3 runs are broken? Did the original processing crash?

LukasNickel:
> @LukasNickel could you check why these 3 runs are broken? Did the original processing crash?

Just tested it again on the `proton_20deg_180deg_run2867___cta-prod3-demo-2147m-LaPalma-baseline.simtel.gz` one.
It is indeed failing on the `process_events` part:

>ERROR [Stage1ProcessorTool] (tool/run): Caught unexpected exception: unpack requires a buffer of 4 bytes
  File "/net/nfshome/home/lnickel/ctapipe/ctapipe/core/tool.py", line 225, in run
    self.start()
  File "/net/nfshome/home/lnickel/ctapipe/ctapipe/tools/stage1.py", line 685, in start
    self._process_events(writer)
  File "/net/nfshome/home/lnickel/ctapipe/ctapipe/tools/stage1.py", line 468, in _process_events
    for event in tqdm(
  File "/home/lnickel/conda/envs/ctapipe_study/lib/python3.8/site-packages/tqdm/std.py", line 1152, in __iter__
    for obj in iterable:
  File "/net/nfshome/home/lnickel/ctapipe/ctapipe/io/eventsource.py", line 254, in __iter__
    for event in self._generator():
  File "/net/nfshome/home/lnickel/ctapipe/ctapipe/io/simteleventsource.py", line 338, in _generator
    yield from self._generate_events()
  File "/net/nfshome/home/lnickel/ctapipe/ctapipe/io/simteleventsource.py", line 355, in _generate_events
    for counter, array_event in enumerate(self.file_):
  File "/home/lnickel/conda/envs/ctapipe_study/lib/python3.8/site-packages/eventio/simtel/simtelfile.py", line 239, in iter_array_events
    self.next_low_level()
  File "/home/lnickel/conda/envs/ctapipe_study/lib/python3.8/site-packages/eventio/simtel/simtelfile.py", line 144, in next_low_level
    self.current_mc_shower = o.parse()
  File "/home/lnickel/conda/envs/ctapipe_study/lib/python3.8/site-packages/eventio/simtel/objects.py", line 1326, in parse
    mc['cmax'] = read_float(byte_stream)
  File "/home/lnickel/conda/envs/ctapipe_study/lib/python3.8/site-packages/eventio/tools.py", line 28, in read_float
    return struct.unpack('<f', f.read(4))[0]
struct.error: unpack requires a buffer of 4 bytes

maxnoe:
@LukasNickel It seems the file was only partially downloaded

ctapipe/tools/dl1_merge.py
```diff
@@ -0,0 +1,327 @@
...
+from tqdm import tqdm
+
+from ..core import Provenance
+from ctapipe.core import Tool, traits
+from traitlets import List
+from ctapipe.instrument import SubarrayDescription
+
+import warnings
+
+warnings.filterwarnings('ignore', category=tables.NaturalNameWarning)
```
maxnoe: Why is this needed? Why are those warnings created?

ctapipe/tools/dl1_merge.py
```diff
@@ -0,0 +1,359 @@
...
+from tqdm import tqdm
+
+from ..core import Provenance
+from ctapipe.core import Tool, traits
+from traitlets import List
+from ctapipe.instrument import SubarrayDescription
+
+import warnings
+
+warnings.filterwarnings("ignore", category=tables.NaturalNameWarning)
```
maxnoe, line 20: Why are  these warnings filtered? Why are they generated in the first place?

ctapipe/tools/dl1_merge.py
```diff
@@ -0,0 +1,359 @@
...
+
+PROV = Provenance()
+
+required_nodes = [
+    "/dl1/monitoring/subarray/pointing",
+    "/dl1/monitoring/telescope/pointing",
+    "/dl1/service/image_statistics",
```
maxnoe, line 27: I don't think these image statistics should be required. There are "nice to have"

ctapipe/tools/dl1_merge.py
```diff
@@ -0,0 +1,359 @@
...
+    "/dl1/service/image_statistics.__table_column_meta__",
+    "/dl1/event/subarray/trigger",
+    "/dl1/event/telescope/trigger",
+    "/dl1/event/telescope/parameters",
```
maxnoe, line 31: Both images and parameters should also be optional, otherwise we cannot merge files that were created without them. (only parameters / only images)

ctapipe/tools/dl1_merge.py
```diff
@@ -0,0 +1,359 @@
...
+    "/dl1/event/telescope/images",
+    "/configuration/simulation/run",
```
maxnoe, line 33: These is simulation info and thus should only be required for simulated files.

I suggest to make a second list for `required_nodes_simulation`

ctapipe/tools/dl1_merge.py
```diff
@@ -0,0 +1,359 @@
...
+    "/simulation/event/subarray/shower",
+    "/simulation/event/telescope/parameters",
+    "/simulation/event/telescope/images",
+    "/simulation/service/shower_distribution",
```
maxnoe, line 37: It might be necessary to not force this. Or mabye add an option `--do-no-require-shower-distribution` since most of our test files are truncated.

ctapipe/tools/dl1_merge.py
```diff
@@ -0,0 +1,359 @@
...
+]
+optional_nodes = [
+    "/simulation/service/shower_distribution",
```
maxnoe, line 40: Wait, this is both here and in `required_nodes`

ctapipe/tools/dl1_merge.py
```diff
@@ -0,0 +1,359 @@
...
+    "/simulation/event/telescope/images",
+    "/simulation/event/telescope/parameters",
+]
+allowlist_simu = [
```
maxnoe, line 44: Be consistent with `allow` vs. `optional`

ctapipe/tools/dl1_merge.py
```diff
@@ -0,0 +1,359 @@
...
+
+    If neither any input files nor any pattern is given, all files from the current directory
+    with the default pattern '*.h5' are taken. For merging all files from a specific
+    directory with a given pattern, use:
+
+    > ctapipe-merge --input-dir=/input/dir/ --output=/path/output_file.h5 --progress
+    --pattern='pattern.*.dl1.h5'
+    """
+    input_dir = traits.Path(
+        help="input dl1-directory", exists=None, directory_ok=True, file_ok=False
```
maxnoe, line 88: This must exist, right? It is an input after all.

ctapipe/tools/dl1_merge.py
```diff
@@ -0,0 +1,359 @@
...
+    ).tag(config=True)
+    input_files = List(default_value=[], help="input dl1-files").tag(config=True)
+    output_path = traits.Path(help="merged-DL1 output filename").tag(config=True)
```
maxnoe, line 91: `directory_ok=False`

ctapipe/tools/dl1_merge.py
```diff
@@ -0,0 +1,359 @@
...
+        help="skip /dl1/event/telescope/parameters and "
+        "/simulation/event/telescope/parameters "
+        "data in output",
+        default_value=False,
+    ).tag(config=True)
+    overwrite = traits.Bool(help="overwrite output file if it exists").tag(config=True)
+    progress_bar = traits.Bool(help="show progress bar during " "processing").tag(
+        config=True
+    )
+    file_pattern = traits.Unicode(
```
maxnoe, line 111: Is a glob pattern enough here or do we need a regex?

ctapipe/tools/dl1_merge.py
```diff
@@ -0,0 +1,359 @@
...
+        default_value="*.h5", help="Give a specific file pattern for the" "input files"
+    ).tag(config=True)
+
+    parser = ArgumentParser()
+    parser.add_argument("input_files", nargs="*", type=Path)
+
+    aliases = {
+        "input-dir": "MergeTool.input_dir",
+        "output": "MergeTool.output_path",
```
maxnoe, line 120: just `o` would be nice to have here

ctapipe/tools/dl1_merge.py
```diff
@@ -0,0 +1,359 @@
...
+    )
+    file_pattern = traits.Unicode(
+        default_value="*.h5", help="Give a specific file pattern for the" "input files"
+    ).tag(config=True)
+
+    parser = ArgumentParser()
+    parser.add_argument("input_files", nargs="*", type=Path)
+
+    aliases = {
+        "input-dir": "MergeTool.input_dir",
```
maxnoe, line 119: just `i` would be nice to have here

ctapipe/tools/dl1_merge.py
```diff
@@ -0,0 +1,359 @@
...
+        "output": "MergeTool.output_path",
+        "pattern": "MergeTool.file_pattern",
```
maxnoe, line 121: just `p` would be nice to have here

ctapipe/tools/dl1_merge.py
```diff
@@ -0,0 +1,359 @@
...
+
+        if self.skip_parameters is True and self.skip_images is True:
+            self.log.warning("Skip-parameters and skip-images are both" "set to True")
+
+        # Get input Files
+        args = self.parser.parse_args(self.extra_args)
+        self.input_files = args.input_files
+        if self.input_dir is not None:
+            self.input_files.extend(sorted(self.input_dir.glob(self.file_pattern)))
+        if not self.input_files:
```
maxnoe, line 171: I think I would just error here at this point.

```
self.log.critical('No input files provided, either provide --input-dir or input files as positional arguments')
sys.exit(1)
```
```

File: ctapipe/tools/dl1_merge.py [hunk 26]
```diff
@@ -0,0 +1,359 @@
...
+                sys.exit(1)
+
+        PROV.add_output_file(str(self.output_path))
+
+        if self.skip_parameters is True and self.skip_images is True:
+            self.log.warning("Skip-parameters and skip-images are both" "set to True")
+
+        # Get input Files
+        args = self.parser.parse_args(self.extra_args)
+        self.input_files = args.input_files
```
maxnoe, line 168: I think to make it work with paths given in the config file, this also needs to be `extend(args.input_files)`

ctapipe/tools/dl1_merge.py
```diff
@@ -0,0 +1,359 @@
...
+        self.first_subarray = SubarrayDescription.from_hdf(self.input_files[0])
+        self.first_subarray.to_hdf(self.output_path)
+        self.output_file = tables.open_file(self.output_path, mode="a")
+
+    def check_file(self, file):
+        # Check that the file is not broken or any node is missing
+        file_path = file.root._v_file.filename
+        current_subarray = SubarrayDescription.from_hdf(file_path)
+        # Check subarrays
+        if self.first_subarray != current_subarray:
```
maxnoe, line 184: I think we should error in this case. We should think about the main use case of this file:

Massive processing of many files on the grid. If a job is successfull, we don't look into the log. So any error should be a real error not just a warning.

maxnoe: Maybe an option `--ignore-broken` could be added, to allow users that "know what they are doing"™ to ignore things like these.

But the default case should be an error with a non-zero exit code.

This goes for everything here. Fail loud, fail early. Otherwise errors will get unnoticed on the grid / clusters and subtle errors hard to find are introduced.

ctapipe/tools/dl1_merge.py
```diff
@@ -0,0 +1,359 @@
...
+            self.log.warning(f"Subarray does not match for {file_path}. " "Skip File.")
+            self.skip_file = True
+
+        # Gives warning if tables for listed nodes in 'optional_nodes'
+        # are missing but continues merging the rest of the file. If any
+        # other node from 'required_nodes' is missing, except the given combinations
+        # of traits and blocklist, the whole file will be skipped.
```
maxnoe, line 191: Same as above. This should create an error by default, not skipping.

ctapipe/tools/dl1_merge.py
```diff
@@ -0,0 +1,359 @@
...
+            if self.skip_parameters is True and node in blocklist_parameters:
+                continue
+            if node in optional_nodes and node not in file:
+                self.log.warning(
+                    f"{node} is not in {file_path}. Continue with "
+                    "merging file. This table will be incomplete "
+                    "or empty"
+                )
+                continue
+            if node not in file:
```
maxnoe, line 208: See comment above.

ctapipe/tools/dl1_merge.py
```diff
@@ -0,0 +1,359 @@
...
+        if self.first_subarray != current_subarray:
+            self.log.warning(f"Subarray does not match for {file_path}. " "Skip File.")
+            self.skip_file = True
+
+        # Gives warning if tables for listed nodes in 'optional_nodes'
+        # are missing but continues merging the rest of the file. If any
+        # other node from 'required_nodes' is missing, except the given combinations
+        # of traits and blocklist, the whole file will be skipped.
+        for node in required_nodes:
+            if self.is_simu is False and node in allowlist_simu:
```
maxnoe, line 193: It might be better to build the allowlists / required lists / ... once in `setup` according to the provided options and then work with single sets in this check here.

maxnoe: Also, membership checks are better done with sets than lists.

ctapipe/tools/dl1_merge.py
```diff
@@ -0,0 +1,359 @@
...
+                )
+                continue
+            if node not in file:
+                self.log.warning(f"{node} is not in {file_path}. Skip file")
+                self.skip_file = True
+
+    def add_image_statistics(self, file):
+        # Creates table for image statistics and adds the entries together.
+        # This does not append rows to the existing table
+        image_statistics_path = allowlist_service[0]
```
maxnoe, line 215: This is not good. If someone reorders allowlist (for whatever reason), this will break. Please just give the needed path here.

ctapipe/tools/dl1_merge.py
```diff
@@ -0,0 +1,359 @@
...
+        # Loop over all nodes listed in required_nodes. Appends table to output_file
+        # if it already exists, otherwise creates group and copies node. If skip_images
+        # or skip_simu_images or skip_parameters flag is True, related groups
+        # will be skipped.
+        for node in required_nodes:
+            if node in file:
+                if self.skip_images is True and node in blocklist_images:
+                    continue
+
+                if self.skip_simu_images is True and node in blocklist_images[0]:
```
maxnoe, line 245: Same as above, please don't rely on the order of things in these lists.

maxnoe: Also, why are you using `node in  blocklist_images[0]`? that tests if a string is in another string, right?

Better use `startswith`, probably like this: `node.startswith('/dl1/image/tel/tel_'`)

ctapipe/tools/dl1_merge.py
```diff
@@ -0,0 +1,359 @@
...
+                    # TODO: remove when we no longer want to support merging 0.8 files.
+                    data = file.root[node][:]
+                    if node == "/dl1/monitoring/subarray/pointing":
+                        data = self.drop_column(data, "tels_with_trigger")
+
+                    output_node = self.output_file.get_node(node)
+                    output_node.append(data)
+                else:
+                    group_path, table_name = os.path.split(node)
+                    if group_path not in self.output_file:
```
maxnoe, line 285: This is the same code as above, right? So please move this to a small method like `get_or_create_group` and use it to avoid duplicatoin.

ctapipe/tools/dl1_merge.py
```diff
@@ -0,0 +1,359 @@
...
+        for i, current_file in enumerate(
+            tqdm(
+                self.input_files,
+                desc="Merging",
+                unit="Files",
+                disable=not self.progress_bar,
+            )
+        ):
+
+            self.skip_file = False
```
maxnoe, line 326: See commets above about skipping files.

maxnoe (changes requested):
This looks very good.

I made some comments inline.

The most important point is, that the default should always be to fail loud and fail early.
This means not just "silently" (as in only logging a warning) skipping over broken files.

As commented inline, this script will most often run unsupervised on the GRID and it is essential that a job fails when something breaks so we are notified of a failure and can investigate.

Nobody will read the logs of successfull jobs unless we find subtle bugs afterwards.

For local use / expert users a flag `--skip-broken-file` could maybe be handy, but like I said, the default should be just to fail with an error message and non-zero exit code for anything going wrong.

LukasNickel:
> I ran the current status on this branch on the output of the stage 1 tool from ctapipe 0.8 and it worked nicely (with `--skip-images --skip-simu-images`):

Have you checked the table attributes/metadata?
They seem to be missing, which would make sense if all you do is adding the nodes under the root

maxnoe:
The first node is copied and then appended to. I thought copying the first node keeps the metadata.

Which metadata is missing?

LukasNickel:
The reference metadata.

```python
import tables

table = tables.open_file(...)
print(table.root._v_attrs)
```

maxnoe:
Ah, at the root! That makes sense, since the root cannot be copied. Good catch.

ctapipe/tools/dl1_merge.py
Hckjs: There is a NaturalNameWarning for `/dl1/service/image_statistics.__table_column_meta__` :
```
/home/jonas/anaconda3/envs/ma/lib/python3.7/site-packages/tables/path.py:155: NaturalNameWarning: object name is not a valid Python identifier: 'image_statistics.__table_column_meta__'; it does not match the pattern ``^[a-zA-Z_][a-zA-Z0-9_]*$``; you will not be able to use natural naming to access this object; using ``getattr()`` will still work, though
  check_attribute_name(name)
```
So you can't reach it just by `file.root.dl1.service.image_statistics.__table_column_meta_` but with `file.root['/dl1/service/image_statistics.__table_column_meta_']` still works. For the way it merges, its not important i think.

ctapipe/tools/dl1_merge.py
Hckjs: So that means it wouldn't be bad if we want to merge images/parameters, but they are missing in some files? It would make a log.warning but continues with merging?

I was thinking at first of something like: If you know, you want to merge files that were created without images/parameters, then you have to flag --skip_images/ --skip_parameters otherwise it will fail (now).
Or could it happen that somehow you want to merge files were (e.g.) half of them are without images/parameters?

kosack:
> Edit: probably this info should not be copied but created, since this is a new operation.

yes, since this is a new product, it needs to be created for it (shouldn't be hard though). We might have to think about how to set a few fields.   In principle these headers + the provenance log allow you to look at a merged DL1 file, and see where it came from (including a list of all inputs files in the provenance log)

ctapipe/tools/dl1_merge.py
Hckjs: It was very unconvenient from me here to use the name of the list 'required_nodes', because it was supposed to have all nodes in it and then just do certain stuff if its e.g. a simu file or so... I will rename it to have a 'all_nodes' -list and then create a set of it ('required_nodes') with all the given requirements like --skip-something in it.

ctapipe/tools/dl1_merge.py
maxnoe: It could always happen that a run did not trigger any telescopes (especially with small proton files). So this should probably be not an error case.

ctapipe/tools/dl1_merge.py
Hckjs: Can't think for now of a good use case for regexp where glob pattern doesn't work or is too intricate, do you?

ctapipe/tools/dl1_merge.py
```diff
@@ -85,27 +85,32 @@ class MergeTool(Tool):
...
     skip_parameters = traits.Bool(
-        help="skip /dl1/event/telescope/parameters and "
-        "/simulation/event/telescope/parameters "
-        "data in output",
+        help="Skip DL1/Event/Telescope and Simulation/Event/Telescope parameters"
+        "in output",
+        default_value=False,
+    ).tag(config=True)
+    skip_broken_files = traits.Bool(
+        help="Skip broken files instead of giving error",
```
maxnoe, line 109: replace "giving" -> "raising an"

ctapipe/tools/dl1_merge.py
```diff
@@ -85,27 +85,32 @@ class MergeTool(Tool):
...
         default_value=False,
     ).tag(config=True)
-    overwrite = traits.Bool(help="overwrite output file if it exists").tag(config=True)
-    progress_bar = traits.Bool(help="show progress bar during " "processing").tag(
+    overwrite = traits.Bool(help="Overwrite output file if it exists").tag(config=True)
+    progress_bar = traits.Bool(help="Show progress bar during " "processing").tag(
```
maxnoe, line 113: Unneeded splitting of the string

ctapipe/tools/dl1_merge.py
```diff
@@ -116,9 +121,9 @@ class MergeTool(Tool):
     parser.add_argument("input_files", nargs="*", type=Path)
 
     aliases = {
-        "input-dir": "MergeTool.input_dir",
-        "output": "MergeTool.output_path",
-        "pattern": "MergeTool.file_pattern",
+        "i": "MergeTool.input_dir",
+        "o": "MergeTool.output_path",
+        "p": "MergeTool.file_pattern",
```
maxnoe, line 126: I wouldn't have replaced them, you added the short forms addionally

maxnoe: So `-p, --pattern`, `-o, --output`, `-i, --input-dir` would be good, I think

ctapipe/tools/dl1_merge.py
```diff
@@ -161,58 +170,79 @@ def setup(self):
...
+            )
+            sys.exit(1)
 
         # create output file with subarray from first file
         self.first_subarray = SubarrayDescription.from_hdf(self.input_files[0])
         self.first_subarray.to_hdf(self.output_path)
         self.output_file = tables.open_file(self.output_path, mode="a")
 
+        # setup required nodes
+        self.usable_nodes = set(all_nodes)
```
maxnoe, line 195: Why not immediately define as sets?

maxnoe: so using `{` instead of `[`

ctapipe/tools/dl1_merge.py
```diff
@@ -161,58 +170,79 @@ def setup(self):
...
         file_path = file.root._v_file.filename
         current_subarray = SubarrayDescription.from_hdf(file_path)
-        # Check subarrays
+
+        # Check subarray
         if self.first_subarray != current_subarray:
-            self.log.warning(f"Subarray does not match for {file_path}. " "Skip File.")
-            self.skip_file = True
+            self.log.critical(f"Subarray does not match for {file_path}")
+            self.broken_file = True
```
maxnoe, line 223: Why not raise the error / log here? Instead of setting a member variable.

This member variable introduces some unneeded state I think.

ctapipe/tools/dl1_merge.py
```diff
@@ -0,0 +1,388 @@
...
+                    f"Output file {self.output_path} exists, "
+                    "use `--overwrite` to overwrite"
+                )
+                sys.exit(1)
+
+        PROV.add_output_file(str(self.output_path))
+
+        if self.skip_parameters is True and self.skip_images is True:
+            self.log.warning("Skip-parameters and skip-images are both set to True")
+            sys.exit(1)
```
maxnoe, line 174: Someone could be interested in e.g. only extracting trigger times or simulation information. So maybe leave the warnings but for sure don't exit.

ctapipe/tools/dl1_merge.py
```diff
@@ -0,0 +1,388 @@
...
+
+        # setup required nodes
+        self.usable_nodes = set(all_nodes)
+
+        if self.skip_simu_images is True:
+            simu_images = ["/simulation/event/telescope/images"]
+            self.usable_nodes = self.usable_nodes - set(simu_images)
+
+        if self.skip_images is True:
+            images_path = [
```
maxnoe, line 202: Move to top and directly define as set?

ctapipe/tools/dl1_merge.py
```diff
@@ -0,0 +1,388 @@
...
+        ):
+
+            self.broken_file = False
+            with tables.open_file(current_file, mode="r") as file:
+                if i == 0:
+                    # Check if first file is simulation
+                    if "/simulation" not in file.root:
+                        self.usable_nodes = self.usable_nodes - set(simu_nodes)
+                    else:
+                        self.log.info("Merging simulation-files")
```
maxnoe, line 358: Add a log also for the other case?

ctapipe/tools/dl1_merge.py
```diff
@@ -0,0 +1,388 @@
...
+
+                self.check_file(file)
```
maxnoe, line 360: Better return True / False than setting a member variable. Avoid such state where possible.

ctapipe/tools/dl1_merge.py
```diff
@@ -0,0 +1,388 @@
...
+
+            PROV.add_input_file(str(current_file))
+            merged_files_counter += 1
+
+        self.log.info(
+            f"{merged_files_counter} out of {len(self.input_files)} files "
+            "has been merged!"
+        )
+
+    def finish(self):
```
maxnoe, line 378: Write the metadata here.

ctapipe/tools/dl1_merge.py
Hckjs: Ah ok that makes sense.

ctapipe/tools/dl1_merge.py
Hckjs: whups, misunderstood that

ctapipe/tools/dl1_merge.py
Hckjs: somehow leftover from black

ctapipe/tools/dl1_merge.py
Hckjs: I thought it might be good to know if anything else is missing in that file, so that the whole check_file() could run through and raising all critical/warnings and in addition to that i would have just one if clause at the end (with 'self.skip_broken_files) to raise the error.

maxnoe: You can still do that with a local variable. 

```
broken = False

if check_one_failed:
    broken = True

if check_two_failed:
   broken = True

return broken
```

maxnoe (dismissed):
Looks good now! Thank you very much.

@kosack Can you have a look over the metadata? I wasn't sure about that.

ctapipe/tools/dl1_merge.py
```diff
@@ -0,0 +1,430 @@
...
+                description="Merged DL1 Data Product",
+                data_category="S",
+                data_level="DL1",
+                data_association="Subarray",
+                data_model_name="ASWG DL1",
+                data_model_version=DL1_DATA_MODEL_VERSION,
+                data_model_url="",
+                format="hdf5",
+            ),
+            process=meta.Process(type_="Simulation", subtype="", id_=0),
```
kosack, line 402: So perhaps for now, only set it to  "simulation" if the `is_simulation` flag is set to True, otherwise it should be "observation".  

For the `id` field, right now it is not defined in the data model doc if this is a merged data file, so I guess 0 is fine, or we could set it to the PROD id (e.g. "Prod3"), but right now I guess there is no way to determine that from the simulation files.   This will be better defined in the next revision of the data model...

kosack (changes requested):
only one minor change in the metadata, but otherwise I think this is ready and good work!

ctapipe/tools/dl1_merge.py
maxnoe: This goes back to the comments I made in the stage1 tool. Shouldn't we expose parts of these metadata fields via the config system, so that users can set things like the contact, email, and prod from the config file or cli?

maxnoe (approved):

kosack (dismissed):

maxnoe:
The difference between the python and the conda build is astropy 4.0 vs. astropy 4.1.

It seems astropy 4.1 is not writing `table_column_meta`?

maxnoe:
Testing locally, the `image_statistics.__tbale_column_meta__` table is still generated with astropy 4.1

maxnoe:
Testing locally, the tool should give a better error for when no arguments are provided:

```
❯ ctapipe-merge
INFO [MergeTool] (tool/initialize): ctapipe version 0.9.0.post23+git21d22a8
INFO [MergeTool] (tool/run): Starting: ctapipe-merge
ERROR [MergeTool] (tool/run): Caught unexpected exception: 'NoneType' object has no attribute 'expanduser'
  File "/home/maxnoe/Uni/CTA/ctapipe/ctapipe/core/tool.py", line 221, in run
    self.setup()
  File "/home/maxnoe/Uni/CTA/ctapipe/ctapipe/tools/dl1_merge.py", line 172, in setup
    self.output_path = self.output_path.expanduser()
AttributeError: 'NoneType' object has no attribute 'expanduser'
INFO [MergeTool] (tool/run): Output:
```

maxnoe:
I get this here, which might explain why the test is failing?

```
❯ ctapipe-merge -o /tmp/merged_dl1.h5 test_dl1.h5 test_dl1_2.h5 --skip-images --overwrite
INFO [MergeTool] (tool/initialize): ctapipe version 0.9.0.post23+git21d22a8
INFO [MergeTool] (tool/run): Starting: ctapipe-merge
WARNING [MergeTool] (dl1_merge/setup): Overwriting /tmp/merged_dl1.h5
INFO [MergeTool] (dl1_merge/start): Merging simulation-files
WARNING [MergeTool] (dl1_merge/check_file_broken): /simulation/service/shower_distribution is not in test_dl1.h5. Continue with merging file. This table will be incomplete or empty
/home/maxnoe/Uni/CTA/ctapipe/.venv/lib/python3.8/site-packages/tables/group.py:1201: UserWarning: problems loading leaf ``/dl1/service/image_statistics.__table_column_meta__``::

  variable length strings are not supported yet

The leaf will become an ``UnImplemented`` node.
  warnings.warn(
/home/maxnoe/Uni/CTA/ctapipe/.venv/lib/python3.8/site-packages/tables/unimplemented.py:86: UserWarning: UnImplemented node '/dl1/service/image_statistics.__table_column_meta__' does not know how to copy itself; skipping
  warnings.warn(
WARNING [MergeTool] (dl1_merge/check_file_broken): /simulation/service/shower_distribution is not in test_dl1_2.h5. Continue with merging file. This table will be incomplete or empty
INFO [MergeTool] (dl1_merge/start): 2 out of 2 files has been merged!
INFO [MergeTool] (tool/run): Finished: ctapipe-merge
INFO [MergeTool] (tool/run): Output: /tmp/merged_dl1.h5
Closing remaining open files:/tmp/merged_dl1.h5...done
```

maxnoe:
I narrowed down the error to the new version of `h5py` (version 3.0) which is used by astropy to write these tables.

I think it becomes more pressing that we move away from the mixture of using h5py through astropy and pytables.

I see two possibilities:

* require h5py < 3 for now
* Use astropy table io also in the merge to to load / append the corresponding tables.

maxnoe:
So according to this: https://docs.h5py.org/en/stable/strings.html

h5py now defaults to variable length strings, which are not implemented in pytables.

maxnoe (approved):

kosack (approved):

maxnoe:
I checked where the coverage is lacking. It's the `input_dir` case, which could be easily tested and the rest is not really easy to write sensical tests at the moment as these concern

a) observations
b) broken files

maxnoe:
I will merge and maybe @Hckjs kann in a new PR add a test for the input_dir / pattern business.

PR #1451, opened 2020-10-09
dump CameraReadout files in ctapipe-dump-instrument

kosack:
Now that CameraDefinitions need both a camgeom and camreadout file, lets generate them

Needed to generate sample .camreadout.fits.gz files for testing purposes in ctapipe-extra

This PR also removes the unused `table_type, table_name` options from `DumpInstrumentTool._get_file_format_info()`


maxnoe (approved):
Are these hierarchical descriptions not relevant anymore?

kosack:
> Are these hierarchical descriptions not relevant anymore

What do you mean by hierarchical description?

maxnoe:
`dl0.tel.svc.camera`

kosack:
Ah, those are from an old version of the data model, so in any case need to change.  However, I'm not sure its the best thing to use for the "role" of the file in the provenance. So far we have no rule for how to name roles, but I guess it's better to say why the data was used, not what kind of data it is (e.g. that the file was opened to construct a  "CameraGeometry" object).

maxnoe:
@kosack the build failed due to some file not found error

kosack:
Ah, yes, the problem is that we do not have an up-to-date optics description in ctapipe-extra.  I will update it, then this should work (the current one uses the old telescope naming scheme, and is named differently).

vuillaut (approved):
The build is still not passing though.

maxnoe:
@kosack the test fails because you used `--infile` instead of `--input`

maxnoe:
Also, you should add the `cwd=tmpdir`

maxnoe (approved):

maxnoe:
After merging, you produce the readout files for the cameras we have in ctapipe extra and add them to ctapipe extra?

kosack:
> After merging, you produce the readout files for the cameras we have in ctapipe extra and add them to ctapipe extra?

Yes, that's the idea. Then we can release another version of ctapipe-extra (or we add them to the server, and merge #1498 )

maxnoe:
Or both, for the moment it is probably the easiest to add them to ctapipe-extra so johan can just download the newest version.

watsonjj (approved):

PR #1461, opened 2020-10-14
Correct pedestal before extracting the pulse, fixes #1460

### LINKED ISSUE #1460: Peak time extraction is biased if there is a residual pedestal
The `CameraCalibrator` only applies the `DL1CameraCalibrationContainer.pedestal_offset` to the integrated charge.

However, this residual pedestal needs also to be taken into account for the peak time extraction, which is biased (at least using our current weighted mean approach) if there is a residual pedestal.

I talked with @watsonjj and we came to the conclusion would be able to cure this:

* change the definition of the `pedestal_offset` to be the pedestal per sample in the waveform instead of a correction for the integral
* Subtract this pedestal from the waveform before doing the extraction

maxnoe:
I implemented the correction discussed in #1460.

I also added a fix for the `SimTelEventSource` that took the wrong normalization for the pedestal.

It was correct just by chance (pedestal seemed to be calculated for the whole readout window for the prod3b and prod5 file I tested).

I don't know if that is the case for all simulations, but I suspect it is.


ctapipe/calib/camera/calibrator.py
```diff
@@ -109,9 +109,21 @@ def _calibrate_dl0(self, event, telid):
...
+
         n_pixels, n_samples = waveforms.shape
+
+        # subtract any remaining pedestal before extraction
+        if dl1_calib.pedestal_offset is not None:
+            # this copies intentionally, we don't want to modify the dl0 data
+            # waveforms have shape (n_pixel, n_samples), pedestals (n_pixels, )
+            # so we need to transpose to be able to utilize broadcasting
+            # and transpose back after to retain the correct shape
+            waveforms = (waveforms.T - dl1_calib.pedestal_offset).T
```
watsonjj, line 125: Is this procedure recommended over broadcasting with np.newaxis?

watsonjj (dismissed):

ctapipe/calib/camera/calibrator.py
maxnoe: It's not an alternative. It's different use cases, right?

I want to subtract from each row, but for broadcasting to work the last dimension has to be the same, so I have to transpose.

ctapipe/calib/camera/calibrator.py
```diff
@@ -109,9 +109,21 @@ def _calibrate_dl0(self, event, telid):
...
+
```
kosack, line 126: Should we really silently ignore missing pedestals here, or issue a warning or exception if pedestals don't exist? If somebody wants to disable it, they should set the pedestal values to all 0, rather than None, I would guess (None means something failed in loading)

maxnoe: For our default case of simtel array mcs, the pedestal is none, because it is already applied at the r1 stage.

So I don't think we should require setting it to 0 if this correction is not necessary.

ctapipe/calib/camera/calibrator.py
maxnoe: Changed to the better readable solution with `[:, np.newaxis]`

ctapipe/calib/camera/calibrator.py
kosack: but I understood that only the dark pedestal was applied, and not the sky pedestal, or has that changed? If not, it would still be good for SimTelEventSource to write a "0" pedestal to the container, to be clear it is not needed.

maxnoe: I don't think simtel differentiates these two pedestals... there is just one pedestal in the monitoring container.

maxnoe: I would make these corrections optional and require that these are either None or an array of len(pixels).

maxnoe:
> Standard pseudo-random generators are not suitable for security/cryptographic purposes.

Thank you codacy for this extremely valuable contribution!

watsonjj (approved):

ctapipe/calib/camera/calibrator.py
kosack: ok for now.

PR #1462, opened 2020-10-15
Fix missing config=True, add test so this never happens again

maxnoe:


kosack:
Would it be possible to just simply add the `config=True` attribute to all traitlets created in a Component class? It's not enough to loop over all traitlets, since there are many "hidden" ones that `traitlets.config` generates, but it would be nice to not have to add the `.tag()` part all the time...  When I implemented Component initially, I couldn't think of a good way, but perhaps I missed something.

kosack (approved):

watsonjj (approved):

maxnoe:
@kosack,  one could use a metaclass to do that in `__new__`, but I don't like to add more `__magic__` to the config system than it already has...

PR #1464, opened 2020-10-15
remove no-longer-used EnergyRegressor and EventClassifier

kosack:
These classes were only used in a very old version of protopipe, and were deprecated for several years.

These two functions, computing energy and gammaness, are now done externally, using either `protopipe.mva` or `aict-tools`, so are no longer needed here


maxnoe (approved):

maxnoe:
I agree with removing those here. But in the end we do want to have functionality like this here, e.g. by moving code form protopipe to ctapipe or by importing and using aict-tools?

kosack:
> But in the end we do want to have functionality like this here, e.g. by moving code form protopipe to ctapipe or by importing and using aict-tools?

That's another discussion to have ,if we want ctapipe to be only event-wise code, and others separate, or we import other thigns and unify it all in ctapipe.  But let's perhaps leave that to a more general architectural discussion.   For now, this code is really not used, and isn't even worth replacing as is. If we add similar functionality, it would just be at the Tool level calling aict-tools or something.

maxnoe:
Yes, fully agree!

HealthyPear (approved):

HealthyPear:
While refactoring, I just realized that part of these classes were still used in protopipe (not the mva module, but the current TRAINING and DL2 tools).

specifically the `model_dict` attribute (basically a deepcopy of the input model by camera) but since it was a class attribute the class was needed to instantiate an object (regressor and classifier respectively)

I will harvest from here what is needed for now and perhaps during the refactoring that I am performing I (we) can think about how to standardize the reading of model files per camera type.

HealthyPear:
ah sorry also the `load` class methods...

PR #1465, opened 2020-10-16
Apply time calib simtel

maxnoe:
This is a follow up to #1463 

Here, I get the required time shift from the simtel laser calibration object and fill it appropriately into the dl1 calibration container.

I simulated flat field events using simtelarray prod5 la palma baseline, on 100 flat field events (times 19 telescopes) the standard deviation of the pulse time goes from 0.56 to 0.15 ns, using this simple script:

```python
from ctapipe.io import SimTelEventSource
from ctapipe.calib import CameraCalibrator
from ctapipe.image.extractor import LocalPeakWindowSum
from collections import defaultdict
import numpy as np
from ctapipe.containers import EventType


path = '../simtel_prod5/Data/sim_telarray/cta-prod5-lapalma/0.0deg/Data/gamma_20deg_180deg_run4___cta-prod5-lapalma_desert-2158m-LaPalma-dark.simtel.zst'


source = SimTelEventSource(path, skip_calibration_events=False)
calib = CameraCalibrator(source.subarray, image_extractor=LocalPeakWindowSum(source.subarray))


times = []

for e in source:
    if e.trigger.event_type != EventType.FLATFIELD:
        continue

    calib(e)
    for tel_id, dl1 in e.dl1.tel.items():
        times.append(dl1.peak_time)

times = np.array(times)

print(np.std(times - np.mean(times, axis=1)[:, np.newaxis]))
```

On the master: 
```
0.5593064
```

Here:
```
0.15312754
```


kosack (approved):

maxnoe:
should this be optional?

kosack:
> should this be optional? Or rather: should maybe the hole time shift business in the CameraCalibrator be optional?

I guess it's good to check if we should be applying this to MCs or not: are they really simulated with a distribution of times per pixel or module? or is this just a measure of the uncertainty due to discrete sampling? (in the second case, I think you would not want to subtract it, since it's statistical not systematic)

maxnoe:
@kosack No this is not the question. We checked the docs of simtel array, see discussion here #1463.

It is correct and should always be correct. Might be nice to be able to disable this correction (in general, not just for the simtel event source) though.

ctapipe/io/simteleventsource.py
```diff
@@ -414,6 +414,13 @@ def _generate_events(self):
                 if r0.num_trig_pix > 0:
                     r0.trig_pix_id = pixel_lists[0]["pixel_list"]
 
+                # get time_shift from laser calibration
+                time_calib = array_event["laser_calibrations"][tel_id]["tm_calib"]
+                pix_index = np.arange(n_pixels)
+
+                dl1_calib = data.calibration.tel[tel_id].dl1
+                dl1_calib.time_shift = time_calib[r1.selected_gain_channel, pix_index]
```
watsonjj, line 422: why is pix_index needed?

maxnoe: Same reason as in the R1 Calibrator. we have a (n_gain, n_pixels) array of time shifts and need to select for each pixel the correct time shift for the selected gain channel.

If you have a better solution, I'll be glad to adapt it also in the other code (which I think you wrote?)

watsonjj: Ah right, yeh. No I don't have a better solution

watsonjj:
I support an option (via trait) to turn off the shifting entirely

watsonjj:
I'm hesitant to approve this without some benchmarking to demonstrate its impact. I am also curious why it hasn't previously been used in other analysis chains...

maxnoe:
The simple check is that the std of the peak times got a lot smaller (from 500 to 150 ps). 

I think applying this correction is the correct thing to do, it is certainly intended this way by sim_telarray.
The question as always then remains how realistic the current values in the simulation are for the telescopes, but this we can only find out by comparing to observed data.

@GernotMaier Do you apply this correction in EventDisplay?

kosack (dismissed):

maxnoe:
@watsonjj the corrections can now be disabled using the boolean options introduced in #1492

kosack (approved):

watsonjj (approved):

Issue #1466, opened 2020-10-16
Follow NEP 29 for python / numpy versions


maxnoe:
The schedule by numpy seems very reasonable:
https://numpy.org/neps/nep-0029-deprecation_policy.html

I'd suggest to follow the same schedule with ctapipe, which would mean we could now be 3.7+ and drop 3.6 from the tests.

And we should probably start testing vs. 3.9 since it is out now.

maxnoe:
Currently, we would be thus

3.7+, numpy 1.17+

maxnoe:
Adapted travis config here #1467

PR #1467, opened 2020-10-16
Require python >= 3.7

maxnoe:
This adresses #1466 , so dropping 3.6 for the next release.

There was also a bug in `ci/install.sh` resulting in not installing the correct python version for conda builds (so we didn't even test on 3.6 since conda switched to 3.7 by default)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -53,6 +53,7 @@ matrix:
           packages:
             - pandoc
             - graphviz
+            - libhdf5-dev # remove when pytables wheel is available for 3.9
 
 before_install:
   - git fetch --tags
--- a/.travis.yml
+++ b/.travis.yml
@@ -47,7 +47,7 @@ matrix:
     - os: linux
       dist: bionic
       language: python
-      python: 3.9
+      python: 3.9-dev
       addons:
         apt:
           packages:
--- a/.travis.yml
+++ b/.travis.yml
@@ -44,16 +44,15 @@ matrix:
             - pandoc
             - graphviz
 
-    - os: linux
-      dist: bionic
-      language: python
-      python: 3.9-dev
-      addons:
-        apt:
-          packages:
-            - pandoc
-            - graphviz
-            - libhdf5-dev # remove when pytables wheel is available for 3.9
+#    - os: linux
+#      dist: bionic
+#      language: python
+#      python: 3.9
+#      addons:
+#        apt:
+#          packages:
+#            - pandoc
+#            - graphviz
 
 before_install:
   - git fetch --tags
--- a/.travis.yml
+++ b/.travis.yml
@@ -8,14 +8,6 @@ env:
 
 matrix:
   include:
-    - name: conda python 3.6
-      os: linux
-      dist: bionic
-      language: generic
-      env:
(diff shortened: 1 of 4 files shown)


kosack (dismissed):

dneise:
There is a timeout in `conda env create -n cta-dev --file environment.yml` on Travis when building for Py3.8 

Is this reason for concern? I restarted the build this morning once .. hoping it was just a mishap on travis.... no luck

ci/install.sh
```diff
@@ -23,7 +23,7 @@ if [[ "$CONDA" == "true" ]]; then
     # Useful for debugging any issues with conda
     conda info -a
 
-	sed -i -e "s/- python=.*/- python=$TRAVIS_PYTHON_VERSION/g" environment.yml
+	sed -i -e "s/- python=.*/- python=$PYTHON_VERSION/g" environment.yml
```
dneise, line 26: I am confused by this change ... looks independent of the Py3.6 -> Py 3.7 step .. was this a bug before?

maxnoe: Yes it was. It was revealed in #1472 because it should have error on 3.6 but didn't since it really was using 3.7

maxnoe:
@dneise it worked yesterday with the first commit... Ah, but that probably was still using the wrong version.

maxnoe:
It is a missing eventio for python 3.8 I think.

dneise:
Erm .. I see to be missing the point here ... it is still failing no?

maxnoe:
Yes, because there is no conda package of eventio for python 3.8. @kosack will upload it shortly.

dneise:
Ah I see ... thanks

.travis.yml
```diff
@@ -54,6 +44,16 @@ matrix:
...
+#    - os: linux
+#      dist: bionic
+#      language: python
+#      python: 3.9
+#      addons:
+#        apt:
+#          packages:
+#            - pandoc
+#            - graphviz
+
```
vuillaut, line 56: why leaving this commented?

maxnoe: I'll remove it. python 3.9 does not work yet

ci/install.sh
```diff
@@ -23,9 +23,9 @@ if [[ "$CONDA" == "true" ]]; then
     # Useful for debugging any issues with conda
     conda info -a
 
-	sed -i -e "s/- python=.*/- python=$TRAVIS_PYTHON_VERSION/g" environment.yml
-	travis_wait 20 conda env create -n cta-dev --file environment.yml
-	conda activate cta-dev
+	sed -i -e "s/- python=.*/- python=$PYTHON_VERSION/g" environment.yml
+	travis_wait 20 conda env create -n travis cta-dev --file environment.yml
```
vuillaut, line 27: remove `cta-dev`

vuillaut (changes requested):

ci/install.sh
maxnoe: how did it work???

vuillaut (approved):

dneise (approved):

PR #1471, opened 2020-10-19
Fix bug in h5_table_to_astropy and HDF5TableWriter

kosack:
## Expected behavior:
```
table = h5_table_to_astropy(dl1_file, path_to_dataset)
table.write("example.ecsv")
```

This currently fails due to several reasons:
1. the _DESC and other metadata were stored as numpy strings, not python strings, causing astropy's io mechanism to fail with unknown type
2. the _DESC metadata were erroneously written for skipped columns in HDF5TableWriter 
3. the _DESC data should be used as a table.column[x].description attribute

## Fix:
- don't write _DESC for skipped columns
- convert all attribute metadata  to strings `str(value)` (or should we check for only things that were strings?) 
- use the .description attribute of Column class in astropy

The output now looks like
image


maxnoe:
> convert all attribute metadata to strings str(value) (or should we check for only things that were strings?)

Yes, please only convert `np.str_` to string. Since ints and floats and all other dtypes are usable here.

ctapipe/io/astropy_helpers.py
```diff
@@ -62,7 +62,7 @@ def h5_table_to_astropy(h5file, path) -> QTable:
             # need to convert to str() here so they are python strings, not
             # numpy strings
             value = table.attrs[attr]
-            other_attrs[attr] = str(value) if isinstance(value, np.str) else value
+            other_attrs[attr] = str(value) if isinstance(value, np.str_) else value
```
maxnoe, line 65:
```
>>> issubclass(np.str_, np.str)
True
```
So using `np.str` is probably more general.

maxnoe (dismissed):

ctapipe/io/astropy_helpers.py
kosack: yes, I had that originally (see the commit message) and it worked, but what is the point of converting a str → str if it is not a numpy str_?  Doesn't hurt, but doesn't do anything either.

kosack: unless we expect other string types, perhaps

watsonjj (dismissed):

ctapipe/io/astropy_helpers.py
maxnoe: Ah, sorry, didn't realize that `np.str` is the python string. Sorry. That's confusing.

kosack:
sorry @watsonjj and @maxnoe , please re-review this.  The tests now work again (they broke due to a typo being accidentally committed)

maxnoe (approved):

kosack:
this is ready, other than perhaps changing the name to `ctapipe.io.read_table` or something simpler than `ctapipe.io.astropy_helpers.h5_table_to_astropy`

maxnoe (approved):

vuillaut (approved):

maxnoe (approved):

PR #1475, opened 2020-10-19
Run tools in tmpdir during tests, do not write provlog if only help was run

maxnoe:


kosack (approved):

maxnoe:
This still creates the log files in the repository. wtf. I need to check again.

maxnoe:
Ok, the stackoverflow answer I found was referring to `py.path.Path` not `pathlib.Path`. The context manager of `Path` does not actually change cwd.

kosack (approved):

ctapipe/core/tool.py
```diff
@@ -238,17 +240,22 @@ def run(self, argv=None):
...
         finally:
-            for activity in Provenance().finished_activities:
-                output_str = " ".join([x["url"] for x in activity.output])
-                self.log.info("Output: %s", output_str)
-
-            self.log.debug("PROVENANCE: '%s'", Provenance().as_json(indent=3))
-            self.provenance_log.parent.mkdir(parents=True, exist_ok=True)
-            with open(self.provenance_log, mode="a+") as provlog:
-                provlog.write(Provenance().as_json(indent=3))
+            print(self.argv)
```
watsonjj, line 243: print statement

ctapipe/core/tool.py
```diff
@@ -238,17 +240,22 @@ def run(self, argv=None):
...
+            if not {"-h", "--help", "--help-all"}.intersection(self.argv):
```
watsonjj, line 244: I don't see a test to make sure the provlog is not written if only help is ran

maxnoe: Good point.

maxnoe: Added the test

ctapipe/core/tool.py
maxnoe: removed

Bultako:
What do you think of extending this feature to a new flag? (not only `--help` `--help-all` or `-h`) 

The provenance log file would be always produced by default, but this behavior could  be deactivated with a flag.

maxnoe:
You deactivating writing it? I would not really want to support that, this is important info.

However, you can do what you can always do if you must: `--provenance-log=/dev/null`

Bultako:
@maxnoe  
I agree it is important info, though as it is now it could be seen as a kind-of imposed non-optional file logging. I can imagine some performance i/o issues in cases where tools could be heavily used (i.e. in a big loop) Anyhow, the `--provenance-log=/dev/null` tip you propose is perfectly fine. Thanks !

maxnoe:
>  can imagine some performance i/o issues in cases where tools could be heavily used (i.e. in a big loop)

I cannot. Writing the provenance log should always be negligible compared to the actual work a tool does (not mentioning the time imports take, etc...). 

Did you actually run into a case where you *measured* that writing the provenance log is a performance bottleneck?

Bultako:
> 
> Did you actually run into a case where you _measured_ that writing the provenance log is a performance bottleneck?

I am actually investigating what it could be a similar case. In LSTOSA, it seems a standard python `log.info()` (where the stream goes into a log file) spends 38% of _measured_ time of a step in which around of ~100 subruns are processed. The log file size could be as big as tens of MB.

maxnoe:
That has nothing to do with provenance log. The problem with info logging are continous small writes (that are probably even flushed at each log call to have them in case of a crash).

The provenance data is written at the end in one chunk. Please don't mix things like these.

kosack:
I think the confusion is just that we call it a "log" - the provenance info so far does not use the logging system, and is written all at once from an in-memory dict, so there should be no performance issues with it. It's only called a "log" since we append to it, and this keep previous runs from being overwritten - so conceptually it is a log of provenance.  In the future, we could figure out better ways to attach this to output files of a tool, but for now that was the simplest.   We could for example even write it into the output HDF file, but then if multiple files are written, it becomes more complicated. 

The ctapipe log (output from the logger) is another thing altogether, and currently defaults to stdout, so no file is created unless the user asks for it.

kosack:
> I am actually investigating what it could be a similar case. In LSTOSA, it seems a standard python log.info() (where the stream goes into a log file) spends 38% of measured time of a step in which around of ~100 subruns are processed. The log file size could be as big as tens of MB.

I guess there you are writing info messages per event or something? That is not a good idea. Log info is intended to be human-readable. If you need per-event info, it should go into the binary output, or if for real logging purposes, should only output every 10000 events or something, with a status/summary.

morcuended:
> I guess there you are writing info messages per event or something? That is not a good idea. Log info is intended to be human-readable. If you need per-event info, it should go into the binary output, or if for real logging purposes, should only output every 10000 events or something, with a status/summary.

Right now we capture the provenance info per sub-run (53000 events).

kosack (approved):

watsonjj (dismissed):

maxnoe:
@kosack @watsonjj sorry, had to fix a conflict, please review again

kosack (approved):

watsonjj (approved):

kosack (approved):

PR #1477, opened 2020-10-20
(Re)move trigger related fields from R0/R1/DL0 Container

maxnoe:


kosack (approved):

thomasarmstrong (commented):
Might we want to keep trigger type for future use? i.e. if specific trigger flags are stored (muon etc), or would this be moved somewhere else?

maxnoe:
It is stored in `event.trigger`, these fields were simply forgotten when we made that transition.

PR #1479, opened 2020-10-20
Do not write tels with trigger in pointing table

maxnoe:
This column is not needed (even unwanted there) and basically makes merging impossible.

### DIFF
--- a/ctapipe/tools/stage1.py
+++ b/ctapipe/tools/stage1.py
@@ -483,7 +483,7 @@ class Stage1ProcessorTool(Tool):
     def _write_telescope_event(self, writer, event):
         """
         add entries to the event/telescope tables for each telescope in a single
-        event
+        even
         """
         # write the telescope tables
         for tel_id, dl1_camera in event.dl1.tel.items():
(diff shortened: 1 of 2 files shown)


kosack:
why was it in there to begin with? Just to write the mean trigger time?   Perhaps we need a better place to keep track of subarray event time.  It's not even clear we have a good definition of subarray time : i.e. is it the the mean telescope time?

kosack (dismissed):

maxnoe:
Yes, to write the time. Which is a hack for simtel array files since these do not have monitoring data with timestamps, so we use the event time (which makes sense in trigger).

maxnoe:
This probably needs further fixing as soon as I can use the LST event source with the dl1 tool and inspect the output.

kosack (approved):

ctapipe/tools/stage1.py
```diff
@@ -483,7 +483,7 @@ def _process_events(self, writer):
     def _write_telescope_event(self, writer, event):
         """
         add entries to the event/telescope tables for each telescope in a single
-        event
+        even
```
vuillaut, line 486: event was right, no?

ctapipe/tools/stage1.py
```diff
@@ -623,6 +624,10 @@ def _setup_writer(self, writer: HDF5TableWriter):
                 writer.exclude(
                     f"/dl1/event/telescope/images/{table_name}", "image_mask"
                 )
+
+            writer.exclude(
+                f"/dl1/monitoring/telescope/trigger/{table_name}", "trigger_pixels"
+            )
```
vuillaut, line 630: isn't this important for real data?

maxnoe: At the DL1 level you want to know which trigger patches contributed to the trigger decision?

maxnoe: Then we would need to convert it to a mask. Currently the list is not writable and this exclude is only to silence the warnings.

vuillaut (approved):

Issue #1482, opened 2020-10-21
HDF5TableReader column transforms failing for enums


LukasNickel:
This occured after merging the current master into #1357.
When I now try to iterate events generated by the DL1EventSource, I get an error stating:
>TypeError("'numpy.str_' object is not callable")

This happens when calling this function:
 https://github.com/cta-observatory/ctapipe/blob/225d5eed16e1b3733252941f0849cf246652f4c4/ctapipe/io/hdf5tableio.py#L391

for the `/dl1/event/subarray/trigger` table.
When the transform function is assigned to the column, the attr / colname are matching the expected (`event_type_ENUM`).
When the function is called, attr is now `time_DESC`, which does not point to an enum, but a string in the attributes. This is the last value from the loop. It seems reasonable to assume, that the value of `attr` is not fixed inside the function, but why did this work before?

As far as I can tell, that code has not been touched in a while and my python version is also fixed, so I really wonder why this is failing now all of a sudden.
I will try to generate a MWE tomorrow to investigate.

kosack:
the code to write column descriptions as `*_DESC` attributes was added recently - could there be an unintended interaction? The enum code should look only for the `*_ENUM` attributes, though

maxnoe:
The full traceback would be helpful

maxnoe:
Shouldn't the DL1EventSource only call it the other way around? `enum_to_int` is the the transform for writing, it should be `int_to_enum` for reading back?

LukasNickel:
> the code to write column descriptions as `*_DESC` attributes was added recently - could there be an unintended interaction? The enum code should look only for the `*_ENUM` attributes, though

The loop goes over all  attributes though. If the ENUM entry was the last one before this change, then that would explain this.

kosack:
> Do we have a notion of inverse transforms for the reader?

Yes

LukasNickel:
```
    def test_max_events(dl1_file):
        max_events = 5
        with DL1EventSource(input_url=dl1_file, max_events=max_events) as source:
            count = 0
            assert source.max_events == max_events  # stop iterating after max_events
            assert len(source) == 20  # total events in file
>           for _ in source:

ctapipe/io/tests/test_dl1eventsource.py:51:
[...]

PR #1483, opened 2020-10-21
Fix enum deserialization in HDF table writer

maxnoe:
Worked before only if there is only one field in the container due to a reference issue


maxnoe:
The test in commit 1 fails on master and commit 2 contains the fix.

LukasNickel:
For context: This fixes #1482. For now we still rely on pickle until we decide whether we want to change that.

LukasNickel (approved):

maxnoe:
The issue is linked, @LukasNickel ;)

LukasNickel:
> The issue is linked, @LukasNickel ;)

I'm blind, mea culpa :)

kosack (approved):

PR #1486, opened 2020-10-23
Use eventio from conda-forge

maxnoe:


kosack (commented):
Is this really necessary? It shouldn't matter where eventio comes from, and since we already have conda-forge as a channel, it should get it from there automatically.  I can also remove it from the cta-observatory channel

maxnoe:
Ah, sorry, I didn't realize that conda-forge is already in the channels list.

Should not matter then, yes!

maxnoe:
@kosack 

conda forge is not in the channels! So yes, this is necessary. Either only for eventio or adding it globally to the channels.

kosack:
> conda forge is not in the channels

I see that at some point it was removed, but some packages like nbsphinx are now marked explicitly.  Ok, then this is ok. 
We could consider just adding the channel, but that does increase the solver time due to how many packages are on conda-forge.

kosack (approved):

HealthyPear (approved):

PR #1487, opened 2020-10-26
Do not calculate errors on time fit parameters, fixes #1420

### LINKED ISSUE #1420: Do not calculate errors in timing parameters
These values are calculated by making another, normal linear regression.

We should remove this since:

* It takes additional time
* The errors are not the ones for the actual fit
* The deviation should be sufficient.
* It even uses a fits weighted with "image" so the errors have even less interpretability for the actual result

maxnoe:
The errors were calculated using a different fit method, which makes them invalid for the actual results.

Since they don't seem to be used anywhere, just removing them is the best option I think.

### DIFF
--- a/ctapipe/containers.py
+++ b/ctapipe/containers.py
@@ -171,9 +171,7 @@ class TimingParametersContainer(Container):
     slope = Field(
         nan / u.m, "Slope of arrival times along main shower axis", unit=1 / u.m
     )
-    slope_err = Field(nan / u.m, "Uncertainty `slope`", unit=1 / u.m)
     intercept = Field(nan, "intercept of arrival times along main shower axis")
-    intercept_err = Field(nan, "Uncertainty `intercept`")
     deviation = Field(
         nan,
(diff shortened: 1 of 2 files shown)


LukasNickel (approved):

vuillaut:
If they were to be used, that would be at a higher level than ctapipe right now (protopipe, lstchain...), no?

maxnoe:
Yes. But they are wrong at the moment, require running a second fitter (which impacts performance quite badly).

vuillaut:
> Yes. But they are wrong at the moment, require running a second fitter (which impacts performance quite badly).

I agree. I was just pointing out that "not used anywhere here" does not mean useless.
For now better remove them than having wrong parameters.

watsonjj (approved):

watsonjj:
This is technically an API change

maxnoe:
Not only technical. But we have already merged some, next will be 0.10

PR #1489, opened 2020-10-26
Remove tels_with_data, fixes #852

### LINKED ISSUE #852: get rid of tels_with_data, and rely only on the Map keys
The reason is that if you have multiple telescope algorithms being run in sequence (like cleaning), one may
drop a tel and not another, so you want to keep track of which telescopes survive at each data level/algorithm set, and not have to update it in 2 places.

This may require some re-thinking of how Containers are  organized.  

Would also be nice to have a standard function for turning a telescope list into a hit-pattern (boolean array or similar) for storage. 

### Use Case:
Imagine a pipeline chain that applies two sets of image cleaning (e.g. one for reconstruction and one for gamma-hadron selection).  In the reconstruction cleaning, 8 telescopes survive and are passed to the Reconstuctor, but in the looser cleaning for discrimination, 15 telescopes survive that are passed to the GammaHadronSelector.  

Having a single top-level tels_with_data (at levels above Dl0, where it makes sense) causes problems for this method.

maxnoe:


vuillaut (approved):

maxnoe:
By the way, these reconstructor tests have much duplicated code which did some really strange ad-hoc trace integration.

watsonjj (approved):

watsonjj:
Actually before merging, @kosack previously said that the tels_with_data is relied on for file writing. Is this still the case?
https://github.com/cta-observatory/ctapipe/issues/722#issuecomment-384017521

maxnoe:
The tels_with_trigger is written in the stage1 tool in the subarray/trigger group.

maxnoe:
The other levels are not really needed I think, you get that for free again by looking at the telescope tables.

The question is maybe if an index table would be helpful for the DL1 event source. E.g. a image_index table and a parameters index table that lists which telescopes have images / parameters.

But I think this is a different issue.

Issue #1330, opened 2020-05-15
Update to bokeh 2


maxnoe:
We are currently fixed on Bokeh v1, but v2 is out.

maxnoe:
This was done in #1494

Issue #1247, opened 2020-04-01
Improve Bokeh Camera Display


kosack:
There is a Bokeh version of the CameraDisplay in `ctapipe.visualization.bokeh.CameraDisplay` now that is much faster than the default MatPlotLib one, and is better suited for interactive use (the matplotlib version is better for making nice PDFs, etc).  However, it's currently not used very often since it's not friendly to use in a notebook, and doesn't have as nice a display as the default one. 

Here I have created an example notebook that demonstrates that we can do it as nicely as the MatPlotLib version, and have it work in a Jupyter Notebook (also in Jupyter Lab).

https://gist.github.com/kosack/aceb8b6a8c6ab616767759ca48609063

Download this notebook and run it to try it out (it doesn't work via GitHub's notebook renderer)

The resulting display looks like this,  handles both hexagonal and square pixels, and updates at > 30 frames per second on my laptop, making it really nice for interactive functions using `ipywidgets.interact()`:

![image](https://user-images.githubusercontent.com/11677812/78160187-de558980-7443-11ea-8c31-06daabe10558.png)

Zooming works properly and in real-time:
![image](https://user-images.githubusercontent.com/11677812/78160721-a26ef400-7444-11ea-95fe-9a3fbd5d98a4.png)

Right now the code is just a proof-of-concept snippet, but I think we could use it to improve the current implementation and automatically push updates to a Notebook for realtime use.  That might break the functionality of the current non-notebook version, which so far is only used in `ctapipe-event-viewer`, which could be modified to be a fully in-notebook display, not a separate application. Or we can make it work in both cases. 

A further development would be to make ctapipe.visualization.CameraDisplay be just an  dummy class that instantiates either an `mpl.CameraDisplay` or `Bokeh.CameraDisplay` upon request (either by delegation, or subclassing).

MaxNoe:
At this point we could then also think about making these two libraries optional dependencies (matplotlib and bokeh). At least one of them should be optional as both are pretty hefty (but matplotlib is more widely used at the moment.)

kosack:
incomplete, but perhaps more useful example

```py3
import numpy as np
import bokeh
from bokeh.io import output_notebook, push_notebook, show
from bokeh.plotting import figure
from ctapipe.instrument import CameraGeometry


def _generate_polygon_vertices(order=6, x=0, y=0, size=1.0, theta0=0.0):
    """
    returns the vertices of a regular polygon of order `order`,
[...]

PR #1491, opened 2020-10-26
Use enum for pixel shapes

maxnoe:
When preparing to update the bokeh event viewer according to #1330 and #1247, 
I found working with these strings very clumsy.

Replaced the string with an enum, which should make many things much more straight forward.

### DIFF
--- a/ctapipe/instrument/camera/geometry.py
+++ b/ctapipe/instrument/camera/geometry.py
@@ -18,11 +18,43 @@ from scipy.spatial import cKDTree as KDTree
 from ctapipe.coordinates import CameraFrame
 from ctapipe.utils import get_table_dataset
 from ctapipe.utils.linalg import rotation_matrix_2d
+from enum import Enum, unique
 
 __all__ = ["CameraGeometry", "UnknownPixelShapeWarning"]
 
 logger = logging.getLogger(__name__)
 CG = TypeVar("CG", bound="CameraGeometry")  # for forward-referencing type hints
+
+
+@unique
+class PixelShape(Enum):
+    CIRCLE = "circle"
+    SQUARE = "square"
+    HEXAGON = "hexagon"
+
+    @classmethod
+    def from_string(cls, name):
+        name = name.lower()
+
+        if name.startswith("hex"):
+            return cls.HEXAGON
+
+        if name.startswith("rect") or name == "square":
+            return cls.SQUARE
+
+        if name.startswith("circ"):
+            return cls.CIRCLE
+
+        raise TypeError(f"Unknown pixel shape {name}")
+
+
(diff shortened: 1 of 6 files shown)


kosack:
Is this really more straightforward? It is right now text metadata, so very human readable, and this changes the data model. I don't see a big advantage, other than avoiding a single string compare (we don't support different pixel types being mixed in a camera)

kosack:
I also see no reason to support "circle" shapes - we always have Winston cones, so the actual area is a hexagon.

ctapipe/instrument/camera/geometry.py
```diff
@@ -371,7 +419,7 @@ def to_table(self):
             [self.pix_id, self.pix_x, self.pix_y, self.pix_area],
             names=["pix_id", "pix_x", "pix_y", "pix_area"],
             meta=dict(
-                PIX_TYPE=self.pix_type,
+                PIX_TYPE=self.pix_type.value,
```
kosack, line 422: again, this is not very readable the user has to know what pixel type this integer corresponds to - why not just use the string representation and then it doesn't break the data model?

kosack:
I think my main issue here is that I want this instrument code to be used outside of ctapipe and it's I/O libraries - already I'm trying to convince the simulation configuration people to use it.  If has too many magic numbers defined in it, the data model becomes too closely tied to ctapipe or simtel.

maxnoe:
@kosack I think you missed that the enum has string values compatible to the string values we had before, this breaks nothing, this only changes in ctapipe internally.

Observe that:
* the tests are not changed
* the data in ctapipe-extra is not changed
* Everything still works

> I also see no reason to support "circle" shapes - we always have Winston cones, so the actual area is a hexagon.

Simtel supports them and some of the cameras defined in ctapipe extra reallly have circular pixels on a hexgrid: https://www.mpi-hd.mpg.de/hfm/CT/tel3.gif

ctapipe/instrument/camera/geometry.py
maxnoe: This is an `Enum` with str values compatible to the ones we had before. Not an `IntEnum`.

maxnoe:
This will prevent typos / enable auto completion. This is a major benefit when developing.

ctapipe/instrument/camera/geometry.py
kosack: ah, I thought they were integer values. Sorry, it does what I suggested then.

kosack:
Ok, sorry I missed that the enum had string values. It's ok then, since it doesn't change the data model.

kosack (approved):

watsonjj (approved):

Issue #1341, opened 2020-05-19
Mean pe time for simulated events


MaxNoe:
simtel array also delivers the individual photon arrival times in each pixel.

I calculated mean arrival time in each pixel from this. The results are looking good.

However as these are lists of lists, calculating the mean in each pixel for each camera is kinda slow (1ms per telescope I found).

I see two possible solutions, calculating the mean time already in the cython part of eventio (would be easy, but right now, eventio does not really do things like this) or using something more suited to variable length arrays, e.g.: https://github.com/scikit-hep/awkward-array

Which might be a good idea anyways (maybe directly in eventio) as many things in eventio are variable length

MaxNoe:
![true_image](https://user-images.githubusercontent.com/5488440/82326140-ab0f8f80-99dc-11ea-86d4-d203c11bd3d8.png)

kosack:
how about numba? We use that already, and it should be relatively fast for list looping

MaxNoe:
```
                    mc.mean_pe_time = np.array([
                        np.mean(t) if len(t) > 0 else np.nan
                        for t in pe["time"]
                    ]).astype('float32')
```

kosack:
I think if you make that a function with @jit, it would work fine - numba unrolls np.mean and I think supports list comprehensions, so not much change needed.

watsonjj:
> simtel array also delivers the individual photon arrival times in each pixel

Is this always delivered, or is it a special option that needs to be activated when running simtelarray and has additional overhead?

MaxNoe:
> Is this always delivered, or is it a special option that needs to be activated when running simtelarray and has additional overhead?

When the true image is there, also the times are there.

But yes, this is a special option that needs to be activated.

MaxNoe:
I tried with numba, but it chokes on the empty lists. awkward is slower, as it also as to first create the list of lists to an awkwardarray and that's basically uses the same loop.

MaxNoe:
I added calculation of the mean time in pyeventio here:
https://github.com/cta-observatory/pyeventio/pull/204/commits/864f193b1a413d300e2a191e58c5a13381815f93

MaxNoe:
eventio 1.2 is now out with mean_time in the output for this.

watsonjj:
[...]

PR #1495, opened 2020-10-27
Refactor simulation event info

### LINKED ISSUE #1493: where to store MC DL1 information in event structure
I'm starting to refactor stage1.py such that the writing is separate from the calculations (creating a DL1Writer Component), and I've noticed that so far we have nearly everything needed in the event structure *except* the `MCDL1CameraContainer`, which is created on-the-fly inside the script, but is never written to the "event".  Now that I'm separating the computation from the writing, everything written must be in the event, so the question is: where would be a good place to put this info?

## Current status:

We have 2 containers for DL1 camera info. 

### Real DL1 images and parameters
**In Event Structure**:  `event.dl1.tel[X]`
**Written to**: */dl1/event/telescope/{images,parameters}/tel_X* in the output
```
ctapipe.containers.DL1CameraContainer:
                         image: Numpy array of camera image, after waveform
                                extraction.Shape: (n_pixel) as a 1-D array with
                                type float32
                     peak_time: Numpy array containing position of the peak of
                                the pulse as determined by the extractor. Shape:
                                (n_pixel) as a 1-D array with type float32
                    image_mask: Boolean numpy array where True means the pixel
                                has passed cleaning. Shape: (n_pixel) as a 1-D
                                array with type bool
                  parameters.*: Parameters derived from images
```

### Simulated DL1 images and parameters
For simulated info we have a similar class:
**In Event Structure**: nowhere
**Written to**: */dl1/simulation/telescope/{images,parameters}/tel_X* in the output
```
ctapipe.containers.MCDL1CameraContainer:
             true_image: Numpy array of camera image in PE as simulated
                                before noise has been added. Shape: (n_pixel) as
                                a 1-D array with type float32
             true_parameters.*: Parameters derived from the true_image
```
 
### Other Simulated info
**In Event Structure**: `event.mc.tel[x]`:
**Written to**: nowhere in the output
```
ctapipe.containers.MCCameraEventContainer:
                    true_image: Numpy array of camera image in PE as simulated
                                before noise has been added. Shape: (n_pixel)
                      dc_to_pe: DC/PE calibration arrays from MC file
                      pedestal: pedestal calibration arrays from MC file
                   azimuth_raw: Raw azimuth angle [radians from N->E] for the
                                telescope
                  altitude_raw: Raw altitude angle [radians] for the telescope
                   azimuth_cor: the tracking Azimuth corrected for pointing
                                errors for the telescope
                  altitude_cor: the tracking Altitude corrected for pointing
                                errors for the telescope
```

## Options

1. get rid of `MCCameraEventContainer` and replace it with `MCDL1CameraContainer`: are any of the other fields still used? I think the pointing already goes in `event.pointing`, and `dc_to_pe` and `pedestal` are not really per-event quantities and are already in `event.calibration.tel[x].dl1` (`DL1CameraCalibrationContainer`). 

2. put MCDL1CameraContainer somewhere in `event.dl1.tel[x]` (a bit ugly, since this is only there for simulations)

Any others?

kosack:
Fixes #1493 

Major changes:
- removed `EventAndMonDataContainer` → merged with DataContainer  (and renamed to `ArrayEventContainer`)
- replaced `MCEventContainer` and `MCCameraEventContainer` with new classes: 
  + `SimulatedEventContainer`
  + `SimulatedShowerContainer`
  + `SimulatedCameraContainer`
- Removed `event.mc.*`, replaced with `event.simulation.*`
- `event.mc.tel[x].dc_to_pe` and `pedestals` from simulations are no longer stored in event...  should discuss if these should return and where (they are still applied however).  I would suggest if needed, these are really R1 calibration info and should go in the same place as r1 calibration info for real data. 
- removed `event.mcheader` (still exists as `source.simulation_config`), and renamed the container to `SimulationConfigContainer`
- use `event.pointing` everywhere, instead of `event.mcheader.run_array_direction[i]` (which was removed)

Todo:
 - [x] fix notebooks
 - [x] fix docs

### Example
```
In [1]: event.simulation
Out[1]:
ctapipe.containers.SimulatedEventContainer:
                      shower.*: True event information
                        tel[*]:

In [2]: event.simulation.shower
Out[2]:
ctapipe.containers.SimulatedShowerContainer:
                        energy: Monte-Carlo Energy [TeV]
                           alt: Monte-carlo altitude [deg]
                            az: Monte-Carlo azimuth [deg]
                        core_x: MC core position [m]
                        core_y: MC core position [m]
                   h_first_int: Height of first interaction [m]
                         x_max: MC Xmax value [g / cm2]
             shower_primary_id: MC shower primary ID 0 (gamma), 1(e-),2(mu-),
                                100*A+Z for nucleons and nuclei,negative for
                                antimatter.

In [3]: event.simulation.tel[1]
Out[3]:
ctapipe.containers.SimulatedCameraContainer:
                    true_image: Numpy array of camera image in PE as simulated
                                before noise has been added. Shape: (n_pixel) as
                                a 1-D array with type float32
             true_parameters.*: Parameters derived from the true_image
```


maxnoe:
Can we then also rename `DataContainer` → `ArrayEventContainer`?

ctapipe/containers.py
```diff
@@ -432,7 +407,30 @@ class MCEventContainer(Container):
         "2(mu-), 100*A+Z for nucleons and nuclei,"
         "negative for antimatter.",
     )
-    tel = Field(Map(MCCameraEventContainer), "map of tel_id to MCCameraEventContainer")
+
+
+class SimulatedCameraContainer(Container):
+    """Contains all fields of the DL1CameraContainer, but adds fields for simulated
```
maxnoe, line 413: Not really true, since we do not have peak times here. (We could calculate them in the simtel event source, see #1341

ctapipe/containers.py
```diff
@@ -922,3 +892,23 @@ class SimulatedShowerDistribution(Container):
         None, "array of core-distance bin lower edges, as in np.histogram", unit=u.m
     )
     histogram = Field(None, "array of histogram entries, size (n_bins_x, n_bins_y)")
+
+
+class DataContainer(Container):
```
maxnoe, line 897: `ArrayEventContainer`?

ctapipe/containers.py
watsonjj: Only problem is it needs to be clear that there is an unknown offset between these simulated mean times, and the times extracted from the waveform

maxnoe: Yes, these times are in the CORSIKA system and I don't know anything in the SimTel output that relates CORSIKA times to the trigger time

maxnoe: I'll write Konrad

kosack: That was an old comment from when this originally inherited from DL1CameraContainer -  I'll just update it to be more accurate. Not sure if we really need the simulated mean times or not. Could be useful for benchmarking, I guess.

ctapipe/containers.py
kosack: Sure, might as well rename it now.  I think at this point it really does only contain event-wise info (finally!).

ctapipe/containers.py
maxnoe: Konrad answered, and it is not easy to get these CORSIKA times converted to times relative to the array trigger, so I would leave it out for someone who really needs this to implement.

maxnoe (dismissed):

vuillaut (commented):
Hi
Is there a _rule_ to choose between _MC_ and _Simu_.
You seem to favor _Simu_ lately but a lot of _MC_ stays and both are used/(re)introduced in this PR.

ctapipe/containers.py
```diff
@@ -432,7 +408,32 @@ class MCEventContainer(Container):
...
+        ImageParametersContainer(), "Parameters derived from the true_image"
+    )
+
+
+class SimulatedEventContainer(Container):
+    shower = Field(SimulatedShowerContainer(), "True event information")
+    tel = Field(Map(SimulatedCameraContainer))
 
 
 class MCHeaderContainer(Container):
```
maxnoe, line 439: This should also be renamed to "SimulationHeaderContainer"

kosack:
I think now I've got rid of all "MC" jargon. I also updated the docstrings for a few of what was the `MCHeaderContainer` (now `SimulationConfigContainer`).  Note that that probably could use some cleanup in the future to be more general - it has a lot of very simtel-specific info like fields and identifiers, but that can come with a later refactoring

maxnoe (approved):

vuillaut (approved):

Issue #1073, opened 2019-05-16
Remote test data vs ctapipe-extra


kosack:
Soon, I would like to start to keep only configuration files and defaults in ctapipe-extra (things like camera geometries, subarray layouts, etc), and move all large test files to a separate server with more space available (to avoid breaking the package size limits)

We now have a place for remote testing data.  @bregeon has uploaded the latest ctapipe-extra info to the publicly available data server, so we can start to migrate any code like notebooks that need larger test data to use that location.  This will be versioned as follows:

```
[server-base-path]/
    - ctapipe-test-data/
         - v1.0.0/
            [files and subdirs]
```

In version 1.0, this is just a copy of ctapipe-extra.  In future versions, we might organize it better.  The next step will be to make some nice utility functions in ctapipe that access this without the user having to give the full path.   In subsequent versions, we can simply symlink files that don't change to save space, and/or add new ones. 

Downloading the data works fine with astropy's `download_file` function as shown here:

![image](https://user-images.githubusercontent.com/11677812/57852060-6c9c2f00-77e2-11e9-8b2b-f678b8625198.png)

kosack:
To avoid this from blowing up into an unmanageable set of files with no idea what is what, please give some suggestions for further organization of this data. E.g.:

* what test data files do we need? e.g Small sample of gammas, protoons, electrons? from which prods?, including what info (PE images?, calibration events?, etc)
* how should we name them to be clear what they are? (e.g. these could be from different productions, different arrays, etc - we need a nice way to clearly state what is used for what purpose)

kosack:
This relates to issues:
* #838 How to add LST test resources
* #738 selectable datasets
* #558 clean up tests that load data

MaxNoe:
Looks good! Can we make a small wrapper function that uses our server by default and a different storage location?

kosack:
> Looks good! Can we make a small wrapper function that uses our server by default and a different storage location?

Yes, I think the best way to start is something similar to the `get_dataset_path` function in ctapipe, something like `fetch_test_dataset(name)`. 

It reminds me of another issue I would like to bring up: having some notion of global "Services", which are basically `Components` who get instantiated and configured every time you load the ctapipe module (where the configuration comes from some default `.ctapiperc` file or similar, or from a Tool's configuration system if it exists). These would be things like this test data fetcher (configurable version and server path), the provenance system (with some configuration options), and for example a random number generator (with configurable seed).

[...]

Issue #849, opened 2018-11-29
Make the example  and test data auto-downloadable, like gammapy


kosack:
- remove from `ctapipe_resources`
- make a command  `ctapipe download test-data --version=0.3`
- if version is omitted, latest version is downloaded
- use the In2p3 server (for now) to store the data sets in versioned subdirs - this is accessible from wget or using the astropy download tools

maxnoe:
This was solved (minus the download command, which I am not sure we need)

PR #1498, opened 2020-10-29
Download test files

maxnoe:
@bregeon for now put the ctapipe extra files on the download server, so I just had to remove the ctapipe-extra installation and add the download instructions to `get_dataset_path`.

For later, we can think of a better file structure on the server.


ctapipe/utils/download.py
```diff
@@ -0,0 +1,133 @@
...
+        raise
+
+    # when successful, move to intended location
+    part_file.rename(path)
+
+
+def get_cache_path(path, cache_name="lstchain"):
+    # need to make it relative
+    path = str(path).lstrip("/")
+    path = Path(os.environ["HOME"]) / ".cache" / cache_name / path
```
maxnoe, line 68: Question to the mac users: Is this also an approproaiate location on mac?

kosack:
Please don't roll your own downloader and just use the one provided in astropy, as it caches automatically and is well debugged. 

We already have a few issues on this: 
 - #1073 (remote test data)
 - #849 (downloadable test data like gammapy)

kosack (changes requested):
see linked issues

ctapipe/utils/download.py
kosack: this is likely forbidden on the grid, where you can only write to the scratch disk. I would also avoid putting things in HOME in general, since that directory is backed up (and often quota-restricted) for most people, and these files should not be.

maxnoe:
The astropy mechanism does not support authentication and does not allow overriding the cache location.

ctapipe/utils/download.py
maxnoe: Why should we event download these things on the GRID?

maxnoe: But I will add an override for the cache.

kosack: We could re-use the same functionality in the future for fetching things like calibration files, perhaps.  I'm not sure we'd ever run these tests on the grid, but in general it's best to support running anywhere we expect to run ctapipe. We already have the idea to do long-term tests on the grid, for example (though we'd use probably DIRAC to fetch the files needed for each job in that case).

maxnoe: I added the env variable, which should also be useful for e.g. testing on our own testing machine with data available (although, the CTAPIPE_SVC_PATH could also be used for that)

maxnoe:
Yes, I know the issues. This is to address parts of them. Not yet the fully structured part, but the downloading part.

kosack:
Do we need authentication? It works with astropy's downloader already.   But the cache location may be a problem - still would be good to maintain less code and use what's already in our main dependency.  

I see gammapy does do their own thing now, however: https://github.com/gammapy/gammapy/tree/master/gammapy/scripts

kosack:
You can set the cache directory for Astropy: 

https://docs.astropy.org/en/stable/api/astropy.config.set_temp_cache.html#astropy.config.set_temp_cache

maxnoe:
According to several people, yes we need authentication as the test data (and especially the newly created gridded test data) won't be public.

Also, I want to reuse these functions for LST, which at the moment definetely needs authentication.

Actually I started writing these function ins `ctapipe_io_lst` but then moved them here.

maxnoe:
Ok, tests are passing now.

I think independent of downloading, I improved the datasets logic a bit by making sure paths are returned (if this regexp group option is not used in `get_all_matching_datasets`.

I think the solution I propose here is relying on a very well tested library (requests) and is not much more code to maintain.

So I think this is preferable to relying on astropy dowload functions.

ctapipe/utils/datasets.py
```diff
@@ -142,12 +152,55 @@ def get_dataset_path(filename):
 
         return ctapipe_resources.get(filename)
 
+    # last, try downloading the data
+    try:
+        return download_file_cached(
+            filename,
+            default_url="http://cccta-dataserver.in2p3.fr/data/ctapipe-extra/v0.3.1/",
```
kosack, line 159: Perhaps at least make this default url a module-level variable (e.g `DEFAULT_DOWNLOAD_URL`), so later we can change it more easily.  Or evem, anticipating selectable datasets, add `DATASET_VERSION = 'v0.3.1'` and then use 

```py3
default_url=f"http://cccta-dataserver.in2p3.fr/data/ctapipe-extra/{DATASET_VERSION}"`
```

That way later we can test other versions by just setting `ctapipe.utils.datasets.DATASET_VERSION` 

Another option would be to do what you do later with ENV_PREFIX, and add the ability to set envvars like `CTAPIPE_DATA_VERSION=v0.3.1` and/or `CTAPIPE_DATA_URL`

kosack (dismissed):

ctapipe/utils/datasets.py
maxnoe: CTAPIPE_DATA_URL is already builtin.

maxnoe:
@kosack I resolved conflicts and created the `ctapipe.utils.datasets.DEFAULT_URL` module level constant.

kosack (approved):

kosack (dismissed):

kosack (approved):

kosack (approved):

ctapipe/conftest.py
```diff
@@ -8,6 +8,15 @@
 
 from ctapipe.io import SimTelEventSource
 from ctapipe.utils import get_dataset_path
+from ctapipe.instrument import CameraGeometry
+
+
+@pytest.fixture(scope="session")
+def camera_geometries():
```
watsonjj, line 15: Is this actually used anywhere?

ctapipe/instrument/camera/tests/test_description.py
```diff
@@ -1,7 +1,7 @@
 from ctapipe.instrument import CameraDescription
 
 
-def test_known_camera_names():
+def test_known_camera_names(camera_geometries):
```
watsonjj, line 4: This argument doesn't seem to be used

maxnoe: The test asserts that at least 5 geometries are known, the fixture ensures that at least 5 geometries have been downloaded, so they are known.

ctapipe/conftest.py
maxnoe: yes, it makes sure we have the required number of known cameras in the test you commented on below.

ctapipe/instrument/camera/tests/test_description.py
watsonjj: But `cams` is from `CameraDescription.get_known_camera_names()`, `camera_geometries` doesn't look like its used...

watsonjj: Oh, so `from_name` will download the files if they aren't already, so that `CameraDescription.get_known_camera_names()` can find them?

watsonjj: It seems a bit of a convoluted way to ensure this for the test (not explicit)

maxnoe: Yes, agrees

maxnoe: We could just remove the `get_known_camera_names` function.

It is not used in any actual analysis and most examples use specific cameras and not something like `get_known_camera_names()[0]`.

kosack: I like the `get_known_camera_names()` it's a nice feature to be able to see what cameras you can construct. I would not like to remove it.   My original suggestion was to have an index database table (just a FITS table with filename, type, etc in columns) on the server, and you can use that to query what files are there without fetching the data

maxnoe: Yes, this was my idea as well. Having this indexes only for things like cameras might me a good compromise to allow full indexing of the server, which was not wished by @bregeon

maxnoe: Shall we merge as as and leave the index files to the planned full restructuring of the test data sets?

watsonjj (approved):

ctapipe/instrument/camera/tests/test_description.py
kosack: yes, that's fine with me

kosack (approved):

Issue #1184, opened 2019-11-22
chaining components that need a final step after all events


kosack:
## Algorithms with event-wise and final calls:

I'd like to come back to the issue of how to chain together a sequence of Components in a Tool.

I'm refactoring #1066 to break it into Components, and I need to make a `DL1EventWriter` component, that writes some final tables when all event writing is done.   The general data flow is roughly like this:

```py3
source = event_source(filename)
subarray = source.subarray

calibrate = CameraCalibrator(..., subarray=subarray)
process_images = ImageProcessor(..., subarray=subarray)
write_dl1 = DL1Writer(..., subarray=subarray)

for event in source:
     calibrate(event)
     process_images(event)
     write_dl1(event)

write_dl1.finish()  # easy to forget to call this...
```

### use context management

```py3
with DL1Writer(...) as write_dl1:
   for event in source:
      calibrate(event)
      process_images(event)
      write_dl1(event)

# cleanup happens automatically at end of "with" block
```

I like this the best if there are only a small number of components with context, but it may get ugly if there are many. Note that python does allow multiple context managers on a line:
```py3
with X() as x , Y() as y, Z() as z:
     ...
```
or there is the `contextlib.ExitStack` class, which is a bit uglier, but works with an arbitrary list of objects (https://docs.python.org/3/library/contextlib.html#contextlib.ExitStack)

### use processor model

A less pythonic alternative would be to use   a *Processor* like interface  (#591) for at least some Components (with minimally `process()` and `finish()` methods, but we could also have an `on_first_event()` method)  That could be done in a destructor ( `__del__`), but then it is called only when the garbage collector decides to call it, so it would likely cause problems. 

As mentioned in #591, one way to go would be to use function composition, chaining , or similar to achieve this. 
```py3
steps = [calibrate, process_images, write_dl1]
map( apply_steps(steps), source )
finish_steps(steps)
```
Note that something similar can still be used in the context management way of doing things.

Right now I'm leaning toward the former (context management) method, but I'm open to suggestions.

kosack:
It occurs to be that in the context management option, we can even implement some delayed initialization using `__enter__()`.  So we can just have configuration stuff in `__init__`, and longer tasks like file handling inside `__enter__`

```py3
In [74]: class C(contextlib.AbstractContextManager):
    ...:     def __init__(self,):
    ...:          print("INIT C")
    ...:     def __enter__(self, ):
    ...:          print("ENTER C")
[...]

PR #1499, opened 2020-10-30
Create a DL1Writer  and ImageProcessor to simplify stage1

kosack:
This is a refactoring of the the `ctapipe-stage1-process` tool  to remove the parts that write DL1 data, and put them into a reusable `Component` called `DL1Writer` that can be called separately. This reduces most of stage1.py into a few lines:

```py3
calibrate = CameraCalibrator(subarray=subarray, parent=self)
process_images = ImageProcessor(subarray=subarray, parent=self)

with DL1Writer(event_source=event_source, parent=self) as write_dl1:
    for event in source:        # does R1,R1 to DL1 step
          calibrate(event)      # does "DL0" to "DL1 image" step
          process_images(event) # does "DL1 image" to "DL1 parameters" step
          write_dl1(event)      # writes a standard DL1 file
```

This **greatly simplifies `stage1.py`**  (who's command-line version is now renamed to just `ctapipe-stage1`) and now it contains only  a loop similar to above.

## TODO:
- [x] ensure all functionality is there (compare output with current stage1.py)
- [x] write tests
- [x] refactor `stage1.py` to use this functionality
- [x] update example *stage1_config.json*
- [x] update docs / tutorials
- [x] write ImageProcessor and use it in stage1.py
- [x] add ImageProcessor tests

## Issues that need resolution:
 - [x] #1184 (chaining components that need a finish funciton) → DEFERRED (so far just call `DL1Writer.finish()` explicitly)
 - [x] #1495 (refactor simulation info, needed since SimulatedCameraContainer is currently missing from the event structure)
 - [x] #1500 (provide access to histograms) → DEFERRED (currently use old method, called explicitly)

## Other Comments
 - This could likely be generalized to be just a DataWriter, that could write DL1 or DL2 data, depending on user options. That would make the next stage (Stage2) much easier to write, and allow things like a training script (à la protopipe) that does both parts of DL1 and DL2 at once. 
- the next step *after this PR* will be to refactor all the image-processing parts of `stage1.py` also into a Component, so that everything is a re-usable brick, and we can later create tools (E.g. `ctapipe-train`) that write out training data, using a mix of DL1 and DL2 stages. I will also switch from CameraFrame Hillas parameters to TelescopeFrame Hillas parameters at that point.


ctapipe/io/dl1writer.py
```diff
@@ -0,0 +1,446 @@
...
+            containers=[event.index, event.trigger],
+        )
+        if self.event_source.is_simulation:
+            self._writer.write(
+                table_name="simulation/event/subarray/shower",
+                containers=[event.index, event.mc],
+            )
+
+        # write telescope event data
+        self._write_telescope_event(self._writer, event)
```
maxnoe, line 155: plural?

kosack: yes, updated.

maxnoe (dismissed):
I don't see that you actually changed the DL1 Tool, so this has now much duplicated code, right?

kosack:
> I don't see that you actually changed the DL1 Tool, so this has now much duplicated code, right

That's why it's still a draft PR - just wanted to show that it was happening and get some general feedback, but I'm nowhere near done with it (see the checklist at the top).  Once I have it working, I'll remove the code in the stage1 tool. Right now it doesn't yet work.

maxnoe:
Oh, ok, sorry. I think this is a very good idea to have the tools be minimal and only composing components!

kosack:
Right now this is at a point where it needs #1495 to be merged first, so I'll wait for that

docs/ctapipe_api/io/index.rst
```diff
@@ -115,6 +115,24 @@ using the `ctapipe.io.HDF5TableReader`, or more generically using the
...
 only supports scalar values).
 
+Writing Output Files:
+=====================
+
+The `DL1Writer` Component allows one to write a series of events (stored in
+`ctapipe.containers.ArrayEventContainer`) to a standardized HDF5 format DL1 file
+following the DL1 data model. This includes all related datasets such as the
+instrument and simulation configuration information, simulated shower and image
+information, and real images and parameters. It can be used in an event loop
```
maxnoe, line 125: real → observed?

I don't like real / simulated as opposites.

kosack: Agreed - I used observed elsewhere, and I think it's nicer than "real"

kosack:
By the way, now that the merge tool is called `ctapipe-merge`, I think we could rename `ctapipe-stage1-process` to just `ctapipe-stage1` either in this PR or another separate one (originally the name was anticipating having several sub-tools like `ctapipe-stage1-merge`, but that is no longer the case).

kosack:
Any idea why overwriting the output file fails for *pip* builds, but not conda? The code seems pretty simple: if overwrite is specified, the file is first unlinked (deleted).  But in the pip build, I get an error that the file can't be opened for output
```
Unable to create file (unable to open file: name = '/tmp/tmp6n4347kc.hdf5', errno = 17, error message = 'File exists', flags = 15, o_flags = c2)
```
and later:
```
Unable to open file (unable to lock file, errno = 11, error message = 'Resource temporarily unavailable')
```
Deleting should be a blocking operation, so not sure why that would happen (and not in conda)

maxnoe:
Is that a question of `h5py` vs. `pytables` not using the same libhdf5 again? So did you somehow change the order of using astropy and tables io?

The order was important to avoid problems like these.

kosack:
> Is that a question of h5py vs. pytables not using the same libhdf5 again? So did you somehow change the order of using astropy and tables io?
> The order was important to avoid problems like these.

It's possible.. let me check.   strange that it woudl still work with conda though, but could be a difference in h5py versions or something.  I also notice that test_tools.py massivly mis-uses NamedTemporaryFile and the tmpdir/tmp_path fixtures of pytest, so I'll re-write that a  bit as well

maxnoe:
> strange that it woudl still work with conda though

The conda packages of h5py and tables are linked vs the same libhdf5, the one that comes with `conda install hdf5`. The pip wheels each bring their own statically linked hdf5 library.

kosack:
The order seems ok the first thing in both the old and new versions is that we create the file by using `subarray.to_hdf(output_path)`.  Then we close that file, and process the rest with pytables.   I compared both versions, and see no difference.  The problem seems only when during `test_merge()`, it tries to overwrite a file.

kosack:
I think it was just an order problem - hopefully fixed now, it was trying to write the processing stats before the file was closed by DL1Writer.

kosack:
This is ready to go now - the complaints from codecov are actually about the ctapipe-merge tool, not this PR

HealthyPear:
>     * the next step _after this PR_ will be to refactor all the image-processing parts of `stage1.py` also into a Component, so that everything is a re-usable brick, and we can later create tools (E.g. `ctapipe-train`) that write out training data, using a mix of DL1 and DL2 stages. I will also switch from CameraFrame Hillas parameters to TelescopeFrame Hillas parameters at that point.

this would be already coincident with the upgraded version of `protopipe.scripts.data_training` no?

ctapipe/image/tests/test_image_processor.py
```diff
@@ -0,0 +1,32 @@
...
+        image_cleaner_type="MARSImageCleaner",
+    )
+
+    assert isinstance(process_images.clean, MARSImageCleaner)
+
+    calibrate(example_event)
+    process_images(example_event)
+
+    for dl1tel in example_event.dl1.tel.values():
+        assert isfinite(dl1tel.image_mask.sum())
```
HealthyPear, line 23: wouldn't be easier (and safer) to assert in a cycle over all the Field keys?

kosack:
> this would be already coincident with the upgraded version of protopipe.scripts.data_training no?

Yes, actually that comment is already out of date in fact - this PR now includes that refactoring.

ctapipe/image/tests/test_image_processor.py
kosack: probably - but this isn't meant to be a full test.  We already test that `hillas_parameters` and other similar functions generate the correct `Containers` as output.  Here I just want to make sure that each parameter container in the event structure is filled with non-default values (so not NaN), so I only really need to test one field from each (and assume the individual tests already cover the details).   basically the test here is "did hillas_parameters get called and store some output", not "did it fill all columns"

ctapipe/tools/stage1.py
```diff
@@ -1,240 +1,81 @@
...
         ),
     }
 
     classes = List(
-        [CameraCalibrator, ImageQualityQuery, EventSource]
+        [CameraCalibrator, EventSource]
         + classes_with_traits(ImageCleaner)
         + classes_with_traits(ImageExtractor)
         + classes_with_traits(GainSelector)
+        + classes_with_traits(DL1Writer)
```
maxnoe, line 74: Isn't `classes_with_traits` though to be for abstract base classes? I think you should just add `DL1Writer` and `ImageProcessor` to the list without the `classes_with_traits` call. But I'm not sure.

maxnoe (dismissed):
This looks very good!

maxnoe:
One minor question, everything else LGTM.

ctapipe/tools/stage1.py
kosack: Yes, I think you are right - I had changed that at some point, but reverted it.  `classes_with_traits` (confusingly named), really just looks for child classes of the parent class.   So far DL1Writer has no sub-classes, so this doesn't really do anything other than include the parent (so this still works, but is probably redundant) .

maxnoe: I think that means it only makes sense to use this where we use the `create_class_enum` approach, right? So for anything where we don't do that, it might actually be misleading, since there is no way the listed configuration options for subclasses actully take effect, since there is no way to make the tool use a subclass.

I would remove it then.

ctapipe/io/tests/test_dl1writer.py
```diff
@@ -0,0 +1,65 @@
...
+
+    Parameters
+    ----------
+    tmpdir :
+        temp directory fixture
+    """
+
+    output_path = Path(tmpdir / "events.dl1.h5")
+    source = event_source(
+        get_dataset_path("gamma_test_large.simtel.gz"),
```
HealthyPear, line 25: you should be able to use Prod3B files now

ctapipe/tools/stage1.py
kosack: ok, will remove

HealthyPear (changes requested):

ctapipe/io/tests/test_dl1writer.py
kosack: yes, but that's for another PR - gamma_test_large will be prod 3 (or 5) in the future once we have the ability to switch between test datasets.

kosack: I have tested this with prod3b externally, however, and it's fine.

kosack: Or do you mean we have a prod3b test file already?

maxnoe: yes, there are 4 `gamma_*_prod3b_test.simtel.gz` files in ctapipe extra (and on the download server)

maxnoe: pointlike La Palma, Point Like paranal, diffuse la palma, divergent la palma

HealthyPear: yes I added prod3b test files sometime ago in ctapipe-extra

kosack: Ok, will switch to one of those.

maxnoe (dismissed):

ctapipe/io/tests/test_dl1writer.py
kosack: now uses the prod3b file, and I updated ctapipe-extra's dependency to require 0.3.1 (since 0.3 didn't have that file), I also created and uploaded ctapipe-extra-0.3.1, since it wasn't in conda

environment.yml
```diff
@@ -43,4 +43,4 @@ dependencies:
   - zlib
   - zstandard
   - conda-forge::eventio>=1.1.1
-  - ctapipe-extra>=0.3
+  - ctapipe-extra>=0.3.1
```
maxnoe, line 46: You also need to change this in `setup.py` or we merge the download branch first

kosack: done

HealthyPear (approved):

maxnoe (approved):

PR #1408, opened 2020-08-18
Direction reconstruction (also) from Telescope Frame

HealthyPear:
`ctapipe.reco.HillasReconstructor` has been modified to account for the possibility of receiving parameters obtained also in the `TelescopeFrame`.

The unit-test `test_reconstruction` has been improved:
- it cleans and parametrizes the image also in the `TelescopeFrame`,
- it compares the reconstruction from both frames
- it is required that all of the shower fields are the same within a relative accuracy of 0.001%
- after adding new test files to ctapipe-extra I now test for on Prod3b files in both parallel and divergent case separately

I have recycled my test scripts to make an updated version of the stereo reconstruction example named `stereo_reconstruction_core_and_direction.py`.

I tested parallel and divergent pointings, on-axis and off-axis.

- reconstruction vs simulation:

<p align="center">

</p>

- pure divergent pointing using `gamma_20deg_180deg_run9___cta-prod3-demo-2147m-LaPalma-baseline.simtel.gz` and 
comparing the old approach (i.e. cleaning and parametrizing in CameraFrame) against the new one (do that in TelescopeFrame)

divergent

- all together (parallel from CameraFrame, parallel from TelescopeFrame and divergent with same pointing from TelescopeFrame - in this case, parallel and divergent must coincide)

all

PR #1504, opened 2020-11-04
Hillas Parameters in TelescopeFrame instead of CameraFrame

kosack:
This PR depends on #1499 , so will wait until that is merged to review (so the diffs don't include those changes, or otherwise choose that PR as the comparison). 

It makes the following changes:
1. pre-computes the `CameraGeometries` in the `TelescopeFrame` and uses those in the parametrization
2. changes the Container data model for `HillasParameters` and `TimingParameters` so that the units are `deg` for positions and distances, and `deg-1` for velocities.  (note that it's then a bit confusing which parameters are in *degrees on a sphere* vs *rotation angle*, but the help text should be clear)

Some open questions:
- **Pointing**:  the pointing of MCs is fixed, so currently we use the same TelescopeFrame geometry for all events and currently there is no mechanism to update the pointing position (or rotation in the case of mis-pointing that includes that). That should work also for real data that tracks a source, since the TelescopeFrame is centered on the alt/az of the pointing of the telescope and the positions `hillas.{x,y}` are deltas from this center position on a sphere. However, when we go to any common frames or use 3D reconstruction, we will have to think about how to apply updated pointing values  or pointing corrections in an efficient way.   This relates to #1408.


HealthyPear:
Now I am a bit confused...I think some of these changes are also in #1408 - am I wrong? should we merge this before? I will try to finish my PR sorry...

kosack:
> Now I am a bit confused...I think some of these changes are also in #1408 - am I wrong? should we merge this before? I will try to finish my PR sorry...

Yes, we should probably merge this into your PR -this was just to fix the stage1.py tool to do it correctly, but obviously it breaks the reconstruction in the process (so would need your PR).  I suggest: 
1. once #1499 is merged, you should update #1408 to the master
2. then merge this PR into #1408, and we can proceed from there and do it all at once.

kosack:
I'm closing this for now and will come back to it. It will be easier to implement after the full refactoring of stage1.py

PR #1507, opened 2020-11-06
Fix naming of image -> true_image for simulated containers.

nbiederbeck:
Also remove not-used containers.

This fixes reading of DL1 from simulations processed with stage1


LukasNickel (approved):

maxnoe (changes requested):
Wouldn't this be easier solved by just adding a `true` prefix to the simulated container?

LukasNickel:
The reading is fine.
It's just, that the fields in the `SimulatedCameraContainer` are named `true_...` so they were not properly assigned.

maxnoe (approved):

kosack:
I already fixed this in #1499 , but doesn't hurt to merge this since I think we did the same thing.

PR #1509, opened 2020-11-06
reduce erroneous warnings in code style checkers

kosack:
Codacy currently gives a lot of false or useless warnings, meaning that most PRs get marked with an X.  This is a first step to avoiding that: it creates a `.pylintrc` file with reasonable options.  So far:

- generated initially using `pylint --generate-rcfile`
- added exclusion for `astropy.units` (all generated members fail erroneously for pylint)
- disabled rules for too-many-arguments and too-many-locals
- increase max parents from 7 to 9 (since Tool classes have 8 of them due to traitlets)
- added x,y,n to list of ok var names

This also updates `setup.cfg`'s `[flake8]` section to disable some rules (though it seems codacy might ignore this, so have to see where to put this config if so):
- check of unused imports in `__init__.py` files (which are normally correct)

Probably more needs to be added.  Please suggest others to disable/relax here.  You can test by downloading this PR and running "pylint ctapipe"

### DIFF
--- a/.pylintrc
+++ b/.pylintrc
@@ -542,7 +542,9 @@ exclude-protected=_asdict,
                   _fields,
                   _replace,
                   _source,
-                  _make
+                  _make,
+                  _v_attrs, 
+                  _v_attrnames
 
 # List of valid names for the first argument in a class method.
 valid-classmethod-first-arg=cls
--- a/.pylintrc
+++ b/.pylintrc
@@ -1,0 +1,592 @@
+[MASTER]
+
+# A comma-separated list of package or module names from where C extensions may
+# be loaded. Extensions are loading into the active Python interpreter and may
+# run arbitrary code.
+extension-pkg-whitelist=
+
+# Specify a score threshold to be exceeded before program exits with error.
+fail-under=10.0
+
+# Add files or directories to the blacklist. They should be base names, not
+# paths.
+ignore=CVS
+
+# Add files or directories matching the regex patterns to the blacklist. The
+# regex matches against base names, not paths.
+ignore-patterns=
+
+# Python code to execute, usually for sys.path manipulation such as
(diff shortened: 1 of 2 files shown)


maxnoe:
I think you can put flake8 rules into `.flake8`

.pylintrc
```diff
@@ -0,0 +1,594 @@
...
+max-public-methods=20
+
+# Maximum number of return / yield for function / method body.
+max-returns=6
+
+# Maximum number of statements in function / method body.
+max-statements=50
+
+# Minimum number of public methods for a class (see R0903).
+min-public-methods=2
```
maxnoe, line 586: I think we have use cases where we want 1 public method (function like) but need to use a component for the configuration system, e.g. the `CameraCalibrator` or `Reconstructor` classes

kosack: done

kosack:
What is strange is that I really can't tell what Codacy is doing sometimes it complains about things that are clearly disabled, and some times not, and sometimes it can't import things like numpy, but not always.  It is very inconsistent.  Hopefully this will help, but I see already in the codacy check for this PR warnings that were disabled.   Perhaps there is a better alternative to codacy.

HealthyPear:
In _pyirf_ I added a `.prospector.yaml` to configure what warnings I want _Codacy_ to ignore.

maxnoe:
That file configures only warnings related to docstrints (pep257)

kosack:
> In pyirf I added a .prospector.yaml to configure what warnings I want Codacy to ignore.

Prospector in codacy just runs pylint and pydocstyle, so hopefully this is equivalent (and lets us run the tools locally as well)

maxnoe (approved):

HealthyPear (approved):

Issue #1077, opened 2019-05-22
provide some numba  versions of standard algorithms


kosack:
This is not required at this point, but still might be interesting to test

Some functions like `tailcuts_clean` (and other cleaning variants), and `hillas_parameters`  are called often (once per event per telescope) and could probably be sped up without major changes to the code, and even without using an external hand-coded library like *hipeCTA*, using only Numba.  This worked well with the charge extraction code, and could be extended to other image functions. 

This may also be useful to the teams doing realtime-analysis prototyping. 

## Example:
Right now, simply applying numba to e.g. `tailcuts_clean` has no effect, since the inputs are a python object (CameraGeometry) and a numpy array, e.g. the function signature is 
```py3
def tailcuts_clean(geom : CameraGeometry, image: np.ndarray) -> np.ndarray:
    ...
```

Since numba doesn't know what to do with `CameraGeometry`, it can't optimize. Splitting this into two functions with simple numpy inputs to one would allow it to work:
```py3
def tailcuts_clean(geom, image):
    return _tailcuts_clean_fast(
        neighbor_matrix=geom.neighbor_matrix_sparse, 
        image=image
    ) 

@numba.njit(parallel=True)
def _tailcuts_clean_fast(...):  # this takes only numpy arrays and is optimizable
    # basically a cut-paste of the current code, after "unpacking" the CameraGeometry
```
This would also allow us to broadcast this same optimized function to a large set of images if in the future we load more than one event at a time. 

## Where applicable:
This pattern could be applied at least to the following functions (probably more), but testing on a few would be interesting.
* `ctapipe.image.cleaning`:
  * `tailcuts_clean()`  (good place to start with a test) 
  * `dilate()`
  * `number_of_islands()`
  * `apply_time_delta_cleaning()`
  * `fact_image_cleaning()`
* `ctapipe.image.hillas.hillas_parameters()` (numba does seem to implement `np.linalg`, etc, so should work)
* `ctapipe.image.timing.timing_parameters()`
* `ctapipe.image.leakage.leakage_parameters()`

kosack:
for reference, Numba currently supports optimization of many numpy functions shown here: https://numba.pydata.org/numba-doc/dev/reference/numpysupported.html#functions

kosack:
[...]

PR #1512, opened 2020-11-12
faster JIT hillas_parameters function

kosack:
I did a quick profile of stage1.py with 8000 events, I found now that the dominant CPU usage is now in `ImageProcessor`, which is slower than the CameraCalibrator (when NeighborPeakWindowSum is used at least). One of the slower parts of that is the `hillas_parameters` function.

## explore if using numba has a significant effect
(this PR may not be merged, it's just a "what if" scenario).  I suspect the overall speed of processing a run will not be largely affected, but this could be interesting for online analysis or RTA. 

This PR refactors the `hillas_parameters()` function  (as in #1077) such that parts of it are compiled using *Numba*. Specifically it defines a `hillas_parameters_fast()` that takes only unitless numpy arrays (and that function is called by the usual `hillas_parameters()` function, so no change of API is made.

While doing this I discovered:
* Numba doesn't support weighting in `np.cov`, so I implemented a very simplistic weighted 2d covariance function using numba (not generally useful, but just the bare minimum to work with the hillas_parameters funciton). Current implementation uses DOF=1, rather than 0
* Numba doesn't support np.ma.masked arrays (so dropped support for them in hillas_parameters) - we didn't use those anyhow, and `hillas_parameters(geom[mask], image[mask])` is preferred and faster
* Numba doesn't support `np.average`, only `np.mean` which doesn't do weighed average, so implemented again a simplistic `weighted_average` function.

### Results:
Tests make using an image of 100 pixels
download-2
`covariance_matrix` is about 13.5x faster than the numpy version

download-4
`hillas_parameters` is about 4x faster than the numpy version, but 35x faster in it's pure form (hillas_parameters_fast with no units overhead)


kosack:
Line profile of the hillas_parameters function (that calls the fast version internally):
```
%lprun -f hillas_parameters_wrapped hillas_parameters_wrapped(g, w0)
Timer unit: 1e-06 s

Total time: 0.000513 s
File: <ipython-input-11-33005bf5af05>
Function: hillas_parameters_wrapped at line 7

Line #      Hits         Time  Per Hit   % Time  Line Contents
==============================================================
     7                                           def hillas_parameters_wrapped(geom, image):
    53         1         12.0     12.0      2.3      unit = geom.pix_x.unit
    54         1         56.0     56.0     10.9      pix_x = Quantity(np.asanyarray(geom.pix_x, dtype=np.float64)).value
    55         1         21.0     21.0      4.1      pix_y = Quantity(np.asanyarray(geom.pix_y, dtype=np.float64)).value
    56         1          7.0      7.0      1.4      image = np.asanyarray(image, dtype=np.float64)
    57         1         20.0     20.0      3.9      if np.ma.is_masked(image):
    58                                                   raise TypeError(
    59                                                       "np.ma.masked arrays are not supported by hillas_parameters(). "
    60                                                       "Use `hillas_parameters(geom[mask], image[mask])` instead."
    61                                                   )
    62         1          1.0      1.0      0.2      msg = "Image and pixel shape do not match"
    63         1          2.0      2.0      0.4      assert pix_x.shape == pix_y.shape == image.shape, msg
    64                                           
    65                                               (
    66                                                   cog_x,
    67                                                   cog_y,
    68                                                   cog_r,
    69                                                   cog_phi,
    70                                                   size,
    71                                                   length,
    72                                                   length_uncertainty,
    73                                                   width,
    74                                                   width_uncertainty,
    75                                                   psi_rad,
    76                                                   skewness_long,
    77                                                   kurtosis_long,
    78         1         43.0     43.0      8.4      ) = hillas_parameters_fast(pix_x, pix_y, image)
    79                                           
    80         1         11.0     11.0      2.1      if np.isnan(size):
    81                                                   raise HillasParameterizationError(
    82                                                       "intensity=0, cannot calculate HillasParameters"
    83                                                   )
    84                                           
    85         1          1.0      1.0      0.2      return HillasParametersContainer(
    86         1         46.0     46.0      9.0          x=u.Quantity(cog_x, unit),
    87         1         21.0     21.0      4.1          y=u.Quantity(cog_y, unit),
    88         1         18.0     18.0      3.5          r=u.Quantity(cog_r, unit),
    89         1        105.0    105.0     20.5          phi=Angle(cog_phi, unit=u.rad),
    90         1          1.0      1.0      0.2          intensity=size,
    91         1         22.0     22.0      4.3          length=u.Quantity(length, unit),
    92         1         18.0     18.0      3.5          length_uncertainty=u.Quantity(length_uncertainty, unit),
    93         1         18.0     18.0      3.5          width=u.Quantity(width, unit),
    94         1         18.0     18.0      3.5          width_uncertainty=u.Quantity(width_uncertainty, unit),
    95         1         51.0     51.0      9.9          psi=Angle(psi_rad, unit=u.rad),
    96         1          1.0      1.0      0.2          skewness=skewness_long,
    97         1         20.0     20.0      3.9          kurtosis=kurtosis_long,
    98                                               )
```

So it looks like as expected units and the use of  `Angle` has a large performance hit.

maxnoe:
@kosack I noted this before, the creating new `Angle` instances is one of the major things right now. 

If you had profiled the hillas parameters before, you'd have seen not much time is spent in the actual computation that you tried to optimize via numba. At least that is what I saw when profiling the stage1 tool and looking at snakeviz output.

Also, not a small amount of time is spent in the rotation into shower coordinates. Which can for sure be made more efficient using numba and is not as much custom code compared to reimplementing `cov` and `average`

kosack:
This version is already faster (4.5x), and it seems getting rid of the Angles and replacing them with just Quantity, I get another factor of 1.5 increase, to nearly 7 times the original speed.  I'll try removing the Angle and see what it breaks - I think it's not really needed anywhere.

maxnoe:
Hi @kosack

We are cleaning up stale pull requests to get a better overview over PRs we want to merge before applying some larger refactorings, e.g. switching to src layout https://github.com/cta-observatory/ctapipe/issues/2078 or implementing CEP-002.

If you think this is still relevant, feel free to update your branch and re-open the pull request or continue discussion in an issue.

PR #1514, opened 2020-11-14
Fix parent handling in EventSource.from_config

maxnoe:
This also moves   io plugin detection into io.__init__, we had a few redundant calls specified at two places before. (nonabstract_children, event source directly before calling nonabstract children).

With this, setting input_url for the event souce should be fully supported now, if tools / components use `EventSource.from_config(parent=self)`.

### DIFF
--- a/ctapipe/core/component.py
+++ b/ctapipe/core/component.py
@@ -11,6 +11,39 @@ from ctapipe.core.plugins import detect_and_import_io_plugins
 __all__ = ["non_abstract_children", "Component", "TelescopeComponent"]
 
 
+def find_config_in_hierarchy(parent, class_name, trait_name):
+    """
+    Find the value of a config item in the hierarchy by going up the hierarchy
+    from the parent and then down again to the child.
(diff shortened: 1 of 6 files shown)


watsonjj:
This sounds like a flaw of the configuration system. Is it appropriate that a fix is only provided for `EventSource.input_url`, instead of a general fix for all Components? Or is `input_url` the only trait that needs this behaviour? (if so could you demonstrate why it is special?)

maxnoe:
No, this is not a limitation of the config system.

EventSource.input_url is special because we need to know input_url *before instance creation* to select the correct event source, but the rraits are only set during instance initialization.

This is the only case (now) where this is necessary.

Actually you wrote this method, didnt you? I just fixed its behaviour when parent is passed.

maxnoe:
The task is basically, to find the correct config value for `input_url` without constructing the class. Which is something traitlets does not seem to support out of the box.

watsonjj (approved):

kosack (approved):

PR #1519, opened 2020-11-16
Make sure get_dataset_path returns Path, fix event source test, fixes #1517

### LINKED ISSUE #1517: test_event_source_config looks broken to me.
There is a test ` def test_event_source_config(): ` in `ctapipe/io/tests/test_event_source.py`  
(and the one below) which seems broken to me. They use two identical strings to see if the config overwrites the `input_url` or the other way around.

_Originally posted by @dneise in https://github.com/cta-observatory/ctapipe/pull/1515#discussion_r523911616_

maxnoe:
fixes #1517

### DIFF
--- a/ctapipe/io/tests/test_event_source.py
+++ b/ctapipe/io/tests/test_event_source.py
@@ -126,20 +126,15 @@ def test_from_config_invalid_type():
         EventSource(config=config, parent=None)
 
 
-def test_event_source_config():
-    dataset1 = get_dataset_path("gamma_test_large.simtel.gz")
-    dataset2 = get_dataset_path("gamma_test_large.simtel.gz")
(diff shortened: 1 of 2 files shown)


kosack (approved):

HealthyPear (commented):
sorry I didn't get it from the issue: is input_url supposed to overwrite the dataset in config or the other way around?

maxnoe:
It is supposed to. More explicit should override less explicit. So kwarg overrides config overrides default.

HealthyPear (approved):

maxnoe:
Sorry, I mixed two things here, which where connected on my machine.

This is both the fix for #1517 and the issue mentioned in the title that `get_dataset_path` would return a `str` when it found the file in `ctapipe-extra` and a `Path` if found anywhere else.

PR #1520, opened 2020-11-16
add DOI link for v0.9.1 in readme

kosack:


README.rst
```diff
@@ -34,6 +36,7 @@ If you use this software for a publication, please cite the proper version using
 
 - v0.7.0 : |doiv07|
 - v0.8.0 : |doiv08|
+- v0.9.1 : |doiv09|
```
maxnoe, line 39: Small change, but maybe put the newest version on top? Sorting anti-chronologically?

maxnoe (approved):

HealthyPear (commented):
Do we list only the most recent bugfix release?
I don't see 0.9.0

maxnoe:
@HealthyPear 0.9 contains a known important bug and was superseded very quickly by 0.9.1. So I think not including it is a sound decision.

kosack:
Yes, the idea is that I will list only the "Stable" version of each major release (well minor using current numbering), so it will just be one link per 0.X release, with the latest patch number.  So for v0.9, it would be v0.9.1.

Semantic versioning implies that 0.9.1 is really the "0.9" release, but with some fixes.

maxnoe (approved):

HealthyPear (approved):

PR #1525, opened 2020-11-17
Fix exclusion filter for pointing table

maxnoe:
The `n_` was missing, resulting in `n_trigger_pixels` being added to the telescope pointing table

### DIFF
--- a/ctapipe/io/dl1writer.py
+++ b/ctapipe/io/dl1writer.py
@@ -311,7 +311,7 @@ class DL1Writer(Component):
                 )
 
             writer.exclude(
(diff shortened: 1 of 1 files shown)


kosack (approved):

kosack:
not sure how that was missed! thanks.  This reminds me we need a "validate" tool (probably just reusing some of the features in the merge tool), so that we can check that all the expected datasets and columns exist in an output file...

maxnoe:
And maybe more importantly, no unexpected columns are there.

HealthyPear (approved):

PR #1531, opened 2020-11-19
Add second set of hillas parameters, only using main island

maxnoe:


HealthyPear:
Why do we need this additional field?

This is basically a different cleaning, so it should be stored in the normal container from a different analysis no?

maxnoe:
What are you saying? That we should recalculate *all* parameters on the main island and have a second `ImageParamersContainer` at the `DL1CameraEvent` level?

I think that is not needed. But I'd like to have the main island hillas parameters, since these might be better suited for stereo reconstruction.

I don't want to replace them, because the other parameters might be better for G/H separation. So I think having to sets of hillas parameters in `ImageParametersContainer` makes sense.

HealthyPear:
I know, this is indeed what protopipe does, but I thought that the setup we have currently was linked to 1 cleaning/parametrization per program execution.

maxnoe:
> I know, this is indeed what protopipe does, but I thought that the setup we have currently was linked to 1 cleaning/parametrization per program execution.

I don't see it as two different cleanings. You can do this for any cleaning that could produce more than one island.

kosack:
It is a different cleaning, as if you use the image_mask in the data, it no longer matches the parameters.   

I also don't see why we get many islands if the thresholds are set correctly, so I would not expect this to affect many events.  Still waiting to see proof, though - has anyone done a study comparing the two?

maxnoe:
> It is a different cleaning, as if you use the image_mask in the data, it no longer matches the parameters.

Not really, since you can always again get the largest island from `image_mask`

kosack:
For me this implies you  want a non-linear workflow that is really a special case:

1. produce DL1-image data
2. from the DL1 image data produce 2 sets of DL2-parameters, one with standard cleaning, one with main-island cleaning
3. then  for example you reconstruct with the island cleaning (which takes standard dl1 parameter files)
4. you then do gamma-hadron separation with the standard cleaning (I guess?)

This could easily be done by having 2 DL1-parameters files derived from the same dl1-image file (so just need to run in two steps) 

Putting specialized hillas parameters in the data model just to support one very specific use case overly complicates it.  If we use other reconstructions or gamma-hadron separations, this is not the workflow (ImPACT for example).   It makes understanding what parameters were used for what difficult.

kosack:
Anyhow, I would not accept this without seeing a study that shows it is strictly necessary.  If it is, then fine, but I would like to keep the data model as simple as possible.

maxnoe:
> . If it is, then fine, but I would like to keep the data model as simple as possible.

Each image parameter is part of the data model definition? That sounds very inflexible to me. I think the data model should not prescribe more than `a table of image parameters` where `image parameter` is defined to be a single number per telescope event

kosack:
Yes, every parameter is part of the data model - subsequent stages of the analysis rely on the names being consistent, so they must be.  If a name changes, the data model version must change, and all things that read that data need updates.  Even if all data are self-consistent inside ctapipe,  remember that there will be analysis steps that are outside ctapipe that rely on the DL1 and DL2 data as well (training, benchmarking).

PR #1539, opened 2020-11-20
Only convert to angle if not alreadey

maxnoe:


kosack (dismissed):

kosack (dismissed):

kosack (approved):

kosack:
Thanks for figuring out the q.info thing! That's really subtle and surprising, but at least they seem to have some ideas to fix it in AstroPy

kosack (dismissed):

kosack (approved):

ctapipe/instrument/camera/geometry.py
```diff
@@ -800,7 +816,9 @@ def position_to_pix_index(self, x, y):
     @staticmethod
     def simtel_shape_to_type(pixel_shape):
         try:
-            return SIMTEL_PIXEL_SHAPES[pixel_shape]
+            shape, rotation = SIMTEL_PIXEL_SHAPES[pixel_shape]
+            # make sure we don't introduce a mutable global state
+            return shape, rotation.copy()
```
LukasNickel, line 821: Why do we need the `.copy()`?

ctapipe/instrument/camera/geometry.py
```diff
@@ -623,12 +636,15 @@ def rotate(self, angle):
             rotation angle with unit (e.g. 12 * u.deg), or "12d"
 
         """
+        angle = Angle(angle)
```
LukasNickel, line 639: Maybe also allow an `Angle` here and avoid the conversion in that case?

ctapipe/instrument/camera/geometry.py
maxnoe: angles are mutable and this injected a global, mutable variable into classes that should not have any reliance on global state.

ctapipe/instrument/camera/geometry.py
maxnoe: Rotate is not called often (not e.g. in every event)

kosack (approved):

HealthyPear (approved):

maxnoe:
On second thought, I think a better / safer way would be to add a `copy=True` keyword to the `__init__` method and use `copy=False` in `__getitem__`.

What do you think?

PR #1541, opened 2020-11-20
Fix missing traceback in colored formatter, fixes #1528

### LINKED ISSUE #1528: Tools do not log traceback on exceptions
For some reason, although `Tool` uses

```
            self.log.exception(f"Caught unexpected exception: {err}")
```

No traceback is logged when an exception  occurs.

maxnoe:


kosack (dismissed):

ctapipe/core/logging.py
```diff
@@ -18,13 +18,8 @@ class ColoredFormatter(logging.Formatter):
...
         """Format the LogRecord."""
-        rec = record.__dict__.copy()
-
-        rec["asctime"] = self.formatTime(record, self.datefmt)
-        rec["levelname"] = apply_colors(record.levelname)
-        rec["message"] = record.getMessage()
-
-        return self._fmt % rec
+        s = super().format(record)
+        return s.replace(record.levelname, apply_colors(record.levelname), 1)
```
kosack, line 22: s is a string here right? is is safe to assume using replace will get the right one? If somebody changes the format string to also have the work INFO  or DEBUG in it somewhere, assuming it is the first one could be bad. (e.g. if you set the string to have the level be the last thing printed, but also have one of those words in the output)

kosack: contrived example, if you had logs that looked like:
```
CTAPIPE-INFO Some output message [INFO]
```
where the part in brackets is the `levelname`, this would fail

kosack (changes requested):

ctapipe/core/logging.py
maxnoe: It would color the wrong info, yes.

maxnoe: I am not sure if supporting this log format (as opposed to the default one we use) is worth copying the log record.

kosack: alternately, just replace all instances. A bit of over-coloring is probably ok

maxnoe: ok, removed the `1`, more colors!

HealthyPear:
Have you considered using [this](https://coloredlogs.readthedocs.io/en/latest/) package to apply colors?

maxnoe:
There is also: https://pypi.org/project/colorlog/ which seems leaner.

kosack (approved):

kosack (approved):

nbiederbeck (approved):

PR #1544, opened 2020-11-20
Move from travis to github actions

maxnoe:


HealthyPear:
Since I need to do the same for protopipe I guess I will get inspired by this PR :)

why is there still Travis-related code if we do not want to use it anymore?

maxnoe:
Where do you mean?

.github/install.sh
```diff
@@ -0,0 +1,31 @@
...
+
+echo "Using conda: $CONDA"
+echo "Using python $TRAVIS_PYTHON_VERSION | $PYTHON_VERSION"
+
+
+if [[ "$CONDA" == "true" ]]; then
+  # Install miniconda following instructions at
+  # http://conda.pydata.org/docs/travis.html
+
+  if [[ "$TRAVIS_OS_NAME" == "linux" ]]; then
```
HealthyPear, line 11: here

.github/install.sh
```diff
@@ -0,0 +1,31 @@
...
+    wget -q https://repo.continuum.io/miniconda/Miniconda3-latest-Linux-x86_64.sh -O miniconda.sh;
+  elif [[ "$TRAVIS_OS_NAME" == "osx" ]]; then
```
HealthyPear, line 13: here

.github/install.sh
```diff
@@ -0,0 +1,31 @@
...
+  source $HOME/miniconda/etc/profile.d/conda.sh
+
+  hash -r
+  conda config --set always_yes yes --set changeps1 no
+  conda update -q conda  # get latest conda version
+  # Useful for debugging any issues with conda
+  conda info -a
+
+  sed -i -e "s/- python=.*/- python=$PYTHON_VERSION/g" environment.yml
+  conda env create -n travis --file environment.yml
```
HealthyPear, line 27: here

kosack:
> Where do you mean?

I thing envvars like TRAVIS_OS_NAME

.github/workflows/ci.yml
```diff
@@ -0,0 +1,30 @@
...
+      - uses: actions/checkout@v2
+      - name: Set up Python ${{ matrix.python-version }}
+        uses: actions/setup-python@v2
+        with:
+          python-version: ${{ matrix.python-version }}
+      - name: Install dependencies
+        run: |
+          source .github/install.sh
+          python --version
+          pip install travis-sphinx codecov pytest-cov
```
HealthyPear, line 23: here

maxnoe:
Hey, I am not finished here, I just started ;)

HealthyPear:
> Hey, I am not finished here, I just started ;)

ah ok sorry! my bad

.github/workflows/ci.yml
kosack: ah, yes, we might have to use something other than travis-sphinx to deploy docs.  It's not really needed, if we add a seperate github action for doc deployment though.  It was only because travis didn't have graphs of actions.  Basically that action should only run if this is a merge to master commit and all CI actions pass

HealthyPear: do you know if we can also build the PR docs like on readthedocs?
I find that tool useful to check the docs without having to build them myself

maxnoe: there is a sphinx action for github actions that even does rest linting: https://github.com/ammaraskar/sphinx-action

maxnoe: done

maxnoe (commented):
Ok, this works now. Will squash.

Let's go ahead and merge and figure out how to deploy the docs and pypi in another PR.

maxnoe:
Also added pypi deploy and docs deploy now

kosack (approved):

HealthyPear:
2 questions:
- could we deploy also to conda from here?
- is it possible to build also the PR version of the docs like on readthedocs?

maxnoe:
> could we deploy also to conda from here?

We moved to conda forge, the process is largely automatized after a pypi release has been made (A bot opens a PR with the changes proposed to build the new conda package)

> is it possible to build also the PR version of the docs like on readthedocs?

The docs are build for pull requests. I don't understand the question about readthedocs.

kosack:
> could we deploy also to conda from here?

we're trying to move to conda-forge, so that would solve that.

> is it possible to build also the PR version of the docs like on readthedocs?

The problem is not generating the docs (they are actuallly generated for every PR, even in Travis), but hosting them on a webserver.  Github-pages is "dumb" in the sense that it just displays things that are dropped into it.  ReadTheDocs is smarter and has mechanisms for switching between versions of the docs to display, which we would have to implement ourselves (so it's non-trivial to support that).  Should probably look into how GammaPy is doing it.

HealthyPear:
> We moved to conda forge, the process is largely automatized after a pypi release has been made (A bot opens a PR with the changes proposed to build the new conda package)

OK

> The docs are build for pull requests, they are only deployed for pushes to master. I don't understand the question about readthedocs.
> 
> See that this PR has a `docs` check?

sorry I meant to see them

when I use readthedocs I have link in the check for peeking at the built docs before merging, so I don't have to checkout the branch or the PR and build them again

Of course this is not necessary, it's just a nice feature to make it easier to check the docs

HealthyPear:
> The problem is not generating the docs (they are actuallly generated for every PR, even in Travis), but hosting them on a webserver. Github-pages is "dumb" in the sense that it just displays things that are dropped into it. ReadTheDocs is smarter and has mechanisms for switching between versions of the docs to display, which we would have to implement ourselves (so it's non-trivial to support that). Should probably look into how GammaPy is doing it. But even there, having docs per PR would not be easy/possible. Per version would be more interesting, but also not trivial.

Sorry I didn't see this!
I guess this answers my question, then.

HealthyPear (approved):

kosack (approved):

adonath:
@kosack The current Gammapy docs setup is here: https://github.com/gammapy/gammapy-docs

Our problem was, that we ran out of resources to build our docs on `readthedocs`. With all the tutorials and example plots etc. our docs build currently takes ~20 min. So we have moved to the following solution:
- Docs are deployed in a separate repo: https://github.com/gammapy/gammapy-docs
- There is one `dev` folder and a new folder for each version
- In a sense we mis-use the repo for file hosting. It works but, is non-ideal, as the size of the repo grows (currently it's ~2 GB). I guess at some point we possibly run into the storage limits imposed by GitHub.  However one could probably just delete the git history or change to a different hosting services...
- The version docs are currently build and uploaded manually as part of the release process. As long as releases are not done that often this is acceptable.
- For the "dev" docs we have a GitHub action, that can be manually triggered. Of course one could trigger on each PR, but typically it's sufficient to build the docs only for relevant PRs, or before developer meetings etc.
- We don't have a nice switching panel for versions as `readthedocs`, but users are notified (e.g. https://docs.gammapy.org/0.17/), if they don't watch the latest version. To switch between the versions one has to manually update the URL (e.g. https://docs.gammapy.org/dev/). This works, however we found that as we introduced the change, many users were not aware of it and looked at a different version of the docs than expected. The version is shown in the upper left but it was not present enough...


All in all the setup is not bad, but should be improved / automatised in some regards.

PR #1546, opened 2020-11-22
Fix DL1 EventSource

### LINKED ISSUE #1545: indexing bug in DL1EventSource
The code here

https://github.com/cta-observatory/ctapipe/blob/a5ad0f181a2f1565324ce6af5c18a806c9b33dd5/ctapipe/io/dl1eventsource.py#L281-L283

adds 1 to the tel index to get the tel_id, but this is not correct in general, particularly if the subarray is a sub-selection of a larger array.   E.g. if we subselect only telescopes `[5,7,9,100]`, the tel_id to index map would be `[5: 0, 1: 7, 2: 9, 3:100]`, which is *not* `tel_index+1`

Therefore this should be changed to use the functions provided by `SubarrayDescription` to take care of this mapping, namely `sub.tel_ids_to_indices(tel_ids)` or `sub.tel_ids_to_mask(tel_ids)`

Example:
```py3
>>> sub = source.subarray.select_subarray("NewSubarray", [1,5,7,10])
>>> sub.tel_ids_to_mask([1,5,7])

array([ True,  True,  True, False])

>>> sub.tel_ids_to_indices([1,5,7,10])

array([0, 1, 2, 3])
```

There is no guarantee that `tel_id - 1 == tel_index`.

LukasNickel:
Small changes to tools and bugfixes in order to work with the DL1EventSource

- Fix #1545 adding a new method to `SubarrayDescription`
- Fix event generation in the `DL1EventSource`: Every second event was skipped, because the iterator was advanced twice. That's a huge bug if it didn't work for some random other reason before
- Rename `dl1eventsource.mc_headers` to `dl1eventsource.simulation_config` in line with the changes to the `SimtelEventSource` and container structure changes. To make this behave the same way the `SimtelEventSource` does when using e.g. the `stage-1` tool, this returns a single container if the file contains only one simulation run. This makes it somewhat inconsistent for merged files, but I think there a a lot of things to clean up anyway for us to fully support merged dl1 files consistently.
- Add tests using the stage1 and muon reconstruction tools on dl1 files


ctapipe/tools/tests/test_tools.py
```diff
@@ -19,10 +21,32 @@
 from pathlib import Path
 
 
+d = tempfile.TemporaryDirectory()
 GAMMA_TEST_LARGE = get_dataset_path("gamma_test_large.simtel.gz")
 LST_MUONS = get_dataset_path("lst_muons.simtel.zst")
 
 
+@pytest.fixture
```
kosack, line 29: Might want to mark all of these new fixtures as `scope='module'` at least, or higher, since they get re-run every time you use the fixture in the default scope ('function'), which is very slow.

ctapipe/tools/tests/test_tools.py
```diff
@@ -252,6 +277,75 @@ def test_stage_1(tmpdir):
             assert "peak_time" in dl1_image.dtype.names
 
 
+def test_stage_1_dl1(tmpdir, dl1_image_file, dl1_parameters_file):
+    from ctapipe.tools.stage1 import Stage1Tool
+
+    config = Path("./examples/stage1_config.json").absolute()
+    # DL1A file as input
+    with tempfile.NamedTemporaryFile(suffix=".hdf5") as f:
```
kosack, line 285: you should not use tempfile.NamedTemporaryFile here, it's better to just use the tmp_dir fixture you are using (the other tests also misuse this a lot as well I notice).  
That ensures the files are written to the temporary directory defined by the *test system*, not some other place (which is confusing when debugging).   Also you don't need to open the files as a filehandle, as NamedTemporaryFile does, only generate a name.  Note that tmpdir is unique for all test runs, so the actual filename does not need to be random at all, you can pick something easy to read like "test_dl1a_file.hdf5":

For example, just do:
```py3
filename = tmp_dir /  "test_dl1a_file.hdf5"
```
Then later just use `f"some string {filename}"` (not `f.name`)

No "with" or anything else is neeeded.  Then

kosack (changes requested):

ctapipe/tools/tests/test_tools.py
LukasNickel: Didn't know of that, thank you

ctapipe/tools/tests/test_tools.py
LukasNickel: That is also much shorter
I copied that from other tests. so its probably the best if I fix them as well.

ctapipe/instrument/subarray.py
```diff
@@ -177,6 +177,23 @@ def tel_ids_to_mask(self, tel_ids):
...
+        ----------
+        tel_mask: array-like
+            Boolean array of length ``num_tels`` with indices of the
+            telescopes in ``tel_ids`` set to True.
+        Returns
+        -------
+        np.array:
+            Array of selected tel_ids
+        """
+        indices = np.where(tel_mask)[0]
```
maxnoe, line 194: The where should not be needed, right? You can directly index `tel_ids` with `tel_mask`

ctapipe/tools/tests/test_tools.py
```diff
@@ -121,22 +124,32 @@ def test_merge(tmpdir):
...
+        == 0
+    )
+
     assert (
         run_tool(
             MergeTool(),
             argv=[
-                f"{input_file_1}",
-                f"{input_file_2}",
+                f"{in_1}",
```
maxnoe: unnecessary fstring, since `in_1` is already a string.

LukasNickel: "He added an unneccesary f-string (Wasn't me)"

ctapipe/io/dl1eventsource.py
```diff
@@ -153,21 +154,27 @@ def datalevels(self):
...
 
     @property
-    def mc_headers(self):
-        return self._mc_headers
+    def simulation_config(self):
+        """
+        Returns the simulation config.
+        In case of a merged file, this will be a list of simulation configs.
+        """
+        if len(self._simulation_configs) == 1:
```
maxnoe, line 167: I would not special case this. Just always return a list or better maybe `dict(obs_id -> config)`?

LukasNickel: I wanted this to behave the same way as the SimtelEventSource for unmerged files.
But we can change it over there as well of course

maxnoe: I think simtel files can also contain multiple corsika runs...

kosack:
This is marked for 0.10.1, but also says "API change". I see only that the mc_header attribute changed to simulation_config (as it should, this was missed in the previous release).  I think that is less of an  API change than a API bugfix, since the API for EventSources already had that attribute (it was just missed in the last version of DL1EventSource).  Are there other larger API changes here? Or can we remove that tag, so it can go in 0.10.1 vs 0.11.0?

LukasNickel:
> This is marked for 0.10.1, but also says "API change". I see only that the mc_header attribute changed to simulation_config (as it should, this was missed in the previous release). I think that is less of an API change than a API bugfix, since the API for EventSources already had that attribute. Are there other larger API changes here? Or can we remove that tag, so it can go in 0.10.1 vs 0.11.0?

I think thats all, so API bugfix might be more accurate

ctapipe/io/dl1eventsource.py
LukasNickel: Then it becomes an api change for sure though as other EventSources change as well

maxnoe: Ok, than I'd say leave it as is.

ctapipe/tools/muon_reconstruction.py
```diff
@@ -98,7 +98,7 @@ def setup(self):
                 "Outputfile {self.output} already exists, use `--overwrite` to overwrite"
             )
 
-        self.source = EventSource(parent=self)
+        self.source = EventSource.from_config(parent=self)
```
maxnoe, line 101: Why this change? The current version should work, there should be no need for `from_config` or `from_url` anymore anywhere.

ctapipe/tools/stage1.py
```diff
@@ -77,12 +77,17 @@ class Stage1Tool(Tool):
     def setup(self):
 
         # setup components:
-        self.event_source = EventSource(parent=self)
+        self.event_source = EventSource.from_config(parent=self)
```
maxnoe, line 80: Same as above, please don't just change things that are supposed to work without creating an issue if they don't

ctapipe/tools/stage1.py
```diff
@@ -77,12 +77,17 @@ class Stage1Tool(Tool):
...
 
         datalevels = self.event_source.datalevels
-        if DataLevel.R1 not in datalevels and DataLevel.DL0 not in datalevels:
+        if (
+            DataLevel.R1 not in datalevels
```
maxnoe, line 84: This is the same but different from the other tool. Create a small method on `EventSource`? E.g. `EventSource.has_datalevel(any_of={...})`

ctapipe/tools/display_dl1.py
```diff
@@ -167,10 +167,14 @@ def __init__(self, **kwargs):
 
     def setup(self):
         self.eventsource = EventSource.from_config(parent=self)
-        if not (
-            set([DataLevel.R0, DataLevel.R1, DataLevel.DL0, DataLevel.DL1_IMAGES])
-            & set(self.eventsource.datalevels)
-        ):
+        compatible_datalevels = [
+            DataLevel.R0,
```
maxnoe, line 171: R0?

kosack (changes requested):
Looks good.  

One minor comment: the `self.datalevels` property is accessed several times in the event loop, so it may be better either as a lazyproperty, or better yet to generate the values in the constructor and just return `self._datalevels` in the property.  The latter may be preferred as right now if you call `close()`, accessing the datalevels property causes an exception since _h5file is no longer open. 

Otherwise, once the conflicts are fixed, this looks ready.

maxnoe (approved):

kosack (approved):

PR #1560, opened 2020-12-03
Fix handling of telescope parameters from cli for string values, fixes #1559

### LINKED ISSUE #1559: Passing TelescopeParameters on the command line does not work
I am investigating, but currently passing telescope components on the command line somehow directly lands in the `TelescopeParamter.trait.validate`, not in `TelescopeParameter.validate`

maxnoe:
I could simplify the telescope parameter a bit and fix the failure for string-valued (e.g. Path) telescope parameters that arised due to how traitlets parses cli options internally.

### DIFF
--- a/ctapipe/core/tests/test_traits.py
+++ b/ctapipe/core/tests/test_traits.py
@@ -433,6 +433,33 @@ def test_telescope_parameter_to_config(mock_subarray):
     assert config["SomeComponent"]["tel_param1"] == [("type", "*", 6.0)]
 
 
+def test_telescope_parameter_from_cli(mock_subarray):
+    """
+    Test we can pass single default for telescope components via cli
+    see #1559
+    """
+
+    from ctapipe.core import Tool, run_tool
+
+    class SomeComponent(TelescopeComponent):
+        path = TelescopeParameter(Path(), default_value=None).tag(config=True)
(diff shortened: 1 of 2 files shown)


kosack (approved):

ctapipe/core/tests/test_traits.py
```diff
@@ -433,6 +433,33 @@ class SomeComponent(TelescopeComponent):
...
+    class TelescopeTool(Tool):
+        def setup(self):
+            self.comp = SomeComponent(subarray=mock_subarray, parent=self)
+
+    tool = TelescopeTool()
+    run_tool(tool)
+    assert tool.comp.path == [("type", "*", None)]
+    assert tool.comp.val == [("type", "*", 1.0)]
+
+    tool = TelescopeTool()
```
HealthyPear, line 457: why the re-instantiation of the tool?
wouldn't the new run overwrite previous traits?

HealthyPear (approved):

ctapipe/core/tests/test_traits.py
maxnoe: I want to be sure that I have a fresh instance and the run before does not interfere. This is about testing the cli and when using Tools, you get a fresh one with each run.

PR #1567, opened 2020-12-09
Fix 1566, telescope parameter cli

maxnoe:
I could not find a better solution, it seems hackish.

But the whole `TelescopeComponent` thing is quite bolted onto traitlets.

### DIFF
--- a/ctapipe/core/traits.py
+++ b/ctapipe/core/traits.py
@@ -6,6 +6,7 @@ from astropy.time import Time
 import pathlib
 from urllib.parse import urlparse
 import os
+from collections.abc import Sequence
 
 from traitlets import (
     Bool,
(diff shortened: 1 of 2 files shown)


ctapipe/core/traits.py
```diff
@@ -422,13 +423,25 @@ def validate(self, obj, value):
         return normalized_value
 
     def set(self, obj, value):
+        # fix for issue #1566.
+        # If a component is in Tool.classes, then a custom argparse parser
+        # gets created that appends the single element to
```
kosack, line 428: to what?

kosack (dismissed):
Seems ok - it's not so important to treat telescope parameters on the command-line, but it's nice if they work in the odd case somebody wants to do it.

maxnoe:
> Seems ok - it's not so important to treat telescope parameters on the command-line, but it's nice if they work in the odd case somebody wants to do it.

The problem arises when wants to set the gloabl default as well.
This is for example about giving the global default to `--CameraCalibrator.apply_waveform_time_shift = False`

maxnoe:
Ok, I think the main problem is the split into two clases. `TelescopeParameter` and `TelescopePatternList` where it should just be one thing!

This messes with how traitlets reads this in. Let me see if I can get rid of this.

maxnoe:
Ok, I remembered why the patternlist is necessary. But I moved the fix to a position where it makes more sense.

kosack (approved):
I see you re-discovered how much of a pain this was to get to work correctly the first time! It took me a while to realize it needed 2 classes, with a custom "klass".  The change looks fine, and doing the conversion from scalar to list in the constructor of TelescopeParameterList is indeed cleaner.

HealthyPear (approved):

Issue #1450, opened 2020-10-09
CameraReadout.from_file() should fail if no camerareadout exists


kosack:
This caused a lot of confusion, and means most of the tests are not correct: 

From readout.py:
```py3
        except FileNotFoundError:
            # TODO: remove case when files have been generated
            logger.warning(
                f"Resorting to default CameraReadout,"
                f" File does not exist: ({tabname})"
            )
            reference_pulse_shape = np.array([norm.pdf(np.arange(96), 48, 6)])
            return cls(
                camera_name=camera_name,
                sampling_rate=u.Quantity(1, u.GHz),
                reference_pulse_shape=reference_pulse_shape,
                reference_pulse_sample_width=u.Quantity(1, u.ns),
            )
```

So if you create a camera like:
```
camera = CameraDescription.from_name("LST")
```

You get a crazy reference pulse shape that doesn't reflect reality (it's 10x too wide), and this is used in tests internally.   I think the correct behavior would be to fail with an error, and we should add correct `*.camerareadout.fits.gz` files in ctapipe-extra so that doesn't happen.

![image](https://user-images.githubusercontent.com/11677812/95596732-0dc9fc00-0a4e-11eb-9198-b3b8b0bf3d17.png)

kosack:
Of course, we don't have camera_readout defs for some of the test cameras (HESS, Whipple-X, etc).  So perhaps instead of simply failing, it should just not create the CameraReadout (leave it as None), that way tests that do not need timing will still work, but those that do not should fail on cameras with no timing info

kosack:
As a quick fix, we should just add `*.camreadout.fits.gz` files to ctapipe-extra.  I opened #1451 to facilitate that.

Issue #1400, opened 2020-07-23
create one instance of the tests with `NUMBA_DISABLE_JIT=1` for the coverage report?


dneise:
Side point: numba jitted functions do not contribute to coverage, we might want to create one instance of the tests with `NUMBA_DISABLE_JIT=1` for the coverage report.

_Originally posted by @MaxNoe in https://github.com/cta-observatory/ctapipe/pull/1398#issuecomment-662900445_

maxnoe:
Unfortunately, as e.g. guvectorize functions don't work in pure python mode.

How to handle coverage with numba is an open issue with them, and I fear we just need to wait until something is implemented in numba itself.

PR #1568, opened 2020-12-29
signal extractor SlidingWindowMaxSum 

jsitarek:
a new simple signal extractor, slightly slower, but with better accuracy (in particular for weak pulses): SlidingWindowMaxSum

It maximizes the sum on "width" consecutive slices

some speed test using 3 trials of 1000 events of LST1 data:
only r0==> r1 calibration: 14.575s, 14.465s, 14.550s
LocalPeakWindowSum (current extractor) 19.853s, 20.188s, 20.698s
MaxWindowSum (new code): 21.731s, 20.813s, 21.132s

one feature can be improved, namely the correction for the signal outside of the integration window, the current code is reusing LocalPeakWindowSum approach assuming that the shift is half of the total window, which is correct only if the pulse is symmetric (which is not really the case)

### DIFF
--- a/ctapipe/image/extractor.py
+++ b/ctapipe/image/extractor.py
@@ -8,10 +8,12 @@ __all__ = [
     "FixedWindowSum",
     "GlobalPeakWindowSum",
     "LocalPeakWindowSum",
+    "SlidingWindowMaxSum",
     "NeighborPeakWindowSum",
     "BaselineSubtractedNeighborPeakWindowSum",
     "TwoPassWindowSum",
     "extract_around_peak",
+    "extract_sliding_window",
     "neighbor_average_waveform",
     "subtract_baseline",
     "integration_correction",
@@ -118,7 +120,77 @@ def extract_around_peak(
     peak_time[0] /= sampling_rate_ghz
     sum_[0] = i_sum
 
+@guvectorize(
+    [
+        (float64[:], int64, float64, float32[:], float32[:]),
+        (float32[:], int64, float64, float32[:], float32[:]),
+    ],
+    "(s),(),()->(),()",
+    nopython=True,
+)
+def extract_sliding_window(
+    waveforms, width, sampling_rate_ghz, sum_, peak_time
+):
+    """
+    This function performs the following operations:
 
+    - Find the largest sum of width consecutive slices 
+    - Obtain the pulse time within a window defined by a peak finding
+    algorithm, using the weighted average of the samples.
+
+    This function is a numpy universal function which defines the operation
+    applied on the waveform for every channel and pixel. Therefore in the
+    code body of this function:
+        - waveforms is a 1D array of size n_samples.
+        - width is integer
+
+    The ret argument is required by numpy to create the numpy array which is
+    returned. It can be ignored when calling this function.
+
+    Parameters
+    ----------
+    waveforms : ndarray
+        Waveforms stored in a numpy array.
+        Shape: (n_pix, n_samples)
+    width : ndarray or int
+        Window size of integration window for each pixel.
+    sampling_rate_ghz : float
+        Sampling rate of the camera, in units of GHz
+        Astropy units should have to_value('GHz') applied before being passed
+    sum_ : ndarray
+        Return argument for ufunc (ignore)
+        Returns the sum of the waveform samples
+    peak_time : ndarray
+        Return argument for ufunc (ignore)
+        Returns the peak_time in units "ns"
+
+    Returns
+    -------
+    charge : ndarray
+        Extracted charge.
+        Shape: (n_pix)
+
+    """
+
+    # first find the cumulative waveform
+    cwf=np.cumsum(waveforms) 
+    cwf=np.concatenate((np.zeros(1), cwf)) # add zero at the begining so it is easier to substract the two arrays later
+    sums=cwf[width:]-cwf[:-width]
+    maxpos=np.argmax(sums) # start of the window with largest sum
+    sum_[0]=sums[maxpos]
+
+    time_num = float64(0.0)
+    time_den = float64(0.0)
+    # now compute the timing as the average of non negative slices
+    for isample in prange(maxpos, maxpos+width):
+        if waveforms[isample] > 0:
+            time_num += waveforms[isample] * isample
+            time_den += waveforms[isample]
+
+    peak_time[0] = time_num / time_den if time_den > 0 else maxpos+0.5*width
+    # Convert to units of ns
+    peak_time[0] /= sampling_rate_ghz
+    
 @njit(parallel=True, cache=True)
 def neighbor_average_waveform(waveforms, neighbors_indices, neighbors_indptr, lwt):
     """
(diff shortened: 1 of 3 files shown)


jsitarek:
I made a few iterations solving things pointed by codacy and coverage checks. 
codacy reports a missing argument in one of the functions, but I think this is just an issue of the paralelization, since the same type of code as in extract_around_peak is used. 
coverage check claims that most of the extract_sliding_window function is not tested, however this function is explicitly tested in test_extractor.py.

I think the code is ready for the review,

kosack (approved):
Looks good to me. 

In fact, I think a similar implementation could be used to speed up the TwoPass "MARS-like" method (which also uses a sliding  window in the first pass).

kosack:
By the way, for the Numba code coverage issue, see #1400

ctapipe/image/extractor.py
```diff
@@ -524,6 +597,66 @@ def __call__(self, waveforms, telid, selected_gain_channel):
...
+    @lru_cache(maxsize=128)
+    def _calculate_correction(self, telid):
+        """
+        Calculate the correction for the extracted charge such that the value
+        returned would equal 1 for a noise-less unit pulse.
+
+        This method is decorated with @lru_cache to ensure it is only
+        calculated once per telescope.
+
+        WARNING: TO BE DONE properly, the current code reuses the function of
```
maxnoe, line 622: Could you implement this directly here, does not sound to complicated?

jsitarek: the main reason why I did not do so is because this feature does not seem to be used (at least in LST), so I did not have a proper set-up to test it, but I can look into making some dummy pulse shape and testing on it.

maxnoe: that would be great

kosack: You have the reference pulse shape in the `CameraDescription`. I guess it's a fairly small effect though, and the correction doesn't really matter much except to get the cleaning thresholds in the same units for all cameras.

jsitarek: Sorry for the delay, I restarted working on this. 
CameraDescription.readout is where the code is taking the pulse shape from. However this is not really reliable. 
If I execute the following code 
`import numpy as np
import astropy.units as u
plt.ion()

from ctapipe.instrument import SubarrayDescription, TelescopeDescription

subarray = SubarrayDescription(
        "LST1",
        tel_positions={1: np.zeros(3) * u.m},
        tel_descriptions={
            1: TelescopeDescription.from_name(
                optics_name="LST", camera_name="LSTCam"
            ),
        },
)
pulse_shape=subarray.tel[1].camera.readout.reference_pulse_shape[ dt=subarray.tel[1].camera.readout.reference_pulse_sample_width
xs=np.arange(len(pulse_shape))*dt
plt.plot(xs, pulse_shape)`
I get the following figure:
pulse_shape_LST
which is a much broader pulse then it should 

The calculation of the correction factor would be much simpler if the pulse shape in this class had the same binning as the actual readout, this is the case in the above example, and one would assume to take it from granted since the shape is taken from the "readout" object, which has the binning embedded, however in the first tests that i was doing in lstchain, when the array was being read from the data the pulse shapes there were actually a delta function with a SSC-like sampling, so obviously  it cannot be taken for granted. 

I will change the code to use a simple conversion and rounding of sampling to make it work also in this more general case, but the whole issue of the LST pulse shape deserves a separate "issue"

kosack: The `from_name()` methods load up a file from ctapipe-extra (which is now a directory on a server rather than a package), and are just meant for unit-testing purposes.  Currently everything in ctapipe-extra is from PROD3 or even PROD2 simulations, so quite out of date for real analysis.  In the future I want to clean that up and have an option to select which "prod" to use, but there has not been manpower for that (see e.g #738 )

If you load real data from a SimTel file or something else supported, the correct waveform should be loaded into the instrument model that you get from `source.subarray`.

kosack: Note taht if you don't want to always load up a file, you can take any reference prod5 file, for example, and run 
```sh
ctapipe-dump-instrument --input=some_reference_file.simtel.gz
```
And it will dump a bunch of FITS files including the Camera geometry and readout definitions to the local directory.  You can then setenv `CTAPIPE_SVC_PATH=[directory where those files are]`, and ctapipe will use that when you run the `from_name()` functions instead of the defaults.

kosack: One more thing: be careful how you plot it - the reference pulse shape in the MCs are *oversampled*, so **not** at the same sampling rate as the waveform! I can't tell in your plot where you get dt from, but I suspect you are using the dt from the waveform, not from the reference pulse.

kosack: Update: I found the actual problem... In fact, the camera readout definition for LSTCam does not even exist in the ctapipe-extra directory on the dataserver. It seems the default behavior is to just return some dummy pulse shape, which has no real meaning (you should see a logger warning message if logging is set up)...   Clearly this is not good behavior. 

here is a comparison with prod3b:
image

kosack: This is also why you get only 1 reference pulse shape if you use `from_name()`... 
I had already opened an issue about this, but obviously forgot: See #1450

kosack:
I opened a PR with at least a temporary fix to the pulse shape problem.  
Until that is accepted, you can also do the following as a hack while testing:
```py3
from ctapipe.utils import datasets
datasets.DEFAULT_URL = "http://cccta-dataserver.in2p3.fr/data/ctapipe-extra/v0.3.2/"
```

jsitarek:
thanks a lot this is really helpful

jsitarek:
I made this correction properly, tested it using the test_extractor.py, and also made an extra test (using the pulse shapes from @kosack suggestion),  so I made the commit, which as you can see however fails:

ctapipe/image/tests/test_concentration.py::test_concentration FAILED     [ 27%]
I've run the same test in my machine, but it goes through.

I've run 
git pull upstream master
to check if the current changes are in conflict with some other changes made in the meantime (I doubt it because I only changed the code of the new extractor which is not yet used anywhere) 

and still test_concentration.py works in my PC.
I checked the other tests and 
tests/test_reducer.py::test_tailcuts_data_volume_reducer FAILED 
also fails in my PC now, however I do not see how this has anything to do with my commit, since a different extractor is used in that test

I suspect that there might have been some other commit done in the meantime that broke those tests.
How should we proceed ?

kosack:
> I made this correction properly, tested it using the test_extractor.py, and also made an extra test (using the pulse shapes from @kosack suggestion), so I made the commit, which as you can see however fails:

See #1588, I have exactly the same problem.  Not sure what the solution is - so far I don't see why it is happening, except for a small change in pixel area (which i still don't understand, as the pixel distances are the same, and I compared the computation to past versions of ctapipe, and it is identical)

ctapipe/image/tests/test_sliding_window_correction.py
```diff
@@ -0,0 +1,52 @@
...
+import astropy.units as u
+from numpy.testing import assert_allclose
+
+from ctapipe.image.extractor import SlidingWindowMaxSum
+from ctapipe.image.toymodel import WaveformModel
+from ctapipe.instrument import SubarrayDescription, TelescopeDescription
+
+from ctapipe.utils import datasets
+
+datasets.DEFAULT_URL = "http://cccta-dataserver.in2p3.fr/data/ctapipe-extra/v0.3.2/"
```
kosack, line 11: Shouldn't really set this globally, since it affects all other tests (and definitely should not be committed when we merge the PR).   Perhaps for now, you might want to use the `monkeypatch` test fixture instead, something like:

```py3
def test_xxx(monkeypatch):
    monkeypatch.setattr(datasets, "DEFAULT_URL", "http://cccta-dataserver.in2p3.fr/data/ctapipe-extra/v0.3.2/")
   # the rest of the test

Comment (jsitarek) [reply ctapipe/image/tests/test_sliding_window_correction.py]: hmm, I put this test explicitely into a separate file to avoid affecting other files with a global variable, because I thought that each file is run independently.
either way, now modified as suggested by you

**Comment by jsitarek** [comment]
now that the other problem with tests is fixed, I'm looking into the issues pointed out by codacy
one is: 

Instance of 'int' has no 'tel' member
643 self.window_width.tel[telid]
the issue makes sense because window_width should be a plain int, however I saw that all the _calculate_correction functions use this, so it is intended, or should I changed for SlidingWindowMaxSum extractor as well as for the other extractors ?

**Comment by jsitarek** [comment]
the failing tests were solved in the other PR, I did some small updates to solve the codacy issues, there are three left:


Instance of 'int' has no 'tel' member
self.window_width.tel[telid]


No value for argument 'sum_' in function call
charge, peak_time = extract_sliding_window(


Instance of 'int' has no 'tel' member
waveforms, self.window_width.tel[telid], self.sampling_rate[telid]

1st and 3rd are somehow strange because the window_width in fact is not int but IntTelescopeParameter, but investigating it I corrected how the width of the integration window was changed in the test file. 
2nd one is I guess also not a problem, but a feature of guvectorize

so, can we merge this PR?

**Comment by jsitarek** [comment]
Hi @maxnoe @kosack 
Please let me know if you want any additional modifications in this PR, or if you can give it already green light for merging

File: ctapipe/image/tests/test_sliding_window_correction.py [hunk 3]
```diff
@@ -0,0 +1,69 @@
...
+from ctapipe.utils import datasets
+
+
+def test_sw_pulse_lst(monkeypatch):
+    """
+    Test function of sliding window extractor for LST camera pulse shape with
+the correction for the integration window completeness
+    """
+
+    with monkeypatch.context() as monkey_patch:
```
maxnoe, line 24: This can be removed now that the other PR is merged.

To update: 
```
git fetch upstream
git merge upstream/master
```

jsitarek: @maxnoe done

kosack (dismissed):

maxnoe (dismissed):

ctapipe/image/tests/test_sliding_window_correction.py
```diff
@@ -0,0 +1,58 @@
...
+
+from ctapipe.image.extractor import SlidingWindowMaxSum, ImageExtractor
+from ctapipe.image.toymodel import WaveformModel
+from ctapipe.instrument import SubarrayDescription, TelescopeDescription
+
+
+def test_sw_pulse_lst():
+    """
+    Test function of sliding window extractor for LST camera pulse shape with
+the correction for the integration window completeness
```
maxnoe, line 19: wrong indentation here

maxnoe:
This PR should not be merged before 0.10.2 is released.

maxnoe (approved):

ctapipe/image/tests/test_sliding_window_correction.py
jsitarek: corrected, I'm not sure why precommit and flake8 did not catch this.
thx @maxnoe for reapproval
@kosack I also need it from you
and then can one of you merge the PR? I do not have permissions for doing it myself.

maxnoe: Yes, will merge as soon as both approvals are there

maxnoe: It is valid code and since it's a test it is probably also not checked by the documentation build

kosack (approved):

PR #1572, opened 2021-01-11
Change definition of concentration_cog, fixes #1571

### LINKED ISSUE #1571: concentration_cog might not refer to cog when only cleaning pixels are provided.
We use 

https://github.com/cta-observatory/ctapipe/blob/f1cd0bb3079bc4d8657bc3917aff09940f8f39ef/ctapipe/image/concentration.py#L32-L33

So taking the three pixels closest to the cog. However, when the pixels below the actual cog are not part of the cleaning pixels and only the cleaning pixels are passed in, this won't work (as happens likely for e.g. muon rings).

We should replace with something like the distance in term of pixel radius, which also will be faster than argsorting.

maxnoe:
This changes the definition of concentration_cog from
"the three closest pixels to the cog" to
"pixel closer to the cog than the pixel diameter".

This has two advantages:
* no need for argsort: 50 % faster
* works if a masked geometry / image is passed as expected, before
  the three closest pixels to the cog could be very far away from the
  cog, e.g. for muon rings.

### DIFF
--- a/ctapipe/containers.py
+++ b/ctapipe/containers.py
@@ -153,7 +153,7 @@ class ConcentrationContainer(Container):
 
     container_prefix = "concentration"
     cog = Field(
-        nan, "Percentage of photo-electrons in the three pixels closest to the cog"
+        nan, "Percentage of photo-electrons inside one pixel diameter of the cog"
     )
     core = Field(nan, "Percentage of photo-electrons inside the hillas ellipse")
     pixel = Field(nan, "Percentage of photo-electrons in the brightest pixel")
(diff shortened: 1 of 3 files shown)


kosack (changes requested):
Need to update the help string in `ctapipe.containers.ConcentrationContainer` as well, since it still says "Percentage of photo-electrons in the three pixels closest to the cog"

maxnoe:
@kosack done

kosack (approved):

HealthyPear (approved):

PR #1580, opened 2021-01-26
Traverse classes attribute in `classes_with_traits` to enable more complete `--help-all`

maxnoe:
Note: traitlets itself does not use `List` but `list` for `classes`, I don't know why we started with that.

This results in `--help-all` to basically show the help for, well, `all`. Including for example the configurable items in the plugin components.

### DIFF
--- a/ctapipe/core/tests/test_tool.py
+++ b/ctapipe/core/tests/test_tool.py
@@ -2,7 +2,7 @@ import os
 import logging
 import tempfile
 import pytest
-from traitlets import Float, TraitError, List, Dict, Int
+from traitlets import Float, TraitError, Dict, Int
 from traitlets.config import Config
 from pathlib import Path
 
@@ -190,7 +190,7 @@ def test_tool_command_line_precedence():
         description = "test"
         userparam = Float(5.0, help="parameter").tag(config=True)
 
-        classes = List([SubComponent])
+        classes = [SubComponent]
         aliases = Dict({"component_param": "SubComponent.component_param"})
 
         def setup(self):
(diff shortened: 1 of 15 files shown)


kosack:
I think it comes from here (the only example of using Application classes included in  Traitlets):

https://github.com/ipython/traitlets/blob/2bb2597224ca5ae485761781b11c06141770f110/examples/myapp.py#L60

```py3
class MyApp(Application):

    name = Unicode('myapp')
    running = Bool(False, 
                   help="Is the app running?").tag(config=True)
    classes = List([Bar, Foo])
    config_file = Unicode('',
                   help="Load this config file").tag(config=True)
    
    aliases = Dict(dict(i='Foo.i',j='Foo.j',name='Foo.name', running='MyApp.running',
                        enabled='Bar.enabled', log_level='MyApp.log_level'))
    
    flags = Dict(dict(enable=({'Bar': {'enabled' : True}}, "Enable Bar"),
                  disable=({'Bar': {'enabled' : False}}, "Disable Bar"),
                  debug=({'MyApp':{'log_level':10}}, "Set loglevel to DEBUG")
            ))
```

maxnoe:
I will check if there is a solution for supporting `List`.

kosack:
> I will check if there is a solution for supporting List.

To be honsest, I prefer using `list`.  I have no idea why they use Dict and List traitlets for those in the example---it seems like just extra work for no gain.

maxnoe:
@kosack the unit tests of traitlets have `classes = [Foo, Bar]` in them.

I think it is as easy as `classes.default()`

kosack (approved):

LukasNickel (approved):

PR #1583, opened 2021-01-28
Faster muon reconstruction using numba

RuneDominik:
Small changes to utilize numba for faster muon reconstruction. The major part in the ~1/3 decreased time required for processing muon events is gained in the direct implementation of a gaussian cdf, where the math modules erf-function is needed since standard numba is not aware of scipy. Further speedup is gained from decorating the chord_length function. This does, however, require the definition of an input and output layout that restricts the usage of the function to inputs of dimension 1 (which should be the use-case). As a result, the corresponding unit-tests are failing, as these test the function with scalar values.

### DIFF
--- a/ctapipe/image/muon/intensity_fitter.py
+++ b/ctapipe/image/muon/intensity_fitter.py
@@ -262,8 +262,6 @@ def image_prediction_no_units(
     # The weight is the integral of the ring's radial gaussian profile inside the
     # ring's width
     delta = pixel_diameter_rad / 2
-
-    # Define Gaussian cdf for numba usage
     cdfs = gaussian_cdf(
         [radial_dist + delta, radial_dist - delta], radius_rad, ring_width_rad
     )
--- a/ctapipe/image/muon/intensity_fitter.py
+++ b/ctapipe/image/muon/intensity_fitter.py
@@ -9,8 +9,8 @@ To do:
 """
 import numpy as np
 
-from math import sqrt, erf
-from numba import vectorize, guvectorize, double, int32
+from math import erf
+from numba import vectorize, guvectorize, double
 
 from scipy.ndimage.filters import correlate1d
 from iminuit import Minuit
@@ -29,7 +29,8 @@ from ...core.traits import FloatTelescopeParameter, IntTelescopeParameter
 # ratio of the areas of the unit circle and a square of side lengths 2
 CIRCLE_SQUARE_AREA_RATIO = np.pi / 4
 
-@guvectorize([(double, double, double[:], double[:])], '(),(),(n)->(n)')
+
+@guvectorize([(double, double, double[:], double[:])], "(),(),(n)->(n)")
 def chord_length(radius, rho, phi, ret_chord):
     """
     Function for integrating the length of a chord across a circle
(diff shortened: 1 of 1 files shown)


ctapipe/image/muon/intensity_fitter.py
```diff
@@ -194,6 +191,10 @@ def image_prediction(
         max_lambda_m=max_lambda.to_value(u.m),
     )
 
+# Define numba vectorized gaussian cdf
```
maxnoe, line 194: Remove this and add a proper docstring to the function.

See: https://numpydoc.readthedocs.io/en/latest/format.html

ctapipe/image/muon/intensity_fitter.py
```diff
@@ -194,6 +191,10 @@ def image_prediction(
...
+@vectorize([double(double, double, double)])
+def gaussian_cdf(x, mu, sig):
+    return 0.5*(1 + erf((x-mu)/(np.sqrt(2*sig**2))))
```
maxnoe, line 197: Please use the black formatting pre-commit hook, see https://cta-observatory.github.io/ctapipe/getting_started/index.html

ctapipe/image/muon/intensity_fitter.py
```diff
@@ -241,7 +242,9 @@ def image_prediction_no_units(
     # The weight is the integral of the ring's radial gaussian profile inside the
     # ring's width
     delta = pixel_diameter_rad / 2
-    cdfs = norm.cdf(
+
+    # Define Gaussian cdf for numba usage
```
maxnoe, line 246: This comment does not really give more information since the function is aptly called `gaussian_cdf`

maxnoe (changes requested):

RuneDominik:
Included @maxnoe requests and swapped guvectorization to normal vectorization, as it includes scalar inputs without performance loss.

kosack:
The pipelines are failing for another reason, see #1585   They should work once that is merged.

ctapipe/image/muon/intensity_fitter.py
```diff
@@ -195,6 +196,28 @@ def image_prediction(
...
+    cumulative distribution function (cdf)
+
+    Parameters
+    ----------
+    x: float
+        point, at which the cdf should be evaluated
+    mu: float
+        gaussian mean
+    sig: float
+        gaussian standart deviation
```
maxnoe, line 212: typo standard.

maxnoe: Also with `vectorize` all values can also be `ndarrays`

ctapipe/image/muon/intensity_fitter.py
```diff
@@ -195,6 +196,28 @@ def image_prediction(
...
+
+    Returns
+    -------
+    float: cdf-value at x
+    """
+    return 0.5 * (1 + erf((x - mu) / (np.sqrt(2 * sig ** 2))))
```
maxnoe, line 218: sqrt is a pretty slow operation. sig is positive. So maybe we can gain something by defining a constant `sqrt2` and removing the sqrt here?

maxnoe (approved):

kosack (approved):

Issue #1456, opened 2020-10-14
Unit of peak_time


maxnoe:
The peak_time container is currently not defined with a unit.

The extractors convert them to ns.

Shouldn't we add this unit to the container / make it a quantity?

watsonjj:
> Shouldn't we add this unit to the container / make it a quantity?

I've generally been against attaching units to the image related values (#684 #1044). But the `peak_time` description should be updated to indicate its units.

> And related: should be also require any timeshifts be in ns instead of in adc sample width?

I think this might complicate the process, as the shifts are generally extracted in sample units, and need to be in sample units for any corrections involving the waveform (before peak_time extraction)...but of course this depends on what we choose to implement in ctapipe for the timing corrections (#1396).

maxnoe:
I agree that it might make it simpler to use samples for the corrections and the conversion is anyway simple, so we just have to decide.

My feeling is, for consistency, it would be better to have peak_time and time_shift in the same units and convert to samples where needed.

kosack:
I think I prefer having it in `ns` and then it's clear that to go to samples, you have to convert back to samples.  Especially if in the future we do some upsampling of the waveforms, this will avoid confusion. Nanoseconds is then a physical unit, and is common for all telescopes, which is nice (same reason we go to PE, and not digital counts).

However, if instead you want to stick with samples, then as elsewhere, it would be good to add the unit name to the Field name if units are not attached, e.g. rename it to `peak_time_samples`, so later it's clear when we do a conversion to `peak_time_ns`.

kosack:
Oh, one more argument for using nanoseconds: we may/will eventually do more pulse analysis than just the peak time in the future. For example extracting also the pulse width, rise time, etc.   In that case, doing that in samples is needlessly confusing.

watsonjj:
Ok, lets specify in the field descriptions that time_shift and peak_time is in units of time ("ns"). Makes sense from their names. But I still would prefer not attaching astropy units to them.

maxnoe:
Agreed.

moralejo:
> Ok, lets specify in the field descriptions that time_shift and peak_time are in units of time ("ns"). Makes sense from their names. But I still would prefer not attaching astropy units to them.

Why not for the peak times?

maxnoe:
I guess because of the computational cost that comes with astropy units. However, the cost is much less when not actually converting units but just accessing the values in the current unit.

[...]

PR #1586, opened 2021-02-01
Refactor conversions, implement possibility to transform images and peak times to int

maxnoe:
This adds configuration options to the dl1 writer to be able to convert images and peak times using an offset, a scale and a dtype.

Defaults are converting to int32 using a scale of 100 for images and of 10 for peak times (basically 2 (1) digit fixed point decimal representation).

filesize of an DL1 LST file:

```
-rw-r--r-- 1 maxnoe maxnoe 701M  1. Feb 15:15 LST-1.Run02610.0000.dl1.h5
-rw-r--r-- 1 maxnoe maxnoe 292M  1. Feb 15:59 LST-1.Run02610.0000_int.dl1.h5
```

So we get a little more than 50% size reduction.

### DIFF
--- a/ctapipe/io/dl1writer.py
+++ b/ctapipe/io/dl1writer.py
@@ -29,8 +29,9 @@ __all__ = ["DL1Writer", "DL1_DATA_MODEL_VERSION", "write_reference_metadata_head
 
 tables.parameters.NODE_CACHE_SLOTS = 3000  # fixes problem with too many datasets
 
-DL1_DATA_MODEL_VERSION = "v1.0.3"
+DL1_DATA_MODEL_VERSION = "v1.1.0"
 DL1_DATA_MODEL_CHANGE_HISTORY = """
+- v1.1.0: images and peak_times can be stored as scaled integers
 - v1.0.3: true_image dtype changed from float32 to int32
 """
 
--- a/ctapipe/io/dl1writer.py
+++ b/ctapipe/io/dl1writer.py
@@ -18,10 +18,11 @@ from ..containers import (
     TelEventIndexContainer,
 )
 from ..core import Component, Container, Field, Provenance, ToolConfigurationError
-from ..core.traits import Bool, CaselessStrEnum, Int, Path
+from ..core.traits import Bool, CaselessStrEnum, Int, Path, Float, Unicode
 from ..io import EventSource, HDF5TableWriter, TableWriter
 from ..io.simteleventsource import SimTelEventSource
 from ..io import metadata as meta
+from ..io.hdf5tableio import ScaleColumnTransform
 from ..instrument import SubarrayDescription
 
 __all__ = ["DL1Writer", "DL1_DATA_MODEL_VERSION", "write_reference_metadata_headers"]
@@ -138,6 +139,16 @@ class DL1Writer(Component):
     ).tag(config=True)
 
     overwrite = Bool(help="overwrite output file if it exists").tag(config=True)
+
+    transform_image = Bool(default_value=False).tag(config=True)
+    image_dtype = Unicode(default_value="int32").tag(config=True)
+    image_offset = Int(default_value=0).tag(config=True)
+    image_scale = Float(default_value=100.0).tag(config=True)
+
+    transform_peak_time = Bool(default_value=False).tag(config=True)
+    peak_time_dtype = Unicode(default_value="int32").tag(config=True)
+    peak_time_offset = Int(default_value=0).tag(config=True)
+    peak_time_scale = Float(default_value=10.0).tag(config=True)
 
     def __init__(self, event_source: EventSource, config=None, parent=None, **kwargs):
         """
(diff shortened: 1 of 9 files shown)


ctapipe/io/tableio.py
```diff
@@ -192,3 +198,117 @@ def open(self, filename, **kwargs):
...
+
+class ColumnTransform(metaclass=ABCMeta):
+    """
+    A Transformation to be applied before serialization / after deserialization.
+
+    The ``TableWriter`` will call the transform on the data to be stored and
+    ``TableReader`` will call `.inverse`` the reverse the transformation 
+    when a transformation is detected in the file through metadata.
+
+    Transformations implemnt ``get_meta`` to provide the necessary metadata
```
nbiederbeck, line 211: typo: implement

ctapipe/io/tests/test_dl1writer.py
```diff
@@ -63,3 +65,76 @@ def test_dl1writer(tmpdir: Path):
...
+
+    for event in EventSource(output_path):
+
+        for tel_id, dl1 in event.dl1.tel.items():
+            original_image = events[event.count].dl1.tel[tel_id].image
+            read_image = dl1.image
+            print()
+            print(original_image[:10])
+            print(read_image[:10])
+            print()
```
nbiederbeck, line 135: do we keep `print` statements in tests?

kosack (dismissed):

ctapipe/io/tests/test_dl1writer.py
maxnoe: nope

ctapipe/io/tableio.py
maxnoe: fixed, thanks

ctapipe/io/dl1writer.py
```diff
@@ -139,6 +140,16 @@ class DL1Writer(Component):
...
 
+    transform_image = Bool(default_value=False).tag(config=True)
+    image_dtype = Unicode(default_value="int32").tag(config=True)
+    image_offset = Int(default_value=0).tag(config=True)
+    image_scale = Float(default_value=100.0).tag(config=True)
+
+    transform_peak_time = Bool(default_value=False).tag(config=True)
+    peak_time_dtype = Unicode(default_value="int32").tag(config=True)
+    peak_time_offset = Int(default_value=0).tag(config=True)
+    peak_time_scale = Float(default_value=10.0).tag(config=True)
```
moralejo: Set both default scales to 100?  Or even 10 for charge (0.1 p.e. seems enough), and 100 for time as we discussed elsewhere?
I don't think the offset is really needed. To use integers we just want a change of units, i.e. pure scaling with no shift. Since these are signed integers I see no need for the offset (but having it there is somewhat confusing, and could eventually be misused).

Also, the idea was to set the units such that this could be "understandable" outside ctapipe, i.e. opening a DL1 file with vitables. Is this already the case? e.g. units 0.1 pe, 0.01 ns

maxnoe: There is a header keyword `SCALE` saved

maxnoe: @moralejo I think you are looking at an old diff, defaults are now 100 for peak time and 10 for images.

maxnoe: If nobody thinks offsets would be useful here, I can remove them.

moralejo: > Offset, Scale and Original dtype are stored in the metadata, yes

But in case there is an offset, the values one could _see_ with e.g. vitables would not just be the right quantity in certain units, right?

maxnoe: vitables will only see the actual values in the table and you'd have too look into the properties for "OFFSET" and "SCALE" to make sense of them.

maxnoe: @moralejo since all these metadata information (units, scales, time formats, whatever) are not hdf5 features per se but additional functionality we implement using hdf5 metadata, this is true for all conversions. vitables will show you plain mjd, the plain value, ... and you'd have to look into the metadata to see if a float was actually a time object.

maxnoe:
@moralejo this is how it looks in vitables:

Screenshot_20210205_152533

ctapipe/io/dl1writer.py
moralejo: Ok, but when you hover the mouse tip over the images table you would simply read e.g. "image_UNIT: 0.1 pe". An offset would make this much less intuitive.

kosack: The offset is just in case it becomes useful for some reason (to get asymmetric  precision for example).  It's the same thing as done in the FITS standard with the BSCALE and BZERO keywords - any FITS image can have them, and it's up to the software looking at the image to apply them or not. If you look at raw fits pixel info, you would have the same misunderstanding if you didn't look in the header.  I think it's ok to keep it in , but set the offset to 0 by default

moralejo: > The offset is just in case it becomes useful for some reason (to get an asymmetric storage range - using more bits to store positive numbers). It's the same thing as done in the FITS standard with the BSCALE and BZERO keywords - any FITS image can have them, and it's up to the software looking at the image to apply them or not. If you look at raw fits pixel info, you would have the same misunderstanding if you didn't look in the header. I think it's ok to keep it in , but set the offset to 0 by default

Ok!

moralejo:
> @moralejo this is how it looks in vitables:
> 
> Screenshot_20210205_152533

Shouldn't we also have image_UNIT and peaktime_UNIT? (at least for times, since p.e. is not a standard unit)

maxnoe:
@moralejo No, image has no units (p.e. is just a number, not a physical unit) and we decided for efficiency reasons not attach units to peak_time.

We can discuss if that should be added in the IO layer, but that's a different issue than this here.
See discussion here: #1456

moralejo:
> @moralejo No, image has no units (p.e. is just a number, not a physical unit) and we decided for efficiency reasons not attach units to peak_time.
> 
> We can discuss if that should be added in the IO layer, but that's a different issue than this here.
> See discussion here: #1456

Ok about image being a number, also because "p.e." is very ambiguous (regarding pulse integration time, use or not of tail corrections, etc).  But I did not see in #1456 any argument against attaching units to peak_time. If there is a reason not to use _UNIT, at least it could go in _DESC.

maxnoe:
Again, this has nothing to do with this PR. If you want to have units attached to peak time, that's something for another PR.

nbiederbeck (approved):

maxnoe:
@kosack @moralejo Could you review again? Let's get this out today if possible.

kosack (approved):

PR #1587, opened 2021-02-01
Add brightest_island function

maxnoe:


ctapipe/image/morphology.py
```diff
@@ -110,28 +110,69 @@ def number_of_island_sizes(island_labels):
...
+
+    # only look at the actual islands, 0 means pixel did not survive cleaning
+    mask = island_labels > 0
+
+    # calculate the sum of image for each island via numpy ufunc magic
+    # basically, each pixel is added to `island_brightness` at its island_label
+    island_brightness = np.zeros(n_islands + 1)
+    np.add.at(island_brightness, island_labels[mask], image[mask])
+
+    # +1 because we discard the 0 "no island" marker above
```
kosack, line 172: Not sure I understand why this comment is associated with this particular line.  Does it apply to line 169 (where there is a +1)?

ctapipe/image/tests/test_morphology.py
```diff
@@ -105,3 +105,22 @@ def test_largest_island():
     assert (mask_largest_one == true_mask_largest_one).all()
     assert (mask_largest_0 == true_mask_largest_0).all()
     assert_allclose(mask_largest, true_mask_largest)
+
+
+def test_brightest_island():
+    # we have two islands, 2 is larger but 1 is brighter
+    n_islands = 2
+    image = np.array([0, 1, 2, 5, 1, 0, 1, 2, 2, 1])
+    island_labels = np.array([0, 0, 1, 1, 1, 0, 2, 2, 2, 2])
```
HealthyPear, line 114: Shouldn't this be
`island_labels = np.array([0, 1, 1, 1, 1, 0, 2, 2, 2, 2])`

ctapipe/image/morphology.py
maxnoe: New code and already outdated docs :cry: Will remove.

ctapipe/image/tests/test_morphology.py
maxnoe: No. It is an artificial test case. I wanted to have a smaller but brighter island. So island 1 has three pixels, island 2 has four but less photons.

HealthyPear: Ah so you are already considering a fictional image cleaning that kills the 2nd pixel?

maxnoe: No, I just made a simple toy example. Don't interpret too much into the specific numbers

ctapipe/image/morphology.py
```diff
@@ -110,28 +110,69 @@ def number_of_island_sizes(island_labels):
...
+    # only look at the actual islands, 0 means pixel did not survive cleaning
+    mask = island_labels > 0
+
+    # calculate the sum of image for each island via numpy ufunc magic
+    # basically, each pixel is added to `island_brightness` at its island_label
+    # +1 because 0 means "no island" and we have up to "n_islands" values.
+    island_brightness = np.zeros(n_islands + 1)
+    np.add.at(island_brightness, island_labels[mask], image[mask])
+
+    brightest_island = np.argmax(island_brightness)
```
kosack, line 173: Shouldn't name your variable `brightest_island`, as that is also the function name

kosack (changes requested):

ctapipe/image/morphology.py
```diff
@@ -110,28 +110,69 @@ def number_of_island_sizes(island_labels):
...
+        The image array
+
+    Returns
+    -------
+    mask : array
+        A boolean mask created from the input labels and filtered for the brightest island.
+        If no islands survived the cleaning the array is all False.
+    """
+
+    if n_islands == 0:
```
HealthyPear, line 161: you can use `island_labels` also for this check

ctapipe/image/morphology.py
HealthyPear: maybe `brightest_label`?

HealthyPear (changes requested):

ctapipe/image/morphology.py
maxnoe: But checking the integer is cheaper than recomputing it using `np.count_nonzero`. Since this information is available, why not use it?

ctapipe/image/morphology.py
maxnoe: That's actually not a problem, since it only assigns a variable in the local scope of the function (and the function is not recursive).

However, to avoid confusion, I changed the variable name.

kosack: Thanks.  it was more to avoid confusion and the linter complaining

kosack (approved):

HealthyPear (approved):

Issue #1315, opened 2020-05-07
Telescope / NominalFrame for reconstruction / visualization


maxnoe:
This is a meta issue for open issues related to improving calculation of image parameters,
reconstruction and plotting.

- [x] Hillas parameter in Telescope / NominalFrame (which? matters for divergent pointing), #1090 
- [ ] Plotting (CameraDisplay). Also for BokehViewer? #1191, #416 
- [ ] Easy way to convert camera geometry between the coordinate frames
- [ ] Pixel areas / diameters in degrees / steradian

HealthyPear:
I am biased since I really need to fix this, but I propose to start from the first one :)

For protopipe, there are 2 different, but related aspects:
- use the _effective_ focal length to radially scale the camera (I think now in ctapipe there is an option to choose, but the values are not in the simtel file, rather in [this table](https://github.com/cta-observatory/ctapipe/files/1332851/flen.pdf)),
- change the order of the operations in a way that the Hillas parameters have already dimension of degrees before the direction reconstruction is performed.

The second aspect is a mixture of two things:
- the reconstructor must change (protopipe uses by default `ctapipe.reco.HillasReconstructor`) in order to not make this transformation
- the transformation itself should happen once at the beginning (no information from the image itself is needed), and this is probably more a pipeline issue
If someone remembers, I already showed a proposal for a way to do this [here](https://indico.cta-observatory.org/event/2656/) (the notebook titled _test_camera_shrinking_).

There I transform to TelescopeFrame, but I imagine that one can do the same for Nominal.
What matters for the final result is the use of the right focal length and the order of the operations.

The only problem for me is that for some reason (either it is supposed to be like this or I made a mistake, but I didn't find it yet) the performance (very simple, just the theta2 distribution right out from  `ctapipe.reco.HillasReconstructor`) using the effective focal length seems bad compared than the one calculated using the focal length currently used in ctapipe (nominal).

So I would propose to start from there (I can move this into a dedicated issue or open a PR right away to test this if we agree).

MaxNoe:
@HealthyPear actually, I think the discussion on which focal length to use for this transformation is a completely different one.

For this issue here, let's assume we have the right one and discuss the necessary steps to calculate this stuff in the respective coordinate frames.

Which focal length to put in is also a very simple change, unlike the general restructuring that's needed to make this happen.

HealthyPear:
[...]

Issue #1090, opened 2019-06-18
Calculate Hillas Parameters in TelescopeFrame


maxnoe:
This was a comment made by Razmik at the meeting in Lugano and something I heard from many people.

We should calculate hillas parameters in telescope frame, so that the units are deg and comparable between telescopes.

thomasgas:
Easy fix: divide the values that we have in meters by the focal length, which gives the result in radians that can be converted to deg?

kosack:
that was the original idea in fact, just somehow never happened since the coordinate frames weren't finished until after the Reconstructor code was written (it originally did the transform internally).  We could transform the cameras first, or transform the hillas ellipse after.  Either way lets you compare.  I don't see this as a major problem right now, since you can still do that, but i guess to be able to compare histograms of width and length, it would be nice to store them that way in the DL1 output. 

If you really want to compare, it should be in something like the `NominalFrame`, since with the `TelescopeFrame` you still have a projection effect if the pointing isn't parallel, but since the HillasReconstructor does things in 3D, that frame is not needed.

kosack:
> Easy fix: divide the values that we have in meters by the focal length, which gives the result in radians that can be converted to deg?

Isn't that basically the transform from Camera to Telescope? You could also load all of the hillas.x and hillas.y values (already computed)  into a SkyCoord and transform to the TelescopeSystem for each telescope and have the same effect.  The question is what is the most useful for storage in DL1? If you later re-measure the real focal length and find it is different from the one you used, you'd have to redo the transform to camera and back again to fix it, but I guess you'd also start from DL0 data in that case, so it wouldn't matter.

kosack:
Of course dealing with width and length, etc,  is slightly different (we'd have to have a transform function for lengths, not just points), so maybe just using the Telescope or Nominal frame to start with is easier to understand (and then the conversion can be removed from the HillasReconstructor).

MaxNoe:
> Of course dealing with width and length,

Not really. As we are dealing with small angles (< 10 degrees for all telescopes).

kosack:
Well, you still need a simple function to convert from meters to degrees. Not difficult, I agree, but right now you can only transform points, not a HillasParametersContainer. Wouldn't be hard to do though.

kosack:
I guess this issue boils down to: *can you make a histogram of Hillas width for all telescopes in such a way as you can compare them, from DL1 output alone?*,

[...]

Issue #674, opened 2018-03-12
Calculation of Hillas Parameters in the Nominal System


ParsonsRD:
From looking through the code the default (only) way to calculate the Hillas parameters seems to be in the camera frame. Probably the default behaviour of this of this code should be to calculate the Hillas parameters in the Nominal shared angular system. Having these values in angular units makes a lot more sense I think as they are then inherently comparable between different telescope types, additionally any further reconstruction will take place in the Nominal system (or the Horizon system).

Although the Hillas parameters can always be converted between systems if a special routine it written (barring any complex non linear Camera to Nominal system conversion), I think having a Hillas with measured in metres is physically meaningless and we should never really want this. 

My suggestion for a new behaviour to fix this and make coordinates easier to use in general would be to store both the pixel positions in the geometry object as a coordinate object and update the array and telescope pointing directions in this object for each event. That way the user can convert to any system they like with a single line without having to get values from the event.inst... object.

@kosack  how does that sound to you?

kosack:
Right now, the HillasReconstructor does not use the coordinate frames, so we have to think of the right way to do it without breaking things.  I sort of want to wait until we have fixed the coordinates module, which doesn't quite do things correctly - I think somebody had mentioned they started migrating it? 

However, in general, this is something that I've wanted to fix at some point - either adding a method to CameraGeometry to convert the pixel positions to other frames, or adding a function that allows us to transform the Hillas parameters themselves afterward  (which may be more efficient).

ParsonsRD:
OK that's fine. I agree we should come up with some code to convert Hillas parameters between systems, but I think probably the default behaviour should be to create them in some kind of angular system as the use cases of these values are many, whereas the use cases on Hillas parameters measured in metres are essentially none (drawing on top of the camera I guess).

I think it's good also to take a closer look at the coordinates code. I would be happy to contribute some time to this. Do we have a plan for how to change things? If not to we have a list of issues with the current implementation?

MaxNoe:
What is the `Nominal System` and is there any document that defines these frames?

ParsonsRD:
This is a major part of the problem, no there isn't yet any documentation of what coordinate systems we should have/use. When I wrote this code I created the systems I though we would need with a heavy leaning on the HESS documentation, but my hope was that a coordinates guidelines document would be available soon after.

All the systems I created are currently only documented in their docstrings
[...]

PR #1591, opened 2021-02-04
Hillas parameters in telescope frame

LukasNickel:
Related discussions about this can be found in:
- #674
- #1090 
- #1315 
- #1408 

This changes the calculation of the image parameter for the dl1 output to be performed in the nominal frame.
For this, new containers have been added. This is kind of the brute force way. Since it is only relevant for a few container fields (x,y,length, width...), we could also make their unit configurable. However, that might lead to confusion, because the field names do not reflect that difference. Or let one container have both fields of which only one is filled? 
Side note: Some code (e.g. calculation of the concentration) explicitly names things x and y, which I did not change.
So maybe a naming of the coordinate axes, that does not indicate the use of a certain coordinate frame could be a way out?

All of these changes are written out by the stage1-tool. Coordinate transformations happen in the ImageProcessor.
Since new frames are constructed for each event, this might be slower. I do not know how to access all pointings at once without iterating over the events though.

Regarding the change of the data model version:
We probably want to collect test files of older versions in order to make sure, that we do not break their functionality.
Is adding dl1 files to ctapipe-extra the way to go or is there another solution?

--------------------------------------------------------
Not longer relevant (keeping it for context):

The more I think about this, the less certain I am if we even want to calculate features in the nominal frame or if a later conversion makes more sense. Anyway, lets discuss this.

**Status:**
For now this adds a new method and a new container for the calculation of the parameters in the nominal frame.
Since conversion from the camera geometry to the nominal frame requires the telescope pointings, I figured it would be easiest to perform this transformation outside and call the function with a SkyCoord containing the pixel positions. After writing the code I learnt that the camera geometry also has a `transform_to` method, so thats another possibility. Would make the API clearer but adds some overhead to the calculation.

There is of course a lot of duplicated code right now since the performed steps are pretty much identical. So if we decide to follow up on this approach, it would be beneficial to put some of that code into an extra function.
Some feature (Intensity/Skewness/...) are identical. If we want to include them in the NominalHillasParametersContainer as well probably depends on whether we want both sets of parameters in stage1 or only one of them.

With the coordinate frames at hand, it is possible to convert from one set of parameters to the other and the results should match.

**Questions / Points to discuss**
- Is this how we want to do it or do we keep the calculation in the camera frame and convert afterwards? 
- The current image processor does not seem to be aware of the telescope pointing (?). This can of course be easily changed but might be an argument towards a calculation in the camera frame and later transformation so long as there will not be more image features, that should be calculated in the nominal frame. (Timing?)
- Do we want to keep both containers or would this replace the calculation in the camera frame? -> Should there be duplicated features like intensity or just the "new" ones? What should be part of the stage-1 tool?

**Images, Cog, length and width in the nominal frame:**
nom_frame


HealthyPear:
> The more I think about this, the less certain I am if we even want to calculate features in the nominal frame or if a later conversion makes more sense.

From the point of view of `ctapipe.reco.HillasReconstructor` it would probably be overkill to calculate the parameters in this frame, for 2 reasons:
- the only practical gain to go from CameraFrame to any sky-based frame is to remove the optics from the equation and better compare the images afterward (this is at the core of #1408 , not only because of the comparison with CTAMARS)
- the only use of the NominalFrame for non-parallel pointing, either from mispointing or from divergent pointing or a mix of the 2 - in the former case we should apply anyway pointing corrections, so the effect would be small, in the latter `ctapipe.reco.HillasReconstructor` already takes care of this by projecting from the sky to a "fake" array-based CameraFrame, thus nullifying the divergent pointing.

Of course, I see no problem in adding the possibility of having the parameters in yet another frame, but that would complicate #1408 and it's encapsulation in stage2 and I think merging that before would make things clearer also for this PR.

LukasNickel:
Yeah, the HillasReconstructor does not really care about this.
It is more about what we want to store in dl1 files rather than the dl2 calculation.

maxnoe:
@LukasNickel 

What we discussed was something else. The task for the next ctapipe release should be to have the DL1 in degrees.
I am no expert in the divergent pointing analysis, so I don't know if it makes sense to go to `NominalFrame`.

For now, let's keep it simple. 

* Do not copy the code, just make sure it works when `pix_x` and `pix_y` are in degrees.
* Use a different container so the difference is transparent
* Change the ImageProcessor to do this by default

LukasNickel:
>     * Do not copy the code, just make sure it works when `pix_x` and `pix_y` are in degrees.
> 
>     * Use a different container so the difference is transparent
> 
>     * Change the ImageProcessor to do this by default

Ok, lets do it like this.

maxnoe:
Since we want to switch to the hillas parameters in degree by default I think it would make sense to actually make the `HillasParametersContainer` the one with the stuff in degrees and have a `CameraHillasParameters` container.

@kosack @HealthyPear opinions?

HealthyPear:
> Since we want to switch to the hillas parameters in degree by default I think it would make sense to actually make the `HillasParametersContainer` the one with the stuff in degrees and have a `CameraHillasParameters` container for the one with unit meters.
> 
> @kosack @HealthyPear opinions?

I like the idea!
Should I change this directly into #1408 or we do it here after the merge?

ctapipe/image/image_processor.py
```diff
@@ -169,8 +171,33 @@ def _process_telescope_event(self, event):
         """
         Loop over telescopes and process the calibrated images into parameters
         """
+        location = EarthLocation.of_site("Roque de los Muchachos")
```
maxnoe, line 174: Wait, are you hardcoding the ORM here for DL1Writer????

maxnoe: Why do you even need a site for using `NominalFrame`?

maxnoe (changes requested):
* Use telescope frame
* Make it default in the image processor
* Make the normal containers without prefix the ones with stuff in degrees
* Keep the old ones with a prefix `Camera....`
* Remove any hard coded sites / array pointings /

ctapipe/image/image_processor.py
LukasNickel: Wasn't meant to stay. I had trouble figuring out the position at first.

kosack:
A few comments:
*  I don't see any reason to keep two copies of the containers.  Just convert everything spatial to TelescopeFrame, and drop the old CameraFrame parts.   (caveat: see last point)
* There is a very good reason *not* to use the NominalFrame (other than it is not used by the HillasReconstructor), and that is that we may update the astrometric (pointing) corrections, which changes the NominalFrame. It is nice to not have to redo DL1 if those change.  Also the center position of the nominal frame is not well defined (by default it is the barycenter of the various telescope pointings, but that may be changed/optimized later).  If we store in TelescopeFrame, it's easy to add any corrections before reconstruction. 
* In principle you don't even need two containers if you want to retain the ability to compute hillas parameters in any frame.  The original idea was that hillas.x could be in meters or degrees.  The only problem is just that the containers so far force a unit.  If we want to support both cases we could make that container have a constructor argument that specifies the unit or something like that.  It could be useful to support the CameraFrame as well,  as the impact point reconstruction  is done in the camera frame - see all the tricks that Michele had to do to have the psi angle in both Telescope and Camera frame without breaking things..., but it's not really necessary
* Another option is to just make a conversion function that converts lengths and angles from one frame to another, and then you can convert HillasParameters (and any other spatial parameters) without having to change the CameraGeometry and re-compute them.

ctapipe/image/image_processor.py
```diff
@@ -169,8 +169,9 @@ def _process_telescope_event(self, event):
         """
         Loop over telescopes and process the calibrated images into parameters
         """
-
         for tel_id, dl1_camera in event.dl1.tel.items():
+            geometry = self.subarray.tel[tel_id].camera.geometry
+            geometry_telescope = geometry.transform_to(self.telescope_frame)
```
kosack, line 174: This is a pretty slow operation (transforming all pixels), and since nothing really changes per event, it could be pre-computed once. Only the focal length matters here, not the pointing, as it is still relative to the camera center.   So we could either store the TelescopeFrame Geometries in a cache in the ImageProcessor (once per telescope).

LukasNickel:
> In principle you don't even need two containers if you want to retain the ability to compute hillas parameters in any frame.  The original idea was that hillas.x could be in meters or degrees.  The only problem is just that the containers so far force a unit.  If we want to support both cases we could make that container have a constructor argument that specifies the unit or something like that. 

I feel like this makes it harder to keep track of the units and coordinate frames. The reconstructor could then not rely on getting a HillasParametersContainer alone, but would also need to check if its in the correct frame. Also `x/y` are heavily connected to the camera frame in my opinion. If its still worth it in order to keep a cleaner looking container structure, I do not really know right now.

> It could be useful to support the CameraFrame as well,  as the impact point reconstruction  is done in the camera frame - see all the tricks that Michele had to do to have the psi angle in both Telescope and Camera frame without breaking things..., but it's not really necessary

Do you mean, that you want to save results in both coordinate frames in the dl1 output? 

> Another option is to just make a conversion function that converts lengths and angles from one frame to another, and then you can convert HillasParameters (and any other spatial parameters) without having to change the CameraGeometry and re-compute them.  That could get rid of the need to store the psi angle in two frames as in #1408

General conversion functions for the dl2 reconstructors could be handy, agreed. If we agree on a list of functions, I can add them here, in a new pr or we see it as part of #1408.

LukasNickel:
Regarding the failing test:
Travis is consistently calculating 0.06638165 deg as the width, while my local tests run just fine with 0.06022414 deg.
Why is there a difference at all? Shouldn't the toymodel produce the same image and the transformation + hillas calculation be identical?

maxnoe:
Will be fixed by merging #1589 or removing your cache

HealthyPear:
I also like the idea of making the HillasContainer a bit more adaptive (I tried to propose this during #1408 in fact).

If there's a way to do this It would be cool to build a HillasContainer with a frame object and a unit, so even if unit is degrees one can differentiate between TelescopeFrame and NominalFrame.

Also because in #1408 I didn't change the Hillass containers, I just write to them in a frame, but I think that for a dummy event the units would be wrong (the values instead would be always either 0 or NaN)

maxnoe:
@HealthyPear No, please don't! We should keep containers simple without any logic other then storing / validating data. I think they are too complex as it is.

HealthyPear:
> @HealthyPear No, please don't! We should keep containers simple without any logic other then storing / validating data. I think they are too complex as it is.

Then if we want to store Hillas parameters in more than one frame we have to create at least 2 of them, one for meters and one for degrees.
I don't see many other alternatives...

maxnoe:
@kosack The simplest solution would be to just drop one of the options.

I disagree that it is as simple as allowing degrees and meters as units, since we should also use the correct names for the coressponding axes. So `fov_lon` / `fov_lat` for the sky frames and `x` and `y` for the camera frames.

So it all gets a  bit more complicated when we want to support both.

However, nothing really prevents us from just *switching* from `CameraFrame` to `TelescopeFrame` with a breaking change I think. I think it would be a sensible decision to do that.

kosack:
> However, nothing really prevents us from just switching from CameraFrame to TelescopeFrame with a breaking change I think. I think it would be a sensible decision to do that.

Yes, that was my original plan. I guess it could be done in a phased way though -start out supporting both frames as we do now with 2 classes (and two modes in #1408), but later remove the bits of the HillasReconstructor that operate in CameraFrame mode once we are sure we are happy (keeping both makes it so much more complex and hard to read).

LukasNickel:
> Will be fixed by merging #1589 or removing your cache

That did not do the trick sadly

HealthyPear:
> > However, nothing really prevents us from just switching from CameraFrame to TelescopeFrame with a breaking change I think. I think it would be a sensible decision to do that.
> 
> Yes, that was my original plan. I guess it could be done in a phased way though -start out supporting both frames as we do now with 2 classes (and two modes in #1408), but later remove the bits of the HillasReconstructor that operate in CameraFrame mode once we are sure we are happy (keeping both makes it so much more complex and hard to read).

As I said, as part of this problem, I believe that now in #1408 there is a problem: a dummy HillasParametersContainer will always have meters as units, even though the analysis is working on degrees....could you cross-check to see if am right?

maxnoe:
@HealthyPear That is not a real problem though. You can instantiate a hillas parameters container with nan values in degrees and make sure that is used for non-paremeterized images. You are not forced to use the default for that.

kosack:
> As I said, as part of this problem, I believe that now in #1408 there is a problem: a dummy HillasParametersContainer will always have meters as units, even though the analysis is working on degrees....could you cross-check to see if am right?

Again, it's a good reason to switch completely to TelescopeFrame.  We can leave the CameraFrame container there in case we want it for visualization, but the default should be TelescopeFrame always.  Then there is no problems with default or dummy containers (event.dl1.tel[x].parameters.hillas should be a HillasParametersContainer that is in the TelescopeFrame).   Then, if we provide a way to convert betwen one to the other, you can still support anything you like (e.g. a function like `hillas_params_to_camera(HillasParametersContainer) → CameraHillasParametersContainer`.  

Nevertheless, i don't really care so much about the (x,y) vs (fov_lon, fov_lat) difference, but being consistent is nice.

kosack:
> That did not do the trick sadly

Are you sure there is nothing in `~/.cache/ctapipe` ? It should have made it give you the same answer as on github.  Since the pixel areas in the camgeom files now use the correct value rather than a guessed one, it changes the predicted image by 3% or so.

HealthyPear:
> @HealthyPear That is not a real problem though. You can instantiate a hillas parameters container with nan values in degrees and make sure that is used for non-paremeterized images. You are not forced to use the default for that.

Oh right, I didn't realize that

```python
h = HillasParametersContainer(x = 4 * u.deg)

h.x
<Quantity 4. deg>

h.y
<Quantity nan m>
```

Nevertheless I think I will need to correct #1408 where I do simply `h = HillasParametersContainer()`

kosack:
> Nevertheless I think I will need to correct #1408 where I do simply h = HillasParametersContainer()
> But then the units problem is not a problem anymore am I right?

No, containers don't check until you call `Container.validate()` (which gets called when writing)

HealthyPear:
Also, I didn't want to sound too much crude, but I also don't quite care about the naming of the coordinates...
In the end it's clear what is X and what is Y in the sky as soon as you say in which reference system you are (like `AltAz()`)

HealthyPear:
> > Nevertheless I think I will need to correct #1408 where I do simply h = HillasParametersContainer()
> > But then the units problem is not a problem anymore am I right?
> 
> No, containers don't check until you call `Container.validate()` (which gets called when writing)

OK but then there IS a problem in the end, am I right?
As soon as we write, if I do not initialize the dummy container with the right units, the check will fail

maxnoe:
> As soon as we write, if I do not initialize the dummy container with the right units, the check will fail

Yes, but you can do that, so not a problem

maxnoe:
> Also, I didn't want to sound too much crude, but I also don't quite care about the naming of the coordinates...

I do care. I think it is important to avoid mental mapping here.

HealthyPear:
> > Also, I didn't want to sound too much crude, but I also don't quite care about the naming of the coordinates...
> 
> I do care. I think it is important to avoid mental mapping here.

Would it be so crazy to add a `frame` Field to `HillasParametersContainer` container and make `hillas_parameters` write it via the `CameraGeometry` object?

In this way, there would never be ambiguity.

maxnoe:
Yes, imho it would.

ctapipe/containers.py
```diff
@@ -118,8 +123,35 @@ class HillasParametersContainer(Container):
...
 
-    skewness = Field(nan, "measure of the asymmetry")
-    kurtosis = Field(nan, "measure of the tailedness")
+
+class HillasParametersContainer(BaseHillasParametersContainer):
+    container_prefix = "hillas"
+    lon = Field(
+        nan * u.deg,
+        "centroid longitude in a sky frame e.g. the telescope frame",
+        unit=u.deg,
```
kosack, line 132: if we stick with this renaming, it might be good to change the help string to say "longitude angle in a spherical system centered on the pointing position (TelescopeFrame)".   Also shouldn't it be fov_lon to be consistent with the Frame? (I don't really mind either way)

maxnoe: +1 for `fov_lon`

kosack:
> I do care. I think it is important to avoid mental mapping here.

Actually, the more I think about it I tend to agree- these lon/lat values cannot be used in the same way as an x,y value as they are not a cartesian projection (i.e. *fov_lon* is degrees along a line of latitude, not degrees on the sky, which differ by 0.4% at 5 degrees in latitude in the camera in this system).

kosack:
I still like the idea of just adding some functions to `ctapipe.coordinates` like `angle_camera_to_telescope(angle)` and `distance_camera_to_telescope(distance)` that convert these quantities, so we can just drop the camera frame parameters.

maxnoe:
@kosack These are spherical coordinates. The difference at the diagonal edges of an sst fov is this:


```
In [1]: from astropy.coordinates import SkyCoord

In [2]: from ctapipe.coordinates import TelescopeFrame

In [3]: coord = SkyCoord(5, 5, unit='deg', frame=TelescopeFrame())

In [4]: center = SkyCoord(0, 0, unit='deg', frame=TelescopeFrame())

In [5]: import numpy as np

In [6]: np.sqrt(coord.fov_lon**2 + coord.fov_lat**2)
Out[6]: <Quantity 7.07106781 deg>

In [7]: coord.separation(center)
Out[7]: <Angle 7.06657439 deg>
```

ctapipe/containers.py
```diff
@@ -213,8 +276,8 @@ class ImageParametersContainer(Container):
     """ Collection of image parameters """
 
     container_prefix = "params"
-    hillas = Field(HillasParametersContainer(), "Hillas Parameters")
-    timing = Field(TimingParametersContainer(), "Timing Parameters")
+    hillas = Field(BaseHillasParametersContainer(), "Hillas Parameters")
```
maxnoe, line 279: That does not make sense. This will be an incomplete container as default. Please use the containers with units degrees as default and add `type=BaseHillasParametersContainer`.

LukasNickel: Oh you just set a type to a field.
Yes, thats much better

ctapipe/io/dl1writer.py
```diff
@@ -28,9 +28,10 @@
 
 tables.parameters.NODE_CACHE_SLOTS = 3000  # fixes problem with too many datasets
 
-DL1_DATA_MODEL_VERSION = "v1.0.3"
+DL1_DATA_MODEL_VERSION = "v1.0.4"
 DL1_DATA_MODEL_CHANGE_HISTORY = """
 - v1.0.3: true_image dtype changed from float32 to int32
+- v1.0.4: hillas and timing parameters saved in telescope frame (degree) as opposed to camera frame (m)
```
maxnoe, line 34: I would say a major update, since also the names changed.

kosack:
> @kosack These are spherical coordinates. The difference at the diagonal edges of an sst fov is this:

Yes, that is exactly what I said above. Not sure I get your point.  At 5.0 degrees offset in longitude from the center of the camera, a difference in spherical latitude is  about 0.4% different than a difference in actual angular separation.  It's small, but have to still be careful not to confuse the two.

ctapipe/containers.py
```diff
@@ -180,6 +218,31 @@ class TimingParametersContainer(Container):
...
+    """
+    Slope and Intercept of a linear regression of the arrival times
+    along the shower main axis in a
+    spherical system centered on the pointing position (TelescopeFrame)
+    """
+
+    container_prefix = "timing"
+    slope = Field(
+        nan / u.deg, "Slope of arrival times along main shower axis", unit=1 / u.deg
+    )
```
kosack, line 243: I'd say we should make the slope in `ns/deg`, if we want this to be comparible for all cameras (the point of going to deg in the first place).  That means we do not want to have samples/deg, since the sampling frequencies are not all 1 GHz.

kosack (changes requested):

ctapipe/containers.py
LukasNickel: Right now the peak time has no unit attached though, right?

kosack: No (they are in samples), but the sampling frequency is available, so the parameters could still be in correct units (as I think Abelardo mentioned).   If the point of this PR is to ensure all parameters can be compared in a telescope-independent way, i.e. removing the effect of camera scale and focal length, then we should also remove the effect of sampling frequency from all timing parameters (the image itself could stay in samples), but we should multiply these parameters by `subarray.tel[tid].camera.readout.sampling_rate`

kosack: It occurs to me that defining the slope as a velocity (deg/ns) might be a nicer unit, but that's another issue.

maxnoe: > No (they are in samples), 

No, peak time is in nanoseconds! But without units attached for performance reasons. But we actually never studied the impact of using a quantity with `ns` everywhere.

kosack: Well if this is already a Quantity, then it should be "ns/deg". We can just attach the ns in the ImageProcessor if that's easiest, but it would be good to have this unit there as it it clearly already confusing.  If it says "ns", then we know there is no need to apply the sampling rate.

kosack: We could then also rename peak_time to peak_time_ns (if no units are there), but the *parameters* which have units already should at the very least be the correct units.

LukasNickel: So we can either:
- Use astropy quantities everywhere
- Hack in a "ns" to make the output consistent

plus renaming the peak time field.
I can have a look at the performance impact of using the correct unit everywhere in the code, but this might be better located in another PR?

maxnoe: Yes, this should be different. Let's focus here on the transformation into telescope frame, not some other connected issues

ctapipe/utils/datasets.py
```diff
@@ -38,7 +38,7 @@ def get_searchpath_dirs(searchpath=os.getenv("CTAPIPE_SVC_PATH")):
     else:
         searchpaths = [Path(p) for p in os.path.expandvars(searchpath).split(":")]
 
-    searchpaths.append(get_cache_path(""))
+    searchpaths.append(get_cache_path(DEFAULT_URL))
```
kosack, line 41: This doesn't seem like it should be part of this PR (and same for all the other changes to datasets.py)

kosack (changes requested):
I only have minor comments about changes that seem to be mixed in there that shouldn't be there, but otherwise this looks ok. Once we can merge it, we can use this 
in #1408 , which is the main thing we need before implementing the rest of the  pipeline.

maxnoe:
@kosack I think these should be gone after the conflicts are resolved since those changes are already in master.

maxnoe:
@kosack Yes, those changes are gone after merging master / fixing conflicts.

kosack (dismissed):

LukasNickel:
The new test for the `hillas_intersection` reconstructor fails, because the reconstructor requires hillas parameters in the camera frame which the test does not provide.
Should I change it (plus the HillasReconstructor while we are at it) to support both parametrizations or only the parameters in the telescope frame? Should not be too much work to support both, but it probably adds complexity we don't really need?

LukasNickel:
> I'd say we limit ourselves here to DL1b and leave the testing/usage of this new feature to subsequent PRs.

Ok, than we just skip the test here?

maxnoe:
> I'd say we limit ourselves here to DL1b and leave the testing/usage of this new feature to subsequent PRs.

I disagree. If we change the default to degrees, we need to make sure that all implemented reconstruction methods can work with that. 

The test fails for a reason. The code needs to be fixed. Just deactivating the test is not the solution.

HealthyPear:
From what I can see from your changes you have introduced a new `CameraHillasParametersContainer` which contains by default quantities in meters, but then in the test you initilize the container with quantities in degrees by converting the camera quantities with the focal length.

Maybe I missed something along the way, but why calling it **Camera**HillasParametersContainer if then you don't use camera-based quantities?

maxnoe:
> Then in this PR leave the default to meters?

No, just fix the code here.

> Never said this....

The rule is not to break the master. You proposed to merge the changes and fix the code in another PR. That is not what should be done.

LukasNickel:
> From what I can see from your changes you have introduced a new `CameraHillasParametersContainer` which contains by default quantities in meters, but then in the test you initilize the container with quantities in degrees by converting the camera quantities with the focal length.
> 
> Maybe I missed something along the way, but why calling it **Camera**HillasParametersContainer if then you don't use camera-based quantities?

You mean the `test_intersection_xmax_reco()` test? 
I assume, that I just replaced HillasParametersContainer with CameraHillasParametersContainer everywhere and didnt realize it was using degrees before. Thats the only reason I can imagine right now.

ctapipe/reco/tests/test_hillas_intersection.py
```diff
@@ -144,9 +144,9 @@ def test_intersection_weighting_spoiled_parameters():
 
     # telescope 2 have a spoiled reconstruction (45 instead of -45)
     hillas_dict = {
-        1: HillasParametersContainer(intensity=10000, psi=-90 * u.deg),
-        2: HillasParametersContainer(intensity=1, psi=45 * u.deg),
-        3: HillasParametersContainer(intensity=10000, psi=0 * u.deg),
+        1: CameraHillasParametersContainer(intensity=10000, psi=-90 * u.deg),
```
maxnoe, line 147: Same as above

ctapipe/reco/tests/test_hillas_intersection.py
```diff
@@ -189,21 +189,21 @@ def test_intersection_nominal_reconstruction():
     cog_coords_nom_3 = cog_coords_camera_3.transform_to(nominal_frame)
 
     #  x-axis is along the altitude and y-axis is along the azimuth
-    hillas_1 = HillasParametersContainer(
+    hillas_1 = CameraHillasParametersContainer(
```
maxnoe, line 192: Same as above.

ctapipe/visualization/tests/test_mpl.py
```diff
@@ -114,16 +114,16 @@ def test_array_display():
 
     # test using hillas params:
     hillas_dict = {
-        1: HillasParametersContainer(length=100.0 * u.m, psi=90 * u.deg),
-        2: HillasParametersContainer(length=20000 * u.cm, psi="95deg"),
+        1: CameraHillasParametersContainer(length=100.0 * u.m, psi=90 * u.deg),
```
maxnoe, line 117: Same

ctapipe/reco/tests/test_hillas_intersection.py
```diff
@@ -119,9 +119,9 @@ def test_intersection_reco_impact_point_tilted():
     tel_y_dict = {1: delta, 2: delta, 3: -delta}
 
     hillas_dict = {
-        1: HillasParametersContainer(intensity=100, psi=-90 * u.deg),
-        2: HillasParametersContainer(intensity=100, psi=-45 * u.deg),
-        3: HillasParametersContainer(intensity=100, psi=0 * u.deg),
+        1: CameraHillasParametersContainer(intensity=100, psi=-90 * u.deg),
```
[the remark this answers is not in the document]
HealthyPear: What is the problem here? psi has always been in degrees no?

HealthyPear:
> > From what I can see from your changes you have introduced a new `CameraHillasParametersContainer` which contains by default quantities in meters, but then in the test you initilize the container with quantities in degrees by converting the camera quantities with the focal length.
> > Maybe I missed something along the way, but why calling it **Camera**HillasParametersContainer if then you don't use camera-based quantities?
> 
> You mean the `test_intersection_xmax_reco()` test?
> I assume, that I just replaced HillasParametersContainer with CameraHillasParametersContainer everywhere and didnt realize it was using degrees before. Thats the only reason I can imagine right now.
> 
> To be honest: That was pretty hacky even before, setting `x/y` to some `fov_lon/lat` in the test

yes but also later there is a test that starts in nominal frame

ctapipe/reco/tests/test_hillas_intersection.py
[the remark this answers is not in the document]
maxnoe: Ah, sorry, parsed it wrong. Was confused that only intensity and psi are set.

ctapipe/reco/tests/test_hillas_intersection.py
maxnoe: But here it is really wrong!

ctapipe/reco/tests/test_hillas_intersection.py
```diff
@@ -189,21 +189,21 @@ def test_intersection_nominal_reconstruction():
...
-    hillas_1 = HillasParametersContainer(
+    hillas_1 = CameraHillasParametersContainer(
         x=cog_coords_nom_1.fov_lat,
         y=cog_coords_nom_1.fov_lon,
         intensity=100,
         psi=0 * u.deg,
     )
 
-    hillas_2 = HillasParametersContainer(
+    hillas_2 = CameraHillasParametersContainer(
```
maxnoe, line 199: And here

ctapipe/reco/tests/test_hillas_intersection.py
```diff
@@ -189,21 +189,21 @@ def test_intersection_nominal_reconstruction():
...
         x=cog_coords_nom_2.fov_lat,
         y=cog_coords_nom_2.fov_lon,
         intensity=100,
         psi=45 * u.deg,
     )
 
-    hillas_3 = HillasParametersContainer(
+    hillas_3 = CameraHillasParametersContainer(
```
maxnoe, line 206: And here.

maxnoe:
Sorry, yes I marked a couple of false positives, but there are some where you really construct a `CameraHillasParametersContainer` with values in degrees (from nominal frame).

HealthyPear:
> > Then in this PR leave the default to meters?
> 
> No, just fix the code here.
> 
> > Never said this....
> 
> The rule is not to break the master. You proposed to merge the changes and fix the code in another PR. That is not what should be done.

not really - what I am proposing is to leave meters as defaults and just add the possibility to have parameters in degrees
you write in the docs that reconstructors expect support only meters and there is no problem...

anyway do as you please , I just wanted to avoid duplicating code between PRs

maxnoe:
I thought we should first merge the other PR adding the possibility to use them in degrees and than this one making it the default. But we could also change the target branch and do them in one go. Merging this into #1408 and not into master.

My main point is: the default should work with the reconstructors.

LukasNickel:
> I thought we should first merge the other PR adding the possibility to use them in degrees and than this one making it the default. But we could also change the target branch and do them in one go. Merging this into #1408 and not into master.

Maybe this is the important point. Is there any work, that still needs to be done in #1408 or just reviews and clean up and is there any reason to stall both PRs any longer (as in first releasing another release without this breaking change)?

ctapipe/image/hillas.py
```diff
@@ -197,9 +194,24 @@ def hillas_parameters(geom, image):
...
+            width=u.Quantity(width, unit),
+            width_uncertainty=u.Quantity(width_uncertainty, unit),
+            psi=Angle(psi, unit=u.rad),
+            skewness=skewness_long,
+            kurtosis=kurtosis_long,
+        )
     return HillasParametersContainer(
-        x=u.Quantity(cog_x, unit),
-        y=u.Quantity(cog_y, unit),
+        fov_lon=u.Quantity(cog_x, unit),
```
maxnoe, line 213: Please double check that this association (fov_lon with x) is correct. I remember it the other way around but I am not sure.

LukasNickel: I think you are correct in general. Will check tomorrow if the x/y are consistent here

maxnoe: Sorry was mistaken. Or confused. 

This is correct. If a camera geometry is in telescope frame, `pix_x` refers to `TelescopeFrame.fov_lon`. 

I was confused by the fact that in the `CameraFrame`, pix_x is refers to (obviously) `CameraFrame.x` which is aligned with `fov_lat`.

HealthyPear:
> I thought we should first merge the other PR adding the possibility to use them in degrees and than this one making it the default. But we could also change the target branch and do them in one go. Merging this into #1408 and not into master.
> 
> My main point is: the default should work with the reconstructors.

If this was the idea from the beginning (merging #1408 first) then I am all for it.
I mistakenly thought the opposite, that's why I was suggesting to leave CameraFrame as default in this PR and let the current tests do their job while updating the docs accordingly.

ctapipe/io/tests/test_dl1eventsource.py
```diff
@@ -68,7 +68,9 @@ def test_simulation_info(dl1_file):
             for tel in event.simulation.tel:
                 assert tel in event.simulation.tel
                 assert event.simulation.tel[tel].true_image is not None
-                assert event.simulation.tel[tel].true_parameters.hillas.x != np.nan
+                assert (
+                    event.simulation.tel[tel].true_parameters.hillas.fov_lon != np.nan
```
maxnoe, line 72: np.isnan, you cannot compare to nan (it will always be false).

ctapipe/io/tests/test_dl1eventsource.py
```diff
@@ -82,7 +84,7 @@ def test_dl1_b_only_data(dl1_parameters_file):
     with DL1EventSource(input_url=dl1_parameters_file) as source:
         for event in source:
             for tel in event.dl1.tel:
-                assert event.dl1.tel[tel].parameters.hillas.x != np.nan
+                assert event.dl1.tel[tel].parameters.hillas.fov_lon != np.nan
```
maxnoe, line 87: same as above

ctapipe/io/tests/test_dl1eventsource.py
LukasNickel: This is actually how its done on the master, good thing you spotted it now

LukasNickel:
Not super happy with adding the `BaseHillasParametersContainer` and `BaseTimingParametersContainer` to `__all__`, but I have not found another way to build the docs.

kosack:
Yes, I think we should merge #1408 first or merge the two together, since there is some overlap, particularly since this now also modifies the HillasReconstructor.  Perhaps open a third branch where the two can be integrated?

kosack (commented):
Since you retain the CameraFrame versions of everything, would it be possible to just add an option to ImageProcessor to select which frame to use? That way we can test both easily.   In #1408 there also remains he possibility to do reconstruction in both frames, so that would make this fully consistent.  

Then, in a future PR we could always clean up and remove all traces of the CameraFrame reconstruction, once we are sure everything is correct.

maxnoe:
@kosack #1408 is already merged

kosack:
> @kosack #1408 is already merged (by you 2.5 weeks ago ;))

:-) Clearly my memory is already on summer vacation.  Then yes, there is no problem with this other than maybe a switch to test out both versions

LukasNickel:
- Added an option to use the camera frame for the `ImageProcessor`. Its not available from the stage1 tool, is that a thing we want/need? I wasnt sure how that would work with the changed datamodel version.
- Merged the master and resolved some conflicts. Due to the changes in the datamodel in the master (indices lookup) this change now leads to version 2.1 instead of 1.3 of the datamodel

ctapipe/image/image_processor.py
```diff
@@ -65,6 +65,7 @@ def __init__(
         self,
         subarray: SubarrayDescription,
         is_simulation,
+        use_telescope_frame=True,
```
kosack, line 68: This should be a Traitlet, not a construction option.   Otherwise it is not configurable. The default should be set to True.

kosack (dismissed):

ctapipe/image/image_processor.py
kosack: That will not change the data model - it is only a configuration option, and the current data model supports both units in the output. However, we may in the future deprecate the CameraFrame one and clean up all that code as I said before, but now we want to be able to run the full analysis twice on a large dataset and ensure they results are compatible with both frames.

maxnoe: While we are at it, please also remove the `is_simulation` option. It should not be needed. Code that needs to check that should just check `if event.simulation is None`

maxnoe: Sorry, not really connected. Will just make a PR

kosack (approved):
I did a quick check running in both the telescope (using this version) and camera frame (using the master version) and comparing the differences.   It all looks good.  There are some slight differences, but this may be due to rounding error.  We may need to investigate a bit though, but that can come later.  For example, not sure why delta alt is not centered on 0: there is a small bias

download

LukasNickel:
Sorry, thought I pushed that yesterday.

@kosack What did you use for that comparison? Diffuse or pointlikes?

HealthyPear:
> Sorry, thought I pushed that yesterday.
> 
> @kosack What did you use for that comparison? Diffuse or pointlikes?

It shouldn't matter, but better check with both.

kosack:
> @kosack What did you use for that comparison? Diffuse or pointlikes?

point-like.  This was just a quick test to make sure it was not crazy.  We should do a proper benchmark later.

HealthyPear:
> > @kosack What did you use for that comparison? Diffuse or pointlikes?
> 
> point-like. This was just a quick test to make sure it was not crazy. We should do a proper benchmark later.

we should have a small diffuse test file with 10 showers if I recall correctly

LukasNickel:
It shouldnt, I am just curious especially if there is a bias

LukasNickel:
> we should have a small diffuse test file with 10 showers if I recall correctly

I dont think 10 showers will suffice for this kind of benchmark

kosack:
@HealthyPear  There's already a unit test for this for HillasReconstructor, `test_CameraFrame_against_TelescopeFrame` so no need here, I just wanted to be sure the mechanism all works with ctapipe-process.

kosack (approved):

maxnoe (approved):

Issue #1599, opened 2021-02-12
flake8 CI


nbiederbeck:
After yesterdays developer seminar I linted ctapipe with flake8 as proposed by [github.com/action/starter-workflows](https://github.com/actions/starter-workflows/blob/c59b62dee0eae1f9f368b7011cf05c2fc42cf084/ci/python-package.yml)

```yaml
    - name: Lint with flake8
      run: |
        # stop the build if there are Python syntax errors or undefined names
        flake8 . --count --select=E9,F63,F7,F82 --show-source --statistics
        # exit-zero treats all errors as warnings. The GitHub editor is 127 chars wide
        flake8 . --count --exit-zero --max-complexity=10 --max-line-length=127 --statistics
```

The results are huge (2 and 61 warnings/errors)
```
$ flake8 . --count --select=E9,F63,F7,F82 --show-source --statistics
./ctapipe/utils/unstructured_interpolator.py:96:31: F632 use ==/!= to compare constant literals (str, bytes, int, float, tuple)
        if self._remember and self._previous_v is not 0:
                              ^
./ctapipe/tests/setup_package.py:2:13: F821 undefined name '_ASTROPY_PACKAGE_NAME_'
    return {_ASTROPY_PACKAGE_NAME_ + ".tests": ["coveragerc"]}
            ^
1     F632 use ==/!= to compare constant literals (str, bytes, int, float, tuple)
1     F821 undefined name '_ASTROPY_PACKAGE_NAME_'
2

$ flake8 . --count --exit-zero --max-complexity=10 --max-line-length=127 --statistics
./ctapipe/containers.py:886:60: E241 multiple spaces after ','
./ctapipe/reco/hillas_intersection.py:60:19: E231 missing whitespace after ','
./ctapipe/reco/ImPACT.py:364:5: C901 'ImPACTReconstructor.get_likelihood' is too complex (11)
./ctapipe/reco/ImPACT.py:404:9: F841 local variable 'azimuth' is assigned to but never used
./ctapipe/reco/ImPACT.py:834:13: F841 local variable 'migrad' is assigned to but never used
./ctapipe/reco/tests/test_HillasReconstructor.py:168:50: W291 trailing whitespace
./ctapipe/reco/tests/test_HillasReconstructor.py:170:61: W291 trailing whitespace
./ctapipe/reco/tests/test_HillasReconstructor.py:203:13: F841 local variable 'e' is assigned to but never used
[...]

PR #1600, opened 2021-02-12
Add black code formatter to environment.

nbiederbeck:
See: https://github.com/cta-observatory/ctapipe/issues/1599#issuecomment-778081216

We add the pre-commit hook which uses the black code formatter, but did not include it in the conda environment

### DIFF
--- a/environment.yml
+++ b/environment.yml
@@ -8,6 +8,7 @@ dependencies:
   - pip
   - astropy
   - h5py=2
+  - black
   - bokeh=1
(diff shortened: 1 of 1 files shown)


kosack:
For me, the one installed by pre-commit still gets installed even if one exists in the environment, but it's true we should have it there, so devs can run it manually.

kosack (approved):

maxnoe (approved):

Issue #1597, opened 2021-02-11
Error in stage1 processing


LukasNickel:
This has the same root cause as #1411 and #1480 (float32 vs np.nan), but since my symptoms differ, I decided to open a new issue.

When using `ctapipe-stage1` with an `ImageQualityQuery`, that actually removes events (such as in the example configuration), the default container is used for these images, which contains `np.nan` values.
Trying to append such a row to an existing table with data type `float32` fails with the following message:

```python
2021-02-11 13:58:13,425 ERROR [ctapipe.ctapipe-stage1.DL1Writer.HDF5TableWriter] (hdf5tableio._append_row): Error writing col intensity_max of container IntensityStatisticsContainer
2021-02-11 13:58:13,425 ERROR [ctapipe.ctapipe-stage1] (tool.run): Caught unexpected exception: invalid type (<class 'float'>) for column ``intensity_max``
Traceback (most recent call last):
  File "tables/tableextension.pyx", line 1584, in tables.tableextension.Row.__setitem__
TypeError

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "/home/lukas/git_repos/ctapipe/ctapipe/core/tool.py", line 260, in run
    self.start()
  File "/home/lukas/git_repos/ctapipe/ctapipe/tools/stage1.py", line 137, in start
    self.write_dl1(event)
  File "/home/lukas/git_repos/ctapipe/ctapipe/io/dl1writer.py", line 209, in __call__
    self._write_telescope_events(self._writer, event)
  File "/home/lukas/git_repos/ctapipe/ctapipe/io/dl1writer.py", line 491, in _write_telescope_events
    *event.simulation.tel[tel_id].true_parameters.values(),
  File "/home/lukas/git_repos/ctapipe/ctapipe/io/hdf5tableio.py", line 317, in write
    self._append_row(table_name, containers)
  File "/home/lukas/git_repos/ctapipe/ctapipe/io/hdf5tableio.py", line 287, in _append_row
    row[colname] = value
  File "tables/tableextension.pyx", line 1589, in tables.tableextension.Row.__setitem__
TypeError: invalid type (<class 'float'>) for column ``intensity_max``
```

No more events are written after this.

This is on python 3.7.9 with pytables 3.6.1

kosack:
CAn you say more about how to reproduce this? Is it when running `ctapipe-merge` or when running `ctapipe-stage1`?

LukasNickel:
[...]

PR #1601, opened 2021-02-12
Match data types of default values and calculated parameters

LukasNickel:
Quick and easy fix for #1597 and connected issues.
It is still not apparent to me why we use single precision at one place and double precision at another though.

### DIFF
--- a/ctapipe/containers.py
+++ b/ctapipe/containers.py
@@ -137,12 +137,12 @@ class LeakageContainer(Container):
         nan, "fraction of pixels after cleaning that are in camera border of width=2"
     )
     intensity_width_1 = Field(
-        nan,
+        np.float32(nan),
         "Intensity in photo-electrons after cleaning"
         " that are in the camera border of width=1 pixel",
     )
     intensity_width_2 = Field(
-        nan,
+        np.float32(nan),
         "Intensity in photo-electrons after cleaning"
         " that are in the camera border of width=2 pixels",
     )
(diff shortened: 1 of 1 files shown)


maxnoe:
Will depend on the input types sometimes. Quantities will cast everything to float64 I think by default. If it isn't a quantity, it probaly inherits the type when using plain numpy transformations.

maxnoe:
I would rather be consistent here.

 I don't think we have any image parameters where we need 64 bits. So let's make sure all are 32? @kosack?

For example in `leakage_parameters`, that would mean doing `

```
return LeakageContainer(
    pixels_width_1=(leakage_pixel1 / geom.n_pixels).astype(np.float32),
   ...
)
```

However, as stated before, I would probably rather implement this by allowing `Field.validate` to cast to the dtype.

maxnoe:
Which can be done using `astype(casting='same_type')` which only allows casts between different sizes of the same type, so e.g. `float32 ↔ float64` but not `int64 ↔ float64`

maxnoe:
I was wrong about quantity, it preserves float dtype:

```
In [10]: image = np.random.normal(size=1855).astype('float32')
In [11]: u.Quantity(image.sum(), u.m).dtype
Out[11]: dtype('float32')
```

LukasNickel:
Do we need the double precision during the calculation?
If not we can just remove the upcasting in the hillas parameters and only worry about numba code

LukasNickel:
Although field.validate seeme like the nicest way to handle this.

maxnoe:
Yes, I think internally using `flaot64` for all image parameters and then when writing casting to `float32` makes most sense

LukasNickel:
Should we upcast in the parameter function itself, like here:
 https://github.com/cta-observatory/ctapipe/blob/23bf35655f9d6bd0215dba024cadc09187c2ac40/ctapipe/image/hillas.py#L111-L113

or would it be easier to upcast the image once in the `ImageProcessor` and remove that code?

maxnoe:
Let's merge this as is for a hotfix so it can go into 0.10.2 and do the conversion stuff for 0.11.

THis way, we might get it to work today for johan to start processing the data on the grid.

maxnoe (approved):

kosack (approved):
it's strange this didn't happen earlier (I never saw it in any of my tests or large productions), but it clearly is wrong.  Maybe an older version of some dependency was more lenient.

PR #1603, opened 2021-02-12
Update stage1 example config with new cleaning levels

maxnoe:


bregeon:
may I ask what happens to the FlashCam cleaning with that config file?
it's default 10/10?

maxnoe:
Good catch, should have been 10/5

bregeon:
ok, that matches what we discussed on Slack.
May I just ask also why having a default instead of defining explicitly numbers for FlashCam as well?

bregeon:
also did we discuss the ImageQualityQuery parameters, are they ok?

kosack (dismissed):

kosack:
> also did we discuss the ImageQualityQuery parameters, are they ok?

Those are pretty reasonable.  Below 50 PE, you get lots of crazy systematics, so as a default/example it's fine. The user can of course choose what they like

maxnoe:
> also did we discuss the ImageQualityQuery parameters, are they ok?

Also, these are only relevant for the parameters, all images are stored.

bregeon (dismissed):
Looks good to me now.

bregeon (approved):
still good for me

maxnoe:
> also did we discuss the ImageQualityQuery parameters, are they ok?

Also, these are only relevant for the parameters, all images are stored.

maxnoe:
@kosack approve again? Then we have this in 0.10.3

HealthyPear (approved):
what about the other SSTs? I thought there were 3 of them in total, but perhaps something changed recently and I didn't follow

maxnoe:
@HealthyPear Prod5 has only one SST.

PR #1604, opened 2021-02-12
Fix error when my true image does not meet quality query

maxnoe:


kosack (approved):

kosack:
Should update the PR description so we don't forget why this was changed:  this caused an exception in the HDF5TableWriter when a true_image couldn't be parameterized, since the dtype of the value to put in the column changed.  This PR fixes that case.

maxnoe:
Done

HealthyPear (approved):

PR #1605, opened 2021-02-12
fix read_table

### LINKED ISSUE #1576: h5_table_to_astropy / read_table loses some column metadata
# Problem

Since we return an `astropy.table.QTable` instead of `astropy.Table` in `h5_table_to_astropy()` (which is also aliased to `read_table()`), any column that contains a unit loses its description metadata, since those columns are stored as astropy `Quantity` arrays and not `Column` objects. 

```py3

params = read_table("events.h5", "/dl1/event/telescope/parameters/tel_001")

print(params.columns['obs_id'].description)
print(params.columns['hillas_width'].description)
```
```
observation identifier
AttributeError: 'Quantity' object has no 'description' member
```

Also writing these back out to a file, we see that all the description metadata is lost whenever there is a Quantity column:

```py3
params.write("test.ecsv")
```
```sh
% head test.ecsv

# %ECSV 0.9
# ---
# datatype:
# - {name: obs_id, datatype: int32, description: observation identifier}
# - {name: event_id, datatype: int64, description: event identifier}
# - {name: tel_id, datatype: int16, description: telescope identifier}
# - {name: hillas_intensity, datatype: float64, description: total intensity (size)}
# - {name: hillas_x, unit: m, datatype: float64}
# - {name: hillas_y, unit: m, datatype: float64}
# - {name: hillas_r, unit: m, datatype: float64}
```

A second issue that that QTable is missing some features of Table, like aggregation of grouped values.
```py3
params.group_by( "tel_id").groups.aggregate(np.mean)  # fails for QTable, but works for Table
```

# Possible Solution:

Return a  `Table` instead of `QTable` by default.

This makes it slightly less nice to access quantity members with units, since by default the unit is not carried over unless the `quantity` attribute is used:

```py3
np.nanmean(params['hillas_width'])  # returns a bare float, "0.2"
np.nanmean(params['hillas_width'].quantity) # returns a quantity  "0.2 m"
```

But it does not lose any information.

kosack:
fixes #1576 

read_table returned *QTables* instead of *Tables*, which lead to lots of confusing and annoyances when working with them, since some columns worked differently than others and metadata being lost.  This way all columns in the tables have a consistent interface. 

This just fixes that to return Table (the user can still convert to QTable if needed).  Technically this is an API change, but since the behavior was bug-like, I'm marking it as a bug to be included in v0.10.3. 

Also renamed the function to `read_table` everywhere.  It was already called `ctapipe.io.read_table`, but was aliased as read_table while privately name `h5_table_to_astropy`, which could be confusing. So no change of API will happen there, as h5_table_to_astropy was undocumented.

### DIFF
--- a/ctapipe/io/astropy_helpers.py
+++ b/ctapipe/io/astropy_helpers.py
@@ -6,7 +6,7 @@ Functions to help adapt internal ctapipe data to astropy formats and conventions
 from pathlib import Path
 
 import tables
-from astropy.table import QTable
+from astropy.table import Table
 from astropy.units import Unit
 from astropy.time import Time
 import numpy as np
@@ -21,7 +21,7 @@ from .hdf5tableio import get_hdf5_attr
 __all__ = ["h5_table_to_astropy"]
 
 
-def h5_table_to_astropy(h5file, path) -> QTable:
+def read_table(h5file, path) -> Table:
     """Get a table from a ctapipe-format HDF5 table as an `astropy.table.QTable`
     object, retaining units. This uses the same unit storage convention as
     defined by the `HDF5TableWriter`, namely that the units are in attributes
@@ -37,7 +37,7 @@ def h5_table_to_astropy(h5file, path) -> QTable:
 
     Returns
     -------
-    astropy.table.QTable:
+    astropy.table.Table:
         table in Astropy Format
 
     """
@@ -85,7 +85,7 @@ def h5_table_to_astropy(h5file, path) -> QTable:
             value = table.attrs[attr]
             other_attrs[attr] = str(value) if isinstance(value, np.str_) else value
 
-    astropy_table = QTable(table[:], meta=other_attrs)
+    astropy_table = Table(table[:], meta=other_attrs)
 
     for column, tr in column_transforms.items():
         astropy_table[column] = tr.inverse(astropy_table[column])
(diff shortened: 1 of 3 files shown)


maxnoe:
@kosack you forgot an import in a test

maxnoe (approved):

ctapipe/io/astropy_helpers.py
```diff
@@ -21,7 +21,7 @@
 __all__ = ["h5_table_to_astropy"]
 
 
-def h5_table_to_astropy(h5file, path) -> QTable:
+def read_table(h5file, path) -> Table:
     """Get a table from a ctapipe-format HDF5 table as an `astropy.table.QTable`
```
HealthyPear, line 25: Probably a picky comment:
technically now this docstring is not completely correct, since what we really return now is a **Table** with the units stored in attributes as metadata

HealthyPear (changes requested):

ctapipe/io/astropy_helpers.py
maxnoe: This is not a picky comment but a good catch of lying documentation.

maxnoe: fixed

maxnoe (approved):

HealthyPear (approved):

PR #1614, opened 2021-02-18
Fix dl1 merge tool not writing correct metadata / stage1 not being able to read merged files

maxnoe:
The subarray pointing table was missing metadata and the merge tool
always set the product version to 1.0, also when merging newer files.

* Check and use CTA PRODUCT DATA MODEL VERSION when merging
* Remove fix for merging ctapipe 0.8 files that lead to the missing
metadata
* Actually run stage1 on the merged file result

This shoudl fix the dl1 merging.

I am not really happy with the tests yet for merging, we at least two runs from the same production to test properly.


ctapipe/utils/index_finder.py
```diff
@@ -20,8 +20,7 @@ class IndexFinder:
     """
 
     def __init__(self, values):
-        if not len(np.unique(values)) == len(values):
-            raise Exception("values contains duplicate entries!")
+        values = np.unique(values)
```
LukasNickel, line 23: This line is not needed / breaks the test.

ctapipe/utils/index_finder.py
```diff
@@ -10,7 +10,7 @@ class IndexFinder:
     Helper class to find the index of the closest matching value in an array/list/...,
     used to locate the pointing of an event based on the trigger time.
     This searches using pythons bisect module.
-    All entries of `values` need to be unique.
+    Duplicated keys will be ignored.
```
LukasNickel, line 13: Values rather than keys?

Also I would add an warning here to avoid hard to find bugs if someone uses this class for other reasons

LukasNickel (changes requested):

maxnoe:
@kosack @LukasNickel @Hckjs I think this is ready now, please review.

Hckjs:
Don't need that anymore then, do we?

https://github.com/cta-observatory/ctapipe/blob/fb9324a762e73fae715982f7d52b364b508ce3a2/ctapipe/tools/dl1_merge.py#L347-L355

maxnoe:
Ah right, yes.

LukasNickel (approved):

Hckjs (approved):

Issue #1353, opened 2020-05-21
merge stage1 and reconstruct-muons


kosack:
There's a lot of overlap between the tools  `ctapipe-stage1-process` and `ctapipe-reconstruct-muons`. Both output DL1 data, with parameters being one of the options.    

I suggest we just drop `ctapipe-reconstruct-muons` and add an option like `--write-muon-parameters=True` (similar to `--write-parameters`) that optionally does the muon detection/reconstruction.   This way the same tool can be used for both purposes. 

That way, the output files are identical (now they differ), and one can even store both muon parameters and images for testing purposes 

`--write-images=True --write-parameters=False --write-muon-parameters=True` would then produce a DL1 file with simulation info, calibrated images, and muon parameters (can even keep the other parameters enabled as well).

MaxNoe:
Agreed

moralejo:
We did a similar thing recently in lstchain. Keep in mind that probably for muon rings the global peak integrator is better: the signal won't have a bias in pixels containing only NSB. So what we did is to re-integrate the waveforms with the same window parameters (width & shift), but using GlobalPeak integrator, whenever an image fulfilled the "muon candidate conditions". With global peak integrator the comparison to MC is more robust against differences in the NSB level.

moralejo:
Yes, certainly we will eventually converge with the ctapipe DL1 stage tool, including the standardized muon analysis. But an intermediate "checkpoint" in which we adopt the standard DL1 format, but keep our current muon analysis, will be helpful I think.

kosack:
PR #1499 does a lot towards making this easy.  I think we just need to create a new class like the ImageProcessor defined there  for muon analysis (MuonProcessor?), and then we call it if that option is enabled.  We will need somewhere in the ArrayEvent data structure to put the muon parameters and optical efficiency parameters , so perhaps under `event.dl1.tel[x].muon.ring_parameters` and `event.dl1.tel[x].muon.efficiency_parameters`, or we add them to the ImageParameters container (but then have to ignore them when not filled)

maxnoe:
agreed.

@moralejo stressed that the muons should have there own extractor, always using `GlobalPeakWindowSum` to minimize noise.

And the muon analysis should only run after passing some relatively hard `ImageQuery` cuts, but I think that's all rather easy with a structure like that.

kosack:
[...]

PR #1618, opened 2021-02-18
Outsourced muon processor

RuneDominik:
Currently, the muon reconstruction tool has most of the code necessary for the muon event procession build directly into the tool itself instead of relying on an external processor as the stage1 tool does with its `image_processor`. Outsourcing the event procession to an external processor would be a necessary step to include the muon analysis into the standard event procession at a later point. This is a first draft of such a processor. All heavy lifting code has been been copied over from the old tool.


ctapipe/image/muon/muon_processor.py
```diff
@@ -0,0 +1,275 @@
+"""
+High level image processing  (ImageProcessor Component)
```
maxnoe, line 2: copy paste?

ctapipe/image/muon/muon_processor.py
```diff
@@ -0,0 +1,275 @@
...
+
+    ratio_width = traits.FloatTelescopeParameter(
+        default_value=1.5,
+        help=(
+            "Ring width for intensity ratio"
+            " computation as multiple of pixel diameter"
+        ),
+    ).tag(config=True)
+
+    overwrite = traits.Bool(
```
maxnoe, line 43: The processor shouldn't need to overwrite anything

ctapipe/image/muon/muon_processor.py
```diff
@@ -0,0 +1,275 @@
...
+    ).tag(config=True)
+
+    pedestal = traits.FloatTelescopeParameter(
+        help="Pedestal noise rms", default_value=1.1
+    ).tag(config=True)
+
+    def __init__(
+        self,
+        subarray: SubarrayDescription,
+        is_simulation,
```
maxnoe, line 62: There is no need to give this information I think. You can just check `if event.simulation is not None`

kosack: I think I did that in stage1 since I didn't have the event yet and wanted to set up things

kosack:
This is a good idea to make a Component for muon analysis like ImageProcessor.  The other thing that was missing (and why I didn't add it to stage1 from the start) was a place to put the muon  information in the event structure and in the output data model written by the DL1Writer.  

One simple option would be to add these parameters to the ImageParametersContainer, and then they will end up in `/dl1/event/telescope/parameters/tel_xxx` along with hillas and other parameters.   

 But since most events do not contain a muon, it may be better to split into two data streams: not writing muon parameters for all events, but only for selected events, perhaps in `/dl1/event/telescope/muon_parameters/tel_XXX` 

This of course depends a lot on the expected use case and work flow.  The requirements for CTA are that cameras already flag muon-like events, and ACADA will write those to a separate stream that can be processed with something like ctapipe-stage1 with muon analysis enabled. In that case, just writing muon info for all events is fine.    However, if this is run with LST data that so far does not have the muon pre-selection feature, then writing to a separate dataset is perhaps best,

moralejo:
> 
> But since most events do not contain a muon, it may be better to split into two data streams: not writing muon parameters for all events, but only for selected events, perhaps in `/dl1/event/telescope/muon_parameters/tel_XXX`
> 

I think that is the reasonable thing to do, and it will be fine also for the case of separate analysis of the muon ring stream, when we have it.

Related question: what will happen with muon ring images which are part of a stereo event? (which in the case of LSTs is all of them, due to the hardware stereo trigger). Will they be written to both streams, the normal one and the muon one? Or only muon rings which trigger a single telescope will go to the dedicated stream?

kosack:
> Related question: what will happen with muon ring images which are part of a stereo event? (which in the case of LSTs is all of them, due to the hardware stereo trigger). Will they be written to both streams, the normal one and the muon one? Or only muon rings which trigger a single telescope will go to the dedicated stream?

Actually that's a good question - I think the muon streams are meant to only be for cameras where you don't get enough muons in regular stereo triggers, so mainly for the SSTs.  For LSTs, you likely get enough muons mixed into stereo events.   So that answers my question above: we should put them in muon_parameters, so we can run the muon detection on stereo data or pre-selected muon streams in the same way.   Alternately, if LSTs also flag mono muon events into a separate muon stream (with no shower inside), then we don't really need to even analyze the mixed muon-shower events, and those can just be rejected in the analysis.  In any case, it's good to support both for the time being.

maxnoe:
@RuneDominik General comment: I think this processor shouldn't do IO. Rather, add a method like `write_muon_information` to the `DL1Writer` and let this `Processor` just fill the muons containers.

ctapipe/image/muon/muon_processor.py
```diff
@@ -0,0 +1,275 @@
...
+        help=(
+            "Minimum number of pixels after cleaning and ring finding"
+            "required to process an event"
+        ),
+        default_value=100,
+    ).tag(config=True)
+
+    pedestal = traits.FloatTelescopeParameter(
+        help="Pedestal noise rms", default_value=1.1
+    ).tag(config=True)
```
kosack, line 57: Is this really needed as a parameter? Shouldn't it come from the monitoring info, as it changes over time during an observation and is pixel-dependent.

ctapipe/image/muon/muon_processor.py
RuneDominik: As far as I know, this parameter is not used any further in the Processor (and should be recreatable from the EventSource at any time), so I removed it for now.

ctapipe/image/muon/muon_processor.py
RuneDominik: This was taken from the earlier iteration of the muon analysis, but changing it to a pixel and time dependent pedestal should be the correct thing to do, at least on the long run (I don't know if the current muon simulations provide this information). How is this information stored?

RuneDominik:
> @RuneDominik General comment: I think this processor shouldn't do IO. Rather, add a method like `write_muon_information` to the `DL1Writer` and let this `Processor` just fill the muons containers.

@maxnoe There is a first version of outsourcing the io to the dl1writer in the latest commits, following the scheme used for the "normal" image procession. Some changes for the containers were needed as well, as the muon containers were not saved to the event and rather written after creation.

ctapipe/image/muon/muon_processor.py
maxnoe: It should in principle be in `event.monitoring.tel[tel_id].pedestal`. But I don't think that we fill that with sensical values right now.

I also a bit confused of the different "pedestals" here. The monitoring pedestal should contain the average NSB in each pixel, right? And the `calibration.tel[dl1].pedestal_offset` is what is more commonly known as "baseline"?

ctapipe/io/dl1writer.py
```diff
@@ -468,6 +487,34 @@ def fill_from_simtel(
...
+            telescope = self._subarray.tel[tel_id]
+            self.log.debug("WRITING TELESCOPE %s: %s", tel_id, telescope)
+
+            tel_index = TelEventIndexContainer(
+                obs_id=event.index.obs_id,
+                event_id=event.index.event_id,
+                tel_id=np.int16(tel_id),
+            )
+
+            if dl1_camera.muon_parameters is not None:
```
maxnoe, line 508: this if should be first to not do operations like creating the container when it is not necessary.

You can avoid the additional indentation level by doing:

```
if dl1_camera.muon_parameters is None:
    continue
```
first

ctapipe/io/dl1writer.py
```diff
@@ -468,6 +487,34 @@ def fill_from_simtel(
...
+                writer.write(
+                    table_name=f"dl1/event/telescope/parameters/muons",
+                    containers=[tel_index, *dl1_camera.muon_parameters.values()],
+                )
+
+        writer.write(
```
maxnoe, line 514: If this is here, is it written twice if we save both muon parameters and images / normal image parameters?

Also it looks like an "if is_simulation" is missing

ctapipe/tools/muon_reconstruction.py
```diff
@@ -24,65 +23,43 @@ class MuonAnalysis(Tool):
...
         MuonIntensityFitter,
+        DL1Writer,
     ]
 
     aliases = {
         "i": "EventSource.input_url",
         "input": "EventSource.input_url",
-        "o": "MuonAnalysis.output",
-        "output": "MuonAnalysis.output",
+        "o": "DL1Writer.output_path",
```
maxnoe, line 39: I think the correct form of adding long / short forms is to use a tuple: `("i", "input")`, have to fix that for the other tools as well.

ctapipe/io/dl1writer.py
RuneDominik: In the current form, a is_muon-flag is set on writer creation, so one instance can only write muons or regular events and this problem should not occure (General Question: Is this wanted behaviour or should one writer instance be able to write both types?). However, since this information is written for both types, having these lines in `__call__` makes things easier.

maxnoe: > General Question: Is this wanted behaviour or should one writer instance be able to write both types 

Yes, we should not have two writer instances writing to the same file. I think this does not even work, since hdf actually blocks files for writing.


Just adding a `write_muon_parameters(self, event)` method seems best to me.

kosack: `simulation/event/subarray/shower` is where the simulation parameters should go, but you shouldn't need to touch that in the muon component - the DL1Writer will put it in the correct place. 

Basically, the old ctapipe-reconstruct-muons script did not follow any consistant data model, so there is no need to have a new tool reproduce the same format.

ctapipe/io/dl1writer.py
```diff
@@ -151,7 +151,14 @@ class DL1Writer(Component):
     peak_time_offset = Int(default_value=0).tag(config=True)
     peak_time_scale = Float(default_value=100.0).tag(config=True)
 
-    def __init__(self, event_source: EventSource, config=None, parent=None, **kwargs):
+    def __init__(
+        self,
+        event_source: EventSource,
+        config=None,
+        parent=None,
+        is_muon=False,
```
kosack, line 159: I don't think you need this attribute. The DL1Writer should never care if this is a muon run or not. It should write muon data if there are muon data in the event, if if not, it skips. it.

kosack:
I think there is a bit of a misunderstanding here: *please do not change DL1Writer in the way you are doing now*.  It should not write out a separate file for muons.  It should be exactly as it was before, but have one additional call to writer.write() that writes the muon parameters, just like where regular hillas parameters are written.  That requires only  1-2 line change.  

The idea is that you write *exactly the same format DL1 file*, and just include an extra dataset if there is muon information.

maxnoe:
@kosack I just discussed the same with @RuneDominik offline, he will make the changes accordingly

RuneDominik:
As discussed with @maxnoe , the latest commit reverted all changes to the dl1writer besides the creation of a `write_muon_events` function that handles all IO operation specific to muons. As a next step, this function could easily be called additionally from the strage1 tool or at each writer call for events that are identified as muons but I suppose such changes would be beyond this PRs scope. This approach should, in the future, allow for writing additional muon information and normal event information to the same file as @kosack suggested.

kosack:
> At this point however, the output is equal to the output of the current muon analysis tool

There is no need for that - in fact, it's much better to use the same format as DL1, as this is DL1 information, and that will make it possible to use the same tools and benchmark scripts to process any DL1 file.  The only reason it was different with the old muon tool was that the DL1 format did not yet exist when it was written.

RuneDominik:
The latest commit introduces the muon analysis into the stage1 tool. Using a new flag, the stage1 tool applies the muon analysis either on all or none events. In the case that muons should be analyzed, the parameters are written to `/dl1/event/telescope/muon_parameters/tel_XXX` (which is, as I understand, one conclusion from the discussion of @kosack  and @moralejo  above), not touching the rest of the IO (meaning, that e.g. hillas parameters are written for muons as well).
To achieve the needed muon-IO, the previously introduced function `write_muon_events` of the DL1Writer instance is called, however, including the muon-IO to `__call__` should be possible if needed/requested. Having the muon analysis in stage1 should intrinsically ensure that the output is a correct DL1 file. This IO function checks if muon parameters were written. In case that e.g. the ring fit fails, no parameters are recorded. It might be reasonable to fill the table with NaNs (or equal) in this case to keep this information.

With this change, the old muon-reconstruction tool might become obsolete. The test `ctapipe/tools/tests/test_tools.py::test_muon_reconstruction` is expected to fail, as this test still tries to find muon parameters in the old group `dl1/event/telescope/parameters/muons`.

I think the next big step would then be to implement a muon selection component so that the muon analysis does not need to be run on all events.

ctapipe/image/muon/muon_processor.py
```diff
@@ -0,0 +1,256 @@
...
+        if np.isnan(
+            [ring.radius.value, ring.center_x.value, ring.center_y.value]
+        ).any():
+            self.log.debug(
+                f"Skipping event {event_id}-{tel_id}: Ring fit did not succeed"
+            )
+            return
+
+        parameters = self.calculate_muon_parameters(tel_id, image, dl1.image_mask, ring)
+        # intensity_fitter does not support a mask yet, set ignored pixels to 0
```
maxnoe, line 152: Can you fix this? This is already implemented. So just pass the mask to the fitter here.

maxnoe: instead of zeroing the image

kosack:
> I think the next big step would then be to implement a muon selection component so that the muon analysis does not need to be run on all events.

FOr the muon selection, please use a `ctapipe.core.qualityquery` , which allows you to assign a user-configurable set of cuts, and it can record the statistics and also generate a cut mask for each event if you want

maxnoe:
@RuneDominik tests are failing since the muon tool test still assumes the older output format, should update the test.

While you are at it, can you also move the tests for the muon tool to `ctapipe/tools/tests/test_muon.py`? The `test_tools` has gotten way too big.

ctapipe/image/muon/muon_processor.py
```diff
@@ -0,0 +1,255 @@
...
+            Configuration specified by config file or cmdline arguments.
+            Used to set traitlet values.
+            This is mutually exclusive with passing a ``parent``.
+        parent: ctapipe.core.Component or ctapipe.core.Tool
+            Parent of this component in the configuration hierarchy,
+            this is mutually exclusive with passing ``config``
+        """
+
+        super().__init__(subarray=subarray, config=config, parent=parent, **kwargs)
+        self.subarray = subarray
```
maxnoe, line 78: This line should be unnecessary, as handled by `super().__init__()`

ctapipe/image/muon/muon_processor.py
```diff
@@ -0,0 +1,255 @@
...
+
+        self.source = EventSource(parent=self)
```
maxnoe, line 80: Why is an event source needed here?

ctapipe/image/muon/muon_processor.py
```diff
@@ -0,0 +1,255 @@
...
+        subarray = self.source.subarray
```
maxnoe, line 81: Again overwriting subarray.

ctapipe/image/muon/muon_processor.py
```diff
@@ -0,0 +1,255 @@
...
+
+        self.pixels_in_tel_frame = {}
+        self.field_of_view = {}
+        self.pixel_widths = {}
+
+        for p in ["min_pixels", "pedestal", "ratio_width", "completeness_threshold"]:
+            getattr(self, p).attach_subarray(self.source.subarray)
+
+    def __call__(self, event: ArrayEventContainer):
+        self._process_array_event(event)
```
maxnoe, line 96: Unnecessary method. Just implement `self_process_array_event` here.

ctapipe/image/muon/muon_processor.py
```diff
@@ -0,0 +1,255 @@
...
+            border = cam.get_border_pixel_mask()
+
+            x, y = self.get_pixel_coords(tel_id)
+            self.field_of_view[tel_id] = np.sqrt(x[border] ** 2 + y[border] ** 2).mean()
+
+        return self.field_of_view[tel_id]
+
+    def get_pixel_width(self, tel_id):
+        """Guesstimate fov radius for telescope with id `tel_id`"""
+        # memoize fov calculation
```
maxnoe, line 233: Could you fix my copy/paste error here?

maxnoe (changes requested):
Looks good aside from the two points

moralejo:
Looks good indeed, thanks for this development.

In the current muon ring analysis code in lstchain we set to 0 the charge of pixels which have been declared unusable by the calibration, as stored in `event.mon.tel[telescope_id].calibration.unusable_pixels`. Actually, for the muon ring maximum likelihood fit, such pixels should be simply ignored. I don't think that is the case now, right?

ctapipe/image/muon/muon_processor.py
```diff
@@ -218,9 +211,9 @@ def calculate_muon_parameters(self, tel_id, image, clean_mask, ring):
 
     def get_fov(self, tel_id):
         """Guesstimate fov radius for telescope with id `tel_id`"""
-        # memoize fov calculation
+        # memorize fov calculation
```
maxnoe, line 214: That was correct: https://en.wikipedia.org/wiki/Memoization

ctapipe/image/muon/muon_processor.py
```diff
@@ -230,7 +223,7 @@ def get_fov(self, tel_id):
 
     def get_pixel_width(self, tel_id):
         """Guesstimate fov radius for telescope with id `tel_id`"""
-        # memoize fov calculation
```
maxnoe: The comment was regarding the wrong docstring and what is calculated (fov instead of pixel width)

ctapipe/image/muon/muon_processor.py
```diff
@@ -230,7 +223,7 @@ def get_fov(self, tel_id):
...
+        # memorize fov calculation
```
maxnoe, line 226: again, memoization is the correct term here

ctapipe/image/muon/muon_processor.py
```diff
@@ -239,9 +232,9 @@ def get_pixel_width(self, tel_id):
 
     def get_pixel_coords(self, tel_id):
         """Get pixel coords in telescope frame for telescope with id `tel_id`"""
-        # memoize transformation
+        # memorize transformation
```
maxnoe, line 235: again, memoization is the correct term here

ctapipe/image/muon/muon_processor.py
maxnoe: Bump. Just remove the zeroing below and add `mask=mask` to the `intensitity_fitter` call below.

RuneDominik:
Now with the right correction. Regarding @moralejo comment. These pixels are indeed not set to zero in the muon analysis, so if they survive the cleaning they are used in the likelihood fit. But if these pixel should not be used for any analysis, disabling them at an earlier stage might be best, instead of disabling them for the muon analysis only.

maxnoe:
@RuneDominik Agreed, this is something not currently handled in ctapipe, so we should do that in a separate PR for everything, not just the muons.

ctapipe/image/muon/muon_processor.py
RuneDominik: This should allready have been implemented in f09c437 or did I have missed something here?

RuneDominik:
General Question: Should the old muon analysis tool be kept or deleted, as the stage1 tool can now handle the muon analysis? Deleting the old tool would make this a breaking change, although I do not know how widely used this old feature is.

maxnoe (dismissed):

maxnoe:
@RuneDominik Yes, please remove the old tool.

ctapipe/image/muon/muon_processor.py
```diff
@@ -68,71 +55,48 @@ class MuonAnalysis(Tool):
...
+            Used to set traitlet values.
+            This is mutually exclusive with passing a ``parent``.
+        parent: ctapipe.core.Component or ctapipe.core.Tool
+            Parent of this component in the configuration hierarchy,
+            this is mutually exclusive with passing ``config``
+        """
+
+        super().__init__(subarray=subarray, config=config, parent=parent, **kwargs)
+
+        self.calib = CameraCalibrator(subarray=self.subarray, parent=self)
```
kosack, line 79: This will either overwrite the calibration or simply not run, depending on the order.  Calibration should not be part of this class - it should assume the calibration is already done in the outer event loop (which it is already), since it is schematically like this:

```py3
for event in source:
     calibrate(event)
     process_images(event)
     process_muons(event)
```

ctapipe/image/muon/muon_processor.py
```diff
@@ -68,71 +55,48 @@ class MuonAnalysis(Tool):
...
-            getattr(self, p).attach_subarray(self.source.subarray)
-
-    def start(self):
-        for event in tqdm(self.source, desc="Processing events: "):
-            self.process_array_event(event)
+            getattr(self, p).attach_subarray(self.subarray)
 
-    def process_array_event(self, event):
+    def __call__(self, event: ArrayEventContainer):
         self.calib(event)
```
kosack, line 92: see previous comment

ctapipe/tools/stage1.py
```diff
@@ -64,10 +70,15 @@ class Stage1Tool(Tool):
             {"Stage1Tool": {"progress_bar": True}},
             "show a progress bar during event processing",
         ),
+        "analyse-muons": (
```
kosack, line 73: I suggest using the `write_muons` parameter of DL1DataWriter here,  to be symmetric with how the other options are specified.  We don't have options like "analyze_parameters". I.e. don't add a parameter here, but use the one from the writer (see how it is done in the loop for image parameters).

We can rethink the options later, but I would like to keep them consistent for now.

kosack (changes requested):

ctapipe/image/muon/muon_processor.py
maxnoe: No, this is intended. We want to be able to use different calibration settings for the Muon events.

This will overwrite the images, but only after those have been written to the output file.

ctapipe/image/muon/muon_processor.py
maxnoe: Again, this is intentional and has also already been discussed before here #1353

ctapipe/image/muon/muon_processor.py
kosack: that sounds like something that will come back to bite us- if two calibrations (one hidden inside another object) overwrites the internal event structure, it means you cannot run this in any order you want, and you may end up with images that are not what you expect.  I really don't like that.

kosack: Would it not be better to just  re-do the image extraction part internally, rather than run an entire CameraCalibrator?  The idea of the calibration steps was that they were modular and could be re-used inside other steps of the analysis

maxnoe: The main point why we wanted to integrate the muon analysis here was to avoid opening the files and doing the low-level calibrations multiple times. As @moralejo pointed out, we need to use a different extraction method for the muons.

If the problem is changing the images in the structure, we can deepcopy the events in the muon processor.

ctapipe/tools/stage1.py
maxnoe: I like `analyze-muons` better, since it is not only writing that is enabled. More calculations are performed.

The writer only writes, it doesn't decided whether the parameters are calculated or not.

ctapipe/image/muon/muon_processor.py
kosack: But you *are* are doing low-level analysis multiple times,.   I do not want a DL1 data model that contains images and parameters, but the parameters are not computed from the images, but rather from some other hidden version of those image.  That is a recipe for confusion. 

I think we should have a call about this - it will not be resolved in the issue tracker.

maxnoe: > I still really don't like the idea that there are parameters computed from images that are not saved, however.

The same happens by default in stage1, parameters are stored but not images.

> Would it not be better to just re-do the image extraction part internally, rather than run an entire CameraCalibrator? 

That would mean also copying all the corrections the that happen inside the `CameraCalibrator`. I really don't see a benefit.

CameraCalibrator does only two things, DVR in R1→ DL0, and image extraction in DL0 → DL1. Sure, we only need the second part, but right now that's not possible, that would mean splitting the DVR and DL0→DL1 part into two components.

maxnoe: > But you are are doing low-level analysis multiple times,.

I am talking here about the LST R0 → R1 corrections, which are the most computationally expensive, not about the image extraction.

maxnoe: I think there are only two possible solutions, we include it here in ctapipe, or lstchain will need to do it in their own version of the stage1 tool.

Which would also already be a big improvement over the current situation, but not what we really want.

moralejo: I agree with @kosack that storing images obtained with a pulse integrator which is not the same with which image parameters  are produced is very misleading. But what is the problem with the deepcopy solution proposed by @maxnoe?  
Muon ring events would have their standard image parameters, from the standard image extraction. Besides, they would have the muon ring parameters obtained with the GlobalPeak pulse integrator. 

Eventually, one might consider to store, for muon rings, _also_ the images extracted with the GlobalPeak in a separate table. 

Is there any technical problem with the above approach?

kosack: I thought about this a bit, and there is a simple solution (and sorry to @RuneDominik  that I think it is similar to what he started with!). 

The answer is that we are really talking about two DL1 files:  one for the regular analysis + parameters as we do now, and one for the muon charge extraction + parameters.   So we should just write both separately.  That way both are correct DL1 files and the "provenance chain" of image → parameters is consistent in each.  That solves also the problem that the length of the muon parameters dataset would not have be the same as the parameter dataset, as you would only write out events for which the muon detection routine thinks there is a muon.  We have an implicit assumption in the DL1 files now that we write parameters for *all* written images, and this way we can keep that assumption. 

So just add an option to set the muon output file (unless you want to auto name it by pre-pending "muon" or something), and add a second `DL1DataWriter` called `muon_writer` where `muon_writer.write()`  is only called for muon-like events.  You can also just overwrite the DL1 container twice as you do now, but we should mention this very explicitly as a caveat in the docs so that if we decide to  change the order of components, parallelize, or something unexpected, it doesn't come back to surprise us.  Ideally do the muon version first, call `muon_writer.write()then overwrite it with the normal DL1 info. 

In the end you will just get two files, as if you ran stage1 twice, but without the extra io overhead, and although this is a bit of a hack, it could just be removed in future versions once we have explicit muon-selected data files.

ctapipe/tools/stage1.py
kosack: I agree, however right now I would like to be consistent with the other options, at least for this minor relase.  For the next major release (containing DL2 info) I am already in the process of re-designing all of the options to be consistent, and there there will be a separation between computing and writing that is clear, but that is really a separate issue that should not be solved here.

ctapipe/image/muon/muon_processor.py
maxnoe: Looking at the containers again, this should be `event.monitoring.tel[tel_id].pedestal.charge_std`, which is currently not filled by the simtel event source, since it depends on the extractor used.

This is connected to the whole calibration chain and how to perform these calibration measurements on simulations, so I think it is out of scope for this PR and we should keep this as a parameter for now until that is worked out.

RuneDominik:
After discussion with @maxnoe, the most recent commit reverts the possibility to specify a unique calibration for the muon analysis. Calling ctapipe-stage1 with --write-muons and the --no-write-parameters option (that is, in this commit, manually set until #1738 is resolved)  results in an output close to the behaviour of the old muon analysis, but, in contrast, metadata is written through the DataWriter so a more "correct" DL1 file should be created. Implementing this results in less dublicated code and can easily be extended by, e.g. a muon-preselection or reenabeling the possibility to specify muon-specific calibration options.

ctapipe/io/datawriter.py
```diff
@@ -500,6 +515,40 @@ def fill_from_simtel(
                         containers=hist_container,
                     )
 
+    def write_muon_events(self, event: ArrayEventContainer):
+        """
+        add muon entries to the event/telescope tables for each muon event
+        """
+        # perform delayed initialization on first event
+        if self._writer is None:
```
maxnoe, line 523: this shouldn't be needed anymore (and will always be False), so just remove the two lines 522ff

maxnoe (changes requested):

ctapipe/io/datawriter.py
```diff
@@ -357,6 +369,9 @@ def _setup_writer(self):
         if self.write_parameters is False:
             writer.exclude("/dl1/event/telescope/images/.*", "image_mask")
 
+        if self.write_muons is False:
```
maxnoe, line 372: I think this is not the way do do this. If `write_muons` is False, we should not do anything. The column exclusions are meant to ignore specific columns of containers we do want to write.

ctapipe/io/datawriter.py
```diff
@@ -500,6 +515,40 @@ def fill_from_simtel(
                         containers=hist_container,
                     )
 
+    def write_muon_events(self, event: ArrayEventContainer):
```
kosack, line 518: probably should make this private like the other `_write*` functions (rename to `_write_muon_events`).  We may at some point create a public API, but I'd avoid it now.

maxnoe: This is called explicitly in stage1

ctapipe/io/datawriter.py
RuneDominik: I did this primarily to suppress warnings, that the muon container is not writable as there is nothing stored in there when the muon-analysis is disabled through `write_muons` == False. Is there another way to achieve this?

ctapipe/io/datawriter.py
RuneDominik: I'd second that. However, this would then require to include `_write_muon_events` into datawriter's `__call__`. At the moment, `self.write_dl1.write_muon_events(event)` is called in stage1 to minimize interaction between those parts of the datawriter that write "normal" events and those that write muons, which implies a public function.

ctapipe/io/datawriter.py
kosack: if the muon analysis is disabled, there should be no call to write(), I would assume?

ctapipe/io/datawriter.py
maxnoe: I think this is a leftover from the different calibration discussion. You cannot have it in `__call__` if you want to be able to apply a different calibration to the muon candidates.

But since we basically removed this discussion here, let's make it consistent and add the muon writing to `__call__`

ctapipe/io/datawriter.py
RuneDominik: In my understanding, the same is done for the parameters container (alongside some other)

```python
# avoid some warnings about unwritable columns (which here are just
# sub-containers)
writer.exclude("dl1/event/subarray/trigger", "tel")
writer.exclude("dl1/monitoring/subarray/pointing", "tel")
writer.exclude("/dl1/event/telescope/images/.*", "parameters")
```
        
As I have implemented the storage for the muon_parameters in reminiscence to the "normal" parameters as sub-container of the  `DL1CameraContainer`, the same Warning for unwritable columns arises even when no muons are written. Afaik, this warning is not given when actual data is stored in this field, hence the if-statement. However, excluding it in any case seems to do no harm so I'll at least change this to be consistent to the above.

RuneDominik:
Close this in favor of #2168.

PR #1620, opened 2021-02-19
Mixed up positions in hillas_intersection direction assignment

ParsonsRD:
It looks like the direction was assigned the wrong way round to the reconstructed coordinates here. In principle errors like this should be caught with a test, the obvious one would be to just check the performance of the reconstructed directed vs the simulated one. However if you do this with the example dataset the performance seems to be pretty bad, so not a great test. Any ideas?


kosack (approved):

maxnoe:
@ParsonsRD I think the problem is having nice bright difffuse events where such a thing would show immediately because the reconstruction has to be good.

Can you try running the test e.g. with the file `get_dataset_path('lst_prod3_calibration_and_mcphotons.simtel.zst')`?

That contains 2 Events for 4 LSTs each at 500 GeV. The first event is a corner clipper in some telescopes, but the second seems like a good test case:

lst_events

HealthyPear (approved):
Yes, by going from CameraFrame to TelescopeFrame or NominalFrame the axes need to be switched.

Yes, there is no such test for this reconstructor at the moment (I think it has never been used for quite some time).

If you want you can add it in this PR: I did what you describe for `ctapipe.reco.HillasReconstructor` in PR #1408 (in particular `ctapipe/reco/tests/test_HillasReconstructor.py`).

kosack (changes requested):
sorry, I should not have approved it until we add a unit test obviously...

I actually didn't realize the hillas_intersection method was still working code -we should also perhaps cross-check it with the HillasReconstructor.

HealthyPear:
we already do it, in `ctapipe/reco/tests/test_reconstruction_methods.py`

HealthyPear (changes requested):
sorry, I should have also clicked the other button :/

ParsonsRD:
> we already do it, in `ctapipe/reco/tests/test_reconstruction_methods.py`
> 
> of course also there there should be the same time of test (or better, the same type of **assertion**)

Yes except this only tests that the code runs and gives a valid answer, not that it is actually reconstruction events properly. This bug did not make this test fail.

ctapipe/reco/hillas_intersection.py
```diff
@@ -168,7 +168,7 @@ def predict(self, hillas_dict, subarray, array_pointing, telescopes_pointings=No
         err_x *= u.rad
         err_y *= u.rad
 
-        nom = SkyCoord(fov_lon=src_x * u.rad, fov_lat=src_y * u.rad, frame=nom_frame)
+        nom = SkyCoord(fov_lat=src_x * u.rad, fov_lon=src_y * u.rad, frame=nom_frame)
```
maxnoe, line 171: This bug would proablly also not have happened, if the variables for called `src_fov_lat` / `src_fov_lon` instead of `x / y` / properly use the nominal / telescope frame.

HealthyPear:
Sorry @ParsonsRD I mistakenly modified your comment....

=================

> Yes except this only tests that the code runs and gives a valid answer, not that it is actually reconstruction events properly. > This bug did not make this test fail.

Indeed, this is why I suggested taking a look at #1408 where I do the test you are referring to

ParsonsRD:
> sorry, I should not have approved it until we add a unit test obviously...
> 
> I actually didn't realize the hillas_intersection method was still working code -we should also perhaps cross-check it with the HillasReconstructor.

I'm trying to resurrect the ImPACT code, but was just looking at this one to get reacquainted with ctapipe (it's easier to debug etc). With respect to the HillasReconstructor code, when I test it on a HESS MC dataset (where I know roughly how well such a Hillas analysis should perform) it doesn't seem to perform well at all in terms of angular resolution.

ctapipe/reco/hillas_intersection.py
maxnoe: Or the bug would just move to the other code point.

So all in all I would say this is another reason to just fully switch to hillas parameters in telescope frame with names field names mapping to the coordinate frame axis names.

ParsonsRD: Agreed, of course when I wrote this code the nominal frame dimensions were called x and y :-).

ParsonsRD:
> Indeed, this is why I suggested taking a look at #1408 where I do the test you are referring to

Ah apologies, misunderstood your comment. 

I'm not sure which test in the PR you mean though. I was thinking something like, perform the reconstruction and check for a small subset of events that the reconstructed position is not too far (say less that half a degree) from the reconstructed position.

HealthyPear:
> > Indeed, this is why I suggested taking a look at #1408 where I do the test you are referring to
> 
> Ah apologies, misunderstood your comment.
> 
> I'm not sure which test in the PR you mean though. I was thinking something like, perform the reconstruction and check for a small subset of events that the reconstructed position is not too far (say less that half a degree) from the reconstructed position.

In `ctapipe/reco/tests/test_HillasReconstructor.py` I do this kind of test for both parallel (`test_parallel_reconstruction`) and divergent paintings `(test_divergent_reconstruction)`.
At the end of each test, I compare reconstruction from  CameraFrame and TelescopeFrame and I check that each quantity is within a certain relative tolerance.

Technically what you propose is a _benchmark_, not a unit-test, that is why I didn't use simulated information in these tests, but I wanted to point you there because the code would be the same.

ParsonsRD:
Aha thanks, I can look at implementing something similar.

maxnoe:
> Technically what you propose is a benchmark, not a unit-test, that is why I didn't use simulated information in these tests, but I wanted to point you there because the code would be the same.

I disagree, the reconstruction of a single, well shaped bright event is a good unit test. The overall performance is a benchmark.

ctapipe/reco/hillas_intersection.py
ParsonsRD: > So all in all I would say this is another reason to just fully switch to hillas parameters in telescope frame with names field names mapping to the coordinate frame axis names.

I completely agree with this statement. The definition of Hillas parameters in the camera frame has always upset me (for many reasons). In HESS we typically define them in the nominal system, however we never divergently point, so the difference to the telescope frame is minimal.

kosack: fortunately, we have #1591  and #1408  coming soon...

kosack:
>  when I test it on a HESS MC dataset (where I know roughly how well such a Hillas analysis should perform) it doesn't seem to perform well at all in terms of angular resolution.

That's the opposite of what Thomas Gasparetto found, I think. Should look into if there was a regression

ParsonsRD:
> 
> That's the opposite of what Thomas Gasparetto found, I think. Should look into if there was a regression

I only did a very quick test, so might not be a real difference. Also, could be something not working well with the HESS simulations. 

If there were problems with this code it could have been due to this bug, as everything would look right on axis and break as you go to larger offsets. I'm not sure who added this line though, so I'm not sure how long it had been in there. Essentially this is a direct reimplementation of Konrad's code, I've tested individual events against these codes and it seems to perform basically identically, so it shouldn't really do much worse unless there is a bug.

ParsonsRD:
Yes, I must have made a mistake before. I find that the 2 Hillas reconstructors give very similar results, with the HillasReconstructor being slightly better.

maxnoe:
@ParsonsRD Can you add the test using the file I proposed? Or should I do it in a follow-up PR?

ParsonsRD:
Sure I can add it, I'll try to take a look later this week.

maxnoe:
@ParsonsRD I added the test as I was looking into this again.

Before the fix, the tests fails with a separation of 1.6 degrees, with the fix, separation between true and reco position is < 0.1 degree.

ParsonsRD:
Thanks @maxnoe

ctapipe/conftest.py
```diff
@@ -62,6 +62,42 @@ def test_my_thing(test_event):
     return deepcopy(_global_example_event)
 
 
+@pytest.fixture(scope="function")
+def gamma_off_axis_500_gev():
```
kosack, line 66: I suggest naming this `gamma_event_off_axis_500_gev`, so it's clear that this fixtures is an event, not some other information. 

It's a but long, but even `subarray_and_gamma_event_off_axis_500_gev` would be more clear, as it's not just an event, but also the subarray description...

ctapipe/conftest.py
```diff
@@ -62,6 +62,42 @@ def test_my_thing(test_event):
     return deepcopy(_global_example_event)
 
 
+@pytest.fixture(scope="function")
```
kosack, line 65: I think the scope should be 'session' here, since with function scope this is re-run for every test (which is a bit slow). This only needs to be loaded once (assuming no test modifies it). 

Alternately, if you want to be careful to avoid modification by tests, you can keep it "function", but  make it a generator and add at the end:
```py3
while True:
     yield deepcopy(event)
```
(I think that works at least) see https://docs.pytest.org/en/stable/fixture.html#dynamic-scope

kosack (changes requested):

ctapipe/reco/tests/test_hillas_intersection.py
```diff
@@ -297,3 +297,40 @@ def test_reconstruction():
         assert fit_result.is_valid
 
     assert reconstructed_events > 0
+
+
+def test_reconstruction_works(gamma_off_axis_500_gev):
+    from ctapipe.image import ImageProcessor
```
HealthyPear, line 303: I don't think this import is needed

HealthyPear (changes requested):

maxnoe:
Just to check what @HealthyPear reported to me, I produced this plot here with the separation for both HillasReconstructor and HillasIntersection for the test event vs. different cleaning levels (boundary = 0.5 * picture threshold for everything):

separation_vs_cleaning_levels

kosack:
The fact that there is any difference between HillasReconstructor and hillas_intersection is somewhat surprising (I suspect a simple bug somewhere), since the methods are nearly identical just implemented differently.  But that's another issue.

maxnoe:
@kosack, No, these methods are not equivalent. One tries to find a single least squares solution for the direction using all the lines the other calculates a weighted average of all pair-wise intersections. 

I am not surprised that the first method performance better.

ctapipe/conftest.py
maxnoe: With the `while True` I get

> fixture function has more than one 'yield':

maxnoe: I think generators are only used to be able to run clean up code after a fixture isn't needed anymore

ctapipe/reco/tests/test_hillas_intersection.py
maxnoe: yes, removed

maxnoe:
@kosack @HealthyPear Comments addressed

maxnoe (approved):

HealthyPear (approved):

ParsonsRD:
I would guess this is either a bug or a difference in the way the pairwise intersections are weighted when the average direction is taken. I'm actually kind of amazed that the reconstruction performs basically flat higher than 4,2 cuts. I would consider these as very loose tailcuts compared to the 5, 10 cuts used in HESS. It might just be that this is a particularly bright event though, so the noise around the image edges doesn't change anything.

> Edit: sorry, it seems I misrembered how the `HillasReconstructor` works. It seems to also to the pairwise intersection for the origin, the least squares is only done for the impact. I wonder why?

Off-topic, but It is also possible to implement fitting style reconstruction that is not quite to heavy as something like ImPACT, for example performing a likelihood fit to the Hillas parameters (I did this many years ago in read_hess for my PhD thesis), but that is a feature proposal for another time I think 😄

maxnoe:
@kosack I think your change requests are addresed

kosack:
> Off-topic, but It is also possible to implement fitting style reconstruction that is not quite to heavy as something like ImPACT, for example performing a likelihood fit to the Hillas parameters (I did this many years ago in read_hess for my PhD thesis), but that is a feature proposal for another time I think 😄

In fact, the original version of HillasReconstructor did exactly that, using scipy.leastsq.  But for some reason it was abandoned. Could be resurrected, I think. Didn't I just open an issue about that? Maybe it was in protopipe...

kosack (approved):

PR #1639, opened 2021-02-26
fix a bug in twopass extractor where the timing slope was wrong

kosack:
This lead to incorrect predicted values.
The bug was just that timing_parameters() was called with a zeroed-out
image, and no image mask was passed. Passing the mask fixes it.

Also some small cleanups and name changes.

The effect was that the second pass had very little effect, and in fact the incorrect predicted window location caused any signal outside the main island to be lost (sub showers, etc).  

### Before
Left is the true_image, right is the extracted one, note the bright sub-shower is lost :
image

### After fix:
Sub shower is recovered, and more charge as well
image

### DIFF
--- a/ctapipe/image/extractor.py
+++ b/ctapipe/image/extractor.py
@@ -997,17 +997,26 @@ class TwoPassWindowSum(ImageExtractor):
         # Indexes of pixels that will need the 2nd pass
         non_core_pixels_mask = charge_1stpass < core_th
 
+        # find all islands using this cleaning
+        num_islands, labels = number_of_islands(camera_geometry, mask_1)
+
+        if num_islands > 0:
+            # ...find the biggest one
+            mask_biggest = largest_island(labels)
+        else:
+            mask_biggest = mask_1
+
         # STEP 4
 
         # if the resulting image has less then 3 pixels
         # or there are more than 3 pixels but all contain a number of
         # photoelectrons above the core threshold
-        if np.count_nonzero(mask_1) < 3:
+        if np.count_nonzero(mask_biggest) < 3:
             # we return the 1st pass information
             # NOTE: In this case, the image was not bright enough!
             # We should label it as "bad and NOT use it"
             return charge_1stpass, pulse_time_1stpass
-        elif len(non_core_pixels_mask) == 0:
+        elif np.count_nonzero(non_core_pixels_mask) == 0:
             # Since all reconstructed charges are above the core threshold,
             # there is no need to perform the 2nd pass.
             # We return the 1st pass information.
@@ -1017,15 +1026,6 @@ class TwoPassWindowSum(ImageExtractor):
 
         # otherwise we proceed by parametrizing the image
         hillas = hillas_parameters(camera_geometry[mask_1], charge_1stpass[mask_1])
-
-        # find all islands using this cleaning
-        num_islands, labels = number_of_islands(camera_geometry, mask_1)
-
-        if num_islands > 0:
-            # ...find the biggest one
-            mask_biggest = largest_island(labels)
-        else:
-            mask_biggest = mask_1
 
         # STEP 5
 
--- a/ctapipe/image/extractor.py
+++ b/ctapipe/image/extractor.py
@@ -361,7 +361,7 @@ class ImageExtractor(TelescopeComponent):
         """
         super().__init__(subarray=subarray, config=config, parent=parent, **kwargs)
 
-        self.sampling_rate = {
+        self.sampling_rate_ghz = {
             telid: telescope.camera.readout.sampling_rate.to_value("GHz")
             for telid, telescope in subarray.tel.items()
         }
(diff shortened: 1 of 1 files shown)


ctapipe/image/extractor.py
```diff
@@ -1032,31 +1035,36 @@ def _apply_second_pass(
         # linear fit of pulse time vs. distance along major image axis
         # using only the main island surviving the preliminary
         # image cleaning
-        timing = timing_parameters(camera_geometry, image_2, pulse_time_1stpass, hillas)
+        timing = timing_parameters(
+            geom=camera_geometry,
+            image=image_2,
+            peak_time=pulse_time_1stpass,
+            hillas_parameters=hillas,
+            cleaning_mask=mask_biggest,
```
kosack, line 1043: this was the main bug, missing the cleaning_mask option

HealthyPear: thanks for having spotted this!

HealthyPear (approved):

kosack:
Another issue here is that the peak_time image returned is the one used for the integration (meaning it is a smooth gradient across the camera, as it is the predicted peak time).  That means it is essentially useless other than to see what the extractor used.  Would it be better to return the first-pass peak time image instead, which is the time of maximum for all pixels after a sliding window filter?

ctapipe/image/extractor.py
maxnoe: Why zero out the image at all? All algorithms should support passing either masked image / geometries or image and mask

kosack: agreed - I didn't make that change here yet since i wanted first to show the minimal bug fix, but it would be faster/easier to just slice the geometry as is done in ImageProcessor.

HealthyPear (approved):

ctapipe/image/extractor.py
```diff
@@ -987,36 +991,23 @@ def _apply_second_pass(
             keep_isolated_pixels=False,
             min_number_picture_neighbors=1,
         )
-        image_1 = charge_1stpass.copy()
-        image_1[~mask_1] = 0
 
         # STEP 3
 
-        # find all islands using this cleaning
-        num_islands, labels = number_of_islands(camera_geometry, mask_1)
```
maxnoe, line 996: Why was this removed? Changing how this code handles islands is not just fixing the bug you mentioned in the title.

HealthyPear: Yes I just noticed this - without this the algorithm behaves differently than the original MARS one (which is the point - if we want we can create variations of course, but the 2nd pass time gradient is supposed to use the main cluster)

HealthyPear (changes requested):

ctapipe/image/extractor.py
kosack: It's not removed - It's just moved a few lines below where it is actually used.

maxnoe: ah, ok, sorry.

HealthyPear: No, it's still different, because now the check on the number of pixels and the parameterization are performed on the multi-cluster image.

Whar instead this algorithm is supposed to do is to clean for the biggest cluster, check if n_pixels > 3, if yes parametrized the image, and then perform the timing fit.

I report here the explanation of Abelardo,

> [...] After this cleaning, only the biggest cluster of pixels ("main island") is kept. If the resulting image has three or more pixels, it is parametrized following the Hillas approach. A linear fit of pulse time vs. distance along major image axis is then performed [...]

kosack: Ah, I see now - it was very confusing to follow with all the copies of the images, and each one being overwritten.  I'll modify it to check the number of pixels in the "biggest island" mask.

ctapipe/image/extractor.py
```diff
@@ -1025,38 +1025,43 @@ def _apply_second_pass(
             return charge_1stpass, pulse_time_1stpass
 
         # otherwise we proceed by parametrizing the image
-        hillas = hillas_parameters(camera_geometry, image_2)
+        hillas = hillas_parameters(camera_geometry[mask_1], charge_1stpass[mask_1])
```
HealthyPear, line 1028: Also here mask_biggest is used

HealthyPear (changes requested):

ctapipe/image/extractor.py
kosack: Ah, in that case we can compute the masked version only once - I'll change it (right now it's sliced two time)

HealthyPear (approved):

maxnoe (approved):

maxnoe:
Should we make 0.10.5 for this? @kosack @HealthyPear ?

HealthyPear:
Yes.

I have the feeling that this bugfix will improve a bit gamma-hadron separation at low energies.
I started a test branch in protopipe to refactor the training part into ctapipe-stage1 + shower geometry and I would like to test it with this bugfix.

PR #1646, opened 2021-03-02
Remove charge resolution tools, fixes #1634

### LINKED ISSUE #1634: remove charge resolution tools
We have a few tools like 
* ctapipe-chargeres-extract
* ctapipe-chargeres-hist
* ctapipe-chargeres-plot

Which probably are not needed anymore, as with the DL1 format and benchmarks, you can plot charge resolution pretty easily.   Have to check that ChargeResolutionCalculator does the same as what are in the protopipe benchmarks, but I think it's covered. @HealthyPear and @watsonjj should give the final say though.

### LINKED ISSUE #1643: Missing f-string in dl1_merge prevents from seeing broken filename

maxnoe:


setup.py
```diff
@@ -25,7 +23,7 @@
     "ctapipe-stage1 = ctapipe.tools.stage1:main",
     "ctapipe-merge = ctapipe.tools.dl1_merge:main",
 ]
-tests_require = ["pytest"]
+tests_require = ["pytest", "pandas>=0.24.0"]
```
kosack, line 26: isn't pandas still used in some of the notebooks generated by the docs?

maxnoe: ha, yes, probably, will also add to `docs`

kosack: that's another issue of course, we could probably clean them up to only use astropy

maxnoe: seems not to be the case, the docs build passes...

maxnoe: or the `read_table` function, yes.

maxnoe: the docs build installed `[all]` not just `[docs]`.

kosack (approved):

HealthyPear (approved):

PR #1651, opened 2021-03-08
Generalize URL for downloading test data

HealthyPear:
This PR make the `DEFAULT_URL`, which is currently

`DEFAULT_URL = "http://cccta-dataserver.in2p3.fr/data/ctapipe-extra/v0.3.3/"`

really the default URL (and not the only choice like it is now) by adding `url=DEFAULT_URL` as a keyword default argument were necessary.

This PR should be complementary to #1638 .

### DIFF
--- a/ctapipe/utils/datasets.py
+++ b/ctapipe/utils/datasets.py
@@ -70,9 +70,7 @@ def find_all_matching_datasets(pattern, searchpath=None, regexp_group=None, url=
 
     if searchpath is None:
         searchpath = os.getenv("CTAPIPE_SVC_PATH")
-    search_path_dirs = get_searchpath_dirs(searchpath, url)
-
-    print(search_path_dirs)
+    search_path_dirs = get_searchpath_dirs(searchpath, url=url)
 
     # first check search path
     for path in search_path_dirs:
@@ -215,7 +213,7 @@ def try_filetypes(basename, role, file_types, url=DEFAULT_URL, **kwargs):
         for ext, reader in file_types.items():
             filename = basename + ext
             try:
-                path = get_dataset_path(filename, url)
+                path = get_dataset_path(filename, url=url)
                 break
             except (FileNotFoundError, HTTPError):
                 pass
--- a/ctapipe/utils/datasets.py
+++ b/ctapipe/utils/datasets.py
@@ -31,19 +31,21 @@ __all__ = ["get_dataset_path", "find_in_path", "find_all_matching_datasets"]
 DEFAULT_URL = "http://cccta-dataserver.in2p3.fr/data/ctapipe-extra/v0.3.3/"
 
 
-def get_searchpath_dirs(searchpath=os.getenv("CTAPIPE_SVC_PATH")):
+def get_searchpath_dirs(searchpath=os.getenv("CTAPIPE_SVC_PATH"), url=DEFAULT_URL):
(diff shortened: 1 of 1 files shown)


ctapipe/utils/datasets.py
```diff
@@ -31,19 +31,21 @@
 DEFAULT_URL = "http://cccta-dataserver.in2p3.fr/data/ctapipe-extra/v0.3.3/"
 
 
-def get_searchpath_dirs(searchpath=os.getenv("CTAPIPE_SVC_PATH")):
+def get_searchpath_dirs(searchpath=os.getenv("CTAPIPE_SVC_PATH"), url=DEFAULT_URL):
     """ returns a list of dirs in specified searchpath"""
+
+    print(searchpath)
```
maxnoe, line 37: debug print

ctapipe/utils/datasets.py
```diff
@@ -31,19 +31,21 @@
...
         searchpaths = [Path(p) for p in os.path.expandvars(searchpath).split(":")]
 
-    searchpaths.append(get_cache_path(DEFAULT_URL))
+    searchpaths.append(get_cache_path(url))
 
     return searchpaths
 
 
-def find_all_matching_datasets(pattern, searchpath=None, regexp_group=None):
+def find_all_matching_datasets(pattern, url=DEFAULT_URL, searchpath=None, regexp_group=None):
```
maxnoe, line 48: Better add new arguments at the end, so you don't break API (since now the order of args has changed)

ctapipe/utils/datasets.py
```diff
@@ -112,15 +116,15 @@ def find_in_path(filename, searchpath):
 
     """
 
-    for directory in get_searchpath_dirs(searchpath):
+    for directory in get_searchpath_dirs(searchpath, url):
```
maxnoe, line 119: `url=url`

HealthyPear: no problem to add this, why is it needed? that function requires now only 2 args

ctapipe/utils/datasets.py
HealthyPear: removed thanks

ctapipe/utils/datasets.py
HealthyPear: done

ctapipe/utils/datasets.py
maxnoe: Not needed. Better style. E.g. safe against adding a third as a second argument, like you did before changing to add it at the end ;)

Always prefer kwargs when possible.

maxnoe (approved):

kosack:
Can you explain the use case for this that can't be solved using $CTAPIPE_SVC_PATH?

HealthyPear:
> Can you explain the use case for this? It's fine to add, but I'm not sure we want people changing this ever.

It's not a change, it's an additional option.

My main drive for this PR was that to set up and execute the integration tests of _protopipe_ (see cta-observatory/protopipe#70) I need to access `http://cccta-dataserver.in2p3.fr/data/protopipe/[ecc...]`
and I didn't want to re-invent the wheel (even though this means that I won't be able to test this until the next release anyway without some ugly copy&paste)

kosack (approved):
ok that's fine

PR #1652, opened 2021-03-09
Fix codecov missing coverage for cli tools

maxnoe:
In lstchain we found that using `pytest --cov=<path>` lead to codecov not showing coverage for command line tools, since it was  not recognizing paths in the xml that were not starting at the base of the repository.


kosack:
I think all tests right now call the Tool classes directly rather than run an external tool, so no change is expected, but i suppose it doesn't hurt to call it this way in any case.

HealthyPear (approved):

PR #1656, opened 2021-03-11
Fix time_delta cleaning 

### LINKED ISSUE #1653: apply_time_delta cleaning should only check neighbors in the mask and is very slow
`apply_time_delta` cleaning is missing the check if the neighbor pixels are part of the cleaning set. It checks all neighbors, not only the once that survived the cleaning until that point.

This is a bug, since this function should be modeled after the fact-tools java version of the cleaning which includes this check:
https://github.com/fact-project/fact-tools/blob/da7d193e0e37aa33fbf23e48fe34e63ac743704a/src/main/java/fact/cleaning/TwoLevelTimeNeighbor.java#L164

Also, `apply_time_delta_cleaning` uses a pure python loop over pixels and is called multiple times.
This is a performance bottleneck.

https://github.com/cta-observatory/ctapipe/blob/91f2b13e0967ac00a01b009e667fc5e402b1d632/ctapipe/image/cleaning.py#L205

LukasNickel:
- fixes #1653

### DIFF
--- a/ctapipe/image/cleaning.py
+++ b/ctapipe/image/cleaning.py
@@ -232,14 +232,13 @@ def apply_time_delta_cleaning(
     `image[~mask] = 0`
 
     """
-    pixels_to_remove = []
-    mask = mask.copy()  # Create copy so orginal is unchanged
-    for pixel in np.where(mask)[0]:
-        neighbors = geom.neighbor_matrix_sparse[pixel].indices
-        time_diff = np.abs(arrival_times[neighbors] - arrival_times[pixel])
-        if sum(time_diff < time_limit) < min_number_neighbors:
-            pixels_to_remove.append(pixel)
-    mask[pixels_to_remove] = False
+    time_diffs = np.abs(arrival_times[mask, None] - arrival_times)
+    # neighboring pixels arriving in the time limit #Fix: and previously selected
+    valid_neighbors = (time_diffs < time_limit) & geom.neighbor_matrix[
+        mask
+    ]  # & mask # for the fix
+    enough_neighbors = valid_neighbors.sum(axis=1) >= min_number_neighbors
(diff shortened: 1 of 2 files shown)


ctapipe/image/cleaning.py
```diff
@@ -232,15 +232,13 @@ def apply_time_delta_cleaning(
...
-        time_diff = np.abs(arrival_times[neighbors] - arrival_times[pixel])
-        if sum(time_diff < time_limit) < min_number_neighbors:
-            pixels_to_remove.append(pixel)
-    mask[pixels_to_remove] = False
-    return mask
+    pixels_to_keep = mask.copy()
+    time_diffs = np.abs(arrival_times[mask, None] - arrival_times)
+    # neighboring pixels arriving in the time limit and previously selected
+    valid_neighbors = (time_diffs < time_limit) & geom.neighbor_matrix[mask] & mask
+    enough_neighbors = valid_neighbors.sum(axis=1) >= min_number_neighbors
```
maxnoe, line 239: `np.count_nonzero` instead of `.sum`

ctapipe/image/cleaning.py
```diff
@@ -232,15 +232,13 @@ def apply_time_delta_cleaning(
...
+    pixels_to_keep[np.where(pixels_to_keep)[0]] &= enough_neighbors
```
maxnoe, line 240: `pixels_to_keep[pixels_to_keep] &= enough_neighbors`

maxnoe:
Can you add a test that makes sure we don't reintroduce the bug.

LukasNickel:
The last test case should fail with the old cleaning behaviour

maxnoe (approved):

kosack (approved):

PR #1658, opened 2021-03-17
Transform warning into error for wrong config

maxnoe:


kosack (dismissed):

maxnoe:
@kosack, I now also added a solution that works with the `Tool`, where only a log warning instead of a proper warning was created.

maxnoe:
Unfortuately, we still get no error if a whole object has the wrong name.

So when we have a `MyTool.MyComponent` and have `config.MyTool.MyWrongComponent`, this is still silently ignored.

kosack (approved):

HealthyPear (approved):

PR #1659, opened 2021-03-19
Use tqdm.auto

maxnoe:


kosack (approved):

kosack:
we might want to specify a minimum TQDM version in setup.py in this case, since I think tqdm.auto did not exist in some older versions (it was tqdm.autonotebook at some point).

HealthyPear (approved):

PR #1666, opened 2021-03-23
Update dl1writer.py docstring

vuillaut:
Please let me know if I missed something but I think the docstring is lying in the dl1_writer, the option is only to write the parameters.

### DIFF
--- a/ctapipe/io/dl1writer.py
+++ b/ctapipe/io/dl1writer.py
@@ -114,7 +114,7 @@ class DL1Writer(Component):
     ).tag(config=True)
 
     write_parameters = Bool(
-        help="Compute and store image parameters", default_value=True
+        help="Store image parameters", default_value=True
     ).tag(config=True)
 
     compression_level = Int(


ctapipe/io/dl1writer.py
```diff
@@ -114,7 +114,7 @@ class DL1Writer(Component):
     ).tag(config=True)
 
     write_parameters = Bool(
-        help="Compute and store image parameters", default_value=True
+        help="Store image parameters", default_value=True
```
maxnoe, line 117: It's kind of true, since the stage1 tool only computes dl1 parameters if this is true, but this is a bad leftover and should be fixed with a proper traitlet for the stage1 tool.

kosack: yes this is just leftover from the refactoring that I did (before this was all part of one script).  I'm already re-working the options for computing and writing completely for the next major release (when DL2 is also included), so just fixing the docstring here for now is fine.

kosack (approved):

ctapipe/io/dl1writer.py
vuillaut: > It's kind of true, since the stage1 tool only computes dl1 parameters if this is true, but this is a bad leftover and should be fixed with a proper traitlet for the stage1 tool.

So just for my understanding, if `False` when using the stage1, the parameters are not computed?
Could you explain me how is this passed through the stage1 to the `ImageProcessor`? I fail to see it.

maxnoe: https://github.com/cta-observatory/ctapipe/blob/83d1a5d5cfc5bb4f79b126152aad3cd0f5cb5b6a/ctapipe/tools/stage1.py#L135

vuillaut: Thanks, I completely missed this.

kosack: it was jut to keep the number of options simple and easy to understand. Genearlly if you don't ask to write out something, there is no reason to compute it, therefore only a single option was needed.  In the future, as we add more parts of the analysis as I said, this will be have to be more flexible (at the expense of simplicity).

maxnoe (approved):

PR #1667, opened 2021-03-23
Remove DL1Writer.setup, just properly initialize in __init__

maxnoe:
Fixes an error with the stage 1 tool @orelgueta observed with a file not containing any physics events

### DIFF
--- a/ctapipe/io/dl1writer.py
+++ b/ctapipe/io/dl1writer.py
@@ -178,6 +178,7 @@ class DL1Writer(Component):
         # setup(), which is called when the first event is read.
 
         self.event_source = event_source
+        self._at_least_one_event = False
         self._is_simulation = event_source.is_simulation
         self._subarray: SubarrayDescription = event_source.subarray
 
@@ -204,11 +205,11 @@ class DL1Writer(Component):
 
     def __call__(self, event: ArrayEventContainer):
         """
-        Write a single event to the output file. On the first event, the output
-        file is set up
+        Write a single event to the output file.
         """
+        self._at_least_one_event = True
 
-        # Write subarray evvent data
+        # Write subarray event data
         self._write_subarray_pointing(event, writer=self._writer)
 
         self.log.debug(f"WRITING EVENT {event.index}")
@@ -228,6 +229,8 @@ class DL1Writer(Component):
     def finish(self):
         """ called after all events are done """
         self.log.info("Finishing DL1 output")
+        if not self._at_least_one_event:
+            self.log.warning("No events have been written to the output file")
         if self._writer:
             if self.write_index_tables:
                 self._generate_indices()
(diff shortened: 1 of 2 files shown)


kosack:
I think there was a very good reason I did this in `setup()` and not in `__init__()`...  I have to remember what it was specifically, but the concept in general was to have delayed initialization so that the construction of a Component has little to no overhead and doesn't create any files.  

I would prefer to keep initiailzation when the first event arrives, but to handle the case where no  setup() happens since no events were found separately

maxnoe:
@kosack I disagree. Opening a writer is expected to create a file, so no suprise there (at least for me), and it writes the simulation configuration and subarray information, which should not depend on a triggered event.

So for me, this is the expected case.

kosack:
@maxnoe I'm not disagreeing, just being cautious since I usually think through many non-obvious use cases when designing code like this.  But since I cannot remember the reason for doing lazy init in the particular case of DL1DataWriter, it seems to be ok.  I meant at some point to make a base class for lazy init Components, and this component in principle would not be one that needs to inherit from it. 

However,  please add a warning in the writer if we get to`finish()` without ever having called __call__() at least once, since with this current fix it would stop the crash that was seen, but would still result in a file empty of events (which was clearly not the expected result).   I guess having a successful exit status is still ok, since it did do what it was asked for, so just a logged warning that the user may have made a mistake in the configuration is enough.  I think just adding an internal state flag or counter would be enought (e.g. in __call__, add `self._call_count += 1`, and warn if `self._call_count==0` in finish())

maxnoe:
>  (which was clearly not the expected result).

But it is an expected result e.g. when producing smaller files using simtel, and yes I agree a warning is probably in order, but it is also not unusual to expect a proper dl1 output file with configuration / shower histograms even when no event triggered.

Official productions should have large enough number of showers that we don't need this, but at least in FACT we had with older versions of CORSIKA compute runs so small that out of a couple of thousand runs, there were a handfull not triggered and that actually resulted in invalid FITS output files from the detector simulation because of exactly this kind of situation: some initilization only happening on the first triggered event.

kosack (changes requested):
It's fine, just add a warning in the case of no events.  Even if that sometimes is expected, it's a good idea.  It could just say something on the lines of :

"WARNING: an output file was written, but no events were processed.  This may be expected for small input simulations where no showers caused a trigger, or may be due to a mis-configuration."

maxnoe:
I added the warning.

kosack (approved):

HealthyPear (commented):
The code looks fine, but this special case of empty I/O now is not tested (even though as @maxnoe said it will be rare with official productions, it can still be possible in testing - like Orel found out)

maxnoe:
@kosack @HealthyPear I added the test for no events being written.

kosack (approved):

HealthyPear (approved):

PR #1672, opened 2021-03-31
Add another cleaning method in cleaning.py

satoshifukami0115:
I tried to implement a new image cleaning method used in MAGIC.
Distributions of a few Hillas parameters are checked in the link below.
https://indico.cta-observatory.org/event/3368/contributions/28400/attachments/19171/26360/CTA_ASWG_20210331_Fukami.pdf

I'm not sure if this method is better than other methods so far, but I think we can have an additional method as one option.
If you have any suggestions or comments, please let me know.


ctapipe/image/cleaning.py
```diff
@@ -241,6 +243,56 @@ def apply_time_delta_cleaning(
...
+
+    A boolean mask of *clean* pixels.  To get a zero-suppressed image and pixel
+    list, use `image[mask], geom.pix_id[mask]`, or to keep the same
+    image size and just set unclean pixels to 0 or similar, use
+    `image[~mask] = 0`
+
+    """
+    pixels_to_remove = []
+    mask = mask.copy()
+    if sum(mask) > 0:
```
maxnoe, line 279: use `np.count_nonzero(mask)` instead, it's faster.

ctapipe/image/cleaning.py
```diff
@@ -241,6 +243,56 @@ def apply_time_delta_cleaning(
...
+
+        # use main island (maximum charge) for time average calculation
+        num_islands, island_labels = number_of_islands(geom, mask)
+        mask_main = island_labels == np.argmax([np.sum(image[np.where(island_labels == l)]) for l in range(1,num_islands+1)]) + 1
+        time_ave = np.average(arrival_times[np.where(mask_main)[0]], weights=image[np.where(mask_main)[0]]**2)
+
+        for pixel in np.where(mask)[0]:
```
maxnoe, line 286: This looks like it can be vectorized. Please check if you can make it work without the for loop just using numpy.

If the problem is the `time_limit`, you can do `limit = np.where(image < (2 * picture_threshold), time_limit, time_limit * 2)`

ctapipe/image/cleaning.py
```diff
@@ -241,6 +243,56 @@ def apply_time_delta_cleaning(
...
+            time_diff = np.abs(arrival_times[pixel] - time_ave)
+            time_limit = time_limit if image[pixel] < picture_thresh * 2 else time_limit * 2
```
maxnoe, line 288: This overwrites time_limit. So for each core pixel, you will increase the time limit by a factor of two of all following pixels. This seems not intended.

Also, what is the intension here? Why are brighter pixels allowed a more lax time requirement? Why the split at 2 * picture_threshold?

ctapipe/image/cleaning.py
```diff
@@ -241,6 +243,56 @@ def apply_time_delta_cleaning(
     return pixels_to_keep
 
 
+def apply_time_average_cleaning(
+    geom, mask, image, arrival_times, picture_thresh, time_limit
+):
+    """
+    Extract all pixels that arrived within a given timeframe 
+    with respect to the time average of the pixels on the main island.
```
maxnoe, line 251: This function does more than the docstring says. It somehow adapts the time limit dependent on the image and on picture threshold.

ctapipe/image/cleaning.py
```diff
@@ -241,6 +243,56 @@ def apply_time_delta_cleaning(
...
+    `image[~mask] = 0`
+
+    """
+    pixels_to_remove = []
+    mask = mask.copy()
+    if sum(mask) > 0:
+
+        # use main island (maximum charge) for time average calculation
+        num_islands, island_labels = number_of_islands(geom, mask)
+        mask_main = island_labels == np.argmax([np.sum(image[np.where(island_labels == l)]) for l in range(1,num_islands+1)]) + 1
```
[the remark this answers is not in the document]
satoshifukami0115: OK, thank you. I would use this after merged.

ctapipe/image/cleaning.py
satoshifukami0115: Done, thanks.

ctapipe/image/cleaning.py
satoshifukami0115: Thank you. I removed for loop using np.where() you suggested.

ctapipe/image/cleaning.py
```diff
@@ -241,6 +243,54 @@ def apply_time_delta_cleaning(
...
+
+    """
+    pixels_to_remove = []
+    mask = mask.copy()
+    if np.count_nonzero(mask) > 0:
+
+        # use main island (maximum charge) for time average calculation
+        num_islands, island_labels = number_of_islands(geom, mask)
+        mask_main = island_labels == np.argmax([np.sum(image[np.where(island_labels == l)]) for l in range(1,num_islands+1)]) + 1
+        time_ave = np.average(arrival_times[np.where(mask_main)[0]], weights=image[np.where(mask_main)[0]]**2)
```
maxnoe, line 284: This can just be `np.average(arrival_times[mask_main], weights=image[mask_main]**2)`, no need to call `np.where`.

ctapipe/image/cleaning.py
satoshifukami0115: This is a way to avoid removing signal pixels located at edges of elongated shower images.
Images with large impact parameters have large time gradient, resulting in large time difference even inside core pixels. If we use time difference from the average, we have to take care of those events.
Of course, the picture_threshold *2, or time_limit *2 can be changed. This is just a fixed number used in MAGIC.

satoshifukami0115: If you have any suggestions, please let me know. This part can be considered further.

ctapipe/image/cleaning.py
satoshifukami0115: Thanks, I will change the docstring accordingly.

ctapipe/image/cleaning.py
satoshifukami0115: Thanks, I modified it.

maxnoe:
@satoshifukami0115 The brightest_island function is now merged and you can update here to use it.

satoshifukami0115:
> @satoshifukami0115 The brightest_island function is now merged and you can update here to use it.
@maxnoe Thank you! I changed the part to use this function.

maxnoe:
This looks good, only a test is missing @satoshifukami0115

satoshifukami0115:
Thank you @maxnoe. The test means to include a test function in this file, right?
https://github.com/cta-observatory/ctapipe/blob/master/ctapipe/image/tests/test_cleaning.py

I will also work on the Cleaner class within this week and let you know once done.

kosack (changes requested):
looks good.  Could you please also run `black` to re-format the code?  Actually the correct method should be to install the git commit hook by first running this as shown in the *Getting Started for Developers* guide for ctapipe:
```
pre-commit install
```
After that, it should be automatically formatted when you commit.  (you may have some commit failures if the code isn't already formatted, so just add/commit twice in that case)

ctapipe/image/cleaning.py
```diff
@@ -440,3 +575,34 @@ def __call__(
...
+    ).tag(config=True)
+
+    def __call__(
+        self, tel_id: int, image: np.ndarray, arrival_times=None
+    ) -> np.ndarray:
+        """
+        Apply MAGIC-like image cleaning with timing information. See `ImageCleaner.__call__()`
+        """
+
+        return mars_cleaning_1st_pass(
```
kosack, line 599: Do you really want to run this function here, and not your `time_constrained_clean()` function?

ctapipe/image/cleaning.py
```diff
@@ -238,6 +240,57 @@ def apply_time_delta_cleaning(
...
+    Returns
+    -------
+
+    A boolean mask of *clean* pixels.  To get a zero-suppressed image and pixel
+    list, use `image[mask], geom.pix_id[mask]`, or to keep the same
+    image size and just set unclean pixels to 0 or similar, use
+    `image[~mask] = 0`
+
+    """
+    pixels_to_remove = []
```
maxnoe, line 277: this is an unused variable?

satoshifukami0115:
> looks good. Could you please also run `black` to re-format the code? Actually the correct method should be to install the git commit hook by first running this as shown in the _Getting Started for Developers_ guide for ctapipe:
> 
> ```
> pre-commit install
> ```
> 
> After that, it should be automatically formatted when you commit. (you may have some commit failures if the code isn't already formatted, so just add/commit twice in that case)

@kosack Thank you for your review. I did not run pre-commit. I will do it and push the branch again.

ctapipe/image/cleaning.py
satoshifukami0115: No, this I should have changed to `time_constrained_clean()`. I will do it now...

ctapipe/image/cleaning.py
satoshifukami0115: Yes, it is. I will remove this line...

ctapipe/image/cleaning.py
```diff
@@ -323,6 +376,88 @@ def fact_image_cleaning(
...
+    # keep core pixels whose arrival times are within a certain time limit of the average                                                                                                                 
+    mask_core = apply_time_average_cleaning(geom, pixels_in_picture, image, arrival_times, picture_thresh, time_limit_core)
+
+    # find boundary pixels that pass a boundary threshold                                                                                                                                                 
+    pixels_above_boundary = image >= boundary_thresh
+    pixels_with_picture_neighbors = geom.neighbor_matrix_sparse.dot(mask_core)
+    mask_boundary = ( pixels_above_boundary & pixels_with_picture_neighbors ) & np.invert(mask_core)
+
+    # keep boundary pixels whose arrival times are within a certain time limit of the neighboring core pixels                                                                                             
+    pixels_to_remove = []
```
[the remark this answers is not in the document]
satoshifukami0115: Yes, same here.

maxnoe:
The pyflakes check is also complaining about some unused variables:

```
Run pyflakes ctapipe
ctapipe/image/cleaning.py:277:5 local variable 'pixels_to_remove' is assigned to but never used
ctapipe/image/cleaning.py:449:5 local variable 'pixels_to_remove' is assigned to but never used
ctapipe/image/tests/test_cleaning.py:1:1 'tkinter.Image' imported but unused
Error: Process completed with exit code 1.

```

satoshifukami0115:
> The pyflakes check is also complaining about some unused variables:
> 
> ```
> Run pyflakes ctapipe
> ctapipe/image/cleaning.py:277:5 local variable 'pixels_to_remove' is assigned to but never used
> ctapipe/image/cleaning.py:449:5 local variable 'pixels_to_remove' is assigned to but never used
> ctapipe/image/tests/test_cleaning.py:1:1 'tkinter.Image' imported but unused
> Error: Process completed with exit code 1.
> ```

Yes, it seems that the `tkinter` module is not used in `ctapipe/image/tests/test_cleaning.py`.
I will remove this line.

satoshifukami0115:
I have performed the modifications you suggested and also confirmed that `make doc` worked. Please let me know if you have additional requests or comments. Thank you again for your comments so far.

kosack (approved):

maxnoe (approved):

maxnoe:
@satoshifukami0115 Sorry for the late reply, this looks good know. Feel free to ping me the next time something is ready for review and I am not reacting for so long.

PR #1673, opened 2021-04-06
DL2 data model and output

kosack:
This PR is the first of several changes needed to support the "stage 2" output (DL1→DL2) and also training output (files with a subset of DL1 + DL2 information in them).  In general the user should see no main change here, ctapipe-stage1 still works as before; it will in the future be superceded (but not replaced) by a version that can optionally also generate DL2 info once this PR is merged.

## The main changes are

* *containers*: cleanups to `event.dl2` to support telescope-wise (`event.dl2.tel` )and shower-wise (`event.dl2.stereo`) dl2 parameters 
* *dl1writer*: renamed `DL1Writer` to `DataWriter`, and now supports two new options `write_stereo_shower` and `write_mono_shower`, which generate the `/dl2/event` part of the data model
* *dl1eventsource*: support reading files that happen to also have DL2 info in them (no change otherwise, still only reads DL1 info, since not sure it's needed to read DL2 eventwise)
* *io.tableio*: added a `TelListToMaskTransform` (which somehow was missed in the previous refactoring of transformations). This is needed to turn a variable-length tel_id list (e.g. which telescopes were used in reconstruction) into a fixed-length bitmask, just like the trigger pattern.
* various minor changes elsewhere to support the renaming of containers and DL1Writer. 

DL2 output supports multiple reconstructions done at once.  In the current DL2 data model it generates looks like this:
image

the Reference metadata are also extended to support multiple data levels , currently as a stringified list in JSON format, but it could be  perhaps better as just a comma-separated list:

image

Data are currently split just like for DL1, but with two  additional group nodes: quantity being reconstructed  (energy, geometry, particle class), and the algorithm name. 

### misc

* should we add DL2_MONO and DL2_STEREO as datalevels? It only makes sense for full reconstruction workflows, not training workflows (where you might want other subsets of DL2, like only the geometry and not the energy) 
    * → **this should be determined later (See note below about data sub-levels)**

## Future Todos (for another PR):
- Update ctapipe-merge to also merge DL2 info → deferred to another PR
- Implement DL0 writing (quite simple to do now), to serve as a testbed for DL0 format


ctapipe/containers.py
```diff
@@ -487,17 +487,19 @@ class TriggerContainer(Container):
...
     tels_with_trigger = Field(
-        [], "List of telescope ids that triggered the array event"
+        None, "List of telescope ids that triggered the array event"
     )
     event_type = Field(EventType.SUBARRAY, "Event type")
     tel = Field(Map(TelescopeTriggerContainer), "telescope-wise trigger information")
 
 
-class ReconstructedShowerContainer(Container):
+class ReconstructedGeometryContainer(Container):
```
HealthyPear, line 496: you should update the one in "__all__" otherwise it won't be read

ctapipe/io/tests/test_datawriter.py
```diff
@@ -1,18 +1,35 @@
...
 from pathlib import Path
 from ctapipe.instrument import SubarrayDescription
 from copy import deepcopy
 import tables
 import logging
+from astropy import units as u
 
 
-def test_dl1writer(tmpdir: Path):
+def generate_dummy_dl2(event):
```
HealthyPear, line 16: should be `test_xxx` otherwise pytest doesn't find it (at least this is what I tested by checking out the PR)

HealthyPear (changes requested):
As we discussed together, at least for stereo I vouch for 

/dl2/event/stereo

   - shower
   - energy
   - classification

and then under each of those the output of the algorithm(s).

For mono, I can't say much since I am not working on that, but I would say to minimize the divergencies and just use the same approach we have now but with only 1 tel_id

I left a couple of comments to start!

ctapipe/containers.py
HealthyPear: this also the reason why the docs fail

ctapipe/io/tests/test_datawriter.py
kosack: that is not a test, but a function to generate some dummy data for one of the tests (it just fills in the DL2 containers, since we don't yet have a reconstructor that does it)

ctapipe/containers.py
kosack: done

ctapipe/io/tests/test_datawriter.py
HealthyPear: ah yes for the roundtrip test ok

ctapipe/containers.py
maxnoe: Is `Geometry` a good word here? More more commong would be `Direction`, right?

kosack: well in principle it's not just the direction, but also the height of shower max, and potentially  (depending on the reco method) other aspects of the shower's geometry like the longitudinal and lateral widths,  Moliere radius, etc. a

kosack:
Note that it would be quite easy to also add the ability to write DL0 data to this, and then we would have a nice testbed for what HDF5 DL0 data could look like.

kosack:
the more I think about it, I think having the option to split the per-tel mono reconstructions into datasets by tel_id or tel_type would still be useful, even if it adds some overhead from having many more tables. First because it then mirrors how DL1 is organized, and second because it makes it easier to  generate mono reconstruction *before* merging files by telescope (e.g. one has the option to do DL1+DL2_mono, merge, and then compute DL2_stereo).   Of course that still is possible with flat mono tables, but it would require a `sort()` operation after appending to do the merge by telescope.

ctapipe/io/tests/test_metadata.py
```diff
@@ -21,8 +21,8 @@ def test_construct_and_write_metadata(tmp_path):
         product=meta.Product(
             description="An Amazing Product",
             creation_time="2020-10-11 15:23:31",
-            data_category="S",
-            data_level="DL1",
+            data_category="Sim",
+            data_level=["DL1"],
```
HealthyPear, line 25: Shouldn't we test also for DL2 metadata here?

HealthyPear (approved):
I think everything strictly needed is there, nonetheless, I'll go again through the points raised, in case I didn't get something,

- JSON list vs simple comma-separated list

What would the consequences of choosing one or the other? could you give more details?

- how to split stereo data

I'll stick to my original proposal (which is done here)

`dl2/event/subarray/*estimated quantity*/*algorithm*`
`dl2/event/telescope/*estimated quantity*/*algorithm*`

- about mono

I don't get what would happen in case one chooses `tel_type` to split
(if I understood correctly for tel_id there is instead no problem we can do as we do in DL1)

if there is only 1 telescope, shouldn't the table be unique anyway?

- should we add DL2_MONO and DL2_STEREO as datalevels?

I think that if we write them down with the same options we should treat them the same, just DL2 - this makes things easier I'd say

kosack:
> JSON list vs simple comma-separated list: What would the consequences of choosing one or the other? could you give more details?

Just aesthetics and maybe simplicity of parsing, no real difference. I guess a simple comma-separated list is simplest

>  how to split stereo data
> I'll stick to my original proposal (which is done here)
> dl2/event/subarray/*estimated quantity*/*algorithm*
> dl2/event/telescope/*estimated quantity*/*algorithm*

That is what we have already, but the question is what to do with the table names under those for the case of the telescope (mono) datasets.  I think I will just mirror DL1, and have them split by tel_id by default (that requires #1690 which is now merged)

> should we add DL2_MONO and DL2_STEREO as datalevels?

It's not so simple, since depending on the use case, you have different things in DL2 data: e.g. you have DL2 MONO/STEREO, or DL2 SHOWER/ENERGY/HADRONNESS (some of which can be missing, for example for training data). So either there is 6 DL2 levels, or we introduce some level-specific DATA_SUBLEVEL as a separate metadata field.  In that case you would have "DL2"  or "DL1" as the datalevel (as specified in the top-level data model document, which does not allow DL1_IMAGE for example), and then each level has it's own sub-levels: 
 *  DL1  sublevels are "IMAGES" and "PARAMETERS"
 *  DL2 sublevels are  "MONO-ENERGY" and "STEREO-ENERGY", "MONO-GEOMETRY", "STEREO-GEOMETRY", "MONO-GAMMANESS", "STEREO-GAMMANESS", ...

For now I will just stick with "DL2", and we can figure out if we want sub-levels later... that would be a nice topic for the data model meeting.

ctapipe/io/datawriter.py
```diff
@@ -53,25 +57,29 @@ def write_reference_metadata_headers(obs_ids, subarray, writer, is_simulation):
         SubarrayDescription to get metadata from
     writer: HDF5TableWriter
         output
+    data_levels: List[DataLevel]
+        list of data levels that were requested/generated
+        (e.g. from `DataWriter.datalevels`)
     """
     activity = PROV.current_activity.provenance
+    category = "Sim" if is_simulation else "Other"
```
maxnoe, line 65: Are these categories defined somewhere? Might an enum make sense?

ctapipe/io/datawriter.py
```diff
@@ -98,33 +105,43 @@ class DL1Writer(Component):
...
+    write_parameters = Bool(
+        help="Store DL1 image parameters if available", default_value=True
+    ).tag(config=True)
+
+    write_stereo_shower = Bool(
+        help="Store DL2 stereo shower geometry if available", default_value=False
+    ).tag(config=True)
+
+    write_mono_shower = Bool(
+        help="Store DL2 stereo mono geometry if available", default_value=False
```
maxnoe, line 134: "stereo mono"?

ctapipe/io/datawriter.py
```diff
@@ -224,7 +241,14 @@ def __call__(self, event: ArrayEventContainer):
             )
 
         # write telescope event data
-        self._write_telescope_events(self._writer, event)
+        self._write_dl1_telescope_events(self._writer, event)
```
maxnoe, line 244: Can we make these methods public? I don't really see a reason why they should be private and it might make using the writer to write specific thinks easier for users

ctapipe/io/hdf5tableio.py
```diff
@@ -127,14 +127,14 @@ def __init__(
         self._group = "/" + group_name
         self.filters = filters
 
-        self.log.debug("h5file: %s", self._h5file)
+        self.log.debug("h5file: %s", self.h5file)
 
     def open(self, filename, **kwargs):
         self.log.debug("kwargs for tables.open_file: %s", kwargs)
-        self._h5file = tables.open_file(filename, **kwargs)
+        self.h5file = tables.open_file(filename, **kwargs)
```
maxnoe, line 134: Any reason in particular to expose the h5file?

ctapipe/io/metadata.py
```diff
@@ -61,9 +60,25 @@ class Product(HasTraits):
     description = Unicode("unknown")
     creation_time = AstroTime()
     id_ = Unicode(help="leave unspecified to automatically generate a UUID")
-    data_category = Enum(["S", "A", "B", "C", "Other"], "Other")
-    data_level = Enum(
-        ["R0", "R1", "DL0", "DL1", "DL2", "DL3", "DL4", "DL5", "DL6", "Other"], "Other"
+    data_category = Enum(["Sim", "A", "B", "C", "Other"], "Other")
+    data_level = List(
+        Enum(
```
maxnoe, line 65: We already have the `ctapipe.io.DataLevel` enum. Would that be usable here?

It is an int enum, maybe we should change to a string enum and then we could just use it here

maxnoe: This should work: `datalevel=List(Enum([level.name for level in DataLevel]))`

ctapipe/io/hdf5tableio.py
kosack: yes, it was already used publicly in a number of places (causing lots of pylint warnings), so for now the simplest thing is to make it public as it's clearly useful. There isn't really a good reason to make it private, unless we later add some new API functions to replace use these use cases.

ctapipe/io/datawriter.py
kosack: I don't think that would break anything, but it may be  confusing since it is a lower-level API function than `write()`.  Normally the user should just call `write(event)` and everything is handled automatically, and if they want something different there should be an option.  If they mix calling write() with calling these lower-level write functions  explicitly, it would lead to duplicated events, but perhaps that is ok.   

If we think the user wants more control, perhaps we should either add more options, or some *hook* functions that can be overridden?  Anyhow, i'm ok making these public, but I think that's better to do in another PR afterward.

ctapipe/io/datawriter.py
kosack: oops. will fix

ctapipe/io/datawriter.py
kosack: yes, they are defined in metadata.py - it only allows "Sim", "A","B","C".  Could use the enum explicitly here though

ctapipe/io/datawriter.py
maxnoe: I was mainly thinking about inheriting from `DL1DataWriter` and overriding these methods with custom implementations. But since the underscore is just convention, I don't have a strong opinion.

ctapipe/io/datawriter.py
kosack: eventually they will maybe need to be set by a user option (since it A/B/C refers to which *official* pipeline produced them: realtime, next-day, offline, but for now don't have much meaning, so "other" is the default).

maxnoe (approved):
Looks good to me.

HealthyPear (approved):

Issue #1627, opened 2021-02-19
`Container.__repr__` should show values.


maxnoe:
The current container `__repr__` is very verbose and especially for containers containing other containers largely unhelpful.

It also doesn't display the actual values, which would be very helpful.

kosack:
I originally implemented it that way, but the output became very large in cases where the values were long and so was pretty ugly.  If you use print() you get the values.  I always thought about swapping that though, and having a `Container.ls()` function that prints the contents of the container, and just have repr and str give back the dictionary. 

Might also implement it like `Component._html_repr_()` where you get a nice notebook version.

kosack:
One issue I have now is `container.to_dict()`, without recursion enabled, gives you a mix of a dict and the reprs of each class, which is not useful.  Probably  `to_dict()` should *always* recurse, or at least by default (which is what you get with str()).

maxnoe:
@kosack I think we should simplify the `__repr__` of a container *instance* to just show the values, essentially making it the same as the dict repr. 

I imagine we should use the current repr for the class, not instances of the class.

PR #1675, opened 2021-04-08
Implement a shower processor

HealthyPear:
This pull request implements a `ShowerProcessor`, similar to the previously implemented `ImageProcessor`.

#### Shower geometry reconstruction is performed with `HillasReconstructor` only

It could be generalized in principle to all `Reconstructor`-based classes, but that class needs to upgraded accordingly and it will probably trigger a different discussion related to e.g. the relation between something like `HillasReconstructor` and `ImPACT`.

I am open to integrate this PR with something like a `reconstructor_type` attribute to `ShowerProcessor` (like `image_cleaner_type` for  `ImageProcessor`), but this would require at least to make `ctapipe.reco.ImPACTReconstructor` not a class based on `Reconstructor`.

#### `ShowerProcessor`as the name says is supposed to process the shower interpreted as an event

Currently when `ShowerProcessor` is called the shower geometry is always reconstructed, being the minimum operation we can do on parametrized images.

It will be also possible to reconstruct energy and particle type, so for the moment, I added them as dummy functions that raises a `NotImplementedError`.

#### Unit-test

The unit-test tests for the following cases:
- successful reconstruction
- unsuccessful reconstruction due to not enough DL1b instances
- call of the currently unsupported energy or particle type reconstructions


ctapipe/reco/shower_processor.py
```diff
@@ -0,0 +1,193 @@
...
+
+DEFAULT_SHOWER_PARAMETERS = ReconstructedShowerContainer()
+
+
+class ShowerQualityQuery(QualityQuery):
+    """Configuring shower-wise data checks."""
+
+    quality_criteria = List(
+        default_value=[
+            ("Finite charge", "lambda hillas: hillas.intensity > 0"),
```
maxnoe, line 29: Finite is not what is tested here. That would be `np.isfinite(hillas.intensity)`, same as above. 

So either use `isinfinite` or call it `Positive charge`

ctapipe/reco/shower_processor.py
```diff
@@ -0,0 +1,193 @@
...
+            ("Finite width", "lambda hillas: hillas.width > 0"),
```
maxnoe, line 30: Doing `hilas.width.value > 0` is much faster than `hillas.width > 0`.

ctapipe/reco/shower_processor.py
```diff
@@ -0,0 +1,193 @@
...
+            ("Finite length", "lambda hillas: hillas.length > 0"),
+            ("Ellipticity > 0.1", "lambda hillas: hillas.width / hillas.length > 0.1"),
```
maxnoe, line 32: How were these cuts here determined? Are they good enough as defaults?

ctapipe/reco/shower_processor.py
```diff
@@ -0,0 +1,193 @@
...
+            ("Ellipticity < 0.6", "lambda hillas: hillas.width / hillas.length < 0.6")
+        ],
+        help=QualityQuery.quality_criteria.help,
+    ).tag(config=True)
+
+
+class ShowerProcessor(Component):
+    """
+    Takes DL1/parameters data and estimates the shower geometry.
+    Should be run after ImageProcessor to produce all DL1b information.
```
maxnoe, line 42: Or with the DL1  EventSource, so maybe just write: Needs DL1_PARAMETERS data as input?

ctapipe/reco/shower_processor.py
HealthyPear: Fixed it (also for the other cuts)

ctapipe/reco/shower_processor.py
HealthyPear: done (also for the other cuts)

ctapipe/reco/shower_processor.py
```diff
@@ -0,0 +1,193 @@
...
+
+    For now the only supported reconstructor is HillasReconstructor.
+    """
+
+    def __init__(
+        self,
+        subarray: SubarrayDescription,
+        is_simulation,
+        reconstruct_energy,
+        classify,
```
maxnoe, line 52: Missing docs for classify

ctapipe/reco/shower_processor.py
```diff
@@ -0,0 +1,193 @@
...
+    Should be run after ImageProcessor to produce all DL1b information.
+
+    For now the only supported reconstructor is HillasReconstructor.
+    """
+
+    def __init__(
+        self,
+        subarray: SubarrayDescription,
+        is_simulation,
+        reconstruct_energy,
```
maxnoe, line 51: Missing docs for reconstruct energy

ctapipe/reco/shower_processor.py
```diff
@@ -0,0 +1,193 @@
...
+            average intensity of the intensities used for reconstruction,
+            measure of algorithm success (if fit),
+            list of tel_ids used if stereo, or None if Mono
+        """
+
+        # Read only valid HillasContainers (minimum condition to continue)
+        hillas_dict = {
+            tel_id: dl1.parameters.hillas
+            for tel_id, dl1 in event.dl1.tel.items()
+            if np.isfinite(event.dl1.tel[tel_id].parameters.hillas.intensity)
```
maxnoe, line 116: Why not use the shower_criteria already here?

ctapipe/reco/shower_processor.py
HealthyPear: The first 3 are minimum conditions I would say, the ellipticity one is the one used in the protopipe/CTAMARS analysis.
These are just defaults, so we can change them if needed.

the defaults of image processor are size >50 and n_pixels> 3 so the first 3 cuts are actually overwritten now that I think about it.

maxnoe: They are needed to ignore non-reconstructed telescope events.

ctapipe/reco/shower_processor.py
```diff
@@ -0,0 +1,193 @@
...
+        # On top of this check if the shower should be considered based
+        # on the user's configuration
+        shower_criteria = [self.check_shower(hillas_dict[tel_id]) for tel_id in hillas_dict]
+        self.log.debug(
+            "shower_criteria: %s",
+            list(zip(self.check_shower.criteria_names[1:], shower_criteria)),
+        )
+
+        # Reconstruct the shower only if all shower criteria are met
+        if np.count_nonzero(shower_criteria) > 2:
```
maxnoe, line 128: You check for at least two telescopes surviving the criteria, but you use all available telescopes here!

I find that surprising. I would expect only those telescopes are used for stereo reconstruction where the requirements are met.

ctapipe/reco/shower_processor.py
HealthyPear: Yes, technically the default shower criteria should also check for this, should we just drop this then?

ctapipe/reco/shower_processor.py
```diff
@@ -0,0 +1,193 @@
...
+
+            return result
+
+        else:
+            return default
+
+    def _reconstruct_energy(self, event: ArrayEventContainer):
+        raise NotImplementedError("TO DO")
+
+    def _estimate_classification(self, event: ArrayEventContainer):
```
maxnoe, line 158: Be more explicit. e.g. `classify_particle_type`

ctapipe/reco/shower_processor.py
HealthyPear: ah right.

maybe a mask on hillas_dict?

ctapipe/reco/shower_processor.py
maxnoe: No, you have to here use it! See comment below.

ctapipe/reco/shower_processor.py
maxnoe: No, just use the quality criterion to create the hillas dict

HealthyPear: ah this is what you meant with the comment above...

ctapipe/reco/shower_processor.py
```diff
@@ -0,0 +1,193 @@
...
+        parent: ctapipe.core.Component or ctapipe.core.Tool
+            Parent of this component in the configuration hierarchy,
+            this is mutually exclusive with passing ``config``
+        """
+
+        super().__init__(config=config, parent=parent, **kwargs)
+        self.subarray = subarray
+        self.check_shower = ShowerQualityQuery(parent=self)
+        self._is_simulation = is_simulation
+        self.reconstruct_energy = reconstruct_energy
```
maxnoe, line 79: This must be traitlets, not kwargs to `__init__`. Otherwise they cannot be set via the configuration.

ctapipe/reco/shower_processor.py
HealthyPear: if we want the defaults to just allow reconstruction, then the ellipticity cut is probably too much

I would at least make sure that width and length are positive, but in case someone changes the defaults of image processor for any reason I would keep also the intensity one

maybe we could add n_pixels also here

the containment cut (or leakage) is probably more related to performance like the ellipticity

maxnoe: These checks are completely independent of the `ImageQualityQuery`, since all events will be processed here. So we need to do all queries we want to do and we cannot rely on a pre-selection by `ImageQualityQuery`.

ctapipe/reco/shower_processor.py
```diff
@@ -0,0 +1,193 @@
...
+        self.classify = classify
+        self.reconstructor = HillasReconstructor()
```
maxnoe, line 81: Add a telescope parameter traitlet `reconstructor_type` and use `from_name`. Like in the `CameraCalibrator` for the extractors.

ctapipe/reco/shower_processor.py
HealthyPear: Ops. I didn't test the configuration form the tool, my bad.

ctapipe/reco/shower_processor.py
```diff
@@ -0,0 +1,193 @@
...
+    Takes DL1/parameters data and estimates the shower geometry.
+    Should be run after ImageProcessor to produce all DL1b information.
+
+    For now the only supported reconstructor is HillasReconstructor.
+    """
+
+    def __init__(
+        self,
+        subarray: SubarrayDescription,
+        is_simulation,
```
maxnoe, line 50: For what do you need `is_simulation`? This shouldn't be necessary. And it can be checked for each event by doing `event.simulation is not None`

ctapipe/reco/shower_processor.py
HealthyPear: If I do this I have to remove `Reconstructor` as a parent class from `Impact` - this is what I meant in the description of the PR.

ctapipe/reco/shower_processor.py
HealthyPear: yes it's an old copy paste from `ImageProcessor` when I was testing, I have to remove it

ctapipe/reco/shower_processor.py
maxnoe: Why?

ctapipe/reco/shower_processor.py
kosack: might be clearer to first build a list of "good" tel_ids insted of a list of shower criteria (which is not indexed by id currently). Something on the lines of the following (though there are multple ways to do it):
```py3
quality_by_tel = {tel_id: self.check_shower(hillas_dict[tel_id]) for tel_id in hillas_dict}
good_tel_ids = [tid for tid,quality in quality_by_tel.items() if all(quality)]
```

Another comment is that you should not apply `check_shower` to  hillas_dict, but to `event.dl1.tel[i].parameters`. That allows one to make selections on *any* parameter, not just hillas parameters.

kosack (changes requested):

ctapipe/reco/shower_processor.py
kosack: I should fix that in ImageProcessor as well - I thought it was very ugly, but actually forgot that event.simulation existed!

maxnoe: We only set it to `None` recently by default. Before you would have to check for nans/specific values.

maxnoe:
@kosack, yes that was what I was essentially also proposing above. Building the hillas dict like:

```
hillas_dict = {
    tel_id: dl1.parameters.hillas
    for tel_id, dl1 in event.dl1.tel.items()
    if all(self.check_event(dl1.parameters))
}
```

ctapipe/reco/shower_processor.py
HealthyPear: because ImPACT as far as I know needs HillasReconstructor to work

in fact  ImPACT-like classes constitute (or should constitute) a different reconstruction class by themselves, for sure more than what Reconstructor does know (which to me now seems at least not up-to-date with what we have in mind). 

Generalizing Reconstructor to initialize all these choices (meaning HillasReconstructor and ImPACT for example) seems too complicated than making ImPACT not rely on Reconstructor at all

especially because after merging #1468 ImPACT will be really different in how it operates

it might work (meaning not crash with defaults) in master now, but as a choice further down the line I'd say it's better to treat it differently

ctapipe/reco/shower_processor.py
HealthyPear: OK I would say then to make it coherent with ImageProcessor, so

- intensity > 50 phe 
- n_pixels > 3

this should automatically ensure that both width and length are positive, though it is possible (but very rare) that width can be 0, so I would also add that as a check.

HealthyPear: Though to access n_pixels I have to start directly from DL1EventSource instead of the hillas dictionary

ctapipe/reco/shower_processor.py
HealthyPear: done

ctapipe/reco/shower_processor.py
HealthyPear: done

ctapipe/reco/shower_processor.py
HealthyPear: done

ctapipe/reco/shower_processor.py
HealthyPear: > Another comment is that you should not apply check_shower to hillas_dict, but to event.dl1.tel[i].parameters. That allows one to make selections on any parameter, not just hillas parameters.

yes in fact to use n_pixels I need to do this

ctapipe/reco/shower_processor.py
HealthyPear: done

ctapipe/reco/shower_processor.py
HealthyPear: done

ctapipe/reco/shower_processor.py
HealthyPear: actually, there is no tool yet...how can I test if this works? I am not really yet used to these traitlets...

HealthyPear: anyway I gues you mean something like some of the options of the image extractors?

HealthyPear: like e.g.

```python
apply_integration_correction = BoolTelescopeParameter(
        default_value=True, help="Apply the integration window correction"
    ).tag(config=True)
```

ctapipe/reco/shower_processor.py
HealthyPear: done

ctapipe/reco/shower_processor.py
HealthyPear: done

ctapipe/reco/shower_processor.py
HealthyPear: done

ctapipe/reco/shower_processor.py
```diff
@@ -108,23 +99,18 @@ def _reconstruct_shower(
...
+            if all(self.check_shower(dl1.parameters))
         }
-
-        # On top of this check if the shower should be considered based
-        # on the user's configuration
-        shower_criteria = [self.check_shower(hillas_dict[tel_id]) for tel_id in hillas_dict]
         self.log.debug(
-            "shower_criteria: %s",
-            list(zip(self.check_shower.criteria_names[1:], shower_criteria)),
+            f"shower_criteria:\n {self.check_shower.to_table()}"
```
maxnoe, line 109: Does this log the shower info for all the events all the time? This might be quite expensive since `to_table` is also called when the level is higher than debug.

HealthyPear: Yes, it should be for all events.
I chose this solution because the fancier hillas_dict uses `self.check_shower` on a telescope-wise quantity, whereas the old syntax on the event, so I didn't want to define `shower_criteria` only for the debug print

HealthyPear: technically it should be the `quality_by_tel` quantity that Karl proposed, but it seemed overly complicated just for the print

ctapipe/reco/shower_processor.py
HealthyPear: done

ctapipe/reco/shower_processor.py
HealthyPear: I think it's ok, let me now if you meant something more specific

ctapipe/reco/shower_processor.py
HealthyPear: as for `classify`

ctapipe/reco/shower_processor.py
HealthyPear: perhaps @kosack can integrate what I said if I missed something from what we discussed

kosack: Yes, we discussed this off-line: the current way things are set up this will not work, except for HillasReconsructor and HillasIntersection (which both inherit from Reconstructor and do actually share a common  predict function), ImPACT works differently, but still inherits from Reconstructor.  It can be considered an "advanced reconstructor" which would be called in a *second step* following the standard reconstructor (since it needs the first-pass reconstruction as input), so it does not make sense to have it an option here.  However, the other reconstructors could be, so just removing the base class Reconstructor from ImPACT would fix the problem (and it doesn't use anything from that base class, so it should not break anything).  But re-factoring the existing reconstructions is out of scope of this PR, so should be done separately.   Once that is done, supporting an option to choose the type should be fine.

ctapipe/reco/shower_processor.py
kosack: Since this is an f-string, it's evaluated for every event, which is bad for performance.  If you change it to:
```python
self.log.debug("shower_criteria:\n %s", self.check_shower.to_table())
```
The call to `to_table()` will only be made if the log-level is debug, I think.  

However, if this is useful information, we should consider just adding it to the Container so it's written out for each event, rather than a debug statement.  If you add a field like "quality_flags", it would get written as a boolean vector column.

HealthyPear: Though this is telescope-wise information, not really shower-wise no?

I thought we decided not to add telescope-wise data to DL2 (I guess the container address you are referring to is `event.dl2.shower["HillasReconstructor"]` for example)

When we print `self.check_shower.to_table()` we obtain something like,

```
   criteria    counts cumulative_counts
-------------- ------ -----------------
         TOTAL      8                 8
      > 50 phe      8                 8
Positive width      8                 8
    > 3 pixels      8                 8
```
but what we really use to filter the hillas dictionaries is 

`self.check_shower(dl1.parameters)`

which is something like `[ True  True  True]` for each `tel_id`

so what do we want to save?

I'd say that a dict of dicts like `{tel_id: {cut_name: bool}}` would be the most informative choice, but the container we are talking about here is DL2...

HealthyPear: For the moment I changed the debug logging to a normal `%` format string

kosack: You actually want to save two things: for each event you want to save the results of check_shower into a Field (it will become a column of boolean arrays when written), and also at the end you want to write the to_table() information to the service group as we did in the DL1 part (that contains overall statistics, but not event-wise information). 

So add a Field to the container like `quality_flags`, and place the output of `check_shower(params)`  there.

HealthyPear: But what you call "boolean arrays" tells you if an image passed **all** the checks, not which specific cuts

maxnoe: It does, you store each flag for each event (a boolean array per event, resulting in a 2d (n_events, n_checks) column)

HealthyPear: but here by event you mean the shower or the image?
because we want to write this to ReconstructedShowerContainer

maxnoe: No, this needs to be written per telescope.

HealthyPear: I mean, when you do `self.check_shower(dl1.parameters)`

you get for example

`[True False True]`

which means there are 3 images and only 1 was not used, because **one** of the cuts was not passed , but you cannot know which one from this information

HealthyPear: ah no sorry

HealthyPear: this is for 1 shower, with 3 images and each boolean is 1 cut from the quality query - right?

ctapipe/reco/shower_processor.py
```diff
@@ -0,0 +1,194 @@
...
+        """
+
+        # Select only images which pass the shower quality criteria
+        hillas_dict = {
+            tel_id: dl1.parameters.hillas
+            for tel_id, dl1 in event.dl1.tel.items()
+            if all(self.check_shower(dl1.parameters))
+        }
+        self.log.debug(
+            "shower_criteria:\n %s", self.check_shower.to_table()
```
maxnoe, line 118: Just because you changed how this is formatted, `to_table` is still a pretty expensive operation that is executed here everytime

HealthyPear: Indeed, but we are proposing here to not input this to the logger, but rather write it directly to output

HealthyPear: such an info is anyway too complex to make sense printing it to prompt

ctapipe/reco/shower_processor.py
HealthyPear: then by using a Map we should link the tel_ids with each cut-related boolean, am I correct?

kosack: You may want to wait until we merge #1673 since that updates the internal data model to have a place to put per-telescope DL2 data, which is missing in this version.  Then you can add this boolean there - I just realized that may be what is causing confusion.

HealthyPear: yes you're right. I'll try to review again that so we can proceed!

ctapipe/reco/shower_processor.py
kosack: if you really want to keep it for now, you can wrap it in:
```py3
if self.log.isEnabledFor(logging.DEBUG):
    self.log.debug(
            "shower_criteria:\n %s", self.check_shower.to_table()
```

That way it's not executed unless the log-level is DEBUG. Another option would be to add a __str__ method to QualityQuery that returns the table in string format, that way if you write:
```py3
self.log.debug("shower_criteria:\n %s", self.check_shower)
```
the table method it is only called if debug is on, and no if-statement is needed

HealthyPear: I like the second proposal

HealthyPear: done

HealthyPear:
if writing the shower criteria to file is urgent we can wait #1673 otherwise we can leave it to a subsequent PR and merge this one which I think now is OK unless there are other comments

ctapipe/reco/shower_processor.py
```diff
@@ -0,0 +1,197 @@
...
+
+class ShowerProcessor(Component):
+    """
+    Needs DL1_PARAMETERS as input.
+    Should be run after ImageProcessor to produce all DL1b information.
+
+    For the moment it only supports the reconstruction of the shower geometry
+    using ctapipe.reco.HillasReconstructor.
+
+    It is planned to support also energy reconstruction and particle type
```
maxnoe, line 48: Then don't add them here please. We made bad experience with stubs. Just don't implement anything that is not yet implemented please.

maxnoe (changes requested):
Please remove all the `raise not ImplementedError` parts.

I would also make the  single functions public API (not leading with an underscore).

ctapipe/reco/shower_processor.py
HealthyPear: this (and the `raise not ImplementedError`s ) was suggested by @kosack but I have not strong preference for now

maxnoe: We made pretty bad experience with this kind of premature API design in the past, I'd say. 

It's not usable now and it might elad to restricting the actual imlpementation when it's done unnecessarily. It's just dead code. And it confuses users.

The methods being there and documented might on first glance lead to users thinking it's already implemented, only to be then badly surprised by `NotImplemented`.

I am really not a fan of having `raise NotImplementedError` functions / methods.

HealthyPear:
> I would also make the single functions public API (not leading with an underscore).

This follows what has been done in `ctapipe.image.ImageProcessor,` also for this I have no strong preference (but if there is a reason to change this, I'd say we have to do it in both of the classes).

ctapipe/reco/shower_processor.py
maxnoe: Ok, then I would just (for now) remove the base class from `ImPACT`.

maxnoe:
Yes, this is also something I wanted to change. I think it would be good to expose also the single steps in the `ImageProcessor` and test them as public API. Not just the `__call__`.

ctapipe/reco/shower_processor.py
HealthyPear: Even if this doesn't really change anything here, I'd do this change separately

maxnoe: ok, so for now, it's always `HillasReconstructor` and we can make another PR for enabling the choice of reconstructor.

HealthyPear:
I have some documentation warnings (which are now treated as errors) which I cannot reproduce locally - they seem to appear only on the CI.

maxnoe:
@HealthyPear I get the same errors locally when also running with the `-n --keep-going -W` options.

We should add those to the makefile, so they are also used always.

ctapipe/reco/shower_processor.py
kosack: You can remove them - in the end the only function the user should call is  the __call__ method anyhow, so if new features are added, it doesn't break anything.  It is important I think to remember that this is intended to also do energy and particle type reco,  so mentioning it in the doc is I think fine. (in case somebody tries to implement it outside).

ctapipe/reco/shower_processor.py
```diff
@@ -0,0 +1,194 @@
...
+        super().__init__(config=config, parent=parent, **kwargs)
+        self.subarray = subarray
+        self.check_shower = ShowerQualityQuery(parent=self)
+        self.reconstructor = HillasReconstructor()
+
+    def _reconstruct_shower(
+        self,
+        event,
+        default=DEFAULT_SHOWER_PARAMETERS,
+    ) -> ReconstructedShowerContainer:
```
kosack, line 89: I think you need to pull in the latest changes from master, since the DL2 PR defined all the classes needed for DL2 storage (and they are no longer called the same things)

kosack: This is `ReconstructedGeometryContainer` now

ctapipe/reco/shower_processor.py
HealthyPear: done

ctapipe/reco/shower_processor.py
```diff
@@ -0,0 +1,176 @@
...
+            return default
+
+    def process_shower_geometry(self, event: ArrayEventContainer):
+        """Record the reconstructed shower geometry into the ArrayEventContainer."""
+
+        shower_geometry = self._reconstruct_shower(event)
+
+        self.log.debug("shower geometry: %s", shower_geometry.as_dict(recursive=True))
+
+        event.dl2.shower["HillasReconstructor"] = shower_geometry
```
kosack, line 160: this is now in `event.dl2.stereo.geometry['HillasReconstructor']`

kosack (changes requested):

ctapipe/reco/shower_processor.py
HealthyPear: yes working on it

kosack:
> Yes, this is also something I wanted to change. I think it would be good to expose also the single steps in the I mageProcessor and test them as public API. Not just the __call__.
>Similar for the steps in Stage1, so pipelines can easier add custom behaviour by inheriting from stage1 and overriding / adding some methods.

Before we do that I would prefer to collect some use cases to see what another pipeline would really want to override. The current private API is in my opinion not really designed for that purpose, so just making it public is not the best design.

ctapipe/reco/shower_processor.py
```diff
@@ -0,0 +1,176 @@
...
+        self,
+        event,
+        default=DEFAULT_SHOWER_PARAMETERS,
+    ) -> ReconstructedGeometryContainer:
+        """Perform shower reconstruction.
+
+        Parameters
+        ----------
+        event : container
+            A `ctapipe` event container
```
kosack, line 95: I guess this needs to be either  \`\`ctapipe``  or else just get rid of the backtics (which make it  a link in Sphinx)

HealthyPear: just fixed, should be fine now

maxnoe:
@HealthyPear I think we can go ahead here independent of that and talk about what needs to be adapted / changed to have a more fine-grained public API later / when I will continue with trying to port the full LST analysis to vanilla ctapipe.

HealthyPear:
> @HealthyPear I think we can go ahead here independent of that and talk about what needs to be adapted / changed to have a more fine-grained public API later / when I will continue with trying to port the full LST analysis to vanilla ctapipe.

Sure for me it's also fine, I do not have preferences.
From the point of view of _protopipe_ I think the ShowerProcessor will be a black box after initialization (unless bugs or design problems arise, which is still possible but that's the reason we have protopipe to test...).

I do not have further modifications unless you or Karl have other comments.

kosack (approved):

ctapipe/reco/__init__.py
```diff
@@ -1,13 +1,12 @@
 # Licensed under a 3-clause BSD style license - see LICENSE.rst
-from .reco_algorithms import Reconstructor
-from .hillas_reconstructor import HillasReconstructor
-from .impact import ImPACTReconstructor
+from .hillas_reconstructor import HillasReconstructor, Reconstructor
```
maxnoe, line 2: Please import the `Reconstructor` from the place where it is defined, not some other module where it is used. It was good before.

ctapipe/reco/shower_processor.py
```diff
@@ -0,0 +1,176 @@
...
+            ("> 3 pixels", "lambda p: p.morphology.num_pixels > 3"),
+        ],
+        help=QualityQuery.quality_criteria.help,
+    ).tag(config=True)
+
+
+class ShowerProcessor(Component):
+    """
+    Needs DL1_PARAMETERS as input.
+    Should be run after ImageProcessor to produce all DL1b information.
```
maxnoe, line 40: "to produce DL1b information" sounds like *this* class produces the dl1b information.

kosack: Also, isn't it DL2?  (and the terms DL1a/b DL2a/b are no longer used - they are not well defined).

maxnoe: I think he meant to say that the ImageProcessor needs to be run to produce the DL1b which is required as input here.

ctapipe/reco/__init__.py
HealthyPear: done

ctapipe/reco/shower_processor.py
HealthyPear: fixed

ctapipe/reco/shower_processor.py
```diff
@@ -0,0 +1,176 @@
...
+                Returning default ReconstructedGeometryContainer container"""
+            )
+            return default
+
+    def process_shower_geometry(self, event: ArrayEventContainer):
+        """Record the reconstructed shower geometry into the ArrayEventContainer."""
+
+        shower_geometry = self.reconstruct_shower(event)
+
+        self.log.debug("shower geometry: %s", shower_geometry.as_dict(recursive=True))
```
maxnoe, line 158: This will also already call `as_dict` for each event, not only when it actually is logged.

ctapipe/reco/shower_processor.py
```diff
@@ -0,0 +1,176 @@
...
+    Should be run after ImageProcessor which produces such information.
+
+    For the moment it only supports the reconstruction of the shower geometry
+    using ctapipe.reco.HillasReconstructor.
+
+    It is planned to support also energy reconstruction and particle type
+    classification which are now disabled by default.
+    """
+
+    reconstruct_energy = Bool(
```
maxnoe, line 49: These two options should also be removed, now that the corresponding methods are gone

ctapipe/reco/shower_processor.py
HealthyPear: how should it be printed? would a simpler

`self.log.debug("shower geometry:\n %s", shower_geometry)`

work?

ctapipe/reco/shower_processor.py
HealthyPear: done

ctapipe/reco/tests/test_shower_processor.py
```diff
@@ -0,0 +1,72 @@
...
+from ctapipe.reco import ShowerProcessor
+
+
+def test_shower_processor_geometry(example_event, example_subarray):
+    """Ensure we get shower geometry when we input an event with parametrized images."""
+
+    calibrate = CameraCalibrator(subarray=example_subarray)
+
+    config = Config()
+    config.ShowerProcessor.reconstruct_energy = False
```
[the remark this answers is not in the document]
HealthyPear: done

ctapipe/reco/shower_processor.py
maxnoe: no, that wouldn't show the values, just the container definition. https://github.com/cta-observatory/ctapipe/issues/1627

HealthyPear: what would be a work-around for now?

maxnoe: not logging here.

HealthyPear: done

ctapipe/reco/shower_processor.py
```diff
@@ -0,0 +1,166 @@
...
+            Parent of this component in the configuration hierarchy,
+            this is mutually exclusive with passing ``config``
+        """
+
+        super().__init__(config=config, parent=parent, **kwargs)
+        self.subarray = subarray
+        self.check_shower = ShowerQualityQuery(parent=self)
+        self.reconstructor = HillasReconstructor()
+
+    def reconstruct_shower(
```
maxnoe, line 75: Since the container is now called `ReconstructedGeometryContainer`, should we call this `reconstruct_geometry`?

HealthyPear: done

maxnoe (approved):

kosack (approved):

PR #1677, opened 2021-04-09
Fix some documentation warnings

HealthyPear:
Just some easy-to-fix warnings that I was finding annoying...


kosack (approved):
thanks!

docs/ctapipe_api/containers.rst
```diff
@@ -1,19 +0,0 @@
-.. _containers:
```
maxnoe, line 1: Why was this file removed?

maxnoe: Ah sorry, duplicate with the `containers/index.rst`

kosack: does that really cause a warning? I use that trick elsewhere to have one item appear at the top of the list, independently of how the glob is sorted (though here that is not needed)... e.g. to get a list where you have everything in alphabetical order, except one item that should be pinned to the top. 

In any case, this is fine here.

maxnoe (approved):

docs/ctapipe_api/containers.rst
HealthyPear: the warning was caused by the fact that both the label and the referenced module were the same, so it was seen by sphinx as a duplication.

In any case as far as I saw it didn't make sense to have them duplicated and I just merged the description to make it more clear.

PR #1680, opened 2021-04-19
Do not use numba cache for lts_linear_fit

maxnoe:
* Numba cache seems broken for this function, it is recreated on every
new process
* That lead to ~6 GB of cache files on a cluster with 23k files
* deactivated the cache for this function for now

I couldn't figure out why the cache does not work properly for this function.

### DIFF
--- a/ctapipe/fitting.py
+++ b/ctapipe/fitting.py
@@ -83,7 +83,7 @@ def _lts_single_sample(X, y, sample_size, max_iterations, eps=1e-12):
     if np.isnan(beta[0]):
         return beta, np.nan
 
-    last_error = residual_sum_of_squares(X[sample], y[sample], beta)
+    error = residual_sum_of_squares(X[sample], y[sample], beta)
 
     for i in range(max_iterations):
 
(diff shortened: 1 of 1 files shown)


LukasNickel (approved):

kosack:
Is it possible it is treating different length inputs as different versions of the function? (e.g. 5 points in the fit vs 6 points)?  Normally, it would only generate new functions for different data types, but it would explain it.  Perghaps giving the function signature in the @jit call would fix it?

kosack (approved):

maxnoe:
> Is it possible it is treating different length inputs as different versions of the function? (e.g. 5 points in the fit vs 6 points)? Normally, it would only generate new functions for different data types, but it would explain it. Perghaps giving the function signature in the @jit call would fix it?

No, this was my first suspicion as well and I ran it like this:

```
for n in range(3, 10000):
     x = np.random.uniform(0, 1, n)
     y = 5 * x + 2
     lts_linear_regression(x, y)
```

and it created one new cache file per run of the script.

PR #1681, opened 2021-04-19
Fix syntax typo in call of TwoPassWindowSum with 2ndPass disabled

HealthyPear:
This typo was resulting in a fatal crash when launching CameraCalibrator with this image extractor when the 2nd pass was disabled (tested from ctapipe-stage1 tool).

Added portion of unit-test which tests also against the case in which 2nd pass is disabled

### DIFF
--- a/ctapipe/image/extractor.py
+++ b/ctapipe/image/extractor.py
@@ -1190,23 +1190,23 @@ class TwoPassWindowSum(ImageExtractor):
             Shape: (n_pix)
         """
 
-        self._charge1, self._pulse_time1, correction1 = self._apply_first_pass(
+        charge1, pulse_time1, correction1 = self._apply_first_pass(
             waveforms, telid
         )
 
         # FIXME: properly make sure that output is 32Bit instead of downcasting here
         if self.disable_second_pass:
             return (
-                (self._charge1 * correction1[selected_gain_channel]).astype("float32"),
-                self._pulse_time1.astype("float32"),
+                (charge1 * correction1[selected_gain_channel]).astype("float32"),
+                pulse_time1.astype("float32"),
             )
 
         charge2, pulse_time2 = self._apply_second_pass(
             waveforms,
             telid,
             selected_gain_channel,
-            self._charge1,
-            self._pulse_time1,
+            charge1,
+            pulse_time1,
             correction1,
         )
         # FIXME: properly make sure that output is 32Bit instead of downcasting here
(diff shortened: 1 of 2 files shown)


ctapipe/image/extractor.py
```diff
@@ -1190,23 +1190,23 @@ def __call__(self, waveforms, telid, selected_gain_channel):
...
+        charge1, pulse_time1, correction1 = self._apply_first_pass(
             waveforms, telid
         )
 
         # FIXME: properly make sure that output is 32Bit instead of downcasting here
         if self.disable_second_pass:
             return (
-                (self.charge1 * correction1[selected_gain_channel]).astype("float32"),
-                self._pulse_time1.astype("float32"),
+                (charge1 * correction1[selected_gain_channel]).astype("float32"),
```
kosack, line 1200: the typo was here I guess, since self.charge1 was used instead of self._charge1.  The reason for making those class variables (rather than local ones) was to  support being able to debug the first and second passes without having to re-compile or re-run the code.  Is that no longer needed? I.e if you want to check the output of the first pass only, you need these.

maxnoe: No, since you can deactivate the second pass. So you can just deactivate the second pass to check only the first pass.

maxnoe: Or you just directly call the `first_pass` and `second_pass` methods.

HealthyPear: this is what I did now: I added a test where I disable 2nd pass (which is indeed a traitlet by itself, so I guess we debug 1st pass by disablng the 2nd)

ctapipe/image/tests/test_extractor.py
```diff
@@ -348,6 +348,22 @@ def test_two_pass_window_sum(subarray):
         assert_allclose(charge, true_charge, rtol=0.1)
         assert_allclose(pulse_time, true_time, rtol=0.1)
 
+    # test only 1st pass
+    extractor.disable_second_pass = False
```
maxnoe: For only the first pass, `disable_second_pass=True`, right?

HealthyPear: yes I should have fixed it few minutes ago

HealthyPear: oh god actually it was correct before ahah
now I get what you mean

kosack (approved):

maxnoe (approved):

PR #1685, opened 2021-04-23
Store package versions in provenance

maxnoe:


kosack:
One question is if we should store all packages here , or limit it to  the ctapipe dependencies as done in ctapipe-info? If running in a shared environment, the list of packages could be very large (in my development env, I have 270 packages installed, but don't really care about the versions of most as they have no effect on the data generated by ctapipe)

maxnoe:
We can think about this, yes. But identifying all transitive dependencies, as is needed for fully reproducing the requirement, might be hard.

kosack:
Yeah, I can't think of a way to get the list of dependencies constantly without just hard-coding it.  Though I didn't look too hard. It may not be possible/easy.

kosack (approved):

maxnoe:
We could use `sys.modules`, which are all the imported modules.

LukasNickel (approved):

PR #1690, opened 2021-05-03
implements add_column_transform_regexp feature

kosack:
This allows one to add column transforms by patterns that match the table name and column name, using re.fullmatch.  It adds a new function `add_column_transform_regexp()`, so the original API does not change.

It also adds an internal helper function `_realize_regexp_transforms()` that converts transform patterns into explicit transform dicts that are applied as before.  This way there is no speed impact compared to the original implementation, as this function is only called during schema generation.

This feature is needed in order to make the DL2 output of #1673 work if we split tables by tel_id, since it allows one to add column transforms to all tables for each algorithm without a huge and complex loop over all possible names.


maxnoe (approved):
LGTM

PR #1693, opened 2021-05-05
Remove unused io modules, fixes #1527

### LINKED ISSUE #1527: More code removals
1. [ ] https://github.com/cta-observatory/ctapipe/blob/master/ctapipe/io/array.py seems not to be used anywhere, has no tests and is probably superseeded by `SubarrayDescription.from_hdf / to_table / ...` 
2. [ ] https://github.com/cta-observatory/ctapipe/blob/master/ctapipe/io/files.py Not used, has a test, but uses IMHO the wrong approach, instead of looking at the actual file contents and checking for magic bytes, it just uses file extension. Also there are very well tested libraries we could use if we really needs this kind of functionality. Mostly stuff like this is implemented in `EventSource.is_compatible` implementations

maxnoe:


LukasNickel:
`ctapipe/ctapipe/io/tests/test_files.py ` needs to be removed as well.
It only tests the `get_file_type` function we want to get rid off

maxnoe:
thx, done

LukasNickel (approved):

kosack (approved):

PR #1697, opened 2021-05-05
Consistent use of allowed_tels 

LukasNickel:
Previously, `allowed_tels` only dropped the telescope events of unselected telescopes.
With the changes in this PR, this behaviour is extended to also select the telescopes in the `Subarraydescription` and the event-wise trigger information.
This makes it easier to continue to work with these files as all information in the file should now be consistent.


ctapipe/io/dl1eventsource.py
```diff
@@ -35,7 +35,7 @@
 logger = logging.getLogger(__name__)
 
 
-COMPATIBLE_DL1_VERSIONS = ["v1.0.0", "v1.0.1", "v1.0.2", "v1.0.3", "v1.1.0"]
+COMPATIBLE_DL1_VERSIONS = ["v1.0.0", "v1.0.1", "v1.0.2", "v1.0.3", "v1.1.0", "v1.2.0"]
```
maxnoe, line 38: We didn't actually change anything in the data model, right?

ctapipe/io/simteleventsource.py
```diff
@@ -340,6 +340,12 @@ def is_compatible(file_path):
 
     @property
     def subarray(self):
+        if self.allowed_tels:
+            return self._subarray_info.select_subarray(
+                self._subarray_info.name
+                + "_subset".join(f"_{t}" for t in self.allowed_tels),
```
maxnoe, line 346: This can get very long. Maybe better use a notation like `1-4,60-80` like implemented in here: https://github.com/cta-observatory/ctapipe/pull/1557

ctapipe/io/simteleventsource.py
```diff
@@ -471,13 +476,17 @@ def _fill_trigger_info(data, array_event):
             data.trigger.event_type = EventType.UNKNOWN
 
         data.trigger.tels_with_trigger = trigger["triggered_telescopes"]
+        if self.allowed_tels:
+            data.trigger.tels_with_trigger = np.intersect1d(
+                data.trigger.tels_with_trigger, np.array(list(self.allowed_tels))
```
maxnoe, line 481: calling `array(list(...` every time is not necessary.

ctapipe/io/simteleventsource.py
```diff
@@ -471,13 +476,17 @@ def _fill_trigger_info(data, array_event):
...
+            )
         central_time = parse_simtel_time(trigger["gps_time"])
         data.trigger.time = central_time
 
         for tel_id, time in zip(
-            trigger["triggered_telescopes"], trigger["trigger_times"]
+            data.trigger.tels_with_trigger, trigger["trigger_times"]
```
maxnoe, line 487: Is this correct? I think you are messing up trigger times and telescope ids here.

ctapipe/io/dl1eventsource.py
LukasNickel: Kind of. Using the stage1-tool with `allowed_tels` now leads to a different file as the subarray is different

ctapipe/io/simteleventsource.py
LukasNickel: Also didnt change the name in the dl1eventsource

ctapipe/io/simteleventsource.py
LukasNickel: Should I call it once in the init?

ctapipe/io/dl1eventsource.py
maxnoe: No, this is not a change to the data model.

ctapipe/io/dl1eventsource.py
```diff
@@ -313,9 +319,13 @@ def _generate_events(self):
             data.count = counter
             data.trigger = trigger
             data.index = index
-            data.trigger.tels_with_trigger = self.subarray.tel_mask_to_tel_ids(
+            data.trigger.tels_with_trigger = self._full_subarray_info.tel_mask_to_tel_ids(
```
kosack, line 322: If you want to be consistent, this should not be the full subarray info, but the selected subarray.  Otherwise we would have to store both in the output file in order to understand how to convert between index and id.   I think there is no real reason to keep the full subarray around. . 

For example, if we subselect only 4 telescopes, it would be correct to have the tel_mask be only 4 elements wide, rather than whatever the super-array was (which could be very large for some simulations).  The SubarrayDescription is always what is used to map indices to tel_ids.

kosack (changes requested):

ctapipe/io/dl1eventsource.py
maxnoe: I think here it needs to be read with the full array (like it is in the file) and then converted to another mask later. So I think this here is correct.

LukasNickel: But this is an array of booleans with length equal to the number of telescopes in the full array.
I think the easiest way to handle it is to select from the full subarray and only choose the allowed telescopes afterwards, because you cant just match that mask with the reduced subarray.

If you see an easier way, I'm open to change it, I just failed to think of one

kosack: ah, yes I understand now why this is not easy (it's a sub-selection a the DL1 level after a possible sub-selection at the DL0 level). Ok, fine then, you can ignore my reques.

ctapipe/io/simteleventsource.py
maxnoe: This should just be `subarray.tel_ids`, which is cached.

maxnoe: Also, https://numpy.org/doc/stable/reference/generated/numpy.intersect1d.html

We can `assume_unique` here, right? No duplicated tel ids

LukasNickel:
Does anyone use the possibility to set `allowed_tels` after initializing the event source?
I am in favor of dropping the support for this in order to avoid blowing up the code and avoiding bugs that might occur in the future.

HealthyPear:
> Does anyone use the possibility to set `allowed_tels` after initializing the event source?
I am in favor of dropping the support for this in order to avoid blowing up the code and avoiding bugs that might occur in the future.

In protopipe I do not do it, so for me it is ok to drop it.

LukasNickel:
> > Does anyone use the possibility to set `allowed_tels` after initializing the event source?
> > I am in favor of dropping the support for this in order to avoid blowing up the code and avoiding bugs that might occur in the future.
> 
> In protopipe I do not do it, so for me it is ok to drop it.

Its more about interactive use I think.
You can avoid defining a new event source this way if you only want to change the subarray.

LukasNickel:
I dont think that exists anymore.
At least I can not find it in the examples.

ctapipe/io/dl1eventsource.py
```diff
@@ -97,9 +97,23 @@ def __init__(self, input_url=None, config=None, parent=None, **kwargs):
 
         self.file_ = tables.open_file(self.input_url)
         self._full_subarray_info = SubarrayDescription.from_hdf(self.input_url)
+
         if self.allowed_tels:
```
maxnoe, line 101: this code should go into a function or method of `EventSource` or probably even into `Subarray.select_subarray` itself.

You have it both here and in the simtel event source.

ctapipe/instrument/subarray.py
```diff
@@ -314,6 +313,24 @@ def select_subarray(self, name, tel_ids):
         tel_positions = {tid: self.positions[tid] for tid in tel_ids}
         tel_descriptions = {tid: self.tel[tid] for tid in tel_ids}
 
+        if not name:
+            # sets don't interact nicely with numpy, so we transform it to a list
+            if isinstance(tel_ids, set):
+                tel_ids = list(tel_ids)
```
maxnoe, line 319: just always do `tel_ids = sorted(tel_ids)`, that will work with any sequence and return a list with the sorted tel ids (which you rely on when using diff).

ctapipe/instrument/subarray.py
```diff
@@ -314,6 +313,24 @@ def select_subarray(self, name, tel_ids):
...
+            name = self.name + "_"
+            tel_id_delta = np.diff(tel_ids)
+            splits = np.where(tel_id_delta > 1)[0] + 1
+            if splits:
+                id_ranges = np.split(tel_ids, splits)
+            else:
+                # If all allowed ids are contiguous integers the split will be empty
+                id_ranges = ((min(tel_ids), max(tel_ids)),)
+            for id_range in id_ranges:
+                name += f",{min(id_range)}-{max(id_range)}"
```
maxnoe, line 332: name += ",".join(... for id_range in id_ranges)

LukasNickel: Actually there are functions to get the ranges in `subarray.py` that are used for the info method.
Can probably use them

ctapipe/io/simteleventsource.py
```diff
@@ -471,13 +466,22 @@ def _fill_trigger_info(data, array_event):
...
+                assume_unique=True,
+            )
         central_time = parse_simtel_time(trigger["gps_time"])
         data.trigger.time = central_time
 
         for tel_id, time in zip(
             trigger["triggered_telescopes"], trigger["trigger_times"]
         ):
-            # telesocpe time is relative to central trigger in ns
+            if self.allowed_tels:
```
maxnoe, line 481: combine to one if:

```
if self.allowed_tels and tel_id not in self.allowed_tels:
```

python won't evaluate the second half if the first is already not true

LukasNickel: I need to remember that one finally

maxnoe (dismissed):

nbiederbeck (dismissed):

maxnoe (approved):

PR #1698, opened 2021-05-06
Fix bugs in TwoPassWindowSum and define better unit-test

HealthyPear:
#### Description

- fixed 2nd pass mask which was not taking into account core pixels outside the main cluster

- if the main cluster has more than 3 pixels it doesn't matter their charge, we go for 2nd pass anyway

- the parameterization and timing fit has to be done on the brightest image and not the biggest

- the edge cases were ill-defined: they depend on the position of the predicted integration window and not only the peak time 

- the edge case for which a predicted pulse time (correction from the previous point: any time the integration window...)  falls after the readout window had the wrong settings for shift and width

you can check this last point with the following code snippet which simulates the case for LSTCam with code from `extract_around_peak,`

```python
n_samples = 30
peak_index = 29
shift = 4                 # before it was 5
width = 6              # before it was 5
start = peak_index - shift
end = start + width
start = max(0, start)
end = min(end, n_samples)
for i in range(start, end):
    print(i)
```

#### TODO

- [x] parameterization and timing fit should be applied on the **brightest** island and not the _biggest_

- [x] "If the predicted time for a pixel leads to a **window** outside the readout window, then the last (or first) 5 samples are integrated."

- [x] I tested these changes directly on CTAMARS data (see cta-observatory/protopipe#129 for preliminary results), but a better unit-test should be defined. The current one in use is not good for image extractors that use a timing fit.
- [x] Understand why the test fails (some bug? too strict? overkill?)


ctapipe/image/extractor.py
```diff
@@ -1006,22 +1003,16 @@ def _apply_second_pass(
         else:
             mask_biggest = mask_clean
 
+        # for all pixels except the core ones in the main island of the
+        # preliminary image, the waveform will be integrated once more (2nd pass)
+
+        mask_2nd_pass = ~mask_biggest | (mask_biggest & (charge_1stpass < core_th))
```
LukasNickel, line 1009: Isn't this easier done with
`mask_2nd_pass = ~ (mask_biggest & (charge_1stpass > core_th))`
?

Also I think this can be done after the part where you check that the image has at least 3 pixels

HealthyPear: > Isn't this easier done with
> `mask_2nd_pass = ~ (mask_biggest & (charge_1stpass > core_th))`?

I wrote in the way that for me was more readable, and I still think it is

> Also I think this can be done after the part where you check that the image has at least 3 pixels

It shouldn't change the result, but yes it makes more sense logically

LukasNickel:
How close do you expect the charge after second pass to match the true charge? The second pass is matching much better than the first one and the charge is always underestimated, which to me indicates like the integration window is just too narrow in these cases. The wiki also mentions a bias due to not taking the pulse tails into account.

ctapipe/image/extractor.py
```diff
@@ -34,7 +34,7 @@
 from scipy.ndimage.filters import convolve1d
 from typing import Tuple
 
-from . import number_of_islands, largest_island, tailcuts_clean
+from . import number_of_islands, largest_island, tailcuts_clean, brightest_island
```
LukasNickel, line 37: `largest islands` is not used anmore

LukasNickel:
It seems like you edited my comment instead of replying, looks a bit weird now

HealthyPear:
> How close do you expect the charge after second pass to match the true charge? The second pass is matching much better than the first one and the charge is always underestimated, which to me indicates like the integration window is just too narrow in these cases.

It is not always sure: 2nd pass is based on a timing fit of a cleaned image, so it is susceptible of error.
The main "problem" here is that this test uses no noise, which is probably not ideal for this image extractor, which is made to recover weak signal charge from boundary pixels of the main (brightest) island + signal pixels killed by the preliminary image cleaning and kill noise - but by killing noise you of course loose a bit of signal.

For these reasons I divided to test only "nice" signal pixels, but it seems that the rtol I am using is too strict.
Also at some point this becomes more of a benchmark than a unit-test...

> The wiki also mentions a bias due to not taking the pulse tails into account. 

And we already correcting for this (integration correction function)

HealthyPear:
> > And we already correcting for this (integration correction function)
> 
> Not really. The integration window corrections assumes you are still integrating the central part of the pulse.
> 
> However, when using the time gradient to set the window, you integrate different parts of the pulse than what the correction was calculated for.

we are re-calculate the correction factors over the new integration windows later

maxnoe:
> we re-calculate the correction factors over the new integration windows later

For a different width, right? Not for the fact that the pulse might not be centered in the window.

HealthyPear:
> For a different width, right? Not for the fact that the pulse might not be centered in the window.

Right. `integration_correction` shouldn't correct for this I think...

Anyway the bias @LukasNickel was referring about I think it's the one which is corrected by `integration_correction`, but now I am not sure about the "full" correction

maxnoe (dismissed):

maxnoe (approved):

LukasNickel:
Maybe a bit more general:
The test image is always the same due to the fixed random seed. Would it not be better (in terms of acting as a regression test in the future) to define the expected expected charge and use a stricter tolerance?
The tolerance is now so big, that small changes in the future might pass the tests unnoticed.

LukasNickel:
> > to define the expected charge
> 
> what do you mean by expected charge?
> isn't it simply the true charge from the generated image?

Is it? The algorithm returns different values now or did I miss a commit?
So, if we assume the algorithm to work correct, the algorithm is correctly returning different values. I would say it is not very helpful to compare this result to the true values if we know/expect it to be different.

It is meant more as a question if people think defining tests with a huge tolerance is helpful in the long term. I have an opinion, but I would like to hear different views.

Anyway it does not necessarily need to block this PR, was just wondering.

maxnoe:
@LukasNickel certainly, have  a large rtol is not really good. However, using a small one that works for this specific random seed is also not better, since subtle changes not changing anything really will also make it fail.

I think we should go ahead here. The unit test tests that the extractor is working and producing sensible results. The rest is probably for benchmarks.

PR #1703, opened 2021-05-07
Raise error in CameraReadout.from_name if no corresponding file can be found

maxnoe:
* Raise error in CameraReadout.from_name if requested data is not available
* Set `readout` to None in `CameraDescription.from_name` if that exception is raised.

### DIFF
--- a/ctapipe/instrument/camera/readout.py
+++ b/ctapipe/instrument/camera/readout.py
@@ -111,25 +111,11 @@ class CameraReadout:
         else:
             verstr = f"-{version:03d}"
 
-        try:
-            tabname = "{camera_name}{verstr}.camreadout".format(
-                camera_name=camera_name, verstr=verstr
-            )
-            table = get_table_dataset(tabname, role="dl0.tel.svc.camera")
-            return CameraReadout.from_table(table)
-        except FileNotFoundError:
-            # TODO: remove case when files have been generated
-            logger.warning(
-                f"Resorting to default CameraReadout,"
-                f" File does not exist: ({tabname})"
-            )
-            reference_pulse_shape = np.array([norm.pdf(np.arange(96), 48, 6)])
-            return cls(
-                camera_name=camera_name,
-                sampling_rate=u.Quantity(1, u.GHz),
-                reference_pulse_shape=reference_pulse_shape,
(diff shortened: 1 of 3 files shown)


ctapipe/instrument/camera/readout.py
```diff
@@ -111,25 +111,11 @@ def from_name(cls, camera_name="NectarCam", version=None):
...
-            )
-            reference_pulse_shape = np.array([norm.pdf(np.arange(96), 48, 6)])
-            return cls(
-                camera_name=camera_name,
-                sampling_rate=u.Quantity(1, u.GHz),
-                reference_pulse_shape=reference_pulse_shape,
-                reference_pulse_sample_width=u.Quantity(1, u.ns),
-            )
+        tabname = "{camera_name}{verstr}.camreadout".format(
+            camera_name=camera_name, verstr=verstr
```
nbiederbeck, line 115: maybe use f-string syntax?

ctapipe/instrument/camera/tests/test_readout.py
```diff
@@ -145,5 +145,8 @@ def test_hashing():
 @pytest.mark.parametrize("camera_name", camera_names)
 def test_camera_from_name(camera_name):
     """ check we can construct all cameras from name"""
-    camera = CameraReadout.from_name(camera_name)
-    assert str(camera) == camera_name
+
+    # these two don't have readout definitions on the dataserver
+    if camera_name not in ["MAGICCam", "Whipple109"]:
```
nbiederbeck, line 150: should we test one of those? I think we only test LSTCam (at least on my machine), which will then not really test the new behaviour introduced in this PR

ctapipe/instrument/camera/readout.py
maxnoe: I didn't add/change this code, just removed a try except around. There is no real reason to remove format

ctapipe/instrument/camera/tests/test_readout.py
maxnoe: done

nbiederbeck (approved):

LukasNickel (approved):

PR #1705, opened 2021-05-07
Update documentation Makefile

HealthyPear:
Use the same building options as in the CI

### DIFF
--- a/docs/Makefile
+++ b/docs/Makefile
@@ -2,7 +2,7 @@
 #
 
 # You can set these variables from the command line.
-SPHINXOPTS    =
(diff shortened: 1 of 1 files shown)


docs/Makefile
```diff
@@ -2,7 +2,7 @@
 #
 
 # You can set these variables from the command line.
-SPHINXOPTS    =
+SPHINXOPTS    = -W --keep-going -n --color -w /tmp/sphinx-log
```
maxnoe, line 5: I think the logfile shouldn't be here. That's only needed on the CI system

HealthyPear: done

maxnoe (approved):

nbiederbeck (approved):

PR #1718, opened 2021-05-12
Add macos-latest to CI workflow jobs

### LINKED ISSUE #1649: Enable mac CI
Github Actions supports macOS CI builds.

We should activate those again.

HealthyPear:
Closes #1649

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -9,9 +9,9 @@ env:
 
 jobs:
   tests:
-    runs-on: [ubuntu-latest, macos-latest]
     strategy:
       matrix:
+        runs-on: [ubuntu-latest, macos-latest]
         python-version: [3.7, 3.8]
         install-method: ["pip", "conda"]
 
@@ -50,7 +50,10 @@ jobs:
       - uses: codecov/codecov-action@v1
 
   docs:
-    runs-on: [ubuntu-latest, macos-latest]
+    strategy:
+      matrix:
+        runs-on: [ubuntu-latest, macos-latest]
+        python-version: [3.7, 3.8]
     steps:
       - uses: actions/checkout@v2
         with:
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -9,9 +9,10 @@ env:
 
 jobs:
   tests:
+    runs-on: ${{ matrix.os }}
     strategy:
       matrix:
-        runs-on: [ubuntu-latest, macos-latest]
+        os: [ubuntu-latest, macos-latest]
         python-version: [3.7, 3.8]
         install-method: ["pip", "conda"]
 
@@ -50,9 +51,10 @@ jobs:
       - uses: codecov/codecov-action@v1
 
   docs:
+    runs-on: ${{ matrix.os }}
     strategy:
       matrix:
-        runs-on: [ubuntu-latest, macos-latest]
+        os: [ubuntu-latest, macos-latest]
         python-version: [3.7, 3.8]
     steps:
       - uses: actions/checkout@v2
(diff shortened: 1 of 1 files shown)


maxnoe:
@HealthyPear I think this is invalid syntax. You need to move the `runs_on` into the `matrix`

kosack (dismissed):

maxnoe:
See a working example here: https://github.com/tudo-astroparticlephysics/PROPOSAL/blob/master/.github/workflows/cpp.yml

HealthyPear:
> See a working example here: https://github.com/tudo-astroparticlephysics/PROPOSAL/blob/master/.github/workflows/cpp.yml

I took it from [https://gist.github.com/Jarmos-san/a1f219934fd23ad6b915a37dd85a2864](url) (test linter)
Apart from having the `runs-on: ${{ matrix.os }}` line after the matrix I don't see differences...

HealthyPear:
The tests seem to fail for this reason

```
Using conda

Collecting package metadata (current_repodata.json): ...working... failed
NotWritableError: The current user does not have write permissions to a required path.
  path: /usr/local/miniconda/pkgs/cache/53d7ef51.json
  uid: 501
  gid: 20

If you feel that permissions on this path are set incorrectly, you can manually
change them by executing

  $ sudo chown 501:20 /usr/local/miniconda/pkgs/cache/53d7ef51.json

In general, it's not advisable to use 'sudo conda'.


Error: Process completed with exit code 1.
```

kosack:
miniconda doesn't need to be installed into /usr/local - is it possible to just install it in $HOME?  That way no sudo is needed

maxnoe:
@kosack it's possible if we install miniconda ourselves, sure. Currently we rely on the setup-python action to install the initial miniconda

HealthyPear:
I found someone else with the same problem, but didn't really understand what is the solution
https://github.com/actions/virtual-environments/issues/1003

We could also use [this GitHub action](https://github.com/marketplace/actions/setup-miniconda) to setup miniconda

I think that it should be sufficient to replace

```yaml
- name: Set up Python ${{ matrix.python-version }}
   uses: actions/setup-python@v2
   with:
          python-version: ${{ matrix.python-version }}
```

with

```yaml
- name: Set up Python ${{ matrix.python-version }}
  uses: conda-incubator/setup-miniconda@v2
   with:
          auto-update-conda: true
          python-version: ${{ matrix.python-version }}
```

maxnoe:
@HealthyPear yes, you can even add options for the environment file and which environment to activate.

This will reduce our conda boilerplate for the conda builds.

HealthyPear:
It finally seems to work!

Though I think the checks still expected will block the merging even with 2 approving reviews since they had a different name (due to the one degree of freedom less)

They should have been substituted by the new ones but they are still there

.github/install.sh
```diff
@@ -2,7 +2,8 @@
 
 
 if [[ "$INSTALL_METHOD" == "conda" ]]; then
-  echo "Using conda"
+  echo "Using conda located at "
+  echo $CONDA
```
HealthyPear, line 6: I added this during some tests, but I think it should remain for debugging

kosack (dismissed):

.github/workflows/ci.yml
```diff
@@ -27,10 +31,10 @@ jobs:
             ~/.cache/ctapipe
           key: ctapipe-test-data
 
-      - name: Set up Python ${{ matrix.python-version }}
-        uses: actions/setup-python@v2
+      - uses: conda-incubator/setup-miniconda@v2
         with:
-          python-version: ${{ matrix.python-version }}
+          auto-activate-base: true
+          activate-environment: ""
```
maxnoe, line 37: Why is this the empty string?

maxnoe (approved):

kosack (approved):

PR #1722, opened 2021-05-18
Use .ecsv extension in dump instrument to enable `Table.read`.

maxnoe:
When using `.ecsv`, astropy table automatically chooses the `ascii.ecsv` format, also when reading.

By using `.ecsv.txt` we prevent simply usung `Table.read(path)`


kosack:
The reason for adding .txt was just so that when clicking the file in Linux/macOS/Windows, a text editor or viewer is able to  opened it (rather than nothing, since ecsv is not a well known format is not registered by default). It also makes QuickLook (in macOS) or Sushi (gnome linux) work, so you can preview the file in a file manager.

maxnoe:
MMh, on my system that's not an issue with the ecsv extension:
```
❯ file MonteCarloArray.optics.ecsv
MonteCarloArray.optics.ecsv: ASCII text
```

And `xdg-open  MonteCarloArray.optics.ecsv`  opens the file in my default text editor.

kosack:
Yeah, seems to work now, so I'm ok with the change.

kosack (approved):

LukasNickel (approved):

kosack (approved):

Issue #1707, opened 2021-05-07
Compoment to worsen the optical PSF at DL1a level


maxnoe:
Same reason as #1706

kosack:
In general it may be nice to have several components (or one big one) that can modify events to worsen them for various studies, e.g where we can introduce random noise into the pixel/waveform values, images (including neighbor pixels effects like cross-talk), image parameters, and reconstruction parameters. ]

This would benefit from some simple coordinated design so that all work in a similar way from a user/config standpoint. I guess for each value type, you would want to specify the random distribution that is applied, for example (Gaussian, Poisson, flat, and their parameters).  This will be useful for e.g. making bracketing IRFs as well.

kosack:
Another option is to not make these Components, but just make some tools that modify DL1 (and soon DL2) files to introduce artifacts, and then no change to the pixel loop is needed.

maxnoe:
Yes, we can have the components and then a tool that modifies an existing DL1 file using them. 

This was about smearing the DL1a images, so directly worsen the optical PSF, which of course could have an impact on the gamma-ray PSF.

Issue #1706, opened 2021-05-07
Component to add noise to images


maxnoe:
For adapting generic simulations to observation conditions, it might be necessary to add noise to images during the dl1 stage.

This was just implemented in lstchain here https://github.com/cta-observatory/cta-lstchain/pull/708/files and seems to greatly improve the match between observed and simulated data.

kosack:
This is what is done in Whipple for many years - it's called "noise padding" and is definitely a good idea to do for mono analysis (for stereo it has less effect). Being able to specify a noise value per pixel would be useful (i.e. for a pair of On/OFF observations, it's useful to increase the noise to the highest of the pedestal variances of each of the two runs in the pair)

kosack:
Such a component would also be useful for dealing with broken/noisy pixels, as it lets you inject a factor of 0 to simulate turning off a pixel, and a high value to simulate high NSB from a star, etc

HealthyPear:
the implementation inside the component could use the already available `ctapipe.image.toymodel.ImageModel`

maxnoe:
@HealthyPear I don't think this has much in common, besides a call to `rng.poisson` or `rng.normal`

watsonjj:
I'd like to ask a question with regards to this and #1707.

Would these features replace fully representing the noise and PSF in the MC model? Or are these features intended for exploratory comparisons between simulations and data, such that the MC model can then be appropriately improved?

The current wording in this issue could suggest that this is intended more as a permanently used feature in the pipeline to improve the matching, rather than an exploratory feature, and could lead to MC models that aren't fully representative of the real telescopes.

maxnoe:
@watsonjj No, at least for LST, this is only to transform the standard simulations to runs that deviate from the norm due to worse mirror alignment or higher NSB noise.

In FACT we tried simulating only the Cherenkov shower signal and add that to observed pedestals, which could be another possible implementation.

But for now, these simple components would just add additional noise / be a smearing convolution. What you do with it, is the users' choice. You could simulate without any noise and add all required noise using this or simulate a perfect mirror and add all PSF (up to a certain degree) witht he psf smearing component, but that's certainly not the intended use case.

moralejo:
[...]

PR #1723, opened 2021-05-19
Components to modify dl1 images

LukasNickel:
As recently discussed (See #1706, #1707 and linked lstchain PR), modifying simulated dl1 images is a useful way to improve data/mc mismatches.
There are different ways to achieve this. I personally feel like it is easiest done in stage-1, although that requires to modify the event loop. Currently I added a `ImageModifier` component, which forced me to also add a `NullModifier`, that does not alter the image. There might be better ways to achieve this in the given traitlets system.
While I feel like this is somewhat hacky, we might want to use this to mask pixels or do other things in real data, so that this would not be used all the time in the end anyway (?).
The advantage of implementing it this way to me is that we already have a tool, that can calculate dl1 from input dl1 files, which is also used for recleaning anyway and also takes care of recalculating parameters. This makes it easy to use from a users perspective and less duplicated effort.
This is certainly a point to be discussed though.

As there are countless ways to modify images and being able to reproduce lstchain results is a plus anyway, I for now implemented the approach @moralejo used in lstchain as a ready to use component. This performs the smearing aswell as the addition of noise.

-------------
- Opinions on using stage-1? Is there any benefit in a new tool?
- Are there better ways than using a dummy modifier? Does it matter?
- It should be easy to perform this on all images at once, just like the cleaning. I think its not worth to do this here though, especially since this would be part of the image processing loop, that could be changed altogether.

-------------
Proof of concept: I processed some simtel events and reprocessed the dl1 file with stage-1, using the `LSTImageModifier`.
noisy_image

sample_charges

### DIFF
--- a/ctapipe/image/modifications.py
+++ b/ctapipe/image/modifications.py
@@ -6,50 +6,55 @@ from ..core.component import TelescopeComponent
 
 
 def mask_broken(geom, pix_ids):
+    """
+    Create a boolean mask with deselected pixels
+    Does this really need a function?
+    """
     mask = np.ones(shape=geom.pix_id.shape, dtype=bool)
     mask[pix_ids] = 0
     return mask
 
 
-def mask_thresholds(image, upper, lower):
-    mask = np.ones(shape=geom.pix_id.shape, dtype=bool)
-    mask[image > upper] = 0
-    mask[image < lower] = 0
+def mask_thresholds(image, upper=None, lower=None):
+    """
+    Create a boolean mask with pixels selected based on thresholds.
+    Does this really need a function?
+    """
+    mask = np.ones_like(image, dtype=bool)
+    if upper:
+        mask[image > upper] = 0
+    if lower:
+        mask[image < lower] = 0
     return mask
 
 
-def add_noise(
-    image,
-    noise_level,
-    rng=None,
-    mask=None,
-    upper_threshold=None,
-    lower_threshold=None,
-    correct_bias=True,
-):
+def add_noise(image, noise_level, rng=None, correct_bias=True):
+    """
+    Create a new image with added poissonian noise
+    """
     # notes:
     # mask and thresholds is probably not very useful
     # mask is only useful if noise level is a scalar, could also provide the array directly
     # maybe only noise level and bias, selection in tool?
     if not rng:
         rng = np.random.default_rng()
-    if mask is None:
-        mask = np.ones(shape=image.shape, dtype=int)
-    else:
-        mask = mask.astype(int)
-    if upper_threshold:
-        mask &= image < upper_threshold
-    if lower_threshold:
-        mask &= image > lower_threshold
-    mask *= noise_level
     noisy_image = image.copy()
-    noisy_image += rng.poisson(mask)
+    noise = rng.poisson(noise_level)
+    noisy_image += noise
     if correct_bias:
-        noisy_image -= mask
+        # which one to use?
+        # Fluctuations that have no bias per pixel
+        noisy_image -= noise_level
+        # No bias over the whole image
+        # noisy_image -= np.mean(noise)
     return noisy_image
 
 
 def smear_image(image, geom, smear_factor):
+    """
+    Create a new image with values smeared to the direct pixel neighbors
+    Pixels at the camera edge lose charge this way.
+    """
     # make clear what smear factor is supposed to mean and that only direct neighbors are taken into account
     # smear factor can be an array as well -> selection in tool possible!
     # a more sophisticated approach might make use of the pixel area, but thats complicated
--- a/ctapipe/image/modifications.py
+++ b/ctapipe/image/modifications.py
@@ -1,0 +1,99 @@
+from abc import abstractmethod
+import numpy as np
+from ..core.component import TelescopeComponent
+
+# the most important part is the selection api!!
+
+
+def mask_broken(geom, pix_ids):
+    mask = np.ones(shape=geom.pix_id.shape, dtype=bool)
+    mask[pix_ids] = 0
+    return mask
+
+
+def mask_thresholds(image, upper, lower):
+    mask = np.ones(shape=geom.pix_id.shape, dtype=bool)
+    mask[image > upper] = 0
+    mask[image < lower] = 0
+    return mask
+
+
+def add_noise(
+    image,
+    noise_level,
+    rng=None,
+    mask=None,
+    upper_threshold=None,
+    lower_threshold=None,
+    correct_bias=True,
+):
+    # notes:
+    # mask and thresholds is probably not very useful
+    # mask is only useful if noise level is a scalar, could also provide the array directly
+    # maybe only noise level and bias, selection in tool?
+    if not rng:
+        rng = np.random.default_rng()
+    if mask is None:
+        mask = np.ones(shape=image.shape, dtype=int)
+    else:
+        mask = mask.astype(int)
+    if upper_threshold:
+        mask &= image < upper_threshold
+    if lower_threshold:
+        mask &= image > lower_threshold
+    mask *= noise_level
+    noisy_image = image.copy()
+    noisy_image += rng.poisson(mask)
+    if correct_bias:
+        noisy_image -= mask
+    return noisy_image
+
+
+def smear_image(image, geom, smear_factor):
+    # make clear what smear factor is supposed to mean and that only direct neighbors are taken into account
+    # smear factor can be an array as well -> selection in tool possible!
+    # a more sophisticated approach might make use of the pixel area, but thats complicated
+    if geom.pix_type.value == "hexagon":
+        max_neighbors = 6
+    elif geom.pix_type.value == "rectangular":  # thats probably labeld differently
+        max_neighbors = 8  # or 4? lookup neighbor matrix for a rect cam
+    else:
+        print(geom.pix_type)
+        raise Exception("Unknown pixel type")
+
+    diffused_image = (
+        (image * geom.neighbor_matrix).sum(axis=1) * smear_factor / max_neighbors
+    )
+    remaining_image = image * (1 - smear_factor)
+    smeared_image = remaining_image + diffused_image
+    return smeared_image
+
+
+class ImageModifier(TelescopeComponent):
+    """
+    ...
+    """
+
+    @abstractmethod
+    def __call__(self, tel_id: int, image: np.ndarray, **kwargs) -> np.ndarray:
+        """
+        Modifies a given image
+
+        Returns
+        -------
+        np.ndarray
+            modified image
+        """
+        pass
+
+
+class ImageSmearer(ImageModifier):
+    pass
+
+
+class NoiseAdder(ImageModifier):
+    pass
+
+
+class PixelValueSetter(ImageModifier):
+    pass
--- a/ctapipe/image/modifications.py
+++ b/ctapipe/image/modifications.py
@@ -3,31 +3,6 @@ import numpy as np
 from ..core.component import TelescopeComponent
 from ..core.traits import FloatTelescopeParameter, BoolTelescopeParameter
 
-# the most important part is the selection api!!
-
-
-def mask_broken(geom, pix_ids):
-    """
-    Create a boolean mask with deselected pixels
-    Does this really need a function?
-    """
-    mask = np.ones(shape=geom.pix_id.shape, dtype=bool)
-    mask[pix_ids] = 0
-    return mask
-
-
-def mask_thresholds(image, upper=None, lower=None):
-    """
-    Create a boolean mask with pixels selected based on thresholds.
-    Does this really need a function?
-    """
-    mask = np.ones_like(image, dtype=bool)
-    if upper:
-        mask[image > upper] = 0
-    if lower:
-        mask[image < lower] = 0
-    return mask
-
 
 def add_noise(image, noise_level, rng=None, correct_bias=True):
     """
@@ -39,11 +14,7 @@ def add_noise(image, noise_level, rng=None, correct_bias=True):
     noise = rng.poisson(noise_level)
     noisy_image += noise
     if correct_bias:
-        # which one to use?
-        # Fluctuations that have no bias per pixel
         noisy_image -= noise_level
-        # No bias over the whole image
-        # noisy_image -= np.mean(noise)
     return noisy_image
 
 
--- a/ctapipe/image/modifications.py
+++ b/ctapipe/image/modifications.py
@@ -33,10 +33,6 @@ def add_noise(image, noise_level, rng=None, correct_bias=True):
     """
     Create a new image with added poissonian noise
     """
-    # notes:
-    # mask and thresholds is probably not very useful
-    # mask is only useful if noise level is a scalar, could also provide the array directly
-    # maybe only noise level and bias, selection in tool?
     if not rng:
         rng = np.random.default_rng()
     noisy_image = image.copy()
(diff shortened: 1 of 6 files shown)


HealthyPear:
Wouldn't be easier to generalize `ctapipe.image.toymodel.ImageModel` and/or what's in that sub-module instead of adding new functions?

LukasNickel:
I think having basic functions is better tbh. It's easier to test this way and if we want to avoid duplicate code we could add this to the toymodel (which I havent done yet).
And in my understanding the toymodel is not meant as a tool to use in analyses, but more of a visualization/testing thing. Correct me if I am wrong.

HealthyPear:
To be clear I am not criticizing what has been added here, just wondering if it makes sense how much code makes sense to replicate wrt what we already have

HealthyPear:
> > I think having basic functions is better tbh. It's easier to test this way and if we want to avoid duplicate code we could add this to the toymodel (which I havent done yet).

`add_noise` is basically the same as `ImageModel.generate_image` so you could just use/modify/generalize that no?

 I would say it' better to have all these functions in the same place and let the `ImageModifier` pick what it needs

> And in my understanding the toymodel is not meant as a tool to use in analyses, but more of a visualization/testing thing. Correct me if I am wrong.

There the signal is created randomly, but nothing prevents you (us) to add a keyword argument `image=None`which if initialized with dl1.image uses that instead of creating the randomized toymodel image

maxnoe:
@HealthyPear As noted in the connected issue, I think using the `ImageModel` in `ModifyImage` is having it backwards.

It should be the other way around. The `ImageModel` should use the `ImageModifier` to add noise / smear the image.

ctapipe/image/modifications.py
```diff
@@ -0,0 +1,113 @@
...
+    Create a new image with values smeared to the direct pixel neighbors
+    Pixels at the camera edge lose charge this way.
+    """
+    # make clear what smear factor is supposed to mean and that only direct neighbors are taken into account
+    # smear factor can be an array as well -> selection in tool possible!
+    # a more sophisticated approach might make use of the pixel area, but thats complicated
+    if geom.pix_type.value == "hexagon":
+        max_neighbors = 6
+    elif geom.pix_type.value == "rectangular":  # thats probably labeld differently
+        max_neighbors = 8  # or 4? lookup neighbor matrix for a rect cam
```
maxnoe, line 32: I think we should use a gaussian kernel here. E.g. not use the same amount for the direct neighbors as for the diagonal neighbors. I don't know how easy that is though.

ctapipe/image/modifications.py
```diff
@@ -0,0 +1,113 @@
...
+    else:
+        print(geom.pix_type)
```
maxnoe, line 34: Don't print, put it into the excpetion.

ctapipe/image/modifications.py
```diff
@@ -0,0 +1,113 @@
...
+
+def smear_image(image, geom, smear_factor):
+    """
+    Create a new image with values smeared to the direct pixel neighbors
+    Pixels at the camera edge lose charge this way.
+    """
+    # make clear what smear factor is supposed to mean and that only direct neighbors are taken into account
+    # smear factor can be an array as well -> selection in tool possible!
+    # a more sophisticated approach might make use of the pixel area, but thats complicated
+    if geom.pix_type.value == "hexagon":
```
maxnoe, line 29: Please use the enum. `geom.pix_type is PixelShape.HEXAGONAL`

ctapipe/image/modifications.py
```diff
@@ -0,0 +1,113 @@
...
+        """
+        pass
+
+
+class NullModifier(ImageModifier):
+    def __call__(self, tel_id: int, image: np.ndarray) -> np.ndarray:
+        return image
+
+
+class LSTImageModifier(ImageModifier):
```
maxnoe, line 69: This is not really specific to LST I think.

ctapipe/image/modifications.py
```diff
@@ -0,0 +1,113 @@
...
+        default_value=8, help="separation between dim and bright pixels"
+    ).tag(config=True)
+    dim_pixel_bias = FloatTelescopeParameter(
+        default_value=0.6, help="extra bias to add in dim pixels"
+    ).tag(config=True)
+    dim_pixel_noise = FloatTelescopeParameter(
+        default_value=1.5, help="expected extra noise in dim pixels"
+    ).tag(config=True)
+    bright_pixel_noise = FloatTelescopeParameter(
+        default_value=1.44, help="expected extra noise in bright pixels"
```
maxnoe, line 87: Why this strange default? 

I would set the defaults all to 0 / 1, so that by default, the ImageModifier is a no-op

ctapipe/image/modifications.py
LukasNickel: I will have a look. It seemed to be a bit tricky, but I havent looked very closely

ctapipe/image/modifications.py
LukasNickel: Copied from the lstchain pr. But I'm fine with that

kosack:
I have a few design questions/comments: 
-  How is this expected to be included in the main analysis (e.g. ctapipe-stage1): a single step in the chain?
-  Is the idea that there can be one and only one ImageModifier that can be chosen by name? what if the user wants to apply two or more modifications?
-  would it be be better to drop the hierarchy and have one big one that has many options (e.g. to be able to smear and add noise at the same time in a generic way?) Or do you expect that each chain would build their own subclass and enable that? (meaning code has to be written to support fancier combinations of modifications?) 

For the question of where to put it: I think it should go in `ctapipe-stage1`, but *also* be a separate tool. That way one can use it either way during analysis, or on an existing DL1 file to generate a new modified file.   That bring up the question of if it should act event-wise or on all events at once (which would be more efficient in the second case)

maxnoe:
>For the question of where to put it: I think it should go in ctapipe-stage1, but also be a separate tool. That way one can use it either way during analysis, or on an existing DL1 file to generate a new modified file.

For that you don't need a separate tool, you just rerun stage1 using the dl1 source.

LukasNickel:
>     * How is this expected to be included in the main analysis (e.g. ctapipe-stage1): a single step in the chain? 

Thats my take. What alternatives do you see?

>     * Is the idea that there can be one and only one ImageModifier that can be chosen by name? what if the user wants to apply two or more modifications? 
>     * would it be be better to drop the hierarchy and have one big one that has many options (e.g. to be able to smear and add noise at the same time in a generic way?) Or do you expect that each chain would build their own subclass and enable that? (meaning code has to be written to support fancier combinations of modifications?)

This is also one of my main concerns. Unfortunately I dont really see a way out. Building one tool with lots of options did not strike me as more elegant though. If you or someone else sees a way out, I would be very interested. Afaik there is no easy way to enable multiple (or an a priori undefined number of) components of the same type in a tool, so we would need a monolithic giant at some point. We could also have one component for each task (adding noise, blurring the light, masking pixels, ...). I settled with one component, but I dont have a hard opinion.

> That brings up the question of if it should act event-wise or on all events at once (which would be more efficient in the second case)

Agreed. I feel like it could be done on all events just like the cleaning could. But since you would need to reclean and recalculate parameters anyway, I figured its not worth the effort to do it independently.

ctapipe/image/modifications.py
LukasNickel: Not really, its just the current lstchain procedure.
I was not sure if we wanted to mirror this behaviour 1to1, so I marked in in the name for now.

ctapipe/image/modifications.py
LukasNickel: So direct neighbors only, but there is a constant factor between diagonal and non-diagonal pixels?
That should be easy. What would be tricky, would be to incorporate the pixel size and also smear to pixels beyond the first neighbor.

LukasNickel:
> > For that you don't need a separate tool, you just rerun stage1 using the dl1 source
> 
> Good point. It's not super efficient, but should work.

It does work and is also done for recleaning dl1 files without modifying the images.
But having the event loop is of course not optimal in terms of runtime.

kosack:
> It does work and is also done for recleaning dl1 files without modifying the images.
But having the event loop is of course not optimal in terms of runtime.

Yeah, let's ignore speed issues for now - that can be a future refactoring of ctapipe in general. The design now is also fine, having a single component with subclasses like the ImageExtractors.  It's not super flexible, but that's ok for now, and I can't think of a better way : even with one big monster class, the order of operations is hard to define (e.g. blur then add noise is not the same thing as add noise then blur).

ctapipe/image/modifications.py
kosack: Just curious: why is this used in the LST chain? Is it basically another calibration to remove some strange effects of the ring sampler?  Is that a calibration we should include somewhere earlier, e.g. where do the values get derived?

maxnoe: Nothing to do with the ring sampler. It's just to make the simulations simulated for specific conditions (extragalactic dark night) fit better for other kinds of observations (e.g. galactiv fov for crab) and a worse PSF than simulated.

maxnoe: @moralejo showed nice results of how these two additions improve the data mc mismatches quite dramatically.

maxnoe: See discussion here: https://github.com/cta-observatory/cta-lstchain/pull/708

kosack: ah, so it's to add NSB essentially - yes hat makes sense (though I don't quite understand the smearing or part or why dim pixels respond differently).  Would  be interesting to test how different adding the noise here is from adding it at the simulation level - that could save some time in prod5 to not simulate different NSB levels if the effect is the same

kosack: It does sounds similar to what we used to do in Whipple: add noise to ON/OFF observations such that you bring the total NSB noise up to the level of the worst of the pair.

moralejo: > > though I don't quite understand the smearing or part or why dim pixels respond differentl
> 
> I think abelardos reasoning for this was that in noise dominated pixels there is a positive bias for peak-finding integrators, since they will look for a high peak, even if there is only noise. So noise is higher in pixels where you integrate over the highest fluctuation as opposed to just integrating over the pulse with the normal noise contribution. This is also the reason why you can give an additional `dim_bias`.

Yes, that is the idea: it is not possible to do a 100% correct tuning at the DL1a level if a peak-search algorithm was used. Only the limit case of bright signals, for which the integration window position won't change upon adding some extra noise, is easy to treat (I consider "bright signals" those above the picture threshold).  What I did in lstchain is an empirical tuning of the noise-only spectrum for interleaved pedestals, which is what I apply for pixels below the picture threshold. 

Bear in mind that the idea is to use this for fine-tuning, it should not replace having MC generated with different levels of NSB (either at the sim_telarray level, or by adding noise in R1 waveforms)

maxnoe:
@LukasNickel  did you see the updated function for the PSF here: https://github.com/cta-observatory/cta-lstchain/pull/719 ?

The first version had some problems discussed in that PR

LukasNickel:
> @LukasNickel did you see the updated function for the PSF here: [cta-observatory/cta-lstchain#719](https://github.com/cta-observatory/cta-lstchain/pull/719) ?
> 
> The first version had some problems discussed in that PR

I did, but would have probably forgotten to incorporate it, thanks

LukasNickel:
Relying on #1776.
Otherwise this doesnt work on dl1 files, that have parameters and images

kosack (approved):

ctapipe/image/modifications.py
```diff
@@ -0,0 +1,231 @@
...
+        Modifies a given image
+
+        Returns
+        -------
+        np.ndarray
+            modified image
+        """
+
+
+class NullModifier(ImageModifier):
```
maxnoe, line 161: Do we really need a "NullModifier"? Can't we just set the traitlet to `None` by default and do:

```
if self.image_transformer is not None:
    image = self.image_transformer(image)
```

Also: we might want to apply more than one, so is there an easy possibility to have multiple? Like

```
{
    "image_transformers": [
        {"cls": "NoiseAdder", ...},
        {"cls"}: "PSFSmearer", ...}
    ]
}
```


just thinking a bit

LukasNickel: 1)
Probably not since we made the noise adder a noop by default anyway.

2)
I like it, but didn't we come to the conclusion that its not really feasible with the config/traitlets system currently? It would make a lot of sense if the order is fixed and classes can be used multiple times. Then you would be really flexible with your modifications. I don't remember exactly what the issue was though. Would this work on the command line?

kosack: That is basically what is done with TelescopeParameters, but in this case, the problem is how to configure different instances of the same Component: the config system doesn't allow that currently.  Could be a future feature if we change config systems though!

LukasNickel: > the problem is how to configure different instances of the same Component
Thanks, that was probably it.

In that case I would propose to ignore the proposal for now because it leads to restrictions that are annoying and not super transparent to the user (mainly because the config files are not strictly checked before executing tools).
But your opinions may vary

ctapipe/image/modifications.py
```diff
@@ -0,0 +1,231 @@
...
+        neighbor_charges = np.random.multinomial(to_smear, smear_probabilities)
+
+        for n in range(n_neighbors):
+            neighbor = neighbors[n]
+            new_image[neighbor] += neighbor_charges[n]
+
+    return new_image
+
+
+def smear_psf_statically(image, geom, smear_factor):
```
maxnoe, line 98: Since we know that this did not yield usable results, please just remove.

ctapipe/image/modifications.py
maxnoe, line 161: Please remove the "NullModifier"

maxnoe (changes requested):
Code itself looks ok, some comments:

* Please remove the first, not-working psf smearing version.
* At the moment, this does not allow to use both NSB addition and PSF smearing at the same time, however, this should be possible. So I'd just add both explictly to the process tool, not configured through a single base class. By default, they should do nothing, e.g. have psf smearing and added noise at 0 and then be not costly to call.

LukasNickel:
Matplotlib errors in seemingly unrelated code only in the push test?
Any idea where that is coming from? I already reran the tests...

maxnoe:
@LukasNickel You need to merge or rebase master, the issue is fixed there. Since the pull request build runs on the merged version, it passes.

ctapipe/image/modifications.py
```diff
@@ -0,0 +1,174 @@
...
+    correct_bias = BoolTelescopeParameter(
+        default_value=True, help="If True subtract the expected noise from the image."
+    ).tag(config=True)
+
+    def __call__(self, tel_id, image, rng=None):
+        # TODO: how to get max_neighbors?
+        # This might be the safest way (including the decision if diagonal neighbors
+        # count for square pixels)
+        geom = self.subarray.tel[tel_id].camera.geometry
+        max_neighbors = geom.neighbor_matrix.sum(axis=1).max()
```
maxnoe, line 152: This should be a property on the geometry and cached, not calculated for each pixel on each event.

ctapipe/image/modifications.py
```diff
@@ -0,0 +1,174 @@
...
+        Modifies a given image
+
+        Returns
+        -------
+        np.ndarray
+            modified image
+        """
+
+
+class NSBNoiseAdder(ImageModifier):
```
maxnoe, line 113: The class should be called differently now. Probably this should just be `ImageModifier`.

LukasNickel: No base class anymore?

maxnoe: Yes.

ctapipe/instrument/camera/geometry.py
```diff
@@ -640,6 +640,10 @@ def neighbors(self):
     def neighbor_matrix(self):
         return self.neighbor_matrix_sparse.A
 
+    @lazyproperty
+    def max_neighbors(self):
+        return self.neighbor_matrix.sum(axis=1).max()
```
maxnoe, line 645: Using the sparse matrix should be better here

ctapipe/image/modifications.py
```diff
@@ -0,0 +1,151 @@
...
+    would change for peak-searching extraction algorithms with different background levels
+    introducing a bias to the charge in dim pixels.
+
+    The performed steps include:
+    - Smearing of the image (simulating a worse PSF)
+    - Adding poissonian noise (different for bright and dim pixels)
+    - Adding a bias to dim pixel charges (see above)
+    """
+
+    smear_factor = FloatTelescopeParameter(
```
maxnoe, line 109: Call this `psf_smear_factor`?

ctapipe/image/modifications.py
```diff
@@ -0,0 +1,151 @@
...
+        default_value=0.0, help="Fraction of light to move to each neighbor"
+    ).tag(config=True)
+    transition_charge = FloatTelescopeParameter(
```
maxnoe, line 112: prefix noise related optoins with `noise_` ?

maxnoe:
@LukasNickel could you implement the random seed stuff we talked about?

I.e. add a `seed` configurable to the `Component` and generate seeds from that for the numba code?

ctapipe/image/modifications.py
```diff
@@ -0,0 +1,181 @@
...
+
+
+def add_noise(image, noise_level, rng=None, correct_bias=True):
+    """
+    Create a new image with added poissonian noise
+    """
+    if not rng:
+        rng = np.random.default_rng()
+    noisy_image = image.copy()
+    noise = rng.poisson(noise_level)
```
maxnoe, line 15: If `noise_level` is a single number (it isn't when used thorugh the Component, but could be if something uses it on its own) each pixel get's the same value added.

ctapipe/image/modifications.py
```diff
@@ -0,0 +1,181 @@
+import numpy as np
+from numba import njit
+from ..core.component import TelescopeComponent
+from ..core.traits import FloatTelescopeParameter, BoolTelescopeParameter, Int
+from ..instrument import SubarrayDescription
+
+
+def add_noise(image, noise_level, rng=None, correct_bias=True):
```
maxnoe, line 8: I'd make this private `_add_noise`

ctapipe/image/modifications.py
```diff
@@ -0,0 +1,181 @@
...
+    noisy_image = image.copy()
+    noise = rng.poisson(noise_level)
+    noisy_image += noise
+    if correct_bias:
+        noisy_image -= noise_level
+    return noisy_image
+
+
+@njit(cache=True)
+def smear_psf_randomly(
```
maxnoe, line 23: I'd make this private as well

maxnoe (changes requested):
You should also add this to the documentation in `docs/ctapipe_api/image/`, look at e.g. `docs/ctapipe_api/image/hillas.rst`

ctapipe/image/modifications.py
maxnoe: done

maxnoe (approved):

ctapipe/image/modifications.py
```diff
@@ -0,0 +1,189 @@
...
+        # remove light from current pixel
+        new_image[pixel] -= to_smear
+
+        # add light to neighbor pixels
+        neighbors = indices[indptr[pixel] : indptr[pixel + 1]]
+        n_neighbors = len(neighbors)
+
+        # all neighbors are equally likely to receive the charge
+        # we always distribute the charge into 6 neighbors, so that charge
+        # on the edges of the camera is lost
```
nbiederbeck, line 93: Is this comment valid? don't we smear into `len(smear_probabilities)` neighbors?

LukasNickel: They do, thats why you lose something
So you split the charge into 6 (for hexagonal pixels) parts, but only add it to existing neighbors.
Say you draw
```
[1,1,3,1,2,3]
```
and have 4 neighbors (because you are at the edge of the camera). Then this
https://github.com/cta-observatory/ctapipe/blob/63ea86ee52ecfd8969c124697d536854dc02b8b7/ctapipe/image/modifications.py#L96-L98

 stops after 4 iterations and you lose
`2+3 = 5` charge

nbiederbeck: ok, got it

nbiederbeck (approved):

nbiederbeck (approved):

maxnoe (approved):

maxnoe (dismissed):

nbiederbeck (dismissed):

nbiederbeck (approved):

maxnoe (approved):

Issue #1632, opened 2021-02-22
Support Prod5 subarray selection


kosack:
It might be nice to provide some functionality be able to create a subarray from a general SubarrayDescription using the info provided here (for example) for Prod5:

https://gitlab.cta-observatory.org/GernotMaier/CTA-Prod5-PerformanceResults/-/blob/master/Layouts/telescope_lists_south/CTA.prod5S-M1C1-10MSTs30SSTs-MSTF.lis

Something like:

```py3
tel_ids, names = read_subarray_info("CTA.prod5S-M1C1-10MSTs30SSTs-MSTF.lis")
subarray = source.subarray.select_subarray("prod5S-M1C1-10MSTs30SSTs", tel_ids)
```
 
These files also provide human-readable names for the telescopes (e.g. "M-08"), so perhaps adding an attribute for that in the SubarrayDescription could also be useful, and allowing one to name them. Note that that is not the same thing as `str(event.tel[14])`, which returns the name of the TelescopeDescription (e.g. `SST_ASTRI_CHEC`), not the name of the individual telescope

```py3
subarray.set_telescope_names( dict(zip(tel_ids, names)) )
```

That could then be used by the ArrayDisplay to show more useful names than just the tel_id integer.   Keeping the integer tel_ids is still useful however, so this should be an additional name, not the tel_id itself.

kosack:
this is also related o #1558

maxnoe:
With the added support for multiple configuration files, I think we are basically there.

E.g. for a test processing of the Prod5b La Palma data using the Alpha Configuration, I just had this configuration file:


```
# telescope ids of the 4 LST + 9 LST La Palma alpha configuration in Prod5b
# (out of 84 telescopes total)
EventSource:
  allowed_tels: [1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 19, 35]
```


and called `ctapipe-process` like this:

```
ctapipe-process --config ctapipe_dl2.yml --config prod5b_lapalma_alpha.yml
```

maxnoe:
Maybe a way to override the name would be good though:
```
name='MonteCarloArray_1-11,19,35'
```

kosack:
Yes, we could add an attribute in EventSource perhaps like `force_subarray_name=Prod5b-alpha` , or even just `subarray_name=` and if it is None, it's either read or  auto-generated from the EventSource, and if not it's set by the user.  That would then propagate to output files nicely.

[...]

Issue #1562, opened 2020-12-04
DataWriter should not write pointing info from each event


maxnoe:
The `DataWriter` should not try to reconstruct the monitoring info by looking into unique pointings. 

Rather, it should have a `write_pointing` method, that gets the info that should be written. 

This requires a common way to store the monitoring info however.

maxnoe:
A simple solution would be to implement a `PointingSource` along the lines of `EventSource`, however for simtel files and standard DLx files, this would mean having to repeat the input file (once for `EventSource`, once for `PointingSource`, which is not nice UX.

```python
event_source = EventSource(input_url)
pointing_source = PointingSource(input_url)
writer = DL1Writer(output_file)
for event in source:
    pointing_source(event)
    proces(event)

for pointing in pointing_source:
    writer.write_pointing(pointing)
```

kosack:
yeah, this was discussed in a few past Issues as well.  

see #1041  and the diagram here: https://github.com/cta-observatory/ctapipe/issues/1165#issuecomment-548311167

maxnoe:
I opened another one because this is partciluar about pointing and the "hack" we currently do in the DL1Writer.

For LST, this has the result of storing pointing info for every event although it would be sufficient to store the drive report values with a resolution of 2s.

kosack:
But right now, the DL1 writer does not write per-event pointing, I think, so it should be ok

maxnoe:
> But right now, the DL1 writer does not write per-event pointing, I think, so it should be ok.

Because for LST real data the pointing is different for each event, it is written for every event.

kosack:
It depends a bit on the design: if we have event-wise (or at least rapidly updated pointing info), we could create a general `MonitoringComponent`  that takes input eventwise containers, and writes out every N events a summary, with columns of value_mean, value_std, value_count to some output stream.

You could add that as a sub-component of DL1Writer, e.g. `DL1Writer.pointing_monitor`, and it would do the right thing

maxnoe:
For the lst data, that wouldn't make too much sense since the values come from the drive report with 2 s resolution, so it would make much more sense to just store these data values.

See: https://github.com/cta-observatory/ctapipe_io_lst/blob/ctapipe_compatible/ctapipe_io_lst/pointing.py

kosack:
> For the lst data, that wouldn't make too much sense since the values come from the drive report with 2 s resolution, so it would make much more sense to just store these data values.

[...]

Issue #1484, opened 2020-10-22
Store event-wise pointing at least for MC


maxnoe:
I think we should *optionally* enable storing `dl1/event/telescope/pointing` with `TelEventIndex` at least for simulations since
the timestamp based approach will not work for mcs since they are produced in parallel.
Alternatively, we should atleast include an `obs_index` in the monitoring table.

This will create problems when merging files that do not only contain a single pointing position.

kosack:
what is `obs_index`? Is that `obs_id`? That would perhaps be better to store in the monitoring table along with the time, since then either could be used to link the dataset (in case the time is meaningless).

maxnoe:
yes, sorry.

kosack:
yes, at least adding an obs_id I think is useful.  Eventwise pointing could be done too, but it's really overkill to have the pointing position per event when it changes on a much slower rate (once per second + interpolation to the event time is usually enough).  So far simulations only have fixed alt/az - we will have to consider something better when and if we do ra/dec simulations.   I had always assumed we'd just have a function that just provides event_wise pointing based on the monitoring data, like a `PointingInterpolator`. For MCs, this could just use the obs_id if we add it, and for observations, it would be a proper interpolation of the actual pointing monitoring data.

PR #1726, opened 2021-05-24
Implement ctapipe-process

kosack:
This is intended to be the minimal change to implement DL2 processing in ctapipe.  The interface is not perfect  (see issues below), but is enough to get started - other issues can be solved in future PRs.

## Features Added:

* Renamed `ctapipe-stage1` → `ctapipe-process`
* added a `ShowerProcessor` to the event loop, so DL2 can optionally be computed and written
* added example configurations for *stage1* (DL1), *stage2* (DL2), and *training* (DL1+DL2)
* defined two internal properties: `should_compute_dl1` and `should_compute_dl2` where the logic for when (re-)computation should occur is placed (see note below).  Right now these are set to very simplistic defaults: compute DL2 if either mono or stereo parameters are requested to be written, and compute DL1 if the user wants to write parameters or DL2 showers. This doesn't cover all useful cases yet.

```sh
ctapipe-process --config stage1.json --input events.simtel.gz --output event.DL1.h5
ctapipe-process --config stage2.json --input events.DL1.h5  --output events.DL2.h5
ctapipe-process --config training.json --input events.simtel.gz --output events.DL1DL2.h5
```

## Minor issue on re-computation logic:

right now the logic for `should_compute_dl1` is to always re-compute DL1_PARAMETERS if DL2 is requested and they do not already exist in the input file.  It may be useful to add add an option like "force_recompute_parameters" in case the user wants to ensure they are re-computed even if they already exist (this can be a separate PR). 

## TODOs:
- [x] clean up recompute logic
- [x] add unit tests for each "configuration" (stage1, stage2, training) and for starting from both simtel and DL1 files. 

### config includes

 Right now the config files have a lot of repetition.  That would benefit from using a config system that allows "includes": which means either switching from JSON to Python-based config files (which is supported already by `traitlets.config`, but is slighly ugly in that every parameter needs to have it's full hierarchy specified, like `c.ShowerProcessor.ShowerQualityQuery=...`), or have  a way to specify multiple `--config` options to a ` Tool`, so that we can do for example:

```sh
ctapipe-process --config stage1_config.json --config cuts_config.json --input gammas.simtel.gz
```
EDIT: this is now in issue #1732 

### subarray selection

to do a full analysis at the DL2 level, one needs to select the appropriate subarray.  This is only possible right now by using the `--allowed-tels` option, which is not so friendly.  Allowing it to be selected by name from a file containing mappings (e.g see #1632) would make this much friendlier.  e.g. add options like 
```
--subarray-definition=prod5_subarrays.yaml --subarray="1LST3MST"
```


ctapipe/tools/process.py
```diff
@@ -109,16 +115,37 @@ def setup(self):
...
+                path="/dl1/service/image_statistics",
+                append=True,
+                serialize_meta=True,
+            )
+
+        if self.should_compute_dl2:
+            shower_stats = self.process_shower.check_shower.to_table(functions=True)
+            shower_stats.write(
+                self.write.output_path,
+                path="/dl2/service/image_statistics",
```
maxnoe, line 145: shower_statistics?

kosack: it's not really shower statistics, it's still cuts on each telescope image/parameters that are later used to reconstruct the shower .  I thought about renaming both to cut_statistics, in fact, but decided not to change the data model.

maxnoe: > but decided not to change the data model.

THis is a new field. so what would a different name change?

kosack: no I mean changing both.  In the end I decided to keep the old name for both, rather than naming both to "cut_statistics".

kosack: both are statistics on images used for X, where X is parameterization (DL1) or reconstruction (DL2).  For a future PR/issue, these are not super useful right now, and would benefit from being broken down by tel_id or something.

HealthyPear (commented):
as a first try, I tested a CTAMARS-like analysis with the following config (I attached also the provenance file for debugging),

[config_provenance.zip](https://github.com/cta-observatory/ctapipe/files/6860616/config_provenance.zip)

using only 10 showers for simplicity as

`ctapipe-process --input XXX --config training.json --max-events 10`

with a Prod3b file from that analysis.

I get suspiciously the following warning,

`2021-07-22 09:26:18,109 WARNING [ctapipe.ctapipe-process.DataWriter.HDF5TableWriter] (hdf5tableio._create_hdf5_table_schema): Column tel_ids of container ReconstructedGeometryContainer in table dl2/event/subarray/geometry/HillasReconstructor not writable, skipping`

and indeed the whole DL2-related table is filled with NaNs (also `core_psi` in the DL1 information).

Perhpas it is a bug in the new `DataWriter` or the way it is called in `ctapipe-process`?

maxnoe:
>  I get suspiciously the following warning,
> 
> 2021-07-22 09:26:18,109 WARNING [ctapipe.ctapipe-process.DataWriter.HDF5TableWriter] (hdf5tableio._create_hdf5_table_schema): Column tel_ids of container ReconstructedGeometryContainer in table dl2/event/subarray/geometry/HillasReconstructor not writable, skipping
> 
> and indeed the whole DL2-related table is filled with NaNs (also core_psi in the DL1 information).


I think one has nothing to do with the others. `tel_ids` is a variable length list, it should either be transformed by the writer into the mask representation (using `subarray.tel_ids_to_mask` ) or it should be ignored.

That you get all nans should only mean that none of the events met the quality criteria for stereo reconstruction.

HealthyPear:
Yes sorry there was a typo in the config I wrote, now it works as expected

kosack:
It should automatically transform that column into a fixed-length boolean array.  @HealthyPear can you say what was the typo in your config that caused it to malfunction? We should give a better error if there is a typo...

HealthyPear:
> It should automatically transform that column into a fixed-length boolean array. @HealthyPear can you say what was the typo in your config that caused it to malfunction? We should give a better error if there is a typo...

Nothing that you can really fix from ctapipe I think: as you can see I set ellipticity > 0.1 two times instead of > 0.1 & < 0.6

maxnoe:
> It should automatically transform that column into a fixed-length boolean array.

That is not the case

> can you say what was the typo in your config that caused it to malfunction? We should give a better error if there is a typo...

Not really something you can catch, the image quality query was defined so that no events could ever survive.

kosack:
>> It should automatically transform that column into a fixed-length boolean array.

> That is not the case


It should be, but I guess only if there is a first event to make it initialize: 
https://github.com/cta-observatory/ctapipe/blob/3e1c3088335e16c2fd3ef3736d076cce98b82fca/ctapipe/io/datawriter.py#L400-L404

maxnoe:
> It should be, but I guess only if there is a first event to make it initialize (with no events, you're correct we can't really check for it)

I think it is because it is `None`, not an empty list. It probably should be an empty list.

HealthyPear:
I noticed that `/dl1/monitoring/telescope/pointing/tel_XXX` doesn't have refernce to either of `["obs_id", "tel_id", "event_id"]`.

In case of simulated data it's OK I guess, but since we are anyway preparing for real data, shouldn't we add such columns that table anyway?

maxnoe:
> In case of simulated data it's OK I guess, but since we are anyway preparing for real data, shouldn't we add such columns that table anyway?

There are several issues connected to this:
* https://github.com/cta-observatory/ctapipe/issues/1562
* https://github.com/cta-observatory/ctapipe/issues/1484
* https://github.com/cta-observatory/ctapipe/issues/1041

TL:DR the current format is actually ideal for  observations (where timestamps are unique and we have pointing information every couple of seconds that needs to interpolated to the timestamp of each event) but is is very suboptimal for the simulations.

HealthyPear (approved):

maxnoe (approved):

PR #1727, opened 2021-05-26
Use jit for eigvals and hillas params

maxnoe:
This is an alternate approach to #1512, which I think is also jitting the eigvals calculation.

Performance on this branch:

```
72.04 ± 0.42 µs
```

Performance on current master:
```
425.73 ± 2.94 µs
```


kosack:
This seems to do exactly what is done in #1512, with nearly identical speed.  What is the difference other inlining a few functions?  Note that #1512 was rejected as premature optimization (though I disagree, as the unit-free version which is 30x faster could replace the version used currently in the RTA, which is custom C++ code).

maxnoe:
I implemented eigh as numba jitted function, which triggered this in the first place. Yes the rest seems identical....

Issue #1139, opened 2019-10-18
CHEC camera geometry converter classes broken for Prod4


STSpencer:
The chec_to_2d_array() camera geometry class no longer works with CHEC Prod4 simulations. This is what you get for a muon ring if you try to use it:
<img width="459" alt="Screen Shot 2019-10-18 at 14 03 12" src="https://user-images.githubusercontent.com/31512502/67107574-9f4cb800-f1c4-11e9-8752-7cde69cf4a5d.png">

The cameradisplay class works fine though using event.inst.subarray.tel[tel_id].camera 
<img width="735" alt="Screen Shot 2019-10-18 at 15 45 19" src="https://user-images.githubusercontent.com/31512502/67107718-e2a72680-f1c4-11e9-9d95-d6ac42453ed1.png">

Is there a workaround to extract a 2d array as an image from the cameradisplay class? I've tried using the new geometry with the old 2d array extractor to no avail.

kosack:
Does the general method work (`convert_geometry_hex1d_to_rect2d`)?  I hadn't even noticed that there were hard-coded versions (probably for speed).  We could consider removing them and using the general method always, even if it's slower.  At least until the camera geometries are fixed long-term.   I'm not sure the hard-coded specific ones were ever properly tested with all prods.

kosack:
Perhaps we should consider replacing all of that functionality with the ones implemented by the ML group? Isn't there now a general library for doing it using multiple methods?

kosack:
never mind... of course CHEC is not hexagonal, so the general method won't work... 

Not sure the best way to proceed - we don't want to break reading older prods for now, and it's not easy to detect which prod is being used

kosack:
I guess the problem is just that the pixel ordering has changed, so the method used now is far too simplistic, and probably needs to be replaced with a version that doesn't assume the pixels are in some specific order (similar to how the hex1d_to_rect2d method).

kosack:
currently the pixel transform is hardcoded (but can be replaced with another):
```py3
def chec_transformation_map():
    # By default, pixels map to the last element of input_img_ext (i.e. NaN)
    img_map = np.full([8 * 6, 8 * 6], -1, dtype=int)

    # Map values
    img_map[:8, 8:-8] = np.arange(8 * 8 * 4).reshape([8, 8 * 4])
    img_map[8:40, :] = np.arange(32 * 48).reshape([32, 48]) + 256
[...]

Issue #1641, opened 2021-03-01
Test that test multiple cameras no longer function


kosack:
We have a few tests that loop over all known camera types (testing many different camera geometries), for example test_geometry_converter.  These no longer seem to run on all cameras, since the datasets were moved to the server from ctapipe-extra.  The reason is that `CameraDescription.get_known_camera_names()` no longer returns all camera types, but only those that have been downloaded and cached recently.  This seems to have been broken by #1498

I suspect those tests do not even run if no camera geometries have been downloaded. 

We should fix this so that the old behavior works, even if it has to be hard-coded.  Ideally, it should be possible to fetch all known camera geometries on the dataset server.

maxnoe:
> I suspect those tests do not even run if no camera geometries have been downloaded.

This is why I added the fixture for several camera geometries

PR #1728, opened 2021-05-26
Generalize image converter for square pixels

LukasNickel:
This add new converter functions to go from 1d to 2d images and backwards with square pixels (which is the much easier case obviously). It is basically the code from #1139 and added to be able to apply a gaussian filter in #1723 (at least for square pixels).

I see some issues with this structure of converters and I am also not really happy with the API, but for now I wanted to make minimal changes to the code. Maybe this is something that could be attached to the camera geometry in the future?
Also this is pretty trivial to vectorize, but I am not sure about the existing code for hexagonal pixels, so I skipped on that as well (and I intend to use it in an event loop for now).

Bonus:
The existing unit test was not really doing much, because it compared the pixel shape to a string and thus skipped most geometries. Thats fixed as well.


LukasNickel:
It looks like the tests are not run at all on the CI.
Are there no known CameraDescriptions?

maxnoe:
See https://github.com/cta-observatory/ctapipe/issues/1641

You need to use the `camera_geometries` fixture to make sure the files are downloaded.

maxnoe:
You can actually parameterize fixtures, so maybe we should add a `camera_geometry` parameterized fixture these test can use:

maxnoe:
So instead of 

```
camera_names = CameraGeometry.get_known_camera_names()

@pytest.parametrize("camera_name", camera_names)
def test_foo(camera_name):
    camera = CameraGeometry.from_name(camera_name)
```

this should work better now:

```
@pytest.fixture(params=["LSTCam", "CHEC", "FlashCam", "NectarCam", "MAGICCam", "FACT"])
def camera_geometry(camera_name):
    return CameraGeometry.from_name(camera_name)

def test_foo(camera):
    ...
```

LukasNickel:
Nice!
We actually have these defined in `conftest.py` already.
I will try to replace it in other tests as well

kosack:
By the way, if you want to make larger changes to the API, that is also fine - this code was used a lot in the studies on wavelet cleaning for example, but those already no longer with with the latest ctapipe and are being re-written, so right now there is no real constraint on the API and a nicer one would be welcome

LukasNickel:
> By the way, if you want to make larger changes to the API, that is also fine

I would like to, but right now I am not sure what the best way would be. At that point it would probably also be good to have a look into the efforts of the ML group as you mentioned above. Probably not worth to rethink this twice.
In my opinion this would best be added to the CameraGeometry so that the user does not need to worry about pixel shapes at all, but there might be better places still.

kosack:
We could maybe add  convenience functions to `CameraGeometry ` as you suggest, something like:
```py3
2d_geom  = geom.to_2d(method='skew')
2d_geom.to_1d()   
```
That just calls these functions in the background.  In the future we would implement more methods or use the ones by the ML group.   The `to_1d()` would need to know what method was used originally and should only work on 2D CameraGeometries, so maybe we would need a subclass CameraGeometry2D as the return type.  That can come with another future PR if you prefer though.

kosack (dismissed):

ctapipe/image/geometry_converter_rect.py
```diff
@@ -0,0 +1,70 @@
...
+    Returns:
+    --------
+    (rows, cols): (ndarray, ndarray)
+        Arrays holding the row and col of each pixel, needed to transform back
+    image_2d: ndarray
+        Square image
+    """
+    size = np.sqrt(geom.pix_area)
+    col = pos_to_index(geom.pix_x, size)
+    row = pos_to_index(geom.pix_y, size)
```
maxnoe, line 45: We could add the `row, col` part to the `CameraGeometry` and cache it. Here we recalculate it each time.

LukasNickel: We could, but for hexagonal pixels its actually a whole new geometry instead and would not modify the existing geometry. To me that lead to the restructuring discussion above, but we can of course add it here already or add some `_transformation_cache`  or smth like that

ctapipe/instrument/camera/geometry.py
```diff
@@ -383,6 +383,36 @@ def _all_pixel_areas_equal(self):
         """
         return ~np.any(~np.isclose(self.pix_area.value, self.pix_area[0].value), axis=0)
 
+    @lazyproperty
+    def _pixel_rows(self):
```
maxnoe, line 387: please make this public. It's useful and there is no reason why this should be hidden.

maxnoe: Also, call it `pixel_row`, we don't use plural for arrays of stuff, compare pixel `x` or `area` or `diameter`

kosack (dismissed):

LukasNickel:
Had to merge the master due to conflicts.

The failing tests do not seem to be related, but mention failed connections to the dataserver. Are there any issues currently?

maxnoe:
@LukasNickel @kosack mentioned connection issues as well for him locally today

kosack (dismissed):

ctapipe/image/geometry_converter_rect.py
```diff
@@ -0,0 +1,56 @@
...
+    Returns:
+    --------
+    (rows, cols): (ndarray, ndarray)
+        Arrays holding the row and col of each pixel, needed to transform back
+    image_2d: ndarray
+        Square image
+    """
+    row = geom.pixel_row
+    col = geom.pixel_column
+    image_square = np.full((row.max() + 1, col.max() + 1), np.nan)
```
maxnoe, line 32: This could be `geometry.n_rows / n_cols`

maxnoe (changes requested):
Acutally now looking at this, I think we should just add a method `to_regular_image` to the geometry. `pixel_row`, `pixel_col` like defined now would also apply to hex images, since that is where the pixels end up in the square image.

ctapipe/image/tests/Untitled.ipynb
```diff
@@ -0,0 +1,6 @@
+{
```
maxnoe, line 1: Untitled empty notebook

LukasNickel:
> Acutally now looking at this, I think we should just add a method `to_regular_image` to the geometry. `pixel_row`, `pixel_col` like defined now would also apply to hex images, since that is where the pixels end up in the square image.

Done (including the reverse method). Some code got moved, the image conversion part is now in `instrument/camera` as its tied to the geometry. The nice thing about this is that the same function can be used for both hexagonal and square pixels.
I merged `pixel_row` and `pixel_col` to avoid having the calculations in the `__init__` or duplicating it. If you feel like we want them available, they can be added again. To me it felt like the `CameraGeometry` api is already pretty cluttered and I dont see a real use outside of debugging (its of course still available, just less accessible).

New Feature:
You can transform `n` images at once as can be done with the cleaning methods (but no transpose needed)

kosack:
I don't quite like the term "regular image", since it is ambiguous. Could you use  a more specific term? For me for a Cherenkov camera, a "Regular image" would be hexagonal, but it seems it's 2D.  Why not just call it 2D and 1D?

kosack:
Another question: looking to the future when we support more ways to do the re-sampling, how will that be handled? Does it still make sense to have it as methods of CameraGeometry, or as functions that take a CameraGeometry?

maxnoe:
>  Does it still make sense to have it as methods of CameraGeometry, or as functions that take a CameraGeometry?

I think it still makes sense and the we can just add a `method=` kwarg to the conversion. However, not all (most?) conversion methods from irregular grids to regular grids will be invertible.


> geom.image_1d_to_cartesian(image1d)? 

Regular is meant for "regular square grid" as opposed to an irregular square (CHEC) or hex grid. 

```
geom.image_{to,from}_regular_square_grid()
```
?

kosack:
A Cartesian image is orthogonal and has regular pixel spacing, so `geom.image_{to,from}_cartesian(image)` would also work (and is a bit shorter)

LukasNickel:
> a "Regular image" would be hexagonal, but it seems it's 2D. Why not just call it 2D and 1D?

2D seems even more ambigous. I would consider a hexagonal grid on a plane to be perfectly 2D.

> Another quibble: methods starting with `to_x()` usually convert the object itself 

I can see that, agreed.

> A Cartesian image is orthogonal and has regular pixel spacing, so `geom.image_{to,from}_cartesian(image)` would also work 

Im voting for this one

maxnoe:
> A Cartesian image is orthogonal and has regular pixel spacing, so geom.image_{to,from}_cartesian(image) would also work

The problem I have with cartesian is that it uses opposite axis (y positive up) direction for the y-axis than the usual image convention (y positive down).


What do we do here? Does "imshow" look like the same?

kosack:
> What do we do here? Does "imshow" look like the same?

The actual projection (how x/y map to pixels) probably doesn't matter, it should just be documented.  Imshow in fact puts Y on the Y axis, but positive down, unlike a normal plot (unless you add an option to flip it), so even using imshow doesn' give what one would expect.  In the end machine learning and cleaning algorithms don't really care I guess.

LukasNickel:
> What do we do here? Does "imshow" look like the same?

Currently the axes are set up in a way, that `imshow` matches the `CameraDisplay` in terms of orientation (for square pixels at least). I found this to be the most natural way of comparing the images.
But this is an argument against `carthesian` for sure.

Naming things is hard...

kosack:
By the way, imshow only defaults to y-negative, but you can have it display either way, so it doesn't matter really which way. It may be clearer to have a traditional x-y axis.  
```py3
plt.imshow(image, origin="lower") 
```
puts 0,0 at the lower left as usual

kosack:
In fact, at one point I had thought of defining a `CameraImage` that was just a tuple of (geom, image_1d_array), but in the end didn't do it for simplicity.  But we could consider that, and then these conversion routines would be methods of CameraImage not CameraGeometry

LukasNickel:
> The real problem is that we have overloaded the word "image"... we could use Cherenkov Image or Camera Image to mean the 1D ones ..

Side Note: We do name the peak times array `image` or `time_image` occasionally as well, so its already ambiguous.
Would you prefer `representation`?

```
image_{to_from}_cartesian_representation()
```

LukasNickel:
> In fact, at one point I had thought of defining a `CameraImage` that was just a tuple of (geom, image_1d_array), but in the end didn't do it for simplicity (often you have many images, and only one geometry for example). But we could consider that, and then these conversion routines would be methods of CameraImage not CameraGeometry. But that is a larger API change

It sounds excessive to create a new object for each image. If we were to perform all operations on `n` images at a time, I would be more in favor for this. Then all user facing functions could get this and avoid the extra `geom` every time.
To me this sounds like we reach the limits of what the `containers` can realistically do and we might need a more generic "datastore" at some point?

kosack:
> It sounds excessive to create a new object for each image.

Yes this was the main reason I did no do it, along with added complexity when writing the data.

kosack (dismissed):

ctapipe/instrument/camera/geometry.py
```diff
@@ -383,6 +387,123 @@ def _all_pixel_areas_equal(self):
...
+                np.rollaxis(np.atleast_3d(cols_2d), 2, 0)
+            )[..., square_mask]
+            pixel_row = rows_1d
+            pixel_column = cols_1d
+
+        elif self.pix_type is PixelShape.SQUARE:
+            pixel_row = get_orthogonal_grid_indices(self.pix_y, np.sqrt(self.pix_area))
+            pixel_column = get_orthogonal_grid_indices(
+                self.pix_x, np.sqrt(self.pix_area)
+            )
```
maxnoe, line 443: This errors with a local variable name error in case of an unsupported pixel shape, we should raise a more helpful error here

ctapipe/instrument/camera/geometry.py
```diff
@@ -383,6 +387,123 @@ def _all_pixel_areas_equal(self):
...
+        Namely, slanting and stretching of the 1d pixel positions
+        to align them nicely.
+        Beware, that this means the pixel geometries on this
+        grid to not match the one in the geometry.
+
+        Returns:
+        --------
+        (rows, columns) of each pixel if transformed onto an orthogonal grid
+        """
+        if self.pix_type == PixelShape.HEXAGON:
```
maxnoe, line 405:
```suggestion
        if self.pix_type in {PixelShape.HEXAGON, PixelShape.CIRCLE}:
```

kosack (dismissed):

ctapipe/instrument/camera/tests/test_geometry.py
```diff
@@ -2,9 +2,30 @@
 import numpy as np
 from astropy import units as u
 from ctapipe.instrument import CameraGeometry, PixelShape
+from ctapipe.image.toymodel import Gaussian
 import pytest
 
 
+def create_mock_image(geom):
```
maxnoe, line 9: Is this function superfluous? It is the same as below and according to codecov not covered by tests

LukasNickel: It is.
I moved the image conversion tests and forgot to remove this as well

kosack (dismissed):

LukasNickel:
Sorry @kosack I forgot to remove a now obsolete import in the last commit, which made the tests fail.

kosack (approved):

maxnoe (approved):

Issue #1729, opened 2021-05-27
[FR - ctapipe-merge]: Add option 'split_datasets_by' 


TjarkMiener:
Would be great to have the ability to merge tel_types at this stage of the pipeline (if stage1 files already have been produced).

LukasNickel:
I am not sure I understand what you are trying to achieve.
You want to merge dl1 files, but only specific telescope types? Or do you want to group telescope types in the output?

TjarkMiener:
I want to group the telescope types in the output with stage1 h5 files as input. The output file looks like this (hopefully the same as if you would run stage1 with `--split_datasets_by tel_type` on simtel files):
![Screenshot from 2021-05-27 12-05-18](https://user-images.githubusercontent.com/37835610/119808017-ff005e80-bee3-11eb-8b84-fe99ef39a3a4.png)

LukasNickel:
I am not sure if this is intended/wanted.
Maybe @kosack can comment.

Why do you need this? It looks organized if you want to work on one telescope type at a time of course, but you should always be able to combine the tables in your own code, especially as this does not even seems to be ordered by trigger time

maxnoe:
@LukasNickel this option is supported by the stage1 tool currently. I think it makes sense to also have it for the merge tool.

We can discuss if we want to keep this option at all, but having this option only for some tools and not all is not great.

vuillaut:
> @LukasNickel this option is supported by the stage1 tool currently. I think it makes sense to also have it for the merge tool.
> 
> We can discuss if we want to keep this option at all, but having this option only for some tools and not all is not great.

Does that mean that there are several DL1 data models?

LukasNickel:
Oh it is supported? In that case we can add it here of course.

maxnoe:
That's splitting hairs :D 

You could call it one data model with two allowed structures...

But you are right. And I wouldn't recommend using the "tel_type" option. See e.g. this connected issue #971, #1426 and some more related to this.

vuillaut:
> > Does that mean that there are several DL1 data models?
> 
> That's splitting hairs :D
> 
> You could call it one data model with two allowed structures...
> 
> But you are right. And I wouldn't recommend using the "tel_type" option. See e.g. this connected issue #971, #1426 and some more related to this.

Is there a big advantage (reading speed?) having telescopes merged per tel_type?
Having a method `read_dl1_table_per_tel_type` or something similar should be enough to cover the "ease of use" advantage.

[...]

PR #1730, opened 2021-05-27
added split_datasets_by option to ctapipe-merge

TjarkMiener:
address #1729


kosack (changes requested):
Good idea.  Could you perhaps add (or extend an existing) test to make sure this functionality works?

maxnoe:
@kosack please see the discussion in #1729 

I am not sure we add this feature and maybe we should also remove the option from the datawriter

TjarkMiener:
@maxnoe this feature might be essential for me. Can we rediscuss this feature once I have some benchmarks? Thanks!

TjarkMiener:
Hi @kosack @maxnoe 
Sorry for leaving this pending for so long! 

For the test, I thought on checking for the table length:
```
def test_allowed_tels_and_split_datasets_by(tmp_path, gamma_dl1_path):
    from ctapipe.tools.dl1_merge import MergeTool

    # create a second file so we can test the split by tel type
    output = tmp_path / "split_datasets_by_tel_type.dl1.h5"

    ret = run_tool(
        MergeTool(),
        argv=[
            str(gamma_dl1_path),
            f"--output={output}",
            "--allowed-tels=[1,2,3,4]",
            "--split_datasets_by=tel_type",
            "--overwrite",
        ],
        cwd=tmp_path,
    )
    assert ret == 0

    original_table_length = 0
    for tel_id in range(1,5):
        original_table_length += len(read_table(str(gamma_dl1_path), f"/dl1/event/telescope/parameters/tel_{tel_id:03d}"))

    merged_table_length = len(read_table(output, f"/dl1/event/telescope/parameters/LST_LST_LSTCam"))

    assert original_table_length == merged_table_length
```

However, this test is failing (works just fine outside of pytest with handmade dl1 files):
```
>           original_table_length += len(read_table(str(gamma_dl1_path), f"/dl1/event/telescope/parameters/tel_{tel_id:03d}"))

ctapipe/tools/tests/test_merge.py:130: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
ctapipe/io/astropy_helpers.py:59: in read_table
    table = h5file.get_node(path)
../../anaconda3/envs/cta/lib/python3.7/site-packages/tables/file.py:1642: in get_node
    node = self._get_node(nodepath)
../../anaconda3/envs/cta/lib/python3.7/site-packages/tables/file.py:1590: in _get_node
    node = self._node_manager.get_node(nodepath)
../../anaconda3/envs/cta/lib/python3.7/site-packages/tables/file.py:432: in get_node
    node = self.node_factory(key)
../../anaconda3/envs/cta/lib/python3.7/site-packages/tables/group.py:1178: in _g_load_child
    node_type = self._g_check_has_child(childname)
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

self = <closed tables.group.RootGroup at 0x7f5d2fe21cb0>, name = '/dl1/event/telescope/parameters/tel_001'

    def _g_check_has_child(self, name):
        """Check whether 'name' is a children of 'self' and return its type."""
    
        # Get the HDF5 name matching the PyTables name.
        node_type = self._g_get_objinfo(name)
        if node_type == "NoSuchNode":
            raise NoSuchNodeError(
                "group ``%s`` does not have a child named ``%s``"
>               % (self._v_pathname, name))
E           tables.exceptions.NoSuchNodeError: group ``/`` does not have a child named ``/dl1/event/telescope/parameters/tel_001``

../../anaconda3/envs/cta/lib/python3.7/site-packages/tables/group.py:395: NoSuchNodeError
```

Can you please help me?

HealthyPear:
The fixture that creates the file stored under `gamma_dl1_path` is using the example JSON configuration file.
If you try to create that file externally you will see (at least I saw) that it returns an empty file.

For these DL1 / merge tests, physics performance is not required, so I'd use a dedicated less restrictive configuration file which makes some images available (the test file is not very big).

I tested it simply by lowering all cleaning threshold to 2/1 phe and got a valid file.

then you need to open the file you created and check which tel_id actually survived, so you can pin point the correct one and not get the error you've shown (`tables.exceptions.NoSuchNodeError: group ``/`` does not have a child named ``/dl1/event/telescope/parameters/tel_001`)

HealthyPear (dismissed):

TjarkMiener:
Thanks for your time reviewing the PR @HealthyPear! I just merged the current master into this PR with minor changes (see [21261dc](https://github.com/cta-observatory/ctapipe/pull/1730/commits/21261dc69cb540a5490540be30bcfe916a79234e)) and the tests are failing. Do you have an idea of what could have gone wrong? Once it's resolved can you please merge this PR? Many thanks in advance!

maxnoe:
This is what the CI system reports (the `pyflakes ctapipe` to be precise):

```
ctapipe/tools/tests/test_merge.py:106:5 'ctapipe.instrument.SubarrayDescription' imported but unused
ctapipe/tools/tests/test_merge.py:145:9 undefined name 'SubarrayDescription'
```

So please check the code in the unit tests and fix the imports. You can also run `pyflakes ctapipe` locally to check before pushing.

kosack:
Once the build is fixed, this is probably ok to merge.  

Just keep in mind that we will very likely *deprecate* the ability to write events grouped by telescope type in the future. That will ensure that the DL1/DL2 format does not have two incompatible variants.  That in turn simplifies a lot of the code needed to read it like the DL1EventSource and the future TableLoader, and makes it so that we can do a full DL0→DL1 production on the Grid only once, not twice with two splittings.  

That will not break the use case of reading data per telescope type, since that is already supported in #1771 which will be merged soon, it would just mean that you might not have files with the split already built-into them in the future. So it's just a heads up, if you have code that relies on this feature, you would need to make a small change.

TjarkMiener:
@maxnoe Thanks for the tip with `pyflakes`! Before I used the GitHub tool to resolve the merging conflicts and the automerge messed up the tests. It should be solved in the next push since `pyflakes` and `pytest` are passing locally for the merger test now. Sorry for the inconvenience!

@kosack Thanks for the heads up! Even though I was hoping that this feature will be supported as long as possible ;), since this is very crucial for us, particularly at training phase. I don't think a full DL0-DL1 with two different splittings is needed or wanted. The purpose of this PR is to by-pass a reproduction from DL0 on the Grid, which would be a waste of resources. What do you think of depreacating, i.e. removing, this splitting feature in the `Process tool`, but keeping it in the `Merger tool`? This will ensure (1) that a full production can only produce one standard format, where the tables are split by tel_ids, and (2) that users, who need the tables split by tel_types, are forced to make an additional step in their analysis chain! I think it would be fair to not support the "modified" format in the DL1EventSource and/or the future TableLoader by just printing a error message "hey don't do this additional merger step". What do you think @kosack @HealthyPear @maxnoe? I'm currently also converting our Deep Learning classes, namely dl1dh and ctlearn, into core components to align the DL pipelines with ctapipe. Once this is completed, it may get clearer why this split by tel_types is so crucial for our approach.

maxnoe:
@TjarkMiener I still don't understand why having it split by telescope type at the file level is *crucial* for you. 

The new table reader provides an API to get the data for a single telescope type efficiently also when the file is split by telid. Did you actually test that having it in a single table in the file is that much faster than merging tables? If yes, what is the bottleneck?

TjarkMiener:
My issue isn't related to the reading speed but rather to the RAM usage. The bottleneck here is the file structure itself, which you can see on [L793-L800](https://github.com/cta-observatory/dl1-data-handler/blob/stage1/dl1_data_handler/reader.py#L793-L800) in my code.  Especially line [L799](https://github.com/cta-observatory/dl1-data-handler/blob/stage1/dl1_data_handler/reader.py#L799) was identified by the memory profiler to be the most consuming one. For the split by tel_types, we only have to call `TABLE._f_get_child()` once and more important outside the for loop over the tel ids as you can in [L830](https://github.com/cta-observatory/dl1-data-handler/blob/stage1/dl1_data_handler/reader.py#L830). Honestly, I was very surprise when I found out that this was causing my RAM problems. My machines equipped with 62GB of RAM weren't able to train a model with 40 SSTs, where the files were split by tel_ids. For the training with files splited by tel types, roughly 10 GBs of RAM is needed (independent from the number of telescopes per type).

kosack:
That seems a bit strange: reading for example 20 per-telescope tables in memory vs one big one should be roughly the same, since the total data size is identical.  The difference should only be the metadata, which will be 20x larger in the split by tels case, but that should be very small compared to the data. I wonder if there is a memory management problem (the table being copied for example, and the original not being garbage collected), or that it is reading all of the tables not just the requested one?  In any case, it's something to look into. 

But that's outside the scope of this PR (maybe should open an issue), and we can of course hold off deprecating the tel_type splitting until that is resolved.

maxnoe:
@kosack That seems specific to the implementation in DL1 data handler

TjarkMiener:
As discussed offline with @maxnoe, this PR requires PR [1784](https://github.com/cta-observatory/ctapipe/pull/1784) (traitlets fix) to pass the automatic checks.

Thanks for the further input @kosack @maxnoe! I have to admit that I adopted this reading structure of the previous DL1DH reader, which was developed by @aribrill. Ari, could you please confirm or dissent my understanding? My understanding is that we never open the table and read the whole table into the memory but rather getting the child of the node via `node._f_get_child(child)` and only retrieving the desired image via pytables indexing, i.e. we only have the requested images in memory (and the child)?

I'm agreeing that ctapipe should make decisions independent from sub-packages based on ctapipe! However, it seems that this feature is pretty handy for the standard pipeline `protopipe` and alternative approaches (at least for the time being).

maxnoe:
> and only retrieving the desired image via pytables indexing, i.e. we only have the requested images (and the child) in memory?

Images, plural. And if I read the code correctly, you do this in a loop appending to a pure python list, for each image opening the table again. That has to be much more inefficient than first buildung an index of what images you want to read an then reading all at once. (In any language)

maxnoe:
@TjarkMiener #1784 is merged, if you update your branch (by rebasing or merging origin/master), the test should pass

ctapipe/tools/tests/test_merge.py
```diff
@@ -122,3 +124,24 @@ def test_allowed_tels(tmp_path, dl1_file, dl1_proton_file):
 
     s = SubarrayDescription.from_hdf(output)
     assert s.tel.keys() == {1, 2}
+
+
+def test_split_datasets_by(tmp_path, dl1_file, dl1_proton_file):
```
maxnoe, line 129: You don't actually validate anything here, you just check that the command runs without problem.

Please add checks that validate that the output file is indeed how it should be. E.g. by opening it with the DL1 event source and iterating over the first couple events.

TjarkMiener:
@maxnoe Let's switch to an issue in DL1DH for this ;-D

ctapipe/tools/tests/test_merge.py
TjarkMiener: Please see above, I suggested checking the total length of the tables. However, the DL1 files are too small to run proper checks as pointed out by @HealthyPear. How should we proceed?

maxnoe: Can we use other files? But then it's a general issue which we can solve after this PR.

We should have files that contain at least some triggered events.

TjarkMiener:
>  #1784 is merged, if you update your branch (by rebasing or merging origin/master), the test should pass

@maxnoe  I successfully `git merge origin/master` into this PR, however `pytest` is still not passing because of the image_conversion test, which haven't been touched by this PR.  
=================================================================================== short test summary info ====================================================================================
FAILED ctapipe/image/tests/test_toy.py::test_intensity[None] - assert 177.0 <= 168
FAILED ctapipe/instrument/camera/tests/test_image_conversion.py::test_single_image[ASTRICam] - numpy.AxisError: axis 1 is out of bounds for array of dimension 0
FAILED ctapipe/instrument/camera/tests/test_image_conversion.py::test_multiple_images[ASTRICam] - numpy.AxisError: axis 1 is out of bounds for array of dimension 1
FAILED ctapipe/instrument/camera/tests/test_image_conversion.py::test_single_image[CHEC] - numpy.AxisError: axis 1 is out of bounds for array of dimension 0
FAILED ctapipe/instrument/camera/tests/test_image_conversion.py::test_multiple_images[CHEC] - numpy.AxisError: axis 1 is out of bounds for array of dimension 1
FAILED ctapipe/instrument/camera/tests/test_image_conversion.py::test_single_image[SCTCam] - numpy.AxisError: axis 1 is out of bounds for array of dimension 0
FAILED ctapipe/instrument/camera/tests/test_image_conversion.py::test_multiple_images[SCTCam] - numpy.AxisError: axis 1 is out of bounds for array of dimension 1
======================================================== 7 failed, 638 passed, 2 skipped, 19 xfailed, 107 warnings in 397.89s (0:06:37) ========================================================

maxnoe:
@TjarkMiener The CI tests are passing. Is your local environment up-to-date?

TjarkMiener:
@maxnoe Seems like the local environment messed things up since the CI tests passed. I cleaned my working directory and reinstalled the master. However, `pytest` is still failing for the same tests as above for me.

``` 
 2030  cd ..
 2031  rm -rf ctapipe
 2032  conda deactivate cta
 2033  conda deactivate
 2034  git clone https://github.com/cta-observatory/ctapipe
 2035  cd ctapipe/
 2036  conda env create -n cta-dev -f environment.yml
 2037  conda activate cta-dev
 2038  pip install -e .
 2039  pytest
```

ctapipe/tools/tests/test_merge.py
```diff
@@ -122,3 +124,24 @@ def test_allowed_tels(tmp_path, dl1_file, dl1_proton_file):
...
+            str(dl1_file),
+            str(dl1_proton_file),
+            f"--output={output}",
+            "--split_datasets_by=tel_type",
+            "--overwrite",
+        ],
+        cwd=tmp_path,
+    )
+    assert ret == 0
+
```
kosack, line 147: I'd suggest adding to this test a check that indeed the output has the right datasets in it (e.g. that if split by tel type, then open the file and check that there is a dataset named by the tel type).

Basically, the test should ensure that the functionality is there.  For example, if later simply ignore the split by by tel_type option and always split only by tel_id,  this test should fail, but right now it does not.

kosack (changes requested):
just a small request: please make the test actually test for the expected result

ctapipe/tools/tests/test_merge.py
kosack: Should be enough to do something like 
```py3
with tables.open_file(output) as h5file:
    assert h5file.get_node("/dl1/event/telescope/parameters/LST_LST_LSTCam") is not None
```

TjarkMiener: > ```
> def test_allowed_tels_and_split_datasets_by(tmp_path, gamma_dl1_path):
>     from ctapipe.tools.dl1_merge import MergeTool
>     from ctapipe.io import read_table
> 
>     # create a second file so we can test the split by tel type
>     output = tmp_path / "split_datasets_by_tel_type.dl1.h5"
> 
>     ret = run_tool(
>         MergeTool(),
>         argv=[
>             str(gamma_dl1_path),
>             f"--output={output}",
>             "--allowed-tels=[1,2,3,4]",
>             "--split_datasets_by=tel_type",
>             "--overwrite",
>         ],
>         cwd=tmp_path,
>     )
>     assert ret == 0
> 
>     original_table_length = 0
>     for tel_id in range(1,5):
>         original_table_length += len(read_table(str(gamma_dl1_path), f"/dl1/event/telescope/parameters/tel_{tel_id:03d}"))
> 
>     merged_table_length = len(read_table(output, f"/dl1/event/telescope/parameters/LST_LST_LSTCam"))
> 
>     assert original_table_length == merged_table_length
> ```
> 
> However, this test is failing (works just fine outside of pytest with handmade dl1 files):
> 
> ```
> >           original_table_length += len(read_table(str(gamma_dl1_path), f"/dl1/event/telescope/parameters/tel_{tel_id:03d}"))
> 
> ctapipe/tools/tests/test_merge.py:130: 
> _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
> ctapipe/io/astropy_helpers.py:59: in read_table
>     table = h5file.get_node(path)
> ../../anaconda3/envs/cta/lib/python3.7/site-packages/tables/file.py:1642: in get_node
>     node = self._get_node(nodepath)
> ../../anaconda3/envs/cta/lib/python3.7/site-packages/tables/file.py:1590: in _get_node
>     node = self._node_manager.get_node(nodepath)
> ../../anaconda3/envs/cta/lib/python3.7/site-packages/tables/file.py:432: in get_node
>     node = self.node_factory(key)
> ../../anaconda3/envs/cta/lib/python3.7/site-packages/tables/group.py:1178: in _g_load_child
>     node_type = self._g_check_has_child(childname)
> _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
> 
> self = <closed tables.group.RootGroup at 0x7f5d2fe21cb0>, name = '/dl1/event/telescope/parameters/tel_001'
> 
>     def _g_check_has_child(self, name):
>         """Check whether 'name' is a children of 'self' and return its type."""
>     
>         # Get the HDF5 name matching the PyTables name.
>         node_type = self._g_get_objinfo(name)
>         if node_type == "NoSuchNode":
>             raise NoSuchNodeError(
>                 "group ``%s`` does not have a child named ``%s``"
> >               % (self._v_pathname, name))
> E           tables.exceptions.NoSuchNodeError: group ``/`` does not have a child named ``/dl1/event/telescope/parameters/tel_001``
> 
> ../../anaconda3/envs/cta/lib/python3.7/site-packages/tables/group.py:395: NoSuchNodeError
> ```

I think the problem is that the node is not created when no triggered event survived? Neither tel_001 nor LST_LST_LSTCam can be found in the "split by tel_ids" files and  "split by tel_types" files, respectively, for the test data.

ctapipe/tools/tests/test_merge.py
```diff
@@ -107,13 +107,15 @@ def test_allowed_tels(tmp_path, dl1_file, dl1_proton_file):
...
     output = tmp_path / "merged_allowed_tels.dl1.h5"
+
     ret = run_tool(
         MergeTool(),
         argv=[
             str(dl1_file),
             str(dl1_proton_file),
             f"--output={output}",
             "--allowed-tels=[1,2]",
+            "--split_datasets_by=tel_type",
```
maxnoe, line 118: Why this change?

ctapipe/tools/tests/test_merge.py
maxnoe: There are events in there, so at least one telescope type is there. You can check which ones are there by inspecting the created file with vitables, something like `vitables /tmp/pytest-of-$USER/pytest-current/<test_name>current/<file>`

ctapipe/tools/tests/test_merge.py
TjarkMiener: One of those automatic code style anaylsis sent a warning (which worried me) that not all edit code in this PR was covered by the pytest, so I added the split also in the allowed tells test. Would it be better to have it separated in an additional test `def test_allowed_tels_and_split_datasets_by`?

ctapipe/tools/tests/test_merge.py
TjarkMiener: Thanks @maxnoe! Triggered events were found for tel id 25, 125 (FlashCam), and 130 (NectarCam). I updated the merger test accordingly and it passes locally.

aribrill:
> Thanks for the further input @kosack @maxnoe! I have to admit that I adopted this reading structure of the previous DL1DH reader, which was developed by @aribrill. Ari, could you please confirm or dissent my understanding? My understanding is that we never open the table and read the whole table into the memory but rather getting the child of the node via `node._f_get_child(child)` and only retrieving the desired image via pytables indexing, i.e. we only have the requested images (and the child) in memory?

This is my understanding as well. I replied in more detail on the DL1DH issue: https://github.com/cta-observatory/dl1-data-handler/issues/104#issuecomment-916362431

maxnoe:
Hi @TjarkMiener,

I am sorry for the late reply and for closing this Pull Request in which you put a lot of work, but as I wrote in the corresponding issue #1729, we don't want to keep supporting to different structures of files, it adds a lot of overhead.

Also see the corresponding issue on removing support on older versions of files here: #1983

maxnoe:
The use case of getting all images by type is now covered by the `ctapipe.io.TableLoader`, which will stack the images from the different files for you. 

Since dl1dh does not actually load all images into ram but accesses the low-level tables, I also don't see a reason why you couldn't do that with the by id structure by randomly selecting a telescope and then randomly selecting an image from that telescope's table (or whatever else the strategy is).

Doing tables by telescope type breaks some assumptions that are made in implementing chunked reading in the table loader, which is important for memory efficient application of machine learning algorithms to large datasets (like merged dl2 test data sets).

So I hope you understand that we didn't take this decision light heartedly.

TjarkMiener:
Thanks for your kind message @maxnoe! The id structure is also supported by dl1dh, but I would need to fix [#104](https://github.com/cta-observatory/dl1-data-handler/issues/104).

PR #1731, opened 2021-05-27
Switch on doctests in the CI

maxnoe:


LukasNickel:
Nice!
Looks like its only the `stereo_reconstruction` example, that has not been updated to API changes in a while.
All the `ctapipe.image.*` import can import from `ctapipe.image` directly and `leakage` is now called `leakage_parameters` (imports and l.64)

kosack (approved):

maxnoe:
The only thing that fails is the camdemo tool. I cannot reproduce locally and there is no indication why this test fails.

Do we really need this tool?

HealthyPear:
> The only thing that fails is the camdemo tool. I cannot reproduce locally and there is no indication why this test fails.

You could add the `--debug` and/or `--log-level` to `run_tool` to check what is failing on the CI (I guess something related to the cached test files or similar, since the code works locally).

HealthyPear (approved):

kosack (approved):

PR #1736, opened 2021-05-31
Adapt subarray for 'allowed_tels' option in merger_tool

Hckjs:
ref to #1697

### DIFF
--- a/ctapipe/tools/dl1_merge.py
+++ b/ctapipe/tools/dl1_merge.py
@@ -204,14 +204,12 @@ class MergeTool(Tool):
             sys.exit(1)
 
         # create output file with subarray from first file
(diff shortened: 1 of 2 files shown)


maxnoe (commented):
Please also add a unit test for this case

LukasNickel:
Not really connected to this PR, but while we are at it:
Is there any reason to have the merge tool tests separate from the other tools? We could just use the fixtures from there.

maxnoe:
@LukasNickel Yes, that the `test_tool` file got way to large and we started to split it up.

Please don't remerge this, rather we should also create own files for the other ones.

If we want to have common fixtures for tests in `tools` we can add a `conftest.py` file into the directory or add the fixtures to `ctapipe/conftest.py`

kosack (approved):

LukasNickel (approved):

PR #1737, opened 2021-05-31
Test on 3.8 and 3.9, use 3.9 in environment.yml

maxnoe:


maxnoe:
There is no conda package of bokeh 1.x for python 3.9 and there seems to be an independent  new issue for opening a hdf5 file multiple times.

LukasNickel:
Pytables seems to put a strict policy in place if the hdf5 version is not newer than 1.8.7:
```
if tuple(map(int, utilsextension.get_hdf5_version().split('-')[0].split('.'))) \
                                                                        < (1, 8, 7):
    _FILE_OPEN_POLICY = 'strict'
else:
    _FILE_OPEN_POLICY = 'default'
```
How is hdf5 installed? The conda packages should all be newer

maxnoe:
@LukasNickel the conda tests passed, so...

With pip, we either get a wheel which includes its own libhdf5 or we compile it against a system hdf5 library.

maxnoe:
Where a version this old comes from, I have no idea since ubuntu should have a much newer version

maxnoe:
Both h5py and tables were installed using wheels in the test that failed.

maxnoe:
The manylinux wheel for tables 3.6.1 has this old version of libhdf5:
```
In [1]: import tables

In [2]: tables.__version__
Out[2]: '3.6.1'

In [3]: tables.hdf5_version
Out[3]: '1.8.5-patch1'```

maxnoe:
See https://github.com/PyTables/PyTables/issues/884

kosack:
probably we should not be opening any HDF5 files more than once at a time... I tried to avoid that in the DataWriter, which only uses a single HDF5TableWriter, and then should only try to write the instrument info and any other astropy tables when that is not yet constructed, or after it is destructed  - did I miss a case perhaps?   Using a more recent libhdf5 is also important, but it shouldn't be this large an issue.

maxnoe:
@kosack It's in the DL1 EventSource I think, which gives a path to many table readers, not an already opened h5file.

maxnoe:
Now it fails because Bokeh 1.x is not available for python 3.9. We need to update...

TjarkMiener:
@maxnoe @kosack Are there any plans to upgrade to Python 3.8+ in the near future? It'd be quite crucial for me (the latest GPUs only support CUDA 11.X which requires TensorFlow v2.5+ and Python 3.8+). At the moment I'm installing ctapipe from source with Python 3.8, which is working alright. Nevertheless, it would be great to upgrade to a later Python version!

maxnoe:
Python 3.8 is already fully supported. What is the problem?

TjarkMiener:
I forgot to rebuild my uploaded conda package with the correct Python version. Somehow the log from the error confused me and after seeing this PR draft I thought it was related. I will just upload a rebuild version, which should solve the problem. Sorry for the false alarm and thanks for the quick response!

PR #1740, opened 2021-06-01
Parallel tests

### LINKED ISSUE #1641: Test that test multiple cameras no longer function

maxnoe:
To speed up the continous integration and running the tests locally, I here enable the possibility to use `pytest-xdist` to run the test suite on multiple cores.

For this, the test discovery has to be deterministic, for which I needed to fix the tests using `get_known_camera_names`, as that list was randomly ordered. I directly went for the solution discussed here #1728, which also fixes #1641.

Second, I also had to fix the usage of temporary paths and fixtures in the hdf5 tests, so they do not rely on the order of the tests and instead use the temp path fixtures correctly.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -54,7 +54,7 @@ jobs:
             source $CONDA/etc/profile.d/conda.sh
             conda activate ci;
           fi
-          pytest --cov --cov-report=xml -n 4 --dist loadscope
+          pytest --cov --cov-report=xml -n auto --dist loadscope
           ctapipe-info --version
 
       - uses: codecov/codecov-action@v1
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -37,23 +37,15 @@ jobs:
           pip freeze
 
       - name: Static codechecks
-        env:
-          INSTALL_METHOD: ${{ matrix.install-method }}
+        # need to use a login shell for the conda setup to work
+        shell: bash -leo pipefail {0}
         run: |
-          if [[ "$INSTALL_METHOD" == "conda" ]]; then
-            source $CONDA/etc/profile.d/conda.sh
-            conda activate ci;
-          fi
           pyflakes ctapipe
 
       - name: Tests
-        env:
-          INSTALL_METHOD: ${{ matrix.install-method }}
+        # need to use a login shell for the conda setup to work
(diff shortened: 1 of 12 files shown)


kosack (dismissed):
great!  I tried this at one point but never got it to work., probably for the reasons mentioned.

maxnoe:
In my local tests, with using 4 cores on a machine that has 4 real cores, test time goes from 5:15 to 2:00 minutes using
```
pytest -n 4 --dist loadscope ctapipe -s  
```

maxnoe:
For the pip run on python 3.7 it went from 12 minutes to 8:45.

maxnoe:
And the conda build from 14 to 12 minutes.

LukasNickel (dismissed):

.github/workflows/ci.yml
```diff
@@ -53,7 +54,7 @@ jobs:
             source $CONDA/etc/profile.d/conda.sh
             conda activate ci;
           fi
-          pytest --cov --cov-report=xml
+          pytest --cov --cov-report=xml -n 4 --dist loadscope
```
kosack, line 57: is `n=4` chosen for a particular reason? Would `n=auto` be better?   If it's for a specific reason, maybe add a comment as to why.

maxnoe: github actions provides two physical cores, aince our tests also do some IO i used more threads than available cores, but I havent measured it.

fine with using auto

kosack (approved):

maxnoe:
I now also switched to using mamba to create the env, which shaved another 2 minutes off of the conda build. So we are now down to 8 minutes for the pip build and 7-9 minutes per build from 12-16 minutes before.

LukasNickel (approved):

kosack (approved):

PR #1741, opened 2021-06-02
Improve dl1 fixtures and fix test concurrency issues

maxnoe:


kosack (approved):
Thanks! Works great locally now! much faster

ctapipe/conftest.py
```diff
@@ -135,3 +136,117 @@ def prod5_proton_simtel_path():
...
+        infile = get_dataset_path("gamma_test_large.simtel.gz")
+
+        argv = [
+            f"--input={infile}",
+            f"--output={output}",
+            "--write-images",
+            "--max-events=20",
+            "--allowed-tels=[1,2,3]",
+        ]
+        assert run_tool(Stage1Tool(), argv=argv, cwd=dl1_tmp_path) == 0
```
nbiederbeck, line 170: Does this create the output file *and* check if it ran correctly? I am recently reading up on coding best practices and this seems like "hidden side effects" (although in this case this side effect is wanted) (I read this line as assert no error, side effect create the file). Maybe split this up into two lines? What are your thoughts on this?

maxnoe: This is not "hidden side effects", this is error handling with return codes. Which is not really pythonic, but this here is about the semantics of running command line tools, not really about python.

nbiederbeck (approved):

PR #1748, opened 2021-06-04
Add no-flags, add short aliases, make aliases more consistent

### LINKED ISSUE #1738: Stage1 tool's `--write-parameters` flag is superflous
The stage1 tool has a flag `-write-paramters` that's not useful since it can only set `DataWriter.write_parameters` to True, which already is the default.

Flags are only useful if they set values other than the default.

There is also currently no nice way to deactivate writing parameters, so the flag should probably be `--no-parameters` instead.

We could also follow a common pattern and provide pairs of flags for everything, so I would propose to change

`--write-images`, `--write-paramters` to `--images`, `--no-images`, `--parameters`, `--no-parameters`.

nbiederbeck:
Fixes #1738

### DIFF
--- a/ctapipe/tools/display_dl1.py
+++ b/ctapipe/tools/display_dl1.py
@@ -4,7 +4,7 @@ Calibrate dl0 data to dl1, and plot the photoelectron images.
 from copy import copy
 from matplotlib import pyplot as plt
 from matplotlib.backends.backend_pdf import PdfPages
-from traitlets import Bool, Dict, Int
+from traitlets import Bool, Int
 
 from ctapipe.calib import CameraCalibrator
 from ctapipe.core import Component, Tool
--- a/ctapipe/tools/display_dl1.py
+++ b/ctapipe/tools/display_dl1.py
@@ -17,7 +17,7 @@ from ctapipe.visualization import CameraDisplay
 
 
 class ImagePlotter(Component):
-    """ Plotter for camera images """
+    """Plotter for camera images"""
 
     display = Bool(
         True, help="Display the photoelectron images on-screen as they are produced."
@@ -136,22 +136,19 @@ class DisplayDL1Calib(Tool):
         help="Telescope to view. Set to None to display all telescopes.",
     ).tag(config=True)
 
-    aliases = Dict(
-        dict(
-            input="EventSource.input_url",
-            max_events="EventSource.max_events",
-            T="DisplayDL1Calib.telescope",
-            O="ImagePlotter.output_path",
+    aliases = {
+        ("i", "input"): "EventSource.input_url",
+        ("m", "max-events"): "EventSource.max_events",
+        "T": "DisplayDL1Calib.telescope",
+        ("o", "output"): "ImagePlotter.output_path",
+    }
+    flags = {
+        "D": (
+            {"ImagePlotter": {"display": True}},
+            "Display the photo-electron images on-screen as they are produced.",
         )
-    )
-    flags = Dict(
-        dict(
-            D=(
-                {"ImagePlotter": {"display": True}},
-                "Display the photo-electron images on-screen as they are produced.",
-            )
-        )
-    )
+    }
+
     classes = [EventSource, ImagePlotter] + traits.classes_with_traits(ImageExtractor)
 
     def __init__(self, **kwargs):
--- a/ctapipe/tools/display_dl1.py
+++ b/ctapipe/tools/display_dl1.py
@@ -139,17 +139,16 @@ class DisplayDL1Calib(Tool):
     aliases = {
         ("i", "input"): "EventSource.input_url",
         ("m", "max-events"): "EventSource.max_events",
-        "T": "DisplayDL1Calib.telescope",
+        ("t", "telescope"): "DisplayDL1Calib.telescope",
         ("o", "output"): "ImagePlotter.output_path",
     }
-    flags = {
-        "D": (
-            {"ImagePlotter": {"display": True}},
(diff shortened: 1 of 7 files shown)


ctapipe/tools/display_dl1.py
```diff
@@ -136,22 +136,19 @@ class DisplayDL1Calib(Tool):
...
-    aliases = Dict(
-        dict(
-            input="EventSource.input_url",
-            max_events="EventSource.max_events",
-            T="DisplayDL1Calib.telescope",
-            O="ImagePlotter.output_path",
+    aliases = {
+        ("i", "input"): "EventSource.input_url",
+        ("m", "max-events"): "EventSource.max_events",
+        "T": "DisplayDL1Calib.telescope",
```
maxnoe, line 142:
```suggestion
        ("t", "telescope"): "DisplayDL1Calib.telescope",
```

ctapipe/tools/display_integrator.py
```diff
@@ -226,24 +226,19 @@ class DisplayIntegrator(Tool):
...
-        dict(
-            f="EventSource.input_url",
-            max_events="EventSource.max_events",
-            E="DisplayIntegrator.event_index",
-            T="DisplayIntegrator.telescope",
-            C="DisplayIntegrator.channel",
+    aliases = {
+        ("i", "input"): "EventSource.input_url",
+        ("m", "max-events"): "EventSource.max_events",
+        "E": "DisplayIntegrator.event_index",
```
maxnoe, line 232: e / event-index, t / telescope / C / channel (I think `c` is config, right`?

ctapipe/tools/display_dl1.py
nbiederbeck: Ahh yes, I wanted to ask if this would be a good change

ctapipe/tools/display_integrator.py
```diff
@@ -226,24 +226,19 @@ class DisplayIntegrator(Tool):
...
+        "T": "DisplayIntegrator.telescope",
+        "C": "DisplayIntegrator.channel",
+    }
+    flags = {
+        "id": (
+            {"DisplayDL1Calib": {"use_event_index": True}},
```
maxnoe, line 238: This should also be a flag, right?

ctapipe/tools/display_integrator.py
nbiederbeck: Will change it as proposed, yes `c` is config

ctapipe/tools/dl1_merge.py
```diff
@@ -166,6 +149,34 @@ class MergeTool(Tool):
...
+            "Don't skip DL1/Event/Telescope and Simulation/Event/Telescope parameters in output",
+        ),
+        flag(
+            "skip-broken-files",
+            "MergeTool.skip_broken_files",
+            "Skip broken files instead of raising an error",
+            "Don't skip broken files instead of raising an error",
+        ),
+    ]
+    for f in _flags:
```
maxnoe, line 178: Why this for loop?

ctapipe/tools/display_integrator.py
nbiederbeck: Was not sure about that one, but yes, why not

maxnoe (changes requested):
The for loop looks strange in the class body,

it should be possible to just define flags like this, right?

```
flags = {
    **flag(...),
    **flag(...),
   ...
}
```

Also, there are some boolean flags remaining, that are not setup using `flag`, can you transform those as well?

nbiederbeck:
I thought the loop is the most elegant solution because I forgot about `**`

nbiederbeck:
do we want a --no-progress flag?

maxnoe:
Yes, I think that makes sense. In case we switch the default for example. Having all boolean flags consistent is a big plus I'd say.

nbiederbeck:
That leads me to another question: the should the `(-f, --overwrite)` flags (and e.g. `(-q, --quiet)`) also become negated? But only the long forms, I suppose? (e.g. no `--no-f`)

maxnoe:
shortform negation doesn't make sense, correct.  How does `flag` do shortforms?

nbiederbeck:
afaik it doesn't. It can't handle short/long tuples, only strings. So I think we need to add a long yes/no flag and the short form we have to build ourselves

maxnoe:
It doesn't. And it also does not make sure that only one of the flags is given.

nbiederbeck:
What do you mean exactly? When I give `--flag --no-flag` it takes the last one. When I give `-l --long` it doesn't really matter

maxnoe:
I somehow expected that traitlets would raise an error for `--flag --no-flag`

ctapipe/tools/tests/test_tools.py
```diff
@@ -221,7 +221,7 @@ def test_display_integrator(tmp_path):
     assert (
         run_tool(
             DisplayIntegrator(),
-            argv=shlex.split(f"--f={GAMMA_TEST_LARGE} " "--max_events=1 "),
+            argv=shlex.split(f"--input={GAMMA_TEST_LARGE} " "--max_events=1 "),
```
maxnoe, line 224: Could you just give the arguments as list? `[f"--input={GAMMA_TEST_LARGE}", "--max_events=1"]`? 

using shlex.split here is not what we want I think. We know, what are the single arguments that need to be quoted together.

nbiederbeck: for all occurences in that file?

maxnoe (approved):

kosack (approved):

Issue #1744, opened 2021-06-03
Things to save from protopipe before the full-refactoring


HealthyPear:
We are near to a complete refactoring of the pipeline based on ctapipe, but if we did it now some things would disappear and that is not good (both from the point-of-view of comparisons with historical pipelines and for any future study).

These features can and should be generalized to account for more use-cases than the ones taken into account currently by protopipe.

This issue is a summary list: I will add new things as I stumble upon them and each related PR should be linked to the respective feature.

- [x] **Image extraction status** (#1810)

This is the equivalent of any other `QualityQuery`-based class we use (in `ImageProcessor` and `ShowerProcessor`) but for image extractors.
It is needed by extractors like `TwoPassWindowSum` (CTAMARS-based analysis), for which it is fundamental to label some events (in this case images, not showers) which do not survive a step during the total image extraction process (in this case a preliminary image cleaning cut in between the two passes).

This will require to modify the `ImageExtractor` class to output a quality query together with charges and peak times.
Of course, for single-pass image extractors, this condition will should be always positive by default and thus it would have no visible effect.

- [x] **Custom trigger conditions**

In the specific case of _protopipe_ I have implemented the LST-stereo trigger condition, required for Prod3b simulations in which this requirement was not simulated

- [x] **calibscale** (#1749)

This should be a simple scale factor for the reconstructed charge extracted from each pixel.
In particular, as reported [here](https://forge.in2p3.fr/projects/step-by-step-reference-mars-analysis/wiki#From-simtelarray-to-DL1a),

> the factor to transform the integrated charges (in ADC counts) into number of photoelectrons is the pixel-wise one calculated by simtelarray (0.92*LasCalData->calib[][][]).

- [x] **Shower core position in TiltedFrame** (#1745)

This allows the pipeline to read from DL2 files this quantity directly in the right frame were telescope-wise impact distances should be computed

- [ ] **Wavelet cleaning**

It has not been used in a long time and it should be added as an `ImageCleaner`, perhaps from a module itself.

- [x] **DL2 event estimation**

This is something that should be added after PR #1726.

There should a function that takes as input 
- what in protopipe I call `TRAINING` data (aka DL1b + shower geometry)
- information from the configuration of the model building process, in particular which features have been used
- the possibility to eval at runtime features that do not come directly from the `TRAINING` file, but that are functions of those (e.g. `np.log10(hillas_intensity)`)
[...]

PR #1749, opened 2021-06-04
Add support for calibration scale and shift in R1

HealthyPear:
This PR adds the possibility to both scale and shift the R1 waveform samples in photoelectrons from `SimtelEventSource`.

The scale factor (called `calib_scale` in simtel and == 0.92 in both HESS and CTAMARS analyses - possibly depends on prod also) is on top of the already present DC_to_PHE factor, which is now explicit.
This is also reported in #1744 .
@moralejo could you check if this corresponds to what you apply in your analysis?

The shift can be useful for simulating a mis-calibration (for example to see how far calibration can drift before we need new IRFs)

### DIFF
--- a/ctapipe/io/tests/test_simteleventsource.py
+++ b/ctapipe/io/tests/test_simteleventsource.py
@@ -12,6 +12,7 @@ from traitlets.config import Config
 
 
 from ctapipe.calib.camera.gainselection import ThresholdGainSelector
+from ctapipe.calib.camera.calibrator import CameraCalibrator
 from ctapipe.io.simteleventsource import SimTelEventSource, apply_simtel_r1_calibration
 from ctapipe.utils import get_dataset_path
 from ctapipe.io import DataLevel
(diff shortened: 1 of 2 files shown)


ctapipe/io/tests/test_simteleventsource.py
```diff
@@ -331,3 +337,43 @@ def test_only_config():
...
+    telid = 25
+
+    with SimTelEventSource(
+        input_url=prod5_gamma_simtel_path, max_events=1
+    ) as source:
+
+        for event in source:
+            pass
+
+    config = Config()
```
maxnoe, line 353: We test the config system elsewhere. For this unit test, it would be much easier to just give the calib scale / calib shift as kwarg to the simtel event source.

HealthyPear: done

kosack (approved):

kosack (approved):

maxnoe (approved):

PR #1752, opened 2021-06-25
include stage1_config.json in package

### LINKED ISSUE #1750: Relative path to config file in tests
The tool tests have the following line:
```python
config = Path("./examples/stage1_config.json").absolute()
```

This failed me on `pytest` *inside* of `ctapipe/tools` instead of `pytest ctapipe/tools`. 

Fix needed or should I just run local test the correct (second) way?

nbiederbeck:
I included the `stage1_config.json` in `ctapipe/examples`, so the tests can use the packaged one. However, I am not sure about the naming of the directory.

This will fix #1750


maxnoe (changes requested):
This is great, thanks.

Two points:

a) I would probably just add the json file into a `resources` directory in `ctapipe/tools/tests/resources`. It is in this case not really an example, but a non-code file needed for the tests.
b) wee need to add `importlib_resources` to the required packages for python < 3.9.

nbiederbeck:
> a) I would probably just add the json file into a `resources` directory in `ctapipe/tools/tests/resources`. It is in this case not really an example, but a non-code file needed for the tests.

Yes, that is what I meant. Not really an example in this case anymore. But should we keep this in the top-level `examples`? So that it is twice in the repo, but once in the package?

> b) wee need to add `importlib_resources` to the required packages for python < 3.9.

Is it possible to require this *only* if python < 3.9 ? I just put it into `tests_require` for now

nbiederbeck:
> Is it possible to require this only if python < 3.9 ? I just put it into tests_require for now

for future reference: yes you can, with [environment markers from pep 508](https://www.python.org/dev/peps/pep-0508/#environment-markers)

setup.py
```diff
@@ -23,7 +23,7 @@
     "ctapipe-stage1 = ctapipe.tools.stage1:main",
     "ctapipe-merge = ctapipe.tools.dl1_merge:main",
 ]
-tests_require = ["pytest", "pandas>=0.24.0"]
+tests_require = ["pytest", "pandas>=0.24.0", "importlib_resources"]
```
maxnoe, line 26: Only for `python < 3.9`, there is a special syntax for this: `importlib_resources;python_version<'3.9'`

nbiederbeck: thanks, done

ctapipe/tools/tests/test_tools.py
```diff
@@ -131,7 +131,7 @@ def _generator(self):
         f.write(b"dummy")
         f.flush()
 
-    config = files("ctapipe.examples").joinpath("stage1_config.json")
+    config = files("ctapipe.tools.tests.resources").joinpath("stage1_config.json")
```
maxnoe, line 134: Do we have `zip_safe=False` in setup.py? Otherwise you should use https://docs.python.org/3/library/importlib.html#importlib.resources.as_file



```
resources = files("ctapipe.tools.tests.resources")
with as_file(resources.joinpath("stage1_config.json")) as config_path:
    ...
```

nbiederbeck: we do: https://github.com/cta-observatory/ctapipe/blob/15cd535e1025352de8477240cd3f8b5da1038147/setup.py#L82

maxnoe (changes requested):
I needed to add "resouces/*" to package data to make setuptools include a json file in `ctapipe/tools/tests/resources/` to the wheels / source dists:

```
 81 ⋮ 81 │    ],
 82 ⋮ 82 │    zip_safe=False,
 83 ⋮ 83 │    entry_points=entry_points,
 84 ⋮    │    package_data={"": ["tools/bokeh/*.yaml", "tools/bokeh/templates/*.html"]},
    ⋮ 84 │    package_data={
    ⋮ 85 │        "": [
    ⋮ 86 │            "tools/bokeh/*.yaml",
    ⋮ 87 │            "tools/bokeh/templates/*.html",
    ⋮ 88 │            "resources/*",
    ⋮ 89 │        ]
    ⋮ 90 │    },
 85 ⋮ 91 │)
```

nbiederbeck:
I can reproduce that issue, after cleaning up (`rm -rf build *.egg-info`)

weird, that the tests still pass ..

kosack (approved):

maxnoe (dismissed):

maxnoe:
Maybe it would be worth to use this setup in the CI to make sure it works.

So I'd suggest to remove the ` -e` in the pip install and use `pytest --pyargs ctapipe` to run the tests.

kosack:
It may be worth waiting for #1726 since that introduces several more "standard" config files that need to be used for testing.  I only haven't had time to write the tests.   

 Of course, we eventually need to create a configuration repo in the CTA-GitLab to keep track of real (non-testing) configurations.

maxnoe:
Wouldn't it make sense to do it the other way around? Merge this here first and then you can just add those new config files to the `ctapipe/tools/tests/resources` folder in the other PR?

kosack (dismissed):

nbiederbeck:
ouch, it seems like I need to review how to do git

nbiederbeck:
sorry if anyone was notified unnecessarily.

I now added the resources for the ProcessorTool and resolved the conflicts that were shown

maxnoe (approved):

kosack:
The only issue I have here is that these configs really are examples, that the user may want to take and modify, not just used for testing.  They are supposed to define the reference analysis. So perhaps we should keep two copies of them? One for tests, and one for examples? Otherwise they are now hard to find.

Or perhaps better yet: we could make a `Tool` that sets up a directory for user analysis (`ctapipe-quickstart` or something), that writes a copy of the standard configs to a user-specified directory with a README or something.  It could even fetch some example files.  Essentially, it would set up a working directory with templates for the configs.  For now, it could be simple, but later it could have some more options.    There is also `ProcessorTool().generate_config_file` (a feature from `trailets.config` that generates a commented config file, but only in Python format)

kosack:
By the way, could we not just put these configs on the *datatserver* in the latest version of *ctapipe-extra* like we do with all other testing files? Thatt would make this PR unneeded, since you could just use `get_dataset_path("stage1_config.json")` ?

maxnoe:
> By the way, could we not just put these configs on the datatserver in the latest version of ctapipe-extra like we do with all other testing files?

I guess that these files will need updates more often than data files. Which makes putting them on the data server really suboptimal, since only very limited number of people have access to them and somebody preparing a PR that needs configs to be adjusted cannot do it while developing. Also, these are small text files, so are well suited for git.

kosack (approved):
Ok let's merge this then, and I'll open a PR for a tool to extract them for users,.

PR #1757, opened 2021-07-22
Configurable contact information for DataWriter

maxnoe:
This makes the Contact info a configurable trait of the `DataWriter` so that users can actually set the proper contact info when running tools.

We probably also want to enable this for other metadata fields, but for the others it was not as clear to me which should be autofilled and which should rely on user input.


kosack (approved):
The only other thing that we could consider is the PROCESS ID and PROCESS SUBTYPE - that is supposed to describe the overall "parent" activity, so the ID could be the MC prod name (which currently is not in the simtel files, so can't be filled automatically), and subtype is not fully defined yet.

LukasNickel (approved):

PR #1758, opened 2021-07-22
Fix docs for Path

maxnoe:


mexanick:
Since there's already open PR considering Path trait, let me ask a question here:
from the source code it appears that it is not possible to set the attribute `allow_none` to` False` and the validation of empty path with attribute `exists` set to `True` doesn't make sense.
Reasons:
https://github.com/cta-observatory/ctapipe/blob/37d901581c478bdfcd56c699feba98588c371046/ctapipe/core/traits.py#L117
prevents setting the `allow_none` attribute:
TypeError: __init__() got multiple values for keyword argument 'allow_none'

and 
https://github.com/cta-observatory/ctapipe/blob/37d901581c478bdfcd56c699feba98588c371046/ctapipe/core/traits.py#L147-L151

will return always `None` if the default value is not set.

Is such behavior is expected? I'd change the hardcoded `allow_none` to a default value with a possibility to override...

kosack (approved):

maxnoe:
@mexanick I had problems making `allow_none=False` work. Let me have a look again, but in the mean time, I found a "nice" solution using the `@default` decorator from traitlets:

```
from traitlets import default, TraitError
from ctapipe.core.traits import Path

class Foo(Component):
    path = Path().tag(config=True)

    @default("path")
    def path_default(self):
        raise TraitError("path is required")

f = Foo()
print(f.path)
```

HealthyPear (approved):

PR #1760, opened 2021-07-23
Implement allow_none=False for Path trait, fixes #1759

### LINKED ISSUE #1759: `Path` trait does not allow to set `allow_none` attribute to `False`
From the source code it appears that it is not possible to set the attribute `allow_none` to` False` and the validation of empty path with attribute `exists` set to `True` doesn't make sense.
Reasons:
https://github.com/cta-observatory/ctapipe/blob/37d901581c478bdfcd56c699feba98588c371046/ctapipe/core/traits.py#L117
prevents setting the `allow_none` attribute:
TypeError: __init__() got multiple values for keyword argument 'allow_none'

and 
https://github.com/cta-observatory/ctapipe/blob/37d901581c478bdfcd56c699feba98588c371046/ctapipe/core/traits.py#L147-L151

will return always `None` if the default value is not set.

Is such behavior is expected? I'd change the hardcoded `allow_none` to a default value with a possibility to override...

_Originally posted by @mexanick in https://github.com/cta-observatory/ctapipe/issues/1758#issuecomment-885000278_

Original answer from @maxnoe

https://github.com/cta-observatory/ctapipe/pull/1758#issuecomment-885029874

maxnoe:
This properly implements the default / validation / allow_none part of the traitlets API for our `Path` traitlet.

### DIFF
--- a/ctapipe/core/component.py
+++ b/ctapipe/core/component.py
@@ -141,6 +141,15 @@ class Component(Configurable, metaclass=AbstractConfigurableMeta):
                 " and not `config`"
             )
 
+        # set up logging (for some reason the logger registered by LoggingConfig
+        # doesn't use a child logger of the parent by default)
+        if self.parent:
+            self.log = self.parent.log.getChild(self.__class__.__name__)
+        else:
+            self.log = getLogger(
+                self.__class__.__module__ + "." + self.__class__.__name__
+            )
+
         # Transform warning about wrong traitlets in the config to an error
         # Only works for Components, unfortunately not for Tools, since
         # Tools use `log.warning` instead of `warnings.warn`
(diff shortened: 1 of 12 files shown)


kosack (dismissed):

kosack (approved):

nbiederbeck (dismissed):

maxnoe:
The `repr_html` of Tool is broken after the tool has been run, due to the subcomponents being setup. 

This is also broken in master, however for some reason the notebook only prints the error in `repr_html`, (I guess jupyter ignores exceptions in repr functions).

For now, I fixed it by just skipping the subcompents in the repr, but maybe we want to do a follow up also including subcomponents in the repr. However, I don't really see a big usecase for a html repr of a `Tool`

nbiederbeck (approved):

kosack (approved):

PR #1766, opened 2021-07-26
Remove cutflow

maxnoe:


kosack:
It's still used heavily in protopipe, so we should keep it a while longer (but perhaps mark it as deprecated)

HealthyPear:
> It's still used heavily in protopipe, so we should keep it a while longer (but perhaps mark it as deprecated). In a not too distant future version of protopipe, however, that code will be totally replaced by `ctapipe-process`, so after protopipe supports that we can remove CutFlow

Yes it is still used.
I see anyway the deprecation notice, so I don't think we should add anything more.

If it is not a problem to keep it (it is not incompatible with current data, isn't it?) we could make this a draft PR, remove the milestone, and as soon as protopipe is using `ctapipe-process` we can reopen it and merge.

kosack (dismissed):

nbiederbeck (approved):

maxnoe:
@kosack You approved this now, but the situation hasn't really changed, has it?

What is the status of protopipe concerning cutflow? If we remove this here, this means either

a) protopipe has to replace cutflow with `QualityQuery` when upgrading to 0.16
b) protopipe needs to include `cutflow` back itself
c) protopipe cannot upgrade to 0.16

kosack (approved):

PR #1767, opened 2021-07-27
WIP: Machine learning module

maxnoe:
Starting to work on the machine learning module for event reconstruction, draft PR to get early feedback.

This will probably get pretty large, so I will try to pick small parts out for individual PRs to make it easier to review.

This will include code  and ideas taken from / adapted from aict-tools and protopipe mva


ctapipe/ml/sklearn.py
```diff
@@ -0,0 +1,137 @@
...
+        self.model = SUPPORTED_MODELS[self.model_cls](**self.model_config)
+
+    def table_to_X(self, table):
+        feature_table = table[self.features]
+        valid = check_valid_rows(feature_table, log=self.log)
+        X = table_to_float(feature_table[valid])
+        return X, valid
+
+    def table_to_y(self, table):
+        return np.array(table[self.target])
```
nbiederbeck, line 59: My gut tells me there should be a dtype here as is done in `table_to_float`. But thinking a little about it that needs to default to float32 for regressors and int for classifiers? Or should the user make sure the correct dtype is supplied? Or is sklearn smart enough that the dtype needs only to be set for `X`? Also, while typing this, I wonder about categorical variables.. but this is maybe not needed for ctapipe? Sorry for this unstructured braindump

maxnoe: It needs to be int for a classifier and float something for a regressor. The dtypes are actually different from model to model depending on the underlying implementation.

nbiederbeck: So we cannot "guess" here and the user has to make sure it works for their usecase, correct?

maxnoe: I think it is safe for now to assume that it is a suitable dtype in the table, yes

ctapipe/ml/tests/test_preprocessing.py
```diff
@@ -0,0 +1,64 @@
...
+
+    t = Table({"a": [1.0, 1e50, np.inf, -np.inf, np.nan], "b": [1, 2, 3, 4, 5]})
+
+    fmax = np.finfo(np.float32).max
+    fmin = np.finfo(np.float32).min
+    expected = np.array(
+        [[1.0, fmax, fmax, fmin, np.nan], [1.0, 2.0, 3.0, 4.0, 5.0]], dtype=np.float32
+    ).T
+
+    array = table_to_float(t)
```
LukasNickel, line 18: Maybe add a comment, that `table_to_float` defaults to `float32`?

ctapipe/ml/tests/test_sklearn.py
```diff
@@ -0,0 +1,139 @@
+import pytest
+
+import numpy as np
+from traitlets.config import Config
+from traitlets import TraitError
+from astropy.table import Table
+from numpy.testing import assert_array_equal
+
+
+def test_supported_regressors():
```
LukasNickel, line 10: If the whitelist approach is to be used for the list of supported models, there should be a test that an unsupported model is not in the `SUPPORTED_REGRESSORS`, e.g. the DummyRegressor

ctapipe/ml/tests/test_sklearn.py
```diff
@@ -0,0 +1,139 @@
...
+    from ctapipe.ml.sklearn import SUPPORTED_REGRESSORS
+    from sklearn.ensemble import RandomForestRegressor
+
+    assert "RandomForestRegressor" in SUPPORTED_REGRESSORS
+    assert SUPPORTED_REGRESSORS["RandomForestRegressor"] is RandomForestRegressor
+
+
+def test_supported_classifiers():
```
LukasNickel, line 18: same here

ctapipe/ml/tests/test_sklearn.py
```diff
@@ -0,0 +1,139 @@
...
+    config = Config(
+        {
+            "Classifier": {
+                "model_cls": "RandomForestClassifier",
+                "model_config": {"n_estimators": 10, "max_depth": 15},
+            }
+        }
+    )
+
+    # should create class with sklearn defaults
```
LukasNickel, line 70: copy/paste.
This is with the config above

ctapipe/ml/tests/test_sklearn.py
```diff
@@ -0,0 +1,139 @@
...
+    from sklearn.datasets import make_blobs
+
+    X, y = make_regression(n_samples=100, n_features=5, n_informative=3, random_state=0)
+    t = Table({f"X{i}": col for i, col in enumerate(X.T)})
+    t["energy"] = y
+    t["X0"][10] = np.nan
+    t["X1"][30] = np.nan
+
+    X, y = make_blobs(n_samples=100, n_features=3, centers=2, random_state=0)
+    for i, col in enumerate(X.T, start=5):
```
LukasNickel, line 38: What exactly is this doing?
It partially overwrites the regression table with blobs to avoid having two tables for the unit tests?

maxnoe: Not partially, the first table has 5 features, we start at 5 here. I create a single table that as features and targets both suitable for regression and classification (like our real data)

LukasNickel: Thx got confused with the transposition

ctapipe/ml/sklearn.py
```diff
@@ -0,0 +1,137 @@
...
+    model_cls = Enum(
+        SUPPORTED_REGRESSORS.keys(), default_value=None, allow_none=False
+    ).tag(config=True)
+
+
+class Classifier(Model):
+    model_cls = Enum(
+        SUPPORTED_CLASSIFIERS.keys(), default_value=None, allow_none=False
+    ).tag(config=True)
+    invalid_class = Integer(-1).tag(config=True)
```
kosack, line 106: would be useful to have a help="xx" string here to explain what this is (also for other parameters)

kosack:
I know it's only a draft, but I put a few minor comments above.  Overall it's nice and easy to use so far!

ctapipe/ml/preprocessing.py
```diff
@@ -0,0 +1,41 @@
...
+    """Convert a table to a float32 array, replacing inf/-inf with float min/max"""
+    X = structured_to_unstructured(table.as_array(), dtype=dtype)
+    np.nan_to_num(X, nan=np.nan, copy=False)
+    return X
+
+
+def check_valid_rows(table: Table, warn=True, log=log) -> np.ndarray:
+    """Check for nans, returning a mask of the rows not containing any nans"""
+
+    nans = np.array([np.isnan(col) for col in table.columns.values()]).T
```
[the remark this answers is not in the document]
maxnoe: Infs are replaced with float min/max in the other preprocessing functions.

ctapipe/ml/sklearn.py
```diff
@@ -0,0 +1,137 @@
...
+import numpy as np
+from traitlets import Dict, List, Unicode, Enum, Integer
+from sklearn.base import is_classifier, is_regressor
+import joblib
+
+from ..core import Component
+from .preprocessing import check_valid_rows, table_to_float
+
+
+sklearn_modules = {
```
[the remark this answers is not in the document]
maxnoe: Didn't know about `all_estimators`, that'll do

kosack:
One feature that I think is very useful is the ability to make features on-the fly that are composed of existing features.  That allows us to also apply transformations to existing features (`log10(energy)`, etc).  Do you have an idea of how you will implement that? In protopipe, @HealthyPear used Pandas's `DataFrame.eval()` to generate new columns on the fly, but that also required making single-row DataFrames in the event-wise predict step, which is maybe a bit awkward. 

```py3
 dataframe.eval(f"{feature_name} = {feature_expression}", inplace=True)
```

maxnoe:
>  Do you have an idea of how you will implement that?

Yes, this is easy using `numexpr`, same as pandas eval

ctapipe/ml/tools/train_energy_regressor.py
```diff
@@ -0,0 +1,122 @@
...
+            random_state=self.rng.integers(0, 2**31 - 1),
+        )
+
+        for (train_indices, test_indices) in tqdm(kfold.split(table), total=n_cv):
+            train = table[train_indices]
+            test = table[test_indices]
+
+            self.model.fit(train)
+            prediction, _ = self.model.predict(test)
+            scores.append(metrics.r2_score(test["true_energy"], prediction))
```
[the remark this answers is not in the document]
maxnoe: But mean accuracy is not a great score... better have roc auc

nbiederbeck:
I'm trying to rebase this branch, which yielded the following uncaught problems:
1. `len(events) < n_cross_val`: we could skip cross validation or fail gracefully
2. no signal or background events (but the other), sklearn will not fit a model due to not being able to calculate ROC AUC with only 1 class in `y_true`

maxnoe:
> len(events) < n_cross_val: we could skip cross validation or fail gracefully

I guess we should just error out here, this is never a situation we'd actually want to be in.

ctapipe/ml/apply.py
```diff
@@ -294,8 +294,10 @@ def __init__(self, model_component, **kwargs):
         self.rng = np.random.default_rng(self.rng_seed)
         if isinstance(self.model_component, ClassificationReconstructor):
             self.calculate_metrics = self._cross_validate_classification
+            self.split_data = StratifiedKFold
```
maxnoe, line 297: StratifiedKFold needs labels, where are those passed?

nbiederbeck: Fixed via #2054

ctapipe/ml/apply.py
```diff
@@ -0,0 +1,381 @@
...
+
+
+class CrossValidator(Component):
+    n_cross_validations = Int(5).tag(config=True)
+    output_path = Path(
+        default_value=None,
+        allow_none=True,
+        directory_ok=False,
+    ).tag(config=True)
+    rng_seed = Int(default_value=1337, help="Seed for the random number generator").tag(
```
nbiederbeck, line 286: The other ML components call this `random_seed`

ctapipe/ml/stereo_combination.py
```diff
@@ -0,0 +1,250 @@
...
+        """
+
+
+class StereoMeanCombiner(StereoCombiner):
+    """
+    Calculates array-wide (stereo) predictions as the mean of
+    the reconstruction on telescope-level with an optional weighting.
+    """
+
+    weights = CaselessStrEnum(
```
nbiederbeck, line 68: `weights` sound like it is an array of weights, I would prefer e.g. `weighting`

ctapipe/ml/tools/apply.py
```diff
@@ -0,0 +1,216 @@
...
+
+    stereo_combiner_type = create_class_enum_trait(
+        base_class=StereoCombiner,
+        default_value="StereoMeanCombiner",
+    ).tag(config=True)
+
+    aliases = {
+        ("i", "input"): "Apply.input_url",
+        "regressor": "Apply.energy_regressor_path",
+        "classifier": "Apply.particle_classifier_path",
```
nbiederbeck, line 62: With Disp coming to ML, we should change this to energy-regressor and particle-classifier

ctapipe/ml/tools/apply.py
```diff
@@ -0,0 +1,216 @@
...
+
+            predictions = reconstructor.predict(tel, table)
+            table = hstack(
+                [table, predictions],
+                join_type="exact",
+                metadata_conflicts="ignore",
+            )
+            write_table(
+                table[["obs_id", "event_id", "tel_id"] + predictions.colnames],
+                self.loader.input_url,
```
nbiederbeck, line 177: Although this is correct, we might prefer `self.output_path`

ctapipe/ml/tools/apply.py
```diff
@@ -0,0 +1,216 @@
...
+        trafo = TelListToMaskTransform(self.loader.subarray)
+        for c in filter(
+            lambda c: c.name.endswith("tel_ids"),
+            stereo_predictions.columns.values(),
+        ):
+            stereo_predictions[c.name] = np.array([trafo(r) for r in c])
+
+        write_table(
+            stereo_predictions,
+            self.loader.input_url,
```
nbiederbeck, line 201: Although this is correct, we might prefer `self.output_path`

ctapipe/ml/tools/tests/test_process.py
```diff
@@ -0,0 +1,98 @@
...
+        f"--output={output}",
+        "--write-images",
+        "--write-showers",
+        f"--energy-regressor={energy_regressor_path}",
+        f"--config={config_path}",
+    ]
+    assert run_tool(ProcessorTool(), argv=argv, cwd=tmp_path, raises=True) == 0
+
+    print(read_table(output, "/dl2/event/telescope/energy/ExtraTreesRegressor/tel_004"))
+    print(read_table(output, "/dl2/event/subarray/energy/ExtraTreesRegressor"))
```
nbiederbeck, line 50: remove prints, add some assertions

ctapipe/ml/tools/tests/test_process.py
```diff
@@ -0,0 +1,98 @@
...
+        f"--config={config_path}",
+    ]
+    assert run_tool(ProcessorTool(), argv=argv, cwd=tmp_path, raises=True) == 0
+
+    print(
+        read_table(
+            output, "/dl2/event/telescope/classification/ExtraTreesClassifier/tel_004"
+        )
+    )
+    print(read_table(output, "/dl2/event/subarray/classification/ExtraTreesClassifier"))
```
nbiederbeck, line 98: remove prints, add some assertions

ctapipe/ml/tools/tests/test_train.py
```diff
@@ -0,0 +1,96 @@
...
+        argv=[
+            "--input=dataset://gamma_diffuse_dl2_train_small.dl2.h5",
+            f"--output={out_file}",
+            f"--config={config}",
+            "--log-level=INFO",
+            f"--CrossValidator.output_path={energy_cv_out_file}",
+        ],
+    )
+    assert ret == 0
+    assert energy_cv_out_file.exists()
```
nbiederbeck, line 78: Should we assert the contents of this file instead of its mere existence? Maybe we then split this out into a `test_cv.py`

ctapipe/ml/tools/train_energy_regressor.py
```diff
@@ -0,0 +1,110 @@
...
+            idx = self.rng.choice(len(table), n_events, replace=False)
+            idx.sort()
+            table = table[idx]
+
+        return table
+
+    def finish(self):
+        self.log.info("Writing output")
+        self.regressor.write(self.output_path)
+        if self.cross_validate.output_path:
```
nbiederbeck, line 100: With this, we always cross validate, but not always write it's results?

ctapipe/ml/tools/train_particle_classifier.py
```diff
@@ -0,0 +1,160 @@
...
+        directory_ok=False,
+    ).tag(config=True)
+
+    n_signal = Int(default_value=None, allow_none=True).tag(config=True)
+    n_background = Int(default_value=None, allow_none=True).tag(config=True)
+    random_seed = Int(default_value=0).tag(config=True)
+
+    aliases = {
+        ("s", "signal"): "TrainParticleIdClassifier.input_url_signal",
+        ("b", "background"): "TrainParticleIdClassifier.input_url_background",
```
nbiederbeck, line 42: we might leave out the short forms

ctapipe/ml/tools/train_particle_classifier.py
```diff
@@ -0,0 +1,160 @@
...
+            "Train on %s signal and %s background events", len(signal), len(background)
+        )
+        return table
+
+    def finish(self):
+        self.log.info("Writing output")
+        self.classifier.write(self.output_path)
+        self.signal_loader.close()
+        self.background_loader.close()
+        if self.cross_validate.output_path:
```
nbiederbeck, line 151: See above

ctapipe/tools/process.py
```diff
@@ -63,16 +64,36 @@ class ProcessorTool(Tool):
...
 
+    stereo_combiner_configs = List(Dict()).tag(config=True)
+
     aliases = {
         ("i", "input"): "EventSource.input_url",
         ("o", "output"): "DataWriter.output_path",
         ("t", "allowed-tels"): "EventSource.allowed_tels",
         ("m", "max-events"): "EventSource.max_events",
+        ("e", "energy-regressor"): "ProcessorTool.energy_regressor_path",
+        ("particle-classifier"): "ProcessorTool.particle_classifier_path",
```
nbiederbeck, line 96: See above. No need for a short form for the energy regressor, but no short form for the particle classifier (`-p` is occupied)

setup.py
```diff
@@ -19,6 +20,9 @@
     "ctapipe-merge = ctapipe.tools.merge:main",
     "ctapipe-fileinfo = ctapipe.tools.fileinfo:main",
     "ctapipe-quickstart = ctapipe.tools.quickstart:main",
+    "ctapipe-ml-train-regressor = ctapipe.ml.tools.train_energy_regressor:main",
+    "ctapipe-ml-train-classifier = ctapipe.ml.tools.train_particle_classifier:main",
```
nbiederbeck, line 24: train-energy-regressor
train-particle-classifier

ctapipe/ml/conftest.py
```diff
@@ -0,0 +1,49 @@
...
+    return tmp_path_factory.mktemp("models")
+
+
+@pytest.fixture(scope="session")
+def energy_regressor_path(model_tmp_path):
+    from ctapipe.ml.tools.train_energy_regressor import TrainEnergyRegressor
+
+    tool = TrainEnergyRegressor()
+    config = resource_file("ml_config.yaml")
+    out_file = model_tmp_path / "energy.pkl"
```
nbiederbeck, line 17: We should use the `FileLock`, as in other tests

ctapipe/ml/conftest.py
```diff
@@ -0,0 +1,49 @@
...
+    return out_file
+
+
+@pytest.fixture(scope="session")
+def particle_classifier_path(model_tmp_path):
+    from ctapipe.ml.tools.train_particle_classifier import TrainParticleIdClassifier
+
+    tool = TrainParticleIdClassifier()
+    config = resource_file("ml_config.yaml")
+    out_file = model_tmp_path / "particle_classifier.pkl"
```
nbiederbeck, line 37: We should use the `FileLock`, as in other tests

ctapipe/ml/apply.py
```diff
@@ -0,0 +1,381 @@
...
+    def __call__(self, event: ArrayEventContainer) -> None:
+        """
+        Apply the quality query and model and fill the corresponding container
+        """
+        for tel_id in event.trigger.tels_with_trigger:
+            table = self._collect_features(event, tel_id)
+            table = self.generate_features(table)
+            mask = self.qualityquery.get_table_mask(table)
+
+            if mask[0]:
```
LukasNickel, line 188:
```python
        mask = self.qualityquery.get_table_mask(table)[0]
            if mask:
```
is more readable imo

ctapipe/ml/apply.py
```diff
@@ -0,0 +1,381 @@
...
+        # gammas have true_shower_primary_id = 0
+        self.model.positive_class = 0
+
+    def __call__(self, event: ArrayEventContainer) -> None:
+        for tel_id in event.trigger.tels_with_trigger:
+            table = self._collect_features(event, tel_id)
+            table = self.generate_features(table)
+            mask = self.qualityquery.get_table_mask(table)
+
+            if mask[0]:
```
LukasNickel, line 242: Same as above

ctapipe/ml/stereo_combination.py
```diff
@@ -0,0 +1,250 @@
...
+    def predict(self, mono_predictions: Table) -> Table:
+        """
+        Calculates the (array-)event-wise mean.
+        Telescope events, that are nan, get discarded.
+        This means you might end up with less events if
+        all telescope predictions of a shower are invalid.
+        """
+
+        prefix = self.algorithm
+        # TODO: Integrate table quality query once its done
```
LukasNickel, line 182: Is available now

ctapipe/ml/tools/tests/test_process.py
```diff
@@ -0,0 +1,98 @@
...
+):
+    from ctapipe.tools.process import ProcessorTool
+
+    output = tmp_path / "gamma_prod5.dl2_energy.h5"
+
+    config_path = tmp_path / "config.json"
+
+    input_url = prod5_gamma_lapalma_simtel_path
+
+    # la palma alpha config
```
LukasNickel, line 18: Is there a reason we need alpha here?
I think it just adds complexity to the test opposed to all telescopes or a smaller subset (same below)

setup.py
```diff
@@ -19,6 +20,9 @@
     "ctapipe-merge = ctapipe.tools.merge:main",
     "ctapipe-fileinfo = ctapipe.tools.fileinfo:main",
     "ctapipe-quickstart = ctapipe.tools.quickstart:main",
+    "ctapipe-ml-train-regressor = ctapipe.ml.tools.train_energy_regressor:main",
```
LukasNickel, line 23: This should have energy in its name

setup.py
LukasNickel, line 24: This should have classifier, id, particle type or smth in its name

ctapipe/containers.py
```diff
@@ -830,8 +830,8 @@ class ParticleClassificationContainer(Container):
         nan,
         (
             "prediction of the classifier, defined between "
-            "[0,1], where values close to 0 are more "
-            "gamma-like, and values close to 1 more "
+            "[0,1], where values close to 1 are more "
```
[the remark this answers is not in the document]
kosack: I'd go for gammaness, since that is usually the goal (to detect gammas).  However, I guess there is also electron analysis...

ctapipe/ml/__init__.py
```diff
@@ -0,0 +1,19 @@
+from .apply import (
+    Reconstructor,
```
kosack, line 2: I think I made this comment a while ago as well, but I think we should avoid calling these "Reconstructors" as that will be confusing compared with those in the reco module.  Predictor or something would be better.

ctapipe/ml/apply.py
```diff
@@ -0,0 +1,381 @@
...
+        result = Table(
+            {
+                f"{self.model.model_cls}_prediction": score,
+                f"{self.model.model_cls}_is_valid": is_valid,
+            }
+        )
+        return result
+
+
+class CrossValidator(Component):
```
kosack, line 279: Missing docstring. What does this component do? Need to explain.

ctapipe/ml/tools/train_energy_regressor.py
```diff
@@ -0,0 +1,110 @@
...
+        allow_none=False,
+        directory_ok=False,
+    ).tag(config=True)
+
+    n_events = Int(default_value=None, allow_none=True).tag(config=True)
+    random_seed = Int(default_value=0).tag(config=True)
+
+    aliases = {
+        ("i", "input"): "TableLoader.input_url",
+        ("o", "output"): "TrainEnergyRegressor.output_path",
```
kosack, line 29: I'd suggest adding an alias for `CrossValidator.output_path` since that is used often (and in the tests). 

Related to that: I added this field to my ml_config.yaml under the EnergyRegressor section, and it did not complain, but also did not write out a file.

ctapipe/ml/apply.py
```diff
@@ -0,0 +1,381 @@
...
+            1,
+            0,
+        )
+        roc_auc = metrics.roc_auc_score(truth, prediction)
+        return prediction, truth, {"ROC AUC": roc_auc}
+
+    def write(self):
+        Provenance().add_output_file(self.output_path, role="ml-cross-validation")
+        for tel_type, results in self.cv_predictions.items():
+            write_table(results, self.output_path, f"cv_predictions_{tel_type}")
```
kosack, line 381: Should add some metadata to these output tables: they should at the very least have a TAB_VER and some description.

kosack (changes requested):

ctapipe/ml/tools/train_energy_regressor.py
kosack: For more info: I ran:
```sh
ctapipe-ml-train-regressor --config ml_config.yaml --input gamma_diffuse_dl2_train_small.dl2.h5 --output energy_regressor.pkl --log-level=INFO
```

And in the ml_config.yaml file, I have this inside the TrainEnergyRegressor: / EnergyRegressor: section
```yaml
...    
    FeatureGenerator:
      features:
        - ["area", "hillas_width * hillas_length"]

    CrossValidator:
      n_cross_validations: 5
      output_path: "energy_cross_validation.h5"  # this is ignored (perhaps also all?) 
```

No output is written by CrossValidate, however when I add `--CrossValidator.output_path="something.h5"` I get an output!.  So something is perhaps wrong with the config system or how the config file is being parsed

kosack: The config I get in the provanance log confirms this (and also: why does it include an EnergyRegressor?)
```
1098   │             "CrossValidator": {
1099   │                "n_cross_validations": 5,
1100   │                "output_path": null,
1101   │                "rng_seed": 1337,
1102   │                "EnergyRegressor": {
1103   │                   "model": null,
1104   │                   "QualityQuery": {
1105   │                      "quality_criteria": [
1106   │                         [
1107   │                            "HillasValid",
1108   │                            "HillasReconstructor_is_valid"
1109   │                         ],
1110   │                         [
1111   │                            "enough intensity",
1112   │                            "hillas_intensity > 50"
1113   │                         ],
1114   │                         [
1115   │                            "Positive width",
1116   │                            "hillas_width > 0"
1117   │                         ],
1118   │                         [
1119   │                            "enough pixels",
1120   │                            "morphology_n_pixels > 3"
1121   │                         ],
1122   │                         [
1123   │                            "not clipped",
1124   │                            "leakage_intensity_width_2 < 0.5"
1125   │                         ]
1126   │                      ]
1127   │                   },
1128   │                   "FeatureGenerator": {
1129   │                      "features": [
1130   │                         [
1131   │                            "area",
1132   │                            "hillas_width * hillas_length"
1133   │                         ]
1134   │                      ]
1135   │                   }
1136   │                }
1137   │             }
1138   │          }
1139   │       },
1140   │       "status": "completed",
1141   │       "duration_min": 0.07745000000008773
1142   │    }
```

ctapipe/ml/sklearn.py
```diff
@@ -0,0 +1,193 @@
...
+
+        return prediction, valid
+
+    def write(self, path):
+        Provenance().add_output_file(path, role="ml-model")
+        with open(path, "wb") as f:
+            joblib.dump(self, f, compress=True)
+
+    @classmethod
+    def load(cls, path, check_cls=True):
```
kosack, line 90: When is this load function supposed to be used? If I run `ctapipe-ml-train-regessor` and run `Regressor.load("output.pkl.gz")` on it's output, it fails, as this pkl file seems to have the regressor and a few other objects in it.    I don't see anywhere where this function is even used, but maybe I missed it?

kosack: It looks like Reconstructor.read() does it's own IO, so maybe we should remove this here? or else Reconstructor.load() should use this function.

ctapipe/ml/sklearn.py
```diff
@@ -0,0 +1,193 @@
...
+        return model
+
+    def __getstate__(self):
+        state = self.__dict__.copy()
+        state["_trait_values"]["parent"] = None
+        state["_trait_notifiers"] = {}
+        return state
+
+
+class Regressor(Model):
```
kosack, line 108: Missing docstring.   This also causes the HTML repr of this class to fail. Really, that's a bug in the Component html repr for another PR, but it's also a good way to find a missing docstring :-)

ctapipe/ml/sklearn.py
```diff
@@ -0,0 +1,193 @@
...
+        if self.log_target:
+            if prediction.unit is not None:
+                prediction.value[valid] = np.exp(prediction[valid].value)
+            else:
+                prediction[valid] = np.exp(prediction[valid])
+
+        return prediction, valid
+
+
+class Classifier(Model):
```
kosack, line 137: Missing docstring

ctapipe/ml/sklearn.py
```diff
@@ -0,0 +1,193 @@
...
+
+from ..core import Component, Provenance
+from .preprocessing import check_valid_rows, table_to_float
+
+SUPPORTED_CLASSIFIERS = dict(all_estimators("classifier"))
+SUPPORTED_REGRESSORS = dict(all_estimators("regressor"))
+SUPPORTED_MODELS = {**SUPPORTED_CLASSIFIERS, **SUPPORTED_REGRESSORS}
+
+
+class Model(Component):
```
kosack, line 20: missing docstring

ctapipe/ml/sklearn.py
```diff
@@ -0,0 +1,193 @@
...
+        # to verify settings
+        self.new_model()
+
+        self.models = {}
+        self.unit = None
+
+    def new_model(self):
+        return SUPPORTED_MODELS[self.model_cls](**self.model_config)
+
+    def table_to_X(self, table):
```
kosack, line 42: missing docstring

ctapipe/ml/sklearn.py
```diff
@@ -0,0 +1,193 @@
...
+        feature_table = table[self.features]
+        valid = check_valid_rows(feature_table, log=self.log)
+        X = table_to_float(feature_table[valid])
+        return X, valid
+
+    def table_to_y(self, table, mask=None):
```
kosack, line 48: missing docstring

ctapipe/ml/sklearn.py
```diff
@@ -0,0 +1,193 @@
...
+        if self.unit is not None:
+            return table[mask][self.target].quantity.to_value(self.unit)
+        return np.array(table[self.target][mask])
+
+    def fit(self, key, table):
```
kosack, line 53: missing docstring

ctapipe/ml/sklearn.py
```diff
@@ -0,0 +1,193 @@
...
+        self.models[key] = self.new_model()
+
+        X, valid = self.table_to_X(table)
+        y = self.table_to_y(table, mask=valid)
+        self.unit = table[self.target].unit
+        self.models[key].fit(X, y)
+
+    def predict(self, key, table):
```
kosack, line 61: missing docstring

kosack (changes requested):
I find the class Hierarchy has some real naming problems.  I suggest trying to alwyas include the parent class in the sub-class name, so it's clear what the hierarchy is (as we do elsewhere). 

For example:
* `EnergyRegressor` is not a subclass of `Regressor`, but of `Reconstructor` !!?  This is super confusing.
* `Reconstructor` is also completely unrelated to the `reco.Reconstructor` (and is not even in the reco module), so I highly suggest changing the name to `Predictor` or something so we don't get confused.
* `Regressor` / `Classifiers` are both subclasses of `Model`, so perhaps call it `RegressionModel` and `ClassificationModel`?


Perhaps:
```mermaid
classDiagram 
   Predictor <|-- RegressionPredictor
   Predictor <|-- ClassificationPredictor
   RegressionPredictor <|-- EnergyRegressionPredictor
   ClassificationPredictor <|-- ParticleClassificationPredictor
```

```mermaid
classDiagram
    Model
    Model <|-- RegressionModel
    Model <|-- ClassificationModel
```

Also, there are many many missing docstrings, some of which even break the html reprs.  The only places where you can skip a docstring is if that method is already documented in a parent class (e.g. fit() in Model should be documented, but sub-classes need not include a docstring as it will be inherited from the parent)

ctapipe/ml/__init__.py
nbiederbeck: For reference: https://github.com/cta-observatory/ctapipe/issues/1955

ctapipe/ml/tools/train_energy_regressor.py
maxnoe: @kosack The ml config example has a wrong indentation for the `CrossvalidatorConfig`. It must be a child of `TrainEnergyRegressor`, not of `EnergyRegressor`

maxnoe:
@kosack I restructured the ml module to simplify the structure. The separate low-level classes only dealing with sklearn are now gone and the hirarchy is:

```
ctapipe.reco.Reconstructor
   ctapipe.ml.SKLearnReconstructor
     ctapipe..ml.SKLearnRegressionReconstructor
       ctapipe.ml.EnergyRegressor   # giving it the full name was too long for my taste, python has introspection
     ctapipe.ml.SKLearnClassificationReconstructor
        ctapipe.ml.ParticleIdClassifier
```

@kosack @LukasBeiske @LukasNickel @nbiederbeck Please have a look again.

ctapipe/conftest.py
```diff
@@ -145,6 +145,13 @@ def prod5_gamma_simtel_path():
     return get_dataset_path("gamma_prod5.simtel.zst")
 
 
+@pytest.fixture(scope="session")
+def prod5_gamma_lapalma_simtel_path():
```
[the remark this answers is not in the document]
maxnoe: To have proton and gamma file with the same subarray I think

ctapipe/ml/conftest.py
maxnoe: done

ctapipe/ml/conftest.py
maxnoe: done

ctapipe/ml/stereo_combination.py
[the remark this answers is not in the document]
maxnoe: What does that mean? Do we want to have a stereo quality query here again? I'd say it's enough to use the `is_valid` for that and use the same quality query as for the mono predictions (otherwise, why have a quality query for the mono predictions other than non-nan?)

ctapipe/ml/sklearn.py
maxnoe: In the refactoring I completely removed the classes formerly in `sklearn.py`, IO now only happens through pickling the reconstructors.

LukasBeiske:
> @kosack I restructured the ml module to simplify the structure. The separate low-level classes only dealing with sklearn are now gone and the hirarchy is:
> 
> ```
> ctapipe.reco.Reconstructor
>    ctapipe.ml.SKLearnReconstructor
>      ctapipe..ml.SKLearnRegressionReconstructor
>        ctapipe.ml.EnergyRegressor   # giving it the full name was too long for my taste, python has introspection
>      ctapipe.ml.SKLearnClassificationReconstructor
>         ctapipe.ml.ParticleIdClassifier
> ```
> 
> @kosack @LukasBeiske @LukasNickel @nbiederbeck Please have a look again.

Looks good. Removing the low-level classes makes everything easier to follow imo. I will change the disp component to be a subclass of `ctapipe.reco.Reconstructor`, as requested.

ctapipe/ml/tools/apply.py
```diff
@@ -0,0 +1,220 @@
...
+        prefix = reconstructor.model_cls
+
+        tel_tables = []
+
+        desc = f"Applying {reconstructor.__class__.__name__}"
+        unit = "telescope"
+        for tel_id, tel in tqdm(self.loader.subarray.tel.items(), desc=desc, unit=unit):
+            if tel not in reconstructor._models:
+                self.log.warning(
+                    "No regressor model for telescope type %s, skipping tel %d",
```
LukasBeiske, line 160: This warning is also used for the particle classifier, so I would cut the "regressor".

maxnoe: done

setup.py
kosack: yes, particularly as there will eventually be a classifier of the event type (reconstruction quality) that is not the same as the particle type classifier.

kosack (commented):
Still need to fix the docs so this module at least appears.

Perhaps we could also modify *stage2_config.py* (or create stage2_full_config.py) to show how one would run an analysis up to stage 2 with pre-trained models for energy and gammaness as inputs, e.g having the new Reconstructors in the list with the proper inputs/outputs.     That could also just be part of the docs, not the standard configs.

ctapipe/containers.py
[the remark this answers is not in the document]
kosack: Yes, there will eventually be an electron classifier for example

maxnoe:
@kosack For now I introduced a `GeometryReconstructor` in between, as I couldn't find a nice solution to read the model yet in a general manner. 

I will have a look again.

ctapipe/ml/stereo_combination.py
```diff
@@ -0,0 +1,249 @@
...
+    weights = CaselessStrEnum(
+        ["none", "intensity", "konrad"],
+        default_value="none",
+    ).tag(config=True)
+
+    def _calculate_weights(self, data):
+        """"""
+
+        if isinstance(data, Container):
+            if self.weights == "intensity":
```
kosack, line 77: Would it make sense here to use an ExpressionEngine rather than hard-coded list of weights? That way one could say
```yaml
weights:  "hillas_intensity * hillas_length / hillas_width"
```

ctapipe/ml/__init__.py
kosack: Now that they are all properly subclasses of `reco.Reconstructor`, I'm happy.

ctapipe/ml/stereo_combination.py
kosack: At the very least, rename the "konrad" option something understandable like "aspect_weighted_intensity" (width/length is the aspect ratio of the ellipse)

maxnoe:
I ran this on the prod5b La Palma Alpha Layput using the config here:
https://github.com/maxnoe/test_ctapipe_ml/

Find some performance metrics here:

https://github.com/maxnoe/test_ctapipe_ml/blob/main/performance.ipynb

Looks good I'd say

ctapipe/reco/sklearn.py
```diff
@@ -0,0 +1,590 @@
...
+        return instance
+
+    @lazyproperty
+    def instrument_table(self):
+        return QTable(self.subarray.to_table("joined"))
+
+    def _new_model(self):
+        return SUPPORTED_MODELS[self.model_cls](**self.model_config)
+
+    def _table_to_X(self, table):
```
LukasNickel, line 188: Although these `table_to` are only used in the fit method below, we might want to document them. It might not be completely obvious why they behave so differently.
Also `valid` might indicate it is the subset of valid `X` as opposed to a boolean mask. Later on we also call it `mask`

I understand the code, but I have also worked on/with it for a while, not sure how clear this is for other people in the future.

ctapipe/reco/sklearn.py
```diff
@@ -0,0 +1,590 @@
...
+    invalid_class = Integer(
+        default_value=-1, help="The label to fill in case no prediction could be made"
+    ).tag(config=True)
+
+    positive_class = Integer(
+        default_value=1,
+        help=(
+            "The label value of the positive class."
+            " ``prediction`` values close to 1.0 mean the event"
+            " belonged likely to this class."
```
LukasNickel, line 294: This is actually true even for the binary case with only one output label, but at some point we should note how the output differs between binary and multi class problems.

ctapipe/reco/sklearn.py
```diff
@@ -0,0 +1,590 @@
...
+    def _get_positive_index(self, key):
+        return np.nonzero(self._models[key].classes_ == self.positive_class)[0][0]
+
+
+class EnergyRegressor(SKLearnRegressionReconstructor):
+    """
+    Use a scikit-learn regression model per telescope type to predict primary energy
+    """
+
+    #: Name of the target table column for training
```
LukasNickel, line 354: I think that is self explanatory

ctapipe/reco/sklearn.py
```diff
@@ -0,0 +1,590 @@
...
+    positive_class = Integer(
+        default_value=0,
+        help="Particle id (in simtel system) of the positive class. Default is 0 for gammas.",
+    ).tag(config=True)
+
+    def __call__(self, event: ArrayEventContainer) -> None:
+        for tel_id in event.trigger.tels_with_trigger:
+            table = collect_features(event, tel_id, self.instrument_table)
+            table = self.generate_features(table)
+            mask = self.qualityquery.get_table_mask(table)
```
LukasNickel, line 432: This is inconsistent with above, where the 0th value is taken directly and the result named `passes_quality_checks` (l366)

ctapipe/reco/sklearn.py
```diff
@@ -0,0 +1,590 @@
...
+
+    def predict_table(self, key, table: Table) -> Table:
+        """Predict on a table of events"""
+        table = self.generate_features(table)
+
+        n_rows = len(table)
+        score = np.full(n_rows, np.nan)
+        is_valid = np.full(n_rows, False)
+
+        mask = self.qualityquery.get_table_mask(table)
```
LukasNickel, line 462: Further up `valid` is used as variable name

ctapipe/reco/stereo_combination.py
```diff
@@ -0,0 +1,291 @@
...
+        """
+
+
+class StereoMeanCombiner(StereoCombiner):
+    """
+    Calculate array-event prediction as (weighted) mean of telescope-wise predictions
+    """
+
+    weights = CaselessStrEnum(
+        ["none", "intensity", "konrad"],
```
LukasNickel, line 84: Do we have a new name for konrad yet?

ctapipe/reco/stereo_combination.py
```diff
@@ -0,0 +1,291 @@
...
+    def predict_table(self, mono_predictions: Table) -> Table:
+        """
+        Calculates the (array-)event-wise mean.
+        Telescope events, that are nan, get discarded.
+        This means you might end up with less events if
+        all telescope predictions of a shower are invalid.
+        """
+
+        prefix = f"{self.prefix}_tel"
+        # TODO: Integrate table quality query once its done
```
LukasNickel, line 211: Do we want to work on that todo?

ctapipe/tools/process.py
```diff
@@ -68,11 +68,15 @@ class ProcessorTool(Tool):
         default_value=False,
     ).tag(config=True)
 
+    stereo_combiner_configs = List(Dict()).tag(config=True)
```
LukasNickel, line 71: What are these for?
Are they not part of the reconstructor configs?

ctapipe/tools/tests/test_apply_modesl.py
```diff
@@ -0,0 +1,138 @@
...
+    for col in "obs_id", "event_id":
+        assert table[col].description == EventIndexContainer.fields[col].description
+
+    for name, field in ReconstructedEnergyContainer.fields.items():
+        colname = f"ExtraTreesRegressor_{name}"
+        assert colname in table.colnames
+        assert table[colname].description == field.description
+
+    loader = TableLoader(output_path, load_dl2=True)
+    events = loader.read_subarray_events()
```
LukasNickel, line 42: Should we test that these are not (all) nan aswell here (and in the other tests)?

ctapipe/tools/tests/test_process_ml.py
```diff
@@ -0,0 +1,97 @@
...
+    argv = [
+        f"--input={input_url}",
+        f"--output={output}",
+        "--write-images",
+        "--write-showers",
+        f"--config={config_path}",
+    ]
+    assert run_tool(ProcessorTool(), argv=argv, cwd=tmp_path, raises=True) == 0
+
+    print(read_table(output, "/dl2/event/telescope/energy/ExtraTreesRegressor/tel_004"))
```
LukasNickel, line 49: Leftover from testing?

ctapipe/tools/tests/test_train.py
```diff
@@ -0,0 +1,96 @@
...
+    energy_cv_out_file = model_tmp_path / "energy_cv_results.h5"
+
+    ret = run_tool(
+        tool,
+        argv=[
+            "--input=dataset://gamma_diffuse_dl2_train_small.dl2.h5",
+            f"--output={out_file}",
+            f"--config={config}",
+            f"--cv-output={energy_cv_out_file}",
+            "--log-level=INFO",
```
LukasNickel, line 74: That has no use here aynmore right?

ctapipe/tools/train_energy_regressor.py
```diff
@@ -0,0 +1,127 @@
...
+
+        self.regressor = EnergyRegressor(self.loader.subarray, parent=self)
+        self.cross_validate = CrossValidator(
+            parent=self, model_component=self.regressor
+        )
+        self.rng = np.random.default_rng(self.random_seed)
+
+    def start(self):
+        """
+        Train models per telescope type using a cross-validation.
```
LukasNickel, line 68: It is performing a cross validation and after that trains the final model

ctapipe/tools/train_particle_classifier.py
```diff
@@ -0,0 +1,173 @@
...
+            parent=self,
+        )
+        self.rng = np.random.default_rng(self.random_seed)
+        self.cross_validate = CrossValidator(
+            parent=self, model_component=self.classifier
+        )
+
+    def start(self):
+        """
+        Train models per telescope type using a cross-validation.
```
LukasNickel, line 101: Same here.
I prefer the way it is phrased in the class docstring

ctapipe/reco/sklearn.py
```diff
@@ -0,0 +1,590 @@
...
+
+    def __call__(self, event: ArrayEventContainer) -> None:
+        """
+        Apply model for a single event and fill result into the event container
+        """
+        for tel_id in event.trigger.tels_with_trigger:
+            table = collect_features(event, tel_id, self.instrument_table)
+            table = self.generate_features(table)
+
+            passes_quality_checks = self.qualityquery.get_table_mask(table)[0]
```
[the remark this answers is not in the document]
maxnoe: a table with one row, yes.

ctapipe/reco/sklearn.py
maxnoe: `#:` makes this a sphinx comment, so it appears in the docs.

ctapipe/tools/process.py
maxnoe: Yes, leftover from a previous version. Not needed anymore.

ctapipe/reco/stereo_combination.py
maxnoe: No. It solve that globally in an another PR. Currently it's the same as in the geometry reconstructors.

ctapipe/reco/stereo_combination.py
maxnoe: Yes, but not in this PR, it's too large already.

ctapipe/tools/tests/test_train.py
maxnoe: It provides valuable info in case the test fails I think (stderror is printed nicely by pytest for failing tests)

ctapipe/ml/apply.py
```diff
@@ -0,0 +1,381 @@
...
+
+    target = None
+    model_cls = Model
+    model = Instance(model_cls, allow_none=True).tag(config=True)
+
+    def __init__(self, subarray, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+        self.subarray = subarray
+        self.qualityquery = QualityQuery(parent=self)
+        self.generate_features = FeatureGenerator(parent=self)
```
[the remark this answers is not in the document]
maxnoe: I like it better to have the instance called like the class, but snakecase. Even if it implements `__call__`.

ctapipe/reco/sklearn.py
```diff
@@ -0,0 +1,602 @@
...
+
+    def fit(self, key, table):
+        """
+        Create and fit a new model for ``key`` using the data in ``table``.
+        """
+        self._models[key] = self._new_model()
+
+        X, valid = self._table_to_X(table)
+        y = self._table_to_y(table, mask=valid)
+        self.unit = table[self.target].unit
```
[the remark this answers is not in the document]
maxnoe: You are right, it's a bit confusing, and I will switch the order here, but actually it doesn't matter. In training, If unit is None, it just returns values and then unit is set then to the unit of the columns. It's important for applying a trained model where the units of the input might be different.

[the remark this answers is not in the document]
LukasBeiske: I see.

maxnoe:
I hope I addressed all review comments. Please have a look again.

ctapipe/utils/download.py
```diff
@@ -56,7 +56,7 @@ def download_file(url, path, auth=None, chunk_size=10240, progress=False):
                 for chunk in r.iter_content(chunk_size=chunk_size):
                     f.write(chunk)
                     pbar.update(len(chunk))
-        except BaseException:  # we really want to catch everything here
+        except BaseException:  # we really want to catch everythin here
```
LukasNickel, line 59: did you revert a commit here?

LukasNickel (dismissed):

ctapipe/utils/download.py
maxnoe: rebased, yes...

LukasNickel (dismissed):

LukasBeiske (approved):

ctapipe/reco/stereo_combination.py
```diff
@@ -0,0 +1,290 @@
...
+
+__all__ = [
+    "StereoCombiner",
+    "StereoMeanCombiner",
+]
+
+
+def _calculate_ufunc_of_telescope_values(tel_data, n_array_events, indices, ufunc):
+    combined_values = np.zeros(n_array_events)
+    ufunc.at(combined_values, indices, tel_data)
```
nbiederbeck, line 30: This works only for additive ufuncs, right? Not for e.g. multiplication. We currently only use it for `np.add` though, but that might become an issue

ctapipe/reco/stereo_combination.py
```diff
@@ -0,0 +1,290 @@
...
+class StereoCombiner(Component):
+    """Base Class for algorithms combining telescope-wise predictions to common prediction"""
+
+    prefix = Unicode(
+        default_value="",
+        help="Prefix to be added to the output container / column names",
+    ).tag(config=True)
+    property = CaselessStrEnum(
+        ["energy", "classification", "direction"],
+        help="Whic property is being combined",
```
nbiederbeck, line 61: typo

ctapipe/reco/stereo_combination.py
```diff
@@ -0,0 +1,290 @@
...
+    def __call__(self, event: ArrayEventContainer) -> None:
+        """
+        Calculate the mean prediction for a single array event.
+        """
+        if self.property == "energy":
+            self._combine_energy(event)
+        elif self.property == "classification":
+            self._combine_classification(event)
+        else:
+            raise NotImplementedError(f"Cannot combine {self.property}")
```
nbiederbeck, line 199: maybe add an "implemented combinations: ..."

ctapipe/tools/apply_models.py
```diff
@@ -0,0 +1,234 @@
...
+
+    Models need to be trained with
+    `~ctapipe.tools.train_energy_regressor.TrainEnergyRegressor`
+    and
+    `~ctapipe.tools.train_particle_classifier.TrainParticleClassifier`.
+    """
+
+    name = "ctapipe-ml-apply"
+    description = __doc__
+    examples = """
```
nbiederbeck, line 36: wrong example

ctapipe/tools/apply_models.py
```diff
@@ -0,0 +1,234 @@
...
+        "particle-classifier": "ApplyModels.particle_classifier_path",
+        ("o", "output"): "ApplyModels.output_path",
+    }
+
+    flags = {
+        **flag(
+            "overwrite",
+            "ApplyModels.overwrite",
+            "Overwrite tables in output file if it exists",
+            "Don't overwrite tables in output file if it exists",
```
nbiederbeck, line 89: ~~it~~ → they

ctapipe/tools/tests/test_process_ml.py
```diff
@@ -0,0 +1,101 @@
...
+        f"--config={config_path}",
+    ]
+    assert run_tool(ProcessorTool(), argv=argv, cwd=tmp_path, raises=True) == 0
+
+    print(
+        read_table(
+            output, "/dl2/event/telescope/classification/ExtraTreesClassifier/tel_004"
+        )
+    )
+    print(read_table(output, "/dl2/event/subarray/classification/ExtraTreesClassifier"))
```
nbiederbeck, line 101: replace with asserts

ctapipe/tools/tests/test_apply_models.py
```diff
@@ -0,0 +1,144 @@
...
+
+    for name, field in ReconstructedEnergyContainer.fields.items():
+        colname = f"ExtraTreesRegressor_{name}"
+        assert colname in table.colnames
+        assert table[colname].description == field.description
+
+    loader = TableLoader(output_path, load_dl2=True)
+    events = loader.read_subarray_events()
+
+    prefix = "ExtraTreesRegressor"
```
nbiederbeck, line 46: could event be in line 34, and maybe needs to be added/confirmed with the (default) configuration used?

ctapipe/tools/tests/test_apply_models.py
```diff
@@ -0,0 +1,144 @@
...
+        assert table[col].description == EventIndexContainer.fields[col].description
+
+    for name, field in ParticleClassificationContainer.fields.items():
+        colname = f"ExtraTreesClassifier_{name}"
+        assert colname in table.colnames
+        assert table[colname].description == field.description
+
+    loader = TableLoader(output_path, load_dl2=True)
+    events = loader.read_subarray_events()
+    assert "ExtraTreesClassifier_prediction" in events.colnames
```
nbiederbeck, line 98: see above with `prefix`

ctapipe/tools/tests/test_apply_models.py
```diff
@@ -0,0 +1,144 @@
...
+    assert ret == 0
+
+    loader = TableLoader(output_path, load_dl2=True)
+
+    events = loader.read_subarray_events()
+    assert "ExtraTreesRegressor_energy" in events.colnames
+    assert "ExtraTreesClassifier_prediction" in events.colnames
+
+    events = loader.read_telescope_events()
+    assert "ExtraTreesClassifier_prediction" in events.colnames
```
nbiederbeck, line 143: see above with `prefix`. Now I see the `prefix` and or configuration can be file-global

ctapipe/tools/train_energy_regressor.py
```diff
@@ -0,0 +1,156 @@
...
+from ctapipe.reco.preprocessing import check_valid_rows
+
+__all__ = [
+    "TrainEnergyRegressor",
+]
+
+
+class TrainEnergyRegressor(Tool):
+    """
+    Tool to train a `~ctapipe.reco.EnergyRegressor` on dl2 data.
```
nbiederbeck, line 16: dl1b / dl2

ctapipe/tools/train_energy_regressor.py
```diff
@@ -0,0 +1,156 @@
...
+
+    The tool first performs a cross validation to give an initial estimate
+    on the quality of the estimation and then finally trains one model
+    per telescope type on the full dataset.
+    """
+
+    name = "ctapipe-train-regressor"
```
nbiederbeck, line 23: train-energy-regressor

ctapipe/tools/train_energy_regressor.py
```diff
@@ -0,0 +1,156 @@
...
+
+    overwrite = Bool(help="overwrite existing output files").tag(config=True)
+
+    random_seed = Int(
+        default_value=0, help="Random seed for sampling and cross validation"
+    ).tag(config=True)
+
+    flags = {
+        **flag(
+            "overwrite",
```
nbiederbeck, line 57: we have `-f` as "force" in the other tools (including ml tools). But I think we can adress the `--overwrite/-f` in another PR to make everything consistent (and maybe inside Tool?)

ctapipe/tools/train_particle_classifier.py
```diff
@@ -0,0 +1,214 @@
...
+from ctapipe.reco.preprocessing import check_valid_rows
+
+__all__ = [
+    "TrainParticleClassifier",
+]
+
+
+class TrainParticleClassifier(Tool):
+    """
+    Tool to train a `~ctapipe.reco.ParticleClassifier` on dl2 data.
```
nbiederbeck, line 17: dl1b / dl2

ctapipe/reco/reconstructor.py
```diff
@@ -62,6 +67,16 @@ def __call__(self, event: ArrayEventContainer):
...
 
+
+class GeometryReconstructor(Reconstructor):
+    """
+    Base class for algorithms predicting only the shower geometry
+    """
+
+    def __init__(self, subarray, **kwargs):
+        super().__init__(subarray, **kwargs)
+        self.check_parameters = StereoQualityQuery(parent=self)
```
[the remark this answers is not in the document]
maxnoe: Indeed...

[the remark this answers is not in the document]
maxnoe: Now we can decide: should we add it already in the baseclass and remove in the children? Or remove from the baseclass?

ctapipe/reco/sklearn.py
```diff
@@ -0,0 +1,614 @@
...
+
+    Keeps a dictionary of sklearn models, the current tools are designed
+    to train one model per telescope type.
+
+    Models can either be instantiated normally, or they can load themselves
+    from a pickled file if the ``load_path`` traitlet is given.
+    """
+
+    #: What particle property is being predicted, one of "energy", "direction", "classification"
+    property = None
```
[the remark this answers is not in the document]
maxnoe: ah wait, it is, sorry...

ctapipe/reco/sklearn.py
```diff
@@ -0,0 +1,614 @@
...
+            Key of the model. Currently always a `~ctapipe.instrument.TelescopeDescription`
+            as we train models per telescope type.
+        table : `~astropy.table.Table`
+            Table of features
+
+        Returns
+        -------
+        table : `~astropy.table.Table`
+            Table(s) with predictions, matches the corresponding
+            container definition(s)
```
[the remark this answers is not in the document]
maxnoe: Yes, in anticipation of the disp regressor, which would return disp tables and geometry tables.

ctapipe/reco/stereo_combination.py
maxnoe: maybe better hard code `add` for now and call it `grouped_add`?

nbiederbeck: yes

ctapipe/tools/apply_models.py
LukasNickel: Then also exist~~s~~

ctapipe/tools/apply_models.py
```diff
@@ -0,0 +1,234 @@
...
+            lambda c: c.name.endswith("telescopes"),
+            stereo_predictions.columns.values(),
+        ):
+            stereo_predictions[c.name] = np.array([trafo(r) for r in c])
+            stereo_predictions[c.name].description = c.description
+
+        write_table(
+            stereo_predictions,
+            self.output_path,
+            f"/dl2/event/subarray/{combiner.property}/{combiner.prefix}",
```
LukasBeiske, line 219: Since `combiner.property` is used in the table path here, it would be helpful to rename the property "direction" to "geometry" so the disp geometry prediction gets saved at the correct place.

LukasBeiske: But I can also change that in the disp PR

maxnoe: done

ctapipe/tools/train_energy_regressor.py
maxnoe: Yes

ctapipe/reco/stereo_combination.py
maxnoe: Done (and moved check to `__init__`)

ctapipe/tools/train_energy_regressor.py
```diff
@@ -0,0 +1,156 @@
...
+            table = table[idx]
+
+        return table
+
+    def finish(self):
+        """
+        Write-out trained models and cross-validation results.
+        """
+        self.log.info("Writing output")
+        self.regressor.write(self.output_path, overwrite=self.overwrite)
```
nbiederbeck, line 145: We should move this to somewhere earlier. I just waited for the training to finish, only to then fail here.

nbiederbeck: But we can think about that in #2111, because all Tools (might) have this issue.

nbiederbeck (approved):

ctapipe/tools/apply_models.py
```diff
@@ -0,0 +1,234 @@
...
+    """
+
+    name = "ctapipe-apply-models"
+    description = __doc__
+    examples = """
+    ctapipe-apply-models \\
+        --input gamma.dl2.h5 \\
+        --energy-regressor energy_regressor.pkl \\
+        --particle-classifier particle-classifier.pkl \\
+        --ouput gamma_applied.dl2.h5
```
kosack, line 41: typo, should be `--output`

ctapipe/reco/sklearn.py
```diff
@@ -0,0 +1,628 @@
...
+
+class CrossValidator(Component):
+    """Class to train sklearn based reconstructors in a cross validation"""
+
+    n_cross_validations = Int(5).tag(config=True)
+
+    output_path = Path(
+        default_value=None,
+        allow_none=True,
+        directory_ok=False,
```
kosack, line 518: please add a help string here.  It should at least explain format is the output in. Is is a PDF plot or a picked class? Impossible to tell.  I assumed pkl like the other output, but then looking at the file, it seems to be hdf5. That's not obvious.

ctapipe/tools/train_energy_regressor.py
```diff
@@ -0,0 +1,156 @@
...
+        --config ml_config.yaml \\
+        --input gamma.dl2.h5 \\
+        --ouput energy_regressor.pkl
+    """
+
+    output_path = Path(
+        default_value=None,
+        allow_none=False,
+        directory_ok=False,
+        help="Ouput path for the trained reconstructor.",
```
kosack, line 37: and here.  Also should mention the file type that is written, so the user knows what file extension to use.

ctapipe/tools/train_energy_regressor.py
```diff
@@ -0,0 +1,156 @@
...
+            "Number of events for training the model."
+            " If not give, all available events will be used."
+        ),
+    ).tag(config=True)
+
+    overwrite = Bool(help="overwrite existing output files").tag(config=True)
+
+    random_seed = Int(
+        default_value=0, help="Random seed for sampling and cross validation"
+    ).tag(config=True)
```
kosack, line 53: It's not clear here if the random seed value of 0 is special or not, or if there is an option to auto-generate one.   
Shouldn't the default be to auto-generate?

kosack (changes requested):
Nice! I tested this by running my own analysis including:
- training an energy regressor
- applying the energy regressor with `ctapipe-apply-models`
- applying the energy regressor with `ctapipe-process`

Everything worked as expected.  I just have a few minor changes to request below

ctapipe/tools/train_energy_regressor.py
kosack: Also, if auto-generated, it should somehow get written to the output (I guess it will be in the pickled class at least)

maxnoe: There is no option to auto-generate one and I think the default should be to be reproducible. There is no reason for auto-generating one I think, since this is not a simulation, it's about splitting the events / sub-sampling them.

maxnoe: 0 is not special here

maxnoe:
@kosack I think I addressed all comments.

nbiederbeck (dismissed):

nbiederbeck (approved):

ctapipe/tools/tests/test_apply_models.py
```diff
@@ -0,0 +1,143 @@
...
+
+    ret = run_tool(
+        ApplyModels(),
+        argv=[
+            f"--input={input_path}",
+            f"--output={output_path}",
+            f"--energy-regressor={energy_regressor_path}",
+            "--StereoMeanCombiner.weights=konrad",
+        ],
+        raises=True,
```
LukasNickel, line 30: What is the purpose of `raises=True` here again?

LukasNickel (approved):

LukasBeiske (approved):

PR #1769, opened 2021-07-30
Add condition option to read_table

maxnoe:
This adds the possibility to use `tables` `read_where` method using a `numexpr` expression to only load rows fulfilling condition.

Very handy for example to load data from only one telescope from a `by_type` table: `condition="tel_id == 1"` or only load reconstructed events: `condition = "hillas_length > 0"`

### DIFF
--- a/ctapipe/io/astropy_helpers.py
+++ b/ctapipe/io/astropy_helpers.py
@@ -30,12 +30,24 @@ def read_table(h5file, path, start=None, stop=None, step=None, condition=None) -
     This uses the same conventions as the `~ctapipe.io.HDF5TableWriter`,
     with the exception of Enums, that will remain as integers.
 
+    (start, stop, step) behave like python slices.
+
     Parameters
     ----------
     h5file: Union[str, Path, tables.file.File]
         input filename or PyTables file handle
     path: str
         path to table in the file
+    start: int or None
+        if given, this is the first row to be loaded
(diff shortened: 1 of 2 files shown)


kosack (approved):
good idea - pytables' in-kernel filters are very efficient as well, so this should speed up any pre-filtering.

kosack:
do you know if the start/stop/step apply before or after the filtering? I was going to implement a chunked iterator wrapping read_table, so that may have an effect.

maxnoe:
> do you know if the start/stop/step apply before or after the filtering? I was going to implement a chunked iterator wrapping read_table, so that may have an effect.

It's before, the rows are first selected based on the slice(start, stop, step) and then filtered by condition:
```
In [13]: f.root.test.read()
Out[13]: 
array([(0, -1.07400584), (1, -0.90504077), (2,  1.22623558),
       (3, -0.3141724 ), (4, -0.26661846), (5, -1.7715964 ),
       (6, -0.61327828), (7,  0.98540159), (8, -0.89481293),
       (9, -0.27684014)], dtype=[('idx', '<i8'), ('value', '<f8')])

In [14]: f.root.test.read_where(step=2, condition='value > 0')
Out[14]: array([(2, 1.22623558)], dtype=[('idx', '<i8'), ('value', '<f8')])
```

HealthyPear (approved):

PR #1770, opened 2021-07-30
Implement a DataLoader

HealthyPear:
#### Description

This PR adds a `DataLoader`, analogous to `DataWriter`.

It builds over the class used during the [DL1 benchmarking workshop](https://gitlab.cta-observatory.org/cta-consortium/aswg/workshops/dl1-benchmarks-workshop).

It should be able to automatically detect the `split_datasets_by` option used to write the file and behave accordingly without specifying this information.

#### Caveats

- for now I added only methods up to DL1b - after #1726 it should be sufficient to add 2 methods for DL2 shower geometry (mono and reco)
- the class allows to extract parameters (+ trigger + pointing) and images separately, perhaps we could merge all together, perhaps not

#### TODO

- [ ] general cleanup (mostly after early feedback)
- [ ] add unit-tests
- [ ] fill docstrings

### DIFF
--- a/ctapipe/io/dataloader.py
+++ b/ctapipe/io/dataloader.py
@@ -1,0 +1,426 @@
+from pathlib import Path
+import re
+
+from astropy.table import Table, join, vstack
+import astropy.units as u
+import numpy as np
+import tables
+from tables import NoSuchNodeError
+from traitlets import Bool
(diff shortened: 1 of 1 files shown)


maxnoe:
See #1771

PR #1771, opened 2021-07-30
TableLoader

maxnoe:
I adapted this from the class in the dl1 benchmarks workshop, but changed the structure a bit.

I added options for what to load and for now a single method `load_telescope_events(tel_id)` to read / join all needed columns for one telescope.

Methods to come:

- loading only stereo data
- loading events for multiple tel_ids (as dict by type or just as single table).
- more?

Needs #1769

### DIFF
--- a/.pylintrc
+++ b/.pylintrc
@@ -425,6 +425,7 @@ good-names=i,
            x,
            y,
            n,
+		   f,
            ex,
            _
 
--- a/ctapipe/io/__init__.py
+++ b/ctapipe/io/__init__.py
@@ -2,7 +2,7 @@ from .eventseeker import EventSeeker
 from .eventsource import EventSource
 from .hdf5tableio import HDF5TableReader, HDF5TableWriter
 from .tableio import TableWriter, TableReader
-from .tableloader import get_structure, TableLoader
+from .tableloader import TableLoader
 from .datalevels import DataLevel
 from .astropy_helpers import read_table
 from .datawriter import DataWriter, DATA_MODEL_VERSION
@@ -22,7 +22,6 @@ __all__ = [
     "HDF5TableReader",
     "TableWriter",
     "TableReader",
-    "get_structure",
     "TableLoader",
     "EventSeeker",
     "EventSource",
--- a/ctapipe/io/__init__.py
+++ b/ctapipe/io/__init__.py
@@ -2,6 +2,7 @@ from .eventseeker import EventSeeker
 from .eventsource import EventSource
 from .hdf5tableio import HDF5TableReader, HDF5TableWriter
 from .tableio import TableWriter, TableReader
+from .tableloader import get_structure, TableLoader
 from .datalevels import DataLevel
 from .astropy_helpers import read_table
 from .datawriter import DataWriter, DATA_MODEL_VERSION
@@ -21,6 +22,8 @@ __all__ = [
     "HDF5TableReader",
     "TableWriter",
     "TableReader",
+    "get_structure",
+    "TableLoader",
     "EventSeeker",
     "EventSource",
     "SimTelEventSource",
--- a/ctapipe/io/tableloader.py
+++ b/ctapipe/io/tableloader.py
@@ -21,6 +21,7 @@ GEOMETRY_GROUP = "/dl2/event/subarray/geometry"
 TRIGGER_TABLE = "/dl1/event/subarray/trigger"
 SHOWER_TABLE = "/simulation/event/subarray/shower"
 TRUE_IMAGES_GROUP = "/simulation/event/telescope/images"
+TRUE_PARAMETERS_GROUP = "/simulation/event/telescope/parameters"
 
 by_id_RE = re.compile(r"tel_\d+")
 
@@ -77,6 +78,8 @@ class TableLoader(Component):
         False, help="load simulated shower information").tag(config=True)
     load_true_images = traits.Bool(
         False, help="load simulated shower images").tag(config=True)
+    load_true_parameters = traits.Bool(
+        False, help="load image parameters obtained from true images").tag(config=True)
     load_trigger = traits.Bool(
         True, help="load subarray trigger information").tag(config=True)
     load_instrument = traits.Bool(
@@ -328,6 +331,20 @@ class TableLoader(Component):
                 table = join(
                     table,
                     true_images,
+                    keys=["obs_id", "event_id", "tel_id"],
+                    join_type="outer",
+                )
+
+        if self.load_true_parameters:
+            true_parameters = self._read_telescope_table(TRUE_PARAMETERS_GROUP, tel_id)
+            for col in set(true_parameters.colnames) - {"obs_id", "event_id", "tel_id"}:
+                true_parameters.rename_column(col, f"true_{col}")
+            if table is None or len(table) == 0:
+                table = true_parameters
+            else:
+                table = join(
+                    table,
+                    true_parameters,
                     keys=["obs_id", "event_id", "tel_id"],
                     join_type="outer",
                 )
--- a/ctapipe/io/tableloader.py
+++ b/ctapipe/io/tableloader.py
@@ -3,17 +3,16 @@ Class and functions to read DL1 (a,b) and/or DL2 (a) data from an HDF5 file prod
 """
 
 import re
-from typing import List, Union
 
 import numpy as np
 import tables
 from astropy.table import join, vstack, Table
 
 from ..core import Component, traits, Provenance
-from ..instrument import SubarrayDescription, TelescopeDescription
+from ..instrument import SubarrayDescription
 from .astropy_helpers import read_table
 
-__all__ = ["get_tel_ids", "get_structure", "TableLoader"]
+__all__ = ["get_structure", "TableLoader"]
 
 PARAMETERS_GROUP = "/dl1/event/telescope/parameters"
 IMAGES_GROUP = "/dl1/event/telescope/images"
@@ -24,24 +23,6 @@ TRUE_IMAGES_GROUP = "/simulation/event/telescope/images"
 TRUE_PARAMETERS_GROUP = "/simulation/event/telescope/parameters"
 
 by_id_RE = re.compile(r"tel_\d+")
-
-
-def get_tel_ids(
-    subarray: SubarrayDescription,
-    telescopes: List[Union[int, str, TelescopeDescription]]
-) -> List[int]:
-    ids = set()
-
-    valid_tel_types = [str(tel_type) for tel_type in subarray.telescope_types]
-
-    for telescope in telescopes:
-        if isinstance(telescope, int):
-            ids.add(telescope)
-        if isinstance(telescope, str) and (telescope not in valid_tel_types):
-            raise ValueError("Invalid telescope type input.")
-        ids.update(subarray.get_tel_ids_for_type(telescope))
-
-    return sorted(ids)
 
 
 def get_structure(h5file):
@@ -71,19 +52,26 @@ class TableLoader(Component):
 
     load_dl1_images = traits.Bool(False, help="load extracted images").tag(config=True)
     load_dl1_parameters = traits.Bool(
-        True, help="load reconstructed image parameters").tag(config=True)
+        True, help="load reconstructed image parameters"
+    ).tag(config=True)
     load_dl2_geometry = traits.Bool(
-        False, help="load reconstructed shower geometry information").tag(config=True)
-    load_simulated = traits.Bool(
-        False, help="load simulated shower information").tag(config=True)
-    load_true_images = traits.Bool(
-        False, help="load simulated shower images").tag(config=True)
+        False, help="load reconstructed shower geometry information"
+    ).tag(config=True)
+    load_simulated = traits.Bool(False, help="load simulated shower information").tag(
+        config=True
+    )
+    load_true_images = traits.Bool(False, help="load simulated shower images").tag(
+        config=True
+    )
     load_true_parameters = traits.Bool(
-        False, help="load image parameters obtained from true images").tag(config=True)
-    load_trigger = traits.Bool(
-        True, help="load subarray trigger information").tag(config=True)
+        False, help="load image parameters obtained from true images"
+    ).tag(config=True)
+    load_trigger = traits.Bool(True, help="load subarray trigger information").tag(
+        config=True
+    )
     load_instrument = traits.Bool(
-        False, help="join subarray instrument information to each event").tag(config=True)
+        False, help="join subarray instrument information to each event"
+    ).tag(config=True)
 
     def __init__(self, input_url=None, **kwargs):
         # enable using input_url as posarg
@@ -157,7 +145,7 @@ class TableLoader(Component):
                     table,
                     self.shower_table,
                     keys=["obs_id", "event_id"],
-                    join_type="inner"
+                    join_type="inner",
                 )
             else:
                 table = self.shower_table
@@ -197,11 +185,13 @@ class TableLoader(Component):
 
     def read_telescope_events(self, tel_ids):
 
-        table = vstack([self.read_telescope_events_for_id(tel_id) for tel_id in tel_ids])
+        table = vstack(
+            [self.read_telescope_events_for_id(tel_id) for tel_id in tel_ids]
+        )
 
         return table
 
-    def read_events(self, labels=None):
+    def read_events(self, telescopes=None):
         """Read telescope-based event information.
 
         Parameters
@@ -211,14 +201,14 @@ class TableLoader(Component):
 
         Returns
         -------
-        table: astropy.io.Table
+        events: astropy.io.Table
             Table with primary columns "obs_id", "event_id" and "tel_id".
         """
 
-        if labels is None:
+        if telescopes is None:
             tel_ids = self.subarray.tel.keys()
         else:
-            tel_ids = get_tel_ids(self.subarray, labels)
+            tel_ids = self.subarray.get_tel_ids(telescopes)
 
         if any([self.load_dl1_images, self.load_dl1_parameters, self.load_true_images]):
             table = self.read_telescope_events(tel_ids)
@@ -229,12 +219,12 @@ class TableLoader(Component):
 
         return table
 
-    def read_events_by_tel_type(self, labels=None):
+    def read_events_by_tel_type(self, telescopes=None):
         """Read telescope-based event information.
 
         Parameters
         ----------
-        labels: List[Union[int, str, TelescopeDescription]]
+        telescopes: List[Union[int, str, TelescopeDescription]]
             Any list containing a combination of telescope IDs or telescope_descriptions.
 
         Returns
@@ -244,10 +234,10 @@ class TableLoader(Component):
             with primary columns "obs_id", "event_id".
         """
 
-        if labels is None:
+        if telescopes is None:
             tel_ids = self.subarray.tel.keys()
         else:
-            tel_ids = get_tel_ids(self.subarray, labels)
+            tel_ids = self.subarray.get_tel_ids(telescopes)
 
         selected_subarray = self.subarray.select_subarray(tel_ids)
         selected_tel_types = selected_subarray.telescope_types
@@ -273,11 +263,7 @@ class TableLoader(Component):
             Dictionary of tables organized by telescope types
             with primary columns "obs_id", "event_id".
         """
-
-        if tel_type is None:
-            raise ValueError("Please, specify a telescope description.")
-        else:
-            tel_ids = self.subarray.get_tel_ids_for_type(tel_type)
+        tel_ids = self.subarray.get_tel_ids_for_type(tel_type)
 
         if any([self.load_dl1_images, self.load_dl1_parameters, self.load_true_images]):
             table = self.read_telescope_events(tel_ids)
--- a/ctapipe/io/tableloader.py
+++ b/ctapipe/io/tableloader.py
@@ -9,7 +9,7 @@ import numpy as np
 import tables
 from astropy.table import join, vstack, Table
 
-from ..core import Component, traits
+from ..core import Component, traits, Provenance
 from ..instrument import SubarrayDescription, TelescopeDescription
 from .astropy_helpers import read_table
 
@@ -86,6 +86,8 @@ class TableLoader(Component):
 
         self.subarray = SubarrayDescription.from_hdf(self.input_url)
         self.h5file = tables.open_file(self.input_url, mode="r")
+
+        Provenance().add_input_file(self.input_url, role="Event data")
 
         try:
             self.structure = get_structure(self.h5file)
--- a/ctapipe/io/tableloader.py
+++ b/ctapipe/io/tableloader.py
@@ -55,7 +55,7 @@ class TableLoader(Component):
 
     load_dl1_images = traits.Bool(False).tag(config=True)
     load_dl1_parameters = traits.Bool(True).tag(config=True)
-    load_dl2 = traits.Bool(False).tag(config=True)
+    load_dl2_geometry = traits.Bool(False).tag(config=True)
     load_simulated = traits.Bool(False).tag(config=True)
     load_true_images = traits.Bool(False).tag(config=True)
     load_trigger = traits.Bool(True).tag(config=True)
@@ -109,6 +109,45 @@ class TableLoader(Component):
 
     def __exit__(self, exc_type, exc_value, traceback):
         self.close()
+
+    def read_subarray_events(self, tel_ids=None):
+
+        if self.shower_table is not None and len(table > 0)):
+            table = join(
+                table,
+                self.shower_table,
+                keys=["obs_id", "event_id"],
+                join_type="inner"
+            )
+
+        if self.load_trigger and len(table > 0):
+            table = join(
+                table,
+                self.trigger_table,
+                keys=["obs_id", "event_id"],
+                join_type="inner",
+            )
+        
+        if self.load_dl2_geometry:
+            
+            g = h5file.root[GEOMETRY_GROUP]
+            
+            for reconstructor in g._v_children:
+                
+                geometry = read_table(self.h5file, f"{GEOMETRY_GROUP}/{reconstructor}")
+                
+                # rename DL2 columns to explicit reconstructor
+                # TBD: we could skip this if only 1 reconstructor is present
+                # or simply find another way to deal with multiple reconstructions
+                for col in set(geometry.colnames) - {"obs_id", "event_id"}:
+                    geometry.rename_column(col, f"{reconstructor}_{col}")
+                
+                if (table is None or len(table)==0) and i==0:
+                    table = geometry
+                else:
+                    table = join(
+                        table, geometry, keys=["obs_id", "event_id"], join_type="left"
+                    )
 
     def read_telescope_events(self, tel_ids=None):
         if tel_ids is None:
@@ -165,19 +204,6 @@ class TableLoader(Component):
                     keys=["obs_id", "event_id", "tel_id"],
                     join_type="outer",
                 )
-
-        if self.shower_table is not None and len(table > 0)):
-            table = join(
-                table, self.shower_table, keys=["obs_id", "event_id"], join_type="inner"
-            )
-
-        if self.load_trigger:
-            table = join(
-                table,
-                self.trigger_table,
-                keys=["obs_id", "event_id"],
-                join_type="inner",
-            )
 
         if self.load_instrument and len(table) > 0:
             table = join(
--- a/ctapipe/io/tableloader.py
+++ b/ctapipe/io/tableloader.py
@@ -8,7 +8,7 @@ from ..core import Component, traits
 from ..instrument import SubarrayDescription
 from .astropy_helpers import read_table
 
-__all__ = ["get_structure", "TableLoader"]
+__all__ = ["get_tel_ids", "get_structure", "TableLoader"]
 
 PARAMETERS_GROUP = "/dl1/event/telescope/parameters"
 IMAGES_GROUP = "/dl1/event/telescope/images"
--- a/ctapipe/io/tableloader.py
+++ b/ctapipe/io/tableloader.py
@@ -1,11 +1,12 @@
 import re
-import numpy as np
+from typing import List, Union
 
+import numpy as np
 import tables
 from astropy.table import join, vstack, Table
 
 from ..core import Component, traits
-from ..instrument import SubarrayDescription
+from ..instrument import SubarrayDescription, TelescopeDescription
 from .astropy_helpers import read_table
 
 __all__ = ["get_tel_ids", "get_structure", "TableLoader"]
@@ -19,6 +20,7 @@ TRUE_IMAGES_GROUP = "/simulation/event/telescope/images"
 
 by_id_RE = re.compile(r"tel_\d+")
 
+
 def get_tel_ids(
     subarray: SubarrayDescription,
     telescopes: List[Union[int, str, TelescopeDescription]]
@@ -31,6 +33,7 @@ def get_tel_ids(
         ids.update(subarray.get_tel_ids_for_type(telescope))
 
     return sorted(ids)
+
 
 def get_structure(h5file):
 
@@ -75,8 +78,6 @@ class TableLoader(Component):
         except ValueError:
             self.structure = None
 
-        self.trigger_table = read_table(self.h5file, TRIGGER_TABLE)
-
         self.instrument_table = None
         if self.load_instrument:
             table = self.subarray.to_table()
@@ -110,55 +111,6 @@ class TableLoader(Component):
     def __exit__(self, exc_type, exc_value, traceback):
         self.close()
 
-    def read_subarray_events(self, tel_ids=None):
-
-        if self.shower_table is not None and len(table > 0)):
-            table = join(
-                table,
-                self.shower_table,
-                keys=["obs_id", "event_id"],
-                join_type="inner"
-            )
-
-        if self.load_trigger and len(table > 0):
-            table = join(
-                table,
-                self.trigger_table,
-                keys=["obs_id", "event_id"],
-                join_type="inner",
-            )
-        
-        if self.load_dl2_geometry:
-            
-            g = h5file.root[GEOMETRY_GROUP]
-            
-            for reconstructor in g._v_children:
-                
-                geometry = read_table(self.h5file, f"{GEOMETRY_GROUP}/{reconstructor}")
-                
-                # rename DL2 columns to explicit reconstructor
-                # TBD: we could skip this if only 1 reconstructor is present
-                # or simply find another way to deal with multiple reconstructions
-                for col in set(geometry.colnames) - {"obs_id", "event_id"}:
-                    geometry.rename_column(col, f"{reconstructor}_{col}")
-                
-                if (table is None or len(table)==0) and i==0:
-                    table = geometry
-                else:
-                    table = join(
-                        table, geometry, keys=["obs_id", "event_id"], join_type="left"
-                    )
-
-    def read_telescope_events(self, tel_ids=None):
-        if tel_ids is None:
-            tel_ids = self.subarray.tel.keys()
-
-        return vstack([self.read_telescope_events_for_id(tel_id) for tel_id in tel_ids])
-
-    def read_telescope_events_for_type(self, tel_type):
-        tel_ids = self.subarray.get_tel_ids_for_type(tel_type)
-        return self.read_telescope_events(tel_ids)
-
     def _read_telescope_table(self, group, tel_id):
         if self.structure == "by_id":
             key = f"{group}/tel_{tel_id:03d}"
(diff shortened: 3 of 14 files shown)


ctapipe/io/tableloader.py
```diff
@@ -0,0 +1,117 @@
...
+from ..instrument import SubarrayDescription
+from .astropy_helpers import read_table
+
+PARAMETERS_GROUP = "/dl1/event/telescope/parameters"
+IMAGES_GROUP = "/dl1/event/telescope/images"
+TRIGGER_TABLE = "/dl1/event/subarray/trigger"
+
+by_id_RE = re.compile(r"tel_\d+")
+
+
```
HealthyPear, line 16: `__all__ = ["TableLoader"]`

and also in `ctapipe.io__init__ `

HealthyPear (changes requested):

HealthyPear:
### Updates

I have improved the current PR with

- loading DL2 shower geometry 
- refactoring to ensure loading of any single table and of every (almost, see below) combination
- added a generic call function that makes the usage blind to dataset split mode 

### Caveats

- loading of trigger + instrument cannot work for now because they refer to different things...
- I read shower geometry by pre-pending the name of the reconstruction algorithm in case there is more than 1, perhaps in the case of only 1 it can be avoided

ctapipe/io/tableloader.py
HealthyPear: fixed

ctapipe/io/tableloader.py
```diff
@@ -0,0 +1,331 @@
...
+            self.log.critical("No data enabled for loading.")
+            self.close()
+            return None # it should probably return some exit status != 0
+
+        if (sum(self.all_loads)>1) and self.load_instrument and (not any([self.load_dl1_images, self.load_dl1_parameters, self.load_true_images])):
+            self.log.critical("""Instrument data requires tel_id-based quantities, 
+            but only event_id-based quantities have been selected together with it.""")
+            self.log.critical("No data enabled for loading.")
+            self.close()
+            return None # it should probably return some exit status != 0
```
maxnoe, line 174: This is not a command line tool. Raise exception on errors. Don't use logging or return codes.

ctapipe/io/tableloader.py
```diff
@@ -0,0 +1,331 @@
...
+        key = f"{group}/{reconstructor}"
+
+        if key in self.h5file:
+            table = read_table(self.h5file, key, condition=None)
+        else:
+            table = Table()
+
+        return table
+
+    def read_data(self, labels=None, sort_by=["obs_id", "event_id"]):
```
maxnoe, line 148: Call this `read_subarray_events`

HealthyPear: done

ctapipe/io/tableloader.py
HealthyPear: Ok for the raises, but why not logging?

Also, what about ``def __exit__(self, exc_type, exc_value, traceback)``?
We don't seem to use the arguments, we just close the file, should we modify/use this or simplify it as only ``def __exit__(self)``?

HealthyPear: done

ctapipe/io/tableloader.py
```diff
@@ -162,16 +162,14 @@ def read_subarray_events(self, labels=None, sort_by=["obs_id", "event_id"]):
             Single table containing the enabled data information."""
         
         if not any(self.all_loads):
-            self.log.critical("No data enabled for loading.")
             self.close()
-            return None # it should probably return some exit status != 0
+            raise ValueError("No data enabled for loading.")
 
         if (sum(self.all_loads)>1) and self.load_instrument and (not any([self.load_dl1_images, self.load_dl1_parameters, self.load_true_images])):
```
maxnoe, line 168: No need to check `sum > 1` since you did that in the previous `if not any` statement

HealthyPear: removed

ctapipe/io/tableloader.py
kosack: >  Also, what about def __exit__(self, exc_type, exc_value, traceback)?

Context manager expect those arguments as part of their protocol, even if you don't use them.  I don't think you are free to remove them

ctapipe/io/tableloader.py
```diff
@@ -119,9 +134,173 @@ def _read_telescope_table(self, group, tel_id):
...
+        
+        return table_dict
+
+    def read_telescope_events(self, tel_ids=None):
+        if tel_ids is None:
+            tel_ids = self.subarray.tel.keys()
+        
         table = None
+        
+        # Joini all the DL1-based data, which depends on the triggered tel_ids
```
kosack, line 230: typo

ctapipe/io/tableloader.py
maxnoe: > Ok for the raises, but why not logging?

Because that's double communication. Logging should happen when desired by the code *catching* the exception, not the code *raising* it. We don't do logging for raised exceptions anywhere else.

maxnoe: > def __exit__(self, exc_type, exc_value, traceback)

This is required by the python data model. You cannot change the signature of the magic methods, since python will call `__exit__` with those arguments, even when they are not used inside:

```
class BadContext:
    def __enter__(self):
        return self
    def __exit__(self):
        pass

with BadContext():
    pass
```

Results in:

```
<ipython-input-1-cba4f71839bc> in <module>
      6 
      7 with BadContext():
----> 8     pass
      9 

TypeError: __exit__() takes 1 positional argument but 4 were given```

File: ctapipe/io/tableloader.py [hunk 6]
```diff
@@ -119,9 +134,173 @@ def _read_telescope_table(self, group, tel_id):
...
+        if (self.load_dl1_images or self.load_dl1_parameters or self.load_true_images) and ((table is None) or (len(table)==0)):
+            table = vstack([self.read_telescope_events_for_id(tel_id) for tel_id in tel_ids])
+        elif (self.load_dl1_images or self.load_dl1_parameters or self.load_true_images) and (len(table) > 0):
+            table = join(
+                table,
+                vstack([self.read_telescope_events_for_id(tel_id, table) for tel_id in tel_ids]),
+                keys=["tel_id"],
+                join_type="inner",
```
kosack, line 238: Are you sure you always want *inner* joins? That would skip rows where one or the other is missing, rather than keep the existing one and give Null values for the other.  It may be ok, but I wonder if there might be cases where we would silently drop events with this.

ctapipe/io/tableloader.py
```diff
@@ -0,0 +1,330 @@
...
+            Columns to sort the final table.
+            By default it is sorted by obs_id and event_id.
+        
+        Returns
+        -------
+        output: astropy.Table
+            Single table containing the enabled data information."""
+        
+        if not any(self.all_loads):
+            self.close()
```
maxnoe, line 165: Don't close! This will prevent people from changing the options and runnig read_subarray_events again.

Close will by called when using this class in a context or when it goes out of scope automatically. There is no need to use close manually anywhere else

ctapipe/io/tableloader.py
```diff
@@ -0,0 +1,330 @@
...
+            raise ValueError("No data enabled for loading.")
+
+        if self.load_instrument and (not any([self.load_dl1_images, self.load_dl1_parameters, self.load_true_images])):
+            self.close()
```
maxnoe, line 169: remove

ctapipe/io/tableloader.py
```diff
@@ -119,9 +134,173 @@ def _read_telescope_table(self, group, tel_id):
...
+            # get reconstructors in file (if more than one)
+            reconstructors = self._get_reconstructors()
+                        
+            for i, reconstructor in enumerate(reconstructors):
+                
+                geometry = self._read_telescope_DL2_geometry(GEOMETRY_GROUP, reconstructor)
+            
+                # rename DL2 columns to explicit reconstructor
+                # TBD: we could skip this if only 1 reconstructor is present
+                # or simply find another way to deal with multiple reconstructions
```
kosack, line 264: I think it's ok to always prepend the reconstructor name.  That way there is no confusion about which information is being used.   In the DL2→DL3 transition, we reduce this to a single reconstruction, but not in DL2 where all tools should assume there are many.

ctapipe/io/tableloader.py
```diff
@@ -0,0 +1,330 @@
...
+            Single table containing the enabled data information."""
+        
+        if not any(self.all_loads):
+            self.close()
+            raise ValueError("No data enabled for loading.")
+
+        if self.load_instrument and (not any([self.load_dl1_images, self.load_dl1_parameters, self.load_true_images])):
+            self.close()
+            raise ValueError("""Instrument data requires tel_id-based quantities, 
+            but only event_id-based quantities have been selected together with it.
```
maxnoe, line 171: This results in strange indentation inside the message. Use implicit concatenation instead:

```
raise ValueError(
    "Instrument data requires tel_id-based quantities,"
    " but only event_id-based quantities have been selected together with it."
)
```
```

File: ctapipe/io/tableloader.py [hunk 11]
```diff
@@ -0,0 +1,330 @@
...
+
+        return table
+
+    def read_subarray_events(self, labels=None, sort_by=["obs_id", "event_id"]):
+        """Read data from file.
+        
+        Parameters
+        ----------
+        labels: list
+            Either tel_ids or tel_types depending on dataset split.
```
maxnoe, line 154: Why use a double meaning here? For the telescope events we have `by_telescope_type` and `by_telescope_id`.

Also, why is this needed for *subarray* events?

ctapipe/io/tableloader.py
```diff
@@ -0,0 +1,330 @@
...
+        key = f"{group}/{reconstructor}"
+
+        if key in self.h5file:
+            table = read_table(self.h5file, key, condition=None)
+        else:
+            table = Table()
+
+        return table
+
+    def read_subarray_events(self, labels=None, sort_by=["obs_id", "event_id"]):
```
maxnoe, line 148: This function does not read `subarray` events. It reads telescope events? 

What is the intention here? What does this provide over `read_telescope_events`?

ctapipe/io/tableloader.py
HealthyPear: because this function does all, that's why it was called more generically `read_data`
you can provide either tel_ids or tel_types regardless of the input file dataset splitting mode

you do not need to use any of the specific functions by*** (which in fact we could make private)

ctapipe/io/tableloader.py
```diff
@@ -0,0 +1,330 @@
...
+        sort_by: list
+            Columns to sort the final table.
+            By default it is sorted by obs_id and event_id.
+        
+        Returns
+        -------
+        output: astropy.Table
+            Single table containing the enabled data information."""
+        
+        if not any(self.all_loads):
```
maxnoe, line 164: This can already be checked in `__init__`

ctapipe/io/tableloader.py
HealthyPear: see above

ctapipe/io/tableloader.py
```diff
@@ -0,0 +1,330 @@
...
+            self.close()
+            raise ValueError("No data enabled for loading.")
+
+        if self.load_instrument and (not any([self.load_dl1_images, self.load_dl1_parameters, self.load_true_images])):
```
maxnoe, line 168: In my implementation, `load_instrument` was an option to enable joining instrument info onto the other information. I think this makes sense, since if someone is only interested in instrument information, we have `SubarrayDescription` also accessible from this loader.

HealthyPear: sure this is still possible, it's just something more I felt we should add if you allow to read any single table

ctapipe/io/tableloader.py
maxnoe, line 148: The return type of this function depends on the structure of the file. I think this is not good, since then the code needs to know it. I implemented the functions so they give a well defined output independent of the structure in the file, so that code can be agnostic about that. I think this is an important property we should keep!

ctapipe/io/tableloader.py
```diff
@@ -0,0 +1,330 @@
...
+                join_type="inner",
+            )
+
+        return table
+
+    def read_telescope_events_for_type(self, tel_type):
+        tel_ids = self.subarray.get_tel_ids_for_type(tel_type)
+        return self.read_telescope_events(tel_ids)
+
+    def read_telescope_events_for_id(self, tel_id, table=None):
```
maxnoe, line 300: What is the `table` argument for?

ctapipe/io/tableloader.py
```diff
@@ -0,0 +1,330 @@
...
+
+        if self.load_dl1_parameters:
+            table = self._read_telescope_table(PARAMETERS_GROUP, tel_id)
```
maxnoe, line 303: This overwrites `table` without checking if it is None. For what is the `table` argument?

ctapipe/io/tableloader.py
```diff
@@ -0,0 +1,330 @@
...
+        return table_dict
+
+    def read_telescope_events(self, tel_ids=None):
+        if tel_ids is None:
+            tel_ids = self.subarray.tel.keys()
+        
+        table = None
+        
+        # Joini all the DL1-based data, which depends on the triggered tel_ids
+        if (self.load_dl1_images or self.load_dl1_parameters or self.load_true_images) and ((table is None) or (len(table)==0)):
```
maxnoe, line 231: These if / elif statements here are not necessary / much more complicated than needed since table is always `None` here, you define it in the line just above.

ctapipe/io/tableloader.py
HealthyPear: Not sure if I get what is the problem here (which should be easily fixable anyway, since we have all the tools here).

I don't understand "since then the code needs to know it."
The code will understand via the conditions what to do and will return what I will write in the docstring for the user (sorry I didn't update it)

HealthyPear: the point is exactly to make the code check the dataset split mode so I don't have to worry about it and with 1 function I can ask for anything (available) regardless of the input file nature

ctapipe/io/tableloader.py
kosack: We can probably drop the load_instrument option: it's only for a very specific use case when doing some debugging plots, and the info is already in the subarray description.  The user can always join it manually if needed - it adds a lot of redundant information (e.g. the telescope position for each event), which normally you don't need.

HealthyPear: that I would be in favor of (since we have `SubarrayDescription` already) but maybe it's Max should say if he wants to keep it since he added it in the first place, maybe he had some reason add this support here

ctapipe/io/tableloader.py
maxnoe: That's not good. We should have a clear api without these double meanings.

ctapipe/io/tableloader.py
HealthyPear: of course you want it together with some other info like he said above, but if you put it as its separate traitlet I (as a user) would expect to be able to export it alone if I wish

another option is to always attach it (are there cases in which we do NOT expect this info to be in a file?)

maxnoe: > We can probably drop the load_instrument option: it's only for a very specific use case when doing some debugging plots, and the info is already in the subarray description. 

No, this is also useful for e.g. building a single machine learning model for multiple telescope types (using focal length / mirror area as features)

ctapipe/io/tableloader.py
HealthyPear: I should be able to split it easily in 2, but I don't see the problem honestly 

you write in the docstring parameters "Either list of tel_ids or list of tel_types" and similar for the returns...

kosack:
This has now gotten pretty complex and hard to follow. My implementation  (on which @HealthyPear's was based) was also pretty ugly, since it was cobbled together during the workshop, and I always intended to throw it away and start again with a cleaner version. I really think we should revert back to the version @maxnoe  made here, which is cleaner:

https://github.com/cta-observatory/ctapipe/pull/1771/commits/4a9f8067073ef2dd7735e4e837667af6cf68e838

And just add the ability to load DL2 info following that form.  Otherwise it is going to be really difficult to maintain.

ctapipe/io/tableloader.py
maxnoe: I am talking about the code calling this one. The structure of the return value of this function depends on the structure in the file, correct?

maxnoe:
@kosack I agree. I just thought @HealthyPear would add two/three little functions to add the dl2 loading, not change the whole thing

ctapipe/io/tableloader.py
HealthyPear: not really, you can read as tel_types from a tel_id file as well as tel_ids from a tel_type file

that is why I did it like this: the return value depends only on the labels (tel_ids or tel_type) that the user asks for.
The usage is completely blind to the dataset split (as it should since it doesn't change the contents that you read)

ctapipe/io/tableloader.py
maxnoe: So if this is basically just `read_telescope_events_by_id` if labels is a list of ids and `read_telescope_events_by_type` if `labels` is a list of types, why is this function so complicated???

ctapipe/io/tableloader.py
HealthyPear: removed

ctapipe/io/tableloader.py
HealthyPear: removed

ctapipe/io/tableloader.py
HealthyPear: removed

ctapipe/io/tableloader.py
HealthyPear: I moved both this and the one checking for instrument data + dl2 data

ctapipe/io/tableloader.py
HealthyPear: I wrote a proper docstring, it should be fool-proof now

ctapipe/io/tableloader.py
HealthyPear: need to check thanks, for now it seems to work...

ctapipe/io/tableloader.py
HealthyPear: fixed thanks

ctapipe/io/tableloader.py
HealthyPear: fixed

ctapipe/io/tableloader.py
HealthyPear: done

ctapipe/io/tableloader.py
HealthyPear: removes

ctapipe/io/tableloader.py
HealthyPear: actually, it was right before, so I reverted it (in fact the unit test was failing again):
the >1 ensures that there is _at least_ `self.load_instrument`, since it will crash if joining it with any combination of event-based data (e.g. trigger+ simulation, dl2 alone, dl2+ trigger, etc...)

ctapipe/io/tableloader.py
HealthyPear: this should be OK now (I left __exit__)

HealthyPear (approved):

ctapipe/io/tableloader.py
HealthyPear: as it is now one can load either only instrument data or attach it to a table containing `tel_id` as a column name (so if at least one between `load_dl1_images`, `load_dl1_parameters` and `load_true_images` has been set to True) resulting in a raised `ValueError` if only `event_id`-based data is loaded with instrument data

ctapipe/io/tableloader.py
HealthyPear: if we keep it like this I guess that in the next tool we need to strip somewhat these prefixes or add a dictionary of supported reconstructions to read them

ctapipe/io/tableloader.py
HealthyPear: perhaps you can check out this PR and test it with `ctapipe-process` in case I messed something

ctapipe/io/tableloader.py
HealthyPear: > So if this is basically just `read_telescope_events_by_id` if labels is a list of ids and `read_telescope_events_by_type` if `labels` is a list of types

No, because it also accounts for the case in which I read from a file (regardless of how it has been split at writing time) and I provide a list of tel_ids that span different tel_types and I still want them to result in tel-type based tables

ctapipe/io/tableloader.py
```diff
@@ -0,0 +1,175 @@
...
+                table = true_images
+            else:
+                table = join(
+                    table,
+                    true_images,
+                    keys=["obs_id", "event_id", "tel_id"],
+                    join_type="outer",
+                )
+
+        if self.shower_table is not None:
```
maxnoe, line 157: missing `and len(table > 0)` here

ctapipe/io/tableloader.py
```diff
@@ -0,0 +1,175 @@
...
+
+        self.subarray = SubarrayDescription.from_hdf(self.input_url)
+        self.h5file = tables.open_file(self.input_url, mode="r")
+
+        try:
+            self.structure = get_structure(self.h5file)
+        except ValueError:
+            self.structure = None
+
+        self.trigger_table = read_table(self.h5file, TRIGGER_TABLE)
```
maxnoe, line 66: trigger table is always loaded, should only be loaded when `self.load_trigger is True`

ctapipe/io/tableloader.py
HealthyPear: fixed

ctapipe/io/tableloader.py
HealthyPear: fixed

ctapipe/io/tableloader.py
HealthyPear: reset to old commit - resolving

ctapipe/io/tableloader.py
HealthyPear: this disussion is still relevant I guess

ctapipe/io/tableloader.py
```diff
@@ -166,12 +166,12 @@ def read_telescope_events_for_id(self, tel_id):
...
                 )
 
-        if self.shower_table is not None:
+        if self.shower_table is not None and len(table > 0)):
             table = join(
                 table, self.shower_table, keys=["obs_id", "event_id"], join_type="inner"
             )
 
-        if self.load_trigger and len(table) > 0:
+        if self.load_trigger:
```
maxnoe, line 174: Why remove the length check? This errors when `len(table) == 0`, astropy cannot join empty tables which is the  reason the check was there

ctapipe/io/tableloader.py
```diff
@@ -0,0 +1,213 @@
...
+
+    def __enter__(self):
+        return self
+
+    def __exit__(self, exc_type, exc_value, traceback):
+        self.close()
+
+    def read_subarray_events(self, tel_ids=None):
+
+        if self.shower_table is not None and len(table > 0)):
```
maxnoe, line 115: There is no `table` here

ctapipe/io/tableloader.py
```diff
@@ -8,7 +8,7 @@
 from ..instrument import SubarrayDescription
 from .astropy_helpers import read_table
 
-__all__ = ["get_structure", "TableLoader"]
+__all__ = ["get_tel_ids", "get_structure", "TableLoader"]
```
maxnoe, line 11: If we  make this a public method (as opposed to just an internal function) I think this should be a method of `SubarrayDescription`.

HealthyPear: Agreed - I put it here now to work on the unit tests, but we can move it later in any case

ctapipe/io/tableloader.py
HealthyPear: reverted

ctapipe/io/tableloader.py
HealthyPear: now there is

ctapipe/io/tableloader.py
HealthyPear: reverted

ctapipe/io/tableloader.py
HealthyPear: done

ctapipe/io/tableloader.py
```diff
@@ -0,0 +1,308 @@
...
+        return table
+
+    def read_telescope_events(self, tel_ids):
+
+        table = vstack([self.read_telescope_events_for_id(tel_id) for tel_id in tel_ids])
+
+        return table
+
+    def read_events(self, labels=None):
+        """Read telescope-based event information.
```
kosack, line 182: Why not read_telescope_events? (the other is `read_subarray_events`).  this confused me, since i expected something only called events to load the shower events, not the telescope ones.

ctapipe/io/tableloader.py
```diff
@@ -0,0 +1,308 @@
...
+
+        Parameters
+        ----------
+        labels: list
+            Any list combination of tel_ids, tel_types, or telescope_descriptions.
```
kosack, line 187: I naively tried `loader.read_events(["tel_ids"])`, assuming I would get multiple tables or something, which failed.  Better docstring is needed.

ctapipe/io/tableloader.py
kosack: also what happens here if I mix two telescope types and have `load_images=True`?  If it fails, this should be explained that it doesn't work for mixed cameras if the number of pixels is different.   Or perhaps this could be a private method and we only expose `read_events_by_tel_type`?

ctapipe/io/tableloader.py
```diff
@@ -0,0 +1,308 @@
...
+        selected_tel_types = selected_subarray.telescope_types
+
+        tables = {}
+        for tel_type in selected_tel_types:
+
+            tables[str(tel_type)] = self.read_events_for_type(tel_type)
+
+        return tables
+
+    def read_events_for_type(self, tel_type):
```
kosack, line 239: Shuold this be public? It confused me what the difference between this and he `read_events_per_tel_type` was (until I tried it).  I think only the latter should be public, and this could be hidden  as private to simplify the API.

ctapipe/io/tableloader.py
```diff
@@ -0,0 +1,308 @@
...
+                optics,
+                keys="optics_index",
+                # conflicts for TAB_VER, TAB_TYPE, not needed here, ignore
+                metadata_conflicts="silent",
+            )
+
+            table.remove_columns(["optics_index", "camera_index"])
+            self.instrument_table = table
+
+        self.shower_table = None
```
kosack, line 98: Why is `shower_table` a public attribute?

ctapipe/io/tableloader.py
```diff
@@ -0,0 +1,308 @@
...
+
+    key = next(iter(g._v_children))
+    if by_id_RE.fullmatch(key):
+        return "by_id"
+    else:
+        return "by_type"
+
+
+class TableLoader(Component):
+    """ A helper class to load and join tables from a DL1 file"""
```
kosack, line 55: Adding a simple example here, or mentioning the main API functions that one would call would help a lot.  It took me a few tries to understand how it works.

kosack (changes requested):
Looks cleaner now.  Other than some small comments (see inline), I think the only things left are:

- Missing documentation: it would be good to have a short tutorial added to the docs on using this.  It will likely be the main access point for people working with ctapipe data.  
- We should also update the`ctapipe.io` docs page to sumamrize the main ways to access data: 
    - event-by-event: EventSource
    - all events at once: TableLoader
    -  just instrument info: SubarrayDescription.from_hdf()

ctapipe/io/tableloader.py
kosack: Since this appears in the HTML repr, it should be very concise, so maybe not a full example here (in the class docstring may be better).  Also it's not a DL1 file... but any ctapipe output file.

ctapipe/io/tableloader.py
HealthyPear: we have now both `read_telescope_events`, which gives tables based on single telescope data and `read_subarray_events` which reads only subarray data

`read_events` is simply the general function used to call the first (if DL1 loadings have been enabled) and then the 2nd (if subarray loadings have been enabled

HealthyPear: yes it fails if we try to stack images of different sizes

do we have a nicer solution for this?

ctapipe/io/tableloader.py
HealthyPear: you don't get multiple tables as you are only giving tel_ids (why a string though? it shouldn't work...)

if you use read_events_by_type you get a table for each tel_id

also the docstring says clearly 

```
Returns
        -------
        table: astropy.io.Table
            Table ...
```

as in 1 Table :)

ctapipe/io/tableloader.py
HealthyPear: I think it should, it's still valid to use only 1 tel_type, though if you do

read_event_by_tel_type(**1 tel type**)

you get the same thing

it's the same for me, we can make it private or not

kosack: it was just to simplify the API, it's not a huge issue to keep it.  I just found that so many ways to read the data were confusing, and there is not an equivalent one for tel_id, but both are handled properly by the higher-level function, so maybe it's better just to expose that.

ctapipe/io/tableloader.py
kosack: that's the point: I followed the docstring, and it didn't work.  I assumed a list with combination of the strings "tel_types", "tel_ids", which is of course totallly wrong, but it's not clear from the docstring.  (I basically tried using the class without any prior knowledge of how i was implemented inside, and failed)

kosack: That may be a case to describe it more formally:
`labels: Union[List[int], List[TelescopeDescription]]`

kosack: (also what is the difference between `tel_type` and `tel_description`? They are both representations of the same thing. A TelescopeDescription can be specified by it's class or string representation, so no need to differentiate.

HealthyPear: It seems a very strict interpretation of the usage to me but anyway... :)

I will explicit the types of the list elements

HealthyPear: > That may be a case to describe it more formally:
> `labels: Union[List[int], List[TelescopeDescription]]`

ah yes like you mentioned in the issue about type checking OK

> A TelescopeDescription can be specified by it's class or string representation, so no need to differentiate.

OK

ctapipe/io/tableloader.py
kosack: there is already a `SubarrayDescripton.get_tel_ids_for_type()`, why not use that?

HealthyPear: that gives all the tel_ids, while this function extracts them from any subset defined by something like `Union[List[int], List[TelescopeDescription]]`

HealthyPear: so we can always get them in case someone does e.g. read_events([1,2,"MST_MST_NectarCam"]) or any other combination

ctapipe/io/tableloader.py
maxnoe: > `labels: Union[List[int], List[TelescopeDescription]]`

Actually it is `List[Union[int, str, TelescopeDescription]]` like in the above `get_tel_ids` function (that we should move to `SubarrayDescription`

maxnoe: If we take the type hint seriously, str needs to be there, otherwise tools will complain about `read_telescope_events(['LST_LST_LSTCam'])

ctapipe/io/tableloader.py
HealthyPear: are you referring to a function doing `by_tel_id`?
for tel_id I don't think it makes sense to have an analog one, as you can use `for_tel_id` to get 1 table for that specific tel_id, whereas it makes sense for tel_type because it keeps together the tel_ids of that tel_type

ctapipe/io/tableloader.py
kosack: Ah, I see it's more general.

ctapipe/io/tableloader.py
kosack: There is no easy solution.  you could just make this private since it is a helper functions used by the higher-level functions. Otherwise, you need to add a check and raise a friendly exception explaining why it doesn't work.

ctapipe/io/tableloader.py
kosack: Ah, it was really not clear that you could mix tel IDs and descriptions in the list.  Adding the correct type annotation helps a lot here.

ctapipe/io/tableloader.py
```diff
@@ -0,0 +1,308 @@
...
+    load_instrument = traits.Bool(False).tag(config=True)
+
+    def __init__(self, input_url=None, **kwargs):
+        # enable using input_url as posarg
+        if input_url not in {None, traits.Undefined}:
+            kwargs["input_url"] = input_url
+        super().__init__(**kwargs)
+
+        self.subarray = SubarrayDescription.from_hdf(self.input_url)
+        self.h5file = tables.open_file(self.input_url, mode="r")
```
kosack, line 74: should add the following after this line:

```py3
Provenance().add_input_file(self.input_url, role="Event data") 
```

ctapipe/io/tableloader.py
HealthyPear: well it doesn't matter if you make it a private method no? if someone wants a single-table with true images from different cameras it will always fail

I think we should raise an error (and also specify it in the docs with a disclaimer)

did I misunderstood what you meant?

ctapipe/io/tableloader.py
```diff
@@ -0,0 +1,308 @@
...
+
+by_id_RE = re.compile(r"tel_\d+")
+
+
+def get_tel_ids(
+    subarray: SubarrayDescription,
+    telescopes: List[Union[int, str, TelescopeDescription]]
+) -> List[int]:
+    ids = set()
+
```
kosack, line 29: might want to also check if a string is given that its a valid TelescopeDescription string to ensure it's not a typo.  Right now, it will silently ignore any mis-spelled telescope strings.

kosack:
One more suggestion: since many of these operations can take a very long time on large files, having some optional user feedback would be nice for when it is used interactively.  In the original implementation there was a progress bar, which could be added in the longer loops over telescopes/types/etc.

HealthyPear:
> One more suggestion: since many of these operations can take a very long time on large files [...]

I think we should make a new PR for the chunk-based reading and try to apply it to `TableLoader`, that would unlock all cases I think

ctapipe/io/tableloader.py
```diff
@@ -0,0 +1,308 @@
...
+            table = self.read_telescope_events(tel_ids)
+        else:
+            table = None
+
+        table = self.read_subarray_events(table)
+
+        return table
+
+    def read_telescope_events_for_id(self, tel_id):
+
```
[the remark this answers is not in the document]
HealthyPear: added

ctapipe/io/tableloader.py
```diff
@@ -0,0 +1,308 @@
...
+            tables[str(tel_type)] = self.read_events_for_type(tel_type)
+
+        return tables
+
+    def read_events_for_type(self, tel_type):
+        """Read telescope-based event information.
+
+        Parameters
+        ----------
+        labels: list
```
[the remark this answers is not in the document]
HealthyPear: fixed

ctapipe/io/tableloader.py
```diff
@@ -0,0 +1,308 @@
...
+
+    input_url = traits.Path(directory_ok=False, exists=True).tag(config=True)
+
+    load_dl1_images = traits.Bool(False).tag(config=True)
+    load_dl1_parameters = traits.Bool(True).tag(config=True)
+    load_dl2_geometry = traits.Bool(False).tag(config=True)
+    load_simulated = traits.Bool(False).tag(config=True)
+    load_true_images = traits.Bool(False).tag(config=True)
+    load_trigger = traits.Bool(True).tag(config=True)
+    load_instrument = traits.Bool(False).tag(config=True)
```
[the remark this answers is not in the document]
HealthyPear: added

ctapipe/io/__init__.py
```diff
@@ -2,6 +2,7 @@
 from .eventsource import EventSource
```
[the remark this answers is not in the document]
HealthyPear: added

ctapipe/io/tableloader.py
HealthyPear: done - let me know if it's enough since it cannot be very long (we have also the docs anyway - I don't think the general user care much about the class docstring...)

ctapipe/io/tableloader.py
HealthyPear: added

ctapipe/io/tableloader.py
kosack: well, if it's private it is not part of the public API, so users should not use it.  Instead they would use functions that call it..  But keeping it and adding an error handler is also fine.

kosack:
> I think we should make a new PR for the chunk-based reading and try to apply it to TableLoader, that would unlock all cases I think

Ok, agreed when it is chunk based, it would be much easier to add a progress function.

ctapipe/io/tableloader.py
```diff
@@ -64,13 +64,19 @@ class TableLoader(Component):
...
-    load_instrument = traits.Bool(False).tag(config=True)
+    load_dl1_images = traits.Bool(False, help="load extracted images").tag(config=True)
+    load_dl1_parameters = traits.Bool(
+        True, help="load reconstructed image parameters").tag(config=True)
+    load_dl2_geometry = traits.Bool(
+        False, help="load reconstructed shower geometry information").tag(config=True)
+    load_simulated = traits.Bool(
+        False, help="load simulated shower information").tag(config=True)
+    load_true_images = traits.Bool(
+        False, help="load simulated shower images").tag(config=True)
```
kosack, line 75: does this include also the simulated shower image parameters?

HealthyPear: `load_true_images` means loading the simulated images, `load_simulated` means  loading the simulated shower information

the fact that I needed to specify it means that perhaps we should rename `load_simulated` into `load_simulated_showers` :)

ctapipe/io/tableloader.py
HealthyPear: done

ctapipe/io/tableloader.py
HealthyPear: I explicited the input types in the docstrings that use the generic `label` as input

ctapipe/io/tableloader.py
kosack: but where are the simulated parameters?  It is still not clear which options loads them.  They are not the same as the images.

maxnoe: Not yet implement I think

HealthyPear: we don't have them I think - that would be another traitlet `load_true_parameters`

HealthyPear: ah no sorry we have them, but we should make it a different traitlet indeed

HealthyPear: Done.

ctapipe/io/tableloader.py
HealthyPear: Indeed, we could move it directly in `read_subarray_event` - anything against this, @maxnoe ?

ctapipe/io/tableloader.py
kosack: I agree, my suggestion was just that if we hide this it is a simpler API, but I'm ok keeping it.

ctapipe/io/tableloader.py
kosack: or if it is really needed, make it `self._shower_table` since generally you don't want the users relying on it.

kosack (approved):

maxnoe:
I'd still like to do some small refactoring on this, probably will make it on monday

maxnoe:
Ok, I came finally around to doing this. I think the code is much cleaner now, especially by itroducing the `_allow_empty_join` function, much of the complicated checking if a table is already there is gone.

There are now only three public methods:

`read_subarray_events`, `read_telescope_events` and `read_telescope_events_by_type`. 

The latter two support the use cases of the previous many functions by having this `List[Union[str, int, TelescopeDescription]]` as argument.

kosack (approved):
Look good! Much simpler, and works in all my use cases.

LukasNickel (commented):
There are a lot of codacy complaints. Some are false positives, some are minor issues. I dont see any of them as severe, but want to bring up the topic before approving.

kosack (changes requested):
I somehow missed all the codacy complaints.  I agree with @LukasNickel that most are minor, but a few should be addressed.  It would be good to at least: 
- document `get_structure` (it is a public function, exported from the module in `__all__`, and is really not clear at all what it does from the name)
- document `read_subarray_events`
- probably want a `# pylint: disable= protected-access` at the top of any files that uses low-level hdf5 commands (like `_v_*`, which give false-positive warnings)
- don't use the variable name `tables` (used a few times), since it shadows the `tables` module, and that could lead to bugs in the future
- either disable the warning, or document each test with at least a few words about what the test does: e.g.
```py3
def test_read_subarray_events(test_file_dl2):
     """ check that we get some expected columns when reading a file with DL2 """
```
 - It's not clear to my why we suddenly are getting `redefined-outer-name` warnings in the tests
 - also not sure why we get a `List[TelescopeDescription] is not iterable` warning here: (aren't lists iterable?)
 
```py3
valid_tel_types = {str(tel_type) for tel_type in self.telescope_types}
```

maxnoe:
@kosack I think I addressed most/all relevant codacy issues

maxnoe:
Ok, now the docs should be fixed as well, at least with regard to table loader and processor tool, for the rest I opened #1786

kosack (approved):

.pylintrc
```diff
@@ -425,6 +425,7 @@ good-names=i,
            x,
            y,
            n,
+		   f,
```
LukasNickel, line 428: The indentation seems off here, tab vs spaces?

LukasNickel (approved):

PR #1776, opened 2021-08-05
Add flags to enforce recomputation of dl1/dl2

LukasNickel:
#1726 removed the possibility to rerun stage1/process with the goal of recomputing the DL1 information on a file with both DL1a and DL1b.
This can be useful to test different cleanings or add noise to the images (#1723).

Could be done without the `if` clauses adding another `or`  aswell. I just felt like its the easiest to understand (especially if the compute logic changes at some point.)

### DIFF
--- a/ctapipe/tools/process.py
+++ b/ctapipe/tools/process.py
@@ -44,6 +44,12 @@ class ProcessorTool(Tool):
     """
 
     progress_bar = Bool(help="show progress bar during processing").tag(config=True)
+    force_recompute_dl1 = Bool(
+        help="Enforce dl1 recomputation even if already present in the input file"
+    ).tag(config=True)
+    force_recompute_dl2 = Bool(
+        help="Enforce dl2 recomputation even if already present in the input file"
+    ).tag(config=True)
 
     aliases = {
         ("i", "input"): "EventSource.input_url",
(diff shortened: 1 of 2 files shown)


HealthyPear:
I think we should specify **DL1b** here, since the input file is one that comes from `ctapipe-process` and not the `simtel` file - unless I misunderstood the context

LukasNickel:
> I think we should specify **DL1b** here, since the input file is one that comes from `ctapipe-process` and not the `simtel` file - unless I misunderstood the context

You mean recomputing DL1b?
I would like to recompute DL1a as well and the code makes no difference: Both are done in the same call to the `ImageProcessor`

ctapipe/tools/process.py
```diff
@@ -70,6 +76,18 @@ class ProcessorTool(Tool):
             "show a progress bar during event processing",
             "don't show a progress bar during event processing",
         ),
+        **flag(
+            "recompute_dl1",
```
kosack, line 80: I suggest calling this  alias `--recompute-dl1` (using a dash not an underscore) as we do with other command-line arguments.  Same for `--recompute-dl2`

kosack (changes requested):
other than a small naming change, this looks fine.

maxnoe:
> I would like to recompute DL1a as well and the code makes no difference: Both are done in the same call to the ImageProcessor

How can you *re*compute DL1a? If it is a simtel file, you always need to calculate it. If it is a DL1 file, you cannot recompute it, since the DL0 is not there.

LukasNickel:
> > I would like to recompute DL1a as well and the code makes no difference: Both are done in the same call to the ImageProcessor
> 
> How can you _re_compute DL1a? If it is a simtel file, you always need to calculate it. If it is a DL1 file, you cannot recompute it, since the DL0 is not there.

You can compute a new cleaning mask from existing dl1 or not?
Are we talking about semantics or am I missing something fundamental here?

maxnoe:
> You can compute a new cleaning mask from existing dl1 or not?

Redoing the cleaning is not what I think of "recomputing dl1a". 
If this is about redoing the cleaning, then yes, that should be possible. Maybe an explicit "redo_cleaning"?

LukasNickel:
> > You can compute a new cleaning mask from existing dl1 or not?
> 
> Redoing the cleaning is not what I think of "recomputing dl1a".
> If this is about redoing the cleaning, then yes, that should be possible. Maybe an explicit "redo_cleaning"?

It is also about modifying the images though (adding noise / psf smearing / ...)

kosack (dismissed):

nbiederbeck (changes requested):

nbiederbeck (approved):

kosack (approved):

PR #1780, opened 2021-08-30
Do not use sphinx action for building docs

maxnoe:
Trying to fix the docs build by not using the sphinx extension.

We loose the automatic PR comments for sphinx warnings though.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -65,18 +65,18 @@ jobs:
         with:
           fetch-depth: 0
 
-      - uses: ammaraskar/sphinx-action@master
-        with:
-          docs-folder: "docs/"
-          pre-build-command: |
-            apt update --yes && apt install --yes git build-essential pandoc
-            pip install -U pip setuptools wheel
-            pip install -e .[docs]
-            git describe --tags
-            python -c 'import ctapipe; print(ctapipe.__version__)'
+      - name: Install doc dependencies
+        run: |
+          apt update --yes && apt install --yes git build-essential pandoc
+          pip install -U pip setuptools wheel
+          pip install -e .[docs]
+          git describe --tags
+          python -c 'import ctapipe; print(ctapipe.__version__)'
 
+      - name: Build docs
+        run: |
           # treat warnings as errors, make sure links are working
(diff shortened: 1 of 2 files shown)


maxnoe:
Ok, I think we should merge this to unblock the other PRs, we can switch back to the sphinx action if/when it gets updated.

kosack (approved):

LukasNickel (approved):

.github/workflows/ci.yml
```diff
@@ -65,18 +65,21 @@ jobs:
...
+        uses: actions/setup-python@v2
         with:
-          docs-folder: "docs/"
-          pre-build-command: |
-            apt update --yes && apt install --yes git build-essential pandoc
-            pip install -U pip setuptools wheel
-            pip install -e .[docs]
-            git describe --tags
-            python -c 'import ctapipe; print(ctapipe.__version__)'
+          python-version: 3.8
```
HealthyPear, line 71: why is it necessary to specify the python version here?

maxnoe: Why should we rely on whatever is the current default?

HealthyPear: I'm asking because just above you use `${{ matrix.python-version }}` which can be also 3.7

maxnoe: This is for the unit tests, which we run for multiple python versions. It does not really make sense to build the documentation using multiple python versions.

HealthyPear: exactly, so it should be "3.8" instead of "${{ matrix.python-version }}" (not a problem, though)

maxnoe: I don't understand.

HealthyPear: yes, but line 68 refers to the `docs` job, not to the `tests` one, so the name of the job should reflect the version we are using, which is only 3.8, that was my confusion (line 30 instead refers to the tests and is correct)

maxnoe: Oh, there was a copy paste leftover above in the `name`. This is only for displaying, it does not influence the build.

HealthyPear: yes I know, that's why I said it's not a problem, it just triggered a confusion from my side

PR #1790, opened 2021-09-21
Add flag to use telescope frame for ctapipe process

maxnoe:


HealthyPear (approved):

LukasNickel:
This is true per default, no?
https://github.com/cta-observatory/ctapipe/blob/7b15bd9258bfc6c686a84b05a7912fb6b6d4c466/ctapipe/image/image_processor.py#L63-L64
So it should rather be a flag to use the camera frame (or we use that as default).

maxnoe:
Oh, right, will fix

LukasNickel:
This looks different than the other flags now, why is that?

maxnoe:
> This looks different than the other flags now, why is that?

The `flag` function just creates a pair of `--option`, `--no-option` flags. Which makes no sense here, since it rather should be `--telescope-frame` and `--camera-frame`, not `--camera-frame` and `--no-camera-frame`.

I left out `--telescope-frame`, as it is the default. We mostly need the `flags` for stuff that is by default true but where we want to switch it off.

LukasNickel (approved):

PR #1794, opened 2021-09-21
Do not guess pixel area when transforming

maxnoe:
Instead of using the guess pixel area method, we now calculate a scale from the transformed unit vectors.

This is both faster and also works well for cameras with non-uniform pixel sizes or gaps between pixels:

### Before
transform_before

### After
transform_after


LukasNickel (approved):

LukasNickel (changes requested):
Sorry, didnt see the failing test, thought that was codacy.

kosack (approved):

kosack:
I suspect the failed test is some random fluctuation... but better check

LukasNickel (approved):

PR #1756, opened 2021-07-22
ctapipe-fileinfo

kosack:
Just a small fast command-line utility to display information about ctapipe output files. 

- the output is all reference headers in YAML format  (hierarchical by default, but a `--flat` option is available)
- you can optionally write an index table for all files with all headers as columns (uses pandas DataFrame's ability to convert a list of dicts into a table, and then uses astropy.table to write it to any format you want, e.g. `files.fits` or `files.ecsv`)
- files not in hdf5 format are skipped (with a message)
- incomplete hdf5 files will return `{}` (no header)
- you can colorize the output by piping it through `pygmentize -l yaml`

### Help: 
```
% ctapipe-fileinfo --help                                                                                        

usage: ctapipe-fileinfo [-h] [--output-table OUTPUT_TABLE] [--flat] FILENAME [FILENAME ...]

    Display information about ctapipe output files (DL1 or DL2 in HDF5 format).
    Optionally create an index table from all headers


positional arguments:
  FILENAME              filenames of files in ctapipe format

optional arguments:
  -h, --help            show this help message and exit
  --output-table OUTPUT_TABLE
                        generate output in tabular format (default: None)
  --flat                show flat header hierarchy (default: False)
```

### Example 2:

```
%  ctapipe-fileinfo *.DL1.h5

proton_20deg_0deg_run153___cta-prod5-paranal_desert-2147m-Paranal-dark_merged.DL1.h5:
    CTA:
        ACTIVITY:
            ID: eb09130c-1a4e-4a0a-96a0-2607ebca70b1
            NAME: ctapipe-merge
            SOFTWARE:
                NAME: ctapipe
                VERSION: 0.10.5
            START:
                TIME: '2021-03-15 15:30:01.948'
            TYPE: software
        CONTACT:
            EMAIL: ''
            NAME: ''
            ORGANIZATION: CTA Consortium
        INSTRUMENT:
            CLASS: Subarray
            ID: MonteCarloArray
            SITE: Other
            SUBTYPE: unspecified
            TYPE: unknown
            VERSION: unknown
        PROCESS:
            ID: '0'
            SUBTYPE: ''
            TYPE: Simulation
        PRODUCT:
            CREATION:
                TIME: '2021-03-15 15:30:57.161'
            DATA:
                ASSOCIATION: Subarray
                CATEGORY: S
                LEVEL: DL1
                MODEL:
                    NAME: ASWG DL1
                    URL: ''
                    VERSION: v1.1.0
[...]

PR #1796, opened 2021-09-24
add a simple method to read metadata as a dictionnary from an hdf5 file and allow writing metadata to any path

vuillaut:


ctapipe/io/metadata.py
```diff
@@ -226,3 +228,21 @@ def write_to_hdf5(metadata, h5file):
         warnings.simplefilter("ignore", NaturalNameWarning)
         for key, value in metadata.items():
             h5file.root._v_attrs[key] = value  # pylint: disable=protected-access
+
+
+def read_metadata(h5filename):
```
maxnoe, line 233: Could we add the possibility of passing an already opened hdf5 file here and a node other than `root`?
This could then simplify a lot of the other metadata accesing functions in the io code.

For how to pass both a path or an already opened h5file, see:

```
    with ExitStack() as stack:


        if isinstance(h5file, (str, Path)):
            h5file = stack.enter_context(tables.open_file(h5file))
        elif isinstance(h5file, tables.file.File):
            pass
        else:
            raise ValueError(
                f"expected a string, Path, or PyTables "
                f"filehandle for argument 'h5file', got {h5file}"
            )
```

vuillaut: Done.

Thank you for making me aware of `ExitStack` - awesome feature

vuillaut:
> Wasn't this already available as a part of PR #1756?

I wasn't aware of the `fileinfo` - thanks.
However, if I am not mistaken, it is meant to be used as an entrypoint, right?

Here I'd like to  introduce a simple function to read metadata in other (any?) context - it could be used be fileinfo however...

maxnoe (approved):
thx, this is handy

HealthyPear:
> However, if I am not mistaken, it is meant to be used as an entrypoint, right?

True, if you want to use that information as part of something else it is better to have it as a method.
Perhaps we can make this method unique and factorize it in ctapipe-info?

> Here I'd like to introduce a simple function to read metadata in other (any?) context - it could be used be fileinfo however...

From the point of view of the code it doesn't change, since this information is stored at the root of the file, so the structure doesn't matter (ctapipe output or any HDF5 file, e.g. protopipe where - ironically - I am just now adding this :) )

kosack (approved):
sounds nice - I was just today wrestling with the AttributeSet object that pytables gives back (which is not very friendly).

kosack (changes requested):
actually, one thing that also comes up: can you add a "path" option, so that you can also get the metadata from some object inside the HDF5 file?
```py3
def read_metadata(h5file, path="/")
     ...

# example:
table_meta = read_metadata(h5file, "/config/simulation/run")
```

vuillaut:
> actually, one thing that also comes up: can you add a "path" option, so that you can also get the metadata from some object inside the HDF5 file? That way we can also use it in the TableReader, etc.
> 
> ```python
> def read_metadata(h5file, path="/")
>      ...
> 
> # example:
> table_meta = read_metadata(h5file, "/config/simulation/run")
> ```

Node.
While at it, I modified `write_to_hdf5` to allow passing a path as well.

vuillaut:
@kosack @maxnoe 
The failing tests are not related to this PR I believe. I merged master and tried restarting them - something else I can do?
In any case, this PR would need your review again.

HealthyPear:
> The failing tests are not related to this PR I believe. I merged master and tried restarting them - something else I can do?

It's a timeout connection error with the dataserver (cccta-dataserver.in2p3.fr).
I also have the same problem with protopipe (data test files are hosted there, like the ones for ctapipe).
Just re-run the jobs from the action page from time to time and it will work again eventually.

kosack (approved):

maxnoe (approved):

PR #1810, opened 2021-11-23
Extractor status flag

kosack:
changes the return value of ImageExtractors to include a is_valid flag, which is also added to the DL1 data model.

- In most extractors (that have no failure modes), the return value is just hard-coded to True
- in TwoPassWindowSum, where the second pass can fail, it returns False if the second pass failed (meaning the resulting images are from the first-pass only), unless it was requested to only run the first pass.

This is needed, since those events where TwoPass fails need to be filtered.   Protopipe already implements this, but it should move to ctapipe.

### DIFF
--- a/ctapipe/image/tests/test_extractor.py
+++ b/ctapipe/image/tests/test_extractor.py
@@ -20,9 +20,7 @@ from ctapipe.image.extractor import (
     FullWaveformSum,
     SlidingWindowMaxSum,
 )
-from ctapipe.image.toymodel import (SkewedGaussian,
-                                    obtain_time_image,
-                                    WaveformModel)
+from ctapipe.image.toymodel import SkewedGaussian, obtain_time_image, WaveformModel
 from ctapipe.instrument import SubarrayDescription, TelescopeDescription
 
 extractors = non_abstract_children(ImageExtractor)
@@ -64,14 +62,15 @@ def subarray():
 
 @pytest.fixture(scope="module")
 def subarray_1_LST():
-    subarray = SubarrayDescription("One LST",
-                tel_positions={1: np.zeros(3) * u.m},
-                tel_descriptions={1: TelescopeDescription.from_name(
-                    optics_name="LST", camera_name="LSTCam"
-                    ),
-                },
-               )
+    subarray = SubarrayDescription(
+        "One LST",
+        tel_positions={1: np.zeros(3) * u.m},
+        tel_descriptions={
+            1: TelescopeDescription.from_name(optics_name="LST", camera_name="LSTCam")
+        },
+    )
     return subarray
+
 
 def get_test_toymodel(subarray, minCharge=100, maxCharge=1000):
     telid = list(subarray.tel.keys())[0]
@@ -295,9 +294,10 @@ def test_integration_correction_outofbounds(subarray):
 def test_extractors(Extractor, toymodel):
     waveforms, subarray, telid, selected_gain_channel, true_charge, true_time = toymodel
     extractor = Extractor(subarray=subarray)
-    charge, peak_time = extractor(waveforms, telid, selected_gain_channel)
+    charge, peak_time, is_valid = extractor(waveforms, telid, selected_gain_channel)
     assert_allclose(charge, true_charge, rtol=0.1)
     assert_allclose(peak_time, true_time, rtol=0.1)
+    assert is_valid == True
 
 
 @pytest.mark.parametrize("Extractor", extractors)
@@ -308,7 +308,9 @@ def test_integration_correction_off(Extractor, toymodel):
 
     waveforms, subarray, telid, selected_gain_channel, true_charge, true_time = toymodel
     extractor = Extractor(subarray=subarray, apply_integration_correction=False)
-    charge, peak_time = extractor(waveforms, telid, selected_gain_channel)
+    charge, peak_time, is_valid = extractor(waveforms, telid, selected_gain_channel)
+
+    assert is_valid == True
 
     # peak time should stay the same
     assert_allclose(peak_time, true_time, rtol=0.1)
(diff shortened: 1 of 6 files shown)


HealthyPear (dismissed):
I don't know how much easy it would be using the artificial image construction classes, but we could probably add a test where the 2-pass fails

otherwise looks OK to me

HealthyPear (approved):

ctapipe/image/tests/test_extractor.py
```diff
@@ -550,8 +558,9 @@ def test_global_peak_window_sum_with_pixel_fraction(subarray):
         apply_integration_correction=False,
     )
 
-    charge, peak_time = extractor(waveforms, tel_id, selected_gain_channel)
+    charge, peak_time, is_valid = extractor(waveforms, tel_id, selected_gain_channel)
```
maxnoe, line 561: At this point, shouldn't the extractor just return the `CameraImageContainer`?

kosack: yes, perhaps that is simpler now that the return value is so complicated.  In the end that's what we fill in.  However, doesn't LST have it's own subclass of this container? Or am I misremembering?

kosack: It's a bit of a design decision: in general we had lower-level classes/functions return bare information so that they can be used without the need for a Container hierarchy, but here is maybe is simpler

maxnoe: >  However, doesn't LST have it's own subclass of this container? Or am I misremembering?

No, "just" for the `ArrayEventContainer` that has an additional `.lst` field.

maxnoe: https://github.com/cta-observatory/ctapipe_io_lst/blob/fd8ae42d2afadf2cbde4f80fa359a51822b9b321/ctapipe_io_lst/containers.py#L117

maxnoe: > It's a bit of a design decision: in general we had lower-level classes/functions return bare information so that they can be used without the need for a Container hierarchy, but here is maybe is simpler

Here it's not a hierarchy, right? It's just like `hillas_parameters` and the other image parameters a leaf node in the container tree.

kosack: Ok, sounds like it doesn't break LSTs, then, so I could just return the container.

maxnoe (changes requested):
I was mistaken about `DL1ImageContainer`, it's actually `DL1CameraContainer` that has the fields filled here directly and the `parameters` subcontainer.

So maybe we should split this into a structure like this:

```
class DL1CameraContainer(Container):
    pixels = DL1PixelContainer()
    parameters = DL1ParametersContainer()

class DL1PixelContainer(Container):
    image = ...
    peak_time = ...
    cleaning_mask = ...
```

?

That would make the DL1a / DL1b split explicit in the container structure.

kosack:
> I was mistaken about `DL1ImageContainer`, it's actually `DL1CameraContainer` that has the fields filled here directly and the `parameters` subcontainer.
> 
> So maybe we should split this into a structure like this:
> 
> ```
> class DL1CameraContainer(Container):
>     pixels = DL1PixelContainer()
>     parameters = DL1ParametersContainer()
> 
> class DL1PixelContainer(Container):
>     image = ...
>     peak_time = ...
>     cleaning_mask = ...
> ```
> 
> ?
> 
> That would make the DL1a / DL1b split explicit in the container structure.

Yes, I actually thought the same thing when writing this.  Containers should be filled all at once, so having the image mask and parameters in there doesn't make sense. 

Technically the mask should be grouped with the parameters and not the other images (even though it is also an image), so it would be better to have a containers for:
* extracted images (output of an ImageExtractor)
* cleaning_mask + parameters (output of ImageProcessor?)

Anyhow, maybe let's make that a separate issue - this one is getting too large, and has too many changes

kosack:
It would be  better to have something like:
```py3
class DL1CameraContainer:   # (maybe rename DL1ImageContainer)
     image = ...
     peak_time = ...

class DL1ParametersContainer:
    image_mask = ...
    parmeters = ImageParametersContainer
```

kosack:
In any case, right now everything still works, this just changed where the DL1CameraContainer was constructed, so it's no slower or worse than before.  We should open another PR or issue to clean up the containers as above, however

maxnoe (approved):

ctapipe/calib/camera/flatfield.py
```diff
@@ -165,28 +166,27 @@ def __init__(self, **kwargs):
...
-        peak_pos = 0
         if self.extractor:
-            charge, peak_pos = self.extractor(
-                waveforms, self.tel_id, selected_gain_channel
-            )
-
-        return charge, peak_pos
+            return self.extractor(waveforms, self.tel_id, selected_gain_channel)
+        else:
+            return DL1CameraContainer(image=0, peak_pos=0, is_valid=False)
```
maxnoe, line 189: Shouldn't this be `None` or `np.zeros(n_pixels)`? Filling a single 0 seems wrong

ctapipe/calib/camera/pedestals.py
```diff
@@ -195,29 +196,27 @@ def __init__(self, **kwargs):
...
-        peak_pos = 0
         if self.extractor:
-            charge, peak_pos = self.extractor(
-                waveforms, self.tel_id, selected_gain_channel
-            )
-
-        return charge, peak_pos
+            return self.extractor(waveforms, self.tel_id, selected_gain_channel)
+        else:
+            return DL1CameraContainer(image=0, peak_pos=0, is_valid=False)
```
maxnoe, line 219: Same as above

ctapipe/calib/camera/flatfield.py
kosack: this comes from Franca's code, so I didn't want to change anything. I guess these values are never relied upon (later the is_valid flag is checked, so these are basically null events),  However, it's true the DataWriter would choke on them, so maybe better to do it right

kosack: or maybe `image=None, peak_pos=None`

LukasNickel (approved):

Issue #1809, opened 2021-11-23
test_toy/test_intensity[None] fails on macOS


kosack:
The following line seems to occasionally  lead to a random test failure:

https://github.com/cta-observatory/ctapipe/blob/22655d189d6fe1330e01891ff12296b621313bdf/ctapipe/image/tests/test_toy.py#L44-L45

```
E       assert 177.0 <= 168
E        +  where 177.0 = <bound method rv_frozen.ppf of <scipy.stats._distn_infrastructure.rv_frozen object at 0x7fdb48888490>>(0.05)
E        +    where <bound method rv_frozen.ppf of <scipy.stats._distn_infrastructure.rv_frozen object at 0x7fdb48888490>> = <scipy.stats._distn_infrastructure.rv_frozen object at 0x7fdb48888490>.ppf
E        +      where <scipy.stats._distn_infrastructure.rv_frozen object at 0x7fdb48888490> = poisson(200)
E        +  and   168 = <built-in method sum of numpy.ndarray object at 0x7fdb203970f0>()
E        +    where <built-in method sum of numpy.ndarray object at 0x7fdb203970f0> = array([0, 0, 0, ..., 0, 0, 0]).sum

ctapipe/image/tests/test_toy.py:45: AssertionError
```

kosack:
actually, it's not intermittent: seems to always fail on my machine, so probably related to a random seed.

kosack:
Looking deeper `pytest ctapipe/image -k toy`, this test passes (at least on a few trials). 

```
pytest ctapipe/image -k toy

platform darwin -- Python 3.8.8, pytest-6.2.2, py-1.10.0, pluggy-0.13.1 -- /Users/kkosack/miniconda3/envs/cta-0.10.5/bin/python
cachedir: .pytest_cache
rootdir: /Users/kkosack/Projects/CTA/Working/ctapipe, configfile: setup.cfg
plugins: cov-2.11.1, xdist-2.2.1, forked-1.3.0
collected 102 items / 95 deselected / 7 selected

ctapipe/image/tests/test_hillas.py::test_with_toy PASSED                                                    [ 14%]
ctapipe/image/tests/test_toy.py::test_intensity[None] PASSED                                                [ 28%]
ctapipe/image/tests/test_toy.py::test_intensity[0] PASSED                                                   [ 42%]
ctapipe/image/tests/test_toy.py::test_skewed PASSED                                                         [ 57%]
[...]

PR #1811, opened 2021-11-26
Make sure a fixed random seed is used for toymodel test

maxnoe:


kosack (approved):
I checked that this works on my macOS machine (unlike the master branch), and it does. Therefore I think this fixes #1809

LukasNickel (approved):

PR #1814, opened 2021-12-07
Read true_image_sum, fixes #1180

### LINKED ISSUE #1180: add mc true image sum
In MCs without the `mc_photo_electron_image` (the true number of PEs on the photosensor plane without NSB noise), it seems there may be at least the sum of this image stored in the MC data:

From @moralejo: 
```
mc_pe_sum = mc_event.mc_pesum.num_pe[itel]
```

We should read this in the SimTelEventSource, and put it in the `MCCameraEventContainer`

maxnoe:


kosack:
What is the difference here between the true_image and true_image_sum? Should it really be a separate attribute ?

ctapipe/containers.py
```diff
@@ -484,6 +484,9 @@ class SimulatedCameraContainer(Container):
 
     container_prefix = ""
 
+    true_image_sum = Field(
+        np.nan, "Sum of the detected Cherenkov photons in the camera"
```
kosack, line 488: maybe add a type here or a note in the description to remind us that it is a scalar not a sum in time?

maxnoe: "Sum of photons over the camera" is pretty clear on that it is a scalar, right?

maxnoe:
@kosack See the connected issue #1180, it is essentially the same as `true_image.sum()`, but simtel can be configured to only store the sum, not the individual pixel values.

ctapipe/containers.py
kosack: I read it as the sum in time, which is why I was confused at first (e.g. the sum for each pixel). Here it's the temporal and spatial sum.

kosack: Maybe we should call it true_intensity? Since intensity is what we call the image sum later.

kosack (approved):

LukasNickel (approved):

PR #1816, opened 2021-12-10
Implement treatment of invalid values in FixedPointColumnTransform, fixes #1815

### LINKED ISSUE #1815: FixedPointColumnTransform does not handle inf, nan
nan and inf are stored MINVAL on writing and not converted back on reading:
```
In [3]: np.array([-np.inf, np.nan, np.inf]).astype(np.int32)
Out[3]: array([-2147483648, -2147483648, -2147483648], dtype=int32)
```

maxnoe:


kosack (dismissed):

kosack (approved):

ctapipe/io/datawriter.py
```diff
@@ -578,6 +578,10 @@ def _write_dl1_telescope_events(
                     )
 
             if self.write_images:
+                if dl1_camera.image is None:
+                    raise ValueError(
+                        "DataWrite.write_images is True but event does not contain image"
```
kosack, line 583: typo

kosack (dismissed):

ctapipe/io/tableio.py
```diff
@@ -350,26 +350,106 @@ class FixedPointColumnTransform(ColumnTransform):
...
+        iinfo = np.iinfo(self.target_dtype)
+
+        # use three highest values for nan markers for unsigned case
+        if self.unsigned:
+            self.neginf = iinfo.max - 2
+            self.nan = iinfo.max - 1
+            self.posinf = iinfo.max
+
+            # this leaves this inclusive range for the valid values
+            self.minval = 0
```
LukasNickel, line 389: Thats obviously the same, but I would personally use `iinfo.min` for consistency or drop a note

ctapipe/io/tableio.py
```diff
@@ -350,26 +350,106 @@ class FixedPointColumnTransform(ColumnTransform):
...
+            self.nan = iinfo.min + 1
+            self.posinf = iinfo.max
+
+            # this leaves this inclusive range for the valid values
+            self.minval = iinfo.min + 2
+            self.maxval = iinfo.max - 1
 
     def __call__(self, value):
-        return (value * self.scale).astype(self.target_dtype) + self.offset
+        is_scalar = np.array(value, copy=False).shape == ()
```
LukasNickel, line 401: Numpy docs suggests `np.ndim(value) == 0`
https://numpy.org/doc/stable/reference/generated/numpy.isscalar.html

maxnoe: This is what is done in astropy. There probably was a problem with `isscalar` and they didn't check maybe?

LukasNickel: `np.isscalar` wont work on 0-dim arrays, `np.ndim` (and this implementation) does.

> If you need a stricter way to identify a numerical scalar, use isinstance(x, numbers.Number), as that returns False for most non-numerical elements such as strings.
In most cases np.ndim(x) == 0 should be used instead of this function, as that will also return true for 0d arrays. This is how numpy overloads functions in the style of the dx arguments to gradient and the bins argument to histogram.

This is probably fine as well. I was just wondering where it comes from

ctapipe/io/tableio.py
maxnoe: Really? I find 0 much clearer than people needing to parse `iinfo.min` and think "ah right, this is unsigned, so this must be 0`....

LukasNickel: For me it was an overhead with all the `iinfo.min/max` above and below.
Its probably pure preference

ctapipe/io/tableio.py
LukasNickel: I fail to see a reason why this would not work so I assume it covers all the relevant cases

LukasNickel (approved):

LukasNickel:
Ah, that started new conversations
Sorry

maxnoe:
@kosack @LukasNickel Rebased to fix the CI, please re-approve

FrancaCassol (approved):

LukasNickel (approved):

Issue #1506, opened 2020-11-05
Remove some tools


maxnoe:
I think we have a lot of tools that are not actually needed and also don't really follow current best practices.

Easiest would be just to remove them:

- [ ] display_events_single_tel
- [ ] display_integrator
- [ ] display_summed_images
- [ ] camdemo

kosack:
We can probably remove some of those, even if some are useful for debugging.   I use display_summed_images and display_events_single_tel frequently for making animations for presentations,  but they could be placed in my own private repo so no problem removing them unless others use them.  I'd keep camdemo, as it was meant to be a general test of how the config, Tools,  and graphics works, so that is harmless and nice to have to demonstrate that things are working (it's the first think I run on a new install or to show people what data looks like).

maxnoe:
Aren't these usecases also supported by `display-dl1` and the `bokeh` viewers?

kosack:
> Aren't these usecases also supported by display-dl1 and the bokeh viewers?

not completely, but would be pretty easy to merge all into one tool

maxnoe:
I am in the process of rewriting the bokeh viewer. Can you make a list of features we want? 
Then I will remove the old tools when their features are included in the bokeh viewer.

I also want to make the bokeh viewer a component that can be optionally enabled in e.g. the stage1 tool to debug things.

moralejo:
> I am in the process of rewriting the bokeh viewer. Can you make a list of features we want?
> Then I will remove the old tools when their features are included in the bokeh viewer.
> 
> I also want to make the bokeh viewer a component that can be optionally enabled in e.g. the stage1 tool to debug things.

Hi, a couple from my side, for the camera display, that we implemented in [https://github.com/cta-observatory/cta-lstchain/blob/master/lstchain/visualization/bokeh.py](url):

- lin/log version of a display, using tabs
- Slider to modify lower and upper end of the z-scale
- accompanying histogram of all camera values and (perhaps) plot of camera values vs. pixel id
- Pixel info shown (id, content, whatever - configurable) when mouse tip hovers over it
- Optional slider to show different versions of the  displayed data. We use it for example to show different runs (when displaying e.g. mean pixel charge in a run)

maxnoe:
done

PR #1827, opened 2022-01-25
Remove outdated tools

maxnoe:
Remove the four tools discussed for removal here #1506


kosack (approved):
while I still use some of these for debugging and outreach, I think it's ok to drop them since now with the DL1/DL2 data model + `ctapipe.io.TableLoader` + Jupyter notebooks, it's easy to replicate all of this functionality, and often simpler.

kosack:
Though I still like camdemo, as it's fun and a nice simple example of writing a Tool and using the InstrumentDescription model!

maxnoe:
> Though I still like camdemo, as it's fun and a nice simple example of writing a Tool and using the InstrumentDescription model!

Then should I move it to examples? Or the Tool documentation?

I am against having something that's just supposed to be a documentation example installed in `PATH` along with the "real" tools.

HealthyPear (approved):

PR #1835, opened 2022-01-28
Raise error for missing focal length

maxnoe:


LukasNickel:
Is 0 a default we expect somewhere?
I would have assumed to check `if not focal_length: ...`

HealthyPear:
> Is 0 a default we expect somewhere? I would have assumed to check `if not focal_length: ...`

yes here
https://github.com/cta-observatory/ctapipe/blob/3e35093f2d2ade576c79b4b326bbab7b5bde298f/ctapipe/coordinates/camera_frame.py#L78

maxnoe:
Astropy would complain about `None` and checking `np.isnan()` is much slower than checking `== 0`, so I left it in.

HealthyPear (changes requested):
also the reverse transformation requires to know the focal length

https://github.com/cta-observatory/ctapipe/blob/3e35093f2d2ade576c79b4b326bbab7b5bde298f/ctapipe/coordinates/camera_frame.py#L176

ctapipe/coordinates/camera_frame.py
```diff
@@ -174,6 +178,8 @@ def telescope_to_camera(telescope_coord, camera_frame):
         y_rotated = x_pos * sinrot + y_pos * cosrot
 
     focal_length = camera_frame.focal_length
+    if focal_length.value == 0:
```
maxnoe, line 181: @HealthyPear here it is

HealthyPear: oh right sorry I overlooked

HealthyPear (approved):

LukasNickel (approved):

PR #1837, opened 2022-02-03
Rename DL1EventSource to HDF5EventSource, add support for reading R1 waveforms as written by DataWriter

maxnoe:
* This makes us able to read R1 data back from files written using the `DataWriter.write_waveforms`
* It also enables *only* writing the R1 waveforms

This was motivated for LST to write out calibrated DL0 calibration events so @FrancaCassol can start developing the calibration pipeline in ctapipe without having to redo the low-level calibrations over and over again.

### DIFF
--- a/ctapipe/io/__init__.py
+++ b/ctapipe/io/__init__.py
@@ -11,7 +11,7 @@ from ..core.plugins import detect_and_import_io_plugins
 
 # import event sources to make them visible to EventSource.from_url
 from .simteleventsource import SimTelEventSource
-from .hdf5eventsource import HDF5EventSource
+from .hdf5eventsource import HDF5EventSource, get_hdf5_datalevels
 
 # import IO plugins with their event sources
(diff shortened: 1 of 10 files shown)


kosack (approved):
Nice refactoring!   Is there a plan to also include DL2 in HDF5EventSource for completeness? (it's not strictly necessary, as there is not a strong use case yet)

maxnoe:
Hi, no directly, but should probably be easy.

This was motivated for LST to write out calibrated DL0 calibration events so @FrancaCassol can start developing the calibration pipeline in ctapipe without having to redo the low-level calibrations over and over again.

kosack:
> @FrancaCassol can start developing the calibration pipeline in ctapipe without having to redo the low-level calibrations over and over again.

That's good use case to know. Maybe add that to the description so it's clear.  Then I'm happy to merge this now, and add any other features later.

FrancaCassol (approved):

PR #1839, opened 2022-02-09
Fix _dev_version being included in wheel

maxnoe:
Building wheels does not honor `MANIFEST.in`, that is only used for source dists.

To exclude the `_dev_version.py` from wheels, it has to be a package excluded in `find_package`

### DIFF
--- a/MANIFEST.in
+++ b/MANIFEST.in
@@ -8,6 +8,6 @@ recursive-include docs *
 prune build
 prune docs/_build
 prune docs/api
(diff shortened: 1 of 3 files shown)


kosack (dismissed):

maxnoe:
@kosack sorry, forgot to change the path of the repo root after moving the file.

kosack (approved):

LukasNickel (approved):

PR #1840, opened 2022-02-11
Add RTD config

maxnoe:


maxnoe:
The build works with the configuration here, unfortunately that means we can not easily get the versions up for the last releases.

We could make branches for them and cherry-pick the config to get them to build.

kosack (approved):

LukasNickel (approved):

PR #1841, opened 2022-02-17
Add event type filter to ctapipe-process

maxnoe:
Allows filtering out un-wanted event types form the output.

### DIFF
--- a/ctapipe/tools/process.py
+++ b/ctapipe/tools/process.py
@@ -13,6 +13,8 @@ from ..image.extractor import ImageExtractor
 from ..io import DataLevel, DataWriter, EventSource, SimTelEventSource
 from ..io.datawriter import DATA_MODEL_VERSION
 from ..reco import ShowerProcessor
+from ..utils import EventTypeFilter
+
 
 COMPATIBLE_DATALEVELS = [
     DataLevel.R1,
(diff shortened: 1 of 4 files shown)


kosack (dismissed):

ctapipe/utils/event_type_filter.py
```diff
@@ -0,0 +1,32 @@
...
+class EventTypeFilter(Component):
+    """Check that an event has one of the allowed types"""
+
+    allowed_types = Set(
+        # add both the enum instance and the integer values to support
+        # giving the integers in config files.
+        trait=Enum(list(EventType) + [t.value for t in EventType]),
+        default_value=None,
+        allow_none=True,
+        help="The allowed types. Set to None to allowe all types.",
```
LukasNickel, line 18:
```suggestion
        help="The allowed types. Set to None to allow all types.",
```

LukasNickel (dismissed):

ctapipe/tools/process.py
```diff
@@ -238,6 +245,7 @@ def start(self):
         self.log.info("(re)compute DL2: %s", self.should_compute_dl2)
         self.event_source.subarray.info(printer=self.log.info)
 
+        event = None
```
nbiederbeck, line 248: Why this line? Everything else inside this function is inside the for-loop, where `event` gets overwritten

maxnoe: It is for the case that the source has no triggered events that `event` is not undefined.

maxnoe: Ah, sorry, it's actually not needed in this PR, comes from other code I was also adding

nbiederbeck: Ok then, can stay here I think

nbiederbeck (dismissed):

nbiederbeck (approved):

kosack (approved):

PR #1854, opened 2022-03-21
Implement public interface for 2d index to/from id

maxnoe:
Add public methods to go from pixel id to index in 2d representation and back. Turns out, due to the flipping, the mapping does not hold. (The new test is failing)


LukasNickel (dismissed):
I am still not 100% why the columns are reversed in the first place, but this way we have a consistent interface

kosack (approved):

PR #1856, opened 2022-03-22
allow YAML and TOML config files and multiple configs

### LINKED ISSUE #1047: add support for YAML config files in core.Tool
Right now `Tools` (and `Components`) are configured via `traitlets.config`, which is nice , but so far only supports JSON config files out of the box (see e.g. the code in lst-chain that uses this already).  It would be nice, and probably quite easy, to add support for YAML config files.  

Internally, traitlets.config just turns the config file into a dict(), and presumably this is done with a call to `json.loads()` somewhere.  It should be straightforward to make it also use `yaml.loads()` if the config file is YAML-based. Probably we can implement this in our Tool class directly, rather than asking for traitlets to support it. Tool is already a sub-class of `traitlets.config.Application`. 

## example:
Current config file (for PedestalCalculator):
```json
{
   "version": 1,

   "PedestalHDF5Writer": {
     "calculator_product": "PedestalIntegrator",
     "output_file":"pedestal.hdf5",
     "log_level":"DEBUG"
   },         
   "EventSource": {
      "input_url": "/ctadata/franca/LST/LST-1.1.Run00168.0000.fits.fz",
      "max_events": 100
   },
   "PedestalCalculator":{
     "sample_size": 90,
     "sample_duration":1000,
     "tel_id":0,
      "charge_product":"LocalPeakIntegrator"
   },
    "LocalPeakIntegrator":{
     "window_shift": 4,
     "window_width":10
    }
}
```

could become this (which is more human-friendly):
```yaml
version: 1
EventSource:
  input_url: /ctadata/franca/LST/LST-1.1.Run00168.0000.fits.fz
  max_events: 100
LocalPeakIntegrator:
  window_shift: 4
  window_width: 10
PedestalCalculator:
  charge_product: LocalPeakIntegrator
  sample_duration: 1000
  sample_size: 90
  tel_id: 0
PedestalHDF5Writer:
  calculator_product: PedestalIntegrator
  log_level: DEBUG
  output_file: pedestal.hdf5
```

### LINKED ISSUE #1732: Ability to use multiple config files
Right now the config files have a lot of repetition, particularly when different output options are used with the same cuts configuration  as in #1726 (the cuts configuration must be copied into each config file, which is error prone).  Having the ability to merge several config files would simplify this. Options are: 

1. using a config system that allows pre-processor "includes" .  This is already possible if we  switch from JSON to Python-based config files for example.  Python configs are already supported already by `traitlets.config`, but are slightly ugly in that every parameter needs to have it's full hierarchy specified, like c.ShowerProcessor.ShowerQualityQuery=...).  Neither YAML nor JSON allow includes natively. 
2. have a way to specify multiple `--config` options to a `Tool`, so that one can do the following to have multiple includes merged (note: order should matter and later configs should overwrite earlier ones)
```
ctapipe-process --config stage1_config.json --config cuts_config.json --input gammas.simtel.gz
```

kosack:
fixes #1047 
fixes #1732

This adds a function to load a YAML or TOML config file just like a JSON or Python one.  There is more to do to make this fully useful, but it works as is.

It also adds the ability to specify multiple config files for a tool (by using repeated `--config` options, or  a list in a config file itself). 

For example the config file can look like:
```yaml
DataWriter:
  Contact:
    # please fill in your contact information here. It will be stored in the
    # output files as provenance information
    name: YOUR-NAME-HERE
    email: YOUREMAIL@EXAMPLE.ORG
    organization: YOUR-ORGANIZATION

  # options that control what is stored in the output file
  overwrite: false # do not overwrite existing files
  write_images: true # store DL1 images
  write_parameters: true # store DL1 parameters
  write_stereo_shower: false # store DL2 stereo geometry

CameraCalibrator:
  image_extractor_type: NeighborPeakWindowSum

ImageProcessor:
  image_cleaner_type: TailcutsImageCleaner

  # make sure you include a configuration for the image cleaner you selected
  # above here. The section named by the image_cleaner_type will be used to
  # configure it.
  TailcutsImageCleaner:
    # the thresholds for this image cleaner must be optimized for the data set
    # you are analyzing. The defaults may not be correct.
    picture_threshold_pe:
      - [type, "*", 10.0]
      - [type, LST_LST_LSTCam, 5.0]
      - [type, MST_MST_NectarCam, 5.0]
      - [type, SST_ASTRI_CHEC, 3.0]
    boundary_threshold_pe:
      - [type, "*", 5.0]
      - [type, LST_LST_LSTCam, 2.5]
      - [type, MST_MST_NectarCam, 2.5]
      - [type, SST_ASTRI_CHEC, 1.5]
    min_picture_neighbors:
      - [type, "*", 2]

  # These specify which images should be parameterized:
  ImageQualityQuery:
    quality_criteria:
      - [enough_pixels, "lambda im: np.count_nonzero(im) > 2"]
      - [enough_charge, "lambda im: im.sum() > 50"]

ShowerProcessor:
  # These specify criteria for telescopes that should be included in stereo
  # reconstruction:
  ShowerQualityQuery:
    quality_criteria:
      - [enough intensity, "lambda p: p.hillas.intensity > 50"]
      - [Positive width, "lambda p: p.hillas.width.value > 0"]
      - [enough pixels, "lambda p: p.morphology.num_pixels > 3"]

```


maxnoe:
I am not such a huge fan of yml anymore... way too complex. I'd go for toml to have a simple config file format does not have the short comings of json / the unlimited power of py files

ctapipe/core/tool.py
```diff
@@ -220,7 +222,7 @@ def _load_tool_config_file(self, path: pathlib.Path):
             self.update_config(config)
         else:
             # fall back to traitlets.config.Application's implementation
-            self.load_config_file(str(path))
+            super(self).load_config_file(str(path))
```
maxnoe, line 225: super doesn't need arguments in python 3

kosack:
Now also supports TOML:
```toml
[DataWriter]
overwrite = false
write_images = true
write_parameters = true
write_stereo_shower = false
write_mono_shower = false
transform_image = true
transform_peak_time = true

[CameraCalibrator]
image_extractor_type = "NeighborPeakWindowSum"

[ImageProcessor]
image_cleaner_type = "TailcutsImageCleaner"

[DataWriter.Contact]
name = "YOUR-NAME-HERE"
email = "YOUREMAIL@EXAMPLE.ORG"
organization = "YOUR-ORGANIZATION"

[ImageProcessor.TailcutsImageCleaner]
picture_threshold_pe = [ [ "type", "*", 10.0,], [ "type", "LST_LST_LSTCam", 5.0,], [ "type", "MST_MST_NectarCam", 5.0,], [ "type", "SST_ASTRI_CHEC", 3.0,],]
boundary_threshold_pe = [ [ "type", "*", 5.0,], [ "type", "LST_LST_LSTCam", 2.5,], [ "type", "MST_MST_NectarCam", 2.5,], [ "type", "SST_ASTRI_CHEC", 1.5,],]
min_picture_neighbors = [ [ "type", "*", 2,],]

[ImageProcessor.ImageQualityQuery]
quality_criteria = [ [ "enough_pixels", "lambda im: np.count_nonzero(im) > 2",], [ "enough_charge", "lambda im: im.sum() > 50",],]

[ShowerProcessor.ShowerQualityQuery]
quality_criteria = [ [ "enough intensity", "lambda p: p.hillas.intensity > 50",], [ "Positive width", "lambda p: p.hillas.width.value > 0",], [ "enough pixels", "lambda p: p.morphology.num_pixels > 3",],]
```

ctapipe/core/tool.py
```diff
@@ -16,6 +16,13 @@
     HAS_YAML = False
     pass  # no support for YAML
 
+try:
+    import toml
```
maxnoe, line 20: Point in case: I think nowadays we want to use `tomli` which is also the library chosen for inclusion in the standard lib for python 3.11: https://peps.python.org/pep-0680/

kosack: `toml` is already installed as part of ctapipe (msut be a dependency of some other package), but sure I can add this one  if it's nicer. I guess the interface is the same though, so can always wait until python 3.11

kosack:
Probably there is a much nicer way to format the TOML version... I need to look at the spec.  That was an auto-generated conversion from the JSON

ctapipe/core/tool.py
```diff
@@ -1,21 +1,41 @@
...
+try:
+    import yaml
+
+    HAS_YAML = True
+except ImportError:
+    HAS_YAML = False
+    pass  # no support for YAML
+
+try:
+    import tomli
```
maxnoe, line 20: `tomli as toml` or `from tomli import load as load_toml`? Otherwise the code below won't run

kosack:
Any idea how to write nested lists  on multiple lines correctly in TOML? I tried what I thought the manual said, and it doesn't work: This Fails to read:
```toml
[ImageProcessor.TailcutsImageCleaner]

picture_threshold_pe = [
    [ "type", "*", 10.0,],
    [ "type", "LST_LST_LSTCam", 5.0,],
    [ "type", "MST_MST_NectarCam", 5.0,],
    [ "type", "SST_ASTRI_CHEC", 3.0,],
]
```

Ah... TOML seems to be super picky about spaces after sections.  Even Github won't render it

kosack:
So TOML so far doesn't work with the `TelescopeParameters`... 

1. It seems to require that all elements of an array are of the same format, so the nested arrays all have to be strings (then it stops complaining about not being able to parse the file
2. TelescopeParameter then of course fails to validate the value since it expects a float not a string.

Point 1  seems overly picky.

maxnoe:
TOML works fine, the `toml` library does not fully support toml 1.0. Using `tomli` works fine:

```
In [1]: import tomli

In [2]: !cat test.toml
[cuts]
foo = [
	["type", "*", 10.0],
	["type", "LST", 4.0],
	["id", 1, 4.0],
]

In [3]: tomli.load(open('./test.toml', "rb"))
Out[3]: {'cuts': {'foo': [['type', '*', 10.0], ['type', 'LST', 4.0], ['id', 1, 4.0]]}}
```

So I'd just try `tomli` and not `toml`

kosack:
Yes tomli seems to work.

ctapipe/core/tool.py
```diff
@@ -119,16 +134,18 @@ def main():
...
-            "command-line parameters"
-        ),
+    config_file = List(
+        trait=Path(
+            exists=True,
+            directory_ok=False,
+            allow_none=True,
+            default_value=None,
+            help=(
+                "name of a configuration file with "
```
maxnoe, line 144: Expand `help` to explain we can have multiple config files and how precedence works?

ctapipe/core/tool.py
```diff
@@ -189,18 +206,38 @@ def initialize(self, argv=None):
         self.update_logging_config()
 
         if self.config_file is not None:
-            self.log.debug(f"Loading config from '{self.config_file}'")
+            self.log.info(f"Loading config from '{self.config_file}'")
```
maxnoe, line 209: Move into loop

maxnoe: `Provenance().add_input(config_file, "config")`?

maxnoe (changes requested):
Do we want to change the format of what `ctapipe-quickstart` puts out to be yaml or toml?

Do we want to add a requirement for one of the more "sane" formats?

ctapipe/core/tool.py
kosack: good point! I'm actually surprised it wasn't there before

kosack:
> Do we want to change the format of what ctapipe-quickstart puts out to be yaml or toml?

Right now it writes all three... But I think I'll clean up the standard config  in a future PR (this one is already too long: it's just suppose to add the ability).  That can be a separate PR, as those files need to be updated and documented better anyhow. WIth this PR, we now have the ability to add comments/documentation.

> Do we want to add a requirement for one of the more "sane" formats?

Can always leave it up to the user, but recommend one of them

maxnoe:
> Right now it writes all three... 

That will confuse people I think. At least we should add a switch `--format=` with some default so that only one variant is created.

ctapipe/core/tool.py
```diff
@@ -189,18 +209,40 @@ def initialize(self, argv=None):
...
 
+    def load_config_file(self, path: Union[str, pathlib.Path]):
+
+        path = pathlib.Path(path)
+
+        if path.suffix in [".yaml", ".yml"] and HAS_YAML:
+            # do our own YAML loading
+            with open(path, "r") as infile:
+                config = Config(yaml.safe_load(infile))
+            self.update_config(config)
```
maxnoe, line 235: I think this overrides previous configs fully. Looking at the traitlets code for `load_config_file`, they do something like this instead:

```
config  = Config()
for config_file in self.config_files:
     config.merge(load(config_file))
self.update_config(config)
```

maxnoe: We should test with two configs setting independent things.

kosack: The main reason they merge a bunch together first was to be able to print an error when one fails, I think (at least that is what I understood from the code), but it would be good to check that this doesn't break anything.  The idea is that  a later config file *should* overwrite earlier settings, but it should not drop old configs if that is what you mean, i.e. that `update_config()` *replaces* rather than *updates* previous ones.

kosack: I'll add a test in any case

maxnoe: I feared that loading two config files with this:

```
{"a": 1}
```

```
{"b": 2}
```

results in just `{"b": 2}` if we do `update` instead of `merge`, but I didn't check

kosack: This is the implementation of `update_config()` in Traitlets:
```py3
   def update_config(self, config):
        """Update config and load the new values"""
        # traitlets prior to 4.2 created a copy of self.config in order to trigger change events.
        # Some projects (IPython < 5) relied upon one side effect of this,
        # that self.config prior to update_config was not modified in-place.
        # For backward-compatibility, we must ensure that self.config
        # is a new object and not modified in-place,
        # but config consumers should not rely on this behavior.
        self.config = deepcopy(self.config)
        # load config
        self._load_config(config)
        # merge it into self.config
        self.config.merge(config)
```
So I think it is ok: "update" here means "load and merge"

ctapipe/tools/quickstart.py
```diff
@@ -29,19 +28,22 @@
 This working directory contains some example configuration files that are useful
 for processing data with `ctapipe-process`. These include:
 
-- stage1_config.json:  generate DL1 data from lower data levels
-- stage2_config.json:  generate DL2 shower geometry from DL1 or lower levels
-- training_config.json: generate both DL1 parameter and DL2 shower geometry data
+- base_config.yaml: stanard configuration options, to be included always
```
maxnoe, line 31: typo: stanard

kosack: fixed

kosack:
This reminds me one of the future PRs has to be to make these example files be generated automatically, including help text from the classes and trait options (similar to what traitlets.config does when it writes out sample python config files).  For that, I will need to also add a way to specify which sections/classes to include in the example.   Right now there are both the internal class and trait descriptions, and comments in the sample config files. That is unnecessary repetition, and things need to be updated in both places.  I have to check if YAML or TOML lets you also include comments in the serialized output, or if it has to be done manually... I think not currently (except for some other implementations of yaml).

kosack:
Will need to update the config files when #1875 is merged

ctapipe/tools/tests/resources/base_config.yaml
```diff
@@ -18,10 +18,13 @@ DataWriter:
     organization: YOUR-ORGANIZATION
 
   # options that control what is stored in the output file
+  # by default here we write nothing (overide these in
```
maxnoe: this looks like the sentence got cut off

maxnoe (approved):
LGTM

LukasNickel (approved):

PR #1857, opened 2022-03-23
Time next neighbor cleaning

LucaMF:


ctapipe/image/time_next_neighbor_cleaning.py
```diff
@@ -0,0 +1,742 @@
...
+    cuts in the Q-dT parameterspace then are determined from the individual
+    pixel rate which gives the rate in which the pixels have a charge above
+    a given threshold. Those graphs are calculated from calibration data
+    (i.e. images without Cherenkov signal but pure night sky background).
+    """
+
+    def __init__(self):
+        self.IPR_dict = None
+        self._multiplicity_dict = {"4nn": 4, "3nn": 3, "2+1": 3, "2nn": 2}
+        self.__geometry = None
```
kosack, line 62: generally you should not use double-underscores unless you really want name-mangling to occur (double underscores affect how  python interprets a member). A single-underscore is fine for showing that this is a protected member.  You could simply make the `geometry` property public to avoid the name collision. .

ctapipe/image/time_next_neighbor_cleaning.py
```diff
@@ -0,0 +1,742 @@
...
+        Load the IPR graphs from a json file. The dictionaries should contain
+        one dictionary for each camera_id with the pixel rates and charges.
+
+        Parameters
+        ----------
+        file: string
+            json file in which the IPR graphs is stored
+        """
+        with open(file, "r") as f:
+            self.IPR_dict = json.load(f)
```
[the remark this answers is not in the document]
kosack: Also is there a well-defined format for this file? Could you use a table format instead? The ctapipe developer guidelines specify always using a format readable by `astropy.table.Table` (FITS or ECSV in particular) for any input or output data when possible.  This lets us also keep track of units for each column, attach observatory metadata, and to write the data later to output files.

ctapipe/image/time_next_neighbor_cleaning.py
```diff
@@ -0,0 +1,742 @@
...
+        self.IPR_dict = self.IPR_dict if self.IPR_dict else {}
+
+        if geo.camera_name not in self.IPR_dict:
+            self.IPR_dict[geo.camera_name] = {
+                "counter": 0,
+                "charge": charge_steps,
+                "npix": np.zeros(charge_steps.shape[0]),
+                "rate": np.zeros(charge_steps.shape[0]),
+                "rate_err": np.zeros(charge_steps.shape[0]),
+            }
```
kosack, line 149: See previous comment. I think using a Table format for this, rather than a  dict would be better.

ctapipe/image/time_next_neighbor_cleaning.py
```diff
@@ -0,0 +1,742 @@
...
+                FlashCam={"4nn": 71559, "3nn": 18319, "2nn": 5124, "2+1": 68112},
+                NectarCam={"4nn": 75302, "3nn": 19269, "2nn": 5394, "2+1": 71712},
+                CHEC={"4nn": 36289, "3nn": 11720, "2nn": 4000, "2+1": 61432},
+            )
+            comb_factor = lookup[camera_name][neighbor_group]
+
+        else:
+            from ctapipe.instrument import CameraGeometry
+
+            geometry = CameraGeometry.from_name(camera_name)
```
kosack, line 265: CameraGeometry.from_name is only used for testing purposes, and does not guarantee you get the correct camera geometry for the file you are reading.  Rather you should let the user pass in a SubarrayDescription to this class (As we do with other classes that need instrument information). Then you can just access all the cameras from that object. 

See how we do it elsewhere. There are generally loops like:
```py3
with EventSource(eventfile) as source:
     cleaner = TimeNextNeighborCleaning(subarray=source.subarray)
     ...
     for event in source:
          cleaner.clean(event)
```

ctapipe/image/time_next_neighbor_cleaning.py
```diff
@@ -0,0 +1,742 @@
...
+    n_neighbors = np.sum(mask * neighbor_matrix, axis=1) - 1
+
+    valid_pixels = n_neighbors >= min_neighbors
+    if keep_neighbors:
+        valid_pixels = dilate_matrix(valid_pixels, neighbor_matrix)
+
+    return valid_pixels & mask
+
+
+class TimeNextNeighborCleaning:
```
kosack, line 46: Should this be a subclass of ImageCleaner? I guess so

kosack: Or at least there should be a wrapper that is a subclass of ImageCleaner that calls this class internally. That would allow you to not only specify options from a config file, but also lets the user choose the image cleaner on the fly. See how other ImageCleaners work.

ctapipe/image/time_next_neighbor_cleaning.py
```diff
@@ -0,0 +1,742 @@
...
+        # TODO: Check the updates on the calculation of neighbors in ctapipe.
+        kdtree, points = self.get_kdtree(mask)
+        dist = self._geometry.guess_pixel_width(
+            self._geometry.pix_x, self._geometry.pix_y
+        )
+
+        # kdtree implementation to get the possible combinations of pairs
+        # within a given distance.
+        combs = kdtree.query_pairs(r=d1 * dist.value)
+        if nfold == "2nn":
```
kosack, line 551: probably best to use an Enum for fold names here.

maxnoe:
@kosack Thanks for the many comments, I also wanted to go over the code and have a look

ctapipe/image/time_next_neighbor_cleaning.py
maxnoe: All this hidden state in the class can go by properly making this a `TelescopeComponent` with a `subarray` and then giving a `tel_id` in `__call__`.

ctapipe/image/time_next_neighbor_cleaning.py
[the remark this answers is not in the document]
maxnoe: Currently it's just json, but we'll change it to something better suited.

ctapipe/image/time_next_neighbor_cleaning.py
```diff
@@ -0,0 +1,742 @@
...
+            array storing the steps in charge used to create the IPR graph
+
+        """
+        for event in source:
+            calibrator.calibrate(event)
+            for tel_id in event.r0.tels_with_data:
+                geometry = event.inst.subarray.tel[tel_id].camera
+                image = event.dl1.tel[tel_id].image[0]
+
+                n_samples = event.r0.tel[tel_id].num_samples
```
maxnoe, line 188: I am not sure it is correct to use the waveform num_samples and the sampling frequency here to estimate the rate.

Also: shouldn't the rate be dimensionless? I.e. just be the percentage of events that survive a given charge threshold?

@kpfrang

kpfrang: If I remember correctly, `sum_time` should really be the total duration (with units) of the trace. If this is correct, the calculation of `sum_time = n_samples * sample_time` seems correct to me.

This calculation is to generate fill the `IPR_dict`  which is equivalent to the graphs that are shown in Fig. 1 of the reference: 
https://galprop.stanford.edu/elibrary/icrc/2013/papers/icrc2013-0655.pdf

maxnoe: I don't understand why a time is needed at all. The goal is to have a "rate" of X percent pedestal events surviving, right?

So why do we need a "real" rate with some timeunit?

ctapipe/image/time_next_neighbor_cleaning.py
```diff
@@ -0,0 +1,742 @@
...
+        if not calculate:
+            lookup = dict(
+                LSTCam={"4nn": 75302, "3nn": 19269, "2nn": 5394, "2+1": 71712},
+                ASTRICam={"4nn": 41953, "3nn": 13548, "2nn": 4624, "2+1": 70904},
+                DigiCam={"4nn": 51843, "3nn": 13327, "2nn": 3744, "2+1": 49320},
+                FlashCam={"4nn": 71559, "3nn": 18319, "2nn": 5124, "2+1": 68112},
+                NectarCam={"4nn": 75302, "3nn": 19269, "2nn": 5394, "2+1": 71712},
+                CHEC={"4nn": 36289, "3nn": 11720, "2nn": 4000, "2+1": 61432},
+            )
+            comb_factor = lookup[camera_name][neighbor_group]
```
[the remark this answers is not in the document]
maxnoe: We should just always calculate and cache.

ctapipe/image/time_next_neighbor_cleaning.py
kpfrang: You can have a look at equations 1 and 2 of the proceeding. I think it gets clear from this.

Seems that I tried to convert the fake probability (which is the input parameter) into the accidental rate R_acc from these formulas in the line below. I don't remember how this is done in EventDisplay though. Probably worth checking this.
https://github.com/cta-observatory/ctapipe/blob/8d614a9a77c1d7b11523b22c9b67764e7c733bb6/ctapipe/image/time_next_neighbor_cleaning.py#L648

maxnoe: Here is a inverse cumulative plot of the pedestal pixel values, pretty much reproduces the figure (without any assumptions about "time"):

pixel_values

maxnoe: @kpfrang Do you remember which extractor you used? If it was the `NeighborPeakWindowSum` I think the times of neighboring pixels are correlated and thus the time range rejection might not work as expected, since that assumes uncorrelated times. Maybe that explains that more pedestal events than your target  rate you survived?

ctapipe/image/time_next_neighbor_cleaning.py
```diff
@@ -0,0 +1,742 @@
...
+    searched for groups of pixels with a given charge and time-coincidence.
+
+    It requires only one parameter which is the fake image proability. The
+    cuts in the Q-dT parameterspace then are determined from the individual
+    pixel rate which gives the rate in which the pixels have a charge above
+    a given threshold. Those graphs are calculated from calibration data
+    (i.e. images without Cherenkov signal but pure night sky background).
+    """
+
+    def __init__(self):
```
maxnoe, line 59: Nothing about the current event / image should be stored on the class. 

So please remove `_camera_name` `__neighbor_group` `_multiplicity` and just give that as arguments to the methods that need it

ctapipe/image/time_next_neighbor_cleaning.py
maxnoe: Change to `TimeNextNeighborCleaning(ImageCleaner)`

ctapipe/image/time_next_neighbor_cleaning.py
```diff
@@ -0,0 +1,742 @@
...
+        self.IPR_dict = None
+        self._multiplicity_dict = {"4nn": 4, "3nn": 3, "2+1": 3, "2nn": 2}
+        self.__geometry = None
+        self._camera_name = None
+        self._IPR = None
+        self.__neighbor_group = None
+        self._multiplicity = None
+
+    @property
+    def _geometry(self):
```
maxnoe, line 69: This method should be removed completely

ctapipe/image/time_next_neighbor_cleaning.py
```diff
@@ -0,0 +1,742 @@
...
+        """
+        Get or set the current `_geometry. Setting it to the new value also
+        sets `_camera_name` and selects the corresponding `_IPR` from `IPR_dict`.
+        """
+        return self.__geometry
+
+    @_geometry.setter
+    def _geometry(self, geometry):
```
maxnoe, line 77: This method should be removed completely

ctapipe/image/time_next_neighbor_cleaning.py
```diff
@@ -0,0 +1,742 @@
...
+        self._camera_name = geometry.camera_name
+        try:
+            self._IPR = self.IPR_dict[geometry.camera_name]
+        except KeyError as e:
+            raise KeyError(
+                f"No IPR graph for {geometry.camera_name} in" f" self.IPR_dict."
+            ) from e
+
+    @property
+    def _neighbor_group(self):
```
maxnoe, line 88: This method should be removed completely

ctapipe/image/time_next_neighbor_cleaning.py
```diff
@@ -0,0 +1,742 @@
...
+        """
+        Get or set the current _neighbor_group. Setting the _neighbor_group to
+        a new value also changes `_multiplicity' to the corresponding value.
+        """
+        return self.__neighbor_group
+
+    @_neighbor_group.setter
```
maxnoe, line 95: This method should be removed completely

ctapipe/image/time_next_neighbor_cleaning.py
```diff
@@ -0,0 +1,742 @@
...
+        trace_length: astropy.units.Quantity
+            total length of the trace, e.g. number of samples times sample_time
+        charge_steps: numpy.ndarray
+            array with the charge steps used to calculate the IPR graphs.
+        """
+
+        self.IPR_dict = self.IPR_dict if self.IPR_dict else {}
+
+        if geo.camera_name not in self.IPR_dict:
+            self.IPR_dict[geo.camera_name] = {
```
maxnoe, line 143: We should use the telescope_type here, not the camera name, probably even telescope_id, since noise rates can be different per telescope

ctapipe/image/time_next_neighbor_cleaning.py
```diff
@@ -0,0 +1,742 @@
...
+                "counter": 0,
+                "charge": charge_steps,
+                "npix": np.zeros(charge_steps.shape[0]),
+                "rate": np.zeros(charge_steps.shape[0]),
+                "rate_err": np.zeros(charge_steps.shape[0]),
+            }
+
+        # count number of pixels
+        self.IPR_dict[geo.camera_name]["counter"] += np.sum(image >= 0)
+        for thbin, charge in enumerate(self.IPR_dict[geo.camera_name]["charge"]):
```
maxnoe, line 153: This is very inefficient. The same can be achieved with a single call to `np.histogram(image, self.charge_edges, cumulative=-1)`

maxnoe: `density=True` to get a "cdf", which is I think what we want

ctapipe/image/time_next_neighbor_cleaning.py
```diff
@@ -0,0 +1,742 @@
...
+            self.IPR_dict[geo.camera_name]["npix"][thbin] += np.sum(image > charge)
+
+        # convert count of number of pixels above threshold into rate
+        sample_to_hz = (
```
maxnoe, line 157: I think we should just remove this and just store the counts / percentages I think.

ctapipe/image/time_next_neighbor_cleaning.py
```diff
@@ -0,0 +1,742 @@
...
+            1 / (trace_length * self.IPR_dict[geo.camera_name]["counter"])
+        ).to("Hz")
+        self.IPR_dict[geo.camera_name]["rate"] = (
+            self.IPR_dict[geo.camera_name]["npix"] * sample_to_hz
+        )
+        self.IPR_dict[geo.camera_name]["rate_err"] = (
+            np.sqrt(self.IPR_dict[geo.camera_name]["npix"]) * sample_to_hz
+        )
+
+    def calculate_ipr_from_source(self, source, calibrator, charge_steps):
```
maxnoe, line 167: This method is I think not really helpful to have here. We should probably just do a `from_images` that can than work with output from the process tool

ctapipe/image/time_next_neighbor_cleaning.py
```diff
@@ -0,0 +1,742 @@
...
+                geometry = event.inst.subarray.tel[tel_id].camera
+                image = event.dl1.tel[tel_id].image[0]
+
+                n_samples = event.r0.tel[tel_id].num_samples
+                sample_time = u.Quantity(event.mc.tel[tel_id].time_slice, u.ns)
+                sum_time = sample_time * n_samples
+
+                self.fill_individual_pixel_rate(geometry, image, sum_time, charge_steps)
+
+    def scaled_combfactor_event_display(self):
```
maxnoe, line 194: This method is not used at all and contains values from EventDisplay "nobody undestands", so please just remove

ctapipe/image/time_next_neighbor_cleaning.py
```diff
@@ -0,0 +1,742 @@
...
+            in time
+        """
+        if neighbor_group is not None:
+            self._neighbor_group = neighbor_group
+
+        ipr_graph = interp1d(
+            self._IPR["charge"], self._IPR["rate"], "linear", fill_value="extrapolate"
+        )
+
+        try:
```
maxnoe, line 309: I think this cannot raise an error anymore, since we now directly use `fill_value="extrapolate"`, so you can remove the `try` and the whole `except` block

ctapipe/image/time_next_neighbor_cleaning.py
```diff
@@ -0,0 +1,742 @@
...
+        neighbor_group: string
+            name of neighbor group. Select from 4nn, 3nn, 2+1 and 2nn
+
+        Returns
+        -------
+        time_coincidence: ndarray
+            array of same shape of `charge` filled with the corresponding cut
+            in time
+        """
+        if neighbor_group is not None:
```
maxnoe, line 302: Do not assign to self here, just use `neighbor_group` as given

ctapipe/image/time_next_neighbor_cleaning.py
kpfrang: Not quite sure, but I believe I used the `NeighborPeakWindowSum`.

ctapipe/image/tests/test_time_next_neighbor_cleaning.py
```diff
@@ -34,3 +34,17 @@ def test_clean(example_subarray, example_event):
         camera.geometry, image, peaktime, sample_time, sum_time=1 * u.ns
     )
     assert len(mask) == camera.geometry.n_pixels
+
+
+def test_calculate_ipr(tmp_path):
+    source = EventSource("dataset://gamma_test_large.simtel.gz")
```
maxnoe, line 40: You can use the file `lst_prod3_calibration_and_mcphotons.simtel.zst` with the option `skip_calibration_events=False` to actually use pedestal events for this test.

ctapipe/image/time_next_neighbor_cleaning.py
```diff
@@ -180,14 +205,15 @@ def calculate_ipr_from_source(self, source, calibrator, charge_steps):
 
         """
         for event in source:
```
maxnoe, line 207: You should probably skip all events that are not of type `SKY_PEDESTAL` like so:

```
if event.trigger.type != EventType.SKY_PEDESTAL:
    continue
```

ctapipe/image/time_next_neighbor_cleaning.py
kosack: could become a TelescopeParameter - that way you have the option to load an IPR profile per type or per telescope

ctapipe/image/time_next_neighbor_cleaning.py
```diff
@@ -0,0 +1,652 @@
...
+        neighbors_p2 = _get_neighbors_of(p2, indptr, indices)
+        common_neighbors = np.intersect1d(neighbors_p1, neighbors_p2).tolist()
+        if len(common_neighbors) == 2:
+            n_1, n_2 = common_neighbors
+            combinations.append((p1, p2, np.uint16(n_1), np.uint16(n_2)))
+
+    return np.array(combinations, dtype=np.uint16)
+
+
+def _get_groups_2_1nn(groups_2nn, indptr, indices):
```
maxnoe, line 60: njit missing?

ctapipe/image/time_next_neighbor_cleaning.py
```diff
@@ -0,0 +1,652 @@
...
+        neighbors_p2 = _get_neighbors_of(p2, indptr, indices)
+        for neighbor in np.intersect1d(neighbors_p1, neighbors_p2):
+            if neighbor > p1:
+                if neighbor > p2:
+                    combinations.append((p1, p2, np.uint16(neighbor)))
+
+    return np.array(combinations, dtype=np.uint16)
+
+
+def _get_groups_4nn_from_2nn(groups_2nn, indptr, indices):
```
maxnoe, line 46: njit missing?

ctapipe/image/time_next_neighbor_cleaning.py
```diff
@@ -0,0 +1,652 @@
...
+    for pixel_id in range(n_pixels):
+        neighbors = _get_neighbors_of(pixel_id, indptr, indices)
+        for neighbor in neighbors:
+            if neighbor > pixel_id:
+                combinations.append((pixel_id, neighbor))
+
+    return np.array(combinations, dtype=np.uint16)
+
+
+def _get_groups_3nn(groups_2nn, indptr, indices):
```
maxnoe, line 32: njit missing?

ctapipe/image/time_next_neighbor_cleaning.py
```diff
@@ -0,0 +1,652 @@
...
+import json
+
+
+def _get_neighbors_of(pixel_id, indptr, indices):
+    first = indptr[pixel_id]
+    last = indptr[pixel_id + 1]
+    return indices[first:last]
+
+
+def _get_groups_2nn(indptr, indices):
```
maxnoe, line 20: njit missing?

ctapipe/image/time_next_neighbor_cleaning.py
```diff
@@ -0,0 +1,652 @@
...
+from ..containers import EventType
+from functools import lru_cache
+from tqdm.auto import tqdm
+
+import astropy.units as u
+import warnings
+import json
+
+
+def _get_neighbors_of(pixel_id, indptr, indices):
```
maxnoe, line 14: njit missing?

ctapipe/image/time_next_neighbor_cleaning.py
```diff
@@ -0,0 +1,652 @@
...
+
+
+def _get_groups_3nn(groups_2nn, indptr, indices):
+    combinations = []
+    for group in groups_2nn:
+        p1, p2 = group
+        neighbors_p1 = _get_neighbors_of(p1, indptr, indices)
+        neighbors_p2 = _get_neighbors_of(p2, indptr, indices)
+        for neighbor in np.intersect1d(neighbors_p1, neighbors_p2):
+            if neighbor > p1:
```
maxnoe, line 39: combine to `if neighbor > p1 and neighbor > p2`

maxnoe: Are you sure it is correct here to request this? Why are there no pixel groups where this does not hold true?

ctapipe/image/time_next_neighbor_cleaning.py
```diff
@@ -0,0 +1,652 @@
...
+    return np.array(combinations, dtype=np.uint16)
+
+
+def _get_groups_4nn_from_2nn(groups_2nn, indptr, indices):
+    combinations = []
+    for group in groups_2nn:
+        p1, p2 = group
+        neighbors_p1 = _get_neighbors_of(p1, indptr, indices)
+        neighbors_p2 = _get_neighbors_of(p2, indptr, indices)
+        common_neighbors = np.intersect1d(neighbors_p1, neighbors_p2).tolist()
```
maxnoe, line 52: why `tolist`?

ctapipe/image/time_next_neighbor_cleaning.py
```diff
@@ -0,0 +1,652 @@
...
+
+@lru_cache(maxsize=None)
+def get_pixel_groups(geometry):
+    return _get_pixel_groups(
+        geometry.neighbor_matrix_sparse.indptr, geometry.neighbor_matrix_sparse.indices
+    )
+
+
+@njit
+def _is_group_in_mask(group, mask):
```
maxnoe, line 98: This is just: `np.all(mask[group])`

PR #1859, opened 2022-03-29
Require python 3.8, astropy 5

maxnoe:
Tests for 3.9, 3.10 to be added after #1330 is merged

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -19,7 +19,7 @@ jobs:
     strategy:
       matrix:
         os: [ubuntu-latest, macos-latest]
-        python-version: [3.7, 3.8]
+        python-version: ["3.8", ]
         install-method: ["pip", "conda"]
     defaults:
       run:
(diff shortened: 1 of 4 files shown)


kosack:
```
        with tables.open_file(path, "r") as f:
>           assert f.root.data.units.attrs["inverse_length_UNIT"] == "m-1"
E           AssertionError: assert 'm**-1' == 'm-1'
E             - m-1
E             + m**-1
E             ?  ++
```
Seems the default string representation for units has changed.  Either we should explicitly specify FITS representation when writing these attributes, or just change the test to read them back as `astropy.Unit` objects and compare those.

maxnoe:
> Seems the default string representation for units has changed. Either we should explicitly specify FITS representation when writing these attributes, or just change the test to read them back as astropy.Unit objects and compare those.

It's not the default, we do specify "vounit" to be compatible with the virtual observatory. Checking the vo units document, `m**-1` is correct.

maxnoe:
There was a new release of the VO Units document, version 1.0, which made this change upstream:

https://github.com/astropy/astropy/pull/11565

kosack:
Ah yes,  so it was a bug in older astropy versions

kosack (approved):

PR #1863, opened 2022-03-30
Use write_table for image statistics

maxnoe:
This was the last table in dl1 output that was still using the astropy mechanism. As of now, we don't have any `table_column_meta` tables in the output

### DIFF
--- a/ctapipe/tools/dl1_merge.py
+++ b/ctapipe/tools/dl1_merge.py
@@ -315,7 +315,8 @@ class MergeTool(Tool):
             )
 
             for node in service_nodes:
-                file.copy_node(node, newparent=target_group)
+                if node in file:
+                    file.copy_node(node, newparent=target_group)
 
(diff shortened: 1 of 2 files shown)


kosack (approved):
Great!

ctapipe/tools/dl1_merge.py
```diff
@@ -315,7 +315,8 @@ def add_image_statistics(self, file):
             )
 
             for node in service_nodes:
-                file.copy_node(node, newparent=target_group)
+                if node in file:
```
LukasNickel, line 318: Is this for the case of only dl1b in the file or whats the purpose here?

maxnoe: No, this is use for the `__table_column_meta` tables from astropy.

LukasNickel (approved):

PR #1864, opened 2022-03-30
Implement merging dl2 subarray data, fixes #1861

### LINKED ISSUE #1861: Add possibility to merge DL2 to merge tool

Merge tool should be able to merge DL2

maxnoe:


ctapipe/tools/merge.py
```diff
@@ -49,7 +37,9 @@
     "/dl1/event/telescope/images",
     "/dl1/service/image_statistics",
     "/dl1/service/image_statistics.__table_column_meta__",
+    "/dl2/event/subarray/geometry",
```
kosack, line 40: Should we also support `/dl2/event/telescope/geometry` in this? ctapipe doesn't yet write it, but I could imagine that lstchain or similar might (if they are using the DataWriter, it would get written). 

 Or we can leave that until it is implemented in ctapipe (once there is a proper machine learning module for mono geometry, energy, etc)

maxnoe: lstchain does not yet use the ctapipe format, so the merge tool is  not applicable anyways.

I though about it, but I went for only merging what we already have. If in future we implement something that would use it, we should make sure the code to merge it gets also added to the merge tool

maxnoe: E.g. I will take care of it when implementing telescope-wise energy / gammaness prediction in the ml module

kosack: Ok sounds good

kosack (approved):

ctapipe/tools/merge.py
```diff
@@ -226,29 +233,49 @@ def setup(self):
...
+                self.log.info("Merging observed data")
+                self.is_simulation = False
+            else:
+                self.log.info("Merging simulated data")
+                self.is_simulation = True
+
+            # do not try to merge optional nodes not present in first file
+            for node in filter(lambda n: n not in h5file, optional_nodes):
+                self.log.info(f"First file does not contain {node}, ignoring")
+                self.usable_nodes.remove(node)
```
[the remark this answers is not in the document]
maxnoe: MMh, not sure. I mainly added it to not get the warning per file that "images" are missing when merging a file that does not contain images. 

Thinking about it, when do we expect groups to miss completely from file?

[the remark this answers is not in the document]
Hckjs: > Thinking about it, when do we expect groups to miss completely from file?

Cant really think of an example for this... So should be fine as it is i think

kosack (approved):

Hckjs (approved):

Hckjs (approved):

LukasNickel (approved):

PR #1865, opened 2022-03-30
Update CI os/install-method/python-version combinations

maxnoe:
This adds and removes some builds in our CI config.

To not do too many different builds, instead of doing the full matrix of os / python-version / install-method I went for a sparse matrix of that.

There are now 6 combinatations covering python 3.8, 3.9, 3.10 with linux vs. macos and pip vs. conda.

On linux, all supported python versions are tested with the conda installation.

Additionally, I test the newest python version also with pip.

On macos, the oldest supported python version is tested with pip and conda

This made sense to me, but I don't have strong opinions on the combinations we want to test. I only felt that going for 18 different combinations is too much.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -18,9 +18,31 @@ jobs:
     runs-on: ${{ matrix.os }}
     strategy:
       matrix:
-        os: [ubuntu-latest, macos-latest]
-        python-version: ["3.8", ]
-        install-method: ["pip", "conda"]
+        include:
+          - os: ubuntu-latest
+            python-version: "3.8"
+            install-method: conda
+
+          - os: ubuntu-latest
+            python-version: "3.9"
+            install-method: conda
+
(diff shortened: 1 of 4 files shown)


kosack (dismissed):
Yes that sounds fine. The build matrix was getting a bit bloated.

kosack:
Related: If 3.10 works, should we drop 3.8?

maxnoe:
> Related: If 3.10 works, should we drop 3.8?

We decided to follow NEP29 along with astropy and gammapy, which would drop python 3.8 support only after 2023-04-13:
https://numpy.org/neps/nep-0029-deprecation_policy.html#drop-schedule

kosack:
Still would be nice to rely on the newer type hinting, etc in 3.9+ (e.g. hints like `list[int]`). Not critical though

kosack (approved):

LukasNickel (approved):

PR #1871, opened 2022-04-05
Update black version

LukasNickel:
- Click 8.1 broke black
- Fixed in black 22.3: https://github.com/psf/black/pull/2966


kosack (approved):
good idea to update, but do we use Click anywhere? It isn't a required dependency at least

maxnoe:
Black uses it and then the pre-commit hook would fail

maxnoe (approved):

PR #1873, opened 2022-04-11
Add switch to ImageProcessor to enable/disable ImageModifier

maxnoe:


LukasNickel:
Didnt we make the modifier a noop by default for this exact reason?
But sure, we can add a switch on top of it as well

kosack (dismissed):

maxnoe:
> Didnt we make the modifier a noop by default for this exact reason?
But sure, we can add a switch on top of it as well

But not in the way I had in mind. It still performance all the steps and is a bit expensive to do that. When I asked that in the review, I expected to see some clear checks  like:


```
if smear_factor > 0:
    # do smearing
```

so it might currently be a no op, but it's not really obvious from the code in `__call__`.

LukasNickel (approved):

maxnoe:
I changed to adding the explicit checks in `ImageModifier`, let me know if you think with that, the switch is still required. For now, I removed it again.

LukasNickel (dismissed):

LukasNickel (dismissed):

kosack (changes requested):
I still think it's necessary to have a global switch to skip the ImageModifier altogether, rather than skipping individual steps inside it if the options are set to nominal values (the latter is also nice, but not sufficient). 

 Reading it as a user, I see that it gets run, and it's in no way clear that it isn't doing something unexpected to the images, and I don't want to have to understand each of the options to ensure that.    I mean how can I be sure that smear factor isn't set to 0.00000001 or something, and i am performing an action that I don't want?    For safety, and to be sure no modification is being done, I would like a kill switch, essentially!

kosack (approved):

PR #1877, opened 2022-04-13
Draft: ML Energy Regressor Component/Tool

nbiederbeck:
This is still an early draft, but I am very open to any feedback.

I'm building an EnergyRegression `Component` for the EventSource Loop and a `Tool` for table-based application of the models.

Currently this is with a single model for all telescopes (types/ids), because I couldn't find a simple way of saving/loading multiple models (e.g. per telescope type) with what is currently available in the `ml`-branch.

*NOTE*: This is a feature built on PR #1767

### DIFF
--- a/ctapipe/ml/apply.py
+++ b/ctapipe/ml/apply.py
@@ -1,4 +1,3 @@
-import numpy as np
 from astropy.table import Table
 from ctapipe.containers import ArrayEventContainer, ReconstructedEnergyContainer
 from ctapipe.core import Component
@@ -36,11 +35,16 @@ class EnergyRegressor(EnergyReconstructor):
         for tel_id in event.trigger.tels_with_trigger:
             features = dict()
 
-            for name, container in event.dl1.tel[tel_id].parameters.as_dict().items():
+            for container in (
+                *event.dl1.tel[tel_id]
+                .parameters.as_dict(add_prefix=True, recursive=True)
+                .values(),
+                *event.dl2.tel[tel_id]
+                .as_dict(add_prefix=True, recursive=True)
+                .values(),
+            ):
                 for key, value in container.items():
-                    # TODO: find a better way of creating these values (prefix?)
-                    features.update({f"{name}_{key}": [value]})
-            # TODO: get all possible features
+                    features.update({key: [value]})
 
             feature_array = Table(features)
 
--- a/ctapipe/ml/apply.py
+++ b/ctapipe/ml/apply.py
@@ -2,11 +2,17 @@ from abc import abstractmethod
 
 import numpy as np
 from astropy.table import Table
-from ctapipe.containers import ArrayEventContainer
+from ctapipe.containers import (
+    ArrayEventContainer,
+    ParticleClassificationContainer,
+    ReconstructedEnergyContainer,
+)
 from ctapipe.core import Component
-from ctapipe.core.traits import Path
 
-from .sklearn import Classifier, Regressor
+from traitlets import Instance
+from .sklearn import Classifier, Regressor, Model
+
+from typing import Tuple
 
 
 class Reconstructor(Component):
@@ -14,12 +20,7 @@ class Reconstructor(Component):
 
     # TODO: update model config (?)
     #       only settings that make sense, e.g. verbose, n_jobs
-
-    model_path = Path(
-        default_value=None,
-        allow_none=False,
-        directory_ok=False,
-    ).tag(config=True)
+    model = Instance(Model).tag(config=True)
 
     def __init__(self, *args, **kwargs):
         super().__init__(*args, **kwargs)
@@ -28,15 +29,15 @@ class Reconstructor(Component):
     def __call__(self, event: ArrayEventContainer) -> None:
         """Event-wise prediction for the EventSource-Loop.
 
-        Fill the event.dl2.<your-feature> container.
+        Fill the event.dl2.<your-feature>[name] container.
 
         Parameters
         ----------
         event: ArrayEventContainer
         """
 
-    def predict(self, table: Table) -> np.array:
-        """Tool"""
+    def predict(self, table: Table) -> Tuple[np.ndarray, np.ndarray]:
+        """Predict on a table of events"""
         return self.model.predict(table)
 
     def _collect_features(self, event: ArrayEventContainer, tel_id: int) -> Table:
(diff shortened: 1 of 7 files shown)


kosack:
How is this different from #1767 ?  

We actually had a class almost exactly like this one (`ctapipe.reco.EnergyRegressor`) in the past,  but it was removed in order to have a more general way of doing all machine learning steps: energy, gammaness, reconstruction quality (e.g. event PSF type), and mono-reconstruction (disp) are all essentially the same train/predict algorithm, so can be done in nearly identical ways.

ctapipe/ml/apply.py
```diff
@@ -0,0 +1,59 @@
...
+        self.model.model.verbose = 0
+        # TODO: use config system
+        self.features = features
+
+    def __call__(self, event: ArrayEventContainer) -> ReconstructedEnergyContainer:
+        """EventSource Loop"""
+        for tel_id in event.trigger.tels_with_trigger:
+            features = dict()
+
+            for name, container in event.dl1.tel[tel_id].parameters.as_dict().items():
```
maxnoe, line 39: `as_dict(prefix=True)`

maxnoe:
> How is this related to (or different from) https://github.com/cta-observatory/ctapipe/pull/1767 ?

This PR targets the branch of #1767, it's not targeting master

kosack:
I added that this is part of #1767 in the info, since that wasn't totally clear.  Then I change my question to: could you make this more general? I think a base class (or Protocol) for all "prediction" classes would be nice, that way Energy, Gammaness, Disp, QualityClass, etc could be done in a similar way.  They all have similar needs. At least the first two need to make predictions per telescope and have a function to combine them, for example.

ctapipe/ml/apply.py
```diff
@@ -0,0 +1,59 @@
...
+            # TODO: since a single nan is enough for a single row to be discarded,
+            #       find it earlier for small performance reasons
+            if np.count_nonzero(feature_array.to_pandas().isna()):
+                continue
+            else:
+                prediction = self.model.predict(feature_array)
+                event.dl2.tel[tel_id].energy.energy = prediction
+                event.dl2.tel[tel_id].energy.is_valid = True
+                event.dl2.tel[tel_id].energy.tel_ids = [tel_id]
+
```
kosack, line 56: Should we also do the stereo prediction here as well ?  In general it would just be a function to call that takes the mono predictions + image parameters and does some sort of weighted average to produce the final prediction.  However, the user might want to be able to choose from several methods, so *maybe a separate Component run afterwards is better*.  What do you think?  (maybe you were already considering that)

A few more things to think about (with no real idea of how to support them):
- There also may be some ML algorithms that input the mono features for all telescopes and output directly a stereo prediction without doing mono predictions first (e.g. a recurrent neural network could potentially do that).  
- Some reconstruction algorithms predict the energy without machine learning (I guess that is ok, they would just write to the appropriate place in the Event Structure).  We might consider if using those predictions as features could still be interesting: e.g. include the ImPACT energy prediction in a Hillas-style energy regressor?

maxnoe: I would separate the actual mono prediction from combining them into a stereo prediction because there might be several methods to do this and even algorithms that might even directly give a stereo prediction (like impact)

kosack: yes, that's basically what I said above - I think it is the way to go.  Probably a skeleton of this should be included in this PR then and we can see how it would work.  The initial implement could be super simple, just take the mono predictions and weight by `hillas.intensity`

nbiederbeck:
I will try to summarize my answers here:
- on neural networks: for this PR I want to concentrate on the "classic" algorithms that are supported by sklearn (i.e. feature-based approaches)
- I plan to (allow to) include stereo parameters where applicable
- for the time being I will limit this to telescope-based predictions (I agree with you that merging that should be a second step)
- I will try to come up with a clever way for base-classes for predictions (regression/classification). I just started with energy, because in #1767 was alreay a Energy Trainer implemented ([b354ed9](https://github.com/cta-observatory/ctapipe/pull/1767/commits/b354ed9b17353cc62e23d97f3fb5cacbd9ee2336))
- I reverted to one model to use the implemented loading/saving (`Model.write`, `Model.load`), instead of pickling a dictionary. Maybe we can find a way to support this more, e.g. someone using different algorithms for different telescope types? That is open for discussion

ctapipe/ml/apply.py
nbiederbeck: done

nbiederbeck:
I wanted to do something different with the git rebase, will fix this after the weekend

ctapipe/ml/apply.py
```diff
@@ -0,0 +1,116 @@
...
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+
+    def __call__(self, event: ArrayEventContainer) -> None:
+        for tel_id in event.trigger.tels_with_trigger:
+            features = self._collect_features(event, tel_id)
+            prediction = self.model.predict(features)
+
+            event.dl2.tel[tel_id].energy.energy = prediction
+            event.dl2.tel[tel_id].energy.is_valid = True
```
maxnoe, line 98: Should be `np.isfinite(energy)`

ctapipe/ml/apply.py
```diff
@@ -0,0 +1,116 @@
...
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+
+    def __call__(self, event: ArrayEventContainer) -> None:
+        for tel_id in event.trigger.tels_with_trigger:
+            features = self._collect_features(event, tel_id)
+            prediction = self.model.predict(features)
+
+            event.dl2.tel[tel_id].classification.prediction = prediction
+            event.dl2.tel[tel_id].classification.is_valid = True
```
maxnoe, line 114: Should use the valid information from the model.predict function

ctapipe/ml/tools/apply_energy_regressor.py
```diff
@@ -0,0 +1,56 @@
...
+        )
+        self.estimator = EnergyRegressor(parent=self)
+
+    def start(self):
+        self.log.info("Applying model")
+        table = self.loader.read_telescope_events()
+
+        prediction = Table({self.prediction_key: self.estimator.predict(table)})
+        prediction[self.prediction_key].description = "Estimated energy"
+        write_table(prediction, self.output_path, "/dl2/telescope/energy")
```
maxnoe, line 49: This should be done by telescope id

PR #1880, opened 2022-04-14
added a better template for release-drafter

kosack:
Auto-categorizes PRs by topic.  Should make releasing a bit nicer.

The template should look similar to what is used in protopipe here: 
https://github.com/cta-observatory/protopipe/releases

(though I removed the unicode icons)

### DIFF
--- a/.github/release-drafter.yml
+++ b/.github/release-drafter.yml
@@ -36,7 +36,7 @@ template: |
 
   $CONTRIBUTORS
 
-  ## What is changed since $PREVIOUS_TAG
+  ## What has changed since $PREVIOUS_TAG
 
   Pull-requests containing changes of multiple nature are repeated.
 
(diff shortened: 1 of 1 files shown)


maxnoe:
That means we have to be more careful with applying labels though, right?

Do we first want to maybe go over the once we want to use? Right now, we have *a lot*

kosack:
> That means we have to be more careful with applying labels though, right?

Yes, though you can still edit it afterward, it just helps to organize the changelog.

> Do we first want to maybe go over the once we want to use? Right now, we have a lot

I was noticing that when i went through them to make these categories.  I think I got it right, but there are a few that are clearly duplicates, and some are "theme" categories (like "calibration", "reconstruction") while others are "change" categories (new  functionality, bug).  

We could clean them up, or at least rename/recolor some of them. But that shouldn't affect this PR much, except if we rename them.

maxnoe (approved):

kosack:
Can we merge this so we can test it for future releases?

maxnoe:
@LukasNickel

maxnoe (approved):

nbiederbeck (approved):

PR #1883, opened 2022-04-19
 Fix setting up logging hierarchy, fixes #1882 

### LINKED ISSUE #1882: Component logging hierarchy setup is broken
**Describe the bug**
Setting up the hierarchy of loggers is currently broken (and at least since 0.12). 

The code in `Component` checks `self.parent` before calling `super().__init__()` so `self.parent` is always None and the hierarchy is not setup correctly.

**To Reproduce**
```
from ctapipe.core import Tool, Component


class Foo(Component):
    def __init__(self, **kwargs):
        super().__init__(**kwargs)
        print(self.log)


class Bar(Component):
    def __init__(self, **kwargs):
        super().__init__(**kwargs)
        print(self.log)
        self.foo = Foo(parent=self)

class Baz(Tool):
    name = 'baz'

    def setup(self):
        self.bar = Bar(parent=self)

    def start(self):
        print(self.log)


if __name__ == '__main__':
    Baz().run()
```

prints

```
<Logger __main__.Bar (WARNING)>
<Logger __main__.Foo (WARNING)>
<Logger ctapipe.baz (WARNING)>
```

**Expected behavior**
Hierarchy is setup correctly.

maxnoe:


kosack (dismissed):

maxnoe:
@kosack Sorry, it has to come before `super().__init__()` to enable using the log in traitlets observers, as it is used in the examples. Remove the `self.` works though.

kosack (approved):

nbiederbeck (approved):

PR #1886, opened 2022-04-20
update README for v0.14.0

kosack:
readme was still on 0.11...


maxnoe:
Should we just remove the list from the Readme and just point to the release page and the latest doi?

The Readme can by definition never be correct on the tag...

kosack:
It's usually the first place people look, but we could just point to the docs which have install instructions (which are also probably out of date... should update them to at least automatically include the correct checkout command).  For now, let's update the README, but I will make a better docs page after

kosack:
However just showing the top-level DOI is probably fine. No need for the full list here.

maxnoe (approved):

nbiederbeck (approved):

PR #1888, opened 2022-04-21
Table quality query

nbiederbeck:
For the ML work in #1877 (and #1767) I implemented a QualityQuery for astropy tables.

In the long run this might best be suited inside the currently existing QualityQuery, but for a quick small PR I added this here.

### DIFF
--- a/ctapipe/core/qualityquery.py
+++ b/ctapipe/core/qualityquery.py
@@ -153,16 +153,64 @@ class TableQualityQuery(Component):
 
     def __init__(self, *args, **kwargs):
         super().__init__(*args, **kwargs)
-        self._selectors = "(" + ") & (".join(self.quality_criteria) + ")"
+
+        self.quality_criteria.insert(0, ("TOTAL", "True"))
+
+        self._selectors = [crit for _, crit in self.quality_criteria]
+        self.selection_function_strings = self._selectors
+        self.criteria_names = [name for name, _ in self.quality_criteria]
+
+        self._counts = np.zeros(len(self._selectors), dtype=np.int64)
+        self._cumulative_counts = np.zeros(len(self._selectors), dtype=np.int64)
 
     def __call__(self, table) -> np.array:
         """Check all quality_criteria on an astropy table."""
-        try:
-            valid = numexpr.evaluate(self._selectors, table)
-        except KeyError as key:
-            raise QualityCriteriaError(
-                f"Couldn't evaluate selection expression '{self._selectors}' "
-                f"because key {key} was not found in table."
-            ) from None
+        valid = []
+        for criterium in self._selectors:
+            if criterium == "True":
+                # otherwise numexpr returns a single True, not an array
+                continue
+            try:
+                valid.append(numexpr.evaluate(criterium, table))
+            except KeyError as key:
+                raise QualityCriteriaError(
+                    f"Couldn't evaluate selection expression '{criterium}' "
+                    f"because key {key} was not found in table."
+                ) from None
 
-        return valid
+        # re-attach TOTAL criterium
+        valid.insert(0, np.ones(len(table), dtype=np.bool_))
+        counts = np.sum(valid, axis=1)
+        self._counts += counts
+        self._cumulative_counts += np.cumsum(counts, axis=0)
+
+        return np.prod(valid[1:], axis=0)  # skip TOTAL criterium
+
+    def to_table(self, functions=False):
+        """
+        Return a tabular view of the latest quality summary
+
+        The columns are
+        - *criteria*: name of each criterion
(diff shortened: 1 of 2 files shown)


maxnoe (changes requested):
This is good and need, however, the main feature of the `QualityQuery` is that it keeps track of the statistics for each of the cut (both individually and cumulative).

Can we find a way to have the normal quality query and the table quality query work the same? Maybe even also use numexpr for the normal quality query instead of these lambdas:

So instead of doing:

```
eval("lambda image: ...")(image), do `numexpr("...", {"image": image, ...})   
```

also for the normal quality query

kosack:
+1 for switching everything to use numexpr.

I think it would also be easy to keep track of statistics here as well, and maybe we could just merge the two classes so the same configuration can be used event-wise or for a full table? It seems to me it would be easy to have it work in both cases.  E.g. just add a `QualityQuery.query_table(table: astropy.table.Table)` method.

nbiederbeck:
>  just add a QualityQuery.query_table(table: astropy.table.Table) method.

with you on that one.  will implement both with numpexpr in a single class

nbiederbeck:
### Problems

1. We have some cuts like `["enough_charge", "lambda im: im.sum() > 50"]`, which numexpr cannot do ([Docs](https://numexpr.readthedocs.io/projects/NumExpr3/en/latest/user_guide.html#supported-reduction-operations)):
```python
>>> import numexpr
>>> import numpy as np
>>> image = np.arange(10)
>>> numexpr.evaluate('sum(image) > 10')
  File "/home/nbiederbeck/.local/conda/envs/cta-dev/lib/python3.8/site-packages/numexpr/necompiler.py", line 820, in evaluate
    compiled_ex = _numexpr_cache[numexpr_key]
KeyError: ('sum(image) > 10', (('optimization', 'aggressive'), ('truediv', False)), (('image', <class 'numpy.int64'>),))

During handling of the above exception, another exception occurred:

  File "<stdin>", line 1, in <module>
  File "/home/nbiederbeck/.local/conda/envs/cta-dev/lib/python3.8/site-packages/numexpr/necompiler.py", line 822, in evaluate
    compiled_ex = _numexpr_cache[numexpr_key] = NumExpr(ex, signature, **context)
  File "/home/nbiederbeck/.local/conda/envs/cta-dev/lib/python3.8/site-packages/numexpr/necompiler.py", line 623, in NumExpr
    return interpreter.NumExpr(inputsig.encode('ascii'),
RuntimeError: invalid program: reduction operations must occur last
```

2. Regarding the table-query and counting: having a single AND-query `(x > 0) & (y > 0)` is effectively n times faster than combining n evaluations (but we can't count anymore)

3. Also using numexpr one cannot use subscriptions (`value.othervalue` or `value['othervalue']`):
```python
>>> import numexpr
>>> import numpy as np
>>> from astropy.table import Table
>>> hillas = Table({'intensity': np.arange(10)})
>>> numexpr.evaluate("hillas['intensity'] > 0")
  File "<stdin>", line 1, in <module>
  File "/home/nbiederbeck/.local/conda/envs/cta-dev/lib/python3.8/site-packages/numexpr/necompiler.py", line 809, in evaluate
    _names_cache[expr_key] = getExprNames(ex, context)
  File "/home/nbiederbeck/.local/conda/envs/cta-dev/lib/python3.8/site-packages/numexpr/necompiler.py", line 696, in getExprNames
    ex = stringToExpression(text, {}, context)
  File "/home/nbiederbeck/.local/conda/envs/cta-dev/lib/python3.8/site-packages/numexpr/necompiler.py", line 289, in stringToExpression
    ex = eval(c, names)
  File "<expr>", line 1, in <module>
TypeError: 'VariableNode' object is not subscriptable
```

### Suggested Solutions
- support in a single class both numexpr and the current lambda-style for the Query and try to be smart when building the `_selectors`
- or: keep it as is in this PR in two seperate classes (one lambda-style event-wise, one numexpr table-wise where the problems above would not occure) (would then still have to implement the counting for table selections)

kosack:
> We have some cuts like ["enough_charge", "lambda im: im.sum() > 50"], which

I haven't looked at this in detail to see what the exact problem is, but is it possible to re-write those cuts in a form that numexpr likes?  those are just examples, so we can easily change them as long as they achieve the same result.

maxnoe:
@kosack Unfortunately we didn't find a way of expressing all queries we will need using numexpr due to this specific current limitation:

https://numexpr.readthedocs.io/projects/NumExpr3/en/latest/user_guide.html#supported-reduction-operations

> Note: because of internal limitations, reduction operations must appear the last in the stack. 

So for some internal reason, numexpr cannot do e.g.: `image.sum() > 50`

maxnoe:
This was superseded by  #1943

PR #1889, opened 2022-04-22
Improved configuration of reference metadata

kosack:
## use case:

The user expects that the metadata in the output file's header is useful, for example to build a file catalog necessary to retrieve the data later.

## what this PR adds:

Part of the reference metadata was already configurable (the Contact info), this adds two additional user-configurable  parts: 
- the `Instrument` metadata (not to be confused with the InstrumentDescription, this is just id information about the instrument, in this case the subarray used). This lets the user set the subarray name, version, and other instrument-related attributes.  If they are not specified, they are automatically filled as before (or left unspecified)
- `CONTEXT` metadata, which is fully user-defined and allows an additional set of context-specific headers to be added as key-value pairs.

## changes

- Added user-defined CONTEXT metadata, which can be added to the config
file
- Made `metadata.Instrument` a configurable class similar to
`metadata.Contact` so values can be overridden in config files

The result is one can create a config file for a subarray definition like:

```yaml
EventSource:
  allowed_tels: [1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 19, 35]

DataWriter:
  Instrument:  # this is the new configurable info
    site: CTA-North
    class_: Subarray
    id_: alpha
    version: Prod5b
```
And that information will be written to the file's reference headers:

```sh
ctapipe-fileinfo output.h5
...
        INSTRUMENT:
            CLASS: Subarray
            ID: alpha
            SITE: CTA-North
            SUBTYPE: unspecified
            TYPE: unspecified
            VERSION: Prod5b
...
```

Similarly the arbitrary CONTEXT metadata can be  set as follows:

```yaml
DataWriter:
  context_metadata:
    "SIMULATION PRODUCTION": prod5b
    "SIMULATION DATASET":  Prod5b_Paranal_AdvancedBaseline_NSB1x_South_40deg
    "DESCRIPTION": "This is just a test"
```
Using spaces between keys allows a hierarchy (we could also support dots or similar)

```
ctapipe-fileinfo output.h5
...
    CONTEXT:
        DESCRIPTION: This is just a test
        SIMULATION:
            DATASET: Prod5b_Paranal_AdvancedBaseline_NSB1x_South_40deg
            PRODUCTION: prod5b
...
```
This is useful for example for storing parameters that are not automatically read from the MC files into the `SimulationConfigContainer` , or for storing things like calibration versions, etc.

## Still todo:
- [x] make tests
## issues
-  due to how the config system works, you cannot "chain" context metadata, meaning if you specify it in two included config files, the latter one fully replaces the former's dictionary, rather than adding to it.
- the `id_` and `class_` attributes have underscores to avoid linter/IDE complaints about python name conflicts (even if they do not conflict).  This is a bit ugly, could consider removing them and just deal with complaints

### DIFF
--- a/ctapipe/io/datawriter.py
+++ b/ctapipe/io/datawriter.py
@@ -148,8 +148,8 @@ class DataWriter(Component):
             "Additional metadata keywords and values that describe this data. "
             "This should be a dictionary where the keys will be appended to the "
             "CONTEXT section of the output file's attributes. Keys can be hierarchical "
-            "by using a space between each level, e.g. `SIMULATION PRODUCTION` would make a key "
-            "PRODUCTION grouped under the key SIMULATION"
+            "by using a space between each level, e.g. `SIMULATION PRODUCTION` "
+            "would make a key PRODUCTION grouped under the key SIMULATION"
         )
     ).tag(config=True)
 
--- a/ctapipe/io/datawriter.py
+++ b/ctapipe/io/datawriter.py
@@ -6,12 +6,12 @@ Class to write DL1 (a,b) and DL2 (a) data from an event stream
 
 import pathlib
 from collections import defaultdict
-from traitlets import Instance, Dict
 from typing import List
 
 import numpy as np
 import tables
 from astropy import units as u
+from traitlets import Dict, Instance
 
 from ..containers import (
     ArrayEventContainer,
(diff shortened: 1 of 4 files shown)


maxnoe:
> the id_ and class_ attributes have underscores to avoid linter/IDE complaints about python name conflicts (even if they do not conflict). 

This is true for `id` which is a builtin function that can be shadowed, but not for `class` which is a keyword.

kosack:
> This is true for id which is a builtin function that can be shadowed, but not for class which is a keyword

True, so maybe I'll just stick with the underscores (since they are at least needed for the id column). It's not too bad.   Or I could rename it `inst_id`.

maxnoe (approved):

PR #1894, opened 2022-04-25
Try building docs on 3.10

maxnoe:


kosack:
```
ImportError: cannot import name 'Union' from 'types' (/opt/hostedtoolcache/Python/3.10.4/x64/lib/python3.10/types.py)
```
I know in 3.10 you an now use a `|` instead of `Union` (`List[int] | int`) , but did they actually remove `Union`?

kosack:
Seems to be fixed in Sphinx 4.2.0 (see e.g. `https://github.com/sphinx-doc/sphinx/issues/9562`).  Maybe also update sphinx

maxnoe:
>  Though I guess that will cause the problems we had earlier...

Yes, this was just to test if we run into the same issue as reported by Gernot on slack. Our docs currently run into the issue that prevent building warning-less with 4.x. Let's hope the automodapi issue gets resolved soon so we can update to 4.x

PR #1896, opened 2022-04-26
Use astropy table indices to fix slow pytables query

maxnoe:
Goes from 1.x Events/s to ~175 Event/s on my machine for looping through a very large merged file (7e6 telescope events)

### DIFF
--- a/ctapipe/io/hdf5eventsource.py
+++ b/ctapipe/io/hdf5eventsource.py
@@ -5,6 +5,8 @@ import numpy as np
 import tables
 from ast import literal_eval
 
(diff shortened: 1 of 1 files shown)


nbiederbeck (approved):

maxnoe:
We just found that creating pytables indices on these columns results in the same speed improvement without any code changes here. So we might not want to merge this.

maxnoe:
Yes, I think so.

maxnoe:
I found a better solution not relying on indices or tables that is 5 times faster for a largish file.

PR #1899, opened 2022-04-26
compute impact distance to shower axis

### LINKED ISSUE #1892: Store telescope impact distance in simulation container
sure, it can be computed from those values, but the calculations are not trivial and require also digging out the telescope positions from the array description (and making sure that everything is in the same reference frame) 
since it is really useful, and  it is just one number per telescope I think it is really worth to keep it stored

_Originally posted by @jsitarek in https://github.com/cta-observatory/ctapipe/issues/1891#issuecomment-1106439147_

kosack:
Creating some functions to correctly compute and store impact distance from the shower axis to the telescopes on the ground.  Fixes #1892 

To do: 
* [x] tests for reading from events
* [x] add to ShowerProcessor (or make a Component)
* [x] figure out where to put this in the output for both reco and simulation 
* [x] update `TableLoader` to read and merge this information
* [x] ~~update `ctapipe-merge` to merge this information~~ → deferred to another PR...

For the output, it could be an array column in the  shower-wise ReconstructedShowerContainer in `/{dl2,simulation}/event/subarray/geometry/HillasReconstructor`, for example, or should we put it per-telescope in the `/{dl2,simulation}/event/telescope/`?  It's really a per-telescope quantity, but it's more useful in the per-shower container.

### DIFF
--- a/ctapipe/conftest.py
+++ b/ctapipe/conftest.py
@@ -176,7 +176,7 @@ def dl2_shower_geometry_file(dl2_tmp_path, prod5_gamma_simtel_path):
             f"--input={prod5_gamma_simtel_path}",
             f"--output={output}",
             "--write-images",
-            "--write-stereo-shower",
+            "--write-showers",
             "--max-events=20",
         ]
         assert run_tool(ProcessorTool(), argv=argv, cwd=dl2_tmp_path) == 0
@@ -202,7 +202,7 @@ def dl2_proton_geometry_file(dl2_tmp_path, prod5_proton_simtel_path):
             f"--input={prod5_proton_simtel_path}",
             f"--output={output}",
             "--write-images",
-            "--write-stereo-shower",
+            "--write-showers",
             "--max-events=20",
         ]
         assert run_tool(ProcessorTool(), argv=argv, cwd=dl2_tmp_path) == 0
@@ -228,7 +228,7 @@ def dl2_shower_geometry_file_type(dl2_tmp_path, prod5_gamma_simtel_path):
             f"--input={prod5_gamma_simtel_path}",
             f"--output={output}",
             "--write-images",
-            "--write-stereo-shower",
+            "--write-showers",
             "--max-events=20",
             "--DataWriter.split_datasets_by=tel_type",
         ]
--- a/ctapipe/containers.py
+++ b/ctapipe/containers.py
@@ -508,6 +508,10 @@ class SimulatedCameraContainer(Container):
         None, "Parameters derived from the true_image", type=ImageParametersContainer
     )
 
+    true_impact_distance = Field(
+        nan * u.m, "Distance from shower axis to telescope position", unit=u.m
+    )
+
 
 class SimulatedEventContainer(Container):
     shower = Field(SimulatedShowerContainer(), "True event information")
--- a/ctapipe/containers.py
+++ b/ctapipe/containers.py
@@ -92,7 +92,7 @@ class EventType(enum.Enum):
 
 
 class EventIndexContainer(Container):
-    """ index columns to include in event lists, common to all data levels"""
+    """index columns to include in event lists, common to all data levels"""
 
     container_prefix = ""  # don't want to prefix these
     obs_id = Field(0, "observation identifier")
@@ -251,7 +251,7 @@ class TimingParametersContainer(BaseTimingParametersContainer):
 
 
 class MorphologyContainer(Container):
-    """ Parameters related to pixels surviving image cleaning """
+    """Parameters related to pixels surviving image cleaning"""
 
     num_pixels = Field(-1, "Number of usable pixels")
     num_islands = Field(-1, "Number of distinct islands in the image")
@@ -287,7 +287,7 @@ class CoreParametersContainer(Container):
 
 
 class ImageParametersContainer(Container):
-    """ Collection of image parameters """
+    """Collection of image parameters"""
 
     container_prefix = "params"
     hillas = Field(
@@ -353,7 +353,7 @@ class DL1CameraContainer(Container):
 
 
 class DL1Container(Container):
-    """ DL1 Calibrated Camera Images and associated data"""
+    """DL1 Calibrated Camera Images and associated data"""
 
     tel = Field(Map(DL1CameraContainer), "map of tel_id to DL1CameraContainer")
 
@@ -475,7 +475,7 @@ class SimulatedShowerContainer(Container):
     core_y = Field(nan * u.m, "Simulated core position (y)", unit=u.m)
     h_first_int = Field(nan * u.m, "Height of first interaction", unit=u.m)
     x_max = Field(
-        nan * u.g / (u.cm ** 2), "Simulated Xmax value", unit=u.g / (u.cm ** 2)
+        nan * u.g / (u.cm**2), "Simulated Xmax value", unit=u.g / (u.cm**2)
     )
     shower_primary_id = Field(
         -1,
@@ -606,37 +606,33 @@ class ReconstructedGeometryContainer(Container):
         nan * u.m, "reconstructed x coordinate of the core position", unit=u.m
     )
     core_y = Field(
-        nan * u.m,
-        "reconstructed y coordinate of the core position",
-        unit=u.m
+        nan * u.m, "reconstructed y coordinate of the core position", unit=u.m
     )
     core_uncert_x = Field(
         nan * u.m,
         "reconstructed core position uncertainty along ground frame X axis",
-        unit=u.m
+        unit=u.m,
     )
     core_uncert_y = Field(
         nan * u.m,
         "reconstructed core position uncertainty along ground frame Y axis",
-        unit=u.m
+        unit=u.m,
     )
     core_tilted_x = Field(
         nan * u.m, "reconstructed x coordinate of the core position", unit=u.m
     )
     core_tilted_y = Field(
-        nan * u.m,
-        "reconstructed y coordinate of the core position",
-        unit=u.m
+        nan * u.m, "reconstructed y coordinate of the core position", unit=u.m
     )
     core_tilted_uncert_x = Field(
         nan * u.m,
         "reconstructed core position uncertainty along tilted frame X axis",
-        unit=u.m
+        unit=u.m,
     )
     core_tilted_uncert_y = Field(
         nan * u.m,
         "reconstructed core position uncertainty along tilted frame Y axis",
-        unit=u.m
+        unit=u.m,
     )
     h_max = Field(nan * u.m, "reconstructed height of the shower maximum", unit=u.m)
     h_max_uncert = Field(nan * u.m, "uncertainty of h_max", unit=u.m)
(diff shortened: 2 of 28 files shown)


kosack:
Computing from a DL2 event table and SubarrayDescription:
```py3
directions = SkyCoord(az = showers['az'], alt=showers['alt'], frame='altaz').cartesian.xyz.T.value
points  = np.column_stack([showers['core_x'], showers['core_y'], np.zeros_like(showers['core_y'])])
test_points = sub.tel_coords.cartesian.xyz.T.to_value("m")

impacts = np.array([impact_distance(p, dir, test_points) for p,dir in zip(points, directions)])
```
(I'm sure there is a more numpy-way to implement this)

image

ctapipe/reco/impact_distance.py
```diff
@@ -0,0 +1,108 @@
...
+        [
+            shower_geom.core_x.to_value(u.m),
+            shower_geom.core_y.to_value(u.m),
+            0,
+        ]
+    )
+
+    # sky direction of the shower (defines the shower axis)
+    sky_direction = SkyCoord(
+        alt=shower_geom.alt, az=shower_geom.az, frame="altaz"
```
jsitarek, line 61: Hi @kosack 
I think I know where the problem might be. 
Geographic azimuth is computed from the north axis towards east, so e.g. alt=0, az=45 deg should correspond to a vector [0.707, -0.707, 0], but when you do SkyCoords (alt, az, frame ="altaz").cartesian.xyz it makes [0.707, 0.707, 0], like it would be computed from x direction towards y. 
If I add here a minus sign before shower_geom.az the results of both methods agree (and I hope that provide correct values :-) )

maxnoe: Maybe related:

maxnoe: @jsitarek But we don't use `AltAz.cartesian` in the TiltedFrame transformations, we build a rotation matrix directly from `alt` and `az`...

kosack: I do use the cartesian attribute when turning the alt/az to a cartesian direction. I should probably just to the matrix explicitly to avoid SkyCoord slowness

maxnoe: https://github.com/astropy/astropy/issues/5300

kosack: Ok should agree now. And the 3d method is even faster without the SkyCoord overhead.

ctapipe/reco/impact_distance.py
```diff
@@ -57,9 +58,10 @@ def shower_impact_distance(
     )
 
     # sky direction of the shower (defines the shower axis)
-    sky_direction = SkyCoord(
-        alt=shower_geom.alt, az=shower_geom.az, frame="altaz"
-    ).cartesian.xyz.value
+    # NOTE: remember that in sky direction the azimuth needs to be negative
+    sky_direction = spherical_to_cartesian(
```
maxnoe, line 62: I think at this point we should define a `altaz_to_righthanded_cartesian` in `ctapipe.coordinates`.

ctapipe/reco/impact_distance.py
```diff
@@ -7,6 +7,7 @@
 from ..containers import ReconstructedGeometryContainer
 from ..coordinates import GroundFrame, TiltedGroundFrame
 from ..instrument.subarray import SubarrayDescription
+from .hillas_reconstructor import spherical_to_cartesian
```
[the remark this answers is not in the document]
kosack: yeah, I spotted that right after committing... seems my editor has auto-import now

ctapipe/coordinates/__init__.py
```diff
@@ -59,3 +61,19 @@ def altaz_to_altaz(from_coo, to_frame):
...
+    See e.g: https://github.com/astropy/astropy/issues/5300
+
+    Parameters
+    ----------
+    alt: u.Quantity
+        altitude
+    az: u.Quantity
+        azimuth
+    """
+    return spherical_to_cartesian(r=0, lat=alt, lon=-az)
```
maxnoe, line 79: r should be 1, right?

kosack: oops, yes.

jsitarek:
and what about the true impact? /event/simulation/... ?

kosack:
> and what about the true impact? /event/simulation/... ?

I haven't added that yet but will do that next if this storage layout looks ok to everyone.  I guess it should be computed in the SimTelEventSource, not in ShowerProcessor, so I'll add it there.

kosack:
Note that this PR seems to slow down data processing quite a bit (factor of 1.5 to 2...) So some optimization is probably needed

maxnoe (commented):
Should we rename the `SimulatedCameraContainer` to `SimulatedTelescopeContainer` and just use `/simulation/event/telescope/tel_xxx`?

kosack:
Just a quick check on the CTA-North data:
image
Looks reasonable, about a 5% reconstruction error.

kosack:
This is ready to  review - i decided to not implement the `ctapipe-merge` functionality since I think that is more complex and  I'll need some help with it - I noticed that DL2 is not very well implemented so far, so we should do that all at once.  That can come in another PR.

maxnoe:
>  I noticed that DL2 is not very well implemented so far, so we should do that all at once. That can come in another PR.

Merging dl2 works fine in the current master, what do you mean with "not very well implemented"?

ctapipe/reco/impact_distance.py
```diff
@@ -0,0 +1,111 @@
...
+        ]
+    )
+
+    # sky direction of the shower (defines the shower axis)
+    sky_direction = altaz_to_righthanded_cartesian(
+        alt=shower_geom.alt, az=shower_geom.az
+    )
+
+    # telescope positions in the ground frame
+    telescope_positions = subarray.tel_coords.cartesian.xyz.to_value("m").T
```
maxnoe, line 68: `u.m` to avoid string parsing

ctapipe/reco/impact_distance.py
```diff
@@ -0,0 +1,111 @@
...
+        impact_distance(
+            point=core_position,
+            direction=sky_direction,
+            test_points=telescope_positions,
+        )
+        * u.m
+    )
+
+
+def shower_impact_distance_with_frames(
```
maxnoe, line 80: Is this still needed now that we established that both are the same?

ctapipe/reco/shower_processor.py
```diff
@@ -69,3 +72,18 @@ def __call__(self, event: ArrayEventContainer):
...
+
+        # for the stereo reconstructor:
+        impact_distances = shower_impact_distance(
+            shower_geom=event.dl2.stereo.geometry[k], subarray=self.subarray
+        )
+
+        for tel_index, impact_distance in enumerate(impact_distances):
+            tel_id = self.subarray.tel_ids[tel_index]
+            event.dl2.tel[tel_id].impact[k] = TelescopeImpactParameterContainer(
+                distance=impact_distance
```
maxnoe, line 88: If we don't fill the uncertainty here, why do we define it at all?

ctapipe/reco/impact_distance.py
kosack: probably not, but it is useful for the test.

ctapipe/reco/shower_processor.py
kosack: More for future use and to remind us to implement i - we have a requirement to estimate uncertainty in reconstructed quantities when possible.  This would just be a propagation of the uncertainty in the core reconstruction, which is also not currently implemented yet.

kosack:
> Merging dl2 works fine in the current master, what do you mean with "not very well implemented"?

That was a poor choice of words. DL2 stereo merging does indeed work, but looking (rather quickly) at the code, I would like to do some refactoring to make it easier to implement the merging of tables that are both telescope-dependent and algorithm-dependent.  For example in the list of tables at the top of merge.py, there isn't even yet a category for those.  Since that will require a bit of thought, I think it's ok to make that a separate PR to avoid this becoming too much bigger.  Shouldn't be hard though, and I'll create an issue as a reminder.

kosack:
@jsitarek could you test this out and see if it meets your use case?

ctapipe/conftest.py
```diff
@@ -218,7 +218,7 @@ def dl2_shower_geometry_file_type(dl2_tmp_path, prod5_gamma_simtel_path):
             f"--input={prod5_gamma_simtel_path}",
             f"--output={output}",
             "--write-images",
-            "--write-stereo-shower",
+            "--write-showers",
```
maxnoe, line 221: Not sure I like the `showers` better than `stereo-shower`. Maybe go with `--write-dl2-geometry`?

kosack: It's showers since it now will write both stereo and mono info that is in the DL2 container, which includes also energy and gammaness predictions if they exist (they don't currently of course since we haven't implemented those, but once they are the DataWriter will not need to change)

maxnoe: What I don't like is this "mental mapping" between names and data levels. I think we should just call the options

```
--write-r1
--write-dl0
--write-dl1-images
--write-dl1-parameters
--write-dl2 (switches on geometry and future energy/gh separation, whatever is available)
```

maxnoe: This PR changed the flags in a breaking way already...

maxnoe (dismissed):

maxnoe (approved):

maxnoe (approved):

ctapipe/containers.py
```diff
@@ -724,14 +735,23 @@ class ReconstructedContainer(Container):
     )
 
 
+class TelescopeReconstructedContainer(ReconstructedContainer):
```
maxnoe, line 738: This removes energy, geometry and classification from the telescope reconstructed parameter, these fields will actually be filled for mono predictions in the ml module in #1767, so they should not be removed I think.

maxnoe: Sorry, didn't see the inheritance, all good

maxnoe (approved):

maxnoe (dismissed):

LukasNickel:
This now has changes from the `ml` branch and the tests are failing due to the changes in `ml/conftest.py`.
Are you sure you rebased against master @maxnoe ?

Looks like [a4ffffd](https://github.com/cta-observatory/ctapipe/pull/1899/commits/a4ffffd608a513da59d5c554c278593eb6bfcb9f) added more than expected

maxnoe:
@LukasNickel thanks for noticing, I added uncommited files from the ml branch by accident, fixed now

maxnoe (dismissed):

ctapipe/io/hdf5eventsource.py
```diff
@@ -345,6 +347,15 @@ def _generate_events(self):
                 prefixes="true",
             )
             data.simulation = SimulatedEventContainer()
+            if "impact" in self.file_.root.simulation.event.telescope:
```
LukasNickel, line 350: Tests fail because there is no telescope group in the test file

maxnoe: There should be...

maxnoe:
> Is it because the loop was over dl1 data before and nothing was written if there was none for the event?

yes, it was looping over dl1.tel instead of tels_with_trigger

LukasNickel (approved):

ctapipe/io/hdf5eventsource.py
maxnoe: fixed

LukasNickel (dismissed):

maxnoe (commented):
I have to say I am still not sure we really want to include this information in files vs. just providing the function and computing on the fly...

kosack:
Now there are some warnings generated that were not there before:
```
2022-06-23 16:53:19,711 WARNING [ctapipe.ctapipe-process.DataWriter.HDF5TableWriter] (hdf5tableio._create_hdf5_table_schema): Column impact with value 
ctapipe.containers.TelescopeImpactParameterContainer:
                      distance: distance of the telescope to the shower axis [m]
               distance_uncert: uncertainty in impact_parameter [m] of type 
<class 'ctapipe.containers.TelescopeImpactParameterContainer'>  
of container SimulatedCameraContainer in table simulation/event/telescope/images/tel_001 not writable, skipping
```

kosack (commented):
looks good now

maxnoe (approved):

LukasNickel (approved):

PR #1901, opened 2022-04-27
Add option to load pointing information with TableLoader, fixes #1900

### LINKED ISSUE #1900: TableLoader: Option to load pointing information
It would be useful to have an option for the `TableLoader` to add (interpolated) pointing informations to the table.

Hckjs:
**Description:**
I've added an option for the `TableLoader` to load pointing information from the subarray table, interpolate and adding them to the table.

**Side-Notes:**

- For now it just adds the (interpoalted) `/subarray/pointing` table instead of the `/telescope/pointing` as its done for the trigger information. It doesnt make any difference for simulation data or just one telescope for now (but its faster this way), but has to be adapted in future (Therefore the TODO note).
- I dont know whether its important to get always sorted tables by the keys, but for some parameter combinations, they are not sorted (e.g. if just `load_trigger=True`, because the astropy_helper function `join_allow_empty` just copies the unsorted table if one side is empty. The astropy `join` function sorts by keys). But i think it wouldn't be a problem to just `group_by(keys)` them, if someone needs a sorted one.

Refers to #1900

### DIFF
--- a/ctapipe/io/tableloader.py
+++ b/ctapipe/io/tableloader.py
@@ -111,9 +111,9 @@ class TableLoader(Component):
     load_instrument = traits.Bool(
         False, help="join subarray instrument information to each event"
     ).tag(config=True)
-    load_pointings = traits.Bool(
-        False, help="join subarray pointings information"
-    ).tag(config=True)
+    load_pointings = traits.Bool(False, help="join subarray pointings information").tag(
+        config=True
+    )
 
     def __init__(self, input_url=None, **kwargs):
         # enable using input_url as posarg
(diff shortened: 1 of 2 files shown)


kosack (changes requested):
I'd recommend not putting the interpolation logic in TableLoader, since it has to be used in other places as well and if we do it differently in two places, it will cause a headache in the future.  Better to have a PointingInterpolator component (or even just a function) in the `calib` module.  It should just take an astropy `Table` as input.

Also, in that you should use `scipy.interpolate.interp1d` instead of `np.interp`, since it is more flexible and can pre-compute the interpolation nodes so it is more efficient if one calls it more than once (as we would in the event loop, but maybe not in TableLoader)

kosack:
see #1902

PR #1903, opened 2022-04-27
Keep z coordinate in tilted ground frame, drop unneeded `project_to_ground`

maxnoe:


kosack (changes requested):
TiltedGroundFrame is also used in `ctapipe/reco/impact.py`.  That probably needs to be checked by @ParsonsRD  to make sure this doesn't break anything.

```
rg TiltedGroundFrame

ctapipe/reco/impact.py
18:    TiltedGroundFrame,
579:            X position of telescope in TiltedGroundFrame
581:            Y position of telescope in TiltedGroundFrame
692:            TiltedGroundFrame(pointing_direction=self.array_direction)
731:        tilted = TiltedGroundFrame(
```

ParsonsRD:
Hi, just checked the code. The only issue I see could be that I define a TiltedGroundFrame at some point with only x and y coordinates. Will neglecting the Z coordinate in the definition throw an error?

maxnoe:
@ParsonsRD yes, I added `z=0 * u.m` where needed.

kosack (approved):

ParsonsRD (approved):

maxnoe:
> Just for my understanding, is this so we can drop the requirement for the projection functions and make it behave like the other coordinate transformations?

Not really, projection is different from transforming. This is to ensure that a 3d point in `GroundFrame` actually stays the same point in `TiltedFrame` to enable back and forth between the two frames.

PR #1910, opened 2022-05-18
Use effective focal length by default, always guess telescope names from nominal

### LINKED ISSUE #1908: HillasReconstructor direction reconstructoin is biased for larger fov offsets
Just that I don't forget, I will provide more info later

**Describe the bug**

**To Reproduce**
1.
2.
3. 

**Expected behavior**

**Supporting information**
If applicable, add screenshots, logs, or a traceback to help illustrate the bug.

**Additional context**

I think the reason might be the nominal vs. effective focal length

maxnoe:
For some tests I could switch to prod5 files containing the information, but for most I had to give the option to revert to nominal.

Fixes #1908

### DIFF
--- a/ctapipe/conftest.py
+++ b/ctapipe/conftest.py
@@ -44,7 +44,8 @@ def _global_example_event():
 
     print("******************** LOAD TEST EVENT ***********************")
 
-    with SimTelEventSource(input_url=filename) as reader:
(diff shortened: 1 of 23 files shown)


LukasNickel:
Can you run black on this? There are some very long lines after adding the focal length to the source calls
Looks good otherwise

kosack (approved):

PR #1911, opened 2022-05-19
improve array display

kosack:
* Adds a `EastingNorthingFrame` to properly display array on ground (since the Corsika/Simtel coordinates are confusing with north going on the X-axis).
* Improve ArrayDisplay to have correct frame attribute names on the x and y axes, and add a radial grid
* use an `EastingNorthingFrame` `ArrayDisplay` in `SubarrayDescription.peek()`

Looks like this:

image


ctapipe/coordinates/ground_frames.py
```diff
@@ -200,3 +221,38 @@ def project_to_ground(tilt_system):
...
+
+@frame_transform_graph.transform(FunctionTransform, GroundFrame, EastingNorthingFrame)
+def ground_to_easting_northing(ground_coords, eastnorth_frame):
+    """
+    convert GroundFrame points into eastings/northings for plotting purposes
+
+    """
+
+    return eastnorth_frame.realize_frame(
+        CartesianRepresentation(
```
maxnoe, line 239: This looks like it could easier be an `AffineTransform` with a simple, constant transform matrix

ctapipe/coordinates/tests/test_coordinates.py
```diff
@@ -268,3 +268,15 @@ def test_ground_frame_roundtrip():
     assert u.isclose(coord.x, back.x)
     assert u.isclose(coord.y, back.y)
     assert u.isclose(coord.z, back.z)
+
+
+def test_ground_to_eastnorth_roundtrip():
```
maxnoe, line 273: Not only testing the round trip but also actual values would be good.

ctapipe/coordinates/ground_frames.py
kosack: good point - I didn't actually look to see what transforms are provided by Astropy since it was so simple, but it would be clearer to use one.

ctapipe/coordinates/tests/test_coordinates.py
kosack: done

kosack:
and for completeness, here is what CTA-South alpha looks like by default:
image

kosack:
> Looks good! Can you apply the same changes to the bokeh display?

I probably won't have time this week, so we can either make a separate PR to update the bokeh version (keeping this PR mostly about the new coordinate Frame), or it can just wait.

ctapipe/coordinates/ground_frames.py
```diff
@@ -200,3 +223,35 @@ def project_to_ground(tilt_system):
...
+def ground_to_ground(ground_coords, ground_frame):
+    """Null transform for going from ground to ground, since there are no
+    attributes of the GroundSystem"""
+    return ground_coords
+
+
+# Matrices for transforming between GroundFrame and EastingNorthingFrame
+NO_OFFSET = CartesianRepresentation(Quantity([0, 0, 0], u.m))
+GROUND_TO_EASTNORTH = np.asarray([[0, -1, 0], [1, 0, 0], [0, 0, 1]])
+EASTNORTH_TO_GROUND = np.asarray([[0, 1, 0], [-1, 0, 0], [0, 0, 1]])
```
maxnoe, line 238: This should be just `GROUND_TO_EASTNORTH.T`, right?

kosack: yes, will do

maxnoe (approved):

maxnoe:
Ok, let's merge this now then

maxnoe (approved):

PR #1913, opened 2022-05-23
Start working on flashcam extractor

maxnoe:


maxnoe:
@fwerner Any updates, it would be great to have this in the next release for CTA South Prod5 and Prod6 analysis

maxnoe:
@fwerner sorry for being annoying, but it would be nice to get a status update on this, especially since prod6 only contains flashcams in the southern array.

If you want to have a co-working session to get this done, I would be glad to help.

fwerner:
Hi Max, no worries - entirely my fault for not communicating. My apologies!

As I mentioned during our discussions the critical part will be the verification - and we have a new student, Clara, who just started working with ctapipe on exactly that. (I believe she has contacted you already.)

So this issue should see some activity in the coming weeks!

maxnoe:
Ok, great. Shall we try to make a meeting to get @clara-escanuela up to speed a bit? THe questions she asked were not directly concerned with this PR, so I didn't make the connection.

fwerner:
Wouldn't hurt - let's coordinate via Slack.

ctapipe/image/extractor.py
```diff
@@ -27,6 +27,7 @@
 import numpy as np
 from numba import float32, float64, guvectorize, int64, njit, prange
 from scipy.ndimage import convolve1d
+from scipy.signal import filtfilt
```
maxnoe, line 30: I found this in the scipy docs for this function:

> The function [sosfiltfilt](https://docs.scipy.org/doc/scipy/reference/generated/scipy.signal.sosfiltfilt.html#scipy.signal.sosfiltfilt) (and filter design using output='sos') should be preferred over [filtfilt](https://docs.scipy.org/doc/scipy/reference/generated/scipy.signal.filtfilt.html#scipy.signal.filtfilt) for most filtering tasks, as second-order sections have fewer numerical problems.

Is this applicable here?

ctapipe/image/extractor.py
```diff
@@ -199,7 +200,8 @@ def extract_sliding_window(waveforms, width, sampling_rate_ghz, sum_, peak_time)
     peak_time[0] /= sampling_rate_ghz
 
 
-@njit(cache=True)
+# FIXME JIT has been disabled because it asserts when passing floating-point waveforms (e.g., FlashCamExtractor, at least on arm64)
```
maxnoe, line 203: It asserts what?

ctapipe/image/extractor.py
```diff
@@ -1287,18 +1289,127 @@ def __call__(
             is_valid=is_valid,
         )
 
-
 @lru_cache
-def exponential_scale_from_pulse_shape(camera: CameraDescription):
-    pulse_shape = camera.readout.reference_pulse_shape
-    sample_width = camera.readout.reference_pulse_sample_width
+def deconvolution_parameters(camera: CameraDescription, upsampling: int, window_width: int, window_shift: int):
+    assert(upsampling > 0)
```
maxnoe, line 1294: Please don't use asserts for checks on input parameters, raise the appropriate exceptions with a nice error message, e.g. in this case, use:

```
if upsampling < 0:
    raise ValueError(f"upsampling must be > 0, got {upsampling}")
```

ctapipe/image/extractor.py
```diff
@@ -1287,18 +1289,127 @@ def __call__(
...
+    assert(window_width > 0)
+
+    ref_pulse_shapes = camera.readout.reference_pulse_shape
+    ref_sample_width = camera.readout.reference_pulse_sample_width.to_value("ns")
+    camera_sample_width = 1.0 / camera.readout.sampling_rate.to_value("GHz")
+
+    assert(camera_sample_width >= ref_sample_width)
```
maxnoe, line 1301: That's asserting a property of the camera in something that has no control at all over the `CameraReadout` content, which I think is a bit strange. If that case should never happen, we should test this elsewhere.

If this case can happen in general, but this method needs it not to happen, again as above, raise a nice `ValueError` instead of asserting.

ctapipe/image/extractor.py
```diff
@@ -1287,18 +1289,127 @@ def __call__(
...
+        self.sampling_rate_ghz = {
+            tel_id: telescope.camera.readout.sampling_rate.to_value("GHz")
+            for tel_id, telescope in subarray.tel.items()
+        }
+
+    def __call__(self, waveforms, tel_id, _, broken_pixels) -> DL1CameraContainer:
+        upsampling = max(1, self.upsampling.tel[tel_id])
+        window_width = max(1, self.window_width.tel[tel_id])
+        window_shift = self.window_shift.tel[tel_id]
+        pzs, gains, shifts = deconvolution_parameters(self.subarray.tel[tel_id].camera, upsampling, window_width, window_shift)
```
maxnoe, line 1385: The call to `deconvolution_parameters` looks expensive (didn't test) but does not change since it only depends on options of the Extractor and on the subarray, so this could also be done only once in `__init__`

maxnoe (commented):
Some general comments:

The defensive asserts are nice for developing I guess, but don't really fit. These things should be handled by the unit tests or proper exceptions, which produce much better error message and are better to handle.

I didn't squint to hard into the actual code, but some of the loops look like they could be done using numpy broadcasting instead of the raw loops.

Removing the numba jit decorator from the `neighbor_waveform_sum` is not okay. If this is a problem on ARM64 or whatever the issue was here, please open a general issue so we can fix it independently from here.

ctapipe/image/extractor.py
fwerner: Nothing anymore - was probably due to a shape mismatch in an earlier version.

ctapipe/image/extractor.py
fwerner: Ok.

A PositiveInteger trait would be even nicer for such parameters - does that exist?

ctapipe/image/extractor.py
fwerner: > That's asserting a property of the camera in something that has no control at all over the `CameraReadout` content, which I think is a bit strange.

Are checks of input parameters (=safe programming) not wanted in ctapipe? If so I can remove those checks.

> If that case should never happen, we should test this elsewhere.

Not sure where - as I can't even find the location of the reference waveforms. Could someone more knowledgable do this?

> If this case can happen in general, but this method needs it not to happen, again as above, raise a nice ValueError instead of asserting.

Will do.

ctapipe/image/extractor.py
fwerner: I was thinking about that - but wasn't sure if the parameters (upsampling, window_width, window_size) are *fixed* after init. Is that the case?

ctapipe/image/extractor.py
maxnoe: It's not documented, but the `traitlets.Int` actually takes a `min` and `max` option:

https://github.com/ipython/traitlets/blob/3d35d531f0386585f47b8af6b3e15343e4bdd48a/traitlets/traitlets.py#L2242-L2251

ctapipe/image/extractor.py
maxnoe: > Not sure where - as I can't even find the location of the reference waveforms. Could someone more knowledgable do this?

These are (at least for simtel files) directly stored in the camera configuration part of the files.

> Are checks of input parameters (=safe programming) not wanted in ctapipe? If so I can remove those checks.

They are wanted in many cases, but then they shouldn't be `assert` statements but the corresponding exception, e.g. `TypeError` or `ValueError` with nice error messages.

fwerner:
> Some general comments:
> 
> The defensive asserts are nice for developing I guess, but don't really fit. These things should be handled by the unit tests or proper exceptions, which produce much better error message and are better to handle.

Thanks for that comment. I tried to change that in the last push - please let me know if that's more suitable.

> I didn't squint to hard into the actual code, but some of the loops look like they could be done using numpy broadcasting instead of the raw loops.

You're probably right; deconvolution_parameters is supposed to be called once per camera type, so I did not give it a hard look yet.

> Removing the numba jit decorator from the `neighbor_waveform_sum` is not okay. If this is a problem on ARM64 or whatever the issue was here, please open a general issue so we can fix it independently from here.

Fixed.

ctapipe/image/extractor.py
maxnoe: Yes, in most cases we assume that configuration cannot change after `__init__`. 

The only way that can happen is in interactive settings like jupyter notebooks, but then you should just run `__init__` again.

fwerner: OK if such behaviour is acceptable - done. 👍

ctapipe/image/extractor.py
fwerner: Thanks; done.

ctapipe/image/extractor.py
fwerner: > > Not sure where - as I can't even find the location of the reference waveforms. Could someone more knowledgable do this?
> 
> These are (at least for simtel files) directly stored in the camera configuration part of the files.

I'm still not sure where to put such a check - my comprehension of this codebase is lacking, sorry. :/

maxnoe: If I understand correctly, you check that the single pe pulse is sampled more finely than the sampling frequency of the actual data, correct?

fwerner: At least as fine as data; yes.

maxnoe: Is that a general requirement or specific to this method?

maxnoe: This in any case needs a very clear error message, because at least in the case of sim_telarray input files, the user's only way to fix this is to resimulate with a finer binned single pe pulse, it's completely outside the control of ctapipe.

ctapipe/image/extractor.py
```diff
@@ -1285,3 +1288,158 @@ def __call__(
...
+        window_width = self.window_width.tel[tel_id]
+        window_shift = self.window_shift.tel[tel_id]
+
+        pzs, gains, shifts = self.deconvolution_pars[tel_id]
+        pz, gain, shift = pzs[0], gains[0], shifts[0]
+
+        bls = waveforms[:, 0]  # FIXME fetch from NSB estimates
+        waveforms = deconvolve(waveforms, bls, upsampling, pz)
+
+        # FIXME near-duplicate of neighbour peak sum for now
```
maxnoe, line 1421: What is different?

ctapipe/image/extractor.py
fwerner: I'd say it's a general requirement - doesn't make sense to use a too coarse impulse response for simulations.

I did this - ok?
```
    if camera_sample_width < ref_sample_width:
        raise ValueError(
            f"ref_sample_width (got {ref_sample_width}) must be equal to or shorter than camera_sample_width (got {camera_sample_width}); need a reference single p.e. pulse shape with finer sampling!"
        )
```

ctapipe/image/extractor.py
fwerner: Not applicable to our simple filter, no. Thanks for checking though.

ctapipe/image/extractor.py
maxnoe: yes, looks good, thanks

ctapipe/resources/base_config.yaml
```diff
@@ -38,7 +38,12 @@ CameraCalibrator:
   #
   # Note this is a telescope-wise parameter, so can be specified per telescope
   # if necessary (see below for an example)
-  image_extractor_type: NeighborPeakWindowSum
+  image_extractor_type:
+    - ['type', '*', 'NeighborPeakWindowSum']
+    - ['type', '*FlashCam', 'FlashCamExtractor']
+
+  FlashCamExtractor:
+    exponential_scale: 7.5
```
kosack, line 46: This option doesn't seem to exist in the Extractor - was it forgotten?

fwerner: I guess Max added this as a placeholder in our first discussion - not needed anymore.

ctapipe/image/extractor.py
```diff
@@ -1285,3 +1288,158 @@ def __call__(
...
+            raise ValueError(
+                "ref_pulse_shape is malformed - peak is not well contained"
+            )
+        gains.append(np.mean(phase_gains))
+        shifts.append(np.mean(phase_shifts))
+
+    return pzs, gains, shifts
+
+
+def deconvolve(waveforms, bls, up: int, pz: float):
```
kosack, line 1354: please add a docstring and avoid abbreviations (use clear variable names for  "up", "pz" and "bls")

ctapipe/image/extractor.py
```diff
@@ -1285,3 +1288,158 @@ def __call__(
             peak_time=pulse_time2.astype("float32"),
             is_valid=is_valid,
         )
+
+
+@lru_cache
+def deconvolution_parameters(
+    camera: CameraDescription, upsampling: int, window_width: int, window_shift: int
+):
```
kosack, line 1296: add docstring

kosack (changes requested):

ctapipe/image/extractor.py
fwerner: Got around to making a first attempt - is that useful enough?

ctapipe/image/extractor.py
```diff
@@ -1352,6 +1362,13 @@ def deconvolution_parameters(
 
     gains, shifts = [], []  # avg. gains and timing shifts of the deconvolved pulses
     for pz, ref_pulse_shape in zip(pzs, ref_pulse_shapes):
+        if time_profile_pdf:  # convolve ref_pulse_shape with time profile PDF
+            t = (
+                np.arange(ref_pulse_shape.size) - ref_pulse_shape.size / 2
+            ) * ref_sample_width_nsec
```
maxnoe, line 1368: [The pulse shape time is available as `subarray.tel[tel_id].camera.readout.ref](https://github.com/cta-observatory/ctapipe/blob/370b6b9a8464de0c3a5a06958ccc8b537d4b4f51/ctapipe/instrument/camera/readout.py#L118)

maxnoe:
@fwerner The fact that I originally opened this Pull Request excludes me as a reviewer.

I will now close this PR and would ask you to open a new one yourself, so that the author of the PR is set properly and I am available as a reviewer.

It seems all comments here were alrady addresses, so that shouldn't  be a problem.

maxnoe:
@fwerner @clara-escanuela 

Could you please reopen this PR? I would be very good to get at least a preliminary version of this in. We shouldn't wait for a full verification of this code on a large simulation dataset / actual flashcam data.

If you think the general algorithm is correct and there are unit tests on a simple data example, we should merge and then the full verification can happen on the main branch or the next released version, iterating from there.

maxnoe:
Sorry, re-open is not enough. You actually need to open a new PR from the same branch

PR #1918, opened 2022-05-24
Refactoring of the simulation pointing

maxnoe:
This should improve the handling of the pointing information and prepare for the proper handling of real observation monitoring data. 

Goal is to remove the "closest timestamp" matching in the hdf5 event source and replace it with filling values from the simulation header for simulations and actual iterpolation from the monitoring table for observations

### DIFF
--- a/ctapipe/containers.py
+++ b/ctapipe/containers.py
@@ -597,10 +597,10 @@ class SimulationConfigContainer(Container):
         np.nan * u.rad,
         "Array Pointing declination, only filled of constant for the run",
     )
-    array_alt = Field(
+    array_altitude = Field(
         np.nan * u.rad, "Array Pointing altitude, only filled of constant for the run"
     )
-    array_az = Field(
+    array_azimuth = Field(
         np.nan * u.rad, "Array Pointing azimuth, only filled of constant for the run"
     )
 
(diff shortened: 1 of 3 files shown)


ctapipe/containers.py
```diff
@@ -572,6 +586,23 @@ class SimulationConfigContainer(Container):
     corsika_high_E_detail = Field(
         nan, "More details on high E interaction model (version etc.)"
     )
+    tracking_mode = Field(
+        TrackingMode.UNKNOWN, "Tracking Mode (see docs of `TrackingMode`)"
+    )
```
kosack, line 591: Might be better to have this all in a SchedulingBlock container, which can be common to both Simulations and Observations.

maxnoe: This is directly from simtel... It's more part of the simtel data model than the CTA one.

ctapipe/containers.py
```diff
@@ -91,8 +91,22 @@ class EventType(enum.Enum):
     UNKNOWN = 255
 
 
+class TrackingMode(enum.Enum):
+    """
+    TrackingMode of the telescopes.
+
+    Defines which coordinates are supposed to be constant during the run.
+    """
+
```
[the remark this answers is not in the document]
maxnoe: One of the comments I was about to make...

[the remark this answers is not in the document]
kosack: Yes, I agree that looking at the current model it mixes two concepts (pointing coordinate system and divergence/convergence).  

But the main thing is that we should agree to implement a "SchedulingBlockContainer" (and any sub-containers) here, and it's likely we will have to at the same time update the official model. We can perhaps use this as a testbed to propose a better organization.     Also I do not like "vergent" as it's not only not a word, but doesn't tell you the sign of the parameter (e.g. if it was called Divergent and the depth is negative, it's clear it is converging)

ctapipe/containers.py
kosack: By why hard-code things to the way SimTel does it? My point is we should have a common definition of pointing information, and then use it for SimTel, some other simulation, or real Observations.

maxnoe:
Closed in favor of #1973

PR #1923, opened 2022-06-02
Store both equivalent and effective focal length in OpticsDescription, add __slots__

maxnoe:
This is adding a new field `effective_focal_length` to the optics description, so we don't need to choose there but when we actually do the transforms.

Like this, it's clear inside ctapipe what focal length is used when and it doesn't depend on options what is inside the OpticsDescription.

The choice options now say which of the focal lengths is attached to the `CameraGeometry.frame` attribute when creating the subarray, so all things loading a subarray (SimTelEventSource, HDF5EventSource, Subarray.from_hdf, TableLoader) now gain this option.

### DIFF
--- a/ctapipe/instrument/optics.py
+++ b/ctapipe/instrument/optics.py
@@ -60,8 +60,8 @@ class OpticsDescription:
         num_mirrors,
         equivalent_focal_length,
         effective_focal_length,
-        mirror_area=None,
-        num_mirror_tiles=None,
+        mirror_area,
+        num_mirror_tiles,
     ):
 
         self.name = name
@@ -84,7 +84,7 @@ class OpticsDescription:
             (
                 round(self.equivalent_focal_length.to_value(u.m), 4),
                 round(effective_focal_length, 4),
-                self.mirror_area,
+                round(self.mirror_area.to_value(u.m**2)),
                 self.num_mirrors,
                 self.num_mirror_tiles,
             )
--- a/ctapipe/instrument/optics.py
+++ b/ctapipe/instrument/optics.py
@@ -41,35 +41,50 @@ class OpticsDescription:
     """
 
     __slots__ = (
+        "name",
+        "effective_focal_length",
         "equivalent_focal_length",
         "mirror_area",
-        "name",
-        "num_mirror_tiles",
         "num_mirrors",
+        "num_mirror_tiles",
     )
 
-    @u.quantity_input(mirror_area=u.m**2, equivalent_focal_length=u.m)
+    @u.quantity_input(
+        mirror_area=u.m**2,
+        equivalent_focal_length=u.m,
+        effective_focal_length=u.m,
+    )
     def __init__(
         self,
         name,
         num_mirrors,
         equivalent_focal_length,
+        effective_focal_length,
         mirror_area=None,
         num_mirror_tiles=None,
     ):
 
         self.name = name
         self.equivalent_focal_length = equivalent_focal_length.to(u.m)
+        self.effective_focal_length = effective_focal_length.to(u.m)
         self.mirror_area = mirror_area
         self.num_mirrors = num_mirrors
         self.num_mirror_tiles = num_mirror_tiles
 
     def __hash__(self):
         """Make this hashable, so it can be used as dict keys or in sets"""
+        # From python >= 3.10, hash of nan is random, we want a fixed hash also for
+        # unknown effective focal length:
+        if np.isnan(self.effective_focal_length.value):
+            effective_focal_length = -1
+        else:
+            effective_focal_length = self.effective_focal_length.to_value(u.m)
+
         return hash(
             (
-                round(self.equivalent_focal_length.to_value(u.m), 3),
-                round(self.mirror_area.to_value(u.m**2), 3),
+                round(self.equivalent_focal_length.to_value(u.m), 4),
+                round(effective_focal_length, 4),
+                self.mirror_area,
                 self.num_mirrors,
                 self.num_mirror_tiles,
             )
@@ -121,7 +136,7 @@ class OpticsDescription:
 
         if version == "1.0":
             mask = table["tel_description"] == name
-        elif version == "2.0":
+        else:
             mask = table["description"] == name
 
         if np.count_nonzero(mask) == 0:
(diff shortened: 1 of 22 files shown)


nbiederbeck (dismissed):

nbiederbeck (dismissed):

maxnoe:
Let's merge this PR and the other one affecting the data model: #1949 and then I can recreate the higher level test data files with the new data model.

ctapipe/instrument/optics.py
```diff
@@ -40,27 +40,49 @@ class OpticsDescription:
...
-    @u.quantity_input(mirror_area=u.m ** 2, equivalent_focal_length=u.m)
+    __slots__ = (
+        "name",
+        "equivalent_focal_length",
+        "effective_focal_length",
+        "mirror_area",
+        "num_mirrors",
+        "num_mirror_tiles",
+    )
+
```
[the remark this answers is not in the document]
maxnoe: Yes.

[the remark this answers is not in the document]
maxnoe: I will make a new PR adding slots for the other ones and also the float treatment that was only introduced for CameraGeometry in #1956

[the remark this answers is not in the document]
maxnoe: But not only, using `__slots__` has several benefits, it also makes attribute access slightly faster and memory consumption lower (but likely no real impact on the total of ctapipe).

nbiederbeck (approved):

ctapipe/io/simteleventsource.py
```diff
@@ -339,19 +349,11 @@ def prepare_subarray_info(self, telescope_descriptions, header):
...
 
-            optics = OpticsDescription(
-                name=telescope.name,
-                num_mirrors=telescope.n_mirrors,
-                equivalent_focal_length=focal_length,
-                mirror_area=mirror_area,
-                num_mirror_tiles=cam_settings["n_mirrors"],
-            )
-
             camera = build_camera(
```
kosack, line 352: One (perhaps less common) use case that might not work here, and also didn't before, is if the user changes the focal length choice after construction.  In that case, the SubarrayDescription should be re-built. :
```py3
with EventSource(filename) as source:
    source.focal_length_choice='nominal'
    source.subarray.tel[1].camera.geometry.frame # still uses effective 
```
We could add a callback to that trait, and either just raise an exception if they try to change it, or else re-generate the CameraFrames.

kosack (changes requested):

ctapipe/io/simteleventsource.py
kosack: added #1997

maxnoe: I think at least `allowed_tels` is also not changeable after construction.

kosack (dismissed):

ctapipe/instrument/optics.py
```diff
@@ -24,10 +24,19 @@ class OpticsDescription:
...
-        optics (as in the Monte-Carlo)
-    mirror_area: float
+    equivalent_focal_length : astropy.units.Quantity[length]
+        Equivalent focal-length of telescope, independent of which type of
+        optics (as in the Monte-Carlo). This is the nominal focal length
+        for single mirror telescopes and the equivalent focal length for dual
+        mirror telescopes.
+    effective_focal_length : astropy.units.Quantity[length]
+        The effective_focal_length is the focal length estimated from
+        ray tracing to correct for koma aberration. It is thus not automatically
```
vuillaut, line 34: I think it's _coma_ in english

ctapipe/instrument/optics.py
```diff
@@ -24,10 +24,19 @@ class OpticsDescription:
...
+        available for all simulations, but only if it was set before hand
```
vuillaut, line 35: _beforehand_

ctapipe/instrument/subarray.py
```diff
@@ -538,7 +547,7 @@ def to_hdf(self, h5file, overwrite=False, mode="a"):
                 meta["reference_itrs_z"] = itrs.z.to_value(u.m)
 
     @classmethod
-    def from_hdf(cls, path):
+    def from_hdf(cls, path, focal_length_choice="effective"):
```
vuillaut, line 550: do we expect other possibilities than equivalent and effective in the future?
If not, a boolean `effective=True` might be less error prone than typing a string?

vuillaut (commented):
nominal and equivalent are the same thing right?
Maybe use a single consistent naming everywhere?

ctapipe/instrument/subarray.py
maxnoe: Good point,  I will use an enum instead

ctapipe/instrument/optics.py
maxnoe: fixed

ctapipe/instrument/optics.py
maxnoe: fixed

maxnoe:
> nominal and equivalent are the same thing right?

Only for single mirror telescopes, this is why the docstring says: "This is the nominal focal length for **single mirror telescopes** and the equivalent focal length for **dual mirror telescopes**". For dual mirror telescopes, there are two nominal focal lengths and the *equivalent* focal length is the thin lens focal length that would give the same image.

ctapipe/instrument/subarray.py
kosack: it's possible that in the future this is a calibration parameter that is measured in-situ as well. I know in HESS, for example, we store both the nominal and measured and have to correct for the difference, and for the HESS CT5 telescope, there is even an autofocus system that can dynamically change the focal length. So I agree that an ENUM is best here, in case we add other ways to obtain this information.

vuillaut:
> > nominal and equivalent are the same thing right?
> 
> Only for single mirror telescopes, this is why the docstring says: "This is the nominal focal length for **single mirror telescopes** and the equivalent focal length for **dual mirror telescopes**". For dual mirror telescopes, there are two nominal focal lengths and the _equivalent_ focal length is the thin lens focal length that would give the same image.

Sorry I read too quickly. Thank you for the explanation.

maxnoe:
@vuillaut @kosack please review again, the num is now implemented.

maxnoe:
Sorry @kosack, had to fix the docs build

vuillaut (approved):

kosack (approved):

PR #1929, opened 2022-06-10
Support reading files created with `extract_calibevent`

maxnoe:
Test file to be added.

### DIFF
--- a/ctapipe/io/simteleventsource.py
+++ b/ctapipe/io/simteleventsource.py
@@ -398,6 +398,8 @@ class SimTelEventSource(EventSource):
         data.meta["origin"] = "hessio"
         data.meta["input_url"] = self.input_url
         data.meta["max_events"] = self.max_events
+        # for events without event_id, we use negative event_ids
+        pseudo_event_id = 0
 
         self._fill_array_pointing(data)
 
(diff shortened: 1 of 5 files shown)


maxnoe:
Unfortunately I introduced a breaking change into eventio 1.9.0, will be fixed soon with a 1.9.1 release

ctapipe/io/simteleventsource.py
```diff
@@ -413,7 +415,11 @@ def _generate_events(self):
             data.simulation.tel.clear()
             data.trigger.tel.clear()
 
-            event_id = array_event.get("event_id", -1)
+            event_id = array_event.get("event_id", 0)
+            if event_id == 0:
```
kosack, line 419: Is there no chance that there is an `event_id=0` that is a real event?  Maybe better to return None as default?

maxnoe: event_id 0 seems exclusively used for calibration events. All real shower events in simtel array are 100 x shower + reuse, so they start at 100. The extract_calibevents start at 1.

kosack (approved):

LukasNickel (approved):

nbiederbeck (approved):

PR #1934, opened 2022-06-14
Load all dl2 available groups in TableLoader

maxnoe:
This replaces the hard coded `geometry` group with loading all groups that are in the `/dl2/event/subarray/<attribute>/<algorithm>` and `/dl2/event/telescope/<attribute>/<algorithm>/tel_<tel_id:03d>` structure.

PR #1935, opened 2022-06-14
Chunked table loader

maxnoe:
Needs #1934

### DIFF
--- a/ctapipe/conftest.py
+++ b/ctapipe/conftest.py
@@ -133,9 +133,7 @@ def subarray_and_event_gamma_off_axis_500_gev(
 
 @pytest.fixture(scope="session")
 def prod5_gamma_simtel_path():
-    return get_dataset_path(
-        "gamma_20deg_0deg_run2___cta-prod5-paranal_desert-2147m-Paranal-dark_cone10-100evts.simtel.zst"
-    )
+    return get_dataset_path("gamma_prod5.simtel.zst")
 
 
 @pytest.fixture(scope="session")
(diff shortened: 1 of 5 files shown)


kosack:
It might be a nicer API to have the division into chunks happen inside, rather than by the user, e.g. rather than have a start/stop, allow the user to specify a `chunk_size` (or a `n_chunks`):

```py3
loader = TableLoader(..., chunk_size=1000)

for chunk in  table_loader.read_subarray_events():
    do_stuff(chunk)
```
or perhaps better, put the chunking options in `read_subarray_events()` and `read_telescope_events()` themselves:

```py3
for chunk in  loader.read_subarray_events(chunk_size=1000):
    do_stuff(chunk)
```

in that case, read_subarray_events returns a *generator* if chunk_size is non-zero, and you can even do things like:

```py3
map( do_stuff, loader.read_subarray_events(chunk_size=1000))
```

maxnoe:
> It might be a nicer API to have the division into chunks happen inside, rather than by the user, e.g. rather than have a start/stop, allow the user to specify a chunk_size (or a n_chunks):

Yes, this is planned (the reason this is still a draft). I only managed the basic building blocks yesterday in an "how many off by one errors can you do in two hours" session

LukasNickel (approved):

kosack (approved):
Very nice!  I did a quick check to make sure the memory is properly handled, and it looks great:

image

kosack:
perhaps also mention the chunked versions in the docstring, or in the general docs.  

Also, a nice example might be making a histogram over all chunks:

```py3
from ctapipe.utils import Histogram

e_dist = Histogram(nbins=[100], ranges=[[-2,2]], axis_names=['$\log10 E_{true}$',])

for chunk in tqdm(loader.read_subarray_events_chunked(50_000)):
    e_dist.fill(np.log10(chunk['true_energy']))

e_dist.draw_1d()
```
image

kosack (approved):

PR #1936, opened 2022-06-15
Add test for TrainEnergyRegressor in #1767

nbiederbeck:
To be merged into `ml` (#1767)

### DIFF
--- a/ctapipe/ml/tools/tests/test_train.py
+++ b/ctapipe/ml/tools/tests/test_train.py
@@ -1,6 +1,5 @@
 from ctapipe.core import run_tool
 from ctapipe.ml import TrainEnergyRegressor
-from ctapipe.utils import get_dataset_path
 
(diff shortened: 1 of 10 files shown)


ctapipe/ml/tools/tests/test_train.py
```diff
@@ -0,0 +1,28 @@
+from ctapipe.core import run_tool
+from ctapipe.ml import TrainEnergyRegressor
+
+try:
+    from importlib.resources import files
+except ImportError:
+    from importlib_resources import files
+
+
+def resource_file(filename):
```
LukasNickel, line 10: This is the same as in the test for process.
Maybe we can collect these in `conftest.py`?

maxnoe: Utils, it's not just for tests

PR #1938, opened 2022-06-17
Fix loading of monoscopic dl2 data in the tableloader

LukasNickel:
There is a small bug in the tableloader, which prevents it from reading monoscopic dl2 data.
Since we do not produce any so far, the tests did not spot it.

For now, I added a dummy table for the test files. We should be able to remove it soon (once the ML module is ready).
(tel 25 and FlashCam, because thats what the tableloader unit tests looks for).

### DIFF
--- a/ctapipe/conftest.py
+++ b/ctapipe/conftest.py
@@ -2,6 +2,8 @@
 common pytest fixtures for tests in ctapipe
 """
 
+from astropy.table import Table
+import astropy.units as u
 from copy import deepcopy
 
 import pytest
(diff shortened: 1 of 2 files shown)


maxnoe (approved):

LukasNickel:
Do we still want his?
I think the fix found its way into the master already via f549dfa6ac183a73787e7d742ded173bfc619bef.
It is still untested, but the ML branch will provide test cases naturally

maxnoe:
Yes, already fixed, closing

PR #1941, opened 2022-06-20
Combining telescope-level dl2 predictions

LukasNickel:
Working on ways to combine the individual predictions to array-wide predictions.
Requires #1938 and could be improved with #1888 and #1935 merged.

This adds components, that get a table of monoscopic predictions, finds groups of unique `(obs_id, event_id)` pairs and
performs some operations on them. Currently this supports the mean and median, but everything available as `numpy.ufunc` is straigthforward to add.
The same is true for measurement functions in `scipy.ndimage`, which is probably even easier, but I already had the mean implemented using `np.add`.

My biggest issue right now is the handling of nans or, more generally, selecting rows to use for the calculation.
I did not find a solution I am really happy with, but it probably comes down to it being a usecase for quality queries on tables.
Maybe one could use the normal query in a loop to hide the issue from the user. I did not time it, but given that we spend most time in lower datalevels anyway, it might not be too bad...


ctapipe/ml/stereo_combination.py
```diff
@@ -0,0 +1,141 @@
...
+import numpy as np
+from astropy.table import Table
+import astropy.units as u
+from scipy.ndimage import median
+from ctapipe.core import Component
+from ctapipe.ml.preprocessing import check_valid_rows
+
+
+def _calculate_ufunc_of_telescope_values(tel_data, n_array_events, indices, ufunc):
+    mean_values = np.zeros(n_array_events)
```
maxnoe, line 11: this is called "mean" although you can use any ufunc...

ctapipe/ml/stereo_combination.py
```diff
@@ -0,0 +1,141 @@
...
+
+
+class StereoCombiner(Component):
+    # TODO: Add quality query (after #1888)
+    def __init__(self, mono_prediction_column, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+        self.mono_prediction_column = mono_prediction_column
+
+    @abstractmethod
+    def __call__(self, mono_predictions: Table) -> np.ndarray:
```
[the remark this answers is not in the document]
LukasNickel: Ok!

ctapipe/ml/stereo_combination.py
```diff
@@ -0,0 +1,141 @@
...
+                np.add,
+            )
+            sum_of_weights = n_tels_per_event
+        stereo_table[f"mean_{self.mono_prediction_column}"] = u.Quantity(
+            sum_prediction / sum_of_weights, prediction_unit, copy=False
+        )
+        return stereo_table
+
+
+class StereoMedianCombiner(StereoCombiner):
```
maxnoe, line 105: I don't think the median is really helpful for the low number of events we have and doing it weighted is also hard with low statistics. So for the first iteration, I'd probably remove it and just put in mean. The mean aggregator should probably also calculate std.

ctapipe/ml/stereo_combination.py
```diff
@@ -0,0 +1,141 @@
...
+    mean_values = np.zeros(n_array_events)
+    ufunc.at(mean_values, indices, tel_data)
+    return mean_values
+
+
+class StereoCombiner(Component):
+    # TODO: Add quality query (after #1888)
+    def __init__(self, mono_prediction_column, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+        self.mono_prediction_column = mono_prediction_column
```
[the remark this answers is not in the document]
LukasNickel: In that case, you should be able to infer the attribute from the container type, right?
With two separate calls, I would then add this as argument of the function, that takes tables. Or maybe obs_id, event_id, prediction individually

ctapipe/ml/stereo_combination.py
```diff
@@ -0,0 +1,141 @@
...
+
+class StereoMeanCombiner(StereoCombiner):
+    """
+    Calculates array-wide (stereo) predictions as the mean of
+    the reconstruction on telescope-level with an optional weighting.
+    """
+
+    def __init__(self, mono_prediction_column, weight_column=None, *args, **kwargs):
+        super().__init__(mono_prediction_column=mono_prediction_column, *args, **kwargs)
+        self.weight_column = weight_column
```
maxnoe, line 49: We won't have a weight column... I think we should implement an enum of weights, e.g. like this:

```
weight = Enum(["intensity", "konrad"])
```

where "intensity" just means weighting by intensity and "konrad" is the definition also used in `HillasReconstructor`. The weight should be calculated here.

ctapipe/ml/stereo_combination.py
```diff
@@ -0,0 +1,141 @@
...
+
+    def _construct_weights(self, table):
+        # TODO: Do we want to calculate weights on the fly or do it externally?
```
maxnoe, line 52: I'd say yes

ctapipe/ml/stereo_combination.py
LukasNickel: Weights make no sense for a median, agreed. Thats why its only in the MeanCombiner.

Std I was just about to add. That is also weird for the median, so yeah I will remove it for now

ctapipe/ml/stereo_combination.py
```diff
@@ -0,0 +1,141 @@
...
+        return stereo_table
+
+
+class StereoMedianCombiner(StereoCombiner):
+    """
+    Calculates array-wide (stereo) predictions as the median of
+    the reconstruction on telescope-level.
+    """
+
+    def __call__(self, mono_predictions: Table) -> Table:
```
maxnoe, line 111: as said above, `__call__` should work on the `ArrayEvent`. The other classes have a `predict(table)` method. This should be the same.

maxnoe (changes requested):
General points:

* Remove median
* Add std computation
* Make sure output column names are matching container structure 
* `__call__(event: ArrayEvent)` and `predict(table: Table)`.

kosack:
This is intended to operate on an existing DL2 file I guess.  It would be nice to also support event-wise prediction, as then we can include this in the list of Reconstructors that are run during `ctapipe-process` in the `ShowerProcessor`.  

Remember that the final goal is to have DL2 files with many reconstructions in them, so that in the DL2→DL3 step we can select which reconstruction to used on an event-by-event basis .

maxnoe:
@kosack yes, this is the idea. Have `__call__(event:ArrayEvent)` for process and `predict(table: Table)` for the `ctapipe-ml-apply` tool

ctapipe/ml/stereo_combination.py
```diff
@@ -44,19 +47,102 @@ class StereoMeanCombiner(StereoCombiner):
...
 
-    def __call__(self, mono_predictions: Table) -> Table:
+    def _calculate_weights(self, dl1):
+        """ """
+        if isinstance(dl1, Container):
+            if self.weights == "intensity":
+                return dl1.hillas["intensity"]
+            elif self.weights == "konrad":
+                return (
+                    dl1.hillas["intensity"] * dl1.hillas["length"] / dl1.hillas["width"]
```
maxnoe, line 67: Using `.hillas` should be *slightly* more performant, since `Container.__getitem__` is implemented as `getattr(self, key)` and the container uses real slots for storage.

ctapipe/ml/stereo_combination.py
```diff
@@ -44,19 +47,102 @@ class StereoMeanCombiner(StereoCombiner):
...
+        elif isinstance(dl1, Table):
+            if self.weights == "intensity":
+                return dl1["hillas_intensity"]
+            elif self.weights == "konrad":
+                return (
+                    dl1["hillas_intensity"] * dl1["hillas_length"] / dl1["hillas_width"]
+                )
+            else:
+                return np.ones(len(dl1))
+        return 1
```
maxnoe, line 80: Shouldn't this be some kind of error instead of returning 1?

ctapipe/ml/stereo_combination.py
```diff
@@ -44,19 +47,102 @@ class StereoMeanCombiner(StereoCombiner):
...
+
+    def _weighted_mean_ufunc(self, tel_values, weights, n_array_events, indices):
```
maxnoe, line 82: Why is this a method and the other one a free function? This could at least be a `staticmethod` since it's not acutally using `self`.

ctapipe/ml/stereo_combination.py
LukasNickel: Probably. I think its a leftover

ctapipe/ml/stereo_combination.py
maxnoe: yes

maxnoe:
@LukasNickel Since you set `algorithm` and `type` in this, we would need a number of these (at least two, to combine energy and gammaness predictions). Is it possible to do something like this?

```
ProcessorTool:
    stereo_combiners:
        - algorithm: RandomForestRegressor
          kind: energy
        - algorithm: RandomForestClassifier
          kind: particle_classification
```

?

maxnoe:
This seems possible using for example:

```
from traitlets import List, Unicode, Dict
from ctapipe.core import Component, Tool
import logging


logging.basicConfig(level=logging.INFO)


class Combiner(Component):
    kind = Unicode().tag(config=True)
    name = Unicode().tag(config=True)

    def __repr__(self):
        return f'{self.__class__.__name__}({self.kind}, {self.name})'


class MyTool(Tool):
    '''Tool'''
    name = 'mytool'
    description = 'stuff'
    combiner_configs = List(Dict()).tag(config=True)


    def setup(self):
        self.combiners = [
            Combiner(**config, parent=self)
            for config in self.combiner_configs
        ]

    def start(self):
        for combiner in self.combiners:
            combiner.log.info('kind=%s name=%s', combiner.kind, combiner.name)



if __name__ == '__main__':
    tool = MyTool()
    tool.run()

```

ctapipe/ml/tools/apply_energy_regressor.py
```diff
@@ -83,6 +93,31 @@ def start(self):
                 mode="a",
                 overwrite=self.overwrite,
             )
+        self.loader.load_instrument = False
+        # TODO: Might need to turn this on if we want to use dl1 features as weights
+        self.loader.load_dl1_parameters = False
```
maxnoe, line 98: This already happens with `weight=konrad`, doesn't it? Add a test for `weight="konrad"`?

maxnoe (approved):

PR #1943, opened 2022-06-22
Add QualityQuery.get_table_mask, compile python code

maxnoe:
In preparation of implemting the quality query for tables, I tried to come up with a way these quality queries are pure expressions, that would be suitable for numexpr also in the "apply to single element case".

I do this here by evaluating the expressions with the kwargs of call as locals. The expressions are compiled to bytecode in `__init__` and then evaled, which should not make a difference in performance.

I haven't measured this yet though.

### DIFF
--- a/ctapipe/core/provenance.py
+++ b/ctapipe/core/provenance.py
@@ -127,7 +127,7 @@ class Provenance(metaclass=Singleton):
         Parameters
         ----------
         config: dict
-            configuration paramters
+            configuration parameters
         """
         self.current_activity.register_config(config)
 
--- a/ctapipe/core/qualityquery.py
+++ b/ctapipe/core/qualityquery.py
@@ -6,7 +6,6 @@ __all__ = ["QualityQuery", "QualityCriteriaError"]
 
 import astropy.units as u  # for use in selection functions
 import numpy as np  # for use in selection functions
-import numexpr
 
 from .component import Component
 from .traits import List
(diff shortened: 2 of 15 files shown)


maxnoe:
```python

In [1]: import numpy as np

In [2]: from astropy.table import Table

In [3]: rng = np.random.default_rng(0)

In [4]: N = rng.poisson(100, size=10000)

In [5]: x = rng.uniform(-5, 5, size=10000)

In [6]: y = rng.uniform(-5, 5, size=10000)

In [7]: 

In [7]: t = Table({'x': x, 'y': y, 'N': N})

In [8]: from ctapipe.core import QualityQuery

In [9]: q = QualityQuery(quality_criteria=[('foo', 'x > 2'), ('bar', 'y < 0'), ('baz'', 'sqrt(x**2 + y**2) < 1')])


In [11]: q.get_table_mask(t)
Out[11]: array([False, False, False, ..., False, False, False])

In [12]: %timeit q.get_table_mask(t)
439 µs ± 1.4 µs per loop (mean ± std. dev. of 7 runs, 1,000 loops each)

In [13]: %timeit q.get_table_mask(t, engine='numexpr')
439 µs ± 1.07 µs per loop (mean ± std. dev. of 7 runs, 1,000 loops each)

In [14]: %timeit q.get_table_mask(t, engine='python')
342 µs ± 1.46 µs per loop (mean ± std. dev. of 7 runs, 1,000 loops each)

In [15]: %timeit q.get_table_mask(t, engine='python')
342 µs ± 1.36 µs per loop (mean ± std. dev. of 7 runs, 1,000 loops each)

In [16]: %timeit q.get_table_mask(t, engine='numexpr')
441 µs ± 349 ns per loop (mean ± std. dev. of 7 runs, 1,000 loops each)
```

maxnoe:
I couldn't really come up with queries where `numexpr` is actually faster than the `python` mode. So we should probably remove the `numexpr` based code.

Maybe play around some and try to see if you can find queries where numexpr is faster.

nbiederbeck (dismissed):

maxnoe:
Removed numexpr, since I could find anything where it was faster

nbiederbeck (approved):

PR #1949, opened 2022-06-24
Rework container prefix handling

maxnoe:
This PR changes our prefix handling to make it consistent over the different IO methods. 

Instead of attaching the algorithm name in the TableLoader to avoid duplicated column names, the container prefixes are already set by the algorithms themselves.

This means the prefixed column names are written into the  file and the loader can just read them.

Also, as the columns include the prefix now, so do the metadata entries.

I use the the algorithm name as prefix for stereo features and `f"{algorithm}_tel"` as prefix for telescope-wise features. In a first iteration I used `mono`, but this is misleading at least in the case of the impact, as it is a information per telescope but derived from a stereo method.

The Container class attribute `container_prefix` was also renamed to the more fitting `default_prefix` and the prefix can now be passed to `Container.__init__`.

### DIFF
--- a/ctapipe/io/hdf5tableio.py
+++ b/ctapipe/io/hdf5tableio.py
@@ -566,7 +566,7 @@ class HDF5TableReader(TableReader):
         if prefixes is False:
             prefixes = ["" for _ in containers]
         elif prefixes is True:
-            prefixes = [container.container_prefix for container in containers]
+            prefixes = [container.default_prefix for container in containers]
         elif isinstance(prefixes, str):
             prefixes = [prefixes for _ in containers]
 
--- a/ctapipe/io/hdf5tableio.py
+++ b/ctapipe/io/hdf5tableio.py
@@ -606,7 +606,7 @@ class HDF5TableReader(TableReader):
                 for fieldname in missing_cols:
                     kwargs[fieldname] = None
 
-                container = cls(**kwargs)
+                container = cls(**kwargs, prefix=prefix)
                 container.meta = self._meta[table_name]
                 ret.append(container)
 
--- a/ctapipe/io/tests/test_hdf5.py
+++ b/ctapipe/io/tests/test_hdf5.py
@@ -571,7 +571,7 @@ def test_column_exclusions(tmp_path):
     tmp_file = tmp_path / "test_column_exclusions.hdf5"
 
     class SomeContainer(Container):
-        container_prefix = ""
+        default_prefix = ""
         hillas_x = Field(None)
         hillas_y = Field(None)
         impact_x = Field(None)
@@ -615,7 +615,7 @@ def test_column_transforms(tmp_path):
     tmp_file = tmp_path / "test_column_transforms.hdf5"
 
     class SomeContainer(Container):
-        container_prefix = ""
+        default_prefix = ""
 
         current = Field(1 * u.A, unit=u.uA)
         time = Field(NAN_TIME)
@@ -648,7 +648,7 @@ def test_fixed_point_column_transform(tmp_path):
     tmp_file = tmp_path / "test_column_transforms.hdf5"
 
     class SomeContainer(Container):
-        container_prefix = ""
+        default_prefix = ""
         image = Field(np.array([np.nan, np.inf, -np.inf]))
 
     cont = SomeContainer()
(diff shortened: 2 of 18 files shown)


nbiederbeck (approved):

maxnoe:
Note: this breaks backwards compatibility with previously produced dl2 files.

LukasNickel (approved):

vuillaut (dismissed):

ctapipe/io/hdf5tableio.py
```diff
@@ -530,6 +536,7 @@ def read(self, table_name, containers, prefixes=False, ignore_columns=None):
         containers : Iterable[ctapipe.core.Container]
             Container classes to fill
         prefix: bool, str or list
```
nbiederbeck, line 538: or None

ctapipe/io/hdf5tableio.py
```diff
@@ -555,7 +562,16 @@ def read(self, table_name, containers, prefixes=False, ignore_columns=None):
...
 
-        if prefixes is False:
+        if prefixes is None:
+            tab = self._h5file.get_node(table_name)
+            try:
+                prefixes = tab._v_attrs["CONTAINER_PREFIXES"].split(",")
+            except KeyError:
+                raise ValueError(
+                    "prefixes=None, but file does not contain prefix information."
+                    " Specify `prefixes` manually."
```
nbiederbeck, line 572: ... or set to True/False ?

I read this as "give a list of prefixes"

ctapipe/io/tests/test_hdf5.py
```diff
@@ -887,3 +887,38 @@ class Container2(Container):
         for string in expected:
             c = next(generator)
             assert c.string == string
+
+
+@pytest.mark.parametrize("prefixes", [None, ("", ""), ("a", "b")])
```
nbiederbeck, line 892: do we have a test for when prefixes is a string?

maxnoe:
Unfortunately, its not as easy as this, as we allow partial reading of the containers...

ctapipe/reco/hillas_reconstructor.py
```diff
@@ -282,7 +285,7 @@ class are set to np.nan
             az_uncert=err_est_dir,
             h_max=h_max,
         )
-
+        result.prefix = self.__class__.__name__
```
nbiederbeck, line 288: is there a reason to set this here and not inside the container instantiation (two lines above)? Same in hillas_intersection

maxnoe: Just that I did it before introducing the prefix setting in `__init__` feature

ctapipe/tools/tests/test_process.py
```diff
@@ -124,21 +126,25 @@ def test_stage_1_dl1(tmp_path, dl1_image_file, dl1_parameters_file):
...
+        argv=[
+            f"--config={config}",
+            f"--input={dl1_parameters_file}",
+            f"--output={tmp_path}/dl1b_from_dl1b.dl1.h5",
+            "--write-parameters",
+            "--overwrite",
+        ],
+        cwd=tmp_path,
     )
+    assert ret == 1
```
nbiederbeck, line 147: why do we want this to fail?

maxnoe: Because calculating image parameters is impossible when no images are in the file.

nbiederbeck: makes sense. thanks

nbiederbeck (approved):

maxnoe:
@kosack done

nbiederbeck (approved):

kosack (commented):
I guess the main data model change here is just in the column metadata, where the prefix is now (I think more correctly) attached?  Is there anything else?  I would just like to be sure so we can keep track. 

```
44172c43657
<                   ATTRIBUTE "morphology_num_small_islands_DESC" {
---
>                   ATTRIBUTE "true_morphology_num_small_islands_DESC" {
```

I think that might warrant an update to the DATA_MODEL_VERSION, which I don't see here (though we could change that before release since there may be a few PRs that modify it).

maxnoe:
>  I would just like to be sure so we can keep track.

The actual columns in the files also now include the prefix.

maxnoe:
@kosack I updated the desccription of the PR with I think all the needed information.

vuillaut (approved):

PR #1950, opened 2022-06-25
Simplify units code in subarray

maxnoe:


LukasNickel:
Why can we not use `pos.to(u.m)` directly without the value dance?

maxnoe:
A list of quantities is not the same as an array quantity.

LukasNickel:
Right, thx

LukasNickel (approved):

PR #1952, opened 2022-06-26
Create a fresh ArrayEventContainer for each event in SimTelEventSource

maxnoe:
Seems to have no impact on performance:

This branch:
```
SimTelEventSource: 1010ev [00:46, 21.70ev/s]
```

Master:
```
SimTelEventSource: 1010ev [00:46, 21.66ev/s]
```

I was bitten by the reuse of the container again, this time it was DL2 information from the previous events not being cleared.

### DIFF
--- a/ctapipe/io/simteleventsource.py
+++ b/ctapipe/io/simteleventsource.py
@@ -1,3 +1,4 @@
+from collections import defaultdict
 import warnings
 from gzip import GzipFile
 from io import BufferedReader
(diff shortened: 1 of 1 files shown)


kosack (dismissed):
yeah if there is no performance impact, this is definitely a good idea. all there reuse was originally just for performance, but i have to admit i never measured it. 

i’m considering even replacing the current Container implementation with a dataclass based one, with frozen attributes. I tried it and it seems to be must as fast and would prevent most accidental changes if we’re required to fill all containers at creation time

maxnoe:
> i’m considering even replacing the current Container implementation with a dataclass based one,

That won't have all our features like prefixes / data types / help strings etc. dataclasses are a very powerful code generator, but in the end you get normal python classes with some fields. We have our own code generator here (that's essentially what the `ContainerMeta` class is, a code generator) that's I think better suited than starting over from dataclasses.

LukasNickel (approved):
I think this is a good idea.
Avoids all the subtle bugs with reused information

nbiederbeck (dismissed):

nbiederbeck (approved):

LukasNickel (approved):

vuillaut (approved):

PR #1957, opened 2022-06-27
Raise exceptions in tools when run with run_tool

maxnoe:
This makes debugging much easier, since you can step into the exception with a debugger.


nbiederbeck:
maybe add a debug flag?

maxnoe:
Good idea but now how it was implemented here, we need something else since we also want to be able to test the return codes

Issue #1961, opened 2022-06-27
Impact reconstructor stores distance for all telescopes


LukasNickel:
**Describe the bug**
Impact reconstructor stores distance for all telescopes in a subarray independent of if they triggered or not.
This can lead to situations where the dl1 table of a telescope is empty, but the dl2 table contains (useless) impact distances, which leads to confusing behaviour of the tableloader (Columns missing from the table, but no error).

**Expected behavior**
I expect only distances to triggered telescopes to be written.

**Supporting information**

https://github.com/cta-observatory/ctapipe/blob/27e5439eaf103d19dbdb3696b6f40536552b9981/ctapipe/reco/shower_processor.py#L81-L89

**Additional context**
It mostly matters for small files, but our dl2 geometry file is such a candidate:

![Screenshot_2022-06-27_17-20-22](https://user-images.githubusercontent.com/15281090/175975640-ea7b0fd7-dcbc-43d1-96b3-a9594f6b3962.png)

PR #1963, opened 2022-06-27
Ml classifier

LukasNickel:
As discussed in the ctapipe meeting, we implemented the training and applying of the particle classifier including the changes in `ctapipe-process`.
To avoid the connected design discussions, we added new tools instead of expanding the existing train/apply tools for the energy regressor.
This can be added at a later stage (once we also have disp models).

Unrelated change:
To fix the unit tests, the test for the apply energy regressor tool now uses the dl1 parameters file and the ml config has beed adapted to not use subarray features. This is due to #1961


LukasNickel:
Failing tests should be on ml branch, maybe due to 1a500c85d0120866ba0610029355fa48cde6828c

PR #1968, opened 2022-06-28
Maintain event order of input file in TableLoader,  fixes #1966

### LINKED ISSUE #1966: TableLoader changes order of events
**Describe the bug**
The `TableLoader`, due to the usage of `join` for astropy tables, orders events by `(obs_id, event_id`, which is not necessarily the order in the input files (e.g. order of the merged files by @arrabito and @bregeon were completely random).

We should either decide / check that we expect file contents to be ordered by obs_id / event_id or fix the ordering in the table loader using the workaround also used by the aict-tools for the same issue:

https://github.com/fact-project/aict-tools/blob/ee091fc933110ff008e32df02ee12a97e412e8b3/aict_tools/io.py#L507-L510

Also see: https://github.com/astropy/astropy/issues/11619

I would first go for sorting back to the original order as requiring files to be merged in obs_id order seems very restrictive, but it does have a performance implication.

maxnoe:
This is needed for for reading the machine learning predictions made by the apply tools back with the hdf5 event source and in general for consistency.

### DIFF
--- a/ctapipe/io/tableloader.py
+++ b/ctapipe/io/tableloader.py
@@ -97,14 +97,6 @@ def _empty_telescope_events_table():
     return Table(names=TELESCOPE_EVENT_KEYS, dtype=[np.int32, np.int64, np.int16])
 
 
-def _empty_subarray_events_table():
-    """
-    Create a new astropy table with correct column names and dtypes
-    for subarray event based data.
-    """
-    return Table(names=SUBARRAY_EVENT_KEYS, dtype=[np.int32, np.int64])
-
-
 def _get_structure(h5file):
     """
     Check if data is stored by telescope id or by telescope type in h5file.
@@ -135,12 +127,18 @@ def _add_column_prefix(table, prefix, ignore=()):
 
 def _join_subarray_events(table1, table2):
     """Outer join two tables on the telescope subarray keys"""
-    return join_allow_empty(table1, table2, SUBARRAY_EVENT_KEYS, "outer")
+    return join_allow_empty(table1, table2, SUBARRAY_EVENT_KEYS, "left")
 
 
 def _join_telescope_events(table1, table2):
     """Outer join two tables on the telescope event keys"""
-    return join_allow_empty(table1, table2, TELESCOPE_EVENT_KEYS, "outer")
+    # we start with an empty table, but after the first non-empty, we perform
+    # left joins
+    if len(table1) == 0:
+        how = "right"
+    else:
+        how = "left"
+    return join_allow_empty(table1, table2, TELESCOPE_EVENT_KEYS, how)
 
 
 class TableLoader(Component):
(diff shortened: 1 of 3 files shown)


ctapipe/io/tests/test_table_loader.py
```diff
@@ -114,20 +118,42 @@ def test_true_parameters(test_file):
...
 
 
+def test_table_loader_keeps_original_order(dl2_merged_file):
+    """Test reading subarray events keeps order in file"""
+    from ctapipe.io.tableloader import TableLoader
+
+    # check that the order is the same as in the file itself
+    trigger = read_table(dl2_merged_file, "/dl1/event/subarray/trigger")
+    # check we actually have unsorted input
+    assert not np.all(np.diff(trigger["obs_id"]) >= 0)
```
vuillaut, line 139: The input could be sorted there, e.g.`[0, 4 ,10]`
I suppose you are assuming obs_id ranges from 0 to len(trigger)?

maxnoe: > The input could be sorted there, e.g. [0, 4 ,10]

Yes, and then the test wouldn't actually test that order is maintained. So the check here is to make sure the test is actually testing what it is supposed to test

ctapipe/io/tests/test_table_loader.py
```diff
@@ -114,20 +118,42 @@ def test_true_parameters(test_file):
...
+    with TableLoader(
+        dl2_merged_file, load_dl2=True, load_simulated=True
+    ) as table_loader:
+        events = table_loader.read_subarray_events()
+        tel_events = table_loader.read_telescope_events()
+
+    assert np.all(events[["obs_id", "event_id"]] == trigger[["obs_id", "event_id"]])
+
+    events = []
+    for g, _ in groupby(tel_events[["obs_id", "event_id"]]):
```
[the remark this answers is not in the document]
maxnoe: Replaced with documented and tested numpy based solution that is more general

vuillaut (dismissed):

ctapipe/io/tableloader.py
```diff
@@ -362,9 +393,11 @@ def read_telescope_events(self, telescopes=None):
             tel_ids = self.subarray.get_tel_ids(telescopes)
 
         table = self._read_telescope_events_for_ids(tel_ids)
+        table = self._join_subarray_info(table)
 
-        if any([self.load_trigger, self.load_simulated, self.load_dl2]):
-            table = self._join_subarray_info(table)
+        table = _join_subarray_events(table, self._get_sort_index())
+        _join_subarray_events(table, self._get_sort_index())
```
nbiederbeck, line 399: I think this line is duplicated

maxnoe: Good catch

maxnoe:
@kosack @nbiederbeck @vuillaut this should be ready now

nbiederbeck (approved):

kosack (approved):

maxnoe:
Unfortunately merging #1935 created a lot of conflicts, have to solve them first

maxnoe:
Solved. Please review again @kosack @nbiederbeck

nbiederbeck (approved):

LukasNickel (approved):

maxnoe:
And again, sorry @LukasNickel @nbiederbeck

nbiederbeck (approved):

LukasNickel (approved):

kosack (approved):

PR #1969, opened 2022-06-28
Fix multiplied metadata when reading multiple containers from a single table (#1785)

### LINKED ISSUE #1785: Messed up metadata reading from multi-container tables
Reading containers from a table with multiple containers in it leads to duplicated metadata, because the `HDF5TableReader` adds all available metadata to all containers. This has a few implications:
- The containers in memory have more metadata than they should have
- If a container is saved in multiple places, the output from running ctapipe-process on a hdf5 file (e.g. to get dl2 data from a dl1 file), can have messed up metadata. I found this because the `EventIndexContainer` had metadata from the `TriggerContainer`, which in turn was saved in other tables with the event index.
- `read_table` fails with such broken files, because it tries to perform transforms based on metadata entries, which dont have matching columns

This should reproduce it:
```python
from ctapipe.io import HDF5TableReader, HDF5TableWriter
from ctapipe.containers import HillasParametersContainer, EventIndexContainer

H = HillasParametersContainer(
    fov_lon=1 * u.deg, fov_lat=1 * u.deg, length=1 * u.deg, width=1 * u.deg
)
I = EventIndexContainer(obs_id=1, event_id=2)

with HDF5TableWriter("test.h5") as writer:
    writer.write("table", [H, I])

with HDF5TableReader("test.h5") as reader:
    _ = reader.read("/table", [HillasParametersContainer(), EventIndexContainer()])
    H2, I2, = next(_)

print(I2.meta)
```

Relevant piece in the `HDF5TableReader`:
https://github.com/cta-observatory/ctapipe/blob/63e0df4337c18386aab3dea0d0654c5ec02b447a/ctapipe/io/hdf5tableio.py#L489-L490

I guess there should be some kind of check to only read the relevant parts.

nbiederbeck:
When reading multiple containers from a single table, table metadata
gets multiplied to all containers. This has impacts for e.g. the
_TRANSFORM metadata, due to columns not found (on containers that have
now those wrong metadata).

This fix introduces a new (temporary) level while reading, that
separates (different) container metadata from table-metadata.

This way the correct metadata gets assigned to the corresponding
containers.

General metadata (e.g. version number) are assigned to all containers.

### DIFF
--- a/ctapipe/io/hdf5tableio.py
+++ b/ctapipe/io/hdf5tableio.py
@@ -8,16 +8,17 @@ from astropy.time import Time
 from astropy.units import Quantity, Unit
 
 import ctapipe
-from .tableio import (
-    StringTransform,
-    TableWriter,
-    TableReader,
-    FixedPointColumnTransform,
-    TimeColumnTransform,
-    EnumColumnTransform,
-    QuantityColumnTransform,
-)
+
 from ..core import Container, Map
+from .tableio import (
+    EnumColumnTransform,
+    FixedPointColumnTransform,
+    QuantityColumnTransform,
+    StringTransform,
+    TableReader,
+    TableWriter,
+    TimeColumnTransform,
+)
 
 __all__ = ["HDF5TableWriter", "HDF5TableReader"]
 
@@ -424,6 +425,7 @@ class HDF5TableReader(TableReader):
         tab = self._h5file.get_node(table_name)
         self._tables[table_name] = tab
         self._map_table_to_containers(table_name, containers, prefixes, ignore_columns)
+        self._handle_metadata(table_name, containers, prefixes, ignore_columns)
         self._map_transforms_from_table_header(table_name)
         return tab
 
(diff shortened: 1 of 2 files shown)


ctapipe/io/hdf5tableio.py
```diff
@@ -505,9 +509,20 @@ def _map_table_to_containers(
                     )
 
             # store the meta
-            self._meta[table_name] = {}
-            for key in tab.attrs._f_list():
-                self._meta[table_name][key] = tab.attrs[key]
+            self._meta[table_name][container.__name__] = {}
+            for key in container.fields:
```
maxnoe, line 513: Wouldn't it make more sense to loop through the available meta and sort that into the containers instead of looking for all possible combinations?

maxnoe: Probably also would be good to break this out into a `handle_metadata` function

nbiederbeck: Either way I need to strip the (possible) prefix and suffix from the meta key in order to correctly assign those values. I found it easier to create the pre-and-suffixed version from the plain one instead of the other way around (stripping prefix and suffix).

nbiederbeck:
You fixed the problem I had regarding #1785 with your work in #1992, thanks.

maxnoe:
This should be fixed by #2003

PR #1973, opened 2022-06-30
ObsConfig data model

kosack:
## Use case:

Access the observation information for both simulations and real data in a common way, from an `EventSource` or other io component.  

## Description

This PR will create an observation configuration data model that can be used by real or simulated observations.

This may not  end up being an exact 1-1 mapping of the CTA SchedulingBlock/ObservationBlock model as there are some issues with the current version, and we would also like to support non-CTA instruments.

Two new Components are defined: `SchedulingBlockContainer` and `ObservationBlockContainer`
```
ctapipe.containers.SchedulingBlockContainer:
                         sb_id: Scheduling block ID with default
                                18446744073709551615 with type <class
                                'numpy.uint64'>
                       sb_type: Type of scheduling block with default
                                SchedulingBlockType.UNKNOWN with type <enum
                                'SchedulingBlockType'>
                   producer_id: Origin of the sb_id, i.e. name of the telescope
                                site or 'simulation' with default unknown with
                                type <class 'str'>
                observing_mode: Defines how observations within the Scheduling
                                Block are distributed in space with default
                                ObservingMode.UNKNOWN with type <enum
                                'ObservingMode'>
                 pointing_mode: Defines how the telescope drives move with
                                default PointingMode.UNKNOWN with type <enum
                                'PointingMode'>

ctapipe.containers.ObservationBlockContainer:
                        obs_id: Observation Block ID with default
                                18446744073709551615 with type <class
                                'numpy.uint64'>
                         sb_id: ID of the parent SchedulingBlock with default
                                18446744073709551615 with type <class
                                'numpy.uint64'>
                   producer_id: Origin of the obs_id, i.e. name of the telescope
                                site or 'simulation' with default unknown with
                                type <class 'str'>
                         state: State of this OB with default
                                ObservationBlockState.UNKNOWN with type <enum
                                'ObservationBlockState'>
         subarray_pointing_lat: latitude of the nominal center coordinate of
                                this observation with default nan deg [deg]
         subarray_pointing_lon: longitude of the nominal center coordinate of
                                this observation with default nan deg [deg]
       subarray_pointing_frame: Frame in which the subarray_target is non-
                                moving. If the frame is altaz, The meaning of
                                lat,lon is altitude,azimuth.  If it is icrs, the
                                meaning is RA,Dec with default
                                CoordinateFrameType.UNKNOWN with type <enum
                                'CoordinateFrameType'>
            scheduled_duration: expected duration from scheduler with default
                                nan min [min]
          scheduled_start_time: expected start time from scheduler with default
                                0.0
             actual_start_time: true start time with default 0.0
               actual_duration: true duration with default nan min [min]
```

The tables of observation and scheduling blocks are also written do the `/configuration/observation` group of all output files

## Example

```py3
from pprint import pprint

with EventSource("gamma.simtel.gz") as source:
    print([x.as_dict() for x in source.scheduling_blocks.values()])
    print([x.as_dict() for x in source.observation_blocks.values()])
```
```
[{'observing_mode': <ObservingMode.UNKNOWN: -1>,
  'pointing_mode': <PointingMode.DRIFT: 1>,
  'producer_id': 'simulation',
  'sb_id': 5,
  'sb_type': <SchedulingBlockType.OBSERVATION: 0>}]

[{'actual_duration': <Quantity nan min>,
  'actual_start_time': <Time object: scale='tai' format='mjd' value=0.0>,
  'obs_id': 5,
  'producer_id': 'simulation',
  'sb_id': 5,
  'scheduled_duration': <Quantity nan min>,
  'scheduled_start_time': <Time object: scale='tai' format='mjd' value=0.0>,
  'state': <ObservationBlockState.COMPLETED_SUCCEDED: 1>,
  'subarray_pointing_frame': <CoordinateFrameType.ALTAZ: 0>,
  'subarray_pointing_lat': <Quantity 1.2217305 deg>,
  'subarray_pointing_lon': <Quantity 0. deg>}]
```

# Todo:

- [x]  Have `EventSources` fill this information for each observation they provide
- [x] Ensure that `ctapipe-merge`  can merge this info
- [x] Decide how to store the info for DIVERGENT/CONVERGENT modes, at minimum using a convergence point and depth (maybe we don't need special modes for this, and just define convergence_depth=0 == PARALLEL)

This can later be used in #1918 for SimTelEventSource to provide the pointing info


ctapipe/containers.py
```diff
@@ -1044,3 +1088,86 @@ class ArrayEventContainer(Container):
...
+    of the SB model, only storing what is necessary for analysis."""
+
+    container_prefix = "sb"
+    id = Field(-1, "Scheduling block ID", type=np.int64)
+    type = Field(
+        SchedulingBlockType.OBSERVATION,
+        description="Type of scheduling block",
+        type=SchedulingBlockType,
+    )
+    observing_mode = Field(ObservingMode.CUSTOM, "Observing mode", type=ObservingMode)
```
maxnoe, line 1104: I think we should always add an `UNKNOWN` -1 value here, not one of the actual possible values

ctapipe/containers.py
```diff
@@ -1044,3 +1088,86 @@ class ArrayEventContainer(Container):
...
+    pointing_mode = Field(
+        None,
+        "Pointing Mode is the coordinate frame in which the pointing position is non-moving",
+        type=CoordinateFrameType,
```
maxnoe, line 1108: Why is this a `CoordinateFrameType` and not a `PointingMode` enum? As mentioned, this could also be an asteroid tracking for example ... So I would really call it `PointingMode` and to keep possible extensions to things not purely related to being constant in a specific frame.

ctapipe/containers.py
```diff
@@ -56,13 +57,56 @@
...
+    WOBBLE = 0
+    ON_OFF = 1
+    GRID = 2
+    CUSTOM = 3
+
+
+class CoordinateFrameType(enum.Enum):
+    ALTAZ = 0
+    ICRS = 1
+    GALACTIC = 2  # is this needed?
```
maxnoe, line 107: No, not for pointing mode, since a fixed pointing in GALACTIC is also fixed in ICRS.

ctapipe/containers.py
```diff
@@ -1044,3 +1088,86 @@ class ArrayEventContainer(Container):
...
+        "Pointing Mode is the coordinate frame in which the pointing position is non-moving",
+        type=CoordinateFrameType,
+    )
+
+
+class ObservationBlockContainer(Container):
+    """Stores information about the observation"""
+
+    container_prefix = "obs"
+    id = Field(None, "Observation Block ID", type=np.int64)
```
maxnoe, line 1116: We call this `obs_id` every where else. This is a bit of a question of naming conventions for foreign key / primary keys.

I normally also use `id` in the actual table and then `<name>_id` when used somewhere else as foreign key, so this fits, but something to be kept in mind.

ctapipe/containers.py
```diff
@@ -1044,3 +1088,86 @@ class ArrayEventContainer(Container):
...
+            "The meaning of lat,lon is altitude,azimuth.  If it is icrs, the meaning is "
+            "RA,Dec"
+        ),
+        type=CoordinateFrameType,
+    )
+
+    scheduled_duration = Field(None, "expected duration from scheduler", unit=u.min)
+    scheduled_start_time = Field(NAN_TIME, "expected start time from scheduler")
+    actual_start_time = Field(NAN_TIME, "true start time")
+    actual_duration = Field(None, "true duration", unit=u.min)
```
maxnoe, line 1150: Why `None` and not `nan`? This is a quantity

maxnoe (changes requested):
HDF5EventSource implementation is missing

For the rest see inline comments.

ctapipe/containers.py
kosack: Ah, I was just playing with the idea that they are actually the same thing and hadn't implemeted the PointingModeType, but I will likely add the PointingModeType.  The reason I was thinking about this was that this defines a CoordinateFrame that is fixed in time: e.g. DRIFT implies a fixed ALTAZ, POINTING (which I don't like because that is not the definition of pointing) implies ICRS,  and if you track the moon for example, it would imply LUNAR)

kosack: So maybe  really we don't even need this mode, we just need a pointing_coordinate that has a frame that is implied to be fixed in time.

ctapipe/containers.py
kosack: true, so I will define this separately

ctapipe/containers.py
maxnoe: The `POINTING` name comes from OGIP.

I also don't like it for the same reason of not being obvious.. We should just call them `TRACKING_ALTAZ` or `TRACKING_ICRS` or similar.

ctapipe/containers.py
```diff
@@ -56,13 +57,67 @@
...
+class SchedulingBlockType(enum.Enum):
+    """
+    Types of Scheduling Block
+    """
+
+    UNKNOWN = -1
+    OBSERVATION = 0
+    CALIBRATION = 1
+    ENGINEERING = 2
+    SIMULATION = 3
```
maxnoe, line 81: Isn't this Orthogonal? Couldn't we have simulated observation/calibration/engineering SBs?

ctapipe/containers.py
```diff
@@ -1044,3 +1099,90 @@ class ArrayEventContainer(Container):
...
+
+    container_prefix = "sb"
+    id = Field(-1, "Scheduling block ID", type=np.int64)
+    type = Field(
+        SchedulingBlockType.OBSERVATION,
+        description="Type of scheduling block",
+        type=SchedulingBlockType,
+    )
+    observing_mode = Field(
+        ObservingMode.CUSTOM,
```
maxnoe, line 1116: unknown.

ctapipe/containers.py
```diff
@@ -1044,3 +1099,90 @@ class ArrayEventContainer(Container):
...
+        "Defines how observations within the Scheduling Block are distributed in space",
+        type=ObservingMode,
+    )
+    pointing_mode = Field(
+        None, "Defines how the telescope drives move", type=PointingMode
```
maxnoe, line 1121: unknown.

ctapipe/io/eventsource.py
```diff
@@ -208,6 +208,14 @@ def subarray(self) -> SubarrayDescription:
 
         """
 
+    @property
+    @abstractmethod
+    def obs_config(self) -> list[ObservationConfigurationContainer]:
```
maxnoe, line 213: This type annotation is only supported from 3.9, we still support 3.8. Use `from typing import List; List[ObservationConfigurationContainer]`. Since this also should be immutable, I'd prefer `Tuple[ObservationConfiguration]`

try_obsconfig_model.ipynb
```diff
@@ -0,0 +1,614 @@
+{
+ "cells": [
```
maxnoe, line 2: Remove or make this a proper `docs/tutorial`?

ctapipe/io/eventsource.py
```diff
@@ -208,6 +208,14 @@ def subarray(self) -> SubarrayDescription:
 
         """
 
+    @property
+    @abstractmethod
+    def obs_config(self) -> ObservationConfigurationContainer:
```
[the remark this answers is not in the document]
kosack: yes, this is already planned

try_obsconfig_model.ipynb
kosack: this was committed accidentally (as were the commits to EventSource) I'll remove both.

ctapipe/containers.py
kosack: yes, will remove

kosack:
Another question: 

`SimTelEventSource` and `HDF5EventSource` only return one` SimulationConfigurationContainer` via their `simulation_config` property .   the `observation_config` should be essentially the same interface.   Should all three support returning a tuple of  both observation and simulation configs? I guess the answer is yes, but that is not what happens now.

ctapipe/io/eventsource.py
```diff
@@ -210,9 +210,9 @@ def subarray(self) -> SubarrayDescription:
 
     @property
     @abstractmethod
-    def obs_config(self) -> list[ObservationConfigurationContainer]:
+    def observation_config(self) -> Dict[ObservationConfigurationContainer]:
```
maxnoe, line 213: Isn't the correct type hint `Dict[int, ObservationConfigurationContainer]`? I.e. including the type of the key?

maxnoe: Yes: https://docs.python.org/3/library/typing.html#typing.Dict

ctapipe/io/hdf5eventsource.py
```diff
@@ -220,13 +221,17 @@ def obs_ids(self):
         return list(np.unique(self.file_.root.dl1.event.subarray.trigger.col("obs_id")))
 
     @property
-    def simulation_config(self):
+    def simulation_config(self) -> Dict[SimulationConfigContainer]:
```
maxnoe, line 224: Same as above

ctapipe/io/hdf5eventsource.py
```diff
@@ -220,13 +221,17 @@ def obs_ids(self):
...
         """
         Returns the simulation config(s) as
         a dict mapping obs_id to the respective config.
         """
         return self._simulation_configs
 
+    @property
+    def observation_config(self) -> Dict[ObservationConfigurationContainer]:
```
maxnoe, line 232: Same as above, Dict[int, ...]

ctapipe/io/eventsource.py
kosack: yes, definitely incorrect

ctapipe/containers.py
```diff
@@ -57,13 +57,65 @@
...
+    UNKNOWN = -1
+    WOBBLE = 0
+    ON_OFF = 1
+    GRID = 2
+    CUSTOM = 3
+
+
+class PointingMode(enum.Enum):
+    UNKNOWN = -1
+    TRACK = 0  # track a fixed point that moves with the sky
```
maxnoe, line 107: If you make these comments like this:

```
#: track a fixed point in ICRS
TRACK = 0
```

they are picked up by sphinx as docstrings, see: 
https://github.com/cta-observatory/ctapipe/blob/8c42f54f805d81b152821360f07974d17acbd7f2/ctapipe/instrument/optics.py#L23

ctapipe/containers.py
```diff
@@ -1143,3 +1195,74 @@ class ArrayEventContainer(Container):
...
+    of the SB model, only storing what is necessary for analysis.
+
+    References
+    ----------
+    .. [sb-ob] Oya et al, Scheduling Block Data Model Specification,
+        CTA-SPE-COM-000000-0003, Issue 1, Rev. c
+    """
+
+    default_prefix = ""
+    sb_id = Field(-1, "Scheduling block ID", type=np.int64)
```
maxnoe, line 1211: in the data model, the sb_id is `uint64`.

ctapipe/containers.py
```diff
@@ -1143,3 +1195,74 @@ class ArrayEventContainer(Container):
...
+    pointing_mode = Field(
+        PointingMode.UNKNOWN, "Defines how the telescope drives move", type=PointingMode
+    )
+
+
+class ObservationBlockContainer(Container):
+    """Stores information about the observation"""
+
+    default_prefix = ""
+    obs_id = Field(None, "Observation Block ID", type=np.int64)
```
maxnoe, line 1231: same here, dl0 data model has uint64. I'd also use a default of `np.iinfo(np.uint64).max` so the column is always written.

ctapipe/containers.py
```diff
@@ -1143,3 +1195,74 @@ class ArrayEventContainer(Container):
...
+    producer_id = Field(
+        "unknown",
+        "Origin of the obs_id, i.e. name of the telescope site or 'simulation'",
```
maxnoe, line 1234: Add `type=str`?

ctapipe/io/eventsource.py
```diff
@@ -205,6 +209,22 @@ def subarray(self) -> SubarrayDescription:
...
+    @abstractmethod
+    def observation_block(self) -> Dict[int, ObservationBlockContainer]:
+        """
+        Obtain the ObservationConfigurations from the EventSource, indexed by obs_id
+        """
+        pass
+
+    @property
+    @abstractmethod
+    def scheduling_block(self) -> Dict[int, SchedulingBlockContainer]:
```
maxnoe, line 222: We have the plural in `obs_ids`.

Also, since we are now requiring `observation_blocks` we could remove the `abstractmethod` from `def obs_ids` and just `return self.observation_blocks.keys()`

ctapipe/io/hdf5eventsource.py
```diff
@@ -243,7 +250,15 @@ def obs_ids(self):
         return list(np.unique(self.file_.root.dl1.event.subarray.trigger.col("obs_id")))
 
     @property
-    def simulation_config(self):
```
maxnoe, line 246: We should probably add `simulation_config(s)` to the `EventSource` as interface, returning `None` for observations?

ctapipe/io/hdf5eventsource.py
```diff
@@ -272,16 +287,32 @@ class ObsIdContainer(Container):
...
+
+    def _parse_sb_and_ob_configs(self):
+        """read Observation and Scheduling block configurations"""
+
+        sb_reader = HDF5TableReader(self.file_).read(
+            "/configuration/observation/scheduling_block",
+            containers=SchedulingBlockContainer,
+        )
+
+        scheduling_blocks = {sb: sb.sb_id for sb in sb_reader}
```
maxnoe, line 307: This dict comprehension looks switched. Shouldn't it be: `{sb.sb_id: sb ...}`

ctapipe/io/hdf5eventsource.py
```diff
@@ -272,16 +287,32 @@ class ObsIdContainer(Container):
...
+
+        ob_reader = HDF5TableReader(self.file_).read(
+            "/configuration/observation/observation_block",
+            containers=ObservationBlockContainer,
+        )
+        observation_blocks = {ob.obs_id: ob for ob in ob_reader}
```
maxnoe, line 313: This dict comprehension looks switched. Shouldn't it be: `{ob.obs_id: ob ...}`?

ctapipe/io/simteleventsource.py
```diff
@@ -816,6 +842,36 @@ def _parse_simulation_header(self):
...
+                state=ObservationBlockState.COMPLETED_SUCCEDED,
+                subarray_pointing_lat=alt * u.deg,
+                subarray_pointing_lon=az * u.deg,
+                subarray_pointing_frame=CoordinateFrameType.ALTAZ,
+            )
+        }
+
+        sb_dict = {
+            obs_id: SchedulingBlockContainer(
+                sb_id=np.int64(obs_id),  # simulations have no SBs, so use the OB id
```
maxnoe, line 866: dtype

ctapipe/io/simteleventsource.py
```diff
@@ -816,6 +842,36 @@ def _parse_simulation_header(self):
...
+        """fill scheduling and observation blocks must be run after the
+        simulation config is filled
+        """
+
+        az, alt = self.file_.header["direction"]
+        obs_id = self.obs_ids[0]
+
+        ob_dict = {
+            obs_id: ObservationBlockContainer(
+                obs_id=np.int64(obs_id),
```
maxnoe, line 855: dtype

maxnoe (changes requested):

ctapipe/io/hdf5eventsource.py
kosack: Yup, clearly wrong

ctapipe/io/hdf5eventsource.py
kosack: Yes, to be symmetric this is a good idea.

ctapipe/io/simteleventsource.py
kosack: The dtype is int64 in the Container definition so it has to be explicitly cast, right?

kosack: It failed to validate the container if I do not do this.

maxnoe: See comments above, the DL0 draft says these should be uint64

maxnoe (changes requested):
The `ObservationBlockContainer` needs an `sb_id` to connect to the `SchedulingBlock`

ctapipe/containers.py
```diff
@@ -104,8 +107,10 @@ class ObservingMode(enum.Enum):
 
 class PointingMode(enum.Enum):
     UNKNOWN = -1
-    TRACK = 0  # track a fixed point that moves with the sky
-    DRIFT = 1  # track a fixed alt/az point
+    #: drives track a fixed point that moves with the sky
```
maxnoe, line 110: A "fixed point" that moves sounds a bit strange, although all in all, it is clear what is meant.

Maybe: "Drives track a fixed position in ICRS".

ctapipe/io/eventsource.py
kosack: good point - I'll drop the requirement to fill obs_ids and make it a non-abstract property of the base class

ctapipe/io/toymodel.py
```diff
@@ -76,12 +79,14 @@ def datalevels(self):
         return (DataLevel.DL1_IMAGES,)
 
     @property
-    def observation_block(self):
-        return dict()
+    def scheduling_blocks(self) -> Dict[int, SchedulingBlockContainer]:
+        return {-1: SchedulingBlockContainer(sb_id=-1, producer_id="ctapipe toymodel")}
```
maxnoe, line 83: -1 doesn't make sense since you now defined it as uint64. Just using the default will keep the invalid value, so maybe just do:

```
sb = SchedulingBlock(produce_id="ctapipe toymode")
return {sb.sb_id: sb}
```

?

ctapipe/io/toymodel.py
```diff
@@ -76,12 +79,14 @@ def datalevels(self):
...
 
     @property
-    def scheduling_block(self):
-        return dict()
+    def observation_blocks(self) -> Dict[int, ObservationBlockContainer]:
+        return {
+            -1: ObservationBlockContainer(obs_id=-1, producer_id="ctapipe toymodel")
```
maxnoe, line 88: same here

ctapipe/containers.py
```diff
@@ -57,18 +57,77 @@
...
+    """How a scheduling block is observed."""
+
+    UNKNOWN = -1
+    WOBBLE = 0
+    ON_OFF = 1
+    GRID = 2
+    CUSTOM = 3
+
+
+class PointingMode(enum.Enum):
```
maxnoe, line 108: Missing docstring

ctapipe/containers.py
```diff
@@ -57,18 +57,77 @@
...
+    UNKNOWN = -1
+    #: drives track a point that moves with the sky
+    TRACK = 0
+    #: drives stay fixed at an alt/az point while the sky drifts by
+    DRIFT = 1
+
+
+class CoordinateFrameType(enum.Enum):
```
maxnoe, line 116: Missing docstring

ctapipe/io/hdf5eventsource.py
```diff
@@ -60,6 +63,7 @@
 
 COMPATIBLE_DATA_MODEL_VERSIONS = [
     "v4.0.0",
+    "v4.1.0",
```
maxnoe, line 66: Since `4.0.0` was never part of a released version of ctapipe, I'd be inclined to not make it a new version and to just include it in version `4.0.0`. In the other PRs changing the date model I only updated the changelog for 4.0, not bumped the version each time.

maxnoe: There should be no reason for us to support reading files without the obsconfig / sb config tables now.

ctapipe/io/simteleventsource.py
```diff
@@ -816,6 +837,38 @@ def _parse_simulation_header(self):
...
+
+        ob_dict = {
+            obs_id: ObservationBlockContainer(
+                obs_id=np.uint64(obs_id),
+                sb_id=np.uint64(obs_id),  # see comment above
+                producer_id="simulation",
+                state=ObservationBlockState.COMPLETED_SUCCEDED,
+                subarray_pointing_lat=alt * u.deg,
+                subarray_pointing_lon=az * u.deg,
+                subarray_pointing_frame=CoordinateFrameType.ALTAZ,
```
maxnoe, line 866: the simtel file contains timestamps for the run here:

```
t = Time(self.file_.header['time'], format='unix')
```

maybe use that for `{scheduled,actual}_start`?

ctapipe/io/hdf5eventsource.py
kosack: ok, I'll unbump it (just wasnt sure if any files were written already with the dev version)

ctapipe/io/simteleventsource.py
kosack: I thought they were basically meaningless, but yes, I can add them.  The duration is at least not something we can use here.

maxnoe (approved):
Looks good now!

ctapipe/containers.py
```diff
@@ -1143,3 +1214,81 @@ class ArrayEventContainer(Container):
...
+        "longitude of the nominal center coordinate of this observation",
+        unit=u.deg,
+    )
+
+    subarray_pointing_frame = Field(
+        CoordinateFrameType.UNKNOWN,
+        (
+            "Frame in which the subarray_target is non-moving. If the frame is altaz, "
+            "The meaning of lat,lon is altitude,azimuth.  If it is icrs, the meaning is "
+            "RA,Dec"
```
maxnoe, line 1284: sorry, one last comment, inconsitent order here:

```
            "The meaning of lat,lon is altitude,azimuth.  If it is icrs, the meaning is "
            "RA,Dec"
```

ra is lon, dec is lat.

kosack: good catch

maxnoe (approved):

maxnoe (approved):

maxnoe (approved):

nbiederbeck (approved):

PR #1976, opened 2022-06-30
recursively validate containers

### LINKED ISSUE #1975: Container.validate() doesn't validate sub-containers
**Describe the bug**

Containers contained in Containers do not get validated with the parent container's `.validate()` method is called.

**To Reproduce**
```py3
from ctapipe.core import Container, Field

class Sub(Container):
    y = Field(0*u.m, "Sub.y", unit=u.m)

class A(Container):
    x = Field(0*u.m, "A.x", unit=u.m)
    sub = Field(None, "sub", type=Sub)

# works:
try: 
    A(x=10*u.s, sub=Sub(y=10*u.m)).validate()
except Exception as err:
    print(err)

#> A Field 'x': the value '10.0 s' (<class 'astropy.units.quantity.Quantity'>) is invalid: : 's' (time) and 'm' (length) are not convertible

# doesn't not raise exception (the sub field doesn't get checked recursively):
try: 
    A(x=10*u.m, sub=Sub(y=10*u.s)).validate()
except Exception as err:
    print(err)
```

**Expected behavior**
Should raise exception if any sub-Container also fails validation

kosack:
fixes #1975

### DIFF
--- a/ctapipe/core/container.py
+++ b/ctapipe/core/container.py
@@ -108,6 +108,10 @@ class Field:
                 f"{errorstr} Should be an instance of {self.type}"
             )
 
+        if isinstance(value, Container):
+            # recursively check sub-containers
+            value.validate()
+
         if self.unit is not None:
             if not isinstance(value, Quantity):
                 raise FieldValidationError(
--- a/ctapipe/core/container.py
+++ b/ctapipe/core/container.py
@@ -111,6 +111,7 @@ class Field:
         if isinstance(value, Container):
             # recursively check sub-containers
             value.validate()
+            return
 
         if isinstance(value, Map):
             for key, map_value in value.items():
@@ -119,6 +120,7 @@ class Field:
                         map_value.validate()
                     except FieldValidationError as err:
                         raise FieldValidationError(f"[{key}]: {err} ")
+            return
 
         if self.unit is not None:
             if not isinstance(value, Quantity):
--- a/ctapipe/core/container.py
+++ b/ctapipe/core/container.py
@@ -115,7 +115,10 @@ class Field:
         if isinstance(value, Map):
             for key, map_value in value.items():
                 if isinstance(map_value, Container):
-                    map_value.validate()
+                    try:
+                        map_value.validate()
+                    except FieldValidationError as err:
+                        raise FieldValidationError(f"[{key}]: {err} ")
 
         if self.unit is not None:
             if not isinstance(value, Quantity):
--- a/ctapipe/core/container.py
+++ b/ctapipe/core/container.py
@@ -112,6 +112,11 @@ class Field:
             # recursively check sub-containers
             value.validate()
 
+        if isinstance(value, Map):
+            for key, map_value in value.items():
+                if isinstance(map_value, Container):
+                    map_value.validate()
+
         if self.unit is not None:
             if not isinstance(value, Quantity):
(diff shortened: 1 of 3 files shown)


ctapipe/core/container.py
```diff
@@ -2,6 +2,7 @@
 from copy import deepcopy
 from pprint import pformat
 from textwrap import wrap
+from typing import Container
```
maxnoe, line 5: Not the container you are looking for

maxnoe (dismissed):

ctapipe/core/container.py
```diff
@@ -108,6 +108,10 @@ def validate(self, value):
                 f"{errorstr} Should be an instance of {self.type}"
             )
 
+        if isinstance(value, Container):
```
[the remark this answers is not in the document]
maxnoe: No, inside of methods the Container is already defined.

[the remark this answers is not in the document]
maxnoe: You can of course refer to the class inside of methods...

kosack:
@maxnoe sorry , you can re-review this now - i added also support for Maps, so you can now do `event.validate()` and it checks everything

ctapipe/core/container.py
```diff
@@ -112,6 +112,11 @@ def validate(self, value):
             # recursively check sub-containers
             value.validate()
 
+        if isinstance(value, Map):
+            for key, map_value in value.items():
```
maxnoe, line 116: key is not needed right? Or better, catch the validation error and report the key so users can better see what failed?

maxnoe:
```
try:
   map_value.validate()
except Exception as e:
   raise FieldValidationError(f"Failed to validate map entry {key}: {e}") from e
```

kosack: Good idea.  Now the resulting error message is for example:
```
FieldValidationError: ArrayEventContainer Field 'dl1': DL1Container Field 'tel': [3]: DL1CameraContainer Field 'parameters': ImageParametersContainer Field 'hillas': HillasParametersContainer Field 'r': the value '10.0 m' (<class 'astropy.units.quantity.Quantity'>) is invalid: : 'm' (length) and 'deg' (angle) are not convertible
```

maxnoe (approved):

maxnoe (approved):

maxnoe:
> sorry , you can re-review this now - i added also support for Maps, so you can now do event.validate() and it checks everything

We should probably add this to a couple of unit tests, e.g. image processor and shower processor tests.

ctapipe/core/container.py
```diff
@@ -108,6 +108,18 @@ def validate(self, value):
                 f"{errorstr} Should be an instance of {self.type}"
             )
 
+        if isinstance(value, Container):
+            # recursively check sub-containers
+            value.validate()
+
+        if isinstance(value, Map):
```
nbiederbeck, line 115: shouldn't this return early? If value is a Map, the other validity checks don't need to run?

same above.

and if yes we may need to create a test that checks this?

kosack: I added some returns for both cases. No need to check other attributes for Map or Container fields.

maxnoe (approved):

nbiederbeck (approved):

Issue #1978, opened 2022-06-30
Remove `{CameraDescription/Optics/Telescope}.from_name` methods


maxnoe:
**Please describe the use case that requires this feature.**

While these are handy shortcuts for tests etc. it turned out to also create a lot of bugs and surprising behavior,
e.g. people analyzing / plotting data simulated with Prod5 using the `from_name` version, which loads an older geometry for most telescopes.

**Describe the solution you'd like**

We should remove these methods and make it clear that the subarray information should always be read from the input file or some other source instead of relying on the `from_name` mechanism.

**Describe alternatives you've considered**

Making the `from_name` system more complex by including more information like versions and prod numbers. But I don't like it, we should just remove the method instead of making the leaky abstraction more complex.

kosack:
The original intent was that we would have a selectable set of input data for this and other tests, e.g. one per prod.  See #738 for example.  It just never got implemented.

maxnoe:
@kosack I think we can better enable this through fixtures, e.g. in the pull request #1985 I introduce fixtures like `subarray_prod5_paranal`.

maxnoe:
There was an issue again: #2157 we should probably go ahead with this...

STSpencer:
So it's been a while since I commented on anything with respect to ctapipe, but since this has an effect on a number of ongoing projects with current generation instruments I'd like to make a request. I agree that getting rid of the from_name functionality makes sense as you propose @maxnoe; I think the big problem is that many users don't understand the different camera model versions. It's also not currently possible (as far as I can tell) to determine the available options for this method (i.e. what files are on the data server) or to tell where the associated camera geometry files actually came from.

Going to a purely from_url/from ctapipe-process file-type system is good (because it's more difficult to abuse it). But somewhere in the documentation I'd ask there be downloadable links to historical camera geometry files, who contributed them, and what exact instrument/simulation model they correspond to. The reality is that we're going to be dealing with (i.e.) CHEC-S on ASTRI data or H.E.S.S. data not in ctapipe-compatible data formats for some time yet.

maxnoe:
> The reality is that we're going to be dealing with (i.e.) CHEC-S on ASTRI data or H.E.S.S. data not in ctapipe-compatible data formats for some time yet.

Just to comment on this part for now: that involves IO plugins and other custom code by the Experte for those systems. These IO plugins should provide the corresponding instrument descriptions by whatever means they deem appropriate and should not rely on the ctapipe test data server to provide them.

[...]

PR #1985, opened 2022-07-01
Use pytest fixtures in unit tests instead of `from_name` methods

maxnoe:
We are preparing a number of changes to the data model, including also the instrument model and dropping support for older versions.

This is very hard to do when the tests are relying on very old files on the data server accessed via `from_name`.

Here I replace all use in our unit tests of `from_name` with pytest fixtures, which are defined mostly in `ctapipe/conftest` so we can change them in a single location.

Also, most fixtures are updated to use prod5 data.


kosack:
what's the reasoning for this?

maxnoe:
@kosack As explained in #1978, I think these methods do more harm than good. 

They make it very easy to make mistakes like loading the wrong instrument compared to the data being analyzed. I stopped counting the occurences I had to change code from doing `from_name` to reading the subarray from the input file / just using the one provided by the event source.

maxnoe:
I think the use case for our unit tests is much better solved by defining common fixtures. 

It reduced the code already by quite a lot. I could also imagine just doing the changes in our tests and keeping `from_name` for now, but the main thing is that it was abused far too often resulting in many errors.

maxnoe:
@kosack I removed the "remove from_name" part for now and only included the changes to the fixtures.

It would make it much easier to test the data model changes without constantly having to update files on the data server to include this.

kosack (approved):

kosack:
This is nice in that we don't rely on the data format.  However, what about tests to read old data formats?  We should still have those (perhaps not now but after the 1.0 release,  to ensure backward compatibility)

maxnoe:
The point is, that we now do want do break backwards compatibilty.

For me this means that tests checking compatibility to old data should be added after we merge the current PRs changing the data model, make a release and first thing we add tests testing files pepduced with that release?

maxnoe:
I would propose the following:

* Merge the remaining, data model changing PRs
* Make the 0.16.0 release
* Process test data files using the released version and add them to the data server in some meaningful structure
* Add (parametrized) tests checking we are still compatible with all data model versions we want to support (i.e. for now the current one for 0.16.0 and the currently implemented on, e.g.:

``
@pytest.fixture("session", params=["v0.16.0", "current"])
def prod5_dl1_files(request, prod5_dl1_file):
    if request.param == "current":
        return prod5_dl1_file
    return get_dataset_path(f"prod5/{request.param}/test.dl1.h5")
```

File: ctapipe/image/tests/test_hillas.py [hunk 1]
```diff
@@ -259,18 +257,18 @@ def test_single_pixel():
...
     """
     np.random.seed(42)
 
     telescope_frame = TelescopeFrame()
-    camera_frame = CameraFrame(focal_length=28 * u.m)
 
-    geom = CameraGeometry.from_name("LSTCam")
-    geom.frame = camera_frame
+    geom = prod5_lst.camera.geometry
+    telescope_frame = TelescopeFrame()
```
nbiederbeck, line 270: duplicated line

maxnoe: fixed

maxnoe:
@kosack @nbiederbeck @LukasNickel Please review again

kosack (approved):

PR #1987, opened 2022-07-01
Fix comparison of timestamps in calibration code

maxnoe:
This code was comparing an astropy `TimeDelta` object to a plain number,
which was supposed to be in seconds. However, astropy treats plain
numbers as days, not seconds. Since astropy 5.1, this creates a warning
and made me find this bug. See
https://github.com/astropy/astropy/issues/12887.

### DIFF
--- a/ctapipe/calib/camera/pedestals.py
+++ b/ctapipe/calib/camera/pedestals.py
@@ -72,8 +72,7 @@ class PedestalCalculator(Component):
         Set to None if no configuration to pass.
 
     kwargs
-
-"""
+    """
 
     tel_id = Int(0, help="id of the telescope to calculate the pedestal values").tag(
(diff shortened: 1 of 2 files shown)


ctapipe/calib/camera/pedestals.py
```diff
@@ -257,7 +256,7 @@ def calculate_pedestals(self, event):
 
         self.collect_sample(dl1.image, pixel_mask)
 
-        sample_age = trigger_time - self.time_start
+        sample_age = (trigger_time - self.time_start).to_value(u.s)
```
maxnoe, line 259: Sorry for the "black noise". This is one of two only real changes to the code.

LukasNickel (approved):

PR #1988, opened 2022-07-01
Improve units code in hillas_parameters

maxnoe:


ctapipe/image/hillas.py
```diff
@@ -105,12 +104,15 @@ def hillas_parameters(geom, image):
         container of hillas parametesr
     """
     unit = geom.pix_x.unit
-    pix_x = Quantity(np.asanyarray(geom.pix_x, dtype=np.float64)).value
-    pix_y = Quantity(np.asanyarray(geom.pix_y, dtype=np.float64)).value
+    pix_x = geom.pix_x.to_value(unit)
```
nbiederbeck, line 107: `thing.to_value(thing.unit)` is not better than `thing.value` or do I misunderstand the reasoning?

LukasNickel (approved):

ctapipe/image/hillas.py
maxnoe: This catches the (hypothetical) case that y and x don't have the same units by first getting the units from x and then using `to_value` on both.

nbiederbeck: Ok. I also see now that the unit gets reattached at the bottom :+1:

nbiederbeck (approved):

vuillaut (approved):

PR #1991, opened 2022-07-02
Replace deepcopying defaults with default_factory

maxnoe:
Applying `deepcopy` to the defaults of a container is quite slow, especially for deep hierarchies, as the copy is not just made once, but at each level.

Here I replace the deepcopying with a `default_factory` approach as known e.g. from the `defaultdict`. This dramatically speeds up container creation, especially for deeply nested containers.

Timing code:

```python
import numpy as np
import timeit
from ctapipe.containers import HillasParametersContainer, ArrayEventContainer
import astropy.units as u

width = u.Quantity(0.5, u.deg)
length = u.Quantity(1.0, u.deg)

code = '''
h = HillasParametersContainer(
    intensity=500,
    width=width,
    length=length,
    prefix="hillas_foo",
)
'''

exec(code)

N = 100
repeat = 1000
times = np.array(timeit.repeat(code, number=N, repeat=repeat, globals=globals()))
print('HillasParametersContainer')
print(f'{np.mean(times / N) * 1e6:.2f} ± {np.std(times / N) * 1e6:.2f} µs')


code = '''
a = ArrayEventContainer()
'''

exec(code)

N = 100
repeat = 100
times = np.array(timeit.repeat(code, number=N, repeat=repeat, globals=globals()))
print('ArrayEventContainer')
print(f'{np.mean(times / N) * 1e6:.2f} ± {np.std(times / N) * 1e6:.2f} µs')
```

Result on master:

```
HillasParametersContainer
14.45 ± 0.24 µs
ArrayEventContainer
161.24 ± 8.34 µs
```

Results on this branch:
```
HillasParametersContainer
2.32 ± 0.10 µs
ArrayEventContainer
10.67 ± 0.42 µs
```

### DIFF
--- a/ctapipe/core/container.py
+++ b/ctapipe/core/container.py
@@ -1,5 +1,5 @@
 from collections import defaultdict
-from copy import deepcopy
+from functools import partial
 from pprint import pformat
 from textwrap import wrap, dedent
 import warnings
@@ -15,6 +15,10 @@ log = logging.getLogger(__name__)
 __all__ = ["Container", "Field", "FieldValidationError", "Map"]
 
 
+def _fqdn(obj):
+    return f"{obj.__module__}.{obj.__qualname__}"
+
+
 class FieldValidationError(ValueError):
     pass
 
@@ -26,8 +30,10 @@ class Field:
     Parameters
     ----------
     default :
-        default value of the item. This will be set when the `Container`
-        is constructed, as well as when  ``Container.reset`` is called
+        Default value of the item. This will be set when the `Container`
+        is constructed, as well as when  ``Container.reset`` is called.
+        This should only be used for immutable values. For mutable values,
+        use ``default_factory`` instead.
     description : str
         Help text associated with the item
     unit : str or astropy.units.core.UnitBase
@@ -46,6 +52,8 @@ class Field:
     max_len : int
         if type is str, max_len is the maximum number of bytes of the utf-8
         encoded string to be used.
+    default_factory : Callable
+        A callable providing a fresh instance as default value.
     """
 
     def __init__(
@@ -59,8 +67,10 @@ class Field:
         ndim=None,
         allow_none=True,
         max_length=None,
+        default_factory=None,
     ):
         self.default = default
+        self.default_factory = default_factory
         self.description = description
         self.unit = Unit(unit) if unit is not None else None
         self.ucd = ucd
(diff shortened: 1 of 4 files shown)


maxnoe:
Timings using a dummy event source only producing new empty `ArrayEventContainer`:

```
❯ python dummy_eventsource.py
100%|██████████████████████████████████████████████████████████████████████████████████████████████| 100000/100000 [00:01<00:00, 75570.20it/s]
❯ git switch master
Switched to branch 'master'
Your branch is up to date with 'origin/master'.
❯ python dummy_eventsource.py
100%|███████████████████████████████████████████████████████████████████████████████████████████████| 100000/100000 [00:16<00:00, 5895.31it/s]

```

maxnoe:
The speedup for the `HDF5EventSource` is also noticeble (~10 %):

```
❯ git switch master
Switched to branch 'master'
Your branch is up to date with 'origin/master'.

❯ ctapipe-process -i /tmp/foo.dl1.h5 -o /tmp/bar.dl2.h5  --overwrite --progress
2022-07-02 13:36:41,405 WARNING [ctapipe.ctapipe-process.DataWriter] (datawriter._setup_output_path): Overwriting /tmp/bar.dl2.h5
HDF5EventSource: 100%|██████████████████| 4529/4529 [00:17<00:00, 261.75ev/s]
❯ git switch default_factories
Switched to branch 'default_factories'
Your branch is up to date with 'origin/default_factories'.
❯ ctapipe-process -i /tmp/foo.dl1.h5 -o /tmp/bar.dl2.h5  --overwrite --progress
2022-07-02 13:37:04,977 WARNING [ctapipe.ctapipe-process.DataWriter] (datawriter._setup_output_path): Overwriting /tmp/bar.dl2.h5
HDF5EventSource: 100%|██████████████████| 4529/4529 [00:15<00:00, 289.49ev/s]
```

maxnoe:
Todo: update container tutorials / docs

maxnoe:
Thinking about it again, I am a bit confused why deepcopy is so slow. The initial object is created once at import time and then deepcopied. Why that is slower than creating the object from scratch, I am not sure. 

We also don't perform copies now on objects provided by default, not default_factory which could (and are in case of quantities) mutable, which could be the reason. But we should take care that something like this does not happen:
This would mutate the default (`nan * u.deg`) for all coming containers:
```
hillas.width[...] = 5 * u.deg
```

Instead 
```
hillas.width = 5 * u.deg
```

But I think this is still a net positive performance / safety tradeoff

kosack (dismissed):

kosack (approved):

ctapipe/core/container.py
```diff
@@ -59,8 +67,10 @@ def __init__(
         ndim=None,
         allow_none=True,
         max_length=None,
+        default_factory=None,
```
nbiederbeck, line 70: since you added `description=` to all containers (in core/container.py), what about using the kwargs-only syntax, to enforce this, here?

nbiederbeck (approved):

PR #1993, opened 2022-07-04
Add flake8 pre-commit hook. Change black hook to also format notebooks.

nbiederbeck:
I like having my code checked by pyflakes/flake8, but I dislike getting told so by our CI.

Personally, I think it is better to have this checked locally, to prevent "fix pyflakes"-commits.

I chose flake8, because it wraps pyflakes among others

--

`black-jupyter` formats python files *and* notebooks

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -1,4 +1,10 @@
 repos:
+  # https://pycqa.github.io/isort/docs/configuration/black_compatibility.html#integration-with-pre-commit
+  - repo: https://github.com/pycqa/isort
+    rev: 5.6.4
+    hooks:
+      - id: isort
+        args: ["--profile", "black", "--filter-files"]
   - repo: https://github.com/psf/black
     rev: 22.6.0
     hooks:
@@ -9,9 +15,3 @@ repos:
     hooks:
       - id: flake8
         args: [--max-line-length=88, --extend-ignore=E203]
-  # https://pycqa.github.io/isort/docs/configuration/black_compatibility.html#integration-with-pre-commit
-  - repo: https://github.com/pycqa/isort
-    rev: 5.6.4
-    hooks:
-      - id: isort
-        args: ["--profile", "black", "--filter-files"]
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -2,4 +2,8 @@ repos:
   - repo: https://github.com/psf/black
     rev: 22.3.0
     hooks:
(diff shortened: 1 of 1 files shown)


kosack:
This means we'll have to fix all the remaining code style warnings (since pycodestyle is run by flake8), but probably that is a good thing!  I just checked and there are not so many (that aren't already disabled in `setup.cfg`)

nbiederbeck:
Yes, but only in files one commits

maxnoe:
> but I dislike getting told so by our CI.

Pyflakes checks for actual errors, not style issues (with the exception of unused imports). This is an important first check in the CI and if that fails, there is no point in running the unit tests in the CI.

Of course it would be best if everyone ran all the checks locally first, but that does not make running it in the CI any less necessary.

nbiederbeck:
**RE pre-commit/CI**: I like having a short/fast feedback loop. Having the CI fail with pyflakes annoys me, and pyflakes is really fast on a local machine (as opposed to the full test-suite). I'm not for replacing, I'm for additional checks that makes the life of developers easier (at least "easier imho").

**RE pyflakes/flake8**: according to pyflakes maintainers (as seen in a [PR I opened](https://github.com/PyCQA/pyflakes/pull/709#issuecomment-1156383725)) the preferred way of using pyflakes as pre-commit hook is via flake8. flake8 is a wrapper around pyflakes and pycodestyle, so we get more than "just" pyflakes.

**RE isort**: I'm in

**RE black compability**: The [black docs](https://black.readthedocs.io/en/stable/guides/using_black_with_other_tools.html) give configurations for other tools, which I added in my last commit.

.pre-commit-config.yaml
```diff
@@ -1,9 +1,17 @@
...
     hooks:
       - id: black-jupyter
+  # https://black.readthedocs.io/en/stable/guides/using_black_with_other_tools.html?highlight=other%20tools#flake8
   - repo: https://github.com/PyCQA/flake8
     rev: 4.0.1
     hooks:
       - id: flake8
+        args: [--max-line-length=88, --extend-ignore=E203]
+  # https://pycqa.github.io/isort/docs/configuration/black_compatibility.html#integration-with-pre-commit
+  - repo: https://github.com/pycqa/isort
```
maxnoe, line 13: Are these run in order? Then it might make sense to run isort then Black, then flake8.

maxnoe (approved):

kosack (approved):

PR #1996, opened 2022-07-05
New 3D shower model 

StFroese:
This is just a draft for now to review some ideas. 

This approach is based on this [paper](https://arxiv.org/pdf/astro-ph/0601373.pdf).

The shower is basically a 3D Gaussian distribution with a width  $\sigma_t $ and length $\sigma_l$.
The orientation is given by the azimuthal and polar angle. Those are used to rotate the covariance matrix.
The barycentre of the shower is calculated using the height of the first interaction in the atmosphere and the length of the shower. The centre is shifted according to the intersection of the shower axis on ground.

The next step is to write an `ImagePredictor` class which will take a subarray, goes through each telescope and integrates the 3D distribution as it can be seen for each pixel in camera. This could be done with scipy's `tplquad` but this is computationally very expensive. A better solution to this is to use the analytical solution provided by the paper (see Appendix).


vuillaut:
Appendix to be found where?

StFroese:
@vuillaut Sry, I meant the appendix of the paper, p.29

vuillaut:
> @vuillaut Sry, I meant the appendix of the paper, p.29

Ha yes thanks !
Please include reference to the paper in the docstring of thhe function as well.

vuillaut:
If reminds me of a forgotten package 😄  
https://github.com/vuillaut/pschitt

ctapipe/image/showermodel.py
```diff
@@ -0,0 +1,102 @@
...
+        first_interaction=u.m,
+        width=u.m,
+        length=u.m,
+    )
+    def __init__(self, n_ch, x, y, phi, theta, first_interaction, width, length):
+        """Create a 3D gaussian shower model for imaging.
+
+        Parameters
+        ----------
+        n_ch : int
```
maxnoe, line 27: While I like sticking to paper notations, here I would go for more talkative variable names...

ctapipe/image/showermodel.py
```diff
@@ -0,0 +1,102 @@
...
+            Number of cherenkov photons in shower
+        x : u.Quantity[length]
+            x coord of shower intersection on ground
+        y : u.Quantity[length]
+            y coord of shower intersection on ground
+        phi : u.Quantity[angle]
+            azimuthal angle defining orientation of shower
+        theta : u.Quantity[angle]
+            polar angle defining orientation of shower
```
maxnoe, line 36: Polar and azimuthal angle is normally the same... Does this mean azimuth and zenith? If yes, you should probably swith to call this `azimuth` and use `altitude` instead of `zenith` to be consistent with the rest of ctapipe.

ctapipe/image/showermodel.py
```diff
@@ -0,0 +1,102 @@
...
+        r = R.from_rotvec([0, self.theta.to_value(u.rad), self.phi.to_value(u.rad)])
+        cov = r.as_matrix().T @ cov @ r.as_matrix()
+
+        gauss = multivariate_normal(
+            mean=self.barycenter.to_value(u.m), cov=cov.to_value(u.m)
+        )
+
+        return self.n_ch * gauss.pdf(np.array([x, y, z]))
+
+    def calcBC(self):
```
kosack, line 73: code style: change to `calculate_barycenter()`  (avoid abbreviations, functions should be snake_case)

kosack: or better yet, just make this a `functools.cached_property` called `barycenter`

ctapipe/image/showermodel.py
StFroese: ok yeah, in spherical coords the azimuthal angle is normally the one that is in the xy-plane and the polar angle is the one that is in the xz-plane. So they are the same as `azimuth` and `zenith`. I wasn't sure about calling them the same here since it is not guaranteed that these values describe the origin of the shower if the intersection with the ground is not at the origin (0,0,0) of this frame.

kosack: You could always just add "relative to the shower core", and then azimuth and zenith are ok.  Though be careful as azimuth may not be defined the same direction or origin as the standard Euler angle phi, which is right-handed and starts on the X-axis.

kosack:
> If reminds me of a forgotten package 😄
> https://github.com/vuillaut/pschitt

I remember remarking once when you presented Pschitt that it could be used for reconstruction!

StFroese:
`Predictor` now generates telescope images for a given shower model and some input parameters for the telescopes/subarray. It's a little bit slow because of some loops (I'm sure this can be done faster :D).
Looks ok for me so far:

Unknown-2


NominalFrame of 4 LSTs:
Unknown-4

StFroese:
I changed the parameter names to more recognizable ones as recommended by Karl. The `_photons` function of the `Predictor` is now vectorized and takes an array of pixels as input :)

ctapipe/image/predictor.py
```diff
@@ -20,7 +20,7 @@ def __init__(
         ----------
         tel_positions : u.Quantity[length]
             Position of the telescopes in the frame of the array
-        tel_pointings : u.Quantity[Angle]
+        tel_pix_coords_altaz : u.Quantity[Angle]
             Pointing of the pixels for each telescope in AltAz
```
maxnoe, line 24: What is the shape here? `[n_telescopes, n_pixels]`? Should be added to the docstring

StFroese: Ah yes sry. I'll change it

StFroese:
We also need to add an energy estimator to the model since we only get the number of photons. The paper recommends obtaining a relation from simulations based on number of photons, the telescope multiplicity, the telescope pointing and the direction of the shower axis. If I understand them correctly they interpolate their data to get a relation $E(N_{photons}, \vec{\theta})$ with $\vec{\theta}$ being the other parameters mentioned.

I think another approach would be to use the EnergyRegressor with the number of photons given as a feature, but I'm not sure. What do you think @kosack @maxnoe ?

kosack:
>  another approach would be to use the EnergyRegressor with the number of photons given as a feature, but I'm not sure. What do you think @kosack @maxnoe ?

It think using an EnergyRegressor is a better idea than using a lookup table as they do. It makes it the same sort of training as for other reconstructions, and we avoid the usual problems with binning and edge cases.

maxnoe:
If we'd really want to use a lookup table, my approach would be to implement one as a custom scikit learn regressor and then we can use the machine learning module structure the same way.

kosack:
> If we'd really want to use a lookup table, my approach would be to implement one as a custom scikit learn regressor and then we can use the machine learning module structure the same way.

This reminds me that we might want to also think about merging the ImPACT models and training with the ML module eventually - they are basically the same thing: a model that predicts an image, and should be trainable using the same mechanism. I'll open an issue to remind us

ctapipe/reco/shower_reconstructor.py
```diff
@@ -0,0 +1,165 @@
...
+            is_valid=True,
+            prefix=self.__class__.__name__,
+        )
+
+    def _fit(self, event):
+        # just call iminiut with likelihood and some seeds
+        # may also need limits
+        seeds = self._seeds(event)
+        minimizer = Minuit(self._likelihood, *seeds)
+        minimizer.limits = [
```
maxnoe, line 68: Hard coded limits. Also, much more readable would be assigning like this:

```
minimizer.limits["x"] = (...,.... )
```

StFroese: Yes I know. This is still work in progress and not working correctly

ctapipe/reco/shower_reconstructor.py
```diff
@@ -0,0 +1,165 @@
...
+from scipy.stats import poisson
+
+from ctapipe.containers import ReconstructedGeometryContainer
+from ctapipe.coordinates import CameraFrame, TelescopeFrame
+from ctapipe.core.traits import Unicode
+from ctapipe.image import GaussianShowermodel, ShowermodelPredictor
+from ctapipe.reco import Reconstructor
+
+
+class ShowerReconstructor(Reconstructor):
```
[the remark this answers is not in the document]
kosack: Following the new scheme, I think `Model3DGeometryReconstuctor` would be correct (and it would then be a subclass of `GeometryReconstructor`.   The word "shower" is not needed, as it is always the case that we are reconstructing showers, so it's redundant (we don't say ShowerHillasReconstructor, for example)

ctapipe/reco/shower_reconstructor.py
kosack: Might want to take a look at ImPACTReconstructor, as it also does likelihood minimization. Perhaps we can extract a common way to do it to avoid repetition.

StFroese:
The reconstructor seems to work now. The images show the reconstruction of a 4 LST event. Upper image is the reconstruction and the lower image is the true one.
Screenshot 2022-10-06 at 17 23 33
Screenshot 2022-10-06 at 17 23 46

Screenshot 2022-10-06 at 17 23 37
Screenshot 2022-10-06 at 17 23 23

ctapipe/reco/shower_reconstructor.py
StFroese: This should be pretty similar now. Generate images -> calculate likelihood -> fit using minuit with seeds from previous reconstructor

maxnoe:
> in the name since it should be able to also reconstruct the energy

I discussed this yesterday with Karl and we came to the conclusion that, no, this does not predict energy. Rather you'd use the number of photons this reconstructor produces as input feature for the machine learning `EnergyRegressor`.

StFroese:
>Rather you'd use the number of photons this reconstructor produces as input feature for the machine learning `EnergyRegressor`.

ok great, but this leaves question where to store this number. The `ReconstructedGeometryContainer` only provides `average_intensity`.

StFroese:
Test now takes ~8s instead of 6min

kosack:
> ok great, but this leaves the question where to store this number. The ReconstructedGeometryContainer only provides average_intensity.

One option is to do what we do with telescopes: you make a subclass of `ReconstructedGeometryContainer` called `Model3DReconstructedGeometryContainer` that adds a field for the photon density, and then you just return that (it will then automatically be written to the output files and can be used as a feature of an ML algorithm).

ctapipe/reco/shower_reconstructor.py
```diff
@@ -39,6 +37,10 @@ def __call__(self, event):
         self.tel_positions = tel_positions
         self.tel_solid_angles = tel_solid_angles
         self.tel_mirror_area = tel_mirror_area
+        self.tel_pix_coords_altaz = self._tel_pix_coords_altaz(event)
+
+        # for likelihood
+        self.DL1CamContainers = event.dl1.tel.items()
```
kosack, line 43: should not use PascalCase here, since it is a variable so pep8 says to use snake_case. (so `self.dl1_camera_containers`

maxnoe: Do we really need to store that as state?

StFroese: Good point. There are two choices I think. Since `likelihood` is only called by Minuit, which only allows `(self, params)` as arguments of the likelihood, it has to either know the dl1 cam containers or the event structure. So its either `self.d1_cam_containers = event.dl1.tel.items()` or `self.event = event`. Which one should I use?

StFroese: Third option would be to just store the dl1 images for the likelihood and the whole container

ctapipe/reco/shower_reconstructor.py
```diff
@@ -0,0 +1,169 @@
...
+            x * u.m,
+            y * u.m,
+            azimuth * u.deg,
+            altitude * u.deg,
+            h_max * u.m,
+            width * u.m,
+            length * u.m,
+        )
+
+        predictor = ShowermodelPredictor(
```
maxnoe, line 106: Why do we create the predictor in each call to the likelihood? That seems expensive!

ctapipe/reco/shower_reconstructor.py
```diff
@@ -0,0 +1,169 @@
...
+        minimizer.errordef = Minuit.LIKELIHOOD
+        minimizer.migrad()
+        fit = minimizer.values
+        fit_errors = minimizer.errors
+
+        return fit, fit_errors
+
+    def _likelihood(self, total_photons, x, y, azimuth, altitude, h_max, width, length):
+
+        model = GaussianShowermodel(
```
maxnoe, line 95: This also seems very expensive, especially the adding of astropy units inside the likelihood.

ctapipe/reco/shower_reconstructor.py
maxnoe: This is a classic case for using a closure, e.g. via partial:


```
def likelihood(..., images):
   ...

m  = Minuit(partial(likelihood, images=mages), ...)
```

ctapipe/reco/shower_reconstructor.py
```diff
@@ -0,0 +1,169 @@
...
+            showermodel=model,
+        )
+
+        prediction = predictor.generate_images()
+
+        log_likelihood = 0
+        for tel_id, DL1CamContainer in self.DL1CamContainers:
+            log_likelihood += np.sum(
+                neg_log_likelihood_approx(
+                    DL1CamContainer.image, prediction[tel_id], 0.5, 2.8
```
maxnoe, line 120: magic numbers?

StFroese: Ah yes. Those are hard coded values taken from the ImPACT reco for the pedestal photon width. The pedestal info is not in this event structure for the test but should be there in real measurements. So I maybe have to to something like conditional thing if this is real data or not

ctapipe/reco/shower_reconstructor.py
StFroese: Ah great, I didn't know this. I'll change it

maxnoe:
This should get also much faster if instead of using mappings of single numpy arrays, you put everything upfront into unit-less numpy array of shape `(n_telescopes, ...)`, e.g. have one array for images and predictions of shape `(n_telescopes, n_pixels)`.

This should even broadcast (or if not we should implement that) with the `pixel_likelihood` function. Having the likelihood perform well is key in these fits and that means usually:

* units should not be used inside. Transform once upfront into well-known units and re-attach units to the results afterwards.
* use arrays of packed information instead of mappings, what I wrote above. E.g. see this PR for the changes made to the other recnostructor: https://github.com/cta-observatory/ctapipe/pull/2036

ctapipe/reco/shower_reconstructor.py
StFroese: Yes, astropy units are the biggest time consuming thing right now (I did profiling). I guess I have drop all units in the model and in the predictor.

ctapipe/image/showermodel_predictor.py
```diff
@@ -0,0 +1,148 @@
+import numpy as np
+from erfa.ufunc import s2p as spherical_to_cartesian
+
+__all__ = ["ShowermodelPredictor"]
+
+
+class ShowermodelPredictor:
+    def __init__(
```
maxnoe, line 8: This should just get a `SubarrayDescription` I think and construct the things from there

ctapipe/reco/shower_reconstructor.py
```diff
@@ -0,0 +1,208 @@
...
+
+    def __init__(self, subarray, **kwargs):
+        super().__init__(subarray=subarray, **kwargs)
+
+    def __call__(self, event):
+        if self.geometry_seed not in event.dl2.stereo.geometry:
+            raise ValueError()
+
+        # get all telescope properties that we need for the ShowermodelPredictor
+        tel_positions = {}
```
maxnoe, line 30: As said above, give the subarray to the `Predictor` and move that to `__init__`. It doesn't change so shouldn't be called in the event loop.

ctapipe/reco/shower_reconstructor.py
```diff
@@ -0,0 +1,208 @@
...
+
+class Model3DGeometryReconstructor(Reconstructor):
+    geometry_seed = Unicode(default_value="HillasReconstructor").tag(config=True)
+
+    def __init__(self, subarray, **kwargs):
+        super().__init__(subarray=subarray, **kwargs)
+
+    def __call__(self, event):
+        if self.geometry_seed not in event.dl2.stereo.geometry:
+            raise ValueError()
```
maxnoe, line 27: Meaning full error please. Also this should be a `KeyError`

ctapipe/reco/shower_reconstructor.py
```diff
@@ -0,0 +1,208 @@
...
+            ].optics.mirror_area.to_value(u.m**2)
+            tel_spe_widths[tel_id] = (
+                spe
+                if (spe := event.mon.tel[tel_id].flatfield.charge_std) is not None
+                else 0.5
+            )
+            tel_pedestial_widths[tel_id] = (
+                ped
+                if (ped := event.mon.tel[tel_id].pedestal.charge_std) is not None
+                else 2.8
```
maxnoe, line 52: magic number

ctapipe/reco/shower_reconstructor.py
```diff
@@ -0,0 +1,208 @@
...
+            tel_solid_angles[tel_id] = geometry.transform_to(
+                TelescopeFrame()
+            ).pix_area.to_value(u.rad**2)
+            tel_mirror_area[tel_id] = self.subarray.tel[
+                tel_id
+            ].optics.mirror_area.to_value(u.m**2)
+            tel_spe_widths[tel_id] = (
+                spe
+                if (spe := event.mon.tel[tel_id].flatfield.charge_std) is not None
+                else 0.5
```
maxnoe, line 47: magic number

StFroese:
So for now the big time consuming parts remaining are the `neg_log_likelihood_approx`, `emission_probability`, `np.clip`, `spherical_to_cartesian` (inside _vec_los).
Nevertheless it's 5-6 times faster now.

@maxnoe I'm not sure how to construct something like an image numpy array `(n_telescopes, n_pixels)` since the cameras can have a different number pixels

ctapipe/reco/shower_reconstructor.py
StFroese: So the problem is that the simulations do not contain the flatfield and pedestal charge information. I need some values in case they are not provided and I don't have a clue which numbers are suitable. I think this would also vary with telescope type. It's also part of the impact code https://github.com/cta-observatory/ctapipe/blob/76c672ed2f8e66b0a31bae8bd53350c1a81c55a5/ctapipe/reco/impact.py#L120

ctapipe/reco/shower_reconstructor.py
```diff
@@ -15,57 +13,49 @@
 )
 from ctapipe.reco.reco_algorithms import Reconstructor
 
+# taken from ImPACT PR for now
+ped_table = {
```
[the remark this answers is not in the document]
maxnoe: Hard coding values makes no sense, its completely dependent on the used extraction.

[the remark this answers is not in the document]
StFroese: Changed it :)

ctapipe/reco/shower_reconstructor.py
```diff
@@ -5,28 +5,34 @@
...
+    ).tag(config=True)
+
+    pedestal_width = FloatTelescopeParameter(
+        default_value=[
+            ["type", "LST_LST_LSTCam", 1.4],
+            ["type", "MST_MST_FlashCam", 1.3],
+            ["type", "MST_MST_NectarCam", 1.3],
+            ["type", "SST_ASTRI_CHEC", 0.5],
+        ],
+        help="Pedestal charge std",
```
maxnoe, line 25: call the parameter `pedestal_charge_std` if that is what it is.

StFroese: yes can do

StFroese: I had to remove the default values per type. This fails for a subarray without SSTs for example

StFroese: It's only one value now

StFroese:
You can take a look at some images like this now:
```python
from ctapipe.reco import HillasReconstructor, Model3DGeometryReconstructor


source = EventSource("/Users/stefan/Downloads/gamma_20deg_180deg_run3___cta-prod6-paranal-2147m-Paranal-dark.dl2.h5", max_events=10)
events = list(source)

hillas_reco = HillasReconstructor(source.subarray)
shower_reco = Model3DGeometryReconstructor(source.subarray)

for event in events:
    hillas_reco(event)
    shower_reco(event)
    shower_reco.peek(event)
```

kosack (dismissed):

maxnoe:
There are conflicts that need to be resolved, please rebase / merge the current main

PR #1998, opened 2022-07-07
Remove restriction on existing file in `EventSource.input_url`, fixes #1821

### LINKED ISSUE #1821: Multiple files as input of EventSource
Currently, giving more than one file as `input_url` to EventSource (specifically I am using the `MAGICEventSource` from https://github.com/cta-observatory/ctapipe_io_magic) e.g. passing a string with wildcards, throws an error (a `TraitError` to be precise).

Is there a way, if the logic allows, to pass more files at a time? The issue I have is that in MAGIC data files (at calibrated level; one file per subrun, and each run has multiple subruns) the pointing information from the drive reports are stored in the same file as the events, and sometimes happens that a given subrun has only one drive report available (this can happen for different reasons e.g. the subrun is too short, usually they are ~1min long and have ~7-8 reports available). This prevents to interpolate the pointing position of the events starting from the drive reports.

Therefore, if I could be able to pass all the subruns belonging to the same run to the `EventSource`, this problem would be no longer there. Also, this would improve the interpolation, since performing the interpolation subrun by subrun will result in extrapolation at the edges of each subrun.

Have you any suggestion on how to do this without introducing changes to the `EventSource` class? Or do you suggest to adopt a solution similar to `ctapipe_io_lst` i.e. having the possibility to give as input a file containing the pointing information from the drive reports? In the latter case it would mean for us to have a pre-processing of the files before the stage1 process (MAGIC calibrated-->DL1) to get the drive reports and create this file to give as input to the `MAGICEventSource`.

Thanks!

maxnoe:
fixes #1821

### DIFF
--- a/ctapipe/io/eventsource.py
+++ b/ctapipe/io/eventsource.py
@@ -1,17 +1,18 @@
 """
 Handles reading of different event/waveform containing files
 """
+import warnings
 from abc import abstractmethod
-from traitlets.config.loader import LazyConfigValue
-from typing import Tuple, List, Generator
+from typing import Generator, List, Tuple
 
+from traitlets.config.loader import LazyConfigValue
+
+from ..containers import ArrayEventContainer
+from ..core import Provenance, ToolConfigurationError
+from ..core.component import Component, find_config_in_hierarchy, non_abstract_children
+from ..core.traits import CInt, Int, Path, Set, TraitError, Undefined
 from ..instrument import SubarrayDescription
 from .datalevels import DataLevel
-from ..containers import ArrayEventContainer
-from ..core import ToolConfigurationError, Provenance
-from ..core.component import Component, non_abstract_children, find_config_in_hierarchy
-from ..core.traits import Path, Int, CInt, Set, Undefined
-
 
 __all__ = ["EventSource"]
 
@@ -83,11 +84,7 @@ class EventSource(Component):
         generated events. If None, all available telescopes are used.
     """
 
-    input_url = Path(
-        directory_ok=False,
-        exists=True,
-        help="Path to the input file containing events.",
-    ).tag(config=True)
+    input_url = Path(help="Path to the input file containing events.").tag(config=True)
 
     max_events = Int(
         None,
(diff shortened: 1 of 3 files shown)


ctapipe/io/eventsource.py
```diff
@@ -298,8 +295,18 @@ def _find_compatible_source(cls, input_url):
...
+                if subcls.is_compatible(input_url):
+                    return subcls
+            except Exception as e:
+                warnings.warn(f"{subcls.__name__}.is_compatible raised exception: {e}")
+
+        # provide a more helpful error for non-existing input_url
+        if not input_url.exists():
+            raise TraitError(
+                f"input_url {input_url} is not an existing file "
+                " and no EventSource implementation claimed compatibility"
```
nbiederbeck, line 308: I think this error would benefit from naming the available eventsources like the error below

nbiederbeck (changes requested):

ctapipe/io/eventsource.py
maxnoe: I explicitly left it out because it is right now kind of assumed that for a non-existing file, the error is a typo / wrong path and people should see that immediately and not be confused by a long list of tried sources.

But I don't have a strong opinion here.

nbiederbeck: I agree

nbiederbeck (approved):

ctapipe/io/hdf5eventsource.py
```diff
@@ -183,21 +184,25 @@ def close(self):
...
+        with tables.open_file(path) as f:
             metadata = f.root._v_attrs
             if "CTA PRODUCT DATA LEVELS" not in metadata._v_attrnames:
                 return False
 
-            # we can now read both R1 and DL1
             datalevels = set(metadata["CTA PRODUCT DATA LEVELS"].split(","))
-            if not datalevels.intersection(("R1", "DL1_IMAGES", "DL1_PARAMETERS")):
+            if not datalevels.intersection(
+                ("R1", "DL1_IMAGES", "DL1_PARAMETERS", "DL2")
```
LukasNickel, line 204: Not really related, but I guess it was missed before. Fine for me to add the DL2 here

LukasNickel (approved):

PR #2000, opened 2022-07-08
Atmosphere models for h_max to X_max conversion

kosack:
### Features

Adds an `AtmosphereDensityProfile` class hierarchy with three implementations:
- `ExponentialAtmosphereDensityProfile`  (purely mathematical)
- `TableAtmsophereDensityProfile` (cubic interpolation of a tabulated model)
- `FiveLayerAtmosphereDensityProfile`  (implements the Corsika 5-layer representation where each layer is either an exponential or linear model)

This adds a new `atmosphere_density_profile`  attribute to EventSource, with an implementation in `SimTelEventSource` (reading them from simtel files) and `HDF5EventSource`, as well as in `DataWriter` (propagating them to output files).

By default SimTelEventSource will return a TableAtmosphereDensityProfile if it is found in the data, otherwise it falls back to a FiveLayerAtmosphereDensityProfile, or None if no profile exists.  This is user-selectable with the `SimTelEventSource.atmosphere_model_choice` option (AUTO, TABLE, FIVELAYER)

### Example:

```py3
from ctapipe.atmosphere import ExponentialAtmosphereDensityProfile, TableAtmosphereDensityProfile

# basic profile, using default parameters
print(density)
print("Density:       ", density(10*u.km))
print("Column Density (0d):", density.line_of_sight_integral(distance=10*u.km, zenith_angle=0*u.deg))
print("Column Density (45d):", density.line_of_sight_integral(distance=10*u.km, zenith_angle=45*u.deg))
```
```
ExponentialAtmosphereDensityProfile(h0=<Quantity 8. km>, rho0=<Quantity 0.00125 g / cm3>)
Density:        0.0003581309960752376 g / cm3
Column Density (0d): 286.50479686019 g / cm2
Column Density (45d): 584.3180219705157 g / cm2
```

Loading one from a simtel or ctapipe file can be done with an EventSource:
```py3
# Table profile loaded from a SimTel file
with EventSource("gammas.simtel.zst", atmosphere_profile_choice="TABLE") as source:
     density_tab = source.atmosphere_density_profile
print(density_tab)
```

```
TableAtmosphereDensityProfile(meta={'obs_level': <Quantity 215800. cm>, 'atmo_id': 99, 
    'atmo_name': b'atmprof_ecmwf_north_winter_fixed.dat', 'htoa': 12000000.0}, rows=55)
```

All profiles can also be plotted:
```py3
density.peek()
density_tab.peek()
```
image


ctapipe/atmosphere/model.py
```diff
@@ -0,0 +1,250 @@
...
+    Returns
+    -------
+    Table:
+        table with columns `height`, `density`, and `column_density`
+        along with associated metadata
+    """
+    import eventio
+
+    tables = []
+    with eventio.SimTelFile(simtelfile) as simtel:
```
maxnoe, line 230: Shouldn't this just be in a `SimTelEventSource.atmospheric_profiles` attribute?

ctapipe/atmosphere/__init__.py
```diff
@@ -0,0 +1 @@
+from .model import *
```
maxnoe, line 1: Start imports and no `__all__` mess with sphinx, please import explicitly and add an `__all__`

ctapipe/atmosphere/model.py
maxnoe: This is also a recent addition in the mc files, so it has to anyways handle the case that non are available.

try_obsconfig_model.ipynb
```diff
@@ -0,0 +1,622 @@
+{
+ "cells": [
```
maxnoe, line 2: Accidental commit?

kosack: strange - I don't know how that got committed...

docs/ctapipe_api/atmosphere/index.rst
```diff
@@ -0,0 +1,31 @@
...
+
+    density_profile = ExponentialAtmosphereDensityProfile(
+          h0 = 7.8 * u.m
+          rho0 = 0.000125 * u.g/u.cm**3
+    )
+
+    # convert a h_max to an x_max:
+
+    h_max = 8 * u.km
+    x_max = density_profile.integral(h_max)  # in g/cm2 units
```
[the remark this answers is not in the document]
kosack: I haven't updated this documentation to reflect the latest version. Will do though

ctapipe/atmosphere/model.py
maxnoe: Indeed, there *should* only be one profile and I only see one in the simtel files.

The read logic in pyeventio just parses the objects and if it encounters an `AtmosphericProfile` object, it appends to the list.

If we are sure, there can only be one, we can change the `SimTelFile` in eventio.

kbernloehr:
Hi Max,

I am not aware of simtel files from a single run with more than
an I/O block of type 1216 - if you found such a case, let me
know how it was produced. If they are identical, using the
first one is OK but there are situations possible where a
first one might just contain part of the information and
might later get complemented by a second such block
with the other part. Typical CORSIKA simulation should
fill both parts right away though.

So, well, there are two parts of content in this data block.
One part, which you probably want to convert, is the copy
of the atmprof*.dat file read from CORSIKA. The other part
is the result of the 5-layer fit used by the CORSIKA EGS part.
The code running with CORSIKA should report both parts,
if available, but you might end up with just one part.
The table part would be missing if you run corsika without
an atmospheric profile table. The other ways to configure
the atmospheric profiles in CORSIKA only set the 5-layer
parameters. In that case there is no table to report.
The thickc() and heighc() functions in mc_atmprof.c are
C implementations of what CORSIKA is doing internally
with the 5-layer atmosphere models.

Too bad that EGS part is hard-coded to use the 5-layer
function but it cannot accurately describe the actual
atmosphere. If you want to lookup some height from an
atmospheric depth in the longitudinal profiles, you first
need to make up your mind which description is the more
appropriate one for the task. Since most Cherenkov light
gets emitted from e+-, the 5-layer description might actually
be closer to what CORSIKA originally generated. Xmax is a
mixed beast, at least for hadron showers, so there would be
no really correct way. Hadrons and muons follow directly
the table profile, if any got set up and you don't run a
CURVED option CORSIKA.

Note that interpolation of the tabulated profiles is quite non-trivial.
In particular, you want to have height(thick(h)) == h, at an
acceptable accuracy.
Linear interpolation in a mostly exponential density profile
will obviously give wrong answers, even though that looks
reversible. The atmo.c code goes through some effort to
cspline interpolate in log(thick) versus height - except at the
top where thick=0 is not suitable for the log while csplines
are not very forgiving for any sudden transitions (jumps in
second derivative). The reversion through the heighx_ and
thickx_ functions in atmo.c are good to about 0.2 mm throughout
the whole atmosphere while direct interpolation methods
might easily miss by a 100 meters.

   Best regards,
     Konrad

Am 11.07.22 um 18:08 schrieb Maximilian Nöthe:
> @maxnoe commented on this pull request.
>
>
>
>> +
> +
> +def read_simtel_profile(simtelfile: str) -> Table:
> +    """
> +    Read an atmosphere profile from a SimTelArray file as an astropy Table
> +
> +    Returns
> +    -------
> +    Table:
> +        table with columns `height`, `density`, and `column_density`
> +        along with associated metadata
> +    """
> +    import eventio
> +
> +    tables = []
> +    with eventio.SimTelFile(simtelfile) as simtel:
>
> @kbernloehr Any reason why there could be more than one atmospheric profile in a simtel file?
>

kosack:
Thanks.  As it's not intended for simulation, the use case for this package probably doesn't require extreme precision in the atmosphere profile or it's interpolation as other systematics dominate, but in any case I'll switch to a log  interpolation and use splines.  I'll also add a disclaimer in the documentation!

kbernloehr:
Hi Karl,

I agree that for Xmax lookup you can be a bit more
relaxed. You definitely want to leave out the last point
(thick=0) but perhaps even the second to last point
to avoid wiggles in the csplines. The shower maximum
will never be at these altitudes.

And, even though all of our available productions used
tabulated atmospheres, you might want to be prepared
for those where the atmospheric profile is defined by other
means (switch to 5-layer description).

    Best,
     Konrad

Am 12.07.22 um 10:11 schrieb Karl Kosack:
> Thanks.  As it's not intended for simulation, the use case for this package probably doesn't require extreme precision in the atmosphere profile or it's interpolation as other systematics dominate, but in any case I'll switch to a log  interpolation and use splines.  I'll also add a disclaimer in the documentation!
>

maxnoe:
> means (switch to 5-layer description).

Do you want to add a class for that description as well, @kosack ? As a fallback if the tabulated version is not available?

kosack:
Interpolation in the wiggly areas looks like this now (with cubic splines used):

image

kosack:
>> means (switch to 5-layer description).

> Do you want to add a class for that description as well, @kosack ? As a fallback if the tabulated version is not available?

Is that documented somewhere and do we have a file that has that format atmosphere? If so, I can add it, or  we could always add a subclass in a later version

kosack:
I added a `FiveLayer` profile. Not sure I did it correctly, but at least the column density profile looks ok.

kosack:
The last item to implement here is the i/o.  Since there could in principle be more than one version of the atmosphere profile in the file (one from the simulation, and one from real measurements for observation data), I think the ones loaded by the SimTelEventSource will go in */simulation/service/atmosphere_profile*, which then allows one to add a */dl2/service/atmosphere_profile* later for the frequently updated/measured one.

kosack:
Strange, I get nearly, but not exactly the same result as @kbernloehr  using the data above:

Thickness is identical:
image

Rho is nearly identical, except at one transition 
image

ctapipe/atmosphere.py
```diff
@@ -0,0 +1,389 @@
...
+
+    @abc.abstractmethod
+    def __call__(self, height: u.Quantity) -> u.Quantity:
+        """
+        Returns
+        -------
+        u.Quantity["g cm-3"]
+            the density at height h
+        """
+        raise NotImplementedError()
```
maxnoe, line 43: not really needed, this will never be called. Just having the docstring is fine (and won't result in codecov complaining)

ctapipe/atmosphere.py
```diff
@@ -0,0 +1,389 @@
...
+
+        .. math:: X(h) = \int_{h}^{\infty} \rho(h') dh'
+
+        Returns
+        -------
+        u.Quantity["g/cm2"]:
+            Integral of the density from height h to infinity
+
+        """
+        raise NotImplementedError()
```
maxnoe, line 58: same as above

ctapipe/atmosphere.py
```diff
@@ -0,0 +1,389 @@
...
+        axis[2].grid(True)
+
+        plt.show()
+
+    @classmethod
+    def from_table(cls, table: Table):
+        """return a subclass of AtmosphereDensityProfile from a serialized
+        table"""
+
+        if "TAB_TYPE" not in table.meta:
```
maxnoe, line 135: We should also add a `TAB_VER`, right?

maxnoe: and check for supported versions as with the instrument tables

maxnoe (changes requested):
I think it's important to also have a table versioning here.

ctapipe/atmosphere.py
```diff
@@ -26,8 +26,12 @@
     "FiveLayerAtmosphereDensityProfile",
 ]
 
+SUPPORTED_TABLE_VERSIONS = [
```
maxnoe, line 29: This should be a set or better even frozenset?

ctapipe/atmosphere.py
```diff
@@ -0,0 +1,397 @@
...
+
+        axis[2].legend(loc="best")
+        axis[2].set_xlabel(
+            f"Zenith Angle $\\Psi$ / {zenith_angle.unit.to_string('latex')}"
+        )
+        axis[2].set_ylabel(f"Column Density / {column_density.unit.to_string('latex')}")
+        axis[2].set_yscale("log")
+        axis[2].grid(True)
+
+        plt.show()
```
maxnoe, line 130: Return the fig, so people can actually save the plot?

ctapipe/atmosphere.py
```diff
@@ -0,0 +1,397 @@
...
+    def from_table(cls, table: Table):
+        """return a subclass of AtmosphereDensityProfile from a serialized
+        table"""
+
+        if "TAB_TYPE" not in table.meta:
+            raise ValueError("expected a TAB_TYPE metadata field")
+
+        if (
+            "TAB_VER" not in table.meta
+            or table.meta["TAB_VER"] not in SUPPORTED_TABLE_VERSIONS
```
maxnoe, line 142: could just be `table.meta.get("TAB_VER", "") not in SUPPORTED_TABLE_VERSIONS`

ctapipe/io/datawriter.py
```diff
@@ -797,3 +801,24 @@ def _write_context_metadata_headers(self):
...
+        Parameters
+        ----------
+        path: str
+            path in the HDF5 file where to place the profile
+
+        """
+
+        profile = self.event_source.atmosphere_density_profile
+
+        if profile and hasattr(profile, "table"):
```
maxnoe, line 818: This just silently fails does not write out the  `profile is not None` but `not hasattr(profile, "table")`

maxnoe (changes requested):

ctapipe/io/datawriter.py
kosack: And only profiles with tables  can be written here (e.g. you can't write an ExponentialAtmosphereDensityProfile currently, though I suppose that could be implemented in the future)

kosack: I suppose a warning that the type of profile is not serializable could be provided.

maxnoe: The exponential profile is a table also, right? A table of parameters

maxnoe (approved):
Sorry, didn't see that the `FiveLayerAtmosphere` also has `.table`.

ctapipe/io/simteleventsource.py
```diff
@@ -274,6 +288,113 @@ def apply_simtel_r1_calibration(
...
+        ):
+            return []
+
+        for atmo in simtel.atmospheric_profiles:
+
+            metadata = dict(
+                obs_level=atmo["obslevel"] * u.cm,
+                atmo_id=atmo["id"],
+                atmo_name=atmo["name"],
+                htoa=atmo["htoa"],  # what is this?,
```
nbiederbeck, line 362: We should answer that question

ctapipe/io/tests/test_simteleventsource.py
```diff
@@ -1,3 +1,5 @@
+""" tests of SimTelEventSource """
+# pylint: disable=import-outside-toplevel
```
nbiederbeck, line 2: For another PR: can we disable import-outside-toplevel for all test files?

ctapipe/io/tests/test_simteleventsource.py
```diff
@@ -517,6 +524,46 @@ def test_simtel_no_metadata(monkeypatch):
...
+def test_load_atmosphere_from_simtel(prod5_gamma_simtel_path):
+    """
+    Load atmosphere from a SimTelEventSource
+    """
+    from ctapipe.atmosphere import (
+        FiveLayerAtmosphereDensityProfile,
+        TableAtmosphereDensityProfile,
+    )
+
+    # old simtel files don't have the profile in them, so a null list should be returned
```
nbiederbeck, line 536: This comment is wrong here? Below is correct.

ctapipe/io/tests/test_simteleventsource.py
kosack: I haven't found a way to filter by filename, but it would be nice!  We could just disable it globally since it's used sparingly and correctly for the most part.

ctapipe/io/simteleventsource.py
kosack: Ah, I found it I think:

```cpp
corsika/atmo.c:
2588:   double htoa = 113e5;  /* Height at top of atmosphere */
```
at least in the version I have downloaded.  I'll update it in the file

ctapipe/io/tests/test_simteleventsource.py
kosack: Yup, cut-and-paste error I think! Fixed.

nbiederbeck (approved):

maxnoe (approved):

PR #2007, opened 2022-07-12
ml join apply tools

nbiederbeck:
This targets the machine learning `ml` branch.

Changes:
1. Copy the whole file `input_url` to `output_path`
2. Join the apply tools into a single tool. This tool loads the given model, checks whether it is a regressor or a classifier model, and instantiates the corresponding `Reconstructor`.

Comments:
1. I'm not really happy with the way I decide what kind of Reconstructor is needed
2. I think I would enjoy having a model_path *list*, and apply the models one after the other. Currently, I need to run the tool more than once


maxnoe:
@nbiederbeck I'd simply add (for now) two traitlets `energy_regressor_path` and `particle_id_classifier` that are both optional, so one can apply either one or both models at the same time.

ctapipe/ml/tools/apply.py
```diff
@@ -0,0 +1,255 @@
...
+                )
+            tables.append(table)
+
+        if len(tables) == 0:
+            raise ValueError("No predictions made for any telescope")
+
+        mono_predictions = vstack(tables)
+        if self.apply_regressor:
+            stereo_predictions = self.regressor_combine.predict(mono_predictions)
+            trafo = TelListToMaskTransform(self.loader.subarray)
```
maxnoe, line 215: Probably we should add a `transformations=dict[str, Transform]` option to `write_table`, as it needs to include the metadata correctly (not for the tellist transform, but in general).

maxnoe:
Need to adapt the tests as well, right?

nbiederbeck:
We can set the `combine_property` via a config file. I think the `StereoCombiner` needs to sit a level deeper now?

ctapipe/ml/tools/apply.py
```diff
@@ -0,0 +1,255 @@
...
+                parent=self,
+            )
+            self.regressor_combine = StereoCombiner.from_name(
+                self.stereo_combiner_type,
+                combine_property="energy",
+                algorithm=self.regressor.model.model_cls,
+                parent=self,
+            )
+            self.apply_regressor = True
+        else:
```
maxnoe, line 108: Ah bit nice code is to have:

```
self.apply_regressor = False
self.apply_classifier = False
```
at the top and then get rid of the else branch.

maxnoe:
This is failing due to the changes in prefix handling. Otherwise this looks good. I'll merge and fix / rebase on the ml branch.

PR #2014, opened 2022-07-14
ExpressionEngine for QualityQuery and FeatureGenerator

nbiederbeck:
Breakout the logic of the expression handling/compiling to be used from the QualityQuery (current) and a FeatureGenerator (new, for the ML functionality).

### DIFF
--- a/ctapipe/core/feature_generator.py
+++ b/ctapipe/core/feature_generator.py
@@ -1,0 +1,30 @@
+from .component import Component
+from .expression_engine import ALLOWED_GLOBALS, ExpressionEngine
+from .traits import List
+
+
+class FeatureGeneratorException(TypeError):
+    """Signal a problem with a user-defined selection criteria function"""
+
+
+class FeatureGenerator(Component):
+    features = List().tag(config=True)
+
+    def __init__(self, config=None, parent=None, **kwargs):
+        super().__init__(config=config, parent=parent, **kwargs)
+
+        self.expression_engine = ExpressionEngine(
+            parent=self, expressions=self.features
+        )
+        self.expressions = self.expression_engine()
+        self.feature_names = [n for n, _ in self.features]
+
+    def __call__(self, table):
+        for e, n in zip(self.expressions, self.feature_names):
+            if n in table.colnames:
+                raise FeatureGeneratorException(f"{n} is already a column of table.")
+            try:
+                table[n] = eval(e, ALLOWED_GLOBALS, table)
+            except Exception:
+                raise FeatureGeneratorException(f"{e} is already a column of table.")
+        return table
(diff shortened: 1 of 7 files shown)


ctapipe/core/feature_generator.py
```diff
@@ -0,0 +1,30 @@
...
+from .expression_engine import ALLOWED_GLOBALS, ExpressionEngine
+from .traits import List
+
+
+class FeatureGeneratorException(TypeError):
+    """Signal a problem with a user-defined selection criteria function"""
+
+
+class FeatureGenerator(Component):
+    features = List().tag(config=True)
```
maxnoe, line 11: I think this should be a dict mapping feature names to expressions.

nbiederbeck: Yes, that would fit better. What do you think of using a dict also in the QualityQuery? Now that we removed the possibility to use `lambda` in the expressions, we can also change the type.

maxnoe: We might depend on order the order of the checks when merging and I am not sure everything we use guarantees dict order (python does, but e.g. json and toml don't, so their python parsers also might not).

nbiederbeck: But this is also the case for here, right? Generating a feature from another generated feature

maxnoe: That would be solvable  though by checking the dependencies, which would be a  lot nicer than requiring the user to make it in order anyways.

ctapipe/core/tests/test_feature_generator.py
```diff
@@ -51,3 +51,19 @@ def test_missing_colname():
...
         generator(table)
+
+
+def test_to_unit():
+    """Test chaning the unit of a feature"""
+    from astropy import units as u
+
+    expressions = {
+        "length_meter": "length.to(u.m)",
+        "log_length_meter": "log10(length.to(u.m).value)",
```
maxnoe, line 62: to_value

nbiederbeck: Not possible on Columns:
```python
>>> from astropy import units as u
>>> from astropy.table import Table

>>> table = Table({'length': u.Quantity([1, 2, 3], unit=u.m)})
>>> table['length'].to_value(u.km)

  File "column_to_value.py", line 6, in <module>
    table['length'].to_value(u.km)
AttributeError: 'Column' object has no attribute 'to_value'
```

maxnoe: .quantity.to_value

nbiederbeck: Would be possible on `QTable` though, but this is not what we use internally

nbiederbeck: thanks

ctapipe/core/feature_generator.py
kosack: It was a list originally because the order does in fact matter here, at least if you want useful statistics you want to be able to specify the order in which to apply the quality checks.

maxnoe (approved):

kosack (approved):

Issue #1037, opened 2019-04-03
rethink calibration and image module coupling


kosack:
Right now there are some dependency loops between `calib` and `image`.  This means we might need to rethink the structure:

* **calib**: is right now intended to be the calculation and application of calibration coefficients, and things like gain-channel selection. It mixes right now *production* of calibration coefficients, and *application*, so it could be split up.  Eventually this might also include pointing corrections, etc, so perhaps the module scope is too large.  It also currently includes the classes that perform trace-integration, but as many have complained, trace integration is *not* a calibration, so it should not go in here. 
* **image**: is anything that processes "images" in the general sense of a 3D or 2D Cherenkov camera data. (3D means space + time, e.g. waveforms for each pixel).  This includes charge extraction (e.g. integration) and de-noising (cleaning), which may be quite closely coupled in more advanced techniques. 

Both of these may be too broad.

kosack:
see also #1035

watsonjj:
Regarding your point on #1035:

> In any case, we need to write down the calibration and integration steps in a clear way somewhere as well, to be clear in what order we should compute and apply calibrations, gain-channel selections, trace integration + denoising

This is my current understanding of the order:

1. waveform calibration 
2. gain-channel selection
3. data volume reduction
4. charge extraction from waveforms (including any filtering or waveform cleaning)
5. Application of calibration of extracted charge into p.e. (or photons)
6. Image cleaning (e.g. tailcuts)

kosack:
where does the application of pedestals, gain-correction, and flatfielding come into that list?

kosack:
So perhaps the calib module can be broken into smaller sub-modules then (either under calib/, or at the top-level)

kosack:
Again, we need to make a dependency diagram or something, since a few of those steps can be iterative and use parts of other modules

watsonjj:
Although the charge extraction/trace integration is considered somewhat separate to the calibration...this seperation is potentially a problem for some calibrations. 

The more commonly known calibration is the pedestal subtraction, flat-field and gain-correction (i.e. conversion into p.e.). This can be applied to the charge after it has been extracted from the waveform. Therefore there is no problem here with charge extraction being decoupled from calibration.

[...]

PR #2023, opened 2022-07-15
Add abstract InvalidPixelHandler and NeighborAverage implementation, apply in CameraCalibrator and TwoPassWindowSum. Pass invalid pixels to extractors.

maxnoe:
* Fill pixel_status in SimTelEventSource
* Ignore broken pixels in GlobalPeakWindowSum and NeighborPeakWindowSum extractors
* Interpolate images and peak_time of broken pixels in `CameraCalibrator` using a simple average of their neighbors


maxnoe:
Before:

before

After:
after

kosack (approved):
Looks like a good start.

This does bring up the fact that we need to discuss how broken pixels will be computed in the final version, particularly those due to high NSB that are determined from high pedestal variances.  Those are normally computed *after* charge extraction, but before parameterization.  But that can come later - just having the mechanism here is enough for now!

setup.cfg
```diff
@@ -38,7 +38,7 @@ exclude=
 max-line-length=90
 #ignore=W291,E303,W391,F403,F401,W503,W1202
 select = C,E,F,W,B,B950
-ignore = E501,W503,E203,W201
+ignore = E501,W503,E203,W201,E712
```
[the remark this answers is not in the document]
maxnoe: Yes, because it complains about numpy arrays being compared to True or False elementwise and made it impossible for me to check in already existing code in a completely unrelated code area.

maxnoe:
> Those are normally computed after charge extraction, but before parameterization.

Those are computed from interleaved pedestals, and the information about current interleaved pedestal monitoring will be available in the current physics event exactly at this position (pedestal_failing_pixels), right?

kosack:
> Ignore broken pixels in GlobalPeakWindowSum and NeighborPeakWindowSum extractors

Looks like also TwoPassWindowSum as well

kosack:
> Those are computed from interleaved pedestals, and the information about current interleaved pedestal monitoring will be available in the current physics event exactly at this position (pedestal_failing_pixels), right?

Yes, I would imagine that was the intent there. Maybe the name is not great (it sounds more like pixels where a pedestal could not be computed, rather than where the pedestal variance is too high or too low).

maxnoe:
> Looks like also TwoPassWindowSum as well

Not exactly. The `broken_pixels` argument had to be added to all extractors, because they share a common API. But only the two extractors mentioned make actual use of this information.

setup.cfg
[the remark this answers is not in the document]
maxnoe: I will try to fix it in another PR (there is no real point in doing this for numpy arrays either, better use `np.any` or `np.all` or whatever is equivalent.

But I hate that these pre-commit hooks force you to do changes in unrelated sections of the files. They should only look at the diff, not at the whole file...

kosack:
> Not exactly. The broken_pixels argument had to be added to all extractors, because they share a common API. But only the two extractors mentioned make actual use of this information.

Then shouldn't TwoPass also use it? In computing the time gradient in the first pass, broken pixels should be ignored.

maxnoe:
> Then shouldn't TwoPass also use it? In computing the time gradient in the first pass, broken pixels should be ignored.

Ignored? Or should I add the interpolation?

kosack (changes requested):
> Ignored? Or should I add the interpolation? (twice?)
>I agree that TwoPass should handle it, but it was not straight forward to me how, which is why I left it out initially

I think it would only be necessary to interpolate the second pass. The first pass just determines the time gradient used for charge extraction - there just skipping broken pixels should be fine.   

Regarding the interpolation, that is something that should be possible to disable - it's not always a good idea to do, especially when there are large regions of broken pixels like a full module or more where it could do more harm than good.  Ideally, we should be able to choose from a list of interpolation methods, including "none", which to my mind means that this should be a separate step in the CameraCalibrator not part of the image extractor itself, implying it should be a hierarchy of Components. Perhaps `BrokenPixelInpainter` (since interpolation is only one way to do it).    That way we can also include other types of broken pixel in the inpainting.

maxnoe:
> , especially when there are large regions of broken pixels like a full module or more where it could do more harm than good

Pixels where no valid neighbors are found are set to 0

maxnoe:
> which to my mind means that this should be a separate step in the CameraCalibrator

It is, the interpolation happens in `CameraCalibrator`. The extractors combining multiple pixels however need to be able to ignore broken pixels.

maxnoe:
> Perhaps BrokenPixelInpainter (since interpolation is only one way to do it). That way we can also include other types of broken pixel in the inpainting.


I started with a Component, then removed it again... The idea here was to start simple and adding another component that needs to have the API `__call__(event)` and does a loop over the telescope events was not something I found attractive.

It all should get better with the `ArrayEventContainer -> TelescopeEventContainer -> [DL0, DL1, DL2,MON]` structure, since then these components concerned with telescope calibration don't have a loop inside and can just take the single telescope event.

maxnoe:
> I think it would only be necessary to interpolate the second pass. The first pass just determines the time gradient used for charge extraction - there just skipping broken pixels should be fine.

That's actually not how it works.

The first pass is just a simple integration for each pixel with some hard coded behavior to make it work exactly like MARS. Which makes it a bit ugly. But since no global or neighbor information is used here, nothing to worry about for the broken pixels.

The second pass starts with the parametrization of the first pass result. It applies a cleaning, thus using neighbor information.

So my strategy would be to add the interpolation in the second pass before the parametrization.

Like this, if only the first pass is used, the interpolation happens in `CameraCalibrator` for all pixels. If both passes are used, the cleaning and parametrization actually handle the broken pixels but the broken pixels are extracted again. Then the final result is interpolated in `CameraCalibrator`.

maxnoe:
An alternative for now would be to use a `TelescopeComponent` with API `__call__(self, tel_id, image, peak_time, broken_pixels)`. 

I can change to that.

maxnoe:
@kosack I introduced the component (I had to also move it to `ctapipe.image` otherwise there would have been a circular import.

I think it also makes sense there and we have anyways this issue here: https://github.com/cta-observatory/ctapipe/issues/1037

ctapipe/image/extractor.py
```diff
@@ -532,13 +550,17 @@ def _calculate_correction(self, telid):
             self.window_shift.tel[telid],
         )
 
-    def __call__(self, waveforms, telid, selected_gain_channel):
+    def __call__(
+        self, waveforms, telid, selected_gain_channel, broken_pixels
```
nbiederbeck, line 554: Do you want to change these to `tel_id` ?

ctapipe/image/tests/test_invalid_pixels.py
```diff
@@ -0,0 +1,61 @@
+import numpy as np
+
+from ctapipe.io.eventsource import EventSource
+
+
+def test_neighor_average(prod5_gamma_simtel_path):
```
nbiederbeck, line 6: typo

ctapipe/io/simteleventsource.py
```diff
@@ -528,6 +530,22 @@ def _generate_events(self):
 
             yield data
 
+    def _get_pixels_status(self, tel_id):
+        tel = self.file_.telescope_descriptions[tel_id]
+        n_pixels = tel["camera_organization"]["n_pixels"]
+        n_gains = tel["camera_organization"]["n_gains"]
+
+        tel = self.file_.telescope_descriptions[tel_id]
```
nbiederbeck, line 538: duplicated line

ctapipe/image/extractor.py
maxnoe: I'll make a follow up PR updating everything at once

ctapipe/io/simteleventsource.py
maxnoe: Merge "conflict", good catch

ctapipe/image/tests/test_invalid_pixels.py
maxnoe: fixed

ctapipe/calib/camera/tests/test_calibrator.py
```diff
@@ -263,3 +264,48 @@ def test_shift_waveforms():
...
+        }
+    )
+    # going to modify this
+    event = deepcopy(example_event)
+    tel_id = list(event.r0.tel)[0]
+    camera = example_subarray.tel[tel_id].camera
+    sampling_rate = camera.readout.sampling_rate.to_value(u.GHz)
+
+    event.mon.tel[tel_id].pixel_status.flatfield_failing_pixels[:, 0] = True
+    event.r1.tel[tel_id].waveform[0:, :] = 0.0
```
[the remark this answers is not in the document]
maxnoe: yes, leftover from when I had only `1:`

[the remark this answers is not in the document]
maxnoe: fixed

LukasNickel (approved):
For me, this looks fine as a building block for the handling of invalid pixels.
Maybe @kosack wants to comment on the structure as he way involved in the earlier discussion

nbiederbeck (approved):

kosack (approved):

PR #2029, opened 2022-07-20
Fix wrong pixel area in nominal frame, fixes #2028

### LINKED ISSUE #2028: CameraDisplay visualisation inaccurate for an offset NominalFrame
**Describe the bug**
When using CameraDisplay to visualise an image in a NominalFrame that has an origin anywhere but at the exact centre of the camera, pixel positions seem generally correct but their sizes are greatly increased to the point they begin obscuring each other.

**To Reproduce**
Here's a relatively tame example of a telescope pointing at 10° altitude and trying to visualise in a NominalFrame with origin at 0°.
```python
import astropy.units as u
from astropy.coordinates import SkyCoord, AltAz
from ctapipe.coordinates import CameraFrame, TelescopeFrame, NominalFrame
from ctapipe.instrument import CameraGeometry
from matplotlib import pyplot as plt

camGeom = CameraGeometry.from_name('FlashCam')

# Set up example physical parameters on the CameraGeometry
focal = 16*u.m
pointing = SkyCoord(alt = 10*u.deg, az = 0*u.deg, frame = AltAz())
camGeom.frame = CameraFrame(focal_length = focal, telescope_pointing = pointing)

# Image data
r = camGeom.guess_radius()
model = Gaussian(x = 0.25*r, y = 0.75*r, width = 0.05*r, length = 0.1*r, psi = '10d')
_,image,_ = model.generate_image(camGeom, intensity = 5000)

# Camera geometry into alternate coordinate systems
horizon = SkyCoord(alt = 0*u.deg, az = 0*u.deg, frame = AltAz())
camGeom_t = camGeom.transform_to(TelescopeFrame(telescope_pointing = pointing))
camGeom_n = camGeom.transform_to(NominalFrame(origin = horizon))

fig,(ax1,ax2) = plt.subplots(1,2)
fig.set_size_inches(16,6)
camDisp_t = CameraDisplay(camGeom_t, ax = ax1, image = image)
camDisp_n = CameraDisplay(camGeom_n, ax = ax2, image = image)
plt.show()
```
img

**Expected behaviour**
Obviously I do expect there to be a distortion of the image as the pointing direction of the telescope gets further away from the origin (the horizon in this case) due to the projection to 2D, but I had anticipated that this would be correctly visualised with pixels changing their shapes. Instead, they seem to stay as a fixed shape and their sizes become wildly incorrect.

**Supporting information**
ctapipe v0.15.0

**Additional context**
I assume visualisation of NominalFrames like this wasn't really foreseen, which I understand. But I was trying to visualise the data in this frame to confirm that the NominalFrame transform was actually doing what I expected. For reasons like this I think it will happen that users attempt to visualise such frames. If meaningful calculation of the projected pixel shapes isn't possible, perhaps ensuring they at least don't blow up in size would be enough to make the visualisation useful.

maxnoe:
For now just a test case

### DIFF
--- a/ctapipe/instrument/camera/tests/test_geometry.py
+++ b/ctapipe/instrument/camera/tests/test_geometry.py
@@ -4,6 +4,7 @@ from copy import deepcopy
 import numpy as np
 import pytest
 from astropy import units as u
+from astropy.coordinates import AltAz, SkyCoord
 
 from ctapipe.instrument import CameraGeometry, PixelShape
 
@@ -295,7 +296,12 @@ def test_camera_from_name(camera_geometry):
 
 def test_camera_coordinate_transform(camera_geometry):
     """test conversion of the coordinates stored in a camera frame"""
-    from ctapipe.coordinates import CameraFrame, EngineeringCameraFrame, TelescopeFrame
+    from ctapipe.coordinates import (
+        CameraFrame,
+        EngineeringCameraFrame,
+        NominalFrame,
+        TelescopeFrame,
+    )
 
     geom = camera_geometry
     trans_geom = geom.transform_to(EngineeringCameraFrame())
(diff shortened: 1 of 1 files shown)


nbiederbeck (dismissed):

LukasNickel (approved):

maxnoe:
Rtol missing from test, please review again @nbiederbeck @LukasNickel

nbiederbeck (approved):

LukasNickel (approved):

PR #2034, opened 2022-07-21
Import all ctapipe_ plugins, not just IO. Add test with dummy plugin

maxnoe:
* Import all plugin packages (moved to __init__.py, removed restriction on `ctapipe_io` to `ctapipe_` prefix
* Test a dummy plugin for unit tests

TODO:

- [ ] Add hooks in ctapipe process to call reconstructor plugins?
- [ ] Add hooks in `DataWriter` so that plugins can store additional information in output files?

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -95,6 +95,7 @@ jobs:
           python --version
           pip install codecov pytest-cov pyflakes pytest-xdist 'coverage!=6.3.0'
           pip install .[all]
+          pip install ./test_plugin
           pip freeze
 
       - name: Static codechecks
--- a/ctapipe/__init__.py
+++ b/ctapipe/__init__.py
@@ -3,6 +3,10 @@ ctapipe - CTA Python pipeline experimental version
 
 Licensed under a 3-clause BSD style license - see LICENSE.rst
 """
+from .core.plugins import IMPORTED_PLUGINS, detect_and_import_plugins
 from .version import __version__
 
-__all__ = ["__version__"]
+__all__ = ["__version__", "IMPORTED_PLUGINS"]
+
+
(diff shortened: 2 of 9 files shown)


kosack:
Sounds reasonable.  This allows also one to replace any class that has a "type" option in the configuration with a custom one. E.g. a new ImageCleaner that isn't already included. 

If all classes can be selected by name at configure-time, perhaps hooks aren't needed explicitly between steps in the event loops. 

E.g. to add a step to shower processor, one only needs to do:
```py3
class MyShowerProcessor(ShowerProcessor):
    def __call__(self, event)
         super().__call__(self, event)
         # do some more stuff
```

Then in the config file, if the plugin is loaded, they can just specify `shower_processor_type=MyShowerProcessor`

```

**Comment by maxnoe** [comment]
> If all classes can be selected by name at configure-time, perhaps hooks aren't needed explicitly between steps in the event loop.

I was thinking about the possibility of adding additional steps (like LST doing the muon calibration) or prototyping deep learning end-to-end or multi output steps. 

Somthing like:
```
post_dl1_components:
     - LSTMuonExtraction
```

and in `process`:

```
self._post_dl1_components = [
     Component.from_name(name, parent=self)
    for name in self.post_dl1_components
]
```


```
for component in self._post_dl1_compoents:
     component(event)
```

kosack:
Ah, ok yes that makes sense. So having a hook for a component before DL1 and after DL1 would be useful.

kosack:
> self._post_dl1_components = [
 >    Component.from_name(name, parent=self)
>    for name in self.post_dl1_components
> ]

Might have to deal with the case where the component is a TelescopeComponent, and therefore needs a SubarrayDescription as well.

maxnoe:
Superseeded by #2101

Issue #1922, opened 2022-05-25
HDF5EventSource only supports image parameters in telescope frame for current files


maxnoe:
**Describe the bug**

The HDF5EventSource checks the data model version to decide whether to read a `HillasParametersContainer` or a `CameraHillasParametersContainer`. This does not support the `--camera-frame` switch of the current `ctapipe-process`.

We should either:
a) Handle current files with hillas parameters in camera frame
b) just drop support for hillas parameters in camera frame everywhere

kosack:
We could also just drop the `--camera-frame` option, as it is no longer needed - it was there to be able to test the implementation in the TelescopeFrame, but as that works, we could remove a lot of code

maxnoe:
This was fixed in  #2079

PR #2036, opened 2022-07-21
Vectorize operations in HillasReconstructor 

maxnoe:
Timing script:

```
from ctapipe.reco import HillasReconstructor
import numpy as np
from timeit import repeat
from cProfile import Profile



source = EventSource("../data/gamma_20deg_180deg_run3___cta-prod6-paranal-2147m-Paranal-dark.dl1.h5", max_events=200)
events = list(source)

reco = HillasReconstructor(subarray=source.subarray)

code = '''
print("start")
for event in events:
    reco(event)
'''

with Profile() as profile:
    exec(code)

profile.dump_stats("hillas_reco.profile")


n_repeat = 5
number = 1
times = np.array(repeat(code, number=number, repeat=n_repeat, globals=globals()))
times /= number
```

On master:
```
11.73 ± 0.31 s
```

on this branch:
```
2.25 ± 0.02 s
```

Units tests are passing, but I will also process a larger DL2 run and check that results are the same.

Most time is not spend in astropy coordinate transforms and project_to_ground.

### DIFF
--- a/ctapipe/coordinates/telescope_frame.py
+++ b/ctapipe/coordinates/telescope_frame.py
@@ -6,23 +6,23 @@ We are just not creating a metaclass and a factory but directly building the
 corresponding class.
 """
 import astropy.units as u
-from astropy.coordinates.matrix_utilities import (
-    rotation_matrix,
-    matrix_product,
-    matrix_transpose,
-)
 from astropy.coordinates import (
-    frame_transform_graph,
-    FunctionTransform,
-    DynamicMatrixTransform,
+    AltAz,
+    Angle,
     BaseCoordinateFrame,
     CoordinateAttribute,
-    TimeAttribute,
+    DynamicMatrixTransform,
     EarthLocationAttribute,
+    FunctionTransform,
     RepresentationMapping,
+    TimeAttribute,
     UnitSphericalRepresentation,
-    Angle,
-    AltAz,
+    frame_transform_graph,
+)
+from astropy.coordinates.matrix_utilities import (
+    matrix_product,
+    matrix_transpose,
+    rotation_matrix,
 )
 
 __all__ = ["TelescopeFrame"]
@@ -77,7 +77,7 @@ class TelescopeFrame(BaseCoordinateFrame):
 
 
 @frame_transform_graph.transform(FunctionTransform, TelescopeFrame, TelescopeFrame)
-def skyoffset_to_skyoffset(from_telescope_coord, to_telescope_frame):
+def telescope_to_telescope(from_telescope_coord, to_telescope_frame):
     """Transform between two skyoffset frames."""
 
     intermediate_from = from_telescope_coord.transform_to(
@@ -90,22 +90,24 @@ def skyoffset_to_skyoffset(from_telescope_coord, to_telescope_frame):
 
 
 @frame_transform_graph.transform(DynamicMatrixTransform, AltAz, TelescopeFrame)
-def reference_to_skyoffset(reference_frame, telescope_frame):
+def altaz_to_telescope(altaz_coord, telescope_frame):
     """Convert a reference coordinate to an sky offset frame."""
 
     # Define rotation matrices along the position angle vector, and
     # relative to the telescope_pointing.
-    telescope_pointing = telescope_frame.telescope_pointing.spherical
+    telescope_pointing = telescope_frame.telescope_pointing.represent_as(
+        UnitSphericalRepresentation
+    )
     mat1 = rotation_matrix(-telescope_pointing.lat, "y")
     mat2 = rotation_matrix(telescope_pointing.lon, "z")
     return matrix_product(mat1, mat2)
 
 
 @frame_transform_graph.transform(DynamicMatrixTransform, TelescopeFrame, AltAz)
-def skyoffset_to_reference(skyoffset_coord, reference_frame):
+def telescope_to_altaz(telescope_coord, altaz_frame):
     """Convert an sky offset frame coordinate to the reference frame"""
 
     # use the forward transform, but just invert it
-    mat = reference_to_skyoffset(reference_frame, skyoffset_coord)
+    mat = altaz_to_telescope(altaz_frame, telescope_coord)
     # transpose is the inverse because mat is a rotation matrix
     return matrix_transpose(mat)
--- a/ctapipe/coordinates/telescope_frame.py
+++ b/ctapipe/coordinates/telescope_frame.py
@@ -28,6 +28,9 @@ from astropy.coordinates.matrix_utilities import (
 __all__ = ["TelescopeFrame"]
 
 
+_wrap_angle = Angle(180, unit=u.deg)
+
+
 class TelescopeFrame(BaseCoordinateFrame):
     """
     Telescope coordinate frame.
@@ -73,7 +76,7 @@ class TelescopeFrame(BaseCoordinateFrame):
 
         # make sure telescope coordinate is in range [-180°, 180°]
         if isinstance(self._data, UnitSphericalRepresentation):
-            self._data.lon.wrap_angle = Angle(180, unit=u.deg)
+            self._data.lon.wrap_angle = _wrap_angle
 
 
 @frame_transform_graph.transform(FunctionTransform, TelescopeFrame, TelescopeFrame)
--- a/ctapipe/reco/hillas_reconstructor.py
+++ b/ctapipe/reco/hillas_reconstructor.py
@@ -8,7 +8,7 @@ from itertools import combinations
 
 import numpy as np
 from astropy import units as u
-from astropy.coordinates import AltAz, SkyCoord, cartesian_to_spherical
+from astropy.coordinates import AltAz, Longitude, SkyCoord, cartesian_to_spherical
 
 from ctapipe.containers import (
     CameraHillasParametersContainer,
@@ -190,9 +190,11 @@ class HillasReconstructor(Reconstructor):
         h_max = self.estimate_h_max(cog_cartesian, telescope_positions)
 
         # az is clockwise, lon counter-clockwise, make sure it stays in [0, 2pi)
-        az = u.Quantity(_two_pi - lon.to_value(u.rad), u.rad)
+        az = Longitude(-lon)
 
-        event.dl2.stereo.geometry[self.__class__.__name__] = ReconstructedGeometryContainer(
+        event.dl2.stereo.geometry[
+            self.__class__.__name__
+        ] = ReconstructedGeometryContainer(
             alt=lat,
             az=az,
             core_x=core_pos_ground.x,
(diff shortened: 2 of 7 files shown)


kosack:
as suggested in #1922 perhaps we should also drop the CameraFrame support in HillasReconstructor (either here or in a separate PR) to simplify this code even more and remove a few branch points.

maxnoe:
@kosack If you look at the code, the branching for `CameraFrame` vs. `TelescopeFrame` is now very minimal, so I'd say not really a concern here.

I am investigating a few issues with astropy concerning performance and I have identified several spots where astropy can be much improved.

For now, I would keep it in the current version I think, we already get a very noticeable speedup.

kosack (approved):
Looks good! 

Before merging, It might be a good idea to do a regression test on a largish number of events to be sure the results have not changed since before the refactoring. (if you have not done that already).

maxnoe:
Will do

kosack (changes requested):
One minor annoyance that would be nice to fix in this is that the azimuths end up in a strange quadrant (negative).   It would be nice for sources with azimuth=180.0 that the reconstructed position is centered there, not at -180.0. It's still correct, but may be confusing when we use that value in a ML algorithm without any transforms.
  
E.g.: can it return values more like the right plot?
image

maxnoe:
Sure, I'll wrap the angle into [-180, 180)

kosack:
Astropy.coord seems to wrap to (0,360], not sure which is better (either way the user has to do a wrap transform if you are one of the usual North or South pointings).   At least we should make all Reconstructors consistent. I think HillasIntersection returns something in (0,360]. Not sure about ImPACT.

maxnoe:
Ok, let's go with the astropy default then

maxnoe:
I processed 100 gamma diffuse runs using 0.16.0 and this branch.

The results are not exactly identical, but they are very close for almost all events and the angular resolution is the same or even slightly better with the new code:

angular_resolution_vectorize_hillas

kosack (dismissed):
I guess the small differences are just due to the different order of operations.  In any case as the results are generally better, I think it's ok.

maxnoe:
Had to rebase, please re-review @kosack @nbiederbeck @LukasNickel

LukasNickel:
`AttributeError: 'Quantity' object has no 'deg' member`

error in docs

ctapipe/reco/hillas_reconstructor.py
```diff
@@ -307,99 +231,107 @@ def initialize_hillas_planes(
...
+
+        for i, (tel_id, hillas) in enumerate(hillas_dict.items()):
+            tel_ids[i] = tel_id
+
+            pointing = event.pointing.tel[tel_id]
+            alt[i] = pointing.altitude.to_value(u.rad)
+            az[i] = pointing.azimuth.to_value(u.rad)
+
+            if isinstance(hillas, CameraHillasParametersContainer):
+                hillas_in_camera_frame = True
```
nbiederbeck, line 266: `hillas_in_camera_frame = isinstance(hillas, CameraHillasParametersContainer)`?

ctapipe/reco/hillas_reconstructor.py
```diff
@@ -165,16 +114,25 @@ class that reconstructs the direction of an atmospheric shower
...
         telframe = TelescopeFrame()
-        self._cam_radius_deg = {}
-        for cam, radius_m in self._cam_radius_m.items():
-            point_cam = SkyCoord(0 * u.m, radius_m, frame=cam.geometry.frame)
+        _cam_radius_deg = {}
+        for cam, radius_m in _cam_radius_m.items():
+            point_cam = SkyCoord(0, radius_m, unit=u.m, frame=cam.geometry.frame)
             point_tel = point_cam.transform_to(telframe)
-            self._cam_radius_deg[cam] = point_tel.fov_lon
+            _cam_radius_deg[cam] = point_tel.fov_lon.to_value(u.deg)
```
LukasNickel, line 127: Not directly related to this PR, but is anyone else really disliking this code block and would rather have the cam radius in meter and degree be a property of the camera geometry?

ctapipe/reco/hillas_reconstructor.py
```diff
@@ -165,16 +114,25 @@ class that reconstructs the direction of an atmospheric shower
...
+
+        # store for each tel_id to avoid costly hash of camera
```
LukasNickel, line 129: For my understanding:
Why not directly assign to `self._cam_radius_x` above?
I am guessing you save calculations by performing them once per camery **type** above and only then looping over the actual telescopes down here?

ctapipe/reco/hillas_reconstructor.py
```diff
@@ -67,10 +69,17 @@ def normalise(vec):
...
+    if vec.ndim == 1:
+        try:
+            return vec / norm
+        except ZeroDivisionError:
+            return vec
+
+    result = np.zeros(vec.shape)
+    mask = norm > 0
+    result[mask] = vec[mask] / norm[mask, np.newaxis]
+    return result
```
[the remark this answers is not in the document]
maxnoe: It's the same... Since the norm is only zero when all entries are zero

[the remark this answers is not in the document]
maxnoe: Returning vec in `else` is to not create a new vector, but could also be `np.zeros(vec.shape)` to have the same copy behavior.

ctapipe/reco/hillas_reconstructor.py
maxnoe: Doesn't really save anything, since the if block contains more statements. Moving it outside the loop is not possible

ctapipe/reco/hillas_reconstructor.py
```diff
@@ -410,35 +342,23 @@ def estimate_direction(self, hillas_planes):
         crossings : shape (n,3) list
             an error estimate
         """
+        index_a, index_b = np.array(list(combinations(range(len(norm)), 2))).T
+        crossings = np.cross(norm[index_a], norm[index_b])
+        mask = crossings[:, 2] < 0
+        crossings[mask] = -crossings[mask]
```
LukasNickel, line 348: Why not directly do `np.cross(norm[index_b], norm[index_a])`?

ctapipe/reco/hillas_reconstructor.py
nbiederbeck: I misread the indentation level of `hillas_in_camera_frame` in L296.

maxnoe:
Errors should be fixed now

ctapipe/reco/hillas_reconstructor.py
```diff
@@ -190,9 +190,11 @@ def __call__(self, event):
         h_max = self.estimate_h_max(cog_cartesian, telescope_positions)
 
         # az is clockwise, lon counter-clockwise, make sure it stays in [0, 2pi)
-        az = u.Quantity(_two_pi - lon.to_value(u.rad), u.rad)
```
LukasNickel: In that case you dont need the `_two_pi` anymore right?

ctapipe/reco/hillas_reconstructor.py
maxnoe: We currently have the geometry strictly in one frame and I quite like this...

maxnoe: This should rather be replaced by dropping support for doing high-level things in camera frame

ctapipe/reco/hillas_reconstructor.py
maxnoe: what would that solve?

maxnoe: switching the argument order inverts the sign for all vectors, which still leaves you with *some* vectors pointing down.

LukasNickel: Ah right, I misunderstood the purpose of this

ctapipe/coordinates/telescope_frame.py
```diff
@@ -90,22 +93,24 @@ def skyoffset_to_skyoffset(from_telescope_coord, to_telescope_frame):
 
 
 @frame_transform_graph.transform(DynamicMatrixTransform, AltAz, TelescopeFrame)
-def reference_to_skyoffset(reference_frame, telescope_frame):
+def altaz_to_telescope(altaz_coord, telescope_frame):
```
LukasNickel, line 96: Codacy noted this:
The `altaz_coord` is not actually used, this does not in fact transform coordinates, but calculates the matrix needed. Same for `telescope_to_altaz`.
And I dont see where this function is used at all in the code. Could you elaborate here?

LukasNickel: Ah, its used in the astropy transforms, my bad

LukasNickel: So then the function signature is required to look like this I assume?

ctapipe/reco/tests/test_HillasReconstructor.py
```diff
@@ -192,6 +152,45 @@ def test_reconstruction_against_simulation(subarray_and_event_gamma_off_axis_500
...
     assert reco_coord.separation(true_coord) < 0.1 * u.deg
 
+    assert u.isclose(result.core_x, event.simulation.shower.core_x, atol=25 * u.m)
+    assert u.isclose(result.core_y, event.simulation.shower.core_y, atol=25 * u.m)
+
+
+def test_reconstruction_against_simulation_camera_frame(
+    subarray_and_event_gamma_off_axis_500_gev,
+):
+    """Reconstruction is here done only in the TelescopeFrame,
```
nbiederbeck, line 162: CameraFrame

ctapipe/coordinates/telescope_frame.py
maxnoe: Yes it is. The only change here is to rename the function from the originally copy/pasted code of `SkyOffsetFrame` transforms in astropy.

maxnoe: Coordinate transform graph functions always have the signature `(from_coord, to_frame)`

ctapipe/reco/tests/test_HillasReconstructor.py
maxnoe: fixed

ctapipe/reco/tests/test_HillasReconstructor.py
```diff
@@ -261,13 +260,32 @@ def test_CameraFrame_against_TelescopeFrame(filename):
             for field, cam in result_camera_frame.items():
                 tel = getattr(result_telescope_frame, field)
 
+                kwargs = dict(rtol=6e-3, equal_nan=True)
```
nbiederbeck, line 263: a reason for `6e-3`?

maxnoe: that was needed to make the tests pass... It's more a regression test for the future.

nbiederbeck (approved):

LukasNickel (approved):

PR #2038, opened 2022-07-25
Add 'raises' to run_tool, which raises exceptions instead of logging them.

nbiederbeck:
With this, we can test that tools fails with the expected Exceptions.

### DIFF
--- a/ctapipe/core/tests/test_tool.py
+++ b/ctapipe/core/tests/test_tool.py
@@ -1,13 +1,14 @@
-import os
-import logging
-import tempfile
-import pytest
 import json
-from traitlets import Float, TraitError, Dict, Int
-from traitlets.config import Config
+import logging
+import os
+import tempfile
 from pathlib import Path
 
-from .. import Tool, Component
+import pytest
+from traitlets import Dict, Float, Int, TraitError
+from traitlets.config import Config
+
+from .. import Component, Tool
 from ..tool import export_tool_config_to_commented_yaml, run_tool
 
 
@@ -33,7 +34,7 @@ def test_tool_simple():
 
 
 def test_tool_version():
-    """ check that the tool gets an automatic version string"""
+    """check that the tool gets an automatic version string"""
 
     class MyTool(Tool):
         description = "test"
@@ -44,7 +45,7 @@ def test_tool_version():
 
 
 def test_provenance_dir():
-    """ check that the tool gets the provenance dir"""
+    """check that the tool gets the provenance dir"""
 
     class MyTool(Tool):
         description = "test"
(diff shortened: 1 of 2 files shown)


ctapipe/core/tests/test_tool.py
```diff
@@ -340,3 +342,28 @@ class MyTool(Tool):
...
+        name = "Bad"
+        description = "This tool raises an exception."
+
+        def start(self):
+            raise ValueError("1 does not equal 0.")
+
+    assert run_tool(ToolGood(), raises=True) == 0
+
+    with pytest.raises(ValueError):
+        ret = run_tool(ToolBad(), raises=True)
```
maxnoe, line 365: Add test that with `raises=False` we get a non-zero exit code for the bad tool

nbiederbeck: done

maxnoe (approved):

ctapipe/core/tests/test_tool.py
```diff
@@ -340,3 +342,30 @@ class MyTool(Tool):
...
+    assert run_tool(ToolGood(), raises=True) == 0
+
+    assert run_tool(ToolBad(), raises=False) == 1
+
+    with pytest.raises(ValueError):
+        ret = run_tool(ToolBad(), raises=True)
+
+    # `ret` does not exist, because `run_tool` raised an Exception
+    with pytest.raises(UnboundLocalError):
+        assert ret == 1
```
maxnoe, line 371: `assert "ret" not in locals()`?

maxnoe: Or just leave it out, there is no possibility of ret being set when the ValueError does not happen

nbiederbeck: > there is no possibility of ret being set when the ValueError does happen

Yes, and I wanted to make that very very clear. But I can remove it

ctapipe/core/tool.py
```diff
@@ -357,6 +358,8 @@ def run(self, argv=None):
             Provenance().finish_activity(activity_name=self.name, status="interrupted")
             exit_status = 130  # Script terminated by Control-C
         except Exception as err:
+            if raises:
```
LukasNickel, line 361: Do we want this to happen before finishing the provenance log?
I realize that this is mostly meant for testing anyway, but would personally still write out the logs

maxnoe: Good point.

maxnoe: Actually, the provenance is being written (it is in the `finallly` block below, which always gets run).

But it is missing now the cause for the finishing of the activity. So we should definitely move the raise below the `finish_activity`.

ctapipe/core/tests/test_tool.py
nbiederbeck: I removed it

ctapipe/core/tool.py
nbiederbeck: Done

LukasNickel (approved):

maxnoe (approved):

Issue #1953, opened 2022-06-26
Rename field `tel_ids` to `tels` for reconstructed containers


maxnoe:
**Please describe the use case that requires this feature.**

After writing to hdf5 files and also when reading back, lists of tel_ids are transformed into boolean masks. 

This means the name `tel_ids` is misleading.

**Describe the solution you'd like**

Rename to `tels` or `telescopes`

maxnoe:
@kosack any opinions? If we want to make this change, we should do it now with the other data model changes

PR #2044, opened 2022-07-28
Rename ReconstructedContainer.tel_ids -> telescopes, fixes 1953

maxnoe:


kosack:
Isn't this less obvious than using tel_ids? There are multiple ways to refer to a telescope:  tel_ids, tel_indices, and tel_names (maybe more).  So just saying  "telescopes" it is not clear which one is expected.  tel_ids makes it clear this is a telescope ID, and not a telescope name, or telescope index.

maxnoe:
@kosack the reasoning here is that e.g. when using the table loader, it is actually a boolean mask, not a list of tel_ids. see #1953

maxnoe:
An alternative would be to just directly fill the mask and call it `tel_mask`?

kosack:
ah, yes, I forgot that there is a translation to a mask when we write it, and then the True values are actually telescope_indices, not tel_ids... so ok, let's keep it as `telescopes`, and the representation of these should be up to the user.  I still like having it a list of tel_ids in the event, since that is somewhat friendlier to read

kosack (approved):

PR #2050, opened 2022-08-03
Merge DL2 tables.

nbiederbeck:
For telescope tables: loop over existing algorithms.
For subarray tables: add them to the merge-set.

### DIFF
--- a/ctapipe/tools/merge.py
+++ b/ctapipe/tools/merge.py
@@ -37,7 +37,13 @@ optional_nodes = {
     "/simulation/event/telescope/impact",
     "/dl1/event/telescope/parameters",
     "/dl1/event/telescope/images",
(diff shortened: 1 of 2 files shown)


kosack (approved):

Hckjs (approved):

ctapipe/tools/merge.py
```diff
@@ -73,6 +79,12 @@
 
 dl2_subarray_nodes = {"/dl2/event/subarray/geometry"}
 
+dl2_algorithm_tel_nodes = {
```
maxnoe, line 82: As with the table loader, I think we should just merge everything that is in `dl2/event/telescope/{kind}/{algorithm}/tel_{tel_id}` and `dl2/event/subarray/{kind}/{algorithm}`

PR #2060, opened 2022-08-26
Use _impact_ instead of _tel_, fixes #2051

### LINKED ISSUE #2051: reconstructed impact container has wrong prefix when loaded
[...] The Container has a prefix of "impact" so the intention was that the column name is: `impact_distance` not `tel_distance`. [...]

_Originally posted by @kosack in https://github.com/cta-observatory/ctapipe/issues/2048#issuecomment-1203837728_

nbiederbeck:


kosack (approved):

nbiederbeck:
This currently breaks three other tests, two of which tests for `tel_distance`.

I'm trying to figure out why the other one fails...

kosack (changes requested):
Need to update the data model version I guess, since this will break scripts reading files that were written with ctapipe-0.16 that assume the old naming (probably there are none, but still good to update).

maxnoe:
The idea here was that all mono, telescope specific features have a tel_ prefix to distinguish them from array ones. This is important e.g. for joining tables in the table loader.

So please at least use `tel_impact`

kosack:
> So please at least use tel_impact

Ok, agreed - I didn't like that the word `impact` got dropped somehow in the last refactoring, but having all telescope-wise parameters start with `tel_` is also very good.  So yes, let's please change it to "tel_impact_distance", which means just changing the prefix to "tel_impact" for that container.

nbiederbeck:
Done. I think I changed it for all DL2 telescope-wise parameters, at least the tests don't tell me otherwise. I added the change to the DataModel description

kosack (approved):

ctapipe/io/datawriter.py
```diff
@@ -46,8 +46,9 @@ def _get_tel_index(event, tel_id):
 #   (meaning readers need to update scripts)
 # - increase the minor number if new columns or datasets are added
 # - increase the patch number if there is a small bugfix to the model.
-DATA_MODEL_VERSION = "v4.0.0"
+DATA_MODEL_VERSION = "v5.0.0"
 DATA_MODEL_CHANGE_HISTORY = """
+- v5.0.0: - Change DL2 telescope-wise container prefixes from {algorithm}_tel to {algorithm}_tel_{kind}
```
maxnoe, line 51: This indicates a much larger change than has actually happened.

We just renamed `tel_distance` to `tel_impact_distance`. I.e. whe changed the prefix of a single container from '' to 'impact'.

Nothing to do with "algorithms" or "kinds".

nbiederbeck: Well yes, but in https://github.com/cta-observatory/ctapipe/blob/6cc4f66f6e61a8da2bbbc72507d66381adde91e0/ctapipe/io/hdf5eventsource.py#L449 I made the changes as indicated in the history. This currently only adds the `_impact_` to tel_distance, but (in future?) other DL2 telescope-wise parameters will have the "kind", in addition to the "algorithm", with the "tel" in between. I wanted to account for this in the history. 

But I understand, currently this changes only the `tel_distance`, and am fine with making that clearer

maxnoe: Sounds good

kosack (approved):

maxnoe:
Let's do a bugfix release first and try to sort out the test data before going to 0.17.

I'll approve, but let's wait with the merging...

maxnoe:
On the other hand, this changes the data model (again) and there shouldn't be any data using 0.16 yet, so let's merge this and then reproduce test data.

maxnoe (approved):

maxnoe (approved):

PR #2064, opened 2022-09-05
Disp reconstruction

LukasBeiske:
Adds functionality to train and apply a regressor and a classifier for origin reconstruction using the disp method.

- The tool `ctapipe-ml-train-disp` is added and can be used to train and validate both models. Application can be done using `ctapipe-ml-apply` and `ctapipe-process` just like particle classification and energy regression. 
- A `DispContainer` is added for storing the predictions of both models. The resulting predictions for altitude and azimuth are stored in the `ReconstructedGeometryContainer`.
- The `DispConverter` component is added to do the conversion from |disp| & sign predictions to alt & az predictions. This component mirrors the design of the other ML components in the sense that the `predict` function works on tables and `__call__` works on individual array events.
- The same tests as for energy regression and particle classification are added. However, due to a known error with the datamodel of the dataset currently used for testing, some of the tests are failing atm. 
- Some variable names can probably be improved (e.g. `disp_regressor` vs. `norm_regressor`)

TODO:
- [ ] Error of mean alt & az predictions 
- [ ] Add option to provide different quality criteria for different telescope types (not necessarily part of this PR)


kosack:
Nice!  

Once this is merged (or as part of it) we should add a stereo `Reconstructor` that takes the per-telescope monoscopic reconstructions and does a weighted average to predict the stereo point (algorithm 3 in the [Hofmann et al 1999 paper](https://ui.adsabs.harvard.edu/link_gateway/1999APh....12..135H/doi:10.1016/S0927-6505(99)00084-5))

ctapipe/coordinates/disp.py
```diff
@@ -0,0 +1,135 @@
...
+        pointing = SkyCoord(alt=pointing_alt, az=pointing_az, frame=AltAz())
+        tel_coord = TelescopeFrame(
+            fov_lon=lon, fov_lat=lat, telescope_pointing=pointing
+        )
+        horizontal_coord = tel_coord.transform_to(AltAz())
+
+    return horizontal_coord.alt.to(u.deg), horizontal_coord.az.to(u.deg)
+
+
+class DispConverter(Component):
```
kosack, line 52: This may be better named `MonoReconstructor` or similar: it reconstructs the shower parameters for a single telescope.

ctapipe/containers.py
```diff
@@ -840,16 +841,29 @@ class ParticleClassificationContainer(Container):
...
 
 
+class DispContainer(Container):
+    """
+    Standard output of disp reconstruction algorithms for origin reconstruction
+    """
+
+    default_prefix = ""
+
+    norm = Field(nan * u.deg, "reconstructed absolute value for disp", unit=u.deg)
```
kosack, line 851: Shouldn't this be called "disp"? Norm is a bit generic.  Or instead it could be better to use "disp" as the `default_prefix`, then the output columns will be "disp_norm" and "disp_sign"

LukasNickel: I think using the prefix is a good idea

ctapipe/coordinates/disp.py
LukasBeiske: Good idea. How about `MonoDispReconstructor` to be a bit more specific and avoid confusion if used in larger tools?

ctapipe/containers.py
LukasBeiske: Agreed. Will change that.

ctapipe/coordinates/disp.py
```diff
@@ -0,0 +1,135 @@
...
+    """Convert (norm, sign) predictions into (alt, az) predictions"""
+
+    # This is a temporary solution for simulations using a single pointing position
+    pointing_altitude = u.Quantity
```
maxnoe, line 56: This assigns the class `u.Quantity` as a class members of `DispConverter`, I think this is not what you want to do here (

LukasBeiske: Oh, yes that's wrong. Would it be correct to use `traitlets.Int` here?

maxnoe: Int is certainly wrong for altitude / azimuth. What's the point here?

LukasBeiske: I meant float, sorry. As the pointing information is not yet loaded with TableLoader, I want to store singular values for `pointing_altitude` and `pointing_azimuth` as a temporary solution for it to work with simulations.

LukasBeiske: This is only used if `predict()` is called. So, `pointing_altitude` and `pointing_azimuth` could also be parameters of `predict()`, if this makes more sense.

ctapipe/coordinates/disp.py
```diff
@@ -0,0 +1,135 @@
...
+        for tel_id in event.trigger.tels_with_trigger:
+            # Maybe NormContainer + SignContainer makes more sense
+            # As is: DispContainer always contains only sign or norm
+            norm = event.dl2.tel[tel_id].disp[regressor_cls].norm
+            valid_norm = event.dl2.tel[tel_id].disp[regressor_cls].norm_is_valid
+            sign = event.dl2.tel[tel_id].disp[classifier_cls].sign
+            valid_sign = event.dl2.tel[tel_id].disp[classifier_cls].sign_is_valid
+
+            if valid_sign:
+                sign = -1 if sign < 0.5 else 1
```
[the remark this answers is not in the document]
LukasBeiske: Atm `predict` effectively returns one of the resulting scores of `sklearn.XYZclassifier.predict_proba()` which is between 0 and 1. 
This score could be rescaled to be between -1 and 1 or `predict` could return the value -1 or 1 instead. However, returning the class label would be a different behaviour from the `ParticleIdClassifier`, but for sign the score is not important, so it might make sense anyway.
Returning the sign score is nice for some performance plots though.

ctapipe/coordinates/disp.py
maxnoe: I would say yes, this makes more sense. Even more sense would be to fill this info in table loader from the now merged obs-config. Should be an easy change and could be done in a PR towards `master` even.

ctapipe/coordinates/disp.py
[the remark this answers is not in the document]
maxnoe: Ah, that means we should change `predict` to `predict_proba` and have a `predict` method that really calls `predict` I would say.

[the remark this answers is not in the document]
maxnoe: I assumed when writing the `Classifier` component, that there isn't a usecase for `predict` and we'd always want the score.

ctapipe/ml/stereo_combination.py
```diff
@@ -40,8 +47,8 @@ def _weighted_mean_ufunc(tel_values, weights, n_array_events, indices):
 
 class StereoCombiner(Component):
     # TODO: Add quality query (after #1888)
-    algorithm = Unicode().tag(config=True)
-    combine_property = CaselessStrEnum(["energy", "classification", "direction"]).tag(
+    algorithm = List(Unicode()).tag(config=True)
```
nbiederbeck, line 50: See my comment below. In any case, we should use the correct plural/singular form of `algorithm(s)`

ctapipe/ml/tools/apply.py
```diff
@@ -186,6 +243,110 @@ def _apply(self, reconstructor, parameter):
 
         return vstack(tables)
 
+    def _apply_disp(self):
+        # Different prefix? Two algorithms -> How to combine them?
```
nbiederbeck, line 247: Yes, different prefix (see my other comment). Applying disp makes only sense if we apply both reconstructors.

ctapipe/ml/tools/tests/test_process.py
```diff
@@ -96,3 +96,60 @@ def test_process_apply_classification(
...
+            output,
+            "/dl2/event/telescope/geometry/ExtraTreesRegressor_ExtraTreesClassifier/tel_004",
+        )
+    )
+    print(
+        read_table(
+            output,
+            "/dl2/event/subarray/geometry/ExtraTreesRegressor_ExtraTreesClassifier",
+        )
+    )
```
nbiederbeck, line 155: replace `print`s with some `assert`s

ctapipe/ml/tools/tests/test_train.py
```diff
@@ -94,3 +120,26 @@ def test_cross_validation_results(model_tmp_path):
...
+            f"--output-regressor={out_file_reg}",
+            f"--output-classifier={out_file_clf}",
+            f"--config={config}",
+            "--log-level=INFO",
+            f"--CrossValidator.output_path={disp_cv_out_file_base}",
+        ],
+    )
+    assert ret == 0
+    assert disp_reg_cv_out_file.exists()
+    assert disp_clf_cv_out_file.exists()
```
nbiederbeck, line 145: This feels off, I cannot articulate why.

ctapipe/ml/tools/train_disp_reconstructor.py
```diff
@@ -0,0 +1,210 @@
...
+from ctapipe.core import Tool
+from ctapipe.core.traits import Bool, Int, Path
+from ctapipe.io import EventSource, TableLoader
+
+from ..apply import CrossValidator, DispClassifier, DispRegressor
+from ..preprocessing import check_valid_rows
+from ..sklearn import Classifier, Regressor
+
+
+def euclidean_distance(x1, y1, x2, y2):
```
nbiederbeck, line 14: might be placed in `utils`, or made private.

ctapipe/ml/tools/train_disp_reconstructor.py
```diff
@@ -0,0 +1,210 @@
...
+        default_value=None, allow_none=False, directory_ok=False
+    ).tag(config=True)
+
+    output_path_clf = Path(
+        default_value=None, allow_none=False, directory_ok=False
+    ).tag(config=True)
+
+    n_events = Int(default_value=None, allow_none=True).tag(config=True)
+    random_seed = Int(default_value=0).tag(config=True)
+    project_disp = Bool(default_value=False).tag(config=True)
```
nbiederbeck, line 34: this needs the `help` argument set

ctapipe/coordinates/disp.py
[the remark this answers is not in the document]
LukasBeiske: The `Classifier` component in `ctapipe/ml/sklearn` does have two functions `predict` and `predict_score`, but the `predict` function of the `ParticleIdClassifier` component in `ctapipe/ml/apply` calls the `predict_score` function of its `Classifier` model. 
So I think renaming `predict` from `ParticleIdClassifier` and `DispClassifier` to `predict_score` and adding a `predict` function to `DispClassifier` which calls the `predict` function of its `Classifier` model, would be the easiest solution.

ctapipe/coordinates/disp.py
LukasBeiske: I changed it to be parameters for now. I can look into the necessary changes to table loader later.

ctapipe/coordinates/disp.py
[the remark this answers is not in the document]
maxnoe: Ah! One level higher. Then no. The `.predict` method of these high-level things should return what we fill into the containers, just as a table. So it is correct that `ParticleIdClassifier.predict` returns the score (because we fill gammaness into the container) but `DispClassifier.predict` should return arrays with `-1` or `1` because that's what is in the `DispContainer`.

I'd also probably wouldn't create two high-level components, just a single one that has the two models:

```
class DispRegressor:
    def __init__(self, ...):
        self.norm_regressor = Regressor(...
        self.sign_classifier = Classifier(... 
```

That would solve the strange "two algorithms" naming as well. Conceptually, this is one 2d regression, not two separate algorithms.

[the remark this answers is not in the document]
LukasBeiske: The `DispContainer` is filled with the sign score (between 0 and 1) atm. 

Would it make sense to make `self.norm_regressor` and `self.sign_classifier` `Reconstructor`s to keep the functionality of separate quality queries and feature generators for both models?

ctapipe/ml/stereo_combination.py
```diff
@@ -156,8 +163,57 @@ def _combine_classification(self, event):
...
+            )
+            valid = False
+
+        event.dl2.stereo.geometry[prefix] = ReconstructedGeometryContainer(
+            alt=mean_altaz.alt.to(u.deg),
+            az=mean_altaz.az.to(u.deg),
+            telescopes=ids,
+            is_valid=valid,
+        )
+        event.dl2.stereo.geometry[prefix].prefix = prefix
```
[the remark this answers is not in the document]
LukasBeiske: Done.

ctapipe/coordinates/disp.py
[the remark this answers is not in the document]
LukasNickel: Separate quality queries? An event having only disp or sign is not really useful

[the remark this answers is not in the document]
LukasBeiske: In regards to the "two algorithms" naming: I was of the impression, that it would be necessary to include the name of the used ML algorithms in the prefix to be able to distinguish if multiple disp models are applied on the same file. Or can/ should this never be the case, anyway?

ctapipe/ml/tools/train_disp_reconstructor.py
LukasBeiske: Do you mean `help` should be added to `n_events` and `random_seed`, too?

ctapipe/ml/tools/tests/test_train.py
LukasBeiske: Do you mean the naming (maybe use `disp_cv_out_file_{reg,clf}` instead)? Or do you mean this will/ should not work correctly?

ctapipe/ml/tools/train_disp_reconstructor.py
LukasBeiske: It is only used once, so removing it all together might be the best solution. This was just laziness on my end and it does not make the code much less readable if this calculation is done in place, I think.

ctapipe/ml/tools/train_disp_reconstructor.py
nbiederbeck: no, those are clear by the name. I have just no idea what `project_disp` does.

ctapipe/coordinates/disp.py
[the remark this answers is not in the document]
maxnoe: I don't really think that's necessary to have separate feature generators and separate quality queries does not make sense (you would apply the models to different events...).

Since even if you are unsure about the classification you cannot really do anything with that info, I'd rather store the sign in the container (maybe you can store the score in another field if it is really useful)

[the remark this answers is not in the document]
LukasBeiske: Ok, I understand. Would it still make sense to have this single `DispRegressor` high-level component be a subclass of `Reconstructor`, since `Reconstructor` is designed to work with a single sklearn model, or would it be better to make it a separate `Component`?

I think storing the score in another field is a good idea, event if it's only for performance plots. If we use a separate function to predict the score (e.g. `predict_score`), we could only do so during cross validation. However, if we want to always store the score and the predicted class label, having both `predict` and `predict_score` seems unnecessary, because we could just decide the predicted class label from the score. And if we always store the score and the predicted class label, `predict` would still return what is filled in the container.

ctapipe/ml/tools/tests/test_process.py
LukasBeiske: Done.

ctapipe/ml/tools/apply.py
```diff
@@ -60,6 +74,8 @@ class Apply(Tool):
         ("i", "input"): "Apply.input_url",
         "regressor": "Apply.energy_regressor_path",
         "classifier": "Apply.particle_classifier_path",
+        "disp-regressor": "Apply.disp_regressor_path",
+        "sign-classifier": "Apply.sign_classifier_path",
```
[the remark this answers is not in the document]
LukasBeiske: Since this is a minor change, I included it here.

ctapipe/coordinates/disp.py
```diff
@@ -0,0 +1,139 @@
...
+from ..core import Component
+
+__all__ = [
+    "horizontal_to_telescope",
+    "telescope_to_horizontal",
+    "MonoDispReconstructor",
+]
+
+
+def horizontal_to_telescope(
```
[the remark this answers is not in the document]
LukasBeiske: These wrapper functions are useful to ignore warnings about missing time and location frame attributes, as they are not needed for this specific transformation. And they improve code readability, in my opinion.

ctapipe/ml/stereo_combination.py
```diff
@@ -156,8 +163,57 @@ def _combine_classification(self, event):
...
+        container.prefix = self.algorithm[0]
+        event.dl2.stereo.classification[self.algorithm[0]] = container
+
+    def _combine_disp(self, event):
+        ids = []
+        alt_values = []
+        az_values = []
+        weights = []
+
+        prefix = self.algorithm[0] + "_" + self.algorithm[1]
```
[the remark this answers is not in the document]
LukasBeiske: I agree. `"disp"` is now implemented as `self.algorithm` and as the prefix, resolving the comments below, too.

ctapipe/ml/tools/tests/test_train.py
LukasBeiske: I changed the naming as stated above. Marking this as resolved for now.

ctapipe/coordinates/disp.py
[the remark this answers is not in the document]
LukasBeiske: I implemented a single `DispReconstructor` component whose `predict` function returns `norm`, `sign` and `sign_score`. I also changed the `DispContainer` to contain these three values.

Regarding the "two algorithms" problem: I changed the prefix (and the `algorithm` for the stereo combiner) to be `"disp"` everywhere. This results in a slightly redundant container structure: `event.dl2.tel[tel_id].disp["disp"] = DispContainer`. However using `event.dl2.tel[tel_id].disp = DispContainer` would require changes in the Datawriter and TableLoader, if I am not mistaken. I don't know if such changes would make sense, but I don't think they would be worth it, in any case.

LukasBeiske:
I noticed that the `log_target` for the `Regressor` component uses the natural logarithm: https://github.com/cta-observatory/ctapipe/blob/552aba3a9a96b63fce8dfbc3fce0cbe0b26c72d4/ctapipe/ml/sklearn.py#L122
Shouldn't this be the base 10 logarithm?

kosack (changes requested):

LukasBeiske:
To have a clean history, I will create a new branch and a new PR containing the latest version of the disp implementation, as soon as the `ml` branch is merged into `master`.
`git rebase` broke or requires reworking many commits whose changes get reverted a few commits later (on this current branch) and doing so isn't worth the effort imo.

PR #2066, opened 2022-09-05
Sort unique cameras/types/optics to generate stable indices in hdf5, fixes #2063

### LINKED ISSUE #2063: Stable optics and camera indices
Minor stylistic request to improve consistency between runs of one production. 

Define an order (i.e. linked to tel ids)

**Additional context**
Screenshot 2022-08-26 at 15 02 10

maxnoe:
Possible solution for #2063 

However, I am not sure we really want to introduce this requirement

### DIFF
--- a/ctapipe/instrument/subarray.py
+++ b/ctapipe/instrument/subarray.py
@@ -394,19 +394,34 @@ class SubarrayDescription:
 
     @lazyproperty
     def telescope_types(self) -> Tuple[TelescopeDescription]:
-        """list of telescope types in the array"""
-        unique_types = {tel for tel in self.tel.values()}
+        """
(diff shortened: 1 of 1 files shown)


kosack:
I don't think it should be a requirement (or relied on), but it is nice just from a user-readablity standpoint that the order doesn't vary unexpectedly.  However, I still wonder why it would for files from the same prod and subarray...

maxnoe:
@kosack see my comment here: https://github.com/cta-observatory/ctapipe/issues/2063#issuecomment-1238105227

nbiederbeck (approved):

kosack (approved):

PR #2072, opened 2022-09-09
Fix unreachable assertion in Provenance unit test

HealthyPear:
The assertion was inside a fixture after a return statement, so it was never called.

I created a new unit-test that uses the fixture and calls that assertion.

### DIFF
--- a/ctapipe/core/tests/test_provenance.py
+++ b/ctapipe/core/tests/test_provenance.py
@@ -19,7 +19,8 @@ def provenance():
     prov.finish_activity("test1")
     return prov
 
-    assert set(prov.finished_activity_names) == {"test2", "test1"}
+def test_provenance_activity_names(provenance):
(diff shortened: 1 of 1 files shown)


ctapipe/core/tests/test_provenance.py
```diff
@@ -19,6 +19,8 @@ def provenance():
     prov.finish_activity("test1")
     return prov
 
+def test_provenance_activity_names(provenance):
+    prov = provenance
```
kosack, line 23: I guess this line is not needed, could just use 
```py
assert set(provenance.finished_activity_names) == {"test2", "test1"}
```

HealthyPear: done

maxnoe (approved):

kosack (approved):

PR #2073, opened 2022-09-16
Improve ShowerProcessor API, fixes #2071

### LINKED ISSUE #2071: Change API of `Reconstructor` and `ShowerProcessor` to allow multiple reconstructions and common predictions

Currently, the `ctapipe.reco.Reconstructor` baseclass has the API `__call__(self, event: ArrayEvent) -> ReconstructedGeoemtryContainer`.

This is inflexible, as it makes it impossible to have reconstruction algorithms that directly predict a combination of geometry, energy and particle type, like ImPACT, machine learning, etc.

Also, `ShowerProcessor` only allows a single algorithm.

Change the API of the general base `Reconstructor` to be `__call__(self, event) -> None`, filling any dl2 container it generates into the event structure.

Change `ShowerReconstructor` to allow a list of `Reconstructor` base classes that are **executed in order**.

I.e. the config should look like this:

```
ProcessorTool:
    ShowerProcessor:
        reconstructors: 
            - HillasReconstructor
            - HillasIntersection
            - EnergyRegressor
            - ImPACT
            - ParticleIDClassifier
        ParticleIDClassifier:
             model_path: ...
        ImPACT:
             template_path:  
             geometry_seed: HillasReconstructor
             energy_seed: RandomForestRegressor 
```

This would still (due to how traitlets works) have the limitation of not being able to e.g. use different configurations of the same reconstructor but would be a big step forward

**Additional context**

This means that also the classes in the new ml branch should then be subclasses of `Reconstructor`

StFroese:
I changed two major things. 
1) `ShowerProcessor.__call__` now no longer fills the Containers from the Reconstructors into the event structure. This is now done directly in the call function of each Reconstructor
2) `ShowerProcessor` now allows a list of multiple reconstructors. I also had to change the `ProcessorTool` to write event statistic tables for a list of reconstructors. 

I also changed almost every config in the `resources` folder and some tests. Please take a look at that 😅


ctapipe/reco/shower_processor.py
```diff
@@ -25,11 +25,14 @@ class ShowerProcessor(Component):
     Input events must already contain dl1 parameters.
     """
 
-    reconstructor_type = create_class_enum_trait(
-        Reconstructor,
-        default_value="HillasReconstructor",
-        help="The stereo geometry reconstructor to be used",
-    )
+    reconstructor_types = List(
```
kosack, line 28: The help message now now longer shows the list of possible reconstructor types (a known problem when we change from an enum  to a list of enums...).  It would be good to explicitly add the class list to the `help=` option as we did elsewhere. 

Here is what we get now:
```
ShowerProcessor(Component) options
--ShowerProcessor.reconstructor_types=<list-item-1>...
    The stereo geometry reconstructors to be used
    Default: ['HillasReconstructor']
```

And this is what was there before (note the list of names):
```
--ShowerProcessor.reconstructor_type=<CaselessStrEnum>
    The stereo geometry reconstructor to be used
    Choices: any of ['HillasReconstructor', 'HillasIntersection'] (case-insensitive)
    Default: 'HillasReconstructor'
```

Could just add this to the help string:
```py3
f"Choices are: {list(Reconstructor.non_abstract_subclasses().keys())}" 
```

ctapipe/reco/tests/test_shower_processor.py
```diff
@@ -1,18 +1,28 @@
...
 
 
-@pytest.mark.parametrize("reconstructor_type", ["HillasReconstructor", "HillasIntersection"])
-def test_shower_processor_geometry(example_event, example_subarray, reconstructor_type):
+@pytest.mark.parametrize(
+    "reconstructor_types",
+    [
+        ["HillasReconstructor"],
+        ["HillasIntersection"],
+        ["HillasReconstructor", "HillasIntersection"],
```
kosack, line 20: To be future-proof here, you might want to use:
```py3
list(Reconstructor.non_abstract_subclasses().keys())
```
rather than an explicit list.  That way if anyone adds a new reconstructor, it is tested

ctapipe/resources/base_config.yaml
```diff
@@ -93,9 +93,9 @@ ImageProcessor:
 # The ShowerProcessor performs the DL1 to DL2a (reconstructed shower geometry)
 # transition. It is run only if the parameter DataWriter.write_showers=True.
 ShowerProcessor:
-  # choose between HillasReconstructor and HillasIntersection (two
+  # choose between HillasReconstructor and HillasIntersection or both (two
   # implementations of the standard stereo line-intersection method)
-  reconstructor_type: HillasReconstructor
+  reconstructor_types: [HillasReconstructor, HillasIntersection]
```
kosack, line 98: In YAML this is maybe better written as:
```yaml
reconstructor_types:
    - HillasReconstructor
    - HillasIntersection
```

kosack (changes requested):

kosack:
In light of this change, we should also review #2052 and perhaps consider making ImPACT a subclass of Reconstructor again, and then it could be included in this list. It is dependent on a previous reconstructor, so some check would be needed to ensure that the user didn't forget to include it's seed reconstructor.

kosack:
I notice we still have a descrepency in the output sign:
image

ctapipe/resources/base_config.yaml
kosack: also not sure if the default config should really include both (remember this config is what a user will get when running ctapipe-quickstart), so perhaps comment-out the HillasIntersection line just to show that it can be enabled, but is not enabled by default

ctapipe/reco/shower_processor.py
```diff
@@ -70,21 +76,23 @@ def __call__(self, event: ArrayEventContainer):
...
-        )
+            # compute and store the impact parameter for each reconstruction
 
-        for tel_id in event.trigger.tels_with_trigger:
-            tel_index = self.subarray.tel_indices[tel_id]
-            event.dl2.tel[tel_id].impact[k] = TelescopeImpactParameterContainer(
-                distance=impact_distances[tel_index],
-                prefix=f"{self.reconstructor_type}_tel",
+            # for the stereo reconstructor:
+            impact_distances = shower_impact_distance(
```
maxnoe, line 87: I think we should move this calculation into the stereo reconstructors or at least check if the reconstructor actually filled the geometry container.

maxnoe (changes requested):
Small comment, otherwise looks good. (And happy that this was as easy as I thought)

ctapipe/reco/hillas_intersection.py
```diff
@@ -248,8 +246,8 @@ def _predict(self, hillas_dict, array_pointing, telescopes_pointings=None):
             az=sky_pos.altaz.az.to(u.rad),
             core_x=grd.x,
             core_y=grd.y,
-            core_tilted_x=core_x,
-            core_tilted_y=core_y,
+            core_tilted_x=u.Quantity(core_x, u.m),
```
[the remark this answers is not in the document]
StFroese: I think the problem is that `reconstruct_tilted` only returns floats and not Quantities. This was already taken into account for `core_tilted_uncert_{x,y} `, which is converted to a Quantity. 
If I remember correctly this resulted in some test failing for the `ProcessorTool`. It was not noticed before because the tests only included `HillasReconstructor` and not `HillasIntersection`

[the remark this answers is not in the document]
maxnoe: Ah ok, so the units were just missing before, great then

ctapipe/reco/shower_processor.py
StFroese: Done

ctapipe/reco/shower_processor.py
StFroese: I moved it to `Reconstructor`

ctapipe/resources/base_config.yaml
```diff
@@ -95,7 +95,9 @@ ImageProcessor:
 ShowerProcessor:
   # choose between HillasReconstructor and HillasIntersection or both (two
   # implementations of the standard stereo line-intersection method)
-  reconstructor_types: [HillasReconstructor, HillasIntersection]
+  reconstructor_types:
+    - HillasReconstructor
+      #  - HillasIntersection
```
maxnoe, line 100: Why commented out?

StFroese: It was recommended by @kosack 
> so perhaps comment-out the HillasIntersection line just to show that it can be enabled, but is not enabled by default

maxnoe: Ah ok, good then

maxnoe (approved):

StFroese:
> In light of this change, we should also review #2052 and perhaps consider making `ImPACTReconstructor` a subclass of `Reconstructor` again, and then it could be included in this list. It is dependent on a previous reconstructor, so some check would be needed to ensure that the user didn't forget to include it's seed reconstructor.

Perhaps we can give the base `Reconstructor` class an attribute `self.required_reconstructors` that defaults to `None` and in case of `ImPACT` or some other reconstructor contains a list of the required reconstructor names. We can check if the names are part of `self.reconstructor_types` inside the `__init__` of the ShowerProcessor and otherwise raise an exception.
This could be done in this PR.

StFroese:
So it results in the config that you suggested in the issue. Great, that makes sense to me.

maxnoe:
@StFroese Can you resolve the conflict? You might need to change more things than the conflict implies to get the correct behavior, since you moved the code.

ctapipe/reco/reco_algorithms.py
```diff
@@ -93,3 +94,18 @@ def _get_telescope_pointings(event):
...
+            shower_geom=event.dl2.stereo.geometry[self.__class__.__name__],
+            subarray=self.subarray,
+        )
+        for tel_id in event.trigger.tels_with_trigger:
+            tel_index = self.subarray.tel_indices[tel_id]
+            event.dl2.tel[tel_id].impact[
+                self.__class__.__name__
+            ] = TelescopeImpactParameterContainer(
+                distance=impact_distances[tel_index],
+                prefix=f"{self.__class__.__name__}_tel",
```
maxnoe, line 110: Here the default prefix should be added, see https://github.com/cta-observatory/ctapipe/pull/2060

StFroese:
> @StFroese Can you resolve the conflict? You might need to change more things than the conflict implies to get the correct behavior, since you moved the code.

Yes

maxnoe:
@StFroese Can you remove all the "HillasIntersection" things from the test configs where not explicitly wanted?

We shouldn't spend too much time in the CI making duplicate reconstructions

StFroese:
> @StFroese Can you remove all the "HillasIntersection" things from the test configs where not explicitly wanted?
> 
> We shouldn't spend too much time in the CI making duplicate reconstructions

Done

maxnoe (approved):

kosack (approved):

PR #2079, opened 2022-09-22
Allow data model version 4 in HDFEventSource, actually check which frame is used for hillas parameters

maxnoe:
This makes to small changes to the HDFEventSource, one to add support for the data model version 4.0, which I think is important since the change was so small and we are planning a release of mc data. 

Second, we no actually check which kind of hillas parameters are in the input file instead of relying on the version in which telescope frame became the default.

### DIFF
--- a/ctapipe/io/hdf5eventsource.py
+++ b/ctapipe/io/hdf5eventsource.py
@@ -367,10 +367,10 @@ class HDF5EventSource(EventSource):
                 }
 
         if DataLevel.DL1_PARAMETERS in self.datalevels:
+            hillas_cls = HillasParametersContainer
+            timing_cls = TimingParametersContainer
+
             if self._is_hillas_in_camera_frame():
-                hillas_cls = HillasParametersContainer
-                timing_cls = TimingParametersContainer
-            else:
                 hillas_cls = CameraHillasParametersContainer
                 timing_cls = CameraTimingParametersContainer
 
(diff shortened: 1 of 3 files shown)


ctapipe/io/hdf5eventsource.py
```diff
@@ -354,8 +367,7 @@ def _generator(self):
                 }
 
         if DataLevel.DL1_PARAMETERS in self.datalevels:
-            # FIXME: check units or config, not version. We have a switch.
-            if self.datamodel_version >= "v2.1.0":
+            if self._is_hillas_in_camera_frame():
```
nbiederbeck, line 370: it's the other way around, isn't it?

nbiederbeck (changes requested):

ctapipe/io/hdf5eventsource.py
```diff
@@ -613,6 +624,11 @@ def _generator(self):
                     for algorithm, readers in algorithms.items():
                         c[algorithm] = next(readers[key])
 
+                        # change prefix to new data model
```
LukasNickel, line 627: how is this `to new data model`?
It is only for the older version, right?

ctapipe/io/hdf5eventsource.py
maxnoe: Good catch

ctapipe/io/hdf5eventsource.py
maxnoe: If we have the old data model, we change the prefix to fit with the new version (so we don't lie when writing events out again).

nbiederbeck (approved):

LukasNickel (approved):

PR #2082, opened 2022-09-23
Fix invalid case in HillasReconstructor and HillasIntersection

maxnoe:
The PR #2073 did only change the "happy path" not  the invalid case, leading to the invalid container not being filled into the event.

### DIFF
--- a/ctapipe/reco/hillas_intersection.py
+++ b/ctapipe/reco/hillas_intersection.py
@@ -103,7 +103,9 @@ class HillasIntersection(Reconstructor):
         try:
             hillas_dict = self._create_hillas_dict(event)
         except (TooFewTelescopesException, InvalidWidthException):
-            return INVALID
+            event.dl2.stereo.geometry[self.__class__.__name__] = INVALID
+            self._store_impact_parameter(event)
+            return
 
         # Due to tracking the pointing of the array will never be a constant
         array_pointing = SkyCoord(
--- a/ctapipe/reco/hillas_reconstructor.py
+++ b/ctapipe/reco/hillas_reconstructor.py
@@ -149,7 +149,9 @@ class HillasReconstructor(Reconstructor):
         try:
             hillas_dict = self._create_hillas_dict(event)
         except (TooFewTelescopesException, InvalidWidthException):
-            return INVALID
+            event.dl2.stereo.geometry[self.__class__.__name__] = INVALID
+            self._store_impact_parameter(event)
+            return
 
         # Here we perform some basic quality checks BEFORE applying reconstruction
         # This should be substituted by a DL1 QualityQuery specific to this
--- a/ctapipe/reco/reco_algorithms.py
+++ b/ctapipe/reco/reco_algorithms.py
@@ -1,5 +1,6 @@
 from abc import abstractmethod
 
+import astropy.units as u
 import numpy as np
 from astropy.coordinates import AltAz, SkyCoord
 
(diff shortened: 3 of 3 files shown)


StFroese (commented):
Ok I see the problem, but why did `test_invalid_events` for the hillas reconstructor test not fail?
I don't really understand right now why `invalid_event.dl2.stereo.geometry` returns `Map(ctapipe.containers.ReconstructedGeometryContainer, {})` but `invalid_event.dl2.stereo.geometry["HillasReconstructor"]` yields
```
{'alt': <Quantity nan deg>,
 'alt_uncert': <Quantity nan deg>,
 'average_intensity': nan,
 'az': <Quantity nan deg>,
 'az_uncert': <Quantity nan deg>,
 'core_tilted_uncert_x': <Quantity nan m>,
 'core_tilted_uncert_y': <Quantity nan m>,
 'core_tilted_x': <Quantity nan m>,
 'core_tilted_y': <Quantity nan m>,
 'core_uncert_x': <Quantity nan m>,
 'core_uncert_y': <Quantity nan m>,
 'core_x': <Quantity nan m>,
 'core_y': <Quantity nan m>,
 'goodness_of_fit': nan,
 'h_max': <Quantity nan m>,
 'h_max_uncert': <Quantity nan m>,
 'is_valid': False,
 'telescopes': None}
```
I thought `Map` is a dictionary which should return `None` in this case.

ctapipe/reco/hillas_intersection.py
```diff
@@ -103,7 +103,9 @@ def __call__(self, event):
         try:
             hillas_dict = self._create_hillas_dict(event)
         except (TooFewTelescopesException, InvalidWidthException):
-            return INVALID
+            event.dl2.stereo.geometry[self.__class__.__name__] = INVALID
```
StFroese, line 106: What I mean is that you could basically just return `None` and do nothing else and  `test_HillasReconstructor.py::test_invalid_events` still works.
So this test only checks if the code successfully runs into the exception and doesn't check if the code fills `INVALID` into the event structure.

StFroese:
> I thought `Map` is a dictionary which should return `None` in this case.
Ah never mind. Of course `invalid_event.dl2.stereo.geometry["HillasReconstructor"]` is a new entry with `nan` elements

ctapipe/reco/hillas_intersection.py
StFroese: May be we can add something like `assert invalid_event.dl2.stereo.geometry` to the test to check if it was filled at all

maxnoe: Yes, the test coverage here was clearly lacking, I will add a test for the invalid case

maxnoe: this is again a case where the fact that Map is a defaultdict and not a dict basically made the test meaningless

maxnoe: We should probably change that...

StFroese: What was the original reason for using a defaultdict instead of a dict?

StFroese:
You can also remove the container from the docstring in the HillasReconstructor call since it is no longer returned. I forgot it.

nbiederbeck (dismissed):

StFroese (approved):

nbiederbeck (approved):

ctapipe/reco/hillas_intersection.py
maxnoe: Convenience in creating the event data structure.
Instead of doing this:

```
event.dl1.tel[tel_id] = DL1CameraContainer()
event.dl1.tel[tel_id].image = ...
```

You could just do `event.dl1.tel[tel_id].image = ...`.

However, this has also worse performance, so we should just change it and add the explicit containers where needed.

PR #2086, opened 2022-09-28
Update citations in README.rst

maxnoe:
The list kept getting out of date and by design, it is impossible to have the correct doi in the README for the release itself.

So I replaced the list of Zenodo records with a link which lists all versions on zenodo and the latest doi again.

I also added the ICRC proceeding.


kosack (approved):

README.rst
```diff
@@ -43,16 +27,30 @@ expert or developer!
...
+      doi = {10.22323/1.395.0744},
+      booktitle = {Proceedings, 37th International Cosmic Ray Conference},
+      year=2021,
+      volume={395},
+      number={744},
+      location={Berlin, Germany},
+    }
+
 
 Installation for Users
```
HealthyPear, line 55: I think we don't really need this part in the README when we can just point to the actual documentation

HealthyPear (approved):

README.rst
HealthyPear: (I mean the installation instructions - from the preview it seems I refer to the citation)

kosack: The documentation is more for developers, not for users (though probably that should be updated at some point to have both)

maxnoe: Let's merge this and I'll make another PR updating the installation instructions

nbiederbeck (approved):

PR #2092, opened 2022-09-30
Pixel monitoring

clara-escanuela:
Added new container and filled in simteleventsource


ctapipe/containers.py
```diff
@@ -570,6 +571,20 @@ class TelescopeImpactParameterContainer(Container):
...
+    nsb_rate = Field(np.int32(-1), "NSB rate in units of p.e. per ns"
+            )
+
+    qe_rel = Field(None, "Quantum efficiency of photon detectors"
+            )
+
+    hv_rel = Field(None, "High voltage")
+    current = Field(None, "Current")
+    fadc_amp_hg = Field(None, "FADC amplification.")
+
```
kosack, line 587: I suggest adding units on all of these Fields (where possible). For example:
```py3
current = Field(None, "Current", unit=u.amp)
```

ctapipe/containers.py
```diff
@@ -570,6 +571,20 @@ class TelescopeImpactParameterContainer(Container):
...
+    """
+    True information about nsb rate, quantum efficiency, high voltage, current and amplification settings for each pixel and telescope.
+    """
+    nsb_rate = Field(np.int32(-1), "NSB rate in units of p.e. per ns"
+            )
+
+    qe_rel = Field(None, "Quantum efficiency of photon detectors"
+            )
+
+    hv_rel = Field(None, "High voltage")
```
kosack, line 584: might want to add in the help string something about why this is "rel".  Does that mean relative? relative to what?

ctapipe/containers.py
```diff
@@ -619,6 +634,10 @@ class SimulatedCameraContainer(Container):
         description="true impact parameter",
     )
 
+    service = Field(
```
kosack, line 637: This should not be in the `SimulatedShowerContainer` because "service" information does not change per shower (anything under shower should be updated per shower).  Instead, it should be under `SimulationConfigContainer` I think (which in any case needs some re-structuring).

kosack (changes requested):

ctapipe/containers.py
kosack: At least that is if we consider this to be Configuration information, and not simulation service information.   If we really want it as Service info, we should start a new top-level container for that.   But I think this is really part of the input configuration of the simulations, so having it under `source.simulation_config` is fine. Where it goes in tthe data file is another option: it could be in either `/configuration/simulation/`  or `/simulation/service/`

ctapipe/io/tableloader.py
```diff
@@ -26,6 +26,7 @@
 TRUE_IMAGES_GROUP = "/simulation/event/telescope/images"
 TRUE_PARAMETERS_GROUP = "/simulation/event/telescope/parameters"
 TRUE_IMPACT_GROUP = "/simulation/event/telescope/impact"
+SERVICE_GROUP = "/simulation/event/telescope/service"
```
maxnoe, line 29: Service, by definition, is not per event. So this should be `/simulation/service/telescope`

maxnoe: or `/service/simulation/telescope`? I am not sure why the configuration is in `/configuration/simulation/run` and not in ``simulation/configuration/run`.

@kosack do we rules from the top level data model for this hierarchy?

ctapipe/containers.py
maxnoe: It's relative to the nominal high voltage, which should be in the camera configuration object.

Same for the `qe_rel`.

Maybe it makes sense to remove the `rel_` here completely and multiply with the value from the initial settings?

maxnoe:
```
   double nsb_pe_rate[H_MAX_PIX];  ///< NSB pixel p.e. rate [p.e./ns]
   double qe_rel[H_MAX_PIX];       ///< Assumed QE/PDE w.r.t. nominal
   double gain_rel[H_MAX_PIX];     ///< Assumed (PMT/common) gain w.r.t. nominal
   double hv_rel[H_MAX_PIX];       ///< Assumed high voltage w.r.t. nominal
   double current[H_MAX_PIX];      ///< Assumed current at pixel 
   double fadc_amp[H_MAX_GAINS][H_MAX_PIX]; ///< Assumed FADC amplitude per mean p.e.
   BYTE disabled[H_MAX_PIX];       ///< Pixel totally off and/or disabled in trigger.
```

kosack: > Maybe it makes sense to remove the rel_ here completely and multiply with the value from the initial settings?

Yes, I think having it be the absolute HV is more clear - the nominal doesn't really matter in practice.

ctapipe/io/tableloader.py
kosack: Yes, `/simulation/service` is correct. There is not top-level `service` package in the data model. 

However, what is considered service vs configuration is not so well defined... Particularly for simulations where the border between what is a configuration parameter and what is a piece of data is somewhat blurry.  E.g. is a QE curve part of the configuration? (probably), but is the per-pixel QE value then also configuration or output? Not sure - I agree here it may be service info, as it is not required to drive the simulation, it's more an output.

Perhaps let's look at it that way:
- things that are part of what the simulation producer would include when launching the sims are in `/configuration/simulation`
- things that are derived during the simulation are in `/simulation/service/`

Therefore this data (e.g. per pixel QE curves) is service data, since it is generaetd from the input QE curves (as a function of wavelength) convolved with the wavelength response of each PMT inside the simulation code.

ctapipe/containers.py
clara-escanuela: But source.simulation_config has no attribute "tel[tel_id]" and this data changes per telescope

maxnoe: @clara-escanuela you can add it :)

maxnoe: This would mean adding a field like this to the `SimulationConfigurationContainer`:

```
tel = Field(default_factory=partial(Map, TelescopeSimulationConfigurationContainer))
```

and defining the `TelescopeConfigurationContainer`. for now this can be what you now call "PixelMonitoring" or it can have that as a child field, I don't have a strong opininon here.

ctapipe/io/simteleventsource.py
```diff
@@ -1012,38 +1016,59 @@ def _parse_simulation_header(self):
...
+
+            laser_container = LaserCalibrationContainer(
+                    calib=calib,
+                    max_int_frac=max_int_frac,
+                    max_pixtm_frac=max_pix_tm_frac,
+                    tm_calib=tm_calib,
+                    flat_fielding=flat_fielding,
+                    )
+
+            monitor_id = (self.file_.camera_monitorings.get(tel_id, {}).get("monitor_id", None))
```
maxnoe, line 1047: You can just directly put it into the container:

LaserCalibrationContainer(
    calib=self.file.get....
)

clara-escanuela:
Coming back to this, What variables do you want to keep?

ctapipe/containers.py
```diff
@@ -570,6 +574,82 @@ class TelescopeImpactParameterContainer(Container):
...
+class CameraMonitoringContainer(Container):
+    """
+    Camera monitoring
+    """
+
+    monitor_id = Field(-1, "Monitoring id incremented with each update")
+    monitor_time = Field(None, "Time when last monitoring data was sent")
+    status_time = Field(None, "Status time")
+    trigger_time = Field(None, "Time when last trigger monitor data was read")
+    trigger_rate = Field(None, "Camera average local trigger rate in Hz")
```
kosack, line 619: I'd suggest adding units to all of these that are applicable rather than mentioning the unit in the help string, e.g.  
```py3
trigger_rate = Field(None, "Camera average local trigger rate", unit=u.s**-1)
data_rate = Field(-1, "Average rate of packed data", unit=u.Mbyte/u.s)
drawer_temp = Field(None, "Drawer temperature", unit=u.deg_C)
```

clara-escanuela:
The implementation sometimes is not very nice, for example in simteleventsource. Also, the merge of two tables may result in some repeated rows.

clara-escanuela:
I think we should close this PR. It is too outdated now. And maybe start a new one from scratch

PR #2093, opened 2022-10-06
Only use `setup.cfg`, automatically include correct python requires in docs

maxnoe:


kosack (changes requested):
Can you update the Makefile? I know it is not strictly required and could be removed, but I find it very useful to remind me of what commands to use to build the package.  I see there are many things to also remove in there.

maxnoe:
Done

setup.cfg
```diff
@@ -1,12 +1,102 @@
...
+[options]
+packages = find:
+python_requires = >=3.8
+zip_safe = False
+install_requires=
+	astropy~=5.0
+	bokeh~=2.0
+	docutils
+	eventio>=1.9.1<2.0.0a0
+	h5py
```
kosack, line 31: Do we still need h5py now that we do our own table writing and reading? I think it was only required for astropy's HDF5 output.

maxnoe: h5py is required to support reading tables written using the astropy mechanism using `ctapipe.io.read_table`. We have a unit test for that and I think it's a nice optional feature.

So I moved h5py from the required dependencies to the optional `tests` and `all` extras and left it in the environment.

Makefile
```diff
@@ -24,39 +24,40 @@ init:
...
 	find . -name "*.so" -exec rm {} \;
 	find . -name __pycache__ | xargs rm -fr
 
 test:
-	$(PYTHON) setup.py test
+	pytest
 
 doc:
-	cd docs && $(MAKE) html
+	cd docs && $(MAKE) html SPHINXOPTS="-W --keep-going -n --color -j auto"
```
nbiederbeck, line 36: With this change, can we change the CI to only `make doc`, i.e. without additionally specifying SPHINXOPTS *again*?

Makefile
```diff
@@ -24,39 +24,40 @@ init:
...
 
-conda:
-	python setup.py bdist_conda
-
 analyze:
 	@pylint ctapipe --ignored-classes=astropy.units
 
-pep8:
-	@pep8 --statistics
+flake8:
```
nbiederbeck, line 46: In addition to the above we can change the CI (and here) to e.g. `make lint` or similar?

Makefile
```diff
@@ -24,39 +24,40 @@ init:
...
+	@flake8 ctapipe
 
 env:
 	conda env create -n cta-dev -f environment.yml
 	source activate cta-dev
 
-trailing-spaces:
-	find $(PROJECT) examples docs -name "*.py" -exec perl -pi -e 's/[ \t]*$$//' {} \;
+develop:
+	pip install -e .
```
nbiederbeck, line 54: Doesn't make use a separate shell, thus this has no effect? (I'm curious)

ctapipe/utils/download.py
```diff
@@ -53,7 +55,7 @@ def download_file(url, path, auth=None, chunk_size=10240, progress=False):
                 for chunk in r.iter_content(chunk_size=chunk_size):
                     f.write(chunk)
                     pbar.update(len(chunk))
-        except:  # we really want to catch everythin here
+        except:  # noqa we really want to catch everythin here
```
nbiederbeck, line 58: everything

setup.cfg
```diff
@@ -1,12 +1,105 @@
...
+    pytest
+    pandas>=0.24.0
+    tomli
+    pytest_astropy_header
+	h5py
+
+
+docs =
+    sphinx~=3.5
+	jinja2~=3.0.2  # for sphinx 3.5, remove when upgrading sphinx
```
nbiederbeck, line 59: can this comment be translated into a dependency, as in line 32 for `imporlib_resources`?

Makefile
maxnoe: yes

Makefile
maxnoe: i like that

Makefile
maxnoe: This works...

setup.cfg
maxnoe: I don't think you can make this dependent on arbitrary versions, only on python versions

kosack (dismissed):

nbiederbeck (approved):

LukasNickel (approved):

PR #2096, opened 2022-10-11
Load observation info with TableLoader; standardize id fields in containers

LukasBeiske:
This adds an option to load the observation information from `/configuration/observation/observation_block` with TableLoader and join it to each event. 

This also standardizes the `obs_id`, `event_id` and `tel_id` fields used in some containers to remove mismatching descriptions which led to some warnings when merging different tables.

### DIFF
--- a/ctapipe/containers.py
+++ b/ctapipe/containers.py
@@ -71,6 +71,11 @@ NAN_TIME = Time(0, format="mjd", scale="tai")
 UNKNOWN_ID = np.uint64(np.iinfo(np.uint64).max)
 
 
+obs_id_field = partial(Field, UNKNOWN_ID, description="Observation Block ID")
+event_id_field = partial(Field, UNKNOWN_ID, description="Array Event ID")
+tel_id_field = partial(Field, UNKNOWN_ID, description="Telescope ID")
+
+
 class SchedulingBlockType(enum.Enum):
     """
     Types of Scheduling Block
@@ -161,8 +166,8 @@ class EventIndexContainer(Container):
     """index columns to include in event lists, common to all data levels"""
 
     default_prefix = ""  # don't want to prefix these
-    obs_id = Field(0, "observation identifier")
-    event_id = Field(0, "event identifier")
+    obs_id = obs_id_field()
+    event_id = event_id_field()
 
 
 class TelEventIndexContainer(Container):
@@ -172,9 +177,9 @@ class TelEventIndexContainer(Container):
     """
 
     default_prefix = ""  # don't want to prefix these
-    obs_id = Field(0, "observation identifier")
-    event_id = Field(0, "event identifier")
-    tel_id = Field(0, "telescope identifier")
+    obs_id = obs_id_field()
+    event_id = event_id_field()
+    tel_id = tel_id_field()
 
 
 class BaseHillasParametersContainer(Container):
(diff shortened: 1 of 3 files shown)


kosack (approved):

ctapipe/containers.py
```diff
@@ -71,6 +71,11 @@
 UNKNOWN_ID = np.uint64(np.iinfo(np.uint64).max)
 
 
+obs_id_field = partial(Field, UNKNOWN_ID, description="Observation Block ID")
+event_id_field = partial(Field, UNKNOWN_ID, description="Array Event ID")
+tel_id_field = partial(Field, UNKNOWN_ID, description="Telescope ID")
```
maxnoe, line 76: According to the data model, tel ids are uint16, not uint64 I think. Let me check.

maxnoe (changes requested):

ctapipe/containers.py
kosack: yes, they are - I missed that.  Since we don't expect even more than a few hundred values there, smaller storage saves some space.

kosack: `obs_ids` could even be  32 bit for that matter if we really want to save some space - here we do store them per event, so it could be useful

maxnoe (approved):

ctapipe/containers.py
maxnoe: The DL0 data model has them as uint64

LukasBeiske: I only changed it for `tel_id` for now.
Should I change `UNKNOWN_ID` to uint32 and introduce `UNKNOWN_EVENT_ID` as the uint64 default? `UNKNOWN_ID` is also used for the scheduling block id, but this wouldn't be a problem, would it?

maxnoe: But you are probably correct, ...

```
In [9]: (np.iinfo(np.uint32).max * 1 * u.min).to(u.year)
Out[9]: <Quantity 8165.95804814 yr>
```

maxnoe: Let's do what the data model does and ask Igor to consider changing it there

ctapipe/containers.py
```diff
@@ -1158,7 +1163,7 @@ class SimulatedShowerDistribution(Container):
 
     default_prefix = ""
 
-    obs_id = Field(-1, description="links to which events this corresponds to")
+    obs_id = obs_id_field()
     hist_id = Field(-1, description="Histogram ID")
```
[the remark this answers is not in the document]
StFroese: Ignore the first comment. I don't know what the data model requires here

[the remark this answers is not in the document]
maxnoe: > Also shower_primary_id in the SimulatedShowerContainer defaults to -1

So the default is a positron, which is kind of not intended.. but also something we do not simulate I think, so not that bad.

But yes, we should switch to something like `np.iinfo(np.int16).max` if that is not a real particle...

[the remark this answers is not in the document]
kosack: Yes, I think setting it to the max is fine - clearly we will never simulate that :-)  But it is possible we would simulate positrons at some point, so better to avoid -1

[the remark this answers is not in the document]
LukasBeiske: Done

StFroese (approved):

StFroese (approved):

Issue #1874, opened 2022-04-11
show all possible values in TelescopeParameters with (auto-generated) CaselessStrEnum traits


kosack:
**Please describe the use case that requires this feature.**

 `create_class_enum_trait()` can generate a configurable `CaseLessStrEnum` trait with all the possible values listed, e.g.

```
 --ImageProcessor.image_cleaner_type=<CaselessStrEnum>
     ImageCleaner to use.
     Choices: any of ['TailcutsImageCleaner', 'MARSImageCleaner', 'FACTImageCleaner'] (case-insensitive)
     Default: 'TailcutsImageCleaner'
```
This is very useful for letting the user know what options can be used.

However, this doesn't work when the trait is part of a TelescopeParameter, as in `CameraCalibrator.extractor_type`:

```
--CameraCalibrator.image_extractor_type=<telescopeparameter-item-1>...
    Name of the ImageExtractor subclass to be used.
    Default: [('type', '*', 'NeighborPeakWindowSum')]
```

**Describe the solution you'd like**

`TelescopeParameters` of `CaselessStrEnums` should show the possible Enum values in their help string.

PR #2102, opened 2022-10-21
Add and use ComponentName traitlet instead of create_class_enum_trait, fixes #2100, #1874

### LINKED ISSUE #2100: Move validation of class enum trait to run time instead of import time

Currently, we define class traitlets like this:

```
class Foo(Component):
    bar_type = create_class_enum_trait(Bar...).tag(config=True)
```

this runs at import time of Foo. Which makes it impossible to use subclasses of `Bar` that are only imported at a later stage.

This is problematic for subclasses defined in other submodules than the base class and the containing component.

THis was triggered by the machine learning module, where `ShowerProcessor` has a class enum trait of `Reconstructor` but because `ctapipe.ml.sklearn` cannot import `Reconstructor` without also importing `ShowerProcessor` (as `ctapipe.reco.__init__` imports it), the ml reconstructors are not yet defined when the class enum trait is created.

This is even more problematic for the case of plugins. 

Move the check if a provided string is an available class to the runtime (i.e. the validate method).

maxnoe:


StFroese (changes requested):

ctapipe/calib/camera/calibrator.py
```diff
@@ -53,19 +53,17 @@ class CameraCalibrator(TelescopeComponent):
         The name of the ImageExtractor subclass to be used for image extraction
     """
 
-    data_volume_reducer_type = create_class_enum_trait(
+    data_volume_reducer_type = ComponentName(
         DataVolumeReducer, default_value="NullDataVolumeReducer"
     ).tag(config=True)
 
     image_extractor_type = TelescopeParameter(
```
StFroese, line 60: This doesn't list the possible values. Also happens for other List traits, e.g. ShowerProcessor. Maybe implement a subclass of List ListComponentName which also returns the subclasses of the trait in the help string and use this for the ShowerProcessor and as the superclass of TelescopeParameter

StFroese (changes requested):

ctapipe/calib/camera/calibrator.py
kosack: Yes, I suggest calling it `ComponentNameList` though (it's a list of component names).  The current help string looks like this:
```
ShowerProcessor(Component) options
--ShowerProcessor.reconstructor_types=<list-item-1>...
    The stereo geometry reconstructors to be used.
    Default: ['HillasReconstructor']
```
Which is not so useful as you don't see what options are available. This is the same issue as #1874 for TelescopeParameters.

maxnoe:
I implemented the ComponentNameList and used it for the ShowerProcessor, here is the help now:

```
ShowerProcessor(Component) options
--ShowerProcessor.reconstructor_types=<componentnamelist-item-1>...
    The stereo reconstructors to be used. Possible values:
    ['HillasReconstructor', 'HillasIntersection']
    Default: ['HillasReconstructor']
```

maxnoe:
And with including the help of the basic trait in the help of the `TelescopeParameter`, it now looks like this for the image extractor:

```
--CameraCalibrator.image_extractor_type=<telescopeparameter-item-1>...
    Name of the ImageExtractor subclass to be used. The name of a ImageExtractor
    subclass. Possible values: ['FullWaveformSum', 'FixedWindowSum',
    'GlobalPeakWindowSum', 'LocalPeakWindowSum', 'SlidingWindowMaxSum',
    'NeighborPeakWindowSum', 'TwoPassWindowSum',
    'BaselineSubtractedNeighborPeakWindowSum']
    Default: [('type', '*', 'NeighborPeakWindowSum')]
```

StFroese (approved):

StFroese (approved):

kosack:
This also now fixes #1874

kosack (approved):

PR #2107, opened 2022-10-26
Support python 3.11

maxnoe:
Python 3.11 changed the behavior of the format string for a `class Foo(str, Enum)` but also introduced a new class `StrEnum` that keeps the old behavior.

I added a `ctapipe.compat` module that encapsulates the version dependent behavior needed for ctapipe and added the previous implementation for `StrEnum` there.

Numba also needs to be at least 0.57 for python 3.11.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -47,6 +47,10 @@ jobs:
             extra-args: ["codecov"]
 
           - os: ubuntu-latest
+            python-version: "3.11"
+            install-method: pip
+
+          - os: ubuntu-latest
             python-version: "3.10"
             install-method: pip
 
--- a/setup.cfg
+++ b/setup.cfg
@@ -33,7 +33,7 @@ install_requires=
     importlib_metadata  ; python_version < "3.10"
     joblib
     matplotlib ~=3.0
-    numba ~=0.56.0
+    numba >=0.56
     numpy ~=1.16
     psutil
     pyyaml >=5.1
--- a/ctapipe/compat.py
+++ b/ctapipe/compat.py
@@ -1,0 +1,17 @@
+"""
+Module for python version compatibility
+"""
+import sys
+
+__all__ = [
+    "StrEnum",
+]
+
+
+if sys.version_info >= (3, 11):
+    from enum import StrEnum
+else:
+    from enum import Enum
+
+    class StrEnum(str, Enum):
+        """Compatibility backfill of StrEnum for python < 3.11"""
--- a/ctapipe/instrument/optics.py
+++ b/ctapipe/instrument/optics.py
@@ -9,6 +9,7 @@ import astropy.units as u
 import numpy as np
 from astropy.table import QTable
 
+from ..compat import StrEnum
 from ..utils import get_table_dataset
 
 logger = logging.getLogger(__name__)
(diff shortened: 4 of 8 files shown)


maxnoe:
Seems we have to wait for a numba release adding python 3.11 support

kosack:
just updated this to master - let's see if there are still issues

maxnoe:
Yes, still no numba for 3.11

maxnoe:
Tests are now passing using the 0.57rc1 nunba pre release

.github/workflows/ci.yml
```diff
@@ -103,6 +107,11 @@ jobs:
         run: |
           tee -a ~/.bash_profile <<<'export PATH="$pythonLocation/bin:$PATH"'
 
+      - if: ${{ matrix.python-version == '3.11' }}
+        # install numba pre-release for python 3.11
+        run: |
+          pip install numba==0.57.0rc1
```
Tobychev, line 113: Is there a way to make this "0.57rc1 or later"?

ctapipe/tools/apply_models.py
```diff
@@ -201,7 +201,7 @@ def _apply(self, reconstructor, tel_tables, start, stop):
                 write_table(
                     table[output_columns],
                     self.output_path,
-                    f"/dl2/event/telescope/{prop}/{prefix}/tel_{tel_id:03d}",
+                    f"/dl2/event/telescope/{prop.value}/{prefix}/tel_{tel_id:03d}",
```
Tobychev, line 204: How is this related to the python 3.11 support?

maxnoe: Checking the chabgelog, there is actually  now
https://docs.python.org/3/library/enum.html#enum.StrEnum

Which might be a better solution, will change

.github/workflows/ci.yml
maxnoe: Yes, but I think I'd rather wait until the release is out and just remove this workaround then

Tobychev (approved):

.github/workflows/ci.yml
maxnoe: Numba 0.57 has been released (at least on PyPI, the conda-forge release is underway), so this was removed

maxnoe:
I am not sure if the CI system allows for such conclusions, but: the python 3.11 test suite ran one minute faster than the python 3.10 tests (11 min vs. 10 min). Which would be a really nice improvement for just updating to a new python minor version release!

Tobychev (approved):

setup.cfg
```diff
@@ -33,7 +33,7 @@ install_requires=
     importlib_metadata  ; python_version < "3.10"
     joblib
     matplotlib ~=3.0
-    numba ~=0.56.0
+    numba >=0.56
```
LukasNickel, line 36: You wrote that we need at least numba 0.57, is that not correct anymore?

kosack (approved):

setup.cfg
maxnoe: For 3.11, 3.10 and below is fine with 0.56

maxnoe: And since versions 0.56 correctly specifies that it is not compatible with 3.11, there is no issue.

The conda-forge package for 0.57 is not yet out, so prescribing 0.57 is not a good idea.

PR #2109, opened 2022-10-27
Allow overriding url via env variable

maxnoe:


ctapipe/utils/datasets.py
```diff
@@ -25,14 +25,28 @@
...
 DEFAULT_URL = "http://cccta-dataserver.in2p3.fr/data/ctapipe-test-data/v1.1.0/"
 
 
-def get_searchpath_dirs(searchpath=None, url=DEFAULT_URL):
+def get_default_url():
+    """Get the default download url for datasets
+
+    First tries to look-up CTAPIPE_DOWNLOAD_URL and then falls back to `DEFAULT_URL`
+    """
+    return os.getenv("CTAPIPE_DOWNLOAD_URL", DEFAULT_URL)
```
kosack, line 42: Maybe call it CTAPIPE_DATASET_URL so that it's clear it is linked to the dataset:// input url pattern

kosack (changes requested):

ctapipe/utils/datasets.py
maxnoe: done

kosack (dismissed):

ctapipe/utils/datasets.py
```diff
@@ -25,14 +25,28 @@
...
+    "get_dataset_path", "find_in_path", "find_all_matching_datasets",
+    "get_default_url",
+]
 
 
 DEFAULT_URL = "http://cccta-dataserver.in2p3.fr/data/ctapipe-test-data/v1.1.0/"
 
 
-def get_searchpath_dirs(searchpath=None, url=DEFAULT_URL):
+def get_default_url():
```
LukasNickel, line 37: I dont think this should be called `get_default_url` if it does more than returning `DEFAULT_URL`.
`get_dataset_url` to be consistent with `get_dataset_path`?

maxnoe: It returns the default for the functions.

E.g. all the functions have `url=None` and if `url is None` they do `get_default_url`. So it is indeed the default url.

LukasNickel: Then maybe remove the variable and put it into the function as its not used outside anymore?

LukasNickel (approved):

kosack (approved):

PR #2121, opened 2022-11-16
Allow using predicted energy in particle classifier

maxnoe:
Currently fails in `apply_models`

### DIFF
--- a/ctapipe/tools/apply_models.py
+++ b/ctapipe/tools/apply_models.py
@@ -112,11 +112,11 @@ class ApplyModels(Tool):
         self.loader = TableLoader(
             parent=self,
             h5file=self.h5file,
-            load_dl1_images=False,
             load_dl1_parameters=True,
             load_dl2=True,
-            load_simulated=True,
             load_instrument=True,
+            load_dl1_images=False,
+            load_simulated=False,
         )
 
         self._reconstructors = []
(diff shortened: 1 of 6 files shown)


maxnoe:
The reason seems to be that the `TableLoader` doesn't read the newly added tables for some reason. I tried adding `self.h5file.flush()`, but that wasn't enough, I had to completely reopen the file.

That seems not like a nice solution, but works for now.

kosack:
> Does HDF5 support concurrent access to a single dataset from multiple processes?

>> If all processes are reading, then, yes, HDF5 (serial) does support this. If there are any processes that are writing, then no, this is not supported. We are working on a "Single Write Multiple Read" (SWMR) feature, which will be available in a future release (expected to be in HDF5-1.10).

There does seem to be some development toward "parallel HDF5", but it seems to not be in the core library

nbiederbeck (approved):

kosack (approved):

ctapipe/tools/apply_models.py
```diff
@@ -146,6 +146,9 @@ def start(self):
                 reconstructor,
             )
             self._combine(reconstructor.stereo_combiner, mono_predictions)
+            self.h5file.close()
```
kosack, line 149: You might want to add a comment here to remind a future developer why this close/open was added, otherwise it may get removed.

maxnoe: I am not really happy yet, there should be a better solution...

kosack (approved):

nbiederbeck (approved):

PR #2128, opened 2022-11-22
Store telescope multiplicity in stereo containers

StFroese:


kosack:
I don't think multiplicity on its own is sufficient.  In protopipe, we found that having an overall multiplicity + the individual multiplicities of each telescope type is useful (e.g. `multiplicity_SST`).  

However all of that information is already in the telescope pattern, and we simply used that to derive the needed features.  Is that not possible now? Wouldn't it be possible to do something like:

```yaml
FeatureGenerator:
       features:
         - ["hillas_multiplicity", "sum(HillasReconstructor_telescopes)"]
```

We could include a function to count by telescope type as well

StFroese:
So we have to store something like `telescope_types=np.array([subarray.tel[tel_id].type.value for tel_id in subarray.tel.keys()])` to the stereo container or is there a simpler solution?
And then just something like (very roughly):
```python
def multiplicity(telescopes, telescope_type=None, type=None):
    if telescope_type is None or type is None:
        return np.sum(telescopes)
    else:
        tel_mask = (telescope_types == type)
        return np.sum(telescopes[tel_mask])
```

StFroese:
This code won't for telescopes being the list of tel_id's and not the mask. So question now is: for which cases is telescopes a mask and not list of tel_id's and why are there two possibilities?

maxnoe:
Indeed, this is problematic. The reason for the mask is simple: storing variable length arrays in most data formats is not possible or hard or not supported by libraries (e.g. hdf5 does allow it in principle, but it's finicky in C and not supported by pytables). astropy tables have the same issue in-memory.

So we convert the list of integers to a mask when storing it in hdf5 files, the hdf5eventsource converts it back but the tableloader loads the masks, since that's the only thing supported by astropy. We could decide to only use masks, but we have code where we need the ids, so it's slower to always convert the mask to ids when not needed.

StFroese:
How about writing the mask and the tel_ids to the container? It would be kind of redundant but it's a clearer definition than having one parameter that could be either of them.

StFroese:
> That provides much more flexibility, as we can define different types of multiplicity-like parameters

```yaml
FeatureGenerator:
       features:
         - ["hillas_multiplicity", "sum(HillasReconstructor_telescopes)"]
         - ["hillas_multiplicity_LST", "sum(HillasReconstructor_telescopes[subarray_type=='LST'])"]
```
This should work...
(is subarray_type the correct key for the table?)

StFroese:
@maxnoe @kosack it's more or less a design decision now whether to store the multiplicities in the containers or generate them as a feature.
I personally like the feature generation approach.

kosack:
I think I prefer it as a feature-generation approach.  The required information is already stored in the files, and every time we add more fields, the file size increases and the data model becomes more complex, so best to store the minimum and generate what is needed (CPU time is currently quite a bit cheaper than storage for CTA right now...).

PR #2131, opened 2022-11-23
Support different frames for set_line_hillas

StFroese:
Works now for other frames, e.g. `EastingNorthing`
Screenshot 2022-11-23 at 16 39 09

### DIFF
--- a/ctapipe/visualization/mpl_array.py
+++ b/ctapipe/visualization/mpl_array.py
@@ -2,7 +2,7 @@ from itertools import cycle
 
 import numpy as np
 from astropy import units as u
-from astropy.coordinates import Angle
+from astropy.coordinates import Angle, SkyCoord
 from matplotlib import pyplot as plt
 from matplotlib.collections import PatchCollection
 from matplotlib.lines import Line2D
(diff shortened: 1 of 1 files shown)


kosack (approved):

kosack:
Note that you seem to to have selected the LaPalma alpha telescope id list but applied it to the Paranal array, so that is why you have such a strange-looking array :-)  You should use the Paranal-alpha list!

kosack:
Also, I notice there is still a bug: the points for the line are not correctly transformed

StFroese:
Oh your right

StFroese:
> Note that you seem to to have selected the LaPalma alpha telescope id list but applied it to the Paranal array, so that is why you have such a strange-looking array :-) You should use the Paranal-alpha list!

I didn't select anything (as far as I know). It's just the file:
```python
source.subarray.info()
```
```
Subarray : Paranal-prod6_1-11,19,35
Num Tels : 13
Footprint: 0.59 km2

      Type       Count  Tel IDs  
---------------- ----- ----------
  LST_LST_LSTcam     4 1-4       
MST_MST_FlashCam     9 5-11,19,35
```

PR #2133, opened 2022-11-23
Chunked model application

maxnoe:


ctapipe/io/tableloader.py
```diff
@@ -378,6 +380,8 @@ def read_subarray_events_chunked(self, chunk_size):
             self.read_subarray_events,
             n_total=len(self),
             chunk_size=chunk_size,
+            args=args,
+            kwargs=kwargs,
```
StFroese, line 384: Why explicit and not `*args` and `**kwargs`?

ctapipe/io/tableloader.py
```diff
@@ -667,5 +674,76 @@ def read_telescope_events_by_type_chunked(self, chunk_size, **kwargs):
...
+
+        for tel_id in by_id.keys():
+            by_id[tel_id] = self._join_subarray_info(
+                by_id[tel_id], subarray_events=subarray_events
+            )
+            self._sort_to_original_order(by_id[tel_id], include_tel_id=True)
+
+        return by_id
+
+    def read_telescope_events_by_id_chunked(self, chunk_size, *args, **kwargs):
```
StFroese, line 730: Very similar to `read_telescope_events_by_type_chunked`. Perhaps one function would be sufficient
```python
     def read_telescope_events_by_chunked(self, chunk_size, events_by, *args, **kwargs):       
         """                                                                                   
         Iterate over chunks of telescope events and return a dict of one table per telescope id.    
                                                                                               
         Parameters                                                                            
         ----------                                                                            
         chunk_size: int                                                                       
             Number of subarray events to load per chunk.                                      
             The telescope tables might be larger or smaller than chunk_size                   
             depending on the selected telescopes.                                             
                                                                                               
         *args, **kwargs are passed to `read_telescope_events_by_id`                           
         """                                                                                   
         return ChunkIterator(                                                                 
             self.__getattribute__(f'read_telescope_events_by_{event_by}),                    
             n_total=len(self),                                                                
             chunk_size=chunk_size,                                                            
             args=args,                                                                        
             kwargs=kwargs,                                                                    
         )              
```

ctapipe/io/tableloader.py
maxnoe: Because you then have to mix args to the called function with kwargs to the ChunkIterator as you cannot give args after kwargs.

maxnoe:
This is successful in that it reduced memory consumption a lot, but it's also  a lot slower and I think it doesn't make sense to merge until we remove the bottlenecks from the table loader for that. I will try to incorporate this.

ctapipe/io/tableloader.py
maxnoe: It's more about having the proper docstring and being discoverable in the documentation, I'd say. The code duplicated is very little.

maxnoe:
Ok, main problem was that adding the telescope trigger information naively was *very* slow, since it required loading the whole trigger telescope for each chunk.

Now, application is *much* faster (even a lot faster than non-chunked, since joining N chunks N times is faster than joining the whole table).

kosack:
Looks good! The effect on memory useage is clear.  I tried here some extreme cases: basically using 100k events (blue) vs 1k events (black): 
image
Of course the optimal for speed is clearly at the higher end, but it's nice we can control the memory usage so well

ctapipe/io/tableloader.py
StFroese: ok

kosack (approved):

StFroese (approved):

PR #2136, opened 2022-11-25
Implement stereo trigger for subarray triggers

maxnoe:


ctapipe/instrument/trigger.py
```diff
@@ -0,0 +1,97 @@
...
+
+    def __init__(self, subarray, *args, **kwargs):
+        super().__init__(subarray, *args, **kwargs)
+
+        self._ids_by_type = {
+            str(type): set(self.subarray.get_tel_ids_for_type(type))
+            for type in self.subarray.telescope_types
+        }
+
+        print("SOFTWARE_TRIGGER", self.min_telescopes, self.min_telescopes_of_type)
```
nbiederbeck, line 47: log?

ctapipe/instrument/trigger.py
```diff
@@ -0,0 +1,97 @@
...
+
+    def __call__(self, event: ArrayEventContainer) -> bool:
+        """
+        Remove telescope events that have not the required number of telescopes of
+        a given type from the subarray event and decide if the event would
+        have triggered the stereo trigger.
```
nbiederbeck, line 53: ~~stereo~~ software?  Otherwise this seems very LST-specific.

ctapipe/instrument/trigger.py
```diff
@@ -0,0 +1,97 @@
...
+    A stereo trigger that can remove telescope events from subarray events.
+
+    This class is needed to correctly handle super-arrays as simulated for
+    CTA and still handle the LST hardware stereo trigger and the normal stereo
+    trigger correctly.
+
+    LSTs have a hardware trigger, that requires at least two LSTs to have a trigger
+    in each event. Thus, after selecting a subarray, it is needed to
+    - Remove single LST telescope events from events
+    - Ignore events with only 1 telescope after this has been applied
```
nbiederbeck, line 17: This can do even more, right? So this is basically an example of what to do with this SoftwareTrigger?  My proposal is to add `Example\n-------` before the last paragraph.

ctapipe/instrument/trigger.py
```diff
@@ -0,0 +1,97 @@
...
+    """
+
+    min_telescopes = Integer(
```
nbiederbeck, line 20: I am not happy with the naming here, as only the help of the other parameter shows, that *after* applying other checks this one is used.  I was very confused when I read the tests (which I did before reading this file), that `min_telescopes` is (intentionally) lower than `sum(min_telescopes_of_type)`

ctapipe/instrument/trigger.py
maxnoe: debug print, no logging required at all..

ctapipe/instrument/trigger.py
maxnoe: This sentence refers to the general stereo trigger of the array, which is not LST specific.

nbiederbeck: Indeed, I was biased by my other comment.  But this leads to another question:  when I read "stereo" (instead of "mono") I think of at least two (>=2), this is not needed here, right? By default a single telescope can trigger this SoftwareTrigger?

ctapipe/instrument/trigger.py
```diff
@@ -0,0 +1,97 @@
...
+        triggered : bool
+            Whether or not this event would have triggered the stereo trigger
+        """
+
+        for tel_type in self.subarray.telescope_types:
+            tel_type_str = str(tel_type)
+            min_tels = self.min_telescopes_of_type.tel[tel_type_str]
+
+            # no need to check telescopes for which we have no min requirement
+            if min_tels == 1:
```
nbiederbeck, line 68: I think the default should be `0`, one could require a single MST for example

ctapipe/instrument/trigger.py
kosack: Perhaps we should explain that In the current sims, the trigger is not really correctly simulated due to the use of super-arrays (and user-selected subarrays).  Therefore, it is still possible to have a triggered event that contains e.g. 1 MST, if there were other MSTs that are not in the currently selected subarray that were triggered.  This  component can also help to simulate a more realistic array trigger, which would reject those events.  But normally, a multiplicity cut of 2 at the reconstruction level is sufficient (but at DL1 those events would still remain, unrealistically). That in particular does not work for the LSTs, as you could get a multiplicity-2 event with  1 LST+1 MST that would be rejected by the real LST trigger. Therefore, having a min multiplicity per type is needed.

ctapipe/instrument/trigger.py
kosack: yeah, I think that wording is fine: there are 2 triggeres: the subarray trigger (SWAT) and the LST-Hardware Stereo Trigger.  Both are examples of stereo triggers, just different implementations.

ctapipe/instrument/trigger.py
maxnoe: ok, yes

ctapipe/instrument/trigger.py
```diff
@@ -0,0 +1,103 @@
...
+    correctly simulated as in that after selecting a realistic subarray, events
+    are still in the data stream where only one telescope of the selected subarray
+    triggered, which would in reality not trigger the stereo trigger.
+
+    An additional complexity is the LST hardware stereo trigger, that forces that
+    an array event has always to contain no or at least two LST telescope events.
+
+    This means that after selectig a subarray, we need to:
+    - Remove LST telescope events from the subarray if only one LST triggered
+    - Ignore events with only 1 telescope after this has been applied
```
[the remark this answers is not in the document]
maxnoe: The default of the class is not to do anything! I will add an example section with the required config for cta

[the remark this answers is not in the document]
nbiederbeck: Yes, and that is good! But "we need to" reads as if this it what happens when using the class as is

nbiederbeck (approved):

ctapipe/instrument/trigger.py
```diff
@@ -0,0 +1,113 @@
...
+            tel_type_str = str(tel_type)
+            min_tels = self.min_telescopes_of_type.tel[tel_type_str]
+
+            # no need to check telescopes for which we have no min requirement
+            if min_tels == 0:
+                continue
+
+            tels_with_trigger = set(event.trigger.tels_with_trigger)
+            tel_ids = self._ids_by_type[tel_type_str]
+            tels_in_event = tels_with_trigger.intersection(tel_ids)
```
StFroese, line 89: very nice use of sets

StFroese (approved):

ctapipe/instrument/tests/test_trigger.py
```diff
@@ -0,0 +1,52 @@
...
+
+
+def test_software_trigger(subarray_prod5_paranal):
+    from ctapipe.instrument.trigger import SoftwareTrigger
+
+    subarray = subarray_prod5_paranal
+    trigger = SoftwareTrigger(
+        subarray=subarray,
+        min_telescopes=2,
+        min_telescopes_by_type=[
```
StFroese, line 11: `min_telescopes_of_type`

StFroese (changes requested):

StFroese (approved):

nbiederbeck (approved):

Issue #2116, opened 2022-11-08
General high level API for `Reconstructor`


maxnoe:
**Please describe the use case that requires this feature.**

In the current master, the only required API of `Reconstructor` is `__call__(self, ArrayEvent)`. The `SKLearnReconstructors` adds to this the `predict_table` method, which performs monoscopic prediction given a key, which at this time must be the `TelescopeDescription`.

It would be nice to have general interface for `Reconstructor`, so that the `ctapipe-apply-models` and `ctapipe-process` tools can apply all valid subclasses of `Reconstructor` by just being given the config and / or a path to read a pickled reconstructor from.

The problem here is that the `predict_table(key, telescope_table)` is specific to reconstructors predicting on one telescope table at a time, i.e. this is not feasible for e.g. the `HillasReconstructor` or machine learning algorithms directly predicting stereo results.

It also does not work with reconstructors able to predict multiple properties (i.e. ImPACT) at the same time.

**Describe the solution you'd like**

Define the interface of `Reconstructor` so that use cases are covered and `ctapipe-process` can work event-wise with all reconstructors and `ctapipe-apply` can work in bulk with all reconstructors.

One possibility would be to have more methods, that can be implemented by reconstructors.

WIP:

```
class Reconstructor:
    def __call__(self, event: ArrayEvent):
        """As before: predict and fill prediction into the given ArrayEvent"""

    def predict_table(self, table) -> tuple[dict[str, Table], dict[str, dict[int, Table]]]:
        """
        Predict on a table as returned by TableLoader.read_telescope_events()

        Returns
        -------
        array_predictions :
            dict mapping properties to tables of stereo predictions, e.g.:
            ``{"energy": Table(), "geometry": Table()}``
        telescope_predictions :
            dict mapping properties to dicts mapping tel_ids to tables of mono predictions, e.g.:
            ``{"energy": {1: Table(), 2: Table(), }}```
        """
```

The problem here is that this would not work for reconstructors needing images, as a single table cannot hold the differently shaped images in the same table.

**Describe alternatives you've considered**

Not really any.

kosack:
We might not really need to have reconstructors like ImPACT be "apply-able", as those can just run in ctapipe-process. It's really only ML-based prediction reconstructions that  need to do a test/train cycle where ctapipe-apply is very useful. Even i a final analysis, the predict can be done in the ctapipe event loop however, once models exist. 

[...]

PR #2138, opened 2022-11-25
Disp reconstruction

LukasBeiske:
This is a continuation of #2064 after #1767 was merged.

For now I only implemented a simple weighted average to combine single telescope predictions into event predictions. The estimation of the error of this average seems to underestimate the error a bit, based on a small MC study I did. 
I will look into other ways to combine the telescope predictions in the future (including other error estimates), but for now this is better than no error at all, in my opinion.


ctapipe/coordinates/disp.py
```diff
@@ -0,0 +1,44 @@
+"""
```
maxnoe, line 1: This module really has nothing to do with disp per-se, so I'd not call it "disp.py".

Maybe these functions are better at `ctapipe.reco.preprocessing`?

ctapipe/coordinates/disp.py
```diff
@@ -0,0 +1,44 @@
...
+        horizontal_coord = SkyCoord(alt=alt, az=az, frame=AltAz())
+        pointing = SkyCoord(alt=pointing_alt, az=pointing_az, frame=AltAz())
+        tel_frame = TelescopeFrame(telescope_pointing=pointing)
+
+        tel_coord = horizontal_coord.transform_to(tel_frame)
+
+    return tel_coord.fov_lon.to(u.deg), tel_coord.fov_lat.to(u.deg)
+
+
+def telescope_to_horizontal(
```
maxnoe, line 31: should add `u.quantity_input` to these functions

ctapipe/reco/sklearn.py
```diff
@@ -507,6 +512,353 @@ def predict_table(self, key, table: Table) -> Table:
...
+    sign_config = Dict({}, help="kwargs for the sklearn classifier").tag(config=True)
+    sign_cls = Enum(
+        SUPPORTED_CLASSIFIERS.keys(),
+        default_value=None,
+        allow_none=True,
+        help="Which scikit-learn classification model to use.",
+    ).tag(config=True)
+
+    sign_invalid_class = Integer(
+        default_value=0,
```
maxnoe, line 548: Why not nan?

ctapipe/reco/sklearn.py
```diff
@@ -507,6 +512,353 @@ def predict_table(self, key, table: Table) -> Table:
...
+
+    def __init__(self, subarray, norm_models=None, sign_models=None, **kwargs):
+        # Run the Component __init__ first to handle the configuration
+        # and make `self.load_path` available
+        Component.__init__(self, **kwargs)
+
+        if self.load_path is None:
+            if self.norm_cls is None or self.sign_cls is None:
+                raise TraitError(
+                    "Must provide `norm_cls` and `sign_cls` if not loading form file"
```
maxnoe, line 572: typo: form

ctapipe/reco/sklearn.py
```diff
@@ -507,6 +512,353 @@ def predict_table(self, key, table: Table) -> Table:
...
+                )
+
+            if subarray is None:
```
maxnoe, line 575: How can this happen? subarray is a posarg in `__init__`

ctapipe/reco/sklearn.py
```diff
@@ -507,6 +512,353 @@ def predict_table(self, key, table: Table) -> Table:
...
+                parent=self,
+            )
+        else:
+            loaded = self.read(self.load_path)
+            if (
+                subarray is not None
+                and loaded.subarray.telescope_types != subarray.telescope_types
+            ):
+                self.log.warning(
+                    "Supplied subarray has different telescopes than subarray loaded form file"
```
maxnoe, line 604: typo: form

ctapipe/reco/sklearn.py
```diff
@@ -507,6 +512,353 @@ def predict_table(self, key, table: Table) -> Table:
...
+                )
+            self.__dict__.update(loaded.__dict__)
+            self.subarray = subarray
+
+    def _new_models(self):
+        norm_regressor = SUPPORTED_REGRESSORS[self.norm_cls](**self.norm_config)
+        sign_classifier = SUPPORTED_CLASSIFIERS[self.sign_cls](**self.sign_config)
+        return norm_regressor, sign_classifier
+
+    def _table_to_X(self, table):
```
maxnoe, line 614: How is this different from the other classes? if it is not different, factor out into a free function.

ctapipe/reco/sklearn.py
```diff
@@ -507,6 +512,353 @@ def predict_table(self, key, table: Table) -> Table:
...
+        """
+        Extract features as numpy ndarray to be given to sklearn from input table
+        """
+        feature_table = table[self.features]
+        valid = check_valid_rows(feature_table, log=self.log)
+        X = table_to_float(feature_table[valid])
+        return X, valid
+
+    def _table_to_y(self, table, mask=None):
```
maxnoe, line 623: same

ctapipe/reco/sklearn.py
```diff
@@ -507,6 +512,353 @@ def predict_table(self, key, table: Table) -> Table:
...
+                f" available regressors: {self._norm_models.keys()}"
+            )
+        if key not in self._sign_models:
+            raise KeyError(
+                f"No classifier available for key {key},"
+                f" available classifiers: {self._sign_models.keys()}"
+            )
+        X, valid = self._table_to_X(table)
+        norm_prediction = np.full(len(table), np.nan)
+        sign_prediction = np.full(len(table), self.sign_invalid_class, dtype=np.int8)
```
maxnoe, line 696: Ah! forget comment above.  But I think the option is not necessary, filling 0 for unknown sign shouldn't something we want to change, right?

ctapipe/reco/sklearn.py
```diff
@@ -507,6 +512,353 @@ def predict_table(self, key, table: Table) -> Table:
...
+                disp_container = DispContainer(
+                    norm=norm[0],
+                    sign=sign[0],
+                    parameters_is_valid=valid[0],
+                )
+
+                if valid:
+                    disp = norm * sign
+
+                    fov_lon = event.dl1.tel[
```
maxnoe, line 740: This code would be much more readable if you did something like this:

```
hillas = event.dl1.tel[tel_id].hillas
```

and then used `hillas` from there on

ctapipe/reco/sklearn.py
```diff
@@ -507,6 +512,353 @@ def predict_table(self, key, table: Table) -> Table:
...
+                        tel_id
+                    ].parameters.hillas.fov_lon + disp * np.cos(
+                        event.dl1.tel[tel_id].parameters.hillas.psi.to(u.rad)
```
maxnoe, line 743: converting to `rad` should not be needed! Either just give psi, or if it is for speed, convert `psi` once to  a plain number like this:

```
psi = hillas.psi.to_value(u.rad)
```

note the `to_value` instead of `to`

ctapipe/reco/sklearn.py
```diff
@@ -507,6 +512,353 @@ def predict_table(self, key, table: Table) -> Table:
...
+        add_defaults_and_meta(
+            disp_result,
+            DispContainer,
+            prefix=self.prefix,
+            stereo=False,
+        )
+
+        disp = norm * sign
+
+        fov_lon = table["hillas_fov_lon"] + disp * np.cos(table["hillas_psi"].to(u.rad))
```
maxnoe, line 828: same here, just using `.quantity` should be enough or convert once before (faster)

ctapipe/containers.py
```diff
@@ -843,6 +849,19 @@ class ParticleClassificationContainer(Container):
...
 
 
+class DispContainer(Container):
+    """
+    Standard output of disp reconstruction algorithms for origin reconstruction
+    """
+
+    default_prefix = "disp"
+
+    norm = Field(nan * u.deg, "reconstructed absolute value for disp", unit=u.deg)
```
maxnoe, line 859: Why don't we just store `norm` with the sign in one float? That would use less space and still have all the information.

ctapipe/reco/sklearn.py
```diff
@@ -575,19 +934,43 @@ def __call__(self, telescope_type, table):
...
+                    truth_sign,
+                    metrics,
+                ) = self.cross_validate(telescope_type, train, test)
+
+                predictions.append(
+                    Table(
+                        {
+                            "cv_fold": np.full(len(truth_norm), fold, dtype=np.uint8),
+                            "tel_type": [str(telescope_type)] * len(truth_norm),
+                            "norm_prediction": cv_prediction_norm,
```
maxnoe, line 952: same here: just use one column each for prediction and truth. Using two columns does not provide additional information.

ctapipe/reco/stereo_combination.py
```diff
@@ -205,6 +213,60 @@ def _combine_classification(self, event):
...
+                dl1 = event.dl1.tel[tel_id].parameters
+                weights.append(self._calculate_weights(dl1) if dl1 else 1)
+                ids.append(tel_id)
+
+        if len(alt_values) > 0:  # by construction len(alt_values) == len(az_values)
+            coord = SkyCoord(
+                alt=alt_values,
+                az=az_values,
+                frame=AltAz(),
+            )
```
StFroese, line 236: `AltAz(alt=alt_values, az=az_values)`

ctapipe/reco/stereo_combination.py
```diff
@@ -205,6 +213,60 @@ def _combine_classification(self, event):
...
+            # https://en.wikipedia.org/wiki/Von_Mises%E2%80%93Fisher_distribution#Mean_direction
+            mono_x, mono_y, mono_z = coord.cartesian.get_xyz()
+            stereo_x = np.average(mono_x, weights=weights)
+            stereo_y = np.average(mono_y, weights=weights)
+            stereo_z = np.average(mono_z, weights=weights)
+
+            mean_cartesian = CartesianRepresentation(x=stereo_x, y=stereo_y, z=stereo_z)
+            mean_spherical = mean_cartesian.represent_as(SphericalRepresentation)
+            mean_altaz = SkyCoord(mean_spherical, frame=AltAz())
```
StFroese, line 245: `AltAz`

ctapipe/reco/stereo_combination.py
```diff
@@ -205,6 +213,60 @@ def _combine_classification(self, event):
...
+
+            # https://en.wikipedia.org/wiki/Directional_statistics#Measures_of_location_and_spread
+            r = mean_spherical.distance.to_value()
+            std = np.sqrt(-2 * np.log(r))
+
+            valid = True
+        else:
+            mean_altaz = SkyCoord(
```
StFroese, line 253: `AltAz`

ctapipe/reco/stereo_combination.py
```diff
@@ -294,6 +359,57 @@ def predict_table(self, mono_predictions: Table) -> Table:
             )
             stereo_table[f"{self.prefix}_is_valid"] = np.isfinite(stereo_energy)
             stereo_table[f"{self.prefix}_goodness_of_fit"] = np.nan
+
+        elif self.property == "geometry":
+            if len(valid_predictions) > 0:
+                coord = SkyCoord(
```
StFroese, line 365: `AltAz`

ctapipe/reco/stereo_combination.py
```diff
@@ -294,6 +359,57 @@ def predict_table(self, mono_predictions: Table) -> Table:
...
+                )
+                stereo_z = _weighted_mean_ufunc(
+                    mono_z, weights, n_array_events, indices[valid]
+                )
+
+                mean_cartesian = CartesianRepresentation(
+                    x=stereo_x, y=stereo_y, z=stereo_z
+                )
+                mean_spherical = mean_cartesian.represent_as(SphericalRepresentation)
+                mean_altaz = SkyCoord(mean_spherical, frame=AltAz())
```
StFroese, line 387: `AltAz`

ctapipe/reco/stereo_combination.py
```diff
@@ -294,6 +359,57 @@ def predict_table(self, mono_predictions: Table) -> Table:
...
+
+                # https://en.wikipedia.org/wiki/Directional_statistics#Measures_of_location_and_spread
+                r = mean_spherical.distance.to_value()
+                std = np.sqrt(-2 * np.log(r))
+            else:
+                mean_altaz = SkyCoord(
```
StFroese, line 393: `AltAz`

ctapipe/reco/sklearn.py
```diff
@@ -507,6 +512,353 @@ def predict_table(self, key, table: Table) -> Table:
...
+                    "Must provide `norm_cls` and `sign_cls` if not loading form file"
+                )
+
+            if subarray is None:
+                raise TypeError(
+                    "__init__() missing 1 required positional argument: 'subarray'"
+                )
+
+            super().__init__(subarray, **kwargs)
+            self.subarray = subarray
```
StFroese, line 581: This should already be the case after `super().__init__`

ctapipe/reco/sklearn.py
```diff
@@ -507,6 +512,353 @@ def predict_table(self, key, table: Table) -> Table:
...
+            loaded = self.read(self.load_path)
+            if (
+                subarray is not None
+                and loaded.subarray.telescope_types != subarray.telescope_types
+            ):
+                self.log.warning(
+                    "Supplied subarray has different telescopes than subarray loaded form file"
+                )
+            self.__dict__.update(loaded.__dict__)
+            self.subarray = subarray
```
StFroese, line 607: No `super().__init__(subarray, **kwargs)` in this case, is this correct?

ctapipe/coordinates/disp.py
```diff
@@ -0,0 +1,44 @@
...
+
+def horizontal_to_telescope(
+    alt: u.Quantity, az: u.Quantity, pointing_alt: u.Quantity, pointing_az: u.Quantity
+):
+    """Transform coordinates form horizontal coordinates into TelescopeFrame"""
+    with warnings.catch_warnings():
+        warnings.simplefilter("ignore", MissingFrameAttributeWarning)
+
+        horizontal_coord = SkyCoord(alt=alt, az=az, frame=AltAz())
+        pointing = SkyCoord(alt=pointing_alt, az=pointing_az, frame=AltAz())
```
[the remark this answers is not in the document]
maxnoe: Good question. Astropy recommended using skycoords over frames with data for most if not all application, but most of the time they can be used interchangebly. The frame version cannot hold attributes unknown to that frame, skycoord can.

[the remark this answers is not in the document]
maxnoe: So we should check what is faster maybe, but the default should probably be a skycoord with frame, not frame with data

ctapipe/tools/train_disp_reconstructor.py
```diff
@@ -0,0 +1,199 @@
...
+    def _get_true_disp(self, table):
+        fov_lon, fov_lat = horizontal_to_telescope(
+            alt=table["true_alt"],
+            az=table["true_az"],
+            pointing_alt=table["subarray_pointing_lat"],
+            pointing_az=table["subarray_pointing_lon"],
+        )
+
+        # numpy's trigonometric functions need radians
+        psi = table["hillas_psi"].quantity.to_value(u.rad)
```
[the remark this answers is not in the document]
maxnoe: But it's much faster to do the conversion once upfront. The comment is misleading though

ctapipe/tools/apply_models.py
```diff
@@ -156,22 +180,41 @@ def start(self):
             self.loader.h5file = self.h5file
 
     def _apply(self, reconstructor):
-        prefix = reconstructor.model_cls
         property = reconstructor.property
+        if isinstance(reconstructor, DispReconstructor):
+            prefix = reconstructor.prefix
+        else:
+            prefix = reconstructor.model_cls
```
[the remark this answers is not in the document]
maxnoe: Good point, will open another pr

ctapipe/reco/stereo_combination.py
maxnoe: See comment above

ctapipe/reco/sklearn.py
maxnoe: Yes, the class is loaded from pickle fully initialised.

ctapipe/coordinates/disp.py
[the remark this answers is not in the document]
StFroese:
```python
from astropy.coordinates import AltAz, SkyCoord
from astropy.coordinates import AltAz, SkyCoord
import astropy.units as u
from ctapipe.coordinates import TelescopeFrame

def skycoord():
    horizontal = SkyCoord(alt=20*u.deg, az=12*u.deg, frame=AltAz())
    pointing = SkyCoord(alt=68*u.deg, az=34*u.deg, frame=AltAz())
    tel_frame = TelescopeFrame(telescope_pointing=pointing)
    horizontal.transform_to(tel_frame)

def altaz():
    horizontal = AltAz(alt=20*u.deg, az=12*u.deg)
    pointing = AltAz(alt=68*u.deg, az=34*u.deg)
    tel_frame = TelescopeFrame(telescope_pointing=pointing)
    horizontal.transform_to(tel_frame)
```
```python
In [79]: %timeit skycoord()
1.42 ms ± 41.2 µs per loop (mean ± std. dev. of 7 runs, 1,000 loops each)

In [80]: %timeit altaz()
469 µs ± 7.23 µs per loop (mean ± std. dev. of 7 runs, 1,000 loops each)

```

ctapipe/tools/train_disp_reconstructor.py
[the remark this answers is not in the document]
StFroese: yes, I was confused by the comment

ctapipe/reco/sklearn.py
StFroese: nice

ctapipe/coordinates/disp.py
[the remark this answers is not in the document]
maxnoe: Could you please test the 1st option with using the same `AltAz()` instance? I.e. doing `frame = AltAz()` and then giving `frame=frame` to both `SkyCoord`?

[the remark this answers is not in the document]
StFroese:
```python
def skycoord():
    frame = AltAz()
    horizontal = SkyCoord(alt=20*u.deg, az=12*u.deg, frame=frame)
    pointing = SkyCoord(alt=68*u.deg, az=34*u.deg, frame=frame)
    tel_frame = TelescopeFrame(telescope_pointing=pointing)
    horizontal.transform_to(tel_frame)
```
```python
In [82]: %timeit test1()
1.37 ms ± 20.2 µs per loop (mean ± std. dev. of 7 runs, 1,000 loops each)
```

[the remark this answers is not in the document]
maxnoe: mmphf, okay, then I'll go scan the codebase...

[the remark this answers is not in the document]
StFroese: Could be because SkyCoord.transform_to will build a new SkyCoord after transformation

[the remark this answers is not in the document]
StFroese: https://github.com/astropy/astropy/blob/7539d76ceae146f930d4473107d9940d2fc0b74f/astropy/coordinates/sky_coordinate.py#L718

```python
In [101]: %timeit SkyCoord(alt=60*u.deg, az=30*u.deg, frame=AltAz())
183 µs ± 3.67 µs per loop (mean ± std. dev. of 7 runs, 10,000 loops each)

In [102]: %timeit AltAz(alt=60*u.deg, az=30*u.deg)
83.5 µs ± 309 ns per loop (mean ± std. dev. of 7 runs, 10,000 loops each)
```

ctapipe/containers.py
```diff
@@ -843,6 +849,19 @@ class ParticleClassificationContainer(Container):
...
+    sign = Field(nan, "reconstructed sign for disp")
+    # "parameters_is_valid" instead of "is_valid" to avoid duplication of "disp_is_valid" column
+    parameters_is_valid = Field(False, "true if the predictions are valid")
```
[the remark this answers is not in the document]
kosack: Can you explain what the problem is?  You should have both a reconstructor prefix and the container prefix, so I don't see how you would get a name collision.  I don't see any need to change the reconstruction container names.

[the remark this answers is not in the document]
maxnoe: At the moment we fill the reconstructor name / model name as the prefix.

I agree that this issue should be solved with a different prefix, not with a custom suffix on the field name

ctapipe/reco/sklearn.py
```diff
@@ -507,6 +512,353 @@ def predict_table(self, key, table: Table) -> Table:
...
+                    parameters_is_valid=False,
+                )
+                altaz_container = ReconstructedGeometryContainer(
+                    alt=u.Quantity(np.nan, u.deg, copy=False),
+                    az=u.Quantity(np.nan, u.deg, copy=False),
+                    is_valid=False,
+                )
+
+            disp_container.prefix = f"{self.prefix}_tel"
+            altaz_container.prefix = f"{self.prefix}_tel"
```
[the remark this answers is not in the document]
maxnoe: Just solved by using a `{self.prefix}_parameters` prefix for the disp container and the pure `{self.prefix}` for the `ReconstructedGeometryContainer`.

ctapipe/containers.py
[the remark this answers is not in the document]
kosack: Also, this should be `parameters_are_valid` to be grammatically correct (or `disp_is_valid`) 

is the issue only when reading from a Table I guess? Is it a problem with mixed telescope vs subarray columns? There I can see you might have name collisions if he prefixes are not set right.   We had this same issue with per-tel columns mixed with subarray columns before, which is why we added `_tel_`.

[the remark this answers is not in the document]
maxnoe: The problem here is that `DispParameters` and `ReconstructedGeometry` get assigned the same prefix, which makes them collide when loaded by `TableLoader` into a single table.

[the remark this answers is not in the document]
maxnoe: This could actually be solved by having `TableLoader` include the full path, e.g. `dl2.event.tel.geometry.DispReconstructor.is_valid` but that will make for some pretty long column names...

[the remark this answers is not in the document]
StFroese: Maybe just include the container name in the prefix?

ctapipe/reco/sklearn.py
LukasBeiske: You mean your comment regarding changing `sign_invalid_class` to `nan`, correct? So setting it to `0` permanently should be fine, right?

maxnoe: Yes, but I'd even prefer the solution to just store a single disp value with sign included (and setting that to nan for `not is_valid)`

ctapipe/coordinates/disp.py
LukasBeiske: Done

ctapipe/coordinates/disp.py
LukasBeiske: Done

ctapipe/reco/sklearn.py
LukasBeiske: Oh I missed that. In 'SKLearnReconstrctor´ it is 
```python
def __init__(self, subarray=None, models=None, **kwargs):
            
            ...
    if subarray is None:
        raise TypeError(
            "__init__() missing 1 required positional argument: 'subarray'"
        )
```
So the error message is a bit misleading since `subarray` isn't a posarg there. Would it be better to always make it a posarg and remove this error message since it is unnecessary?

maxnoe: I think the `None` case is needed for `.read()` ing the pickled class on its own.

LukasBeiske: Ok. Then I will do it the same way in `DispReconstructor`

ctapipe/reco/sklearn.py
LukasBeiske: True. I will change it in `SKLearnReconstructor`, too, to be consistent.

ctapipe/reco/sklearn.py
LukasBeiske: fixed

ctapipe/reco/sklearn.py
LukasBeiske: fixed

ctapipe/reco/sklearn.py
LukasBeiske: Done

ctapipe/reco/sklearn.py
LukasBeiske: I will convert it to a plain number once, to be consistent with `TrainDispReconstructor._get_true_disp`

ctapipe/reco/sklearn.py
LukasBeiske: Done, see above

ctapipe/reco/sklearn.py
LukasBeiske: Done

ctapipe/containers.py
LukasBeiske: What should this be called? Since `disp_parameters_disp` doesn't make much sense imo

maxnoe: It's still the `norm`, it just includes the sign (as opposed to predicting dx / dy as lstchain used to do).

LukasBeiske: Done

ctapipe/reco/stereo_combination.py
LukasBeiske: Done

LukasBeiske:
I seem to have missed that commit [71672d8](https://github.com/cta-observatory/ctapipe/pull/2138/commits/71672d849274133f0de27471da145e2f49e66815) breaks `apply-models` for the energy regressor (haven't tested the particle classifier yet) which is also why some tests are failing atm. 
However, I don't understand which change causes this error, especially since I did not actually change anything at the point in the code where the error occurs.
```
  File "/scratch/lbeiske/envs/cta/lib/python3.8/site-packages/tables/table.py", line 2192, in append
    wbufRA = np.rec.array(rows, dtype=self._v_dtype)
  File "/scratch/lbeiske/envs/cta/lib/python3.8/site-packages/numpy/core/records.py", line 1078, in array
    new = obj.view(dtype)
ValueError: When changing to a larger dtype, its size must be a divisor of the total size in bytes of the last axis of the array.

During handling of the above exception, another exception occurred:

  File "/net/nfshome/home/lbeiske/ctapipe/ctapipe/core/tool.py", line 358, in run
    self.start()
  File "/net/nfshome/home/lbeiske/ctapipe/ctapipe/tools/apply_models.py", line 178, in start
    self._apply(reconstructor)
  File "/net/nfshome/home/lbeiske/ctapipe/ctapipe/tools/apply_models.py", line 272, in _apply
    write_table(
  File "/net/nfshome/home/lbeiske/ctapipe/ctapipe/io/astropy_helpers.py", line 211, in write_table
    h5_table.append(table.as_array())
  File "/scratch/lbeiske/envs/cta/lib/python3.8/site-packages/tables/table.py", line 2194, in append
    raise ValueError("rows parameter cannot be converted into a "
ValueError: rows parameter cannot be converted into a recarray object compliant with table '/dl2/event/telescope/energy/RandomForestRegressor/tel_001 (Table(18158,)fletcher32, shuffle, blosc:zstd(5)) '''. The error was: <When changing to a larger dtype, its size must be a divisor of the total size in bytes of the last axis of the array.>
```

nbiederbeck:
We found the issue in https://github.com/cta-observatory/ctapipe/blob/f789773f94bc5f33b33258dab13b5d4bd6c467dd/ctapipe/io/tableloader.py#L326-L329

```python
(Pdb) table['obs_id'].dtype
dtype('int32')
(Pdb) observation_table['obs_id'].dtype
dtype('uint64')
```

The join makes a float64 out of this.

ctapipe/io/tableloader.py
```diff
@@ -101,7 +101,7 @@ def _empty_telescope_events_table():
     Create a new astropy table with correct column names and dtypes
     for telescope event based data.
     """
-    return Table(names=TELESCOPE_EVENT_KEYS, dtype=[np.int32, np.int64, np.int16])
```
nbiederbeck, line 104: Is this change compatible with the data model?

maxnoe: Yes, that's the point, right?

https://github.com/cta-observatory/ctapipe/blob/ed5fdcc75744c8eb8e5a516b25b540ae1a367abf/ctapipe/containers.py#L76-L78

nbiederbeck: Indeed.  I had in mind some smaller dtypes.

maxnoe: Yes, that's the point, right?
dtypes here:

https://github.com/cta-observatory/ctapipe/blob/ed5fdcc75744c8eb8e5a516b25b540ae1a367abf/ctapipe/containers.py#L76-L78

ctapipe/reco/preprocessing.py
```diff
@@ -48,6 +56,16 @@ def check_valid_rows(table: Table, warn=True, log=LOG) -> np.ndarray:
     return valid
 
 
+def table_to_X(table: Table, features: list, log):
+    """
+    Extract features as numpy ndarray to be given to sklearn from input table
```
nbiederbeck, line 61: docstring needs to be more precise: this function selects all valid features, which means only those where there is no nan ?

ctapipe/reco/stereo_combination.py
```diff
@@ -205,6 +208,55 @@ def _combine_classification(self, event):
...
+                alt=u.Quantity(np.nan, u.deg, copy=False),
+                az=u.Quantity(np.nan, u.deg, copy=False),
+            )
+            std = np.nan
+            valid = False
+
+        event.dl2.stereo.geometry[self.prefix] = ReconstructedGeometryContainer(
+            alt=mean_altaz.alt.to(u.deg),
+            az=mean_altaz.az.to(u.deg),
+            ang_distance_uncert=u.Quantity(np.rad2deg(std), u.deg, copy=False),
```
nbiederbeck, line 254: do we need the `.to(u.deg)` or does the container handle `rad`? same for predict_table below

ctapipe/tools/tests/test_train.py
```diff
@@ -16,6 +16,12 @@ def test_train_particle_classifier(particle_classifier_path):
     ParticleClassifier.read(particle_classifier_path)
 
 
+def test_train_disp_reconstructor(disp_reconstructor_path):
```
nbiederbeck, line 19: Maybe for another PR: this would better be called `test_read_trained_{}` ? The training happens in the fixture `{}_path`.

ctapipe/tools/train_disp_reconstructor.py
```diff
@@ -0,0 +1,196 @@
...
+
+    overwrite = Bool(help="overwrite existing output files").tag(config=True)
+
+    random_seed = Int(
+        default_value=0, help="Random seed for sampling and cross validation"
+    ).tag(config=True)
+
+    project_disp = Bool(
+        default_value=False,
+        help="Project true source position on main shower axis for true |disp| calculation",
```
nbiederbeck, line 57: Please clarify what the differences are between True and False here

ctapipe/reco/sklearn.py
```diff
@@ -507,6 +501,324 @@ def predict_table(self, key, table: Table) -> Table:
...
+            DispContainer,
+            prefix=f"{self.prefix}_parameter",
+            stereo=False,
+        )
+
+        psi = table["hillas_psi"].quantity.to_value(u.rad)
+        fov_lon = table["hillas_fov_lon"] + disp * np.cos(psi)
+        fov_lat = table["hillas_fov_lat"] + disp * np.sin(psi)
+
+        # For now: Assume parallel pointing for each run
```
nbiederbeck, line 791: Why? can't we transform elementwise?

Might throw a warning here until changed

ctapipe/reco/stereo_combination.py
maxnoe: THe container handles rad, it will be converted to the unit specified in the container when writing

ctapipe/reco/sklearn.py
maxnoe: > Why? can't we transform elementwise?

Pointing information for simulations is not really well-defined in the dl1 output. Until we resolve this general issue, I think this is ok.

ctapipe/reco/preprocessing.py
LukasBeiske: I think only rows containing nans will be dropped by `check_valid_rows`, so `X` only contains events for which all features are valid. I will update the docstring accordingly.

ctapipe/reco/stereo_combination.py
LukasBeiske: Does this also work in `predict_table`? Since it is never saved in a container in that case.

ctapipe/reco/sklearn.py
LukasBeiske: Added a warning.

ctapipe/tools/train_disp_reconstructor.py
LukasBeiske: Done.

nbiederbeck (dismissed):

ctapipe/reco/sklearn.py
```diff
@@ -788,7 +788,8 @@ def predict_table(self, key, table: Table) -> Tuple[Table, Table]:
         fov_lon = table["hillas_fov_lon"] + disp * np.cos(psi)
         fov_lat = table["hillas_fov_lat"] + disp * np.sin(psi)
 
-        # For now: Assume parallel pointing for each run
+        # FIXME: Assume parallel pointing for each run
+        self.log.warning("Assuming parallel pointing for each run")
```
maxnoe, line 792: Assuming "constant and parallel" pointing, right?

LukasBeiske: The pointing is loaded from the observation block. How would non-constant pointing be handled there?

maxnoe:
@LukasBeiske Could you show some performance plots if you think this is ready?

LukasBeiske:
Sure. [cv_disp.pdf](https://github.com/cta-observatory/ctapipe/files/10186763/cv_disp.pdf) these are the performance plots for the model I am currently using.

maxnoe:
Why is core-y so important? That's confusing me a bit.

Could you show the *single* LST performance? I.e. no stereo features for LST?

LukasBeiske:
For mono I used this config:
```yaml
TrainDispReconstructor:
  CrossValidator:
    n_cross_validations: 5

  random_seed: 42

  DispReconstructor:
    norm_cls: RandomForestRegressor
    norm_config:
      n_estimators: 100
      max_features: "sqrt"
      min_samples_leaf: 0.00001
      n_jobs: 40

    log_target: True

    sign_cls: RandomForestClassifier
    sign_config:
      n_estimators: 100
      max_features: "sqrt"
      min_samples_split: 0.00001
      n_jobs: 40

    features:
      - hillas_intensity
      - hillas_length
      - hillas_width
      - hillas_skewness
      - hillas_kurtosis
      - timing_slope
      - timing_deviation
      - peak_time_std
      - concentration_cog
      - leakage_intensity_width_2
      - morphology_n_pixels
      - scaled_length
      - scaled_width
      - RandomForestRegressor_energy
      - log_abs_timing_slope

    QualityQuery:
      quality_criteria:
        - ["enough intensity", "hillas_intensity > 50"]
        - ["Positive width", "hillas_width > 0"]
        - ["enough pixels", "morphology_n_pixels > 3"]
        - ["not clipped", "leakage_intensity_width_2 < 0.5"]

    FeatureGenerator:
      features:
        - ["log_abs_timing_slope", "log(abs(timing_slope))"]
```
And the performance plots still look pretty good imo (sign performs slightly worse): [cv_disp_mono.pdf](https://github.com/cta-observatory/ctapipe/files/10187475/cv_disp_mono.pdf)

ctapipe/reco/preprocessing.py
```diff
@@ -48,6 +56,17 @@ def check_valid_rows(table: Table, warn=True, log=LOG) -> np.ndarray:
     return valid
 
 
+def table_to_X(table: Table, features: list, log):
```
maxnoe, line 59: type hint could be `List[str]`

maxnoe: and could also be given for `log`

ctapipe/reco/sklearn.py
```diff
@@ -507,6 +501,325 @@ def predict_table(self, key, table: Table) -> Table:
...
+    def _new_models(self):
+        norm_regressor = SUPPORTED_REGRESSORS[self.norm_cls](**self.norm_config)
+        sign_classifier = SUPPORTED_CLASSIFIERS[self.sign_cls](**self.sign_config)
+        return norm_regressor, sign_classifier
+
+    def _table_to_y(self, table, mask=None):
+        """
+        Extract target values as numpy array from input table
+        """
+        # make sure we use the unit that was used during training
```
maxnoe, line 599: this could be simplified to:

```
if self.unit is not None:
    norm = table[mask][self.target].quantity.to_value(self.unit)
else:
    norm = np.array(table[self.target][mask])

abs_norm = np.abs(norm)
sign_norm = np.sign(norm)

if self.log_target:
    # check < 0
    abs_norm = np.log(abs_norm)

return abs_norm, sign_norm
```

ctapipe/reco/sklearn.py
```diff
@@ -507,6 +501,325 @@ def predict_table(self, key, table: Table) -> Table:
...
+
+    def fit(self, key, table):
+        """
+        Create and fit new models for ``key`` using the data in ``table``.
+        """
+        self._norm_models[key], self._sign_models[key] = self._new_models()
+
+        X, valid = table_to_X(table, self.features, self.log)
+        self.unit = table[self.target].unit
+        norm_y, sign_y = self._table_to_y(table, mask=valid)
```
maxnoe, line 621: I would just call it `norm` and `sign`. `norm_y` is confusing...

ctapipe/reco/sklearn.py
```diff
@@ -507,6 +501,325 @@ def predict_table(self, key, table: Table) -> Table:
...
+                f"No regressor available for key {key},"
+                f" available regressors: {self._norm_models.keys()}"
+            )
+        if key not in self._sign_models:
+            raise KeyError(
+                f"No classifier available for key {key},"
+                f" available classifiers: {self._sign_models.keys()}"
+            )
+        X, valid = table_to_X(table, self.features, self.log)
+        norm_prediction = np.full(len(table), np.nan)
```
maxnoe, line 667: you can use a single predictions array now and do `prediction[valid] *= self._sign_models[key].predict(X)` which will be faster and save memory.

ctapipe/reco/sklearn.py
```diff
@@ -565,7 +881,11 @@ def __call__(self, telescope_type, table):
             random_state=self.rng.integers(0, 2**31 - 1),
         )
 
-        cv_it = kfold.split(table, table[self.model_component.target])
+        if isinstance(self.model_component, DispReconstructor):
+            cv_it = kfold.split(table, np.sign(table[self.model_component.target]))
```
maxnoe, line 885: Not really sure if stratified kfold on the sign is really required, it should be 50/50 really already by randomly splitting.

ctapipe/tools/apply_models.py
```diff
@@ -176,14 +202,30 @@ def _apply(self, reconstructor):
...
-                    self.log.warning(
-                        "No model in %s for telescope type %s, skipping tel %d",
-                        reconstructor,
-                        tel,
-                        tel_id,
-                    )
-                    continue
+                if isinstance(reconstructor, DispReconstructor):
+                    if tel not in reconstructor._norm_models:
+                        self.log.warning(
```
maxnoe, line 207: It cannot really be the case that only one is missing, right?

maxnoe (changes requested):
Thanks, this looks already really good. A couple of last comments.

ctapipe/reco/preprocessing.py
LukasBeiske: Do you mean `log=LOG` (and `LOG=logging.getLogger(__name__)`) as for `check_valid_rows`, or do you mean `log: List[str]`?

maxnoe: The type of log is `logging.Logger`, so it should be `log: logging.Logger`.

LukasBeiske: I think passing a default would be better, to keep it consistent with the other functions.

ctapipe/reco/sklearn.py
LukasBeiske: Done

ctapipe/reco/sklearn.py
LukasBeiske: Done

ctapipe/reco/sklearn.py
maxnoe: Good point, either we move it above so we get the nice error or we remove it and just assume this cannot happen

ctapipe/reco/sklearn.py
LukasBeiske: Can we be sure of that even if a small sample is used for training?

ctapipe/reco/sklearn.py
LukasBeiske: I think removing it here is fine, since this should never fail if a correct `mask` is passed.

ctapipe/resources/train_disp_reconstructor.yaml
```diff
@@ -0,0 +1,71 @@
...
+      - leakage_pixels_width_1
+      - leakage_pixels_width_2
+      - concentration_cog
+      - concentration_core
+      - morphology_n_islands
+      - morphology_n_large_islands
+      - morphology_n_medium_islands
+      - morphology_n_pixels
+      - morphology_n_small_islands
+      - HillasReconstructor_tel_impact_distance
```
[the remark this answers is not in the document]
LukasBeiske: Should I remove the predicted energy as well, since the stereo parameters are used for that meaning mono events would not be predicable if the energy prediction is used here?

ctapipe/tools/apply_models.py
LukasBeiske: Yes. I changed it to only check for a regressor.

ctapipe/tools/apply_models.py
```diff
@@ -201,31 +201,18 @@ def _apply(self, reconstructor):
...
-                    if tel not in reconstructor._models:
-                        self.log.warning(
-                            "No model in %s for telescope type %s, skipping tel %d",
-                            reconstructor,
-                            tel,
-                            tel_id,
-                        )
-                        continue
+                if (
+                    tel not in reconstructor._norm_models
```
maxnoe, line 205: If you'd use a tuple of (disp_model, sign_model) in a self._models dict, the API of the different reconstructors would basically be the same here

LukasBeiske: Good idea, I changed it.

ctapipe/reco/preprocessing.py
```diff
@@ -102,3 +122,33 @@ def collect_features(
         features.update(subarray_table.loc[tel_id])
 
     return Table({k: [v] for k, v in features.items()})
+
+
+@u.quantity_input(alt=u.deg, az=u.deg, pointing_alt=u.deg, pointing_az=u.deg)
+def horizontal_to_telescope(alt, az, pointing_alt, pointing_az):
+    """Transform coordinates form horizontal coordinates into TelescopeFrame"""
```
maxnoe, line 129: Typo `form`

ctapipe/reco/sklearn.py
```diff
@@ -507,6 +501,318 @@ def predict_table(self, key, table: Table) -> Table:
...
+                if valid:
+                    hillas = event.dl1.tel[tel_id].parameters.hillas
+                    psi = hillas.psi.to_value(u.rad)
+
+                    fov_lon = hillas.fov_lon + disp * np.cos(psi)
+                    fov_lat = hillas.fov_lat + disp * np.sin(psi)
+                    alt, az = telescope_to_horizontal(
+                        lon=fov_lon,
+                        lat=fov_lat,
+                        pointing_alt=event.pointing.tel[tel_id].altitude.to(u.deg),
```
maxnoe, line 710: Unnecessary conversion. Either convert `to_value` if something unit-less is required or drop the `.to(u.deg)`.

maxnoe: Also, using the table method here instead of the normal transformation seems strange. I'd rather see the `TelescopeFrame(...).transform_to(AltAz())`

ctapipe/reco/sklearn.py
```diff
@@ -507,6 +501,318 @@ def predict_table(self, key, table: Table) -> Table:
...
+        )
+        add_defaults_and_meta(
+            disp_result,
+            DispContainer,
+            prefix=f"{self.prefix}_parameter",
+            stereo=False,
+        )
+
+        psi = table["hillas_psi"].quantity.to_value(u.rad)
+        fov_lon = table["hillas_fov_lon"] + disp * np.cos(psi)
```
maxnoe, line 781: This should be `table["hillas_fov_lon"].quantity`, right?

ctapipe/reco/sklearn.py
LukasBeiske: Done

ctapipe/reco/sklearn.py
LukasBeiske: Yes, fixed.

maxnoe (approved):

nbiederbeck (approved):

maxnoe:
@LukasBeiske Please rebase to fix CI build

nbiederbeck (approved):

maxnoe (approved):

ctapipe/reco/stereo_combination.py
```diff
@@ -102,10 +105,10 @@ class StereoMeanCombiner(StereoCombiner):
     def __init__(self, *args, **kwargs):
         super().__init__(*args, **kwargs)
 
-        if self.property not in {"energy", "classification"}:
+        if self.property not in {"energy", "classification", "geometry"}:
```
kosack, line 108: It would be better to use an Enum here rather than an ad-hoc string.  Here all possibly values are known in advance, so it's not arbitrary.  Enums are better for typos and speed (later you do a lot of compares per event to  e.g.  `if self.property == 'geometry':`

Same for the Reconstructor itself, where there is `property="geometry"`

kosack (changes requested):

ctapipe/reco/stereo_combination.py
kosack: THough this is more general than this PR, and  seems that this will be addressed  in #2116 , so I think we can ignore it here.

kosack (approved):

ctapipe/reco/stereo_combination.py
maxnoe: Yes, let's merge this in and implement #2116 in one go for all the reconstructors

PR #2141, opened 2022-11-28
Drop support for CameraFrame in ImageProcessor

### LINKED ISSUE #2061: remove support for calculating Hillas parameters in CameraFrame
see older issue: #1983 

Now that everything works consistently in the TelescopeFrame for reconstruction, we should remove the redundant option to do it in the CameraFrame.   In general doing parameters in the TelescopeFrame is better as they can be directly compared between telescope optics types.

StFroese:
Initially I wanted to remove the option to use the CameraFrame for calculating the Hillas parameters, but I had to change a lot of stuff and ended up removing the CameraFrame completely from the ImageProcessor.

fixes #2061


kosack:
Thanks this will clean up a lot of redundant code! It's clear a lot of pieces need to be touched, so that is ok.  The only consequence is that it means pointing corrections and the focal length are "baked in" to the DL1 parameters (assuming they are applied in the CameraFrame → TelescopeFrame transition), but I think that is a small price to pay for simplicity, and if the need arises we can add a way to correct them.

kosack:
You'll need to update some of the tutorials as well, as some were written before there was a TelescopeFrame I'm afraid!

kosack (commented):
Since so much had to change, it may be a good idea for somebody to run an analysis (both regular reconstruction as well as muon analysis) on a large number of events and compare the results to the old version before we merge this.  Just in case it introduces some unexpected typos.

maxnoe:
There is a lot of code in `ctapipe/reco/hillas_reconstructor` and `ctapipe/reco/hillas_intersection` dealing with parameters in camera frame, these should also be removed.

StFroese:
@maxnoe should this be in this PR since this one is just ImageProcessor? I would do a separate PR for the ShowerProcessor once this is merged

ctapipe/reco/hillas_intersection.py
```diff
@@ -191,24 +186,15 @@ def _predict(self, hillas_dict, array_pointing, telescopes_pointings=None):
...
-                cog_coords = SkyCoord(
-                    fov_lon=hillas.fov_lon,
-                    fov_lat=hillas.fov_lat,
-                    frame=telescope_frame,
-                )
-                cog_coords_nom = cog_coords.transform_to(nom_frame)
+            telescope_frame = TelescopeFrame(
+                telescope_pointing=telescopes_pointings[tel_id]
+            )
+            cog_coords = SkyCoord(
```
maxnoe, line 192: Re our discussion about SkyCoord vs. Frames: could you also check here if doing `TelescopeFrame(fov_lon, ...)` is faster?

StFroese: Yes it's faster. For 200 events:
With SkyCoord: `1.06 ± 0.01`
With Frame: `0.86±0.01`

StFroese:
I checked the dl1 parameters for this branch and master for regular and for the muon analysis. Works fine. 
Also updated the jupyter notebooks in the docs

ctapipe/io/hdf5eventsource.py
```diff
@@ -351,18 +349,6 @@ def _parse_sb_and_ob_configs(self):
 
         return scheduling_blocks, observation_blocks
 
-    def _is_hillas_in_camera_frame(self):
```
LukasNickel, line 354: Just to be sure:
We want to completely drop reading files created in the CameraFrame?

ctapipe/reco/impact.py
```diff
@@ -204,8 +204,8 @@ def get_hillas_mean(self):
         tel_num = 0
 
         for hillas in self.hillas_parameters:
-            peak_x[tel_num] = hillas.x.to(u.rad).value  # Fill up array
-            peak_y[tel_num] = hillas.y.to(u.rad).value
+            peak_x[tel_num] = hillas.fov_lon.to(u.rad).value  # Fill up array
+            peak_y[tel_num] = hillas.fov_lat.to(u.rad).value
```
LukasNickel, line 208: Its not really x and y anymore, but I guess thats fine here

ctapipe/image/extractor.py
```diff
@@ -1051,7 +1052,10 @@ def _apply_second_pass(
         # Apply correction to 1st pass charges
         charge_1stpass = charge_1stpass_uncorrected * correction[selected_gain_channel]
 
-        camera_geometry = self.subarray.tel[tel_id].camera.geometry
+        # do it in TelescopeFrame
```
LukasNickel, line 1055: That comment is not very useful

docs/examples/camera_display.ipynb
```diff
@@ -91,9 +92,20 @@
   },
   {
    "cell_type": "code",
-   "execution_count": null,
+   "execution_count": 5,
    "metadata": {},
-   "outputs": [],
+   "outputs": [
```
LukasNickel, line 97: We wanted to strip all notebooks before committing, right?

docs/changes/2141.api.rst
```diff
@@ -0,0 +1,4 @@
+Drop support for ``CameraFrame`` in ``ImageProcessor``
+
+``ImageProcessor`` now only uses the ``TelescopeFrame`` to perform all image operations.
+This also drops support for calculating the ``HillasParameters`` in ``CameraFrame``.
```
LukasNickel, line 4: And reading such files!

LukasNickel (changes requested):
Looks almost done to me. Some comments:
- If we do not support loading camera frame files, we need to put that in the changelog. I am fine with it, it is bound to happen at some point
- Strip notebooks
- Fix merge conflicts

ctapipe/image/extractor.py
StFroese: I guess that was just a note for myself😅

ctapipe/reco/impact.py
StFroese: It would be a small change to rename it but I'm really hesitant to do changes to the ImPACT code since #2052 is still open.

LukasNickel: Yeah, that is why I meant its fine.
Lets not touch IMPACT code more than we need to

ctapipe/io/hdf5eventsource.py
StFroese: I think so. Like you said, it will happen at some point. In this case it is just the result of getting rid of `CameraHillasParametersContainer`

ctapipe/image/extractor.py
```diff
@@ -1051,7 +1052,9 @@ def _apply_second_pass(
         # Apply correction to 1st pass charges
         charge_1stpass = charge_1stpass_uncorrected * correction[selected_gain_channel]
 
-        camera_geometry = self.subarray.tel[tel_id].camera.geometry
+        camera_geometry = self.subarray.tel[tel_id].camera.geometry.transform_to(
+            TelescopeFrame()
```
kosack, line 1056: this might need some thought - we shouldn't be transforming the geometry for each event *inside* a calibration/image-extraction algorithm like this as that adds quite a bit of overhead.   This algorithm does not care about the frame of the camera geometry or pointing corrections, so it should not require a transform. 

Since we do eventually need to transform for the ImageProcessor, it may be better to do the transform outside of these algorithms, perhaps in the CameraCalibrator (which eventually should also apply pointing correcions), so that a single TelescopeFrame is available to all algorithms that need it.  That would avoid re-doing the transform many times.

kosack (changes requested):

ctapipe/image/extractor.py
maxnoe: Yes, this should be done inside the `ProcessorTool` in `setup` after writing out the subarray description but before setting up telescope components. The telescope components should just get a subarray with cameras in `CameraFrame`.

maxnoe: Maybe add `transform_camera_geometries` to `SubarrayDescription` returning a new Subarray with transformed geometries

kosack: Is creating a whole new SubarrayDescription each time efficient? This was an issue I think we discussed a while back (need to find where): during analysis, the TelescopeFrame needs to be updated once per event for observations, and could be simplified to only once per run for  simulations in fixed Alt/Az mode.

The TelescopeFrame requires: 
* the latest pointing position, which changes in time in ICRS tracking mode (the default for most observations)
* pre-computed pointing corrections which could include not just a shift in RA/Dec, but also a rotation (this is not currently implemented, but will need to be done eventually)

maxnoe: Ah, sorry, forgot about pointing corrections.

But it stands: we could also do the pointing corrections in the transformation from / to TelescopeFrame to higher frames. That would actually make the most sense to me, since pointing direction is not important for the `CameraFrame` <-> `TelescopeFrame` transformation.

To me that means we should transform the whole subarray *once* to `TelescopeFrame` and apply corrections in the transformation `AltAz` <-> `TelescopeFrame`, we don't need to change telescope frame for this, as the attributes needed for the transformation can be specified on the `SkyCoord` as seen here:

https://github.com/cta-observatory/ctapipe/blob/8a7a367dfe4a2f4e9bdf0e9bf774be29bf03b75f/ctapipe/visualization/mpl_camera.py#L434-L439

kosack: Remember there are in principle rotation corrections and maybe others, so you have to apply pointing corrections _before_ Hillas parameterization (doing it while going to Alt/Az wouldn't work), as you need to have correctly shifted/rotated parameters before you do shower geometry reconstruction.  But that can be solved when the ShowerProcessor is moved to TelescopeFrame later (right now e.g. HillasReconstructor supports both frames, but that's ugly).

ctapipe/image/tests/test_hillas.py
```diff
@@ -3,26 +3,23 @@
...
 from ctapipe.instrument import CameraGeometry, SubarrayDescription
 
 
 def create_sample_image(
     psi="-30d",
-    x=0.2 * u.m,
-    y=0.3 * u.m,
-    width=0.05 * u.m,
-    length=0.15 * u.m,
+    x=0.2 * u.deg,
```
maxnoe, line 19: Did you check that these images make sense when just changing meters to degrees?

ctapipe/image/toymodel.py
```diff
@@ -80,8 +87,8 @@ def obtain_time_image(x, y, centroid_x, centroid_y, psi, time_gradient, time_int
 
     """
     longitudinal, _ = camera_to_shower_coordinates(x, y, centroid_x, centroid_y, psi)
-    longitudinal_m = longitudinal.to_value(u.m)
-    time_gradient_ns_m = time_gradient.to_value(u.ns / u.m)
+    longitudinal_m = longitudinal.to_value(u.deg)
```
maxnoe, line 90: variable names still contain the old units

docs/changes/2134.bug.rst
```diff
@@ -0,0 +1 @@
+Fix for Hillas lines in ``ArrayDisplay`` being wrong in the new ``EastingNorthingFrame``.
```
maxnoe, line 1: Is that part of this PR?

docs/changes/2155.bug.rst
```diff
@@ -0,0 +1 @@
+Replace usage of ``$HOME`` with ``Path.home()`` for cross-platform compatibility.
```
maxnoe, line 1: Not part of this PR

docs/changes/2134.bug.rst
StFroese: I think something went wrong while rebasing. It's not part of the PR, I thinks it's an old one

ctapipe/image/tests/test_hillas.py
StFroese: I checked for the notebooks, I can recheck for all other files

maxnoe:
superseded by #2396

PR #2149, opened 2022-12-05
Add debugging chapter to docs

StFroese:


StFroese:
@maxnoe btw docs are also failing here. Could be an issue on master

kosack:
I would guess pycharm ignores this, but I can check.  In general developers can just set a `PYTHONDEBUG` environment variable to set which debugger to use, but I agree having a more sensible default will lead to more devs using a debugger, which is good!

StFroese:
After a small discussion we decided to not force this option to the user but rather tell the devs in the docs about debugging and the option to use this as a default

PR #2151, opened 2022-12-05
fwhm, rise/fall time

clara-escanuela:


ctapipe/image/extractor.py
```diff
@@ -345,6 +345,94 @@ def integration_correction(
...
+    if peak_index == -1:  # take the maximum as a default
+        peak_index = np.argmax(waveforms, axis=-1)
+
+    nsamples = len(waveforms)
+
+    fwhm = []
+    rise_time = []
+    fall_time = []
+
+    for i in prange(0, nsamples):
```
maxnoe, line 388: prange implies `numba`? But the function is not decorated with `njit`

maxnoe: We also found in the past, that doing things in parallel at the waveform level is not actually faster than just using 1 core, probably due to the cache / data moving overhead.

ctapipe/image/extractor.py
```diff
@@ -345,6 +345,94 @@ def integration_correction(
...
+        waveform = waveforms[i]
+        n = waveform.size
+        peak = max(0, min(peak_index[i], n - 1))
```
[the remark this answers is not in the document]
clara-escanuela: If you select the peak index manually, maybe? I added that just to make sure but maybe it is not necessary.

kosack:
a good unit test here might be to cross-check with `scipy.signal.peak_widths` (which I think only works for 1D signals):
``` py3
width, width_height, l_ips, r_ips = peak_widths(
    signal,
    peaks=[peak,],
    rel_height=0.5,
)
```
 Just a quick test on some waveforms shows this function works reasonably well, as long as the peak is not at the edges:
image

ctapipe/image/extractor.py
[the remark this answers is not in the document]
StFroese: Ok, but in that case the user should either take a look at the waveform first or this should raise an exception instead of fixing it for the user.

clara-escanuela:
Alternatively to this implementation for fwhm, one could do the same as for time over threshold with a slightly worst result but quicker and maybe interpolate before.

ctapipe/image/extractor.py
```diff
@@ -350,9 +350,16 @@ def integration_correction(
     return correction
 
 
-def time_parameters(
-    waveforms,
-):
+@guvectorize(
+    [
+        (float64[:], float32[:], float32[:], float32[:]),
```
maxnoe, line 355: In guvectorize, you need to be carefull about the order of the signatures.

The first possible one is taken, where possible also includes "can be cast to".  So you need to have the float32 before float64, otherwise it will happily cast the float32 waveform to float64:

https://numba.readthedocs.io/en/stable/user/vectorize.html#the-guvectorize-decorator

> If you pass several signatures, beware that you have to pass most specific signatures before least specific ones (e.g., single-precision floats before double-precision floats), otherwise type-based dispatching will not work as expected:

clara-escanuela: > Also, you change a waveform into single numbers, so I think the signatures should be:
> 
> ```
> (float32[:], float32, float32, float32)
> ```
> 
> ?

Following the scheme in extract_around_peak, it should be an array. At the end, I write fwhm_arry[0]=...

kosack (approved):
Looks fine.  In a future PR, we can add it perhaps as an optional output of the ImageExtractors, so if enabled, you get images of t_rise, t_fall, and t_over_threshold.

clara-escanuela:
Do you know why some tests, unrelated to the implementation, are failing?

kosack:
> Do you know why some tests, unrelated to the implementation, are failing?

Yes, it's due to  a change in scikit learn that causes a problem with astropy.  @maxnoe opened a PR to fix it, so shortly you should be able to just merge or rebase the main branch and your tests should pass again.

kosack (approved):

ctapipe/image/extractor.py
```diff
@@ -351,6 +351,118 @@ def integration_correction(
...
+@guvectorize(
+    [
+        (float32[:], float32[:], float32[:], float32[:]),
+        (float64[:], float32[:], float32[:], float32[:]),
+    ],
+    "(s)->(),(),()",
+    nopython=True,
+    cache=True,
+)
+def time_parameters(waveform, fwhm_arr, rise_time_arr, fall_time_arr):
```
maxnoe, line 363: I think this new method is missing in `__all__` at the top.

ctapipe/image/extractor.py
```diff
@@ -351,6 +351,118 @@ def integration_correction(
...
+        if (yj - yi) != 0:
+            if yi >= half_amplitude >= yj:
+                fwhm_left = ti - (half_amplitude - yi) / (yj - yi)
+            if yi >= ampl_90percent >= yj:
+                rt_90 = ti - (ampl_90percent - yi) / (yj - yi)
+            if yi >= ampl_10percent >= yj:
+                rt_10 = ti - (ampl_10percent - yi) / (yj - yi)
+
+    fwhm = 0.0
+    if None not in (fwhm_right, fwhm_left):
```
maxnoe, line 428: You initialize these two with `np.nan` above, so `None` will never be in `(fwhm_right, fwhm_left)`.

I think you can just subtract the two anyways, and the result will be nan if either is nan. Check if there is a warning and maybe silence it in case we don't want to have a warning like "invalid valid encountered in subtract".

ctapipe/image/extractor.py
```diff
@@ -351,6 +351,118 @@ def integration_correction(
...
+        fwhm = fwhm_right - fwhm_left
+
+    rise_time = 0.0
+    if None not in (rt_90, rt_10):
```
maxnoe, line 432: Same as above with the other if

ctapipe/image/extractor.py
```diff
@@ -351,6 +351,118 @@ def integration_correction(
...
+        rise_time = rt_90 - rt_10
+
+    fall_time = 0.0
+    if None not in (ft_90, ft_10):
```
maxnoe, line 436: Same as above with the other if

ctapipe/image/extractor.py
```diff
@@ -351,6 +351,118 @@ def integration_correction(
...
+        fall_time = ft_10 - ft_90
+
+    fwhm_arr[0] = fwhm
+    rise_time_arr[0] = rise_time
+    fall_time_arr[0] = fall_time
+
+
+def time_over_threshold(waveforms, thr):
```
maxnoe, line 444: Technically, this is "samples above threshold", for "time over threshold you would have to divide by the sampling frequency.

maxnoe (changes requested):
See inline comments

PR #2152, opened 2022-12-05
fix import error in metadata: fails on systems without `pwd`

kosack:
The standard library module `pwd` (password database access) doesn't exist on all systems (e.g. Windows or WASM), so don't rely on it. This just replaces the call to it with a default value if it is not there.

This feature was only used in `ctapipe.core.metadata.Contact` to get the user's name if they forgot to explicitly include it, so it should not break things if it fails (which it currently does).

### DIFF
--- a/ctapipe/io/metadata.py
+++ b/ctapipe/io/metadata.py
@@ -23,7 +23,6 @@ them (as in `Activity.from_provenance()`)
 
 """
 import os
-import pwd
 import uuid
 import warnings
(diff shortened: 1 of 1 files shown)


StFroese (approved):

maxnoe:
We need to merge #2150 first to fix the docs build

maxnoe (approved):

nbiederbeck (approved):

PR #2146, opened 2022-12-02
Make errors out of DeprecationWarnings to be caught in CI.

nbiederbeck:
This change catches DeprecationWarnings and displays them as errors in pytest.  I hope this way we can catch those more easily.

We can of course add more Warnings here if needed.

PR #2153, opened 2022-12-06
Error deprecationwarnings 2

nbiederbeck:
Rebased commits.

I have no idea why the CI fails in the other PR, so I hope this fixes it maybe.

Supersedes #2146.

### DIFF
--- a/ctapipe/core/tests/test_container.py
+++ b/ctapipe/core/tests/test_container.py
@@ -1,10 +1,8 @@
 # Licensed under a 3-clause BSD style license - see LICENSE.rst
-import warnings
 from functools import partial
 
 import numpy as np
 import pytest
-
 from astropy import units as u
 
 from ctapipe.core import Container, DeprecatedField, Field, FieldValidationError, Map
@@ -212,7 +210,7 @@ def test_container_brackets():
 
 def test_deprecated_field():
 
-    with warnings.catch_warnings(record=True) as warning:
+    with pytest.warns(DeprecationWarning, match="answer to all questions"):
 
         class ExampleContainer(Container):
             answer = DeprecatedField(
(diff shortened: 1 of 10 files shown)


maxnoe:
Is conftest raising a DeprecationError on import and that's somehow converted to this non-descript error?

maxnoe:
I think this is an error related to editable installs. Can you check if `SETUPTOOLS_ENABLE_FEATURES="legacy-editable" pip install -e ../ctapipe_io_lst` fixes it?

nbiederbeck:
why `ctapipe_io_lst`?

I think it has to do something with the documentation tests, when I locally run the exact command from the CI I get the same error:
```
pytest -x -n0 --dist loadscope --cov --cov-report=xml --doctest-modules --doctest-glob='*.rst' --ignore=docs/conf.py ctapipe docs
```

when running without `docs` I don't get the error, and just `docs` (without `ctapipe`) does not raise this error.

maxnoe:
> why ctapipe_io_lst?

Copy pase, of course `.`

maxnoe:
Sorry, the CI does not make an editable install here. However the issue still seems to be that pytest checks that the imported module is at the expected path, which it is not (It checks for the installed path but finds the path in the cwd). So this would also be fixed by switching to the `src/` layout. The question is why this breaks now.

maxnoe:
Is that maybe a `DeprecationWarning`? That different paths in the future will raise an error?

maxnoe:
Astropy runs the tests in a tmpdir, and passes the absolute path to the docs dir, so maybe try:

` pytest --pyargs ctapipe $HOME/ctapipe/docs` in the CI? (Not sure about the path of the repository in the CI

nbiederbeck:
Regarding astropy and testing: should we think about using [`tox`](https://tox.wiki/en/latest/index.html)?

Our environment.yml uses python 3.9, the docs CI uses python 3.8 and the CI tests use 3.8, 3.9 and 3.10. This made debugging this issue in "identical" systems really hard.

maxnoe:
Think yes, however one issue I think is that tox does not integrate well with conda envs, which is currently the recommended way to install ctapipe, so we would anyway need more? Maybe I am wrong and tox now integrates with conda?

nbiederbeck:
https://github.com/tox-dev/tox-conda

maxnoe (approved):

PR #2155, opened 2022-12-07
Don't use $HOME, use Path.home() instead

kosack:
We look for the envvar `HOME` to get the cache directory, but really should just call `Path.home()`


maxnoe (approved):
Nice!

nbiederbeck (approved):
Should even work on windows, as it is basically calling [`os.path.expanduser('~')`](https://docs.python.org/3/library/os.path.html#os.path.expanduser)

PR #2156, opened 2022-12-07
Build on windows

kosack:
Just a quick test to see what breaks if ctapipe is installed in windows...


nbiederbeck:
Current error should be fixed with `numexpr==v2.8.1`

maxnoe:
Other than the strange MemoryErrors, the CI now seems to pass...

PR #2160, opened 2022-12-09
Rename Attributes to Parameters in docstrings

StFroese:


ctapipe/atmosphere.py
```diff
@@ -169,7 +169,7 @@ class ExponentialAtmosphereDensityProfile(AtmosphereDensityProfile):
         density_profile.peek()
 
 
-    Attributes
+    Parameters
```
maxnoe, line 172: But attributes is correct. It's about the class attributes. Parameters are function arguments

StFroese: you're totally right

StFroese: don't know why sphinx didn't want to build it on my machine

StFroese: I'll close this again

PR #2161, opened 2022-12-09
Improve likelihood muon intensity fitter

StFroese:
Use already implemented likelihood for muon intensity fit and catch `np.log(0)` case in likelihood

### DIFF
--- a/ctapipe/image/pixel_likelihood.py
+++ b/ctapipe/image/pixel_likelihood.py
@@ -41,6 +41,8 @@ __all__ = [
     "chi_squared",
 ]
 
+EPSILON = 5.0e-324
+
 
 class PixelLikelihoodError(RuntimeError):
     pass
@@ -91,9 +93,9 @@ def neg_log_likelihood_approx(image, prediction, spe_width, pedestal):
     -------
     float
     """
-    theta = pedestal ** 2 + prediction * (1 + spe_width ** 2)
+    theta = pedestal**2 + prediction * (1 + spe_width**2)
 
-    neg_log_l = np.log(theta) + (image - prediction) ** 2 / theta
+    neg_log_l = np.log(theta + EPSILON) + (image - prediction) ** 2 / theta
 
     return np.sum(neg_log_l)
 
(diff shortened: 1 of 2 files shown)


StFroese:
It's not really intuitive that this can be smaller than zero
https://github.com/cta-observatory/ctapipe/blob/ed5fdcc75744c8eb8e5a516b25b540ae1a367abf/ctapipe/image/pixel_likelihood.py#L94

nbiederbeck:
Is it feasible to not evaluate the `0` case? Width e.g. some `np.where`?

Where does `1e-323` come from? I would suppose you could use some
```python
minimal_positive_float = np.nextafter(np.float64(0), np.float64(0))
```
which is even a little smaller: `5e-324` :wink:

StFroese:
Maybe yes. The idea was to use the same safe log as in iminuit 

https://github.com/scikit-hep/iminuit/blob/e6365013ef1362b06b5acc8cd28106b61cf4bfca/src/iminuit/cost.py#L73-L75

ctapipe/image/pixel_likelihood.py
```diff
@@ -91,9 +91,9 @@ def neg_log_likelihood_approx(image, prediction, spe_width, pedestal):
     -------
     float
     """
-    theta = pedestal ** 2 + prediction * (1 + spe_width ** 2)
+    theta = pedestal**2 + prediction * (1 + spe_width**2)
 
-    neg_log_l = np.log(theta) + (image - prediction) ** 2 / theta
+    neg_log_l = np.log(theta + 1e-323) + (image - prediction) ** 2 / theta
```
kosack, line 96: I would suggest defining `EPSILON = 1.0e-323` at the top of the file and avoid a bare number here (and also below).

ctapipe/image/pixel_likelihood.py
```diff
@@ -134,16 +134,16 @@ def neg_log_likelihood_numeric(
...
-            prediction ** n
+            prediction**n
             * np.exp(-prediction)
             / theta
             * np.exp(-((image - n) ** 2) / (2 * theta))
         )
         likelihood += _l
 
-    return -np.sum(np.log(likelihood))
+    return -np.sum(np.log(likelihood + 1e-323))
```
kosack, line 146: here too

ctapipe/image/pixel_likelihood.py
```diff
@@ -206,8 +206,8 @@ def mean_poisson_likelihood_gaussian(prediction, spe_width, pedestal):
     -------
     float
     """
-    theta = pedestal ** 2 + prediction * (1 + spe_width ** 2)
-    mean_log_likelihood = 1 + np.log(2 * np.pi) + np.log(theta)
+    theta = pedestal**2 + prediction * (1 + spe_width**2)
+    mean_log_likelihood = 1 + np.log(2 * np.pi) + np.log(theta + 1e-323)
```
kosack, line 210: and here

kosack (changes requested):

nbiederbeck (approved):

kosack (approved):

PR #2167, opened 2022-12-14
Implement transformation from GroundFrame to EarthLocation

maxnoe:


ctapipe/coordinates/tests/test_ground_frame.py
```diff
@@ -0,0 +1,54 @@
+import astropy.units as u
+import pytest
+from astropy.coordinates import EarthLocation
+
+# LST-1, MAGIC-1, MAGIC-2
+locations = [
```
nbiederbeck, line 6: If these are somewhat "official", we might include these somewhere else?  E.g. `ctapipe/utils` or whatever.  Might be beneficial for visualizations.

nbiederbeck: These are used for tests, one should rely on the positions in the `subarray`.

nbiederbeck (dismissed):

nbiederbeck (approved):

nbiederbeck (approved):

kosack (approved):

PR #2172, opened 2022-12-19
Add quality query to display dl1 tool

maxnoe:


nbiederbeck (approved):

RuneDominik:
Is there a reason to divert from the usual usage of `QualityQuerys` by not creating a child class here? Even though the default should be to plot all images, which is what is realized here.

maxnoe:
> Is there a reason to divert from the usual usage of QualityQuerys by not creating a child class here? Even though the default should be to plot all images, which is what is realized here.

Subclassing is only really needed to 
a) change the defaults, or
b) add support for configuring two quality queries next to one-another in the config hierarchy

RuneDominik (approved):

RuneDominik (approved):

kosack (approved):

PR #2173, opened 2022-12-20
Fix docs build for traitlets 5.8 and array field defaults

maxnoe:
This fixes an issue we found when building the lstchain docs. Arrays were included with all numbers spanning multiple lines which made the sphinx build fail.


nbiederbeck (approved):

nbiederbeck (approved):
changelog entry is missing

maxnoe:
@nbiederbeck We said that we start with changelogs using towncrier after the next release because we already have merged a lot of PRs without changelog.

PR #2176, opened 2022-12-21
Run changelog check only on PRs, add support for no-changelog-needed …

maxnoe:
…label

### DIFF
--- a/.github/workflows/changelog.yml
+++ b/.github/workflows/changelog.yml
@@ -1,0 +1,24 @@
+name: Changelog
+
+on:
+  pull_request:
+    # should also be re-run when changing labels
+    types: [opened, reopened, labeled, unlabeled, synchronize]
+
+env:
+  FRAGMENT_NAME: "docs/changes/${{ github.event.number }}.*.rst"
+
+jobs:
+  changelog:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v3
+        with:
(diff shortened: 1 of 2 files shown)


LukasNickel (approved):

StFroese (approved):

.github/workflows/changelog.yml
```diff
@@ -0,0 +1,24 @@
...
+jobs:
+  changelog:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v3
+        with:
+          fetch-depth: 0
+
+      - name: Check for news fragment
+        if: ${{ ! contains( github.event.pull_request.labels.*.name, 'no-changelog-needed')}}
```
StFroese, line 20: This seems to not work. Check still fails in this PR

maxnoe: ? All ten checks passed, including their changelog check

StFroese: Ok, I see. That’s weird I re-ran it two times after you added the label and they both failed. Maybe it was just a bug in the GitHub app…

StFroese: image

StFroese: Definitely a bug with the app

maxnoe: 1 failed, 10 passed? But there are only 10 checks...

Issue #2164, opened 2022-12-13
ctapipe-merge does not treat metadata 


maxnoe:
**Describe the bug**
After merging, contact and site info are empty defaults. We should take them from the input files if no other information is provided via the config system.

**To Reproduce**

Merge files that do have contact info / subarray info defined and look at `ctapipe-fileinfo merged_file.h5`

Tobychev:
@maxnoe did #2179 solve this issue?

maxnoe:
Yes, indeed, thanks!

Issue #2130, opened 2022-11-23
Allow dropping dl1 and telescope-wise information in `apply-models` or add tool for that.


maxnoe:
**Please describe the use case that requires this feature.**
Currently, `apply-models` just copies the input file, with everything inside it.

We should add options – or a second tool – to only copy subsets, e.g. the metadata + instrument + event/subarray data

StFroese:
Ah ok. So something like `apply-model --reduced-data`

nbiederbeck:
Maybe `ctapipe-merge` would also benefit from some sort of Copy Component

Issue #2057, opened 2022-08-16
Using `allowed_tels` at different stages of the pipeline results in broken files


nbiederbeck:
*Disclaimer: I will update this description once I have a reproducible example and narrowed down the cause*

**Describe the bug**
We were calculating the angular resolution for some subarrays. 
1) Simtel(allowed_tels) -> DL1 ; DL1(all tels) -> DL2 works (ang-res of 0.5)
2) Simtel(all tels) -> DL1 ; DL1(allowed_tels) -> DL2 fails (ang-res of 5).

My guess is that in the second case the reading *with the HDF5EventSource* is buggy and the true-reconstructed association is wrong.

**To Reproduce**
*Will update*

**Expected behavior**
It should not make a difference where in the pipeline telescopes get selected

nbiederbeck:
I created this test that fails with a given file, but with our current test files I cannot reproduce this:
```python

def test_read_from_simtel_and_dl1(prod5_proton_simtel_path, tmp_path):
    """In #2057 reading a subset of allowed tels from
    simtel yields another result as reading from DL1,
    which was produced with all tels.

    This test has three steps:
    1) Create a DL2 file from simtel.
    2) Create a DL1 file from simtel.
    3) Create from that DL1 file another DL2 file.

    Keep in mind that both DL2 allowed_tels need to be the same,
    but different from the allowed_tels in the simtel->DL1 step!
    """
    from numpy.testing import assert_allclose

    from ctapipe.core import run_tool
    from ctapipe.io import TableLoader
    from ctapipe.tools.process import ProcessorTool

    input_path = "/cephfs/projects/CTA/simulations/simtel/Prod5/LaPalma/NSB1x/gamma-diffuse/00xxx/gamma_20deg_0deg_run954___cta-prod5b-lapalma_desert-2158m-LaPalma-dark_cone10.simtel.zst"
    allowed_tels_lapalma_alpha = [
        f"--EventSource.allowed_tels={i}"
        for i in (1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 19, 35)
    ]
    allowed_msts_lapalma_alpha = [
        f"--EventSource.allowed_tels={i}" for i in (5, 6, 7, 8, 9, 10, 11, 19, 35)
    ]

    # 1) Create DL2 from simtel.

    dl2_from_simtel = tmp_path / "from_simtel.dl2.h5"
    argv = [
        f"--input={input_path}",
[...]

PR #2179, opened 2023-01-04
Add HDF5Merger Component, use in MergeTool

maxnoe:
This is WIP  for introducing the `HDF5Merger` component and using it in the merge tool to address several open issues:

* #2164
* #2130 
* #2057 

A new component `HDF5Merger` is introduced, which allows finegrained control via traitlets of what is included or not included in the merge and has a `__call__` method that adds a single file to the output.

This component is called repeatedly in merge and could be used to copy the input file in the `apply-models` tool so that one can drop dl1 or even all telescope-wise information in that tool. (not yet done)


ctapipe/tools/merge.py
```diff
@@ -116,126 +30,99 @@ class MergeTool(Tool):
     If no pattern is given, all .h5 files of the given directory will be taken as input.
     """
     input_dir = traits.Path(
-        help="Input dl1-directory",
         default_value=None,
+        help="Input dl1-directory",
```
kosack, line 34: change to:

```python
 help=(
    "directory containing ctapipe data files to merge.  "
    "Use `file_pattern` to set which files are considered. "  
)
```

ctapipe/tools/merge.py
```diff
@@ -116,126 +30,99 @@ class MergeTool(Tool):
...
         allow_none=True,
         exists=True,
         directory_ok=True,
         file_ok=False,
     ).tag(config=True)
+
     input_files = List(
         traits.Path(exists=True, directory_ok=False),
         default_value=[],
         help="Input dl1-files",
```
kosack, line 44: same here, change to:

```python
help="Explicit list of ctapipe data files to merge."
```

ctapipe/tools/merge.py
maxnoe: sh / bash have a maximum number of command line arguments, which is not really large, this was the reason for having this.

maxnoe: Seems it's 2 million, at least on my machine, which should be plenty... But I remember having issues and having to write a config file for the merge tool

maxnoe:
This is ready for review now

ctapipe/io/tests/test_merge.py
```diff
@@ -0,0 +1,129 @@
...
+
+def compare_stats_table(in1, in2, merged, table, required=True):
+    t1 = read_table(in1, table) if table in in1 else None
+    t2 = read_table(in2, table) if table in in2 else None
+
+    if required and t1 is None:
+        raise ValueError(f" table {table} not present in {in1.filename}")
+
+    if required and t2 is None:
+        raise ValueError(f" table {table} not present in {in1.filename}")
```
LukasNickel, line 38: ~1~ 2

ctapipe/io/tests/test_merge.py
```diff
@@ -0,0 +1,129 @@
...
+    merged = tables.open_file(output)
+    with in1, in2, merged:
+        for table in tables_to_check:
+            tables_checked += compare_table(in1, in2, merged, table)
+
+        print(f"Checked {tables_checked} tables")
+        assert tables_checked > 0
+
+        for table in statistics_tables:
+            tables_checked += compare_stats_table(
```
LukasNickel, line 108: Here you would also need to check, that the counter incremented. This just makes sure, that the exception is not raised.

ctapipe/io/tests/test_merge.py
```diff
@@ -0,0 +1,129 @@
...
+
+    in1 = tables.open_file(gamma_train_clf)
+    in2 = tables.open_file(proton_train_clf)
+    merged = tables.open_file(output)
+    with in1, in2, merged:
+        for table in tables_to_check:
+            tables_checked += compare_table(in1, in2, merged, table)
+
+        print(f"Checked {tables_checked} tables")
+        assert tables_checked > 0
```
LukasNickel, line 105: I dont think just checking `>0` is sufficient.
Are there missing tables in the input?
If not, it should just be `len(tables_to_check) == tables_check`.

ctapipe/io/hdf5merger.py
```diff
@@ -0,0 +1,467 @@
...
+
+class CannotMerge(IOError):
+    """Raised when trying to merge incompatible files"""
+
+
+def split_h5path(path):
+    """
+    Split a path inside an hdf5 file into parent / child
+    """
+    head, _, tail = path.rstrip("/").rpartition("/")
```
LukasNickel, line 84: Why would there be a trailing `/`?

Optionally: `head, tail = path.rsplit("/", 1)`, but his is fine as well.

ctapipe/io/hdf5merger.py
```diff
@@ -0,0 +1,467 @@
...
+
+
+class HDF5Merger(Component):
+    """
+    Class to copy / append / merge ctapipe hdf5 files
+    """
+
+    output_path = traits.Path(directory_ok=False).tag(config=True)
+
+    overwrite = traits.Bool(
```
LukasNickel, line 95: Do you define it for the reference to `append` in the help string?
Otherwise its not needed anymore.

ctapipe/io/hdf5merger.py
```diff
@@ -0,0 +1,467 @@
...
+        False,
+        help="If true, the ``output_path`` is overwritten in case it exists. See also ``append``",
+    ).tag(config=True)
+
+    append = traits.Bool(
+        False,
+        help="If true, the ``output_path`` is appended to. See also ``overwrite``",
+    ).tag(config=True)
+
+    telescope_events = traits.Bool(
```
LukasNickel, line 105: Not sure if these are very descriptive.
I can live with `telescope_events`, but `simulation` is really vague.
Maybe `keep_telescope_events/...` or `add_...`?

ctapipe/io/hdf5merger.py
```diff
@@ -0,0 +1,467 @@
...
+    ).tag(config=True)
+
+    def __init__(self, output_path=None, **kwargs):
+        # enable using output_path as posarg
+        if output_path not in {None, traits.Undefined}:
+            kwargs["output_path"] = output_path
+
+        super().__init__(**kwargs)
+
+        appending = self.output_path.exists()
```
LukasNickel, line 158: I think it would be more readable, if this was named `output_exists` and
then later 
```python
appending = output_exists and not overwrite
```
or
```python
if overwrite:
    appending = False
else:
    appending = output_exists
```
or just dont use the variable `appending` and do
```python
if output_exists and not overwrite:
# Check files are compatible
```
since its only ever used there.

ctapipe/io/hdf5merger.py
```diff
@@ -0,0 +1,467 @@
...
+            # focal length choice doesn't matter here, set to equivalent so we don't get
+            # an error if only the effective focal length is available in the file
+            self.subarray = SubarrayDescription.from_hdf(
+                self.h5file, focal_length_choice=FocalLengthKind.EQUIVALENT
+            )
+            self.required_nodes = _get_required_nodes(self.h5file)
+
+    def __call__(self, other: Union[str, Path, tables.File]):
+        """
+        Append file ``other`` to the ouput file
```
LukasNickel, line 205: typo: output

ctapipe/io/hdf5merger.py
```diff
@@ -0,0 +1,467 @@
...
+            if node_path not in other.root:
+                raise CannotMerge(
+                    f"Required node {node_path} not found in {other.filename}"
+                )
+
+    def _append(self, other):
+        # Configuration
+        self._append_subarray(other)
+
+        key = "/configuration/observation/scheduling_block"
```
LukasNickel, line 263: This could be a loop over keys.
There are very few special cases here

ctapipe/io/hdf5merger.py
```diff
@@ -0,0 +1,467 @@
...
+            other_meta = metadata.Reference.from_dict(metadata.read_metadata(other))
+        except Exception:
+            raise CannotMerge(
+                f"CTA Rerence meta not found in input file: {other.filename}"
+            )
+
+        if self.meta is None:
+            self.meta = other_meta
+            self.data_model_version = self.meta.product.data_model_version
+            metadata.write_to_hdf5(self.meta.to_dict(), self.h5file)
```
LukasNickel, line 244: It did come as a surprise here, that this function actually writes the metadata and not just checks

ctapipe/io/hdf5merger.py
```diff
@@ -0,0 +1,467 @@
...
+
+            output_table.append(input_table.astype(output_table.dtype))
+
+        else:
+            self._get_or_create_group(group_path)
+
+            if filter_columns is None:
+                self._copy_node(file, table)
+            else:
+                input_table = recarray_drop_columns(table[:], filter_columns)
```
LukasNickel, line 425: Just to make sure: This keeps the metadata and not just the values?

ctapipe/io/hdf5merger.py
```diff
@@ -0,0 +1,467 @@
...
+    def _copy_node(self, file, node):
+        group_path, _ = split_h5path(node._v_pathname)
+        target_group = self._get_or_create_group(group_path)
+        file.copy_node(node, newparent=target_group)
+
+    def _add_statistics_table(self, file: tables.File, input_table: tables.Table):
+        """
+        Creates table for image statistics and adds the entries together.
+
+        This does not append rows to the existing table
```
LukasNickel, line 452: but instead sums the values.

LukasNickel (changes requested):

ctapipe/io/tests/test_merge.py
maxnoe: We check more tbales than tables_to_check because tables_to_check contains groups of tel and algorithm tables.

ctapipe/io/hdf5merger.py
maxnoe: > Why would there be a trailing / ?

Why not? That's a valid path for a group `/foo/mygroup/`

ctapipe/io/hdf5merger.py
maxnoe: It is. You can either overwrite (truncating an existing file) or append (add to an existing file) or fail if the output file exists. So we need both options.

ctapipe/io/hdf5merger.py
maxnoe: > simulation is really vague.

It's not "vague", it's just a lot of stuff. If you say `simulation=False`, nothing in `/simulation` will be copied, that's rather straight forward, or not?

ctapipe/io/tests/test_merge.py
LukasNickel: Are there?
It is constructed via a loop over the telescopes:
https://github.com/cta-observatory/ctapipe/blob/aafe5fe1c880ddcd163eb7b9ed552e5c669cc626/ctapipe/io/tests/test_merge.py#L87-L90

ctapipe/io/hdf5merger.py
LukasNickel: Right, its a component, not a tool...

ctapipe/io/hdf5merger.py
maxnoe: Every case is a special case in which options are looked at to decide what is included or not...

maxnoe: I am open for another solution, but I tried and it resulted in the equivalent number of lines in a dict that had keys, categories, and what is stored (table, algorithm group, tel table group, etc.)

LukasNickel: But is this not 
```python
for key in keys:
    if key in other.root:
        self.required_nodes.add(key)
        self._append_table(other, other.root[key])
```
for most of them?
And then you could still keep the special cases as they are right now

ctapipe/io/hdf5merger.py
maxnoe: yes, true, will change.

ctapipe/io/hdf5merger.py
maxnoe: First line ;)

ctapipe/io/hdf5merger.py
maxnoe: Good catch. It's a copy of before, but I think it's not copied.

ctapipe/io/tests/test_merge.py
maxnoe: ha, right

ctapipe/io/hdf5merger.py
maxnoe: Not, you are missing the traitlets that select if that particular key should be written

maxnoe: and the different kinds of `_append_XXX`

ctapipe/io/tests/test_merge.py
maxnoe: > It is constructed via a loop over the telescopes:

Exactly, and not all telescopes have telescope events in all input files, thus the number of tables_to_check is larger than the actually checked tables...

ctapipe/io/hdf5merger.py
```diff
@@ -0,0 +1,470 @@
...
+        # output file existed, so read subarray and data model version to make sure
+        # any file given matches what we already have
+        if appending:
+            try:
+                self.meta = metadata.Reference.from_dict(
+                    metadata.read_metadata(self.h5file)
+                )
+            except Exception:
+                raise CannotMerge(
+                    f"CTA Rerence meta not found in existing output file: {self.output_path}"
```
kosack, line 191: typo: Rerence → Reference

kosack, line 191: could you use self._read_meta() here?

ctapipe/io/hdf5merger.py
```diff
@@ -0,0 +1,470 @@
...
+            self.h5file.root._v_attrs["CTA PRODUCT CREATION TIME"] = time.iso
+            self.h5file.root._v_attrs["CTA PRODUCT ID"] = id_
+        self.h5file.flush()
+
+    def _read_meta(self, h5file):
+        try:
+            return metadata.Reference.from_dict(metadata.read_metadata(h5file))
+        except Exception:
+            raise CannotMerge(
+                f"CTA Rerence meta not found in input file: {h5file.filename}"
```
kosack, line 244: typo

ctapipe/io/hdf5merger.py
```diff
@@ -0,0 +1,470 @@
...
+        with exit_stack:
+            # first file to be merged
+            if self.meta is None:
+                self.meta = self._read_meta(other)
+                self.data_model_version = self.meta.product.data_model_version
+                metadata.write_to_hdf5(self.meta.to_dict(), self.h5file)
+            else:
+                self._check_can_merge(other)
+
+            Provenance().add_input_file(other.filename)
```
kosack, line 220: add a descriptive role, "data product to merge"

ctapipe/io/hdf5merger.py
kosack: I think at least a few could be collapsed into loop within each option group, but of course that doesn't do so much to simplify the code. You could make a dict of dicts to map the options to the keys to check related to that option, but again, it may be less readable that way, so maybe not worth the effort.

ctapipe/io/hdf5merger.py
maxnoe: Yes, was introduced later and forgot to use everywhere

ctapipe/io/tests/test_merge.py
LukasNickel: Ok fine.
But then you could still check against the actual number of tables since the input file is fixed. It would make for a good regression test I think.

maxnoe: done

ctapipe/io/hdf5merger.py
maxnoe: I actually had that and found it less readable, but I am open to suggestions...

ctapipe/io/tests/test_merge.py
```diff
@@ -32,10 +32,10 @@ def compare_stats_table(in1, in2, merged, table):
     t2 = read_table(in2, table) if table in in2 else None
 
     if t1 is None:
-        raise ValueError(f" table {table} not present in {in1.filename}")
```
LukasNickel: Shouldn't `read_table` throw an error like this already if the table is not in the file? Then this code is not needed and neither is the `if table in file else None` above

maxnoe: From the debugging point of view, it would be better to land in the failing test, right? Not in the `read_table` method?

ctapipe/io/hdf5merger.py
maxnoe: I merged the ones which are exactly the same, didn't reduce it by much though.

ctapipe/io/hdf5merger.py
maxnoe: Fixed in f2b33ec2a7b9c

LukasNickel (approved):

kosack (approved):

Issue #1552, opened 2020-11-24
Remove special treatment of `q.info` in `CameraGeometry.to_table` once astropy 4.3 is released


maxnoe:
The code is not needed anymore since astropy 4.3 will fix the performance bottleneck for indexing

PR #2181, opened 2023-01-04
Remove workaround for astropy bug fixed in astropy 4.3

maxnoe:
Fixes https://github.com/cta-observatory/ctapipe/issues/1552

### DIFF
--- a/ctapipe/instrument/camera/geometry.py
+++ b/ctapipe/instrument/camera/geometry.py
@@ -602,6 +602,14 @@ class CameraGeometry:
                 CAM_ROT=self.cam_rotation.deg,
             ),
         )
+
+        # clear `info` member from quantities set by table creation
+        # which impacts indexing performance because it is deepcopied
+        # in Quantity.__getitem__, see https://github.com/astropy/astropy/issues/11066
+        for q in (self.pix_id, self.pix_x, self.pix_y, self.pix_area):
+            if hasattr(q, "__dict__"):
+                q.__dict__.pop("info", None)
+
         return t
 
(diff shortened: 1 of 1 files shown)


maxnoe:
Need to check if actually fixed.

maxnoe:
So, turns out, with our quickfix, a camera geometry read from file is faster after running to table then before:

```python
import numpy as np

s = SubarrayDescription.from_hdf('../public_dl2/data/gamma-diffuse.dl2.h5')
cam = s.tel[1].camera.geometry

pix = np.random.choice(np.arange(len(cam)), 50)

print("Before to_table")
get_ipython().run_line_magic('timeit', 'cam[pix]')

t = cam.to_table()
print("After to_table")
get_ipython().run_line_magic('timeit', 'cam[pix]')
```


On master:
```
Before to_table
27.2 µs ± 2.59 µs per loop (mean ± std. dev. of 7 runs, 10,000 loops each)
After to_table
20.9 µs ± 1.33 µs per loop (mean ± std. dev. of 7 runs, 10,000 loops each)
```

Here:
```
Before to_table
27.9 µs ± 670 ns per loop (mean ± std. dev. of 7 runs, 10,000 loops each)
After to_table
26.4 µs ± 726 ns per loop (mean ± std. dev. of 7 runs, 10,000 loops each)
```

This means to me, instead of removing this workaround, we should probably move it to `CameraGeometry.__init__`.

Astropy deepcopying these meta info objects in all operations is really annoying for performance, I opened a couple of issues, some got fixed already and are pending a new release.

PR #2182, opened 2023-01-04
Rename Containers 'Camera' to 'Telescope'

### LINKED ISSUE #2170: Rename  *Camera*Containers to *Telescope*Containers
Yes, I'd go so far to rename most of the containers from camera to telescope. After all, we store them in .tel not .cam

_Originally posted by @maxnoe in https://github.com/cta-observatory/ctapipe/pull/2168#discussion_r1050462522_

kosack:
Fixes #2170

Here's a quick first pass thanks to the magic of refactoring tools...

### DIFF
--- a/ctapipe/containers.py
+++ b/ctapipe/containers.py
@@ -14,10 +14,10 @@ from .core import Container, Field, Map
 __all__ = [
     "ArrayEventContainer",
     "ConcentrationContainer",
-    "DL0CameraContainer",
+    "DL0TelescopeContainer",
     "DL0Container",
     "DL1CameraCalibrationContainer",
-    "DL1CameraContainer",
+    "DL1TelescopeContainer",
     "DL1Container",
     "DL2Container",
     "EventCalibrationContainer",
@@ -29,7 +29,7 @@ __all__ = [
     "CoreParametersContainer",
     "ImageParametersContainer",
     "LeakageContainer",
-    "MonitoringCameraContainer",
+    "TelescopeMonitoringContainer",
     "MonitoringContainer",
     "MorphologyContainer",
     "BaseHillasParametersContainer",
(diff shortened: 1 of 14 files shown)


maxnoe:
@kosack fix merged in #2183 , you can rebase master

ctapipe/containers.py
```diff
@@ -29,7 +29,7 @@
     "CoreParametersContainer",
     "ImageParametersContainer",
     "LeakageContainer",
-    "MonitoringCameraContainer",
+    "TelescopeMonitoringContainer",
```
nbiederbeck, line 32: Why the change of the order of Monitoring and Telescope?

See my general comment: this change is no longer greppable.

nbiederbeck (commented):
I grepped for `cameracontainer`, and found those in docstrings:

```rg
ctapipe/image/extractor.py
405:        DL1CameraContainer:

ctapipe/calib/camera/pedestals.py
53:    Fills the MonitoringCameraContainer.PedestalContainer on the base of a given pedestal sample.
92:        Fills the MonitoringCameraContainer.PedestalContainer on the base of a given pedestal sample.
211:        DL1CameraContainer

ctapipe/calib/camera/flatfield.py
23:    Fills the MonitoringCameraContainer.FlatfieldContainer on the base of a given
65:        Fills the MonitoringCameraContainer.FlatfieldContainer on the base of a given
182:        DL1CameraContainer
```

ctapipe/containers.py
kosack: For class naming usually you prepend less general terms, so it may even be better to change all to the form: `DL1Container` + `TelecopeDL1Container`, rather than `DL1Container`+`DL1TelescopeContainer`.   But the current form does also make some sense, it just reads poorly

nbiederbeck: FWIW, I also favor `DL1Container` + `TelescopeDL1Container`, then the former has the implicit `Array`-DL1Container

maxnoe:
Since this will be breaking a lot of code, we might want to tackle this (very easy change) with the inverted container hierarchy, to only majorly break the container structure once.

But that will be a much larger change.


On the other hand, this change is trivial for us, but results in large breakage and no real advantage, so it is not urgent. The other change however, will make much code much simpler (i.e. remove all internal loops in components only caring about single telescopes).

ctapipe/containers.py
kosack: We could even explicitly call it ArrayDL1Container

kosack:
> Since this will be breaking a lot of code, we might want to tackle this (very easy change) with the inverted container hierarchy, to only majorly break the container structure once.

Yes, this may be a better idea - then the renaming can be done at the same time

maxnoe:
So shall we close this in favor of #2204 ?

kosack:
closing since superseded by #2204

PR #2188, opened 2023-01-06
Flashcam extractor

clara-escanuela:


ctapipe/image/tests/test_extractor.py
```diff
@@ -617,3 +635,46 @@ def test_global_peak_window_sum_with_pixel_fraction(subarray):
...
+
+    # Test on prod5 simulations
+    with EventSource(prod5_gamma_simtel_path) as source:
+        subarray = source.subarray
+        extractor = FlashCamExtractor(subarray)
+        for event in source:
+            for tel_id in subarray.get_tel_ids_for_type("MST_MST_FlashCam"):
+                true_charge = event.simulation.tel[tel_id].true_image
+                if true_charge is None:
+                    continue  # telescope did not trigger
```
maxnoe, line 666: Or the input file does not contain true images! If you want to check something, you should do expicitly the check for what you are checking, i.e. here it would be:

```
if tel_id in event.trigger.tels_with_trigger
```

Or maybe just loop over the `tels_with_trigger`,  using a filter function to only yield flashcam ids:

```
def is_flashcam(tel_id):
    return subarray.tel[1].camera.name == "FlashCam" 

for tel_id in filter(is_flashcam, event.trigger.tels_with_trigger):
    
```

ctapipe/image/extractor.py
```diff
@@ -1285,3 +1290,373 @@ def __call__(
...
+    Parameters
+    ----------
+    waveforms : ndarray
+        Waveforms stored in a numpy array.
+        Shape: (n_pix, n_samples)
+    peak_index : ndarray or int
+        Peak index for each pixel.
+    rel_descend_limit : ndarray or float
+        Fraction of the peak value down to which samples are accumulated in the centroid calculation.
+    cog : ndarray
```
maxnoe, line 1487: cog seems a strange name for this, maybe better `centroids`?

ctapipe/image/extractor.py
```diff
@@ -1285,3 +1290,373 @@ def __call__(
...
+    descend_limit = rel_descend_limit * peak_amplitude
+
+    sum = float64(0.0)
+    jsum = float64(0.0)
+    j = peak_index
+    while j >= 0 and waveforms[j] > descend_limit:
+        sum += waveforms[j]
+        jsum += j * waveforms[j]
+        j -= 1
+        if waveforms[j] > peak_amplitude:
```
maxnoe, line 1516: How can the waveform be larger than the peak amplitude?

ctapipe/image/extractor.py
```diff
@@ -1285,3 +1290,373 @@ def __call__(
...
+            tel_id: telescope.camera.readout.sampling_rate.to_value("GHz")
+            for tel_id, telescope in subarray.tel.items()
+        }
+
+        def time_profile_pdf_gen(std_dev: float):
+            if std_dev == 0:
+                return None
+            return scipy.stats.norm(0.0, std_dev).pdf
+
+        self.deconvolution_pars = {
```
maxnoe, line 1591: If you store the deconvolution parameters here, is the `lru_cache` still needed?

ctapipe/image/extractor.py
```diff
@@ -1285,3 +1290,373 @@ def __call__(
...
+        integration_window_shift = self.window_shift.tel[tel_id]
+        neighbour_sum_clipping = self.neighbour_sum_clipping.tel[tel_id]
+        leading_edge_timing = self.leading_edge_timing.tel[tel_id]
+        leading_edge_rel_descend_limit = self.leading_edge_rel_descend_limit.tel[tel_id]
+
+        pzs, gains, shifts, pz2ds = self.deconvolution_pars[tel_id]
+        pz, gain, shift, pz2d = pzs[0], gains[0], shifts[0], pz2ds[0]
+
+        if leading_edge_timing:
+            d_waveforms = deconvolve(waveforms, 0.0, upsampling, 1)
```
maxnoe, line 1623: This is calculated up here but only used much further down in another if statement. 

You should move it down to the place where it is actually used.

ctapipe/image/tests/test_extractor.py
```diff
@@ -617,3 +635,46 @@ def test_global_peak_window_sum_with_pixel_fraction(subarray):
 
     expected = np.average([29, 30, 31], weights=[5, 10, 3])
     assert np.allclose(dl1.peak_time[bright_pixels], expected / sample_rate)
+
+
+def test_flashcam_extractor(toymodel_1_MST_FC, prod5_gamma_simtel_path):
```
maxnoe, line 640: I'd prefer a test similar to the other ones on artificial input, not on an actually simulated event, where you test the different options result in the expected outcome, which is hard to do on simulated events without a "truth" for the waveform / charge including noise / peak time.

maxnoe: At least additionally. Testing that it works by comparing with a simulated true image is also fine, but you needed to specify a rather high threshold and tolerance, which makes this test brittle / not very sensitive to actual bugs.

ctapipe/image/extractor.py
```diff
@@ -1285,3 +1290,373 @@ def __call__(
...
+    leading_edge_shifts : list of floats
+        Offset of the leading edge peak w.r.t. the deconvolved peak.
+    """
+    if upsampling < 1:
+        raise ValueError(f"upsampling must be > 0, got {upsampling}")
+    if window_width < 1:
+        raise ValueError(f"window_width must be > 0, got {window_width}")
+
+    ref_pulse_shapes = camera.readout.reference_pulse_shape
+    ref_sample_width_nsec = camera.readout.reference_pulse_sample_width.to_value("ns")
```
maxnoe, line 1347: Use `u.ns` and `u.GHz`, this avoids string parsing

ctapipe/image/extractor.py
```diff
@@ -1285,3 +1290,373 @@ def __call__(
...
+
+        integral = (
+            ref_pulse_shape.sum() * ref_sample_width_nsec / camera_sample_width_nsec
+        )
+        phase_gains, phase_shifts, phase_pz2d_shifts = [], [], []
+        for phase in range(avg_step):
+            x = np.atleast_2d(ref_pulse_shape[phase::avg_step])
+            y = deconvolve(x, 0.0, upsampling, pz)[0]
+            i_max = y.argmax()
+            start = i_max - window_shift  # TODO should use extract_around_peak here
```
maxnoe, line 1388: Please fix todo if really needed or remove

ctapipe/image/extractor.py
```diff
@@ -1285,3 +1290,373 @@ def __call__(
...
+    Returns
+    -------
+    deconvolved_waveforms : ndarray
+        Deconvolved and upsampled waveforms stored in a numpy array.
+        Shape: (n_pix, upsampling * n_samples)
+    """
+    deconvolved_waveforms = np.atleast_2d(waveforms) - np.atleast_2d(baselines).T
+    deconvolved_waveforms[:, 1:] -= pole_zero * deconvolved_waveforms[:, :-1]
+    deconvolved_waveforms[:, 0] = 0
+    if upsampling > 1:
```
maxnoe, line 1453: filtfilt is quite slow and the filter here is rather simple. Is there a straight-forward way to implement this? Then the whole think might be numba jittable

ctapipe/image/extractor.py
```diff
@@ -1285,3 +1290,373 @@ def __call__(
...
+    cog : ndarray
+        Return argument for ufunc (ignore)
+        Returns the peak centroid in units "samples"
+
+    Returns
+    -------
+    cog : ndarray
+        Peak centroid in units "samples"
+    """
+    cog[0] = peak_index  # preload in case of errors
```
maxnoe, line 1496: What kinds of errors are expected here?

ctapipe/image/extractor.py
```diff
@@ -1285,3 +1290,373 @@ def __call__(
...
+
+    nsamples = waveforms.size
```
maxnoe, line 1498: `n_samples`

ctapipe/image/extractor.py
```diff
@@ -1285,3 +1290,373 @@ def __call__(
...
+    if nsamples == 0:
+        return
+
+    peak_index = max(0, min(peak_index, nsamples - 1))
```
maxnoe, line 1502: I think you should just raise an error here if the peak index is out of bounds and not silently truncate it to the valid range.

ctapipe/image/extractor.py
```diff
@@ -1285,3 +1290,373 @@ def __call__(
...
+        jsum += j * waveforms[j]
+        j += 1
+        if waveforms[j] > peak_amplitude:
+            descend_limit = rel_descend_limit * peak_amplitude
+
+    if sum != 0.0:
+        cog[0] = jsum / sum
+
+
+class FlashCamExtractor(ImageExtractor):
```
maxnoe, line 1531: Missing docstring. Looking at the complexity, this needs a lot of information and references. At least one to the design of the flashcam electronics that make this necessary and a general description of the approach.

maxnoe:
Hi @clara-escanuela 

In your PR the CI is failing, at least with the static code checks (some style issues).

We recently also updated the way these checks are done, so please rebase the current main branch and fix the issues

clara-escanuela:
Only one check fails (codacy). I checked and there are three critical points:
1) It complains because centroids is not defined in the function but that is because, before deconvolve_parameters, there is a structure which "predefines" the variable. Same for critical point 2).
3) Error about dl1.image not being subscriptable.  I checked and dl1.image is an array, as expected, so I don't understand the error message. And the flashcam test passes.

maxnoe:
@clara-escanuela Yes, unfortunately codacy tends to create some false positives.

We are planning to replace it, right now don't worry too much. I will have do another  review pass here again soon but it already looks quite good.

clara-escanuela:
@maxnoe this implementation is in very good agreement with filtfilt and is way faster. You could check again the speed, should be better. The charge resolution does not change at all, so the upsampling is good. Although, maybe the function should be defined in another file where some other general tools are also implemented, so that people can have use it for upsampling.

ctapipe/image/extractor.py
```diff
@@ -1285,3 +1289,400 @@ def __call__(
...
+    -------
+    deconvolved_waveforms : ndarray
+        Deconvolved and upsampled waveforms stored in a numpy array.
+        Shape: (n_pix, upsampling * n_samples)
+    """
+    deconvolved_waveforms = np.atleast_2d(waveforms) - np.atleast_2d(baselines).T
+    deconvolved_waveforms[:, 1:] -= pole_zero * deconvolved_waveforms[:, :-1]
+    deconvolved_waveforms[:, 0] = 0
+
+    def filtfilt_custom(signal, filt):
```
maxnoe, line 1454: This function does not seem to need a closure, so I'd define it as a private function `_filtfilt_fast` outside of this here.

I'd also add a reference to the scipy issue about the scipy version being slow on ND input.

ctapipe/image/extractor.py
```diff
@@ -1285,3 +1289,400 @@ def __call__(
...
+        j += 1
+        if waveforms[j] > peak_amplitude:
+            descend_limit = rel_descend_limit * peak_amplitude
+
+    if sum_ != 0.0:
+        centroids[0] = jsum / sum_
+
+
+class FlashCamExtractor(ImageExtractor):
+    """
```
maxnoe, line 1541: This docstring is missing a small header line like "Image extractor applying signal preprocessing for FlashCam"

maxnoe (commented):
This looks great, and with the latest changes it also got a lot faster.

Checking a north alpha run with only flashcams in `allowed_tels` I get:

```
FlashCam:           00:54, 17.14ev/s
NeighborPeak: : 00:21, 44.24ev/s
``` 

Which seems like what we can expect since the flashcam extractor is essentially neighbor peak with addional steps. So for me this looks good enough to merge this here and further optimizations if possible can come in a follow up PR. 


Just a last minor comment on the new filtfilt function and I think we are ready to go.

Thanks a lot for the great work!

clara-escanuela:
@fwerner are you also happy with the result?

maxnoe (approved):

ctapipe/image/extractor.py
```diff
@@ -1285,3 +1289,406 @@ def __call__(
...
+            t_waveforms,
+            peak_index,
+            integration_window_width,
+            integration_window_shift,
+            self.sampling_rate_ghz[tel_id] * upsampling,
+        )
+
+        if leading_edge_timing:
+            d_waveforms = deconvolve(waveforms, 0.0, upsampling, 1)
+            peak_time = adaptive_centroid(
```
Tobychev, line 1682: Also here I can only see three arguments being given?

ctapipe/image/extractor.py
```diff
@@ -1285,3 +1289,406 @@ def __call__(
...
+            y = deconvolve(x, 0.0, upsampling, pz)[0]
+            i_max = y.argmax()
+            start = i_max - window_shift
+            stop = start + window_width
+            if start >= 0 and stop <= y.size:
+                if leading_edge_timing:
+                    d = deconvolve(x, 0.0, upsampling, 1)[0]
+                    d_pk_idx = d.argmax()
+                    phase_pz2d_shifts.append(i_max - d_pk_idx)
+                    time = adaptive_centroid(
```
Tobychev, line 1395: `adaptive_centroid` seems to take 4 arguments, but I can only see that 3 are provided, is this intentional? (Found by codacy)

ctapipe/image/tests/test_extractor.py
Tobychev, line 1: I think it is necessary to tests to check that `adaptive_centroid`, `deconvolve` and `deconvolution_parameters` produce the expected results on their own: relying on just one integrated test to catch regressions will make updates that touch more that one of these functions more tricky to debug.

maxnoe: Good point

ctapipe/image/extractor.py
maxnoe: This is how the `guvectorize` decorator works, it changes the signature and the missing argument is actually the return value, that gets pre-allocated by the decorator, passed to the compiled function as mutable input argument but then returned. 

https://numba.pydata.org/numba-doc/dev/user/vectorize.html#the-guvectorize-decorator

maxnoe: Unfortunately, this is somewhat opaque to static analysis

ctapipe/image/extractor.py
```diff
@@ -1285,3 +1289,406 @@ def __call__(
...
+        filt_weighted = filt / upsampling
+        signal = np.repeat(deconvolved_waveforms, upsampling, axis=-1)
+        return __filtfilt_fast(signal, filt_weighted)
+
+    return deconvolved_waveforms
+
+
+@guvectorize(
+    [
+        (float64[:], int64, float64, float32[:]),
```
maxnoe, line 1475: The guvectorize docs state:

> If there are several signatures, they must be ordered from the more specific to the least specific. Otherwise, Numpy’s type-based dispatching may not work as expected. For example, the following is wrong:
> ```@vectorize(["float64(float64)", "float32(float32)"])
def f(x): ...
```

So I think we need to switch the order here

File: ctapipe/image/tests/test_extractor.py [hunk 19]
```diff
@@ -675,6 +678,51 @@ def test_adaptive_centroid(toymodel_mst_fc):
...
+
+    upsampling = 4
+    filt = np.ones(upsampling)
+    filt_weighted = filt / upsampling
+    signal = np.repeat(waveforms, upsampling, axis=-1)
+    up_waveforms = __filtfilt_fast(signal, filt_weighted)
+
+    assert (
+        up_waveforms
+        - filtfilt(
```
maxnoe, line 718: Nice! comparing our custom solution with scipy makes sense

maxnoe (approved):

maxnoe (approved):

kosack (approved):
This looks very good! I don't see any issues with the code, and it seems it's well-covered by unit tests.

I did a few ad-hoc tests with some data as well as look at images with both NeighborPeak and FlashCamExtractor, and everything looks reasonable. 

One can see clearly that the background is now reduced significantly while the signal is maintained over what is given by a naive application of NeighborPeak to the FlashCam events. 
FC_NP_compare
extractor_compare_2
extractor_compare_1

maxnoe:
@kosack sorry, I forgot to post the plots, I did some similar tests and came to the same conclusion.

maxnoe:
Y-shaped shower in the second example? @jsitarek :D

clara-escanuela:
You can also see that here, 
image

PR #2190, opened 2023-01-06
Refactor initialization of telescope parameters, move to own module

maxnoe:
I came across code I didn't first understand in the initialization of the `TelescopeParameter` traits, trying to improve the code, I ran into a circular import problem (as the straight forward code would need to check `isinstance(trait, TelescopeParameter` and `TelescopeComponent` is in a module imported by `ctapipe.core.traits`).

So the best solution was to move the `TelescopeComponent` / `TelescopeParameter` related things into their own submodule `ctapipe.core.telescope_component`.

I think this is an improvement.

The main change (aside from moving things around) is replacing this code:

```python
# configure all of the TelescopeParameters
for trait in list(self.class_traits()):
    try:
        getattr(self, trait).attach_subarray(subarray)
    except (AttributeError, TypeError):
        pass
```

with

```python
# configure all of the TelescopeParameters
for attr, trait in self.class_traits().items():
    if not isinstance(trait, TelescopeParameter):
        continue

    # trait is the TelescopeParameter descriptor at the class,
    # need to get the value at the instance, which will be a TelescopePatternList
    pattern_list = getattr(self, attr)
    if pattern_list is not None:
        pattern_list.attach_subarray(subarray)
```

which is I think much clearer.

This also has the additional benefit of not silently ignoring errors that could potentially happen in `attach_subarray`.


ctapipe/core/telescope_component.py
```diff
@@ -0,0 +1,382 @@
...
+        parent : ctapipe.core.Tool
+            Tool executable that is calling this component.
+            Passes the correct logger and configuration to the component.
+            This argument is typically only specified when using this method
+            from within a Tool (config need not be passed if parent is used).
+        kwargs
+
+        Returns
+        -------
+        instace
```
nbiederbeck, line 73:
```suggestion
        Instance
```

ctapipe/core/telescope_component.py
```diff
@@ -0,0 +1,382 @@
...
+        """Validate and then append a new value"""
+        super().append(self.single_to_pattern(value))
+
+    def attach_subarray(self, subarray):
+        """
+        Register a SubarrayDescription so that the user-specified values can be
+        looked up by tel_id. This must be done before using the ``.tel[x]`` property
+        """
+        self._subarray = subarray
+        self._lookup.attach_subarray(subarray)
```
nbiederbeck, line 130: Does this work when in init `self._lookup = None`?

ctapipe/core/telescope_component.py
```diff
@@ -0,0 +1,382 @@
...
+        """
+        self._telescope_parameter_list = copy.deepcopy(telescope_parameter_list)
+        self._value_for_tel_id = None
+        self._value_for_type = None
+        self._subarray = None
+        self._subarray_global_value = Undefined
+        self._type_strs = None
+        for param in telescope_parameter_list:
+            if param[1] == "*":
+                self._subarray_global_value = param[2]
```
nbiederbeck, line 152: This takes the last `*` as the default, overwriting previous. Is this intended and documented?

ctapipe/core/telescope_component.py
```diff
@@ -0,0 +1,382 @@
...
+        if not isinstance(trait, TraitType):
+            raise TypeError("trait must be a TraitType instance")
+
+        self._trait = trait
+        self.allow_none = kwargs.get("allow_none", False)
+        if default_value != Undefined:
+            default_value = self.validate(self, default_value)
+
+        if "help" not in kwargs:
+            self.help = "A TelescopeParameter"
```
nbiederbeck, line 297: Maybe this should raise an exception, to have everything documented?

ctapipe/core/tests/test_telescope_component.py
```diff
@@ -0,0 +1,365 @@
...
+                ("type", "*", 10.0),
+                ("type", "LST_LST_LSTCam", 100.0),
+                ("type", "*", 200.0),  # should overwrite everything with 200.0
+                ("id", 100, 300.0),
+            ]
+        )
+
+    comp = SomeComponent()
+
+    # need to mock a SubarrayDescription
```
nbiederbeck, line 215: you did this in a fixture, right?

ctapipe/core/tests/test_telescope_component.py
```diff
@@ -0,0 +1,365 @@
...
+                ("type", "LST_LST_LSTCam", 100.0),
+                ("id", 3, 200.0),
+            ]
+        )
+
+        tel_param3 = FloatTelescopeParameter(
+            default_value=[
+                ("type", "*", 10.0),
+                ("type", "LST_LST_LSTCam", 100.0),
+                ("type", "*", 200.0),  # should overwrite everything with 200.0
```
nbiederbeck, line 208:
```suggestion
                ("type", "*", 200.0),  # should overwrite everything above with 200.0
```

ctapipe/core/telescope_component.py
nbiederbeck: I assume this would not necessarily be the case, but we might type-hint in `__init__`

ctapipe/core/telescope_component.py
nbiederbeck: It is documented in L.257

nbiederbeck (approved):

kosack (approved):
yes, this is much easier to follow. I didn't like relying on an exception to type-check

PR #2199, opened 2023-01-10
Convert container input in validate

### LINKED ISSUE #2147: Container.validate should convert input if possible

Currently, if the field value has not the correct type, `Container.validate` raises an error. This is very annoying for cases like e.g. the SchedulingBlockContainer where we define `sb_id` to be `np.uint64`: setting it to a python int is not possible, one has explicitly to set `np.uint64(value)`.

Validate should convert types if possible.

Let behavior stay like now, probably needs type conversion code at many places.

LukasNickel:
Fixes #2147 

Noteworthy things:
- I used `np.can_cast` to check if the conversion is possible. I hope to avoid most `np.nan` issues this way. 
- Converting and then checking for equal values would be insufficient in the nan case. One would need to compare only the non-nan values (which is possible, I just decided against it).
- `can_cast` fails for scalar strings, which is why I only check numbers in the scalar case

### DIFF
--- a/ctapipe/core/container.py
+++ b/ctapipe/core/container.py
@@ -190,18 +190,23 @@ class Field:
                     f"{errorstr} Should have dimensionality {self.ndim}"
                 )
             if value.dtype != self.dtype:
-                raise FieldValidationError(
-                    f"{errorstr} Has dtype "
-                    f"{value.dtype}, should have dtype"
-                    f" {self.dtype}"
-                )
-        else:
+                if np.can_cast(value, self.dtype, casting="safe"):
+                    value = value.astype(self.dtype)
+                else:
+                    raise FieldValidationError(
+                        f"{errorstr} Has dtype "
+                        f"{value.dtype}, should have dtype"
+                        f" {self.dtype}"
+                    )
             # not a numpy array
             if self.dtype is not None:
-                if not isinstance(value, self.dtype.type):
-                    raise FieldValidationError(
-                        f"{errorstr} Should have numpy dtype {self.dtype}"
-                    )
+                if np.can_cast(value, self.dtype, casting="safe"):
+                    value = value.astype(self.dtype)
+                else:
+                    if not isinstance(value, self.dtype.type):
+                        raise FieldValidationError(
+                            f"{errorstr} Should have numpy dtype {self.dtype}"
+                        )
 
 
 class DeprecatedField(Field):
--- a/ctapipe/core/container.py
+++ b/ctapipe/core/container.py
@@ -190,20 +190,20 @@ class Field:
                     f"{errorstr} Should have dimensionality {self.ndim}"
                 )
             if value.dtype != self.dtype:
-                if np.can_cast(value, self.dtype, casting="safe"):
-                    value = value.astype(self.dtype)
-                else:
+                try:
+                    value = value.astype(self.dtype, casting="same_kind")
+                except TypeError:
                     raise FieldValidationError(
                         f"{errorstr} Has dtype "
                         f"{value.dtype}, should have dtype"
-                        f" {self.dtype}"
+                        f" {self.dtype} and could not cast it."
                     )
         else:
             # not a numpy array
             if self.dtype is not None:
                 # Workaround for string input
                 if np.issubdtype(type(value), np.number) and (
-                    np.can_cast(value, self.dtype, casting="safe")
+                    np.can_cast(value, self.dtype, casting="same_kind")
                 ):
                     value = self.dtype.type(value)
                 else:
--- a/ctapipe/core/container.py
+++ b/ctapipe/core/container.py
@@ -198,10 +198,14 @@ class Field:
                         f"{value.dtype}, should have dtype"
                         f" {self.dtype}"
                     )
+        else:
             # not a numpy array
             if self.dtype is not None:
-                if np.can_cast(value, self.dtype, casting="safe"):
-                    value = value.astype(self.dtype)
+                # Workaround for string input
+                if np.issubdtype(type(value), np.number) and (
+                    np.can_cast(value, self.dtype, casting="safe")
+                ):
+                    value = self.dtype.type(value)
                 else:
                     if not isinstance(value, self.dtype.type):
                         raise FieldValidationError(
--- a/ctapipe/core/container.py
+++ b/ctapipe/core/container.py
@@ -189,28 +189,19 @@ class Field:
                 raise FieldValidationError(
                     f"{errorstr} Should have dimensionality {self.ndim}"
                 )
-            if value.dtype != self.dtype:
-                try:
-                    value = value.astype(self.dtype, casting="same_kind")
-                except TypeError:
-                    raise FieldValidationError(
-                        f"{errorstr} Has dtype "
-                        f"{value.dtype}, should have dtype"
-                        f" {self.dtype} and could not cast it."
-                    )
+
         else:
-            # not a numpy array
             if self.dtype is not None:
-                # Workaround for string input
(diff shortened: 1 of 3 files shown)


nbiederbeck (approved):

ctapipe/core/container.py
```diff
@@ -190,18 +190,27 @@ def validate(self, value):
...
+                    )
         else:
             # not a numpy array
             if self.dtype is not None:
-                if not isinstance(value, self.dtype.type):
-                    raise FieldValidationError(
-                        f"{errorstr} Should have numpy dtype {self.dtype}"
-                    )
+                # Workaround for string input
+                if np.issubdtype(type(value), np.number) and (
```
maxnoe, line 205: I think this workaround is (at least at the moment) not needed, as we don't support writing string arrays and python str are handled via type, not dtype.

ctapipe/core/container.py
```diff
@@ -190,18 +190,27 @@ def validate(self, value):
                     f"{errorstr} Should have dimensionality {self.ndim}"
                 )
             if value.dtype != self.dtype:
-                raise FieldValidationError(
-                    f"{errorstr} Has dtype "
-                    f"{value.dtype}, should have dtype"
-                    f" {self.dtype}"
-                )
+                if np.can_cast(value, self.dtype, casting="safe"):
```
maxnoe, line 193: What is the advantage here over just calling `np.astype(...,, casting="safe")` and catching the error if needed?

That should be faster for the "happy path"

ctapipe/core/container.py
LukasNickel: Some form of workaround is still needed for the case of the Field requiring a non-string and the user submitting a string.

maxnoe: See proposal above. Just reraise a `ValidationError("Validation failed") from err`

StFroese (approved):

docs/changes/2199.maintenance.rst
```diff
@@ -0,0 +1 @@
+Fields now convert input to their dtype if possible
```
nbiederbeck, line 1: When validating ..

ctapipe/core/tests/test_container.py
```diff
@@ -236,26 +236,32 @@ def test_field_validation():
     field_f.validate(np.ones((2, 2), dtype=np.float64))
 
     #   test dtype
+    field_f.validate(np.ones((2, 2), dtype=np.float32))
+    field_f.validate(np.ones((2, 2), dtype=np.int32))
     with pytest.raises(FieldValidationError):
```
nbiederbeck, line 241: `pytest.raises` has the `match` keyword, this allows us to match on `ndarray`, `dimensionality` and `dtype`, depending on why these tests should fail.  I think this increases the ease of understanding these tests.

nbiederbeck (approved):

kosack (commented):
I'm not sure this is a good idea - validate is a slow operation, so is not meant to run for every event, and currently we only validate the first event.  So having validate potentially change the data would break things.

maxnoe:
At least optionally, I think that could of caught a lot of errors...

kosack:
> At least optionally, I think that could of caught a lot of errors...

Optionally would be fine, but in normal operation it's usually not necessary and as it recursively visits all members, it just adds some overhead.   I still would not like that a command called `validate() → bool` affects the data itself, however! It should only check, not modify

LukasNickel:
> Did we actually measure this? Can we make it fast enough to be able to to run validate (or "coerce") for every container we write?

In the current implementation it is actually very slow. I ran `ctapipe-process` on a gamma file without validation and without any config, which took ~160s. 
Always validating adds another 200s, so basically doubling that. It is mostly the construction of the error string in
https://github.com/cta-observatory/ctapipe/blob/900631e73db2a7a36bbcf5905511fddc6bbbb103/ctapipe/core/container.py#L150
because the container `__str__` representation does some fancy `pprint` stuff.
I moved that into the `except` clauses and it looks much more manageable:
Screenshot_2023-01-11_16-27-39

It is still a sizeable chunk, but most of the time is spend dealing with units, which means we might be able to improve on that.
Currently in this very naive approach it:
- Doubles the time spent in `DataWriter`
- Increases total processing time from 160 to 180s 
Only for my test case of course, but I would assume that the numbers are representative at least for the simtel -> dl1 step.
It would be a very large fraction of processing time for processing dl1 files as thats usually faster.

> I still would not like that a command called `validate() → bool` affects the data itself, however! It should only check, not modify

I can see that. Are you generally opposed to this implicit transformations or do you feel like it should be part of the `DataWriter`?

StFroese:
So most of the time is spend here, right?

https://github.com/cta-observatory/ctapipe/blob/6e6e62f7f309d52124ef13eeed1786fc2ec9bbb3/ctapipe/core/container.py#L176-L177

Can you try converting the unit of the value instead the value directly?
```python
import astropy.units as u

a = 123 * u.deg
unit = u.rad
```
```python
In [15]: %timeit a.to(unit)
3.89 µs ± 4.66 ns per loop (mean ± std. dev. of 7 runs, 100,000 loops each)

In [16]: %timeit a.unit.to(unit)
1.52 µs ± 11.2 ns per loop (mean ± std. dev. of 7 runs, 1,000,000 loops each)
```

maxnoe:
> Can you try converting the unit of the value instead the value directly?

Just check `value.unit.is_equivalent(self.unit)`.

StFroese:
Little bit slower actually:
```python
In [22]: %timeit a.unit.is_equivalent(unit)
2.03 µs ± 9.9 ns per loop (mean ± std. dev. of 7 runs, 100,000 loops each)
```

maxnoe:
Interesting, I thought I fixed that... https://github.com/astropy/astropy/issues/13495

maxnoe:
@LukasNickel Can you open a PR only fixing that the error message is created always and not just when needed for now?

kosack:
I still don't see what problem this PR is trying to fix, except for a minor annoyance for the sb_id.

My main complaint here was that this changes the original design where the idea was:
* validation is there to check that the types are correct, and if incorrect this is a development mistake and the developer should fix the type they are putting in the event structure
* validation should not modify the data, only complain that the data are not valid! (validation != type conversion)
* Generally, using the wrong type is a code error, not a runtime error: I don't see a case where the incorrect type could be used for some events, but not others (except perhaps for null/unfilled events, but that could be fixed by moving where validate happens). 
* types should *not* be automatically coerced, rather the developer should do the conversion explicitly when filling the event structure, so it is clear that the type is changing.  Otherwise, this can lead to unexpected results in the future.

maxnoe:
@kosack We had many issues in the past with things being upcasted e.g. from float32 to float64 and then making sure again that what we write out if float32. I think it is entirely reasonable to define the intended storage type at the container field, doing computations in memory with whatever is most efficient and then converting to the intended storage type when writing out.

This is I think even more needed for types such as int8 / uint8 etc. that will be upcasted to int32 / int64 for any computations.

kosack:
> This is I think even more needed for types such as int8 / uint8 etc. that will be upcasted to int32 / int64 for any computations.

But isn't that exactly the point? You do a computation, and then should explicitly  and clearly drop precision when filling the event.  Otherwise, it's very easy to miss that the value was silently truncated to int8, leading to unexpected behavior.  It's more the  "explicit is better than implicit" argument.

maxnoe:
> You do a computation, and then should explicitly and clearly drop precision when filling the event.

But stuff gets computed and is used later by other things in computations. Doing these conversions at the IO boundary makes more sense to me than at every component boundary.

maxnoe:
Closing for now, let's discuss back in issues

PR #2203, opened 2023-01-10
Add coordinate overlay plotting to CameraDisplay

maxnoe:
A follow-up to the old PR in #1138, I here add a new method `CameraDisplay.overlay_coordinate` that can plot any `SkyCoor` into the `CameraDisplay` by first transforming it into the frame of the display's camera geometry.

This is the result of the image produced in the unit test:

coord_overlay


kosack (approved):

vuillaut (approved):
Thank you for reviving this!

Issue #2139, opened 2022-11-25
Telescope trigger information is stored in big table for all telescopes, not in tables per telescope


maxnoe:
**Describe the bug**

We write the telescope wise trigger information into one large table with all telescopes. This is different from every other telescope-wise quantity we store in hdf5 and it makes the loading of this information for a single telescope much slower (since you have to read the whole table and then filter for the telescope).

We should probably change this before 1.0 to be consistent in the data format.

maxnoe:
In the current ACADA DPPS ICD, there will actually be a single file for all telescope trigger information. So maybe we don't split it up?

However, I think this file will be indexed by `obs_id, trigger_id`, also including information on rejected triggers. So maybe it's something entirely different.

PR #2204, opened 2023-01-10
First step of implementing CEP-002: Change container structure

### LINKED ISSUE #1165: Discussion on restructuring Event Container structure

### LINKED ISSUE #2170: Rename  *Camera*Containers to *Telescope*Containers

### LINKED ISSUE #2374: Should trigger / pointing information be stored into in dl0?
At the moment we write the trigger information into the `/dl1/event/{subarray,telescope}` groups.

However, this is really dl0 information and we do not apply any correction which would justify making it dl1.

I guess this comes from "historic" times, where we wrote "pure" dl1 files and the data model document was not yet written.

I would propose that when we solve https://github.com/cta-observatory/ctapipe/issues/2139 we also change the data writer to write this into `/dl0/` instead.

### LINKED ISSUE #2952: R1/DL0 telescope event type is also stored in trigger.
**Describe the bug**
We store the R1/DL0 telescope event type in three containers, R1, DL0 and TelescopeTrigger.

**To Reproduce**

**Expected behavior**

Need to define which is the canonical place for this information and how it is moved through data levels..

maxnoe:
- Closes #1165
- Closes #2170 
- Closes #2374 
- Closes #2952


maxnoe:
Before I continue changing everything, could you review the new container structure?

I think it's already a big step in the right direction, as the structure is much simpler, we could get rid of many indirections...

nbiederbeck:
This comment just looked at `ctapipe/containers.py`.  In general I think the structure looks fine, but while we're at it, some comments:

- For the ``EventContainers``: and why do we have ``count``?

- Should there be a general ``MuonContainer``?

- Why does the ``TelescopePointingContainer`` have no ra/dec?
Or why does ``ArrayPointingContainer`` have it?

- Should we rename the ``ParticleClassificationContainer`` and ``TelescopeImpactParameterContainer`` to be more consistent with ``ReconstructedEnergyContainer`` and ``ReconstructedGeometryContainer``? E.g. ``ReconstructedParticleContainer``, ``ReconstructedImpactContainer``.

- Is ``Reconstructed[…].telescopes`` sensible to be None if Mono? Or should it just hold the corresponding tel_id and be of type int?

- Is ``TelescopeTriggerContainer.n_trigger_pixels`` with default -1 sensible? Or is 0 better to have an unsigned int?

- ``SimulationConfigContainer`` have many nans, but str seems more appropriate given the description.  Or None.

- ``TelescopeSimulationContainer.impact`` is not called ``true_impact``.

- ``MorphologyContainer``, is -1 a better default than 0?  Does it make a difference of default 0 and filled 0?

- Is ``CameraTimingParametersContainer`` still needed?

- ``LeakageContainer.intensity_width_{1,2}`` is ambiguously documented (intensity vs. fraction of intensity, as of the docstring).

- There are some ``Camera``-containers, which I think we can get rid of?

maxnoe:
> Why does the TelescopePointingContainer have no ra/dec?

Because all our telescopes are alt/az mounts, so this is the actual information that is monitored.

> Or why does ArrayPointingContainer have it?

Because that's the reference point the array is tracking, which might be in sky-fixed coordinates.

> TelescopeSimulationContainer.impact is not called true_impact.

Isn't this accomplished by adding the `true_` prefix in the data writer?

> Is TelescopeTriggerContainer.n_trigger_pixels with default -1 sensible? Or is 0 better to have an unsigned int?

I think it's important to distinguish invalid from valid data, 0 trigger pixels are possible (e.g. for pedestal events which are randomly triggered), so better distinguish. Same for the other positive integers.

maxnoe:
> For the EventContainers: and why do we have count?

Just nice to have I guess.

maxnoe:
> Is CameraTimingParametersContainer still needed?

That's connected to all the other `Camera...` containers and to #2141, we need them as long as we still support computing image parameters in `CameraFrame`

StFroese:
The `ArrayEventContainer` contains the DL2 data (geometry, energy, classification) of the event in the `ArrayDL2Container`.
It also contains the telescopes as a map of `TelescopeEventContainer`'s which contains all data levels like r0... and also dl2. dl2 in this case is a `TelescopeDL2Container` which is a subclass of the `ArrayDL2Container` again.

Why do we save the array event reconstructions at telescope level again? This is a little bit confusing to me

maxnoe:
> Why do we save the array event reconstructions at telescope level again? This is a little bit confusing to me

The `DL2TelescopeContainer` is for monoscopic predictions, not for repeating the subarray info.

The inheritance here is just to get the same fields, but I agree that it's confusing that `isinstance(DL2TelescopeContainer, DL2SubarrayContainer)`. I had an intermediate version where both just inherited from `Container` and then duplicated the field definitions.

StFroese:
> The inheritance here is just to get the same fields, but I agree that it's confusing that `isinstance(DL2TelescopeContainer, DL2SubarrayContainer)`. I had an intermediate version where both just inherited from `Container` and then duplicated the field definitions.

Can we add a `ReconstrucedShowerContainer` which contains the fields and `DL2{Telescope|Subarray}Container` inherit from that?

ctapipe/image/image_processor.py
```diff
@@ -209,51 +210,49 @@ def _parameterize_image(
         # parameterization
         return default
 
-    def _process_telescope_event(self, event):
+    def _process_telescope_event(self, tel_event):
```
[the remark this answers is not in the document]
maxnoe: It's a `TelescopeEventContainer`, but yes, I'll add type hints for these. However, these are not *terribly* helpful, since at least my lsp only looks at the class variable, not the instance variables or rather assumes they are of the same type.

maxnoe:
Hi all,

I rebased this on the current main and squashed the commits into one, to ease follow-up rebases when we make updates to main.

I'd appreciate a fresh review now.

I suggest to first only review the container structure in [ctapipe.containers](https://github.com/cta-observatory/ctapipe/pull/2204/files#diff-c81e53c1444d1fa2b1b800ff04655cae8cdffda5dee470ebafbe0b0f48781458) (beware, this is collapsed by default) before reviewing any of the resulting code changes.

Once we are happy with the container structure and renaming, I'll fix up the rest of the code and we can commence the review there.

ctapipe/containers.py
```diff
@@ -759,12 +748,11 @@ class SimulatedCameraContainer(Container):
     )
 
 
-class SimulatedEventContainer(Container):
+class SimulationSubarrayContainer(Container):
     shower = Field(
```
maxnoe, line 752: Right now, simulationsubarrayContainer has a single subcontainer, shower. 

We could just attach the `shower` container at the `SubarrayEventContainer` to get rid of one level of indirection.

ccossou:
> > For the EventContainers: and why do we have count?
> 
> Just nice to have I guess.

On this part. This, to me, is related to how people parse the file and shouldn't be in the container itself, or if it is, should be hidden and not publicly exposed (even if in python, private just means harder to type). But that's a minor comment and not the main purpose of CEP-002.

ctapipe/image/image_processor.py
```diff
@@ -209,51 +211,48 @@ def _parameterize_image(
...
+        ):
+            simulation = tel_event.simulation
+            simulation.true_parameters = self._parameterize_image(
+                tel_id,
+                image=simulation.true_image,
+                signal_pixels=simulation.true_image > 0,
+                peak_time=None,  # true image from simulation has no peak time
+                default=DEFAULT_TRUE_IMAGE_PARAMETERS,
             )
+            for container in simulation.true_parameters.values():
```
ccossou, line 251: If the prefix doesn't start with "true_", doesn't that mean there's something wrong with the file and we should raise an error rather than silently correcting it?

ctapipe/image/image_processor.py
```diff
@@ -209,51 +211,48 @@ def _parameterize_image(
...
 
-            dl1_camera.parameters = self._parameterize_image(
-                tel_id=tel_id,
-                image=dl1_camera.image,
-                signal_pixels=dl1_camera.image_mask,
-                peak_time=dl1_camera.peak_time,
-                default=self.default_image_container,
+        self.log.debug("params: %s", dl1.parameters.as_dict(recursive=True))
+
+        if (
```
ccossou, line 239: Why do we test for tel_event.simulation here? I struggle to understand the practical case. Would it be valid if true_image is not None but tel_event.simulation is None?

ccossou:
(Sorry, commented on file without realizing it wasn't container.py anymore, got carried away.)

ctapipe/image/image_processor.py
maxnoe: There are observed events, where `event.simulation is None` and there are simulations, which don't have the `true_image` information, so `event.simulation` is there, but `event.simulation.true_image` is None.

We need to test both.

maxnoe: And python only evaluates the second expression if the first ist true, this is why we need to check first `tel_event.simulation is not None` and then (which will only be checked if we do have simulation) check if the true image is available

ctapipe/image/image_processor.py
maxnoe: > If the prefix doesn't start with "true_", doesn't that mean there's something wrong with the file 

Has nothing to do with the input file, since the container is created by the `_parametrize_image` methods, but you are right, this is confusing.

I will check if it is needed / can be improved.

ctapipe/containers.py
```diff
@@ -1020,35 +983,26 @@ class ReconstructedContainer(Container):
     )
 
 
-class TelescopeReconstructedContainer(ReconstructedContainer):
-    """Telescope-wise reconstructed quantities"""
+class DL2SubarrayContainer(ReconstructedShowerContainer):
+    """
+    DL2 subarray-wise information (reconstructed air shower properties)
+    """
+
```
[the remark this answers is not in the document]
maxnoe: This comes from a review comment before, where it was remarked that it's weird that the `DL2TelescopeContainer` inherits from the `DL2SubarrayContainer`.

So instead, we have a common base class `ReconstructedShowerContainer` for both, but the `DL2SubarrayContainer` doesn't actually define more fields than the base class.

ctapipe/containers.py
```diff
@@ -1020,35 +983,26 @@ class ReconstructedContainer(Container):
...
+
+class DL2TelescopeContainer(ReconstructedShowerContainer):
```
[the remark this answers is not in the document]
maxnoe: It should be the type of `TelescopeEventContainer.dl2`, let me check

ctapipe/containers.py
```diff
@@ -1368,43 +1289,88 @@ class SimulatedShowerDistribution(Container):
...
-    dl2 = Field(default_factory=DL2Container, description="DL2 reconstruction info")
-    simulation = Field(
-        None, description="Simulated Event Information", type=SimulatedEventContainer
+    dl1 = Field(
+        default_factory=DL1TelescopeContainer, description="DL1 images and parameters"
     )
-    trigger = Field(
-        default_factory=TriggerContainer, description="central trigger information"
+    dl2 = Field(
+        default_factory=DL2TelescopeContainer,
```
maxnoe, line 1316: @Tobychev here it is used. If it is missing in `__all__` I'll add it

Tobychev: Yeah, looks to be missing from `__all__` to me.

maxnoe:
I edited the notebooks using jupytext in the notebook interface but that resulted in some broken formatting in the rendered docs.

How can we configure jupytext to actually use the format expected by sphinx-gallery?

StFroese:
Hi @maxnoe, not sure atm. My guess would be configuring a jupytext.toml.
Maybe take a look here https://jupytext.readthedocs.io/en/latest/formats-scripts.html

maxnoe:
It seems that sphinx-gallery expects rest, but jupytext only supports markdown...

maxnoe:
Wow, resolving the conflicts here was no fun.... I'd really appreciate if we can tackle this one next. 

I will do a self-review again and it would be nice if others could start reviewing the structure in `src/ctapipe/containers.py`

One particular issue (noted with a fixme in the CameraCalibrator) is that we (in accordance with the CTAO data model) moved the trigger information to dl0, but the `SimTelEventSource` and also other event source implementations are allowed to return event data with only R1 inside. 

This has the issue of the `CameraCalibrator.r1_to_dl0` method needing to fill trigger information... which it now does by checking if there is a pre-existing dl0 container. Not so nice conceptually.

I see two solutions:

* Putting trigger info also at R1 so that e.g. CameraCalibrator can just forward it from r1 to dl0
* Requiring event sources to provide dl0, removing the dvr step from `CameraCalibrator` so that it only deals with the calibration from dl0 to dl1

I think the latter is the option that is more closely aligned with how we expect actual CTAO operations to be.

Studies for DVR can still be done, if we for example at dvr options to simteleventsource or in an intermediate step.

src/ctapipe/calib/camera/calibrator.py
```diff
@@ -291,13 +300,14 @@ def __call__(self, event):
...
         """
-        # TODO: How to handle different calibrations depending on tel_id?
-        tel = event.r1.tel or event.dl0.tel or event.dl1.tel
-        for tel_id in tel.keys():
-            self._calibrate_dl0(event, tel_id)
-            self._calibrate_dl1(event, tel_id)
+        for tel_event in event.tel.values():
+            self.calibrate_tel_event(tel_event)
+
+    def calibrate_tel_event(self, tel_event):
```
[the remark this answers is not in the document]
maxnoe: Yes, it's essentially a preparation for changing the API. 

As discussed in CEP 2, things that will only ever work on telescope events, should work on telescope events.

So once we also change that (which I would do in a second PR), We remove the `__call__` with the internal loop and `calibrate_tel_event` becomes the new `__call__`.

src/ctapipe/containers.py
```diff
@@ -1388,43 +1322,88 @@ class SimulatedShowerDistribution(Container):
...
+        default_factory=DL0SubarrayContainer,
+        description="DL0 subarray wide information",
+    )
+
+    dl2 = Field(
+        default_factory=DL2SubarrayContainer,
+        description="DL2 reconstructed event properties",
+    )
+
+    pointing = Field(
```
maxnoe, line 1396: Move into dl0?

Tobychev: Not sure, is it very helpful? 

Putting pointing in dl0 means you can't make geometry reconstruction using pure dl2 files, but maybe that's fine? Not sure what other costs there are.

src/ctapipe/io/datawriter.py
```diff
@@ -297,50 +289,180 @@ def __enter__(self):
...
-        if self.write_r1_waveforms:
-            self._write_r1_telescope_events(event)
+    def _write_telescope_event(self, tel_event):
+        table_name = self.table_name(tel_event.index.tel_id)
+        self._writer.write(
+            f"simulation/event/telescope/impact/{table_name}",
+            [tel_event.index, tel_event.simulation.impact],
+        )
+
+        if tel_event.simulation is not None:
```
Tobychev, line 326: Why is `is not None` needed here?

src/ctapipe/io/datawriter.py
```diff
@@ -297,50 +289,180 @@ def __enter__(self):
...
+        table_name = self.table_name(tel_event.index.tel_id)
+        tel_event.r1.prefix = ""
+        self._writer.write(
+            f"r1/event/telescope/{table_name}",
+            [tel_event.index, tel_event.r1],
+        )
+
+    def _write_dl0_telescope(self, tel_event: TelescopeEventContainer):
+        self._writer.write(
+            "dl0/event/telescope/trigger",
```
Tobychev, line 366: Shouldn't there be a `{table_name}` here? Maybe I'm misunderstanding what the `_telescope` part of the function name implies.

src/ctapipe/io/datawriter.py
```diff
@@ -353,19 +475,21 @@ def _write_constant_pointing(self, event):
...
+
+        if tel_id in self._constant_telescope_pointing_written[obs_id]:
+            return
+
+        index = TelescopeConfigurationIndexContainer(
+            obs_id=obs_id,
+            tel_id=tel_id,
+        )
+        self._writer.write(
+            f"configuration/telescope/pointing/tel_{tel_id:03d}",
```
Tobychev, line 489: shouldn't this use `self.table_name` to construct a table name?

src/ctapipe/io/hdf5eventsource.py
```diff
@@ -622,58 +641,46 @@ def _generator(self):
...
+                    tel_id,
+                    time=tel_trigger.time,
+                    tel_pointing_interpolator=pointing_interpolator,
+                )
 
                 if DataLevel.R1 in self.datalevels:
-                    data.r1.tel[tel_id] = next(waveform_readers[key])
+                    containers["r1"] = next(waveform_readers[key])
 
-                    # TODO: Delete if we drop support for datamodel versions < v6.0.0
```
Tobychev, line 656: This comment states that the deletion of the following line implies a change in supported data model versions, is this true? In that case the changelog needs to mention it.

kosack (dismissed):
I don't see any major problem.

examples/tutorials/raw_data_exploration.py
```diff
@@ -73,49 +68,48 @@
 
 print(event.simulation.shower)
 
-######################################################################
-print(event.r0.tel.keys())
-
+# %%
+print(event.tel.keys())
```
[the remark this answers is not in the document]
maxnoe: This prints the `.keys()`, so only the tel_ids

examples/visualization/array_display.py
```diff
@@ -22,7 +23,7 @@
 
 plt.rcParams["figure.figsize"] = (8, 6)
 
-######################################################################
+""
```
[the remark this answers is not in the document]
maxnoe: reverted the change, this was caused by opening the example scripts using jupyter lab with the jupytext extension, which unfortunately breaks the sphinx-gallery syntax.

src/ctapipe/image/reducer.py
```diff
@@ -178,15 +178,13 @@ def __init__(
             self.image_extractor_type = [("type", "*", name)]
             self.image_extractors[name] = image_extractor
 
-    def select_pixels(self, waveforms, tel_id=None, selected_gain_channel=None):
+    def select_pixels(self, waveforms, tel_id, selected_gain_channel=None):
```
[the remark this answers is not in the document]
maxnoe: Because it was already a required value (leaving it None would result in an exception).

src/ctapipe/io/datawriter.py
maxnoe: Because observed data do not have simulation information

src/ctapipe/io/datawriter.py
maxnoe: This is an open issue, see https://github.com/cta-observatory/ctapipe/issues/2139.

I didn't change the data format here, so the trigger table doesn't have a `{table_name}`

src/ctapipe/io/datawriter.py
maxnoe: Good catch. Fixed.

src/ctapipe/io/hdf5eventsource.py
maxnoe: This is a merge artifact I think... will fix.

src/ctapipe/io/tableloader.py
```diff
@@ -269,7 +272,19 @@ def __exit__(self, exc_type, exc_value, traceback):
 
     def __len__(self):
         """Number of subarray events in input file"""
-        return self.h5file.root[TRIGGER_TABLE].shape[0]
+        return self.h5file.root[self._trigger_table].shape[0]
+
+    @property
+    def _trigger_table(self):
+        if TRIGGER_TABLE_OLD in self.h5file.root:
```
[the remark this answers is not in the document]
maxnoe: This should never happen and it will break other things (e.g. the reading in HDF5EventSource)

src/ctapipe/containers.py
```diff
@@ -262,6 +272,25 @@ class TelEventIndexContainer(Container):
     tel_id = tel_id_field()
 
 
+class TelescopeTriggerContainer(Container):
+    default_prefix = ""
```
Tobychev, line 276: No docstring here?

src/ctapipe/containers.py
```diff
@@ -262,6 +272,25 @@ class TelEventIndexContainer(Container):
...
+    time = Field(NAN_TIME, description="Telescope trigger time")
+    n_trigger_pixels = Field(
+        -1, description="Number of trigger groups (sectors) listed"
+    )
+    trigger_pixels = Field(None, description="pixels involved in the camera trigger")
+    event_type = Field(EventType.UNKNOWN, description="Event type")
+
+
+class SubarrayTriggerContainer(Container):
+    default_prefix = ""
```
Tobychev, line 286: No docstring here?

src/ctapipe/containers.py
```diff
@@ -773,12 +775,11 @@ class SimulatedCameraContainer(Container):
     )
 
 
-class SimulatedEventContainer(Container):
+class SimulationSubarrayContainer(Container):
```
Tobychev, line 778: No docstring here?

src/ctapipe/containers.py
```diff
@@ -1081,18 +1048,14 @@ class TelescopePointingContainer(Container):
...
-class PointingContainer(Container):
-    tel = Field(
-        default_factory=partial(Map, TelescopePointingContainer),
-        description="Telescope pointing positions",
-    )
-    array_azimuth = Field(nan * u.rad, "Array pointing azimuth", unit=u.rad)
-    array_altitude = Field(nan * u.rad, "Array pointing altitude", unit=u.rad)
-    array_ra = Field(nan * u.rad, "Array pointing right ascension", unit=u.rad)
-    array_dec = Field(nan * u.rad, "Array pointing declination", unit=u.rad)
+class SubarrayPointingContainer(Container):
```
Tobychev, line 1051: No docstring here?

src/ctapipe/containers.py
```diff
@@ -773,12 +775,11 @@ class SimulatedCameraContainer(Container):
     )
 
 
-class SimulatedEventContainer(Container):
+class SimulationSubarrayContainer(Container):
     shower = Field(
         default_factory=SimulatedShowerContainer,
         description="True event information",
     )
-    tel = Field(default_factory=partial(Map, SimulatedCameraContainer))
```
Tobychev, line 781: With the tel removed this container now only contains one thing, so its not clear what the extra indirection step achieves.

src/ctapipe/containers.py
Tobychev: There is also pointing information in the `tel` field below, although it is per telescope rather than for the subarray. 

Putting the subarray pointing info into the dl0 now seems strange, are there any quantities that depend on the subarray pointing at dl0? Isn't the telescope pointing the natural pointing until you do stereo?

maxnoe:
@kosack @mexanick This is now rebased on current main (not fun!) and locally tests are passing again, but I see a lot of warnings I didn't expect, so I'll continue tomorrow.

In any case, I would be very helpful if you could already start reviewing the changes in the container definitions [`src/ctapipe/containers.py`](https://github.com/cta-observatory/ctapipe/pull/2204/changes#diff-60b9243a1ce3cea2ac93c95f087222d3bf98571ad6afcfc56d4515f75d5b1c22)

and maybe play around a bit interactively with the new structure.

maxnoe:
There is one change in the file format here as well, not just to the in-memory containers: to mirror the move of trigger information to dl0, this also addresses https://github.com/cta-observatory/ctapipe/issues/2374 (and #2952) which better conforms with the CTAO data model.

This however should be implemented in a backwards-compatible way (reading old files with these containers stored in `/dl1` is still supported).

docs/changes/2204.api.rst
```diff
@@ -0,0 +1,23 @@
...
+Whereas now, the subarray event has a collection of telescope events,
+which then contain the different data levels:
+
+.. code:: python
+
+   event.tel[tel_id].dl1.parameters.hillas
+
+Fore more details, see :ref:`cep-002`.
+
+With this, also the container structure itself has been largely refactored.
```
maxnoe, line 23: Add migration guide details here.

src/ctapipe/io/hdf5dataformat.py
```diff
@@ -89,6 +89,8 @@
 R1_GROUP = "/r1"
 R1_TEL_GROUP = "/r1/event/telescope"
 DL0_GROUP = "/dl0"
+DL0_SUBARRAY_TRIGGER_TABLE = "/dl0/event/subarray/trigger"
+DL0_TEL_TRIGGER_TABLE = "/dl0/event/telescope/trigger"
 DL1_GROUP = "/dl1"
 DL1_SUBARRAY_GROUP = "/dl1/event/subarray"
 DL1_SUBARRAY_TRIGGER_TABLE = "/dl1/event/subarray/trigger"
```
mexanick, line 96:
```suggestion
```

src/ctapipe/io/hdf5dataformat.py
mexanick, line 98:
```suggestion
```

mexanick (commented):
I think the DL1 trigger table paths should be removed as that tables do not exist anymore

maxnoe:
> I think the DL1 trigger table paths should be removed as that tables do not exist anymore

These are crucial for backwards compatibility with older data and are used in e.g. the `HDF5EventSource` and `TableLoader`. 

We do not remove compatibility with older data formats in this pull request.

kosack (commented):
**EDIT**: never mind, I was looking at `RandomForestRegressor_energy` and not `RandomForestRegressor_tel_energy`

~I think there is an iteration bug:  I loaded an events file with TableLoader and with EventSource, and compared the first event's energy and get different results for the telescope-wise energy:~

kosack (dismissed):
Looks very good! i don't see any major issues, and anything else can be cleaned up in future PRs.   I guess you still want to add a migration guide, but I can re-approve once that is done.

kosack:
Note that I opened #2997 to fix a long-standing bug, which made exploring the new structure a bit less painful interactively.

maxnoe:
FYI: here is a PR in the `ctapipe_io_zfits` plugin working with this branch: https://github.com/cta-observatory/ctapipe_io_zfits/pull/22/changes

kosack (approved):

PR #2207, opened 2023-01-11
Change actions for mamba / pip install methods.

nbiederbeck:
for mamba: use mamba-org/provision-with-micromamba
for pip: use actions/setup-python


StFroese:
Should we also update the installation instruction in the README and use mamba instead of conda?

kosack:
Eventually, mamba won't be needed explicitly as the mamba solver will be included in a future version of conda (I think it's experimentally there already), but for now I think it's fine to just recommend mamba, as we know that installing packages from conda-forge with conda can be a slow nightmare due.

maxnoe:
I think we should always prefer recommending the open-source builds by conda-forge / mamba over the closed source (binary) blob packages by Anaconda Inc.

maxnoe (dismissed):

maxnoe (dismissed):

LukasNickel:
There has been a new matplotlib release yesterday (3.6.3), that the pip install uses, but mamba does not.
It is not immediately obvious to me however what change in matplotlib would cause the tests to fail now.
https://matplotlib.org/stable/users/github_stats.html

maxnoe (approved):

RuneDominik (approved):

StFroese (approved):

Issue #2087, opened 2022-09-28
Update installation instructions in README and Docs


maxnoe:
**Please describe the use case that requires this feature.**

As discussed here #2086, we should simplify installation instructions in the README and move the more complex setups into the documentation

maxnoe:
Done in #2208

PR #2208, opened 2023-01-12
Update README

StFroese:
- replaced anaconda install badge, since the link was no longer available
- use mamba instead of conda

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -7,7 +7,7 @@ ctapipe |ci| |codacy| |coverage| |conda| |doilatest|
     :alt: Test Status
 .. |codacy|  image:: https://api.codacy.com/project/badge/Grade/6192b471956b4cc684130c80c8214115
   :target: https://www.codacy.com/gh/cta-observatory/ctapipe?utm_source=github.com&amp;utm_medium=referral&amp;utm_content=cta-observatory/ctapipe&amp;utm_campaign=Badge_Grade
-.. |conda| image:: https://anaconda.org/conda-forge/ctapipe/badges/installer/conda.svg
+.. |conda| image:: https://anaconda.org/conda-forge/ctapipe/badges/version.svg
   :target: https://anaconda.org/conda-forge/ctapipe
 .. |coverage| image:: https://codecov.io/gh/cta-observatory/ctapipe/branch/master/graph/badge.svg
   :target: https://codecov.io/gh/cta-observatory/ctapipe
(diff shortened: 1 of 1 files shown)


maxnoe:
See also #2087

kosack (approved):

LukasNickel (dismissed):

StFroese:
closes #2087

docs/getting_started_users/index.rst
```diff
@@ -0,0 +1,48 @@
...
+How to get a specific version
++++++++++++++++++++++++++++++
+
+To install a specific version of ``ctapipe`` you can use the following commands::
+
+  CTAPIPE_VER=0.17.0
+  wget https://raw.githubusercontent.com/cta-observatory/ctapipe/v$CTAPIPE_VER/environment.yml
+  mamba env create -n cta -f environment.yml
+  mamba activate cta
+  mamba install -c conda-forge ctapipe=$CTAPIPE_VER
```
nbiederbeck, line 44: This is not really installing a specific version. That would be
```
pip install ctapipe==0.17.0  # or
mamba install ctapipe=0.17.0
```

This is setting up the development environment (although with a different name).

I think this part should move to getting started for devs (excluding changing the name).

docs/getting_started_users/index.rst
```diff
@@ -0,0 +1,48 @@
...
+
+**Note**: this environment contains many useful packages that are not strictly requirements of ctapipe.
+To get only ctapipe and its direct dependencies, just do ``mamba install -c conda-forge ctapipe[=<version>]`` in an environment
```
nbiederbeck, line 47: Ha I see now that this is what I said in the other comment.  For users I think I would like the other way around:
1. installing latest
2. installing a specific version

For users and developers alike it is strongly advised to create virtual (conda/mamba) environments, but this section of the docs mixes environments and versions in a confusing way.

nbiederbeck (changes requested):

docs/getting_started/index.rst
```diff
@@ -146,6 +146,18 @@ to your local working copy):
...
+--------------------------------------------
+Creating an environment for specific version
+--------------------------------------------
+
+A virtual environment for a specific version using the release on conda-forge can be created with::
+
+  CTAPIPE_VER=0.17.0
+  wget https://raw.githubusercontent.com/cta-observatory/ctapipe/v$CTAPIPE_VER/environment.yml
+  mamba env create -n cta-$CTAPIPE_VER -f environment.yml
+  mamba activate cta-0.17.0
```
nbiederbeck, line 158:
```suggestion
  mamba activate cta-$CTAPIPE_VER
```

nbiederbeck (changes requested):

docs/getting_started/index.rst
maxnoe: I'd remove this option completely. The `environment.yml` is for development (and then you just have it locally) and the CI (which is something we maybe don't even want in on env, since it e.g. doesn't make sense to install vitables in the ci).

People should install ctapipe directly, or add it to their env files. Not use the one from this repo.

nbiederbeck (approved):

kosack (approved):

Issue #1777, opened 2021-08-24
Pedestal subration broken for sim_telarray DST0 files


ParsonsRD:
Hi all,
The reading of the DST0 files (essentially dl1 data level) produced by Konrad's sim_telarray is currently broken. When reading these files the pedestal subtraction is currently not performed correctly. This seems to be due to the following code in simteleventsource.py:

`                pedestal = cam_mon["pedestal"] / cam_mon["n_ped_slices"]
`
In this case  cam_mon["n_ped_slices"] will have a value of whatever the number of slices was in the original simulation (typically something large like 40), but in the DST0 the pedestal stored is that of the integrated signal, i.e. a number of slices = 1. So will result in a subtraction of a far too low value. As far as I know there is no flag in the data to confirm that you are reading a DST0 file so a fix is not obvious. One option could be to use the n_samples length of the waveform from 

`                n_gains, n_pixels, n_samples = adc_samples.shape
`

As this should always be the same as the pedestal samples for a sim_telarray output file, but one for a DST0 file. But there may be safer alternatives out there.

kosack:
Do you have an example DST0 file we could add to the test suite? 

Right now `SimTelEventSource` only supports raw simtel files,  but it should be easy to implement handling the DST format ones, either in SimTelEventSource itself, or a derived version.  At the very least it should fail with an error if a DST0 file is used, so our current `is_compatible()` function is not sufficient.

ParsonsRD:
Sure attached is my 10 event example file that I use for the template generation test functions. I found this issue when updating the template generation to v0.11.0.

[gamma_HESS_example.simhess.gz](https://github.com/cta-observatory/ctapipe/files/7039448/gamma_HESS_example.simhess.gz)

maxnoe:
Sorry for only coming back to this now. The file you provided only contains `ADCSums`, not `ADCSamples`:


In the case where now adc samples are found in the `SimTelEventSource`, `adc_samples` is filled from `adc_sums`, upshaed to a 3d array.

From todays point of view, this is certainly wrong and should not be done, instead we should directly fill the DL1 images using the ADCSums, right?

https://github.com/cta-observatory/ctapipe/blob/6217b8517e02d5f811f0a1f65ca25cf1e3b079e6/ctapipe/io/simteleventsource.py#L762-L765

PR #2209, opened 2023-01-12
Start working on reading adc_sums

maxnoe:
These seem to be raw adc sums that need to be calibrated, we need to figure out how to correctly convert to pe.

They also have both gains, so we also somehow need to perform a gain selection.

Maybe @ParsonsRD knows? If not I'll contact konrad

### DIFF
--- a/ctapipe/io/simteleventsource.py
+++ b/ctapipe/io/simteleventsource.py
@@ -828,7 +828,8 @@ class SimTelEventSource(EventSource):
 
                 adc_sums = telescope_event.get("adc_sums")
                 if adc_sums is not None:
-                    data.dl1.tel[tel_id].image = adc_sums
+                    image_hg = (adc_sums[0] - cam_mon["pedestal"][0]) * dc_to_pe[0]
+                    data.dl1.tel[tel_id].image = image_hg
 
             yield data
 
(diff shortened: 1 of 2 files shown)


maxnoe:
This is intended  to fix #1777 

If we don't want this, we should close #1777 as "won't fix"

kosack:
> This is intended to fix https://github.com/cta-observatory/ctapipe/issues/1777

Ah, right, in the raw data, this has no useful meaning (it's just the full sum of the waveform), however, for DST-style output (which is closer to DL1), it's the DL1 image.   So if we want to use it, we have to do the same calibration as for the simulated waveform, but using the integrated  pedestal.  But in that case the EventSource should then also report that it outputs DL1_IMAGE information, and not DL0 since the waveforms are no longer there.  

Is it worth supporting that? Not sure - do we have a lot of Simtel DST data where we don't already have the raw data?

maxnoe:
> Is it worth supporting that? Not sure - do we have a lot of Simtel DST data where we don't already have the raw data?

Have? No. I don't think so. But I think for data storage reasons, @ParsonsRD uses it in the ImPACT training workflow.

maxnoe:
Closing for now. Will reopen if need arises and we have a solution.

PR #2217, opened 2023-01-13
Add option to configure ml prefixes

LukasBeiske:
This adds the option to configure the prefixes of the ml models, while still using the model class as the default.
Related: https://github.com/cta-observatory/ctapipe/pull/2138#discussion_r1032774404

### DIFF
--- a/ctapipe/reco/sklearn.py
+++ b/ctapipe/reco/sklearn.py
@@ -83,6 +83,9 @@ class SKLearnReconstructor(Reconstructor):
     #: Name of the target column in training table
     target: str = None
 
+    prefix = Unicode(
+        default_value=None, allow_none=True, help="Prefix for output of this model"
+    ).tag(config=True)
     features = List(Unicode(), help="Features to use for this model").tag(config=True)
     model_config = Dict({}, help="kwargs for the sklearn model").tag(config=True)
     model_cls = Enum(SUPPORTED_MODELS.keys(), default_value=None, allow_none=True).tag(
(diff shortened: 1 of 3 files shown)


ctapipe/reco/sklearn.py
```diff
@@ -83,6 +83,9 @@ class SKLearnReconstructor(Reconstructor):
     #: Name of the target column in training table
     target: str = None
 
+    prefix = Unicode(
+        default_value=None, allow_none=True, help="Prefix for output of this model"
```
maxnoe, line 87: Add that if None, the configured model_cls name will be used

ctapipe/reco/sklearn.py
```diff
@@ -117,6 +120,10 @@ def __init__(self, subarray=None, models=None, **kwargs):
                     "__init__() missing 1 required positional argument: 'subarray'"
                 )
 
+            if self.prefix is None:
+                # Default prefix is model_cls
+                self.prefix = self.model_cls
```
kosack, line 125: You should either mention that the default is the model class in the help string above, or you might even be able to use a `@default` decorator to link them

kosack (changes requested):

ctapipe/reco/sklearn.py
LukasBeiske: I am not sure how to use the decorator for this. I would have to make a separate function which sets the prefix, right? I think mentioning the default in the help string is the easier solution in that case.

kosack (approved):

ctapipe/reco/sklearn.py
StFroese: Is it possible to make default_value=model_cls?

maxnoe (approved):

PR #2220, opened 2023-01-16
Make devs aware of debugging and `ipdb` in docs

StFroese:


StFroese:
discussed in #2149

kosack:
It would be good to also show how to run a command-line tool in a debugger with and without having to add a `breakpoint()` line.    I think if you add a `breakpoint()` with an appropriate PYTHONDEBUG set, running e.g. `ctapipe-process` should work fine.    But sometimes you don't want to edit the code, and instead can run the tool directly in debug mode like this:

```sh
python -m pdb -m ctapipe.tools.process [usual command-line options]
```

nbiederbeck:
I would also add `pytest --pdb` in this section

Tobychev:
I recently found [pudb]( https://documen.tician.de/pudb/index.html) which is a pretty nice if somewhat quirky terminal debugger.

kosack:
It seem the jobs here are failing, and there is no "restart" button - perhaps push a blank commit and we can just review/merge this?

maxnoe:
This needs a rebase with main, to e.g. include the fixes for the memory problem (which is I think the issue for the CI failure here).

The reason for the absence of a restart button is probably that github doesn't allow restarting jobs indefinitely.

maxnoe:
Rebased vs. main now

kosack (approved):

Tobychev (approved):

PR #2221, opened 2023-01-16
Switch to jupytext MyST markdown

### LINKED ISSUE #2099: Use jupytext for notebooks in the documentation?

Looking at diffs and quickly updating notebooks in the docs currently requires external tools like *nbdime* or *Review Notebook App* due to the fact that the `*.ipynb`  format is JSON-based.  JSON is also not very GIT-friendly.

jupytext (https://jupytext.readthedocs.io) solves this by using a line-based format for notebooks (basically a python script with special comments to store the metadata or a markdown file with similar). 

It would be nice to use this and migrate the current docs to this format.   

Disadvantage: It would add one more dependency on the jupytext package for documentation building, and would also require anyone using the notebooks outside of the docs to install this plugin.   Also, not sure that GitHub auto-renders jupytext docs, probably not, except if the MarkDown version is used  (but in any case viewing in GitHub may not be a big problem as the docs are rendered to readthedocs)

StFroese:
fixes #2099


maxnoe:
Can we use the same as gammapy here? I like that in that format, the notebooks are actually valid python scripts that can also be run normally.

maxnoe:
Out of the box? Or with some extension installed?

kosack:
fixes #2099

kosack:
> Can we use the same as gammapy here? I like that in that format, the notebooks are actually valid python scripts that can also be run normally.

Does jupyter itself support the format GammaPy uses, or do they export to that after editing? With Jupytext, you can use jupyter directly to edit/view the files I think, which is nice.

maxnoe:
The system gammapy uses is sphinx-gallery: https://sphinx-gallery.github.io/

kosack:
> The system gammapy uses is sphinx-gallery: https://sphinx-gallery.github.io/

That is nice as it lets you download the result as a notebook or a script...

maxnoe:
Loading the script directly with jupytext also seems to work in that format

maxnoe:
See https://jupytext.readthedocs.io/en/latest/formats.html#sphinx-gallery-scripts

.github/workflows/ci.yml
```diff
@@ -130,7 +130,7 @@ jobs:
       - name: Install doc dependencies
         run: |
           sudo apt update --yes && sudo apt install --yes git build-essential pandoc graphviz
-          pip install -U pip towncrier
+          pip install -U pip towncrier myst-nb
```
nbiederbeck, line 133: should we not include those now in `setup.cfg` under `options.extras_require.docs`?

maxnoe:
@StFroese Could you make a new attempt here with the same format Gammapy uses and the current main?

Please add the newly needed packages to the correct extras in `setup.cfg` and to the conda development environment in `environment.yml`.

StFroese:
@maxnoe can do. I'll open a new PR for this

kosack:
> @StFroese Could you make a new attempt here with the same format Gammapy uses and the current main?

I have a quick question, not having worked with the sphinx-gallery text format.  Can you open notebooks written in this format directly in Jupyter(lab)? Or do you have to convert to notebook first, edit, and convert back?  Or is it just a different variant of format supported by the jupytext plugin?

maxnoe:
@kosack When you have the jupytext extension installed, they open just fine in jupyter {lab,notebook}

StFroese:
> Can you open notebooks written in this format directly in Jupyter(lab)? Or do you have to convert to notebook first, edit, and convert back? Or is it just a different variant of format supported by the jupytext plugin?

I'm a little confused too about this. Seems like you have to use .py script with sphinx-gallery instead of notebooks. Am I right @maxnoe?
This seems to be a kind of alternative maybe https://sphinx-nbexamples.readthedocs.io/en/latest/

kosack:
Replaced by #2377

PR #2222, opened 2023-01-17
Fix numba version to 0.56

maxnoe:
Creating the fresh env locally, mamba thought it could install numpy 1.24 with numba 0.53, which are incompatible, but that is not specified on the numba 0.53 package on conda-forge.

So here we explictly add the most recent numba version, which results in mamba correctly resolving it to the numpy 1.23 + numba 0.56 pair

### DIFF
--- a/environment.yml
+++ b/environment.yml
@@ -19,7 +19,7 @@ dependencies:
   - joblib
   - jupyter
   - matplotlib
-  - numba
+  - numba=0.56
   - numpy>=1.17
   - numpydoc
   - pandas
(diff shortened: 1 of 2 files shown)


kosack:
is there an issue opened for the numba package spec on conda-forge? Seems like they should fix that

maxnoe:
Yes, several... I guess they didn't go as far back as 0.53 though. We should just require a recent version

kosack (approved):

nbiederbeck (approved):

PR #2229, opened 2023-01-20
Implement part of common reconstructor API

maxnoe:
* Move `.read` to `Reconstructor`
* Introduce `ReconstructionProperty` enum
* Let `predict_telescope_table` return a `dict[Property, Table]`
* Unify cli arguments of `ctapipe-process` and `ctapipe-apply-models` to have `--reconstructor` for each `Reconstructor` subclass to apply.

* Remove the aliases for the reconstructor load paths in `ctapipe-process` and the hacky workaround to make it not necessary to specify reconstructor types for them in the config.

Associated issues: #2116

### DIFF
--- a/ctapipe/reco/reconstructor.py
+++ b/ctapipe/reco/reconstructor.py
@@ -91,7 +91,27 @@ class Reconstructor(TelescopeComponent):
         """
 
     @classmethod
-    def read(cls, path, **kwargs):
+    def read(cls, path, parent=None, subarray=None, **kwargs):
+        """Read a joblib-pickled reconstructor from ``path``
+
+        Parameters
+        ----------
+        path : str or pathlib.Path
+            Path to a Reconstructor instance pickled using joblib
+        parent : None or Component or Tool
+            Attach a new parent to the loaded class, this will properly
+        subarray : SubarrayDescription
+            Attach a new subarray to the loaded reconstructor
+            A warning will be raised if the telescope types of the
+            subarray stored in the pickled class do not match with the
+            provided subarray.
+
+        **kwargs are set on the loaded instance
+
+        Returns
+        -------
+        Reconstructor instance loaded from file
+        """
         with open(path, "rb") as f:
             instance = joblib.load(f)
 
@@ -101,11 +121,11 @@ class Reconstructor(TelescopeComponent):
             )
 
         # first deal with kwargs that would need "special" treatmet, parent and subarray
-        if parent := kwargs.pop("parent", None):
+        if parent is not None:
             instance.parent = weakref.proxy(parent)
             instance.log = parent.log.getChild(instance.__class__.__name__)
 
-        if subarray := kwargs.pop("subarray", None):
+        if subarray is not None:
             if instance.subarray.telescope_types != subarray.telescope_types:
                 instance.log.warning(
                     "Supplied subarray has different telescopes than subarray loaded from file"
--- a/ctapipe/reco/reconstructor.py
+++ b/ctapipe/reco/reconstructor.py
@@ -95,15 +95,25 @@ class Reconstructor(TelescopeComponent):
         with open(path, "rb") as f:
             instance = joblib.load(f)
 
-        for attr, value in kwargs.items():
-            if attr == "parent":
-                value = weakref.proxy(value)
-            setattr(instance, attr, value)
-
         if not isinstance(instance, cls):
             raise TypeError(
                 f"{path} did not contain an instance of {cls}, got {instance}"
             )
+
+        # first deal with kwargs that would need "special" treatmet, parent and subarray
+        if parent := kwargs.pop("parent", None):
+            instance.parent = weakref.proxy(parent)
+            instance.log = parent.log.getChild(instance.__class__.__name__)
+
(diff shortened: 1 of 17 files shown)


ctapipe/tools/process.py
```diff
@@ -74,9 +74,7 @@ class ProcessorTool(Tool):
         ("o", "output"): "DataWriter.output_path",
         ("t", "allowed-tels"): "EventSource.allowed_tels",
         ("m", "max-events"): "EventSource.max_events",
-        "energy-regressor": "ShowerProcessor.EnergyRegressor.load_path",
-        "particle-classifier": "ShowerProcessor.ParticleClassifier.load_path",
-        "disp-reconstructor": "ShowerProcessor.DispReconstructor.load_path",
+        "reconstructor": "ShowerProcessor.reconstructor_paths",
```
kosack, line 77: Have you considered how will this work with non-ML reconstructors or those that  don't have input files/serialized versions?
* ImPACT (not  a ML reco, but it does have a model as input, so we could just implement `read()` and it should work)
* Model3D (no input path needed - how would you specify it to be run?)

maxnoe: Yes, those are still configuratble via the name list `reconstructor_types` as before. We just *additionally* have the list of paths.

kosack: Thanks, I missed that..  One issue is that reconstructors listed in `reconstuctor_types` are always applied before those listed in `reconstructor_paths`, which then limits the user choice in order (i.e. Model3D, which needs a preliminary geometry as a seed can only get that seed from a reconstructor listed in the types list).   It's a little bit arbitrary.  Though I had trouble coming up with something better. Originall I thought to just have the  input file come in the config for the specific reconstructor itself:

```yaml
ShowerProcessor:
    reconstuctors: 
        - HillasReconstructor
        - EnergyRegressor
        - HillasIntersection
        - ImPACTReconstructor
        - Model3DReconstructor

ImPACTReconstructor:
    model_path: "impact-model.h5"
    geometry_seed: HillasReconstructor
    energy_seed: EnergyRegressor

EnergyRegressor:
    model_path: "energy-regressor-v1.h5"
...
```

Though in that case, you can't have multiply `EnergyRegressors` trained on different parameters, which could be a nice feature (e.g. an EnergyRegressor that uses the Hillas geometry, and another that uses the Model3D geometry. 

So I guess for now, having these two separate (but somewhat equal) lists are ok.

maxnoe: Yes, that was the other option, however that required quite "hacky" or complex code in both the `__init__` methods of such reconstructors that support reading from a pickled (or another) path and to enable the aliases in the tools.

We could leave out  aliases, that would mean having something like `--reconstructor-types EnergyRegressor --EnergyRegressor.load_path=path/to/energy.pkl` or doing it via config files. 

I thought the two lists was the better solution for now. Not by much though, I could also re-add the complex load coad in the `__init__` method.

maxnoe:
@kosack @LukasNickel could you review? Would be great to have this to start the second part.

LukasNickel (approved):

ctapipe/tools/process.py
kosack: > We could leave out aliases, that would mean having something like --reconstructor-types EnergyRegressor --EnergyRegressor.load_path=path/to/energy.pkl or doing it via config files.

That actually is cleaner I think.  I can't see that we will be using the command-line to change the reconstruct list very often so it doesn't have to be pretty — it would really be something you configure once via a config file.

maxnoe: I tend to agree, so I will probably revert the adding of the path here and add back to `load_path` magic to the reconstructors that need it. Will convert this  back to draft state

maxnoe: I changed that part back and adapted the CLI of `process` accordingly. It is more flexible but not as nice as before. But I guess we will mostly use "apply-models" or config files, so maybe consistency is more important than nice cli options.

LukasNickel (dismissed):

maxnoe:
@kosack could you review here?

kosack (changes requested):
I would expect `HillasReconstructor` and `HillasIntersection` to have a `property` attribute that tells me they reconstruct GEOMETRY, but they seems to be missing.  Is there a reason for that?

They both inherit the `Reconstructor.read()` method as well, which might be best to override as a no-op in those classes, or simply return self

ctapipe/reco/reconstructor.py
```diff
@@ -66,6 +90,34 @@ def __call__(self, event: ArrayEventContainer):
             reconstructed stereo geometry and telescope-wise impact position.
         """
 
+    @classmethod
+    def read(cls, path, **kwargs):
```
kosack, line 94: add a docstring here to explain what this does aand what is expected from `path`.   It is  not obvious that path should be an instance of this class pickled with joblib for example.

maxnoe: done

maxnoe:
@kosack not sure about the :+1:  : do you want me to add the `properties` back? Or is it fine as it is?

kosack (approved):

LukasNickel (approved):

PR #2234, opened 2023-01-25
make qualityquery a telescopecomponent

### LINKED ISSUE #2158: Make QualityQuery a TelescopeComponent and list of checks a TelescopeParameter

To support different cuts per telescope type.

Should be straight forward.

nbiederbeck:
Fixes #2158. Still a draft.

This is currently ugly, as we now `yield from` a list with single value. This would need a change to ExpressionEngine, which should be fine also in this PR. Need to check the calls in FeatureGenerator first.

Tests fail because I need to attach the `Subarray` everywhere


ctapipe/core/qualityquery.py
```diff
@@ -93,7 +98,7 @@ def __str__(self):
         """Print a formatted string representation of the entire table."""
         return str(self.to_table())
 
-    def __call__(self, **kwargs) -> np.ndarray:
+    def __call__(self, tel_id, **kwargs) -> np.ndarray:
```
maxnoe, line 101: Since the quality query supports being called with the telescope type, I think we should also support this here. E.g. do not use `tel_id` but `key: [int, str, TelescopeDescription]`?

ctapipe/core/qualityquery.py
```diff
@@ -132,11 +138,13 @@ def get_table_mask(self, table):
...
             Boolean mask of valid entries.
         """
-        n_criteria = len(self.quality_criteria) + 1
+        n_criteria = len(self.criteria_names) + 1
         result = np.ones((n_criteria, len(table)), dtype=bool)
 
-        for i, res in enumerate(self.engine(table), start=1):
-            result[i] = res
+        for i, (_, engine) in enumerate(self.engines, start=1):
+            for tel_id in np.unique(table["tel_id"]):
```
maxnoe, line 145: Is this faster than group_by and then resorting to the original order? It seems like the more costly operation...

ctapipe/core/qualityquery.py
```diff
@@ -25,38 +31,45 @@ class QualityQuery(Component):
...
+            self.engines.append((name, lookup))
 
         # arrays for recording overall statistics, add one for total count
-        n = len(self.quality_criteria) + 1
+        n = len(self.criteria_names) + 1
         self._counts = np.zeros(n, dtype=np.int64)
         self._cumulative_counts = np.zeros(n, dtype=np.int64)
 
-    def to_table(self, functions=False):
+    def to_table(self, functions=False) -> Table:
```
maxnoe, line 72: kwarg now unused?

ctapipe/core/tests/test_qualityquery.py
```diff
@@ -8,33 +8,33 @@
...
 
-def test_selector():
+def test_selector(subarray_prod5_paranal):
     """test the functionality of an example Selector subclass"""
 
     class ExampleQualityQuery(QualityQuery):
         """Available variables: x"""
 
         quality_criteria = List(
-            default_value=[
```
maxnoe, line 18: I'd leave this at it is (it should still be supported since we treat single values as `type *` and just add a new test with the full telescope-parameters version.

ctapipe/core/tests/test_qualityquery.py
```diff
@@ -8,33 +8,33 @@
...
+                ("a_value_not_too_high", [("type", "*", "x < 100")]),
+                ("smallish", [("type", "*", "x < np.sqrt(100)")]),
             ],
         ).tag(config=True)
 
-    query = ExampleQualityQuery()
+    query = ExampleQualityQuery(subarray=subarray_prod5_paranal)
 
-    criteria1 = query(x=0)  # pass smallish
+    criteria1 = query(tel_id=1, x=0)  # pass smallish
```
maxnoe, line 27: if you set the `key` to default value None, this test won't need to be adapted at all.

ctapipe/core/qualityquery.py
maxnoe: And if you use `key=None`, the API is backwards compatible,i.e. `__call__(...)` without a key will take the global default value, which is what the current configurations will be parsed into.

ctapipe/core/tests/test_qualityquery.py
```diff
@@ -126,22 +140,87 @@ def test_table_mask():
...
+    query = QualityQuery(
+        quality_criteria=[
+            ("foo", [("type", "*", "x > 0")]),
+            ("bar", [("type", "*", "x > 1"), ("id", "1", "x > 2")]),
+            ("baz", "x > 3"),
+        ],
+        subarray=subarray_prod5_paranal,
+    )
+
+    query(tel_id=1, x=0)
```
maxnoe, line 184: check the result

ctapipe/core/tests/test_qualityquery.py
```diff
@@ -126,22 +140,87 @@ def test_table_mask():
...
+    query = QualityQuery(
+        quality_criteria=[
+            ("foo", [("type", "*", "x > 0"), ("id", "2", "x < 0")]),
+        ],
+        subarray=subarray_prod5_paranal,
+    )
+
+    x = 1
+    assert query(x=x, tel_id=1)
+    assert not query(x=x, tel_id=2)
```
maxnoe, line 209: Shouldn't this return a tuple? I.e. wouldn't you need to test the single elements?

docs/changes/2234.feature.rst
```diff
@@ -0,0 +1,5 @@
+``QualityQuery`` is now a ``TelescopeComponent``.
+
+You can now give different quality queries per telescope type or id.
+Change is backwards compatible, older configurations still work and
+set the query-expression to the global value of the TelescopeComponent.
```
maxnoe, line 5: But as of now the API of QualityQuery is changed in an incompatible way. If you include the `key=None`, it will be truly backwards compatible.

ctapipe/core/qualityquery.py
nbiederbeck: I didn't do it at first as astropy table group_by does not give you the corresponding tel_id, as pandas does it. I can check

ctapipe/core/qualityquery.py
nbiederbeck: The processtool uses it: https://github.com/cta-observatory/ctapipe/blob/c1f7ba0994e485d983d5fd861d55563682c0c591/ctapipe/tools/process.py#L283

We need to decide whether (and how) to support this.  I left this here as unused kwarg to not change to ProcessTool

maxnoe: I meant inside the function body.

So I'd say that this info is available from the configuration, which is stored in provenance. Alternatively, it could be stored as metadata on the table, not in the table itself?

nbiederbeck: I think config and provenance is enough

ctapipe/core/tests/test_qualityquery.py
nbiederbeck: array with single item: `foo`

ctapipe/core/qualityquery.py
maxnoe: See https://docs.astropy.org/en/stable/table/operations.html#manipulating-groups

```
for key, group in zip(obs_by_name.groups.keys, obs_by_name.groups):
```

nbiederbeck: changed it to use group_by and not rely on masks

ctapipe/core/qualityquery.py
```diff
@@ -147,10 +147,14 @@ def get_table_mask(self, table):
         n_criteria = len(self.criteria_names) + 1
         result = np.ones((n_criteria, len(table)), dtype=bool)
 
+        table["__index__"] = range(len(table))
+
+        by_id = table.group_by("tel_id")
```
maxnoe, line 152: table.group_by makes a full copy of the input table. As provided in the astropy slack by tomaldcroft, the best solution seems create a new table only with the tel_ids and the index to use that for group_by:

```
import numpy as np

table = Table({
    "event_id": [1, 2, 3, 4, 5, 6, 7, 8],
    "tel_id": [1, 1, 2, 1, 2, 2, 1, 3],
    "value": np.random.normal(size=8),
})


def apply_mask(table):
    cuts = {1: 0.0, 2: 1.0, 3: 2.0}
    index_table = Table({"tel_id": table["tel_id"], "index": np.arange(len(table))})
    grouped = index_table.group_by("tel_id")
    del index_table

    table["mask"] = False

    for key, group_index in zip(grouped.groups.keys, grouped.groups):
        tel_id = key["tel_id"]
        index = group_index["index"]
        table["mask"][index] = table["value"][index] >= cuts[tel_id]


apply_mask(table)
print(table)

```

ctapipe/reco/tests/test_sklearn.py
```diff
@@ -58,6 +58,7 @@ def example_table():
         n_samples=100, n_features=5, n_informative=3, random_state=0
     )
     t = Table({f"X{i}": col for i, col in enumerate(X.T)})
+    t["tel_id"] = np.arange(len(t))
```
nbiederbeck, line 61: Maybe it is better to not groupby tel id if it is not there instead of relying on it being attached? Still would have the default value of the TelescopeParameter

ctapipe/core/qualityquery.py
```diff
@@ -132,11 +144,18 @@ def get_table_mask(self, table):
         mask : np.ndarray[bool]
             Boolean mask of valid entries.
         """
-        n_criteria = len(self.quality_criteria) + 1
+        n_criteria = len(self.criteria_names) + 1
         result = np.ones((n_criteria, len(table)), dtype=bool)
 
-        for i, res in enumerate(self.engine(table), start=1):
-            result[i] = res
+        index_table = Table({"tel_id": table["tel_id"], "index": np.arange(len(table))})
```
maxnoe, line 150: MMh, not sure we should always assume `tel_id` here... 

Couldn't we add also here the `key` argument? So e.g. `get_table_mask(table, key="tel_id")` would use the telescope-wise value but `get_table_mask(table, key=None)` would use the global default (which might be the only value specified

nbiederbeck: Added the `key` here (which is the column name of the thing to group-by) and use it in the sklearn reconstructors

ctapipe/core/qualityquery.py
```diff
@@ -25,38 +33,45 @@ class QualityQuery(Component):
...
+
+        self._compile()
+
+    def _compile(self):
+        self.engines = []
+        self.criteria_names = []
+
+        for name, criteria_list in self.quality_criteria:
+
+            if not isinstance(criteria_list, (tuple, list, TelescopePatternList)):
```
maxnoe, line 55: How can this happen? The traitlets system should make sure that this is always a `TelescopePatternList`

nbiederbeck: Yes, it should, but I can't seem to get it to work correctly.

nbiederbeck: fixed

ctapipe/core/qualityquery.py
```diff
@@ -93,12 +96,18 @@ def __str__(self):
         """Print a formatted string representation of the entire table."""
         return str(self.to_table())
 
-    def __call__(self, **kwargs) -> np.ndarray:
+    def __call__(
+        self,
+        key: Union[int, str, Literal[None]] = None,
```
maxnoe, line 101: TelescopeType also, right?

ctapipe/core/telescope_component.py
```diff
@@ -110,7 +110,7 @@ def single_to_pattern(value):
         # make sure we only change things that are not already a
         # pattern tuple
         if (
-            not isinstance(value, tuple)
+            not isinstance(value, (list, tuple))
```
maxnoe, line 113: This change makes me a bit nervous... The mechanics of deciding of something was already a default value or not where already pretty brittle before. What is the reason to  also check for a list here?

docs/changes/2234.feature.rst
maxnoe, line 5: Since `QualityQuery` now require the subarray, it's not fully API compatible. Only with respect to that existing configurations will still work the same for tools that were updated.

Which makes me wonder if we instead of making the `QualityQuery` a telescope component we should instead add a `TelescopeQualityQuery`... Do have have a use-case for a `QualityQuery` that does not need a subarray?

ctapipe/core/telescope_component.py
nbiederbeck: Is not needed.  Left-over from the typing problem

ctapipe/core/qualityquery.py
nbiederbeck: `TelescopeDescription`, right? → circular import ...

docs/changes/2234.feature.rst
nbiederbeck: > Since QualityQuery now require the subarray, it's not fully API compatible. Only with respect to that existing configurations will still work the same for tools that were updated.

Right, will change that.

> Do have have a use-case for a QualityQuery that does not need a subarray?

We have `ctapipe.reco.reconstructor.StereoQualityQuery`, which does not necessarily need a subarray I think

nbiederbeck: Setting `subarray=None` (which I now implemented as default) can be used as a ~`Telescope`~`Component`.  Giving a `key` on calling fails with an okayish error:

```
ValueError: TelescopeParameterLookup: No subarray attached, call `attach_subarray` first before trying to access a value by tel_id
```

ctapipe/core/qualityquery.py
maxnoe: mmh, okay...

maxnoe (approved):

kosack (changes requested):
The only thing I dislike here is that the list of functions is no longer stored in the output table of QualityQuery.  That means you lose an important piece of metadata.  Can you add that back in?  It could simply contain the string representation of the TelescopeParameter, which is ugly, but would at least keep the information

maxnoe:
@kosack we discussed this here offline and concluded that this information is stored in the configuration (and thus the provenance). But adding it back e.g. is metadata of that table would also be possible.

kosack:
> @kosack we discussed this here offline and concluded that this information is stored in the configuration (and thus the provenance). But adding it back e.g. is metadata of that table would also be possible.

I think it's useful to have it in the table as either metadata or better as a column since it's very useful information and is probably not something the user should have to look in provenance for. The short name often doesn't give enough information, often (e.g. the name might be "large enough signal", but knowing that the cut was exactly 50 PE is critical for understanding.  Also this allows QualityQueries to be read from a Table and re-constructed with no information loss.

kosack:
Thinking about it further, it could instead be a table per TelescopeParameter pattern instead of one per telescope, if you want to save some space

kosack:
The fact is the output table in its current form no longer sufficient if the cuts can change per telescope/type - it doesn't fully help to understand which cuts are too strict/loose, which is the point of keeping cut statistics.

maxnoe:
> The fact is the output table in its current form no longer sufficient if the cuts can change per telescope/type - it doesn't fully help to understand which cuts are too strict/loose, which is the point of keeping cut statistics.

Yes, that's true. However we would need to completely change it, since the current format just sums up all.


So you say we should actually have separate `QualityQuery` and `TelescopeQualityQuery` components and then also record per telescope how many events survived?

This sounds good, but it is much more work, so I'd reschedule to 0.19

kosack:
> So you say we should actually have separate QualityQuery and TelescopeQualityQuery components and then also record per telescope how many events survived?
> This sounds good, but it is much more work, so I'd reschedule to 0.19

It may be that you don't need both: I think just inside QualityQuery we can have tables per-telescope (now that you can specify a tel_id), and also one that combines all. Yes, it's a bit larger change, but not too complicated.   Yes, we can postpone this, or split it into two.

maxnoe:
I'll close this PR here as it is very far behind main. I think we still want this feature, but it's easier to start from current main cherry-picking than updating the branch here.

PR #2235, opened 2023-01-26
Check that subarrays from to-be-predicted file match that of all reconstructors (match training-subarray)

nbiederbeck:


ctapipe/tools/apply_models.py
```diff
@@ -154,6 +154,16 @@ def setup(self):
                 )
             )
 
+        self._check_subarrays()
+
+    def _check_subarrays(self):
+        for reconstructor in self._reconstructors:
+            if self.loader.subarray != reconstructor.subarray:
```
kosack, line 161: Do you really need to check that they are identical subarrays, or rather check that they contain the same telescope *types*? Since the models are per-telescope-type, the energy reconstruction as is will still work if you use a model trained for 2 LSTs and you give it an array of 4, for example .

And for real observed data, there will likely be cases where we have for example a subarray with a single telescope missing where you still want to reconstruct energy using the original model (for large arrays the difference would be small, even in the case of some sort of more complex stereo model)

So maybe:
```py
if set(self.loader.subarray.telescope_types) != set(recnostructor.subarray.telescope_types):
```

maxnoe: > Do you really need to check that they are identical subarrays, or rather check that they contain the same telescope types?

I'd say it depends on whether stereoscopic features (e.g. HillasReconstructor) have been used or not.

nbiederbeck: Sorry that I opened the PR before opening the issue, I will copy these comments there and let's discuss in the issue #2236.

maxnoe:
The `Reconstructor.read` method used by the apply-tool compares that the telescope types are identical (issuing a warning if not).

Do we need more checks?

PR #2241, opened 2023-02-02
Simplify ChunkIterator, return start, stop, table

maxnoe:
* Instead of the iterator protocol, it now uses the sequence protocol. This allows random access and simplifies the internal book-keeping.

* Instead of storing start, stop on the Iterator, they are returned together with the table(s).

This made the implementation of the apply tool reading the data only once much easier.

### DIFF
--- a/ctapipe/io/tableloader.py
+++ b/ctapipe/io/tableloader.py
@@ -3,7 +3,7 @@ Class and related functions to read DL1 (a,b) and/or DL2 (a) data
 from an HDF5 file produced with ctapipe-process.
 """
 import warnings
-from collections import defaultdict
+from collections import defaultdict, namedtuple
 from pathlib import Path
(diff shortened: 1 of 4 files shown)


LukasNickel (approved):

LukasNickel (approved):

kosack (commented):
Nice.  this is easier to use as well.

Perhaps add `collections.abc.Sequence` as a base class for ChunkedIterator just for developers to understand what kind of object this is.

maxnoe:
> Perhaps add collections.abc.Sequence as a base class for ChunkedIterator just for developers to understand what kind of object this is.

That would add a ton of mix-in methods like "in", "count", "index" etc. which I am not sure is useful.

kosack (approved):

PR #2249, opened 2023-02-08
Refactor apply tool to use HDF5Merger, read data only once

### LINKED ISSUE #2130: Allow dropping dl1 and telescope-wise information in `apply-models` or add tool for that.

### LINKED ISSUE #2248: FeatureGenerator modifies and returns input table
Currently, the `FeatureGenerator` modifies and returns the input table.

This is a counter-intuitive API.
We should either:

* have it mutate the input and not return
* Create a new table and return that

Part of why I thought we need #2240 was that the reconstructors modify the input table, which should probably not be the case.

A couple of solutions I came up with:

* functional approach: return new table with added columns
* mutating approach: remove the return, allow the override

Both have significant drawbacks: I think it is never a good idea to mutate input in that way (reconstructors.predict_table should produce a prediction, not have side effects on the input table). But creating a full copy will also use much more memory.

I think a good solution would be to return a table only containing the new features and havng the `table_to_X` method of the sklearn reconstructures collect features from both the input table and the generated features table to keep both cleanly separated.

### LINKED ISSUE #2250: TableLoader breaks event order of telescope events when joining observation information
When joining the by-observation-block info, tables are sorted by obs_id and obs_id is converted to float

maxnoe:
Refactor ctapipe-apply models
    
* The tool now uses the HDF5Merger component for copying data from
the input file, which makes it possible to select what data should
be copied.

* Data is only read once in chunks from the input files and all
reconstructors are applied in order to each of the chunks.

For this, it was needed to not modify the input table in the feature generator, which I opted to do by the shallow copy approach as discussed here: #2248 

Fixes #2248 
Fixes #2130
Fixes #2250

### DIFF
--- a/ctapipe/tools/apply_models.py
+++ b/ctapipe/tools/apply_models.py
@@ -1,16 +1,14 @@
 """
 Tool to apply machine learning models in bulk (as opposed to event by event).
 """
-import shutil
-
 import numpy as np
 import tables
-from astropy.table.operations import vstack
+from astropy.table import Table, join, vstack
 from tqdm.auto import tqdm
 
 from ctapipe.core.tool import Tool
-from ctapipe.core.traits import Integer, List, Path, classes_with_traits
-from ctapipe.io import TableLoader, write_table
+from ctapipe.core.traits import Bool, Integer, List, Path, classes_with_traits, flag
+from ctapipe.io import HDF5Merger, TableLoader, write_table
 from ctapipe.io.astropy_helpers import read_table
 from ctapipe.io.tableio import TelListToMaskTransform
 from ctapipe.io.tableloader import _join_subarray_events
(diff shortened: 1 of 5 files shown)


ctapipe/tools/apply_models.py
```diff
@@ -100,84 +138,72 @@ def setup(self):
...
 
-                    write_table(
-                        table[output_columns],
-                        self.output_path,
-                        f"/dl2/event/telescope/{prop}/{prefix}/tel_{tel_id:03d}",
-                        append=True,
-                    )
-                tel_tables.append(table)
+            if tel not in reconstructor._models:
+                self.log.warning(
```
LukasNickel, line 172: Should this be a warning or raise an error?
I am not entirely sure yet what I would expect to happen if my reconstructor has not been trained for the complete array.

maxnoe: It should probably be an error, yes. Will have to check how that works with our test files though.

kosack (approved):

maxnoe:
When running this using the test_ctapipe_ml repo, I get warnings from table loader that for some columns the table order doesn't match, which I think means that we are writing out predictions in the wrong order. I have to investigate why and why the tests do not catch that.

maxnoe:
The single telescope predictions are in the wrong order. and the obs-id for some reason is a float.

Will investigate and add tests.

maxnoe:
Okay, issue in tableloader found and fixed

LukasNickel (approved):

ctapipe/tools/apply_models.py
maxnoe: Since this is not a change to before, let's discuss in an issue

kosack (approved):

PR #2259, opened 2023-02-15
Fix deprecation warning due to pkg_resources

maxnoe:
pkg_resources is deprecated in general and importlib.{resources,metadata} designed to replace it.

Recently, just importing pkg_resources raises a `DeprecationWarning`.

This also replaces an external dependency with using python's stdlib (or a backport for older versions that do not yet have everything we need)

### DIFF
--- a/ctapipe/core/provenance.py
+++ b/ctapipe/core/provenance.py
@@ -17,14 +17,17 @@ from importlib import import_module
 from os.path import abspath
 from pathlib import Path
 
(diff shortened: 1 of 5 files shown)


kosack (approved):

Tobychev:
This just needs a description explaining why the changes are needed and then I can approve :)

Tobychev (approved):

PR #2261, opened 2023-02-16
Fix wrong activity meta being written into output files

### LINKED ISSUE #2260: CTA ACTIVITY END TIME doesn't appear in Reference metadata
**Describe the bug**

the `CTA ACTIVITY STOP TIME` header is missing in output files.  This is because the provenance info that records it happens after DataWriter is already finished, so we never implemented it in DataWriter.  We should maybe just hack it to be written as the time when DataWriter is finished, not getting it from the provenance module, since that is only available after `Provenance().end_activity()` is called. 

**Additional context**

This is not a critical bug, as this info is recorded in the provenance log, but it can be useful info to attach to the file for bookeeping purposes, i.e. to check how long it took to create a specific file without reading the provenance log.  However, the file's last access time could also work there.

maxnoe:
This also adds the stop time.

Fixes #2260

### DIFF
--- a/ctapipe/core/provenance.py
+++ b/ctapipe/core/provenance.py
@@ -88,9 +88,9 @@ class Provenance(metaclass=Singleton):
 
     def _get_current_or_start_activity(self):
         if self.current_activity is None:
-            log.warning(
-                "No activity has been started yet."
-                " Provenance().start_activity(<name>) should be called explicitly."
+            log.info(
+                "No activity has been explicitly started, starting new default activity."
+                " Consider calling Provenance().start_activity(<name>) explicitly."
             )
             self.start_activity()
 
--- a/ctapipe/core/provenance.py
+++ b/ctapipe/core/provenance.py
@@ -86,6 +86,16 @@ class Provenance(metaclass=Singleton):
         self._activities.append(activity)
         log.debug(f"started activity: {activity_name}")
 
+    def _get_current_or_start_activity(self):
+        if self.current_activity is None:
+            log.warning(
+                "No activity has been started yet."
+                " Provenance().start_activity(<name>) should be called explicitly."
+            )
+            self.start_activity()
+
+        return self.current_activity
+
     def add_input_file(self, filename, role=None):
         """register an input to the current activity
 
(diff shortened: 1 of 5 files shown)


kosack (dismissed):

maxnoe:
Tests are currently failing because we as it is now, the .current_activity property does not autostart one when it doesn't exist. However, I think this is better, and we should move the starting a new activity to another place so it is clear whether something is currently running,  just finished or nothing has been started yet

LukasNickel (dismissed):

ctapipe/core/provenance.py
```diff
@@ -86,6 +86,16 @@ def start_activity(self, activity_name=sys.executable):
         self._activities.append(activity)
         log.debug(f"started activity: {activity_name}")
 
+    def _get_current_or_start_activity(self):
+        if self.current_activity is None:
+            log.warning(
+                "No activity has been started yet."
+                " Provenance().start_activity(<name>) should be called explicitly."
+            )
+            self.start_activity()
```
Tobychev, line 95: Since getting here is worth a warning, it is probably a good idea to give the activity a special name like "not-properly-started" to encourage developers to start them properly.

ctapipe/core/provenance.py
```diff
@@ -156,9 +174,8 @@ def activity(self, name):
     @property
     def current_activity(self):
```
Tobychev, line 175: It is not clear to me why the code got moved from here to the entirely new `_get_current_or_start_activity`?

Within `provenance.py` itself it looks like the new implementation is just adding a new level if indirection but will otherwise execute in exactly the same way, and the actual fix of the bug in `datawriter.py` does not use this new function either.

maxnoe: The difference is now we only do that in the methods that attach new information (add_config, add_input, add_output), not in the ones that just access the current activity for reading.

The problem was that the `DataWriter` just by accessing the `current_activity` would create a fresh one and write the info of that and not write the info of the last finished activity.

ctapipe/core/provenance.py
maxnoe: After adding this warning, it popped up a lot in interactive use. I think this is over-the-top, I will remove the warning again and change to info.

kosack (approved):

Issue #2264, opened 2023-02-21
HillasIntersection produces error on badly reconstructed events


gschwefer:
**Describe the bug**
When using the `HillasIntersection` algorithm, the direction reconstruction in the nominal frame can lead to reconstructed values of `src_fov_lat` with an absolute value greater than 90 degrees. This occurs when there are few telescope images (i.e. usually 2) in the reconstruction which have very close values of `hillas_psi` . The ellipses are then almost parallel and the major axes intersect at great distances from the camera center.

 When this results in `src_fov_lat` with an absolute value greater than 90 degrees, `SkyCoord` here

https://github.com/cta-observatory/ctapipe/blob/6c4bd9cc73d0845feb7f8abd08dcc45993edb67f/ctapipe/reco/hillas_intersection.py#L228-L230

throws a corresponding error. In `ctapipe-process`, the processing will be aborted.

**Expected behavior**
Since these events are so badly reconstructed that they will be thrown away for any analysis anyway, this case may as well result in an invalid reconstruction for the event. This would allow processing to continue. I'm open to better solutions, that just seems like the easiest.

**Additional context**

For the `HillasReconstructor`, this is only a problem if all reconstructed major axes are _exactly_ aligned.

gschwefer:
If the fix suggested above is ok, I could open a pull request very quickly. Is this something that should also go in a unit test?

maxnoe:
> , this case may as well result in an invalid reconstruction for the event. 

Yes, that's the right thing to do. Check the value of `src_fov_lat` and if it is larger than 90 degrees return the invalid result container.

The actual threshold could probably be even lower than the threshold that triggers the error in astropy and fov_lon and fov_lat aren't really different in this regard.

Maybe check for 45 degrees in both?

gschwefer:
Yes, sounds good

kosack:
This looks related to #1200 as well

PR #2265, opened 2023-02-22
Fix HillasIntersection error for badly reconstructed events

gschwefer:
Fix for issue #2264. Events with large reconstructed angular distances from the camera center now return the INVALID container.


ctapipe/reco/hillas_intersection.py
```diff
@@ -218,6 +218,15 @@ def _predict(self, hillas_dict, array_pointing, telescopes_pointings=None):
         src_fov_lon, src_fov_lat, err_fov_lon, err_fov_lat = self.reconstruct_nominal(
             hillas_dict_mod
         )
+
+        # Catch events reconstructed at great angular distance from camera center
+        # and retrun INVALID container to prevent SkyCoord error below.
+        if (
+            np.abs(src_fov_lat) * u.rad > 45 * u.deg
```
maxnoe, line 225: I'd define the 45 ° limit in a named constant on top, something like `FOV_ANGULAR_DISTANCE_LIMIT = 45 * u.deg` and then compare to `FOV_ANGULAR_DISTANCE_LIMIT.to_value(u.rad)` instead of multiplying both sides with the unit (faster).

Maybe you can also put this part into a function so it looks like this:

```
if far_outside_fov(src_fov_lat, src_fov_lon):
    return INVALID
```

maxnoe (changes requested):
Could you add a test for this case? I.e. by constructing an `ArrayEvent` with specific values in the `HillasParameterContainer` that triggers this issue?

ctapipe/reco/hillas_intersection.py
```diff
@@ -43,6 +43,29 @@
     prefix="HillasIntersection",
 )
 
+FOV_ANGULAR_DISTANCE_LIMIT = 45 * u.deg
+
+
+def far_outside_fov(fov_lat, fov_lon):
```
maxnoe, line 49: I'd make this private (i.e. add an underscore in front)

gschwefer: Yes, no problem

ctapipe/reco/hillas_intersection.py
```diff
@@ -218,6 +241,12 @@ def _predict(self, hillas_dict, array_pointing, telescopes_pointings=None):
         src_fov_lon, src_fov_lat, err_fov_lon, err_fov_lat = self.reconstruct_nominal(
             hillas_dict_mod
         )
+
+        # Catch events reconstructed at great angular distance from camera center
+        # and retrun INVALID container to prevent SkyCoord error below.
+        if far_outside_fov(src_fov_lat * u.rad, src_fov_lon * u.rad):
```
maxnoe, line 247: It's relatively expensive to create new, single value quantities like this:
```
In [1]: import astropy.units as u

In [2]: %timeit 5 * u.deg
6.11 µs ± 285 ns per loop (mean ± std. dev. of 7 runs, 100,000 loops each)

In [3]: %timeit 5.0 * u.deg
5.33 µs ± 116 ns per loop (mean ± std. dev. of 7 runs, 100,000 loops each)

In [4]: %timeit u.Quantity(5.0, u.deg)
3.14 µs ± 194 ns per loop (mean ± std. dev. of 7 runs, 100,000 loops each)
```

Together with the above comment, I'd not use astropy quantities here. You know these are in radians, so I'd implement the method above purely on floats. The constant could be defined like this:

```
FOV_ANGULAR_DISTANCE_LIMIT_RAD = (45 * u.deg).to_value(u.rad)
```

maxnoe (commented):
Looks good, aside from two minor comments regarding performance and visibility

gschwefer:
Thanks!  I will implement them and get back to you.

ctapipe/reco/hillas_intersection.py
```diff
@@ -43,6 +43,29 @@
...
+    fov_lon : u.Quantity[angle]
+        Longitude in TelescopeFrame or NominalFrame
+
+    Returns
+    -------
+    bool
+    """
+    lat_outside_fov = np.abs(fov_lat) > FOV_ANGULAR_DISTANCE_LIMIT_RAD
+    lon_outside_fov = np.abs(fov_lon) > FOV_ANGULAR_DISTANCE_LIMIT_RAD
+    return lat_outside_fov or lon_outside_fov
```
kosack, line 67: fov_lat and fov_lon are spherical coordinates, so when talking about large angles, I think this should be defined as an angular distance on a sphere, not just a cut on lat and lon. Remember that in spherical coordinates, degrees in lat are in general smaller on the sky than degrees in lon, except a the equator, so just comparing them each like this will give different results for different latitudes. 

It might  be better to use `SkyCoord.separation()` instead

kosack (changes requested):

ctapipe/reco/hillas_intersection.py
kosack:
```py3
point = SkyCoord(fov_lat, fov_lon, frame=TelescopeFrame)
center = SkyCoord(0*u.deg, 0*u.deg, frame=TelescopeFrame)
return point.angular_separation(center).to_value(rad) > FOV_ANGULAR_DISTANCE_LIMIT_RAD
```

maxnoe: >  Remember that in spherical coordinates, degrees in lon are in general smaller on the sky than degrees in lat, except a the equator (here the center of the FOV),

We are at the equator (or even more precisely the crossing of zero meridian and equator) and want to remove things that are far away from it, so for all intents and purposes, I think we can treat lat and lon as equivalent here.

It's also not something we need to precise about, since it is really just some arbitrary limit to prevent an error from happening. 

computing the full angular separation will be much more expensive.

If you think it is really necessary, it would be much faster to call it directly using `astropy.coordiantes.angle_utilities.angular_separation` instead of constructing skycoordinates.

However, I don't think we need to care about cutting out a circle vs. a square here and should chose the fastest, sensible method.

maxnoe:
````
In [10]: np.rad2deg(angular_separation(np.pi /4, 0, 0, 0))
Out[10]: 45.0

In [11]: np.rad2deg(angular_separation(np.pi /2, 0, 0, 0))
Out[11]: 90.0

In [12]: np.rad2deg(angular_separation(0, np.pi/2, 0, 0))
Out[12]: 90.0

In [13]: np.rad2deg(angular_separation(0, np.pi/4, 0, 0))
Out[13]: 45.0
```

maxnoe:
@gschwefer Can you merge or rebase with the current `main` branch to get the correct CI definition so we can get this in?

gschwefer:
Yes I will

ctapipe/reco/hillas_intersection.py
maxnoe: @kosack Could you comment if you think it is necessary for this check to use true angular separation or the current, slightly faster solution is fine?

kosack (approved):

maxnoe (approved):

PR #2267, opened 2023-02-24
Use ruff instead of isort/flake8 in pre-commit, add pre-commit CI workflow

nbiederbeck:
A few weeks ago I found [ruff](https://github.com/charliermarsh/ruff), which is a python linter and fixer. According to [testimonials](https://github.com/charliermarsh/ruff#testimonials) and personal experience it is super-fast.  It checks everything from `flake8` to `isort` to `pylint` (and `black` is currently being implemented) in a single run:
```
$ hyperfine 'ruff .' 'flake8 .' -i -r 100
Benchmark 1: ruff .
  Time (mean ± σ):      58.0 ms ±   4.6 ms    [User: 32.9 ms, System: 83.4 ms]
  Range (min … max):    43.2 ms …  66.8 ms    100 runs

  Warning: Ignoring non-zero exit code.

Benchmark 2: flake8 .
  Time (mean ± σ):     706.4 ms ±  18.0 ms    [User: 12423.5 ms, System: 782.9 ms]
  Range (min … max):   677.1 ms … 767.5 ms    100 runs

  Warning: Ignoring non-zero exit code.

Summary
  'ruff .' ran
   12.19 ± 1.01 times faster than 'flake8 .'
```

* This PR uses ruff instead of isort/flake8 in the pre-commit config and adds a new CI workflow that runs `pre-commit` on the changed files. Additionally pylint-checks are done by ruff. 
* Ruff ignores `PLR2004 Magic value used in comparison, consider replacing 2 with a constant variable` in test files.
* This PR removes the codacy configuration, which had only pylint configured, because that is taken care of by pre-commit and the CI
* I added a commit that changes a file with errors, for debugging the CI and displaying how it works. That one must be reverted before merging.

I also moved some `pip installs` from the workflow files into the `setup.cfg`, to clean up a little.

### DIFF
--- a/ctapipe/io/astropy_helpers.py
+++ b/ctapipe/io/astropy_helpers.py
@@ -1,5 +1,6 @@
 #!/usr/bin/env python3
 """
+
 Functions to help adapt internal ctapipe data to astropy formats and conventions
 """
 import os
@@ -65,7 +66,6 @@ def read_table(
     """
 
     with ExitStack() as stack:
-
         if not isinstance(h5file, tables.File):
             h5file = stack.enter_context(tables.open_file(h5file))
 
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -66,5 +66,7 @@ filterwarnings = [
         path = ""
 
 [tool.ruff]
- # F: pyflakes, E/W: pycodestyle, I: isort, N: pep8-naming, PL: pylint
- select = ["F", "E", "W", "I", "N", "PL"]
+# F: pyflakes, E/W: pycodestyle, I: isort, N: pep8-naming, PL: pylint
+select = ["F", "E", "W", "I", "N", "PL"]
+[tool.ruff.per-file-ignores]
+"test_*.py" = ["PLR2004"]
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -64,3 +64,7 @@ filterwarnings = [
     [[tool.towncrier.section]]
         name = ""
         path = ""
+
+[tool.ruff]
+ # F: pyflakes, E/W: pycodestyle, I: isort, N: pep8-naming, PL: pylint
+ select = ["F", "E", "W", "I", "N", "PL"]
--- a/setup.cfg
+++ b/setup.cfg
@@ -49,6 +49,9 @@ install_requires=
 [options.extras_require]
 tests =
     pytest
+    pytest-cov
+    pytest-xdist
+    coverage != 6.3.0
     pandas ~=1.0
     tomli
     pytest_astropy_header
@@ -67,6 +70,7 @@ docs =
     graphviz
     pandas
     ipython
+    restructuredtext-lint
 
 dev =
     setuptools_scm[toml]
--- a/.codacy.yml
+++ b/.codacy.yml
@@ -1,9 +1,0 @@
----
-engines:
- pylint:
-   enabled: true
-   python_version: 3
- pyflakes:
-   disable:
-     - F999
-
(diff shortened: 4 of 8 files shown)


nbiederbeck:
Please check how much ruff/pylint finds: https://github.com/cta-observatory/ctapipe/actions/runs/4261854717/jobs/7416665240

pyproject.toml
```diff
@@ -64,3 +64,9 @@ filterwarnings = [
     [[tool.towncrier.section]]
         name = ""
         path = ""
+
+[tool.ruff]
```
Tobychev, line 68: Ruff specifies that compatibility with Black requires setting the same line length for both, so it's probably best to set it explicitly for ruff as well as for Black.

ctapipe/io/astropy_helpers.py
```diff
@@ -1,5 +1,6 @@
 #!/usr/bin/env python3
 """
+
```
Tobychev, line 3: What are these new lines about? Are they from differences between the formatting produced by the current linters and the one you propose?

setup.cfg
```diff
@@ -49,6 +49,9 @@ install_requires=
 [options.extras_require]
 tests =
     pytest
+    pytest-cov
+    pytest-xdist
+    coverage != 6.3.0
```
Tobychev, line 54: Are these additions required to switch to ruff? Or are they part of that commit to be reverted you mentioned? 

If they aren't required to run ruff you should update the pull request to describe the reason they are added.

kosack:
I think this issue is not named correctly: this PR is not replacing _codacy_ (which is a web front-end viewer for code quality that runs independently of the CI), but replacing _pylint_ in the CI with _ruff_.

For the codacy replacement, I think soon we will migrate to using SonarQube provided by CTAO/DPPS (See #2211).  SonarQube runs SonarScanner locally in the CI, and then reports results to the front-end. But that can be independent of something like ruff/pylint running directly in the CI

nbiederbeck:
You both are right, I got carried away while working on this. This PR does the following (and I will happily split it up if necessary):

1. Remove the `codacy.yml`. I incorrectly assumed that would be enough to disable codacy. Since it didn't complain I thought I was successful. That is wrong.
2. Move some `pip install`s from the CI to `setup.cfg`. That way one can locally do the same as in the CI, without looking into the workflow file. (*)
3. Add a linting workflow that uses [pre-commit/action](https://github.com/pre-commit/action), it simply runs all pre-commit hooks on all changed files in any PR. (*)
4. We only checked for `flake8` compliance in the CI and additionally `isort` and `black` via pre-commit hooks. With (3.) we now check everything locally and in the CI. We only used pylint via codacy. The change to `ruff` enables us to (quickly) do the pylint-checks during pre-commit and CI.

I'm pushing for the items marked with (*), as I think they make the developers and maintainers life easier. The other two are open for discussion and maybe even worth a CEP.

ctapipe/io/astropy_helpers.py
nbiederbeck: I changed this file to show how the CI behaves when one submits a wrongly formatted file.  This should not be part of the PR.  I wanted to avoid opening a PR changing the CI, where everything works (could easily be false positives).

setup.cfg
nbiederbeck: Those are moved from `.github/workflows/ci.yml`, that the CI setup does not differ from local installs.

nbiederbeck:
I close this because I think the scope is too large and should be done in separate PRs, as done in #2276.

Issue #2177, opened 2023-01-04
QualityQuery statistics are written to hdf5 file without the <kind> group


maxnoe:
**Describe the bug**

```
❯ h5dump -n gamma-diffuse_with_images_00.dl2.h5 | grep statistics
 dataset    /dl1/service/image_statistics
 group      /dl2/service/tel_event_statistics
 dataset    /dl2/service/tel_event_statistics/HillasReconstructor
```

These should have the prefix `geometry` / `energy`, right?

Tobychev:
@kosack I think this is a question for you?

kosack:
it's from the machine learning tools, but yes it should have a prefix since there are multiple quality queries.

PR #2268, opened 2023-02-24
Add stale workflow

nbiederbeck:
This adds the [stale action](https://github.com/actions/stale) from github.

to be discussed: the number of days before issues/pr get labeled as `stale` and how long after that before it gets closed.

Currently this does not label PRs that have a milestone.

I could not find a way to include the number of comments as a filter, but anyways I think it is ok to mark long discussions as stale, when they are old. In any case, one can just remove the `stale`-label and the counter starts again.

### DIFF
--- a/.github/workflows/stale.yml
+++ b/.github/workflows/stale.yml
@@ -1,0 +1,20 @@
+name: 'Close stale issues and PRs'
+on:
+  schedule:
+    - cron: '30 1 * * *'
+  workflow_call:
+
+jobs:
+  stale:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/stale@v7
+        with:
+          stale-issue-message: 'This issue is stale because it has been open 30 days with no activity. Remove stale label or comment or this will be closed in 5 days.'
+          days-before-issue-stale: 30
+          days-before-issue-close: 5
+          stale-pr-message: 'This PR is stale because it has been open 45 days with no activity. Remove stale label or comment or this will be closed in 10 days.'
+          days-before-pr-stale: 45
+          days-before-pr-close: 10
+          exempt-all-pr-milestones: true  # don't stale PRs that have a milestone attached
+          debug-only: true


nbiederbeck:
Please take a look at the output: https://github.com/cta-observatory/ctapipe/actions/runs/4261911561/jobs/7416787168

Only 30 items (PRs+Issues) are processed due to smart rate-limiting on the API: https://github.com/actions/stale#operations-per-run

Tobychev:
> Please take a look at the output: https://github.com/cta-observatory/ctapipe/actions/runs/4261911561/jobs/7416787168
The linked output is unfortunately not very suitable to check how well the labelling works, clicking on each issue or PR to find the verdict and then looking up the actual issue gets old pretty quick. 

Also, I think it would be more useful for judging utility if you run it on the oldest issues first, starting with the newest issues means it mostly doesn't label anything as stale. 

Finally, I picked one issue from the output without much tough and it sort of _looks_ stale but also hard to classify as stale purely on metrics like activity, and indeed it was judged fresh by the filter... I'm talking about #2177 which sees Max asking a question and not getting any reply since almost two months ago.

My conclusion from that single example is that the tool will always be rather blunt, so it'll be mostly to prod us into taking some sort manual action. Maybe we can think of having something like "stale days", where we run the action with aggressive settings and then go trough all the issues and consider them manually. 

Another conclusion is it should _never_ close issues as things stands, we don't have the stream of duped issues requiring it.

nbiederbeck:
@Tobychev I agree with all you said. Some comments:

There is a setting to *never close* issues.  We can use it.  I assume authors and commenters get notified when a label gets added to a PR/issue.  I would argue that the "stale" label (or we rename it to anything else) should be seen as a "do you have more information now than x weeks ago" instead of a "no one is working on this, will close".

If we decide to use this action, I think we should manually run it on *all* PRs/issues once (e.g. during merge) and then repeatedly, e.g. once a week.  See above, this should just (imo) be a "ping" to authors.  We could argue for Monday or Friday mornings (or both): we would already have something to discuss (any new information on old issues) or have the whole week to work on it.

Tobychev (dismissed):

maxnoe:
Let me remark here that we went over old issues like once or twice a year (or also after releases have been made).

Many of the issues we have open and that are open since a long time are design issues that will need to be addressed in future or feature requests that are valid but weren't high priority until now.

So what in your point of view is really the added benefit of running this? We are not really intending on closing such issues, right?

I am a big opponent of the practice  of some open source projects that close bug reports or valid feature requests just because some amount of time without activity has elapsed.
Nothing more annoying than running into a bug, seeing the corresponding bug report closed only to find it has not been actually fixed but just closed due to "inactivity".

nbiederbeck:
> So what in your point of view is really the added benefit of running this?

Automation.  

We agree on *not closing*.  I think it saves some time having a system that automatically goes "over old issues like once or twice a year" (e.g. set the number of days to 180 should do the trick?) Or even run this job after the release job (potentially with different settings).

kosack:
I think if the issue is important enough that we should address it before for example v1.0.0, then we should assign a milestone (in which case the issue will not be marked as stale)

Tobychev (approved):

PR #2273, opened 2023-03-01
Fix sorting in TableLoader

LukasBeiske:
This removes a redundant sorting in `TableLoader` which caused the sorting index to be deleted before the final sort to original order when reading events.

### DIFF
--- a/ctapipe/io/tableloader.py
+++ b/ctapipe/io/tableloader.py
@@ -349,19 +349,12 @@ class TableLoader(Component):
         # casts the obs_id in the joint result to float.
         obs_table["obs_id"] = obs_table["obs_id"].astype(table["obs_id"].dtype)
 
-        # to be able to sort to original table order
-        self._add_index_if_needed(table)
-
         joint = join_allow_empty(
             table,
             obs_table,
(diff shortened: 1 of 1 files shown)


maxnoe:
This is precisely the code I added in https://github.com/cta-observatory/ctapipe/pull/2249/commits/9bd7b1443e97e2f3338c62cc12d2afcc9c8f5bc4

to fix  this: https://github.com/cta-observatory/ctapipe/pull/2249#issuecomment-1423051936

So this sort is really, really not redundant, which is also why now the unit tests are failing with the error that when obs_info is loaded, the order of the events is wrong:

```
FAILED ../../home/runner/work/ctapipe/ctapipe/io/tests/test_table_loader.py::test_order_merged - ValueError: Tables have subarray events in different order
FAILED ../../home/runner/work/ctapipe/ctapipe/tools/tests/test_apply_models.py::test_apply_all - ValueError: Tables have subarray events in different order
```

LukasBeiske:
Oh, sorry. I thought this was something I missed when I last changed that function.

Closed in favor of #2274

PR #2275, opened 2023-03-01
Image-fitting algorithm

clara-escanuela:


clara-escanuela:
This is my first step towards this new image-fitting method. Although I would like to get suggestions about:
1) How to make ellipsoid.py more general and include an option for PDFs
2) I want to now integrate this into HillasReconstructor. How would you proceed?

clara-escanuela:
I also think that the parameter boundaries should be calculated in ellipsoid.py and not being optional. So maybe I could include another function in ellipsoid.py that finds the limits of each parameter

maxnoe:
Let's do one thing after another. Adding new image parameters is already a large enough pull request, let's not mix it with changing the HillasReconstructor

ctapipe/image/toymodel.py
```diff
@@ -369,3 +371,72 @@ def pdf(self, x, y):
         r = np.sqrt((x - self.x) ** 2 + (y - self.y) ** 2)
 
         return norm(self.radius.to_value(u.m), self.sigma.to_value(u.m)).pdf(r)
+
+
+class SkewedCauchy(ImageModel):
```
maxnoe, line 376: It's not a `SkewedCauchy`, is it? It's a SkewedGaussian for the longitudinal and a normal cauchy for trans, right?

clara-escanuela: Yes. I will clarify

ctapipe/image/ellipsoid.py
```diff
@@ -0,0 +1,494 @@
...
+def sensible_boundaries(geometry, cleaned_image, pdf):
+    """
+    Computes boundaries of the fit based on the deviation from Hillas parameters
+    Parameters
+    ----------
+    geometry: ctapipe.instrument.CameraGeometry
+        Camera geometry
+    cleaned_image: ndarray
+        Charge for each pixel, cleaning mask should be applied
+    pdf: str
```
kosack, line 93: Might be useful to use an Enum here and elsewhere for `pdf` (for speed and correctness and to make it easier to include in a config file later).

ctapipe/image/ellipsoid.py
```diff
@@ -0,0 +1,494 @@
...
+    geometry : ctapipe.instrument.CameraGeometry
+        Camera geometry
+    image : ndarray
+        Charge in each pixel, no cleaning mask should be applied
+    dilated_mask : boolean
+        mask after image cleaning and dilation
+    x0 : dict
+       seeds of the fit
+    pdf: str
+       name of the PDF, options = "Gaussian", "Laplace", "Skewed"
```
kosack, line 152: same here, use an Enum

ctapipe/image/ellipsoid.py
```diff
@@ -0,0 +1,494 @@
...
+    width_min, width_max = x0["width"].value, x0["width"].value + width_unc.to_value(
+        unit
+    )
+
+    scale = length_min / np.sqrt(1 - 2 / np.pi)
+    skew_min, skew_max = max(-0.99, x0["skewness"] - 0.3), min(
+        0.99, x0["skewness"] + 0.3
+    )
+
+    if pdf == "Gaussian":
```
kosack, line 214: this then becomes, e.g. `PDFType.Gaussian`

clara-escanuela: Is that what you were asking for?

clara-escanuela:
Here are some examples of the performance of the fit when applying a 2D Gaussian:
image

clara-escanuela:
Maybe this is also interesting. For these tests, I am using an array of 14 MSTs and the HillasReconstructor, telescope multiplicity cut >= 5 triggered telescopes.
image

maxnoe:
@clara-escanuela Could you also include please the standard hillas parameters based approach without applying a centroid cut?

maxnoe:
As discussed via slack, I don't think that the centroid is a good criterion, the `leakage_*` are better parameters. 

Thinking about it: since you get a predicted image from the model, and the model pdf is normalized, you should very easily be able to compute the expected containment from the image fit as `(predicted_image.sum() / normalization)` , which might be an interesting parameter.

maxnoe:
And when you do the cut on the centroid: is it the "new" centroid from the image fit for the fit or the same hillas centroid for both?

clara-escanuela:
> And when you do the cut on the centroid: is it the "new" centroid from the image fit for the fit or the same hillas centroid for both?

The centroid from Hillas. I wanted to compare the exact same events.

clara-escanuela:
> @clara-escanuela Could you also include please the standard hillas parameters based approach without applying a centroid cut?

Here I am plotting Hillas when I do not apply any cut on the centroid distance (light blue), Hillas when I apply a cut on centroid (need to change definition) at 0.8m (dark blue), and 2D Fit with no cut on centroid. I fit all images and not only truncated. It would be interesting to see how this plot looks like when fitting only truncated images to check what it is the actual impact on only truncated images. 

image

ctapipe/containers.py
```diff
@@ -216,6 +217,37 @@ class CameraHillasParametersContainer(BaseHillasParametersContainer):
...
+    Hillas Parameters after fitting. The cog position
+    is given in meter from the camera center.
+    """
+
+    skewness_uncertainty = Field(nan, "measure of skewness uncertainty")
+    likelihood = Field(nan, "measure of likelihood")
+    goodness_of_fit = Field(
+        nan, "measure of goodness of fit, mean likelihood subtracted to the likelihood"
+    )
+    n_pix_fit = Field(nan, "number of pixels used in the fit")
```
maxnoe, line 231: Please avoid such unneeded abbreviations.

ctapipe/containers.py
```diff
@@ -216,6 +217,37 @@ class CameraHillasParametersContainer(BaseHillasParametersContainer):
     psi = Field(nan * u.deg, "rotation angle of ellipse", unit=u.deg)
 
 
+class ImageFitParametersContainer(CameraHillasParametersContainer):
```
maxnoe, line 220: We added support for the toy models in telescope frame now, so it should be possible (and should be the default) that this happens also in telescope frame (i.e. using fov_lon / fov_lat in degrees instead of x, y in m)

ctapipe/image/ellipsoid.py
```diff
@@ -0,0 +1,443 @@
...
+
+class PDFType(Enum):
+    gaussian = "gaussian"
+    laplace = "laplace"
+    skewed = "skewed"
+
+
+def create_initial_guess(geometry, image, size):
+    """
+    This function computes the seeds of the fit with the Hillas parameters
```
maxnoe, line 50: I think you need empty lines before headings in the docstrings. At least it is much more readable

ctapipe/image/ellipsoid.py
```diff
@@ -0,0 +1,443 @@
...
+        improve performance.
+    size: float/int
+        Total charge after cleaning and dilation
+    Returns
+    -------
+    initial_guess: dict
+    Seed parameters of the fit.
+    """
+    unit = geometry.pix_x.unit
+    hillas = hillas_parameters(geometry, image)  # compute Hillas parameters
```
maxnoe, line 66: Why compute hillas parameters again instead of just taking them as input?

ctapipe/image/ellipsoid.py
```diff
@@ -0,0 +1,443 @@
...
+    initial_guess["skewness"] = hillas.skewness
+
+    if (hillas.width.to_value(unit) == 0) or (hillas.length.to_value(unit) == 0):
+        raise ImageFitParameterizationError("Hillas width and/or length is zero")
+
+    initial_guess["amplitude"] = size
+    return initial_guess
+
+
+def dilation(n, cleaned_mask, geometry):
```
maxnoe, line 90: Pretty sure this already exists in `ctapipe.image.cleaning`, no need to redefine

ctapipe/image/ellipsoid.py
```diff
@@ -0,0 +1,443 @@
...
+        Camera geometry
+    cleaned_image: ndarray
+        Charge for each pixel, cleaning mask should be applied
+    pdf: PDFType instance
+        e.g. PDFType("gaussian")
+    Returns
+    -------
+    list of boundaries
+    """
+    hillas = hillas_parameters(geometry, cleaned_image)
```
maxnoe, line 128: Again a recomputation of parameters...

ctapipe/image/ellipsoid.py
```diff
@@ -0,0 +1,443 @@
...
+
+    unit = geometry.pix_x.unit
+    camera_radius = geometry.guess_radius()
```
maxnoe, line 131: Is this memoized? If not it might be expensive to call this in the event loop

ctapipe/image/ellipsoid.py
```diff
@@ -0,0 +1,443 @@
...
+
+    cogx_min, cogx_max = np.sign(hillas.x) * min(
+        np.abs(hillas.x - u.Quantity(0.2, unit)), camera_radius
```
maxnoe, line 134: magic number 0.2?

ctapipe/image/ellipsoid.py
```diff
@@ -0,0 +1,443 @@
...
+from ctapipe.image.hillas import hillas_parameters
+from ctapipe.image.pixel_likelihood import (
+    mean_poisson_likelihood_gaussian,
+    neg_log_likelihood_approx,
+)
+from ctapipe.image.toymodel import SkewedGaussian, SkewedGaussianLaplace
+
+from ..containers import ImageFitParametersContainer
+
+PED_TABLE = {
```
maxnoe, line 24: Where do these come from? Why these keys?

ctapipe/image/toymodel.py
```diff
@@ -369,3 +399,88 @@ def pdf(self, x, y):
         r = np.sqrt((x - self.x) ** 2 + (y - self.y) ** 2)
 
         return norm(self.radius.to_value(u.m), self.sigma.to_value(u.m)).pdf(r)
+
+
+class SkewedGaussianLaplace(ImageModel):
```
maxnoe, line 404: Is there a physical motivation for this? Does it fit better?

ctapipe/image/ellipsoid.py
kosack: it is current not memoized, so it will be quite slow to call this frequently. I didn't add a @lazyproperty to this call, since this is only one way to compute the edge of the camera (not sure it works well in all cases). It was just to give a quick way of computing a reasonable value.  Ideally we would have a `geom.radius` property that is  supplied externally like the pixel positions or telescope focal length, etc, and only guessed if none is provided.

ctapipe/image/ellipsoid.py
```diff
@@ -0,0 +1,443 @@
...
+    "LSTCam": 2.8,
+    "NectarCam": 2.3,
+    "FlashCam": 2.3,
+    "SST-Camera": 0.5,
+    "CHEC": 0.5,
+    "DUMMY": 0,
+    "testcam": 0,
+}
+SPE_WIDTH = 0.5
```
kosack, line 33: I guess this is also not the correct value, and you should get it from the appropriate Container.

ctapipe/image/ellipsoid.py
kosack: actually, we only have `dilate` which does a single dilation. THe loop for multiple is not there, but this code should in any case go into `cleaning.py` (maybe call it `dilate_multiple`)

kosack: or use `functools.reduce` to do the dilate(dilate(dilate()) call chaining

ctapipe/image/ellipsoid.py
clara-escanuela: I do not think the pedestal width is available at the moment. I used those values that I found in Dan's code as he also had to load the pedestal width like that.

ctapipe/image/ellipsoid.py
clara-escanuela: I also think this is not available. At least I could not find the SPE width

ctapipe/image/toymodel.py
clara-escanuela: We thought this could better estimate the width. But, after some tests, I do not think it works better than the 2D Skewed Gaussian so I think I could remove it

ctapipe/containers.py
```diff
@@ -308,6 +341,37 @@ class HillasParametersContainer(BaseHillasParametersContainer):
...
+    """
+
+    skewness_uncertainty = Field(nan, "measure of skewness uncertainty")
+    likelihood = Field(nan, "measure of likelihood")
+    goodness_of_fit = Field(
+        nan, "measure of goodness of fit, mean likelihood subtracted to the likelihood"
+    )
+    n_pixels = Field(nan, "number of pixels used in the fit")
+    free_parameters = Field(nan, "number of free parameters")
+    is_valid = Field(False, "True if the fit is valid")
```
maxnoe, line 357: What's the difference between `is_valid` and `is_accurate`?

ctapipe/image/ellipsoid.py
```diff
@@ -0,0 +1,407 @@
...
+    if (hillas.width.to_value(unit) == 0) or (hillas.length.to_value(unit) == 0):
+        raise ImageFitParameterizationError("Hillas width and/or length is zero")
+
+    initial_guess["amplitude"] = size
+    return initial_guess
+
+
+def boundaries(geometry, image, dilated_mask, hillas, pdf):
+    """
+
```
maxnoe, line 91: No empty line here

ctapipe/image/ellipsoid.py
```diff
@@ -0,0 +1,407 @@
...
+    -------
+    list of boundaries
+    """
+    x = geometry.pix_x.value
+    y = geometry.pix_y.value
+    unit = geometry.pix_x.unit
+    camera_radius = geometry.radius.to_value(unit)
+
+    # Using dilated image
+    row_image = image.copy()
```
maxnoe, line 117: Why do you set the pixels to 0 in a copy of the image instead of just using the surviving pixels like we e.g. do with the hillas parameters and the muon fit?

ctapipe/image/ellipsoid.py
```diff
@@ -0,0 +1,407 @@
...
+class ImageFitParameterizationError(RuntimeError):
+    pass
+
+
+def image_fit_parameters(
+    geom,
+    image,
+    n_row,
+    cleaned_mask,
+    pdf=PDFType("skewed"),
```
maxnoe, line 185: `PDFType.skewed`

ctapipe/image/ellipsoid.py
```diff
@@ -0,0 +1,407 @@
...
+    bounds=None,
+):
+    """
+
+    Computes image parameters for a given shower image.
```
maxnoe, line 190: No empty line before, one empty line after this one

ctapipe/image/ellipsoid.py
```diff
@@ -0,0 +1,407 @@
...
+    pars = m.values
+    errors = m.errors
+
+    like_array = likelihood
+    like_array *= dilated_mask
+    goodness_of_fit = np.sum(
+        like_array[like_array > 0]
+        - mean_poisson_likelihood_gaussian(
+            pdf_dict[pdf](
+                pars[0] * unit,
```
maxnoe, line 313: This won't work for different kinds of PDFs that have different numbers of parameters, does it?

ctapipe/image/ellipsoid.py
```diff
@@ -0,0 +1,407 @@
...
+            ).pdf(geom.pix_x, geom.pix_y),
+            SPE_WIDTH,
+            pedestal,
+        )
+    )
+
+    fit_rcog = np.linalg.norm([pars[0], pars[1]])
+    fit_phi = np.arctan2(pars[1], pars[0])
+
+    b = pars[1] ** 2 + pars[0] ** 2
```
maxnoe, line 329: These lines deserve some comments...

ctapipe/image/ellipsoid.py
```diff
@@ -0,0 +1,407 @@
...
+    fit_rcog_err = np.sqrt(
+        pars[0] ** 2 / b * errors[0] ** 2 + pars[1] ** 2 / b * errors[1] ** 2
+    )
+
+    delta_x = pix_x.value - pars[0]
+    delta_y = pix_y.value - pars[1]
+
+    longitudinal = delta_x * np.cos(pars[2]) + delta_y * np.sin(pars[2])
+
+    m4_long = np.average(longitudinal**4, weights=dilated_image)
```
maxnoe, line 342: Also compute skewness long in case of non-skewed distribution?

PR #2279, opened 2023-03-02
Fix transform of CameraGeometry

### LINKED ISSUE #2277: transforming geometry loses frame attributes
**Describe the bug**

Transforming a CameraGeometry that has a `focal_length` attribute to `EngineeringCameraFrame` resets that frame attribute.

**To Reproduce**

```py
from ctapipe.coordinates import EngineeringCameraFrame

subarray = SubarrayDescription.read("dataset://gamma_prod5.simtel.zst")
geom = subarray.tel[100].camera.geometry
print("OLD", geom.frame)
new_geom = geom.transform_to(EngineeringCameraFrame())
print("NEW", new_geom.frame)
```
```
OLD <CameraFrame Frame (focal_length=16.445049285888672 m, rotation=0.0 rad, telescope_pointing=None, obstime=None, location=None)>
NEW <EngineeringCameraFrame Frame (focal_length=0.0 m, rotation=0.0 rad, telescope_pointing=None, obstime=None, location=None, n_mirrors=1)>
```

**Expected behavior**
Frame attributes are retained, so further transforms (like back to TelescopeFrame) work

kosack:
Allow a Frame class to be passed so attributes are retained

Fixes #2277

### DIFF
--- a/ctapipe/coordinates/__init__.py
+++ b/ctapipe/coordinates/__init__.py
@@ -2,6 +2,7 @@
 Coordinates.
 """
 import warnings
+from typing import Union
 
 from astropy.coordinates import (
     CIRS,
@@ -9,6 +10,7 @@ from astropy.coordinates import (
     FunctionTransformWithFiniteDifference,
     frame_transform_graph,
 )
+from astropy.coordinates.baseframe import BaseCoordinateFrame
 
 from .camera_frame import CameraFrame, EngineeringCameraFrame
 from .ground_frames import (
@@ -42,9 +44,16 @@ class MissingFrameAttributeWarning(Warning):
     pass
 
 
-def get_representation_component_names(frame):
-    """Return the component names of a Frame (or SkyCoord)"""
-    return tuple(frame.get_representation_component_names().keys())
+def get_representation_component_names(
+    frame: Union[BaseCoordinateFrame, type[BaseCoordinateFrame]]
+):
+    """
+    Return the component names of a Frame (or SkyCoord)
+    """
+    if isinstance(frame, type):
+        return tuple(frame().get_representation_component_names().keys())
+    else:
+        return tuple(frame.get_representation_component_names().keys())
 
 
 # astropy requires all AltAz to have locations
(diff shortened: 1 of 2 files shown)


kosack:
TelescopeFrame will still lose the info, but I guess that is fundamentally because it doesn't contain that attribute.

```py3
sub = SubarrayDescription.read("dataset://gamma_prod5.simtel.zst")
geom = sub.tel[100].camera.geometry
print(geom.frame)
print(geom.transform_to(EngineeringCameraFrame).frame)
print(geom.transform_to(TelescopeFrame).frame)
```
```
<CameraFrame Frame (focal_length=16.445049285888672 m, rotation=0.0 rad, telescope_pointing=None, obstime=None, location=None)>
<EngineeringCameraFrame Coordinate (focal_length=16.445049285888672 m, rotation=0.0 rad, telescope_pointing=None, obstime=None, location=None, n_mirrors=1): (x, y) in m
    [( 0.   ,  0.   ), ( 0.   , -0.049), (-0.049,  0.   )]>
<TelescopeFrame Coordinate (telescope_pointing=None, obstime=None, location=None): (fov_lon, fov_lat) in deg
    [(0.        , 0.        ), (0.        , 0.17071965),
     (0.17071965, 0.        )]>
```

kosack:
from the [Astropy docs](https://docs.astropy.org/en/stable/coordinates/transforming.html#examples):
> The same lower-level frame classes also have a [transform_to()](https://docs.astropy.org/en/stable/api/astropy.coordinates.BaseCoordinateFrame.html#astropy.coordinates.BaseCoordinateFrame.transform_to) method that works the same as above, but they do not support attribute-style access. They are also **subtly different in that they only use frame attributes present in the initial or final frame**, while [SkyCoord](https://docs.astropy.org/en/stable/api/astropy.coordinates.SkyCoord.html#astropy.coordinates.SkyCoord) objects use any frame attributes they have for all transformation steps. So [SkyCoord](https://docs.astropy.org/en/stable/api/astropy.coordinates.SkyCoord.html#astropy.coordinates.SkyCoord) can always transform from one frame to another and back again without change, while low-level classes may lose information and hence often do not round-trip.

kosack:
as mentioned in the issue, I think I will modify this to store the pix_x, pix_y coordines internally in a SkyCoord, rather than constructing a SkyCoord during transform.  That will allow all frame attributes to be propegated, even ones not explicitly in the target frame.  That means one will be able to transform from CameraFrame to TelescopeFrame and back again without re-defining the parameters.

PR #2281, opened 2023-03-07
require lint CI before testing

nbiederbeck:


maxnoe (approved):
I think this is good to spare some computing resources

PR #2287, opened 2023-03-13
Fix overwrite not working for ctapipe-apply-models

maxnoe:
Because there was no alias defined for the `HDF5Merger` component the `--overwrite` option had no effect for `ctapipe-apply-models`.

### DIFF
--- a/ctapipe/tools/apply_models.py
+++ b/ctapipe/tools/apply_models.py
@@ -111,6 +111,13 @@ class ApplyModels(Tool):
             "Include true images",
             "Exclude true images",
         ),
+        "overwrite": (
+            {
+                "H5Merger": {"overwrite": True},
+                "ApplyModels": {"overwrite": True},
+            },
+            "Overwrite output file if it exists",
+        ),
     }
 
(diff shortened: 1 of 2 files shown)


Tobychev:
@maxnoe I edited the description with that I guessed the fix was, if you confirm the guess is accurate I can approve.

maxnoe:
@Tobychev thanks, yes!

Tobychev (approved):

kosack:
> Because there was no alias defined for the HDF5Merger component the --overwrite option had no effect 

Doesn't this mean that there simply wasn't an `--overwrite` option at all in the Tool? if so, passing `--overwrite` on the command-line should have raised an exception due to an unknown alias, right?

maxnoe:
> Doesn't this mean that there simply wasn't an --overwrite option at all in the Tool? if so, passing --overwrite on the command-line should have raised an exception due to an unknown alias, right?

The option now lives on the `Tool` itself, it only needs to be overridden if a tool as a sub-component that also needs the option.

nbiederbeck (approved):

Issue #2291, opened 2023-03-17
Support for advanced reconstructors in API


kosack:
**Please describe the use case that requires this feature.**

Soon we will want to support reconstructors that can reconstruct more than one `ReconstructionProperty` (For example ImPACT , #2052, or fancier ML methods where geometry, energy, and gammaness can all be predicted in one go).   The current design supports this, but is somewhat inconsistent  so could use some small changes. 

**Describe the solution you'd like**
A few things I noticed when looking at the current implementation:

* I would expect to be able to construct any `Reconstructor` , e.g. `reco = SomeReconstructor()`,  and check its `reco.property` field to see what it reconstructs. THis means putting it in the base class, and verifying it is set perhaps in the base `__init__`
* Neither `HillasReconstructor` nor `HillasIntersection` seem to set the _property_ field (so it's used inconsistently, and only by the sklearn reconstructors so far).  Would be nice to use it everywhere.  
* would be useful to allow `Reconstructor`s to have a _list_ of `ReconstructionProperties` rather than a single one, to support the above use case
* the `GeometryReconstructor` base class is poorly named, as it is really specific to Hillas-parameter based methods that would be inappropriate for e.g. a deep-learning or model based geometry reconstruction, and it does not set `property=ReconstructionProperty.geometry` in any case.  `HillasGeometryReconstructor` would be a be better name

PR #2292, opened 2023-03-17
Make ReconstructionProperties into a Flag enum to support same reconstructor providing many properties

Tobychev:
A first start at addressing the need in #2291 of a reconstructor indicating it reconstructs many properties. 

I changed  `ReconstructionProperties` into a enum of `Flag` type, meaning a reconstructor can say it provides both geometry and particle type by setting
`self.property = ReconstructionProperty.ENERGY | ReconstructionProperty.PARTICLE_TYPE`

It is also possible to pick out a list of properties a reconstructor supports given a flag combination in `self.property` using this (slightly unreadable) list comprehension
```
        properties = [
            self.property & itm
            for itm in self.supported
            if self.property & itm in ReconstructionProperty
        ]
```

### DIFF
--- a/ctapipe/reco/stereo_combination.py
+++ b/ctapipe/reco/stereo_combination.py
@@ -19,9 +19,9 @@ from ..containers import (
 from .utils import add_defaults_and_meta
 
 _containers = {
-    "energy": ReconstructedEnergyContainer,
-    "classification": ParticleClassificationContainer,
-    "geometry": ReconstructedGeometryContainer,
+    ReconstructionProperty.ENERGY: ReconstructedEnergyContainer,
+    ReconstructionProperty.PARTICLE_TYPE: ParticleClassificationContainer,
+    ReconstructionProperty.GEOMETRY: ReconstructedGeometryContainer,
 }
 
 __all__ = [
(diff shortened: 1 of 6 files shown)


ctapipe/reco/stereo_combination.py
```diff
@@ -107,12 +107,12 @@ class StereoMeanCombiner(StereoCombiner):
     def __init__(self, *args, **kwargs):
         super().__init__(*args, **kwargs)
 
-        supported = {
+        self.supported = {
```
kosack, line 110: You might want to make self.supported by a dispatch dictionary, like this:
```py3
self.supported_combiner = {
  ReconstructionProperty.ENERGY: self._combine_energy   
  ReconstructionProperty.PARICLE_TYPE: self._combine_particle_type
  ReconstructionProperty.GEOMETRY: self._combine_geometry
}
```

Then it simplifies life a later since you can do something similar to 
```py3
# combine all properties that can be combined
for itm in ReconstructionProperty:
    if itm in self.supported_combiner and itm in self.property:
        self.supported_combiner[itm](event)
```

with no need to even make the the `properties` list comprehension or do any bitwise checking.   In fact, since you make `properties` a list anyhow, you aren'y really using the fact that `self.property` is a `Flag` enum

maxnoe:
Do we really want a `FlagEnum` that is using binary operations to store the values instead of just a tuple of the enum like we e.g. do for `DataLevel`?

What is the advantage of using `FlagEnum`?

Tobychev:
I didn't think of using a tuple of enums, compared with that case I'm not sure there is any benefit. I can change it.

Tobychev:
@maxnoe You can't properly do it with an tuple, because the StereoCombiner takes the properties via traitlets and traitlets only support type checking fixed length tuples. But traitlets can typecheck variable size sets, so I used that instead. 

However, I realised that the assumption that a reconstructor only does one thing seems to be pretty deeply encoded in the `predict_table` functions via assumptions about the `prefix` used to name columns, and I don't have a good idea for how to resolve that.

PR #2293, opened 2023-03-18
Rename GeometryReco to HillasGeometryReco

Tobychev:
Fixes parts of #2291, specifically

> the GeometryReconstructor base class is poorly named, as it is really specific to Hillas-parameter based methods that would be inappropriate for e.g. a deep-learning or model based geometry reconstruction, and it does not set property=ReconstructionProperty.geometry in any case. HillasGeometryReconstructor would be a be better name

### DIFF
--- a/ctapipe/reco/__init__.py
+++ b/ctapipe/reco/__init__.py
@@ -5,7 +5,11 @@
 from .hillas_intersection import HillasIntersection
 from .hillas_reconstructor import HillasReconstructor
 from .impact import ImPACTReconstructor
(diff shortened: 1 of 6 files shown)


HealthyPear:
If `GeometryReconstructor` is a "Base class for algorithms predicting only the shower geometry", shouldn't `HillasGeometryReconstructor` be a child class of this implementation calling the `HillasReconstructor`?

Tobychev:
Updated the docstring after @HealthyPear's comment

kosack (approved):

maxnoe (approved):

PR #2299, opened 2023-03-30
Adapt default config for flashcam, bring in line with current grid production

maxnoe:
After #2188, we can adjust the default cleaning configuration for flashcam back to the same as for NectarCam

Also adjusted some other settings to the same values we used for the latest grid production.


Tobychev (commented):
As this "enables" (sort of) FlashCam it is probably worth having a changelog entry for it?

maxnoe:
> As this "enables" (sort of) FlashCam it is probably worth having a changelog entry for it?

There is one aleady for this from the flashcam PR. I think that's sufficient. I will have a look when rendering the full changelog for 0.19

LukasNickel (approved):

PR #2303, opened 2023-03-31
Improve info tool dependencies option

HealthyPear:
While porting this for SWGO software I modified the code to read the dependencies at runtime from `setup.cfg` (like we do already for the Sphinx documentation configuration).

While doing this I noticed that the code using `importlib.import_module` sometimes gives rise to a `ModuleNotFoundError`exception, so I added it to the first `try/except` block to fallback to `importlib.metadata.version` and only if that also fails we give up.

I also used [pytest-console-scripts](https://github.com/kvas-it/pytest-console-scripts) to create 1 test for each flag (but see [this issue](https://github.com/kvas-it/pytest-console-scripts/issues/57) I opened there, even though I tested each of them separately).

### DIFF
--- a/ctapipe/tools/info.py
+++ b/ctapipe/tools/info.py
@@ -1,7 +1,10 @@
 # Licensed under a 3-clause BSD style license - see LICENSE.rst
 """ print information about ctapipe and its command-line tools. """
+from configparser import ConfigParser
 import logging
 import os
+from pathlib import Path
+from re import split
 import sys
 
 from ..core import Provenance, get_module_version
@@ -16,26 +19,41 @@ else:
 
 __all__ = ["info"]
 
-# TODO: this list should be global (or generated at install time)
-_dependencies = sorted(
-    [
-        "astropy",
-        "matplotlib",
-        "numpy",
-        "traitlets",
-        "sklearn",
-        "scipy",
-        "numba",
-        "pytest",
-        "iminuit",
-        "tables",
-        "eventio",
-    ]
-)
+conf = ConfigParser()
+conf.read(Path(__file__).parent / Path("../../setup.cfg"))
+setup_cfg_options = dict(conf.items("options"))
+setup_cfg_options_extras = dict(conf.items("options.extras_require"))
+
+_dependencies = sorted(setup_cfg_options["install_requires"].split("\n")[1:])
 
 _optional_dependencies = sorted(
-    ["ctapipe_resources", "pytest", "graphviz", "matplotlib"]
+    setup_cfg_options_extras["tests"].split("\n")[1:]
+    + setup_cfg_options_extras["docs"].split("\n")[1:]
 )
+
+
+def get_package_name_setuptools(name):
+    """Extract a package name from setuptools syntax.
+
+    Parameters
+    ----------
+    name: str
+        Name of the package as specified in files like
+        ``setup.cfg``, ``setup.py`` or ``pyproject.toml``.
+
+    Returns
+    -------
+    module_name: str
+        Name of package without version specification.
+    """
+
+    version_delimeters = [" ", "=", "~", ">", "<"]
+    if any([i in name for i in version_delimeters]):
+        package_name = split(r"|".join(version_delimeters), name)[0]
+    else:
+        package_name = name
+
+    return package_name
 
 
 def main(args=None):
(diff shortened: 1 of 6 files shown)


HealthyPear:
It seems that the online tests fail to find the `option` part of `setup.cfg` or the file itself.

Locally it works as expected.

maxnoe:
The tests are run on the installed ctapipe, which doesn't have a setup.cfg

HealthyPear:
> The tests are run on the installed ctapipe, which doesn't have a setup.cfg

Dammit...

And I guess this
https://github.com/pypa/packaging-problems/issues/215
kills the possibility to use e.g `importlib.metadata` to access this information from user-like installations, am I right?

Can't we run the tests on the development version and in a separate job just test that the user installation simply doesn't crash? The code to test is the same.

HealthyPear:
It seems that there is at least `importlib.metadata.requires`
`

kosack:
You can't read _setup.cfg_ in `ctapipe-info`, since that file is outside of the package.    But perhaps there is a way to do something similar to how we do the version string: there a file gets generated during the install process that becomes part of the package.  So in the install, you could generate a "dependencies" list and then that would be available in the package. 

The main issue I have with that is that `ctapipe-info` should also list *optional* dependencies, since that info is very useful to know for debugging.   But I suppose you could parse `options.extras_require `, but then you might get lots of random things about the documentation building, and not generally optional run-time packages like bokeh (which now is required, but should be optional)

maxnoe:
In general, any functionality like this should use the new importlib.metadata (importlib_metadata before 3.9) module prodiving such information independently of the setup.cfg / pyproject.toml or whatever packaging system is used.

HealthyPear:
I used `importlib.metadata` to extract the information about the complete set of packages and I added the splitting of optional packages based on the extras provided by the package metadata.

The internal approach is somewhat more brute-force than I wished: it appears that from Python3.10 that library allows using JSON to explore the metadata and simplifies it a bit, but since we test from 3.9 I had to keep things compatible with the older API.

The unit-tests I added based on [pytest-console-scripts](https://github.com/kvas-it/pytest-console-scripts) now work as intended for some reason.

HealthyPear:
On a side note: shouldn't `ctapipe-info --resources` also show the existing cached resources (i.e. `~/.cache/ctapipe/....`)?

kosack (approved):

kosack:
> On a side note: shouldn't ctapipe-info --resources also show the existing cached resources (i.e. ~/.cache/ctapipe/....)?

It does:

```sh
> ctapipe-info --resources  
*** ctapipe resources ***

CTAPIPE_SVC_PATH: (directories where resources are searched)
	 no path is set

RESOURCE NAME                  : LOCATION
----------------------------------------------------------------------
ASTRICam.camgeom.fits.gz       : ~/.cache/ctapipe/cccta-dataser
CHEC.camgeom.fits.gz           : ~/.cache/ctapipe/cccta-dataser
DigiCam.camgeom.fits.gz        : ~/.cache/ctapipe/cccta-dataser
FACT.camgeom.fits.gz           : ~/.cache/ctapipe/cccta-dataser
```

ctapipe/tools/info.py
```diff
@@ -160,17 +147,32 @@ def _info_tools():
...
+    optional_dependencies = {extra: [] for extra in extras}
 
-    for name in _optional_dependencies:
-        version = get_module_version(name)
-        print(f"{name:>20s} -- {version}")
+    required_dependencies = []
+    for package in all_dependencies:
+        if "extra" in package:
+            for extra in extras:
+                if extra in package:
```
maxnoe, line 162: there is the same if twice here

maxnoe (changes requested):

ctapipe/tools/info.py
```diff
@@ -160,17 +147,32 @@ def _info_tools():
 
 def _info_dependencies():
     """Print info about dependencies."""
-    print("\n*** ctapipe core dependencies ***\n")
 
-    for name in _dependencies:
-        version = get_module_version(name)
-        print(f"{name:>20s} -- {version}")
+    meta = metadata("ctapipe")
+    extras = [v for k, v in meta.items() if k == "Provides-Extra"]
```
maxnoe, line 152: This can be:
```
In [10]: meta.get_all("Provides-Extra")
Out[10]: ['all', 'dev', 'docs', 'tests']
```

PR #2304, opened 2023-04-05
cep 2: event structure

maxnoe:


kosack (commented):
Could be useful to mention what this change would affect: namely all code that uses ctapipe's event-level API would need to be updated. Perhaps giving an example (or a few common examples) of what needs to be done to upgrade that software, e.g.:
```py3
# The orignal:
event.dl0.tel[15].waveform

# becomes
event.tel[15].dl0.waveform
```

kosack:
Related note: I wonder if we could provide a [_rope_](https://github.com/python-rope/rope) or [_bowler_](https://github.com/facebookincubator/Bowler) rule to automatically refactor code (as a first pass?)

docs/development/ceps/index.rst
Tobychev, line 1: We should probably not accept changes to the the index page in regular CEP submissions...

maxnoe: How else am I supposed to include the cep in the rendered output?

Tobychev: Oh yeah, forgot we had to disable the glob if the directory was empty... well that's annoying : / but I shan't object further.

docs/development/ceps/proposed/cep-002.rst
Tobychev, line 1: I'm missing more concrete explanation and in particular real examples of what the current problem is, and likewise a concrete example of how the code will look with the proposed solution. 

For example, from what Karl commented it sounds like all `tel[tel_id]` uses would be impacted, currently there are 309 lines of those in the codebase (tests and docs included): will all, most, or only some of these be affected? Will the new proposed change mean that most of these  calls become redundant?

docs/development/ceps/proposed/cep-002.rst
```diff
@@ -0,0 +1,95 @@
...
+Currently, the hierarchy of the ctapipe ``ArrayEvent`` container has the data
+levels first and then each data level has a ``Map`` ``tel`` for telescope specific
+information in subcontainers.
+This CEP proposes to change this structure to have a container ``TelescopeEvent``
+as parent for all telescope-wise information that then contains the data levels.
+
+Proposed new structure
+======================
+
+This CEP proposes to change the current layout of ``ArrayEvent`` from having multiple
```
Tobychev, line 27: Do you mean `ArrayEventContainer`?

docs/development/ceps/proposed/cep-002.rst
```diff
@@ -0,0 +1,95 @@
...
+
+The main structure after the change will look like this:
+
+.. code-block::
+
+   SubarrayEvent
+   - index
+   - trigger
+   - simulation
+   - dl0
```
Tobychev, line 44: What are these "dlX" things? For example the current `DL0Container` looks to mostly just have a telescope-wise map of `DL0CameraContainer`s, but the proposal is to put every telescope-wise thing into a telescope container, right? 

So does the fact that you wrote dlX in both the SubarrayEvent and the TelescopeEvent mean that the current containers are going to be split somehow?

maxnoe: The `DLXContainer` s would lose the `.tel` attribute and only store array wide information. The current *CameraContainers will be renamed to *TelescopeContainer and be subcontainers of the `TelescopeEventContainer`.

If there is no array wide information (currently) at a specific data level, this container would no longer exist. But that is also connected to the issue that we currently have some information not sorted into data levels that should be (e.g. trigger should probably be in DL0 and then there is DL0 / subarray / trigger

RuneDominik:
Looking at the proposed structure the muon information is missing. In the current stucture it is listed on the same structure-level as the data-levels. What would be the new structure for it?

ccossou:
I have a few comment/questions:
* dl3 combine multiple telescopes, so that datalevel probably can't be at the same level in the data structure than others. If that's true, it would be weird to have dlX not at the same place in the structure. 
* in the even of an observation block being too big, and if we have to implement a segmented datastructure to store all DL0 information on disk in multiple sub-files, would the old or new structure be better shaped to handle such an evolution?

My uneducated self would tend to prefer the following structure:
`dl1.telX.event.waveform`

Meaning that the array of event would be directly packed as data level then telescope when it make sense, or for dl3, only have `dl3.event`. 

I don't know if from a code point of view this make sense, but that's how I'd do it.

maxnoe:
>    dl3 combine multiple telescopes, so that datalevel probably can't be at the same level in the data structure than others. If that's true, it would be weird to have dlX not at the same place in the structure.

Yes, we probably won't have `TelescopeEventContainer.dl3` , but only `ArrayEventContainer.dl3`

>    in the event of an observation block being too big, and if we have to implement a segmented datastructure to store all DL0 information on disk in multiple sub-files, would the old or new structure be better shaped to handle such an evolution?

> My uneducated guess would tend to prefer the following structure: dl1.telX.event.waveform

That's what we have now and this CEP is proposing to change that.

I will expand the text to better explain, but I think it's the opposite: we will have DL0 event files per telescope, so we might want to process DL0 to DL1 for telescopes individually, which will be much more natural if we have multiple data levels inside one `TelescopeEvent` (to perform the dl0 to dl1 transformation on one telescope) than having to create `ArrayEvents` for each `Telescope` with only one telescope in each of the data levels.

It is also much more natural then to merge multiple `TelescopeEvent` s into one `ArrayEvent`.

By the way: this is also how sim_telarray output is organized:

```
ArrayEvent[2010](event_id=9610)
     TriggerInformation[2009](event_id=9610)
     TelescopeEvent[2225](telescope_id=25)
         TelescopeEventHeader[2011](telescope_id=25)
         ADCSamples[2013](telescope_id=25)
         PixelTiming[2016](telescope_id=25)
         ImageParameters[2014](telescope_id=25)
         PixelList[2027](telescope_id=25, code=1, kind=selected)
     TelescopeEvent[3225](telescope_id=125)
         TelescopeEventHeader[2011](telescope_id=125)
         ADCSamples[2013](telescope_id=125)
         PixelTiming[2016](telescope_id=125)
         ImageParameters[2014](telescope_id=125)
         PixelList[2027](telescope_id=125, code=1, kind=selected)
     TelescopeEvent[3230](telescope_id=130)
         TelescopeEventHeader[2011](telescope_id=130)
         ADCSamples[2013](telescope_id=130)
         PixelTiming[2016](telescope_id=130)
         ImageParameters[2014](telescope_id=130)
         PixelList[2027](telescope_id=130, code=0, kind=triggered)
         PixelList[2027](telescope_id=130, code=1, kind=selected)
```

kosack:
> Looking at the proposed structure the muon information is missing. In the current stucture it is listed on the same structure-level as the data-levels. What would be the new structure for it?

Muon info are just DL1 parameters, so would appear under DL1. However, we might need to have DL1 split by a Map of strings similar to DL2, so that we can support multiple images/cleanings

maxnoe:
I updated the cep with a couple of before / after code examples and some more text

docs/development/ceps/proposed/cep-002.rst
```diff
@@ -0,0 +1,194 @@
...
+E.g. in the ``CameraCalibrator``, ``ImageProcessor`` and more classes,
+we currently have to provide the ``ArrayEventContainer``,
+so that higher data levels can be filled from lower data levels, although only one telescope
+is processed at a time.
+
+In the current scheme (simplified from ctapipe-process) it looks like this:
+
+.. code-block:: python
+
+   for array_event in source:
```
Tobychev, line 87: I think it is worth lifting up the hidden loops and writing them out explicitly in this example to make the example more clear, even if it means that the depiction of the event loop is less accurate.

docs/development/ceps/proposed/cep-002.rst
```diff
@@ -0,0 +1,194 @@
...
+
+   SubarrayEvent
+   - index: SubarrayEventIndex
+   - trigger: SubarrayTrigger
+   - simulation: SimulatedShower
+   - dl0: DL0Subarray
+   - dl1: DL1Subarray
+   - dl2: DL2Subarray
+   - dl3: DL3Subarray
+   - tel: Map[int, TelescopeEvent] # or just a list? TelescopeEvent stores the tel_id
```
Tobychev, line 56: This question mark seems like the only remaining issue to decide on. I'm not sure what the idea with the map is, maybe you could describe what options you are envisioning?

docs/development/ceps/proposed/cep-002.rst
Tobychev, line 1: Overall I think this looks pretty clear, and the reasons for carrying out these changes are sound. 

So after resolving the two other minor comments I recommend we change the status to "Accepted" and start working on the change.  Then we record in this CEP some typical changes that are needed to refactor the code to fit the new structure, and every gnarly special case along with its solution. 

If nothing fatal comes up during this work we then eventually merge the changed and change the status to "Implemented". Because these changes can be expected to break a lot of downstream code in the production chains, we should give them ample warning before finally merging the feature branch,

kosack (dismissed):
Looks good with the new examples added.

kosack:
> "Accepted" and start working on the change. Then we record in this CEP some typical changes that are needed to refactor the code to fit the new structure, and every gnarly special case along with its solution.

One question is when to involve other stakeholders, i.e. the LST, NectarCam, CHEC/SSTCam, (and maybe MAGIC?) teams that are using ctapipe already.  Should we ask for comments or just go ahead? At the very least, we might want to demo it at a weekly meeting and invite them to join

docs/development/ceps/proposed/cep-002.rst
maxnoe: done

docs/development/ceps/proposed/cep-002.rst
maxnoe: No, I think also having the map is handy

kosack:
Oh, one thing I just realized is not mentioned here that could be useful is the canonical way to determine if a telescope has information at a particular data level.   Now that there is only one tel_id map, just looking at `tel.keys()` is not sufficient, so it would be good to show how it should be done.

In the example:
```py3
hillas_dicts = {
    tel_id: dl1.parameters.hillas
    for tel_id, dl1 in event.dl1.items()
}

#After:

hillas_dicts = {
    tel_id: tel_event.dl1.parameters.hillas
    for tel_id, tel_event in event.tel.items()
}
```
the case where there is no `dl1` object  or no `tel_event.dl1.parameters` for a given tel isn't explained.

maxnoe:
How we work at the moment is that all telescopes get all data levels at the same time and we fill invalid values for telescopes that do fulfil some criteria (e.g. we will have the image parameters, but they might be nan if not enough pixels survived the cleaning).

So the situation you mentioned, that there are dl1 containers for some telescopes but not others cannot really happen.

The full loop in the hillas reconstructors looks like this now:

```
        hillas_dict = {
            tel_id: dl1.parameters.hillas
            for tel_id, dl1 in event.dl1.tel.items()
            if all(self.quality_query(parameters=dl1.parameters))
        }
```

and will look like this after the change:


```
        hillas_dict = {
            tel_id: tel_event.dl1.parameters.hillas
            for tel_id, tel_event in event.tel.items()
            if all(self.quality_query(parameters=tel_event.dl1.parameters))
        }
```

kosack (approved):

docs/development/ceps/proposed/cep-002.rst
```diff
@@ -0,0 +1,204 @@
...
+and the ``ImageProcessor`` here, although both of these do not access any subarray-wide data at all.
+
+Using the new structure, these classes will get a single ``TelescopeEventContainer`` and the loop
+can to be moved outside those classes to a single place:
+
+.. code-block:: python
+
+   for array_event in source:
+       for telescope_event in array_event.tel.values():
+           calibrator(telescope_event)
```
Tobychev, line 112: Do I understand it correctly if I interpret the example code as saying the calibrator in the new structure does not need either of the two loops present in the old example? 

Eliminating _two_ loops over all events seems like a signification savings, and it is worth spelling it out properly in the text.

maxnoe: Nothing here changes the amount of work done. Since the calibration is quite work-intensive, I doubt it matters for performance.

maxnoe: Nothing here changes the amount of work done, the difference will only be the python overhead of the loops.

Since the calibration is quite work-intensive, I doubt it matters for performance.

maxnoe: Essentially we change from:


```
for array_event in source:
    for tel_event in array_event:
        step1(tel_event)
    for tel_event in array_event:
        step2(tel_event)
```


to


```
for array_event in source:
    for tel_event in array_event:
        step1(tel_event)
        step2(tel_event)
```


Ignoring where the loops are (inside or outside of `CameraCalibrator`.

Tobychev:
To restate what I said in inline comment (in violation of CEP 1!), I think this CEP looks pretty clear, and the reasons for carrying out these changes are sound. 

> One question is when to involve other stakeholders, i.e. the LST, NectarCam, CHEC/SSTCam, (and maybe MAGIC?) teams that are using ctapipe already. Should we ask for comments or just go ahead? At the very least, we might want to demo it at a weekly meeting and invite them to join

I just re-read CEP 1, and one thing I think we've forgotten is announcing this CEP according to the workflow we defined, in particular I think this haven't happened:

> an announcement with a link to the pull request should be sent both to the ctapipe mailing list and the ctapipe coordinator list.

so some sort of announcement that will reach more people is already mandated. Separately, I think prodding those "external" stakeholders to leave a review (or at least proof -of-skimming) is valuable, but I'll let @kosack judge if a mailing list announcement is a firm enough poke to get some response.

maxnoe:
I made one last change (removing `trigger` from the top-level of both SubarrayEvent and TelescopeEvent.

This is related to #2374, trigger info should be part of DL0.

Tobychev (dismissed):

kosack (dismissed):

maxnoe:
I started implementing a zfits event source following the current draft of the ACADA DPPS ICD, in part as preparation for the ACADA LST integration tests on La Palma.

On thing I realized is that with this proposal here, it would be possible to implement subarray event sources and telescope event sources.

The latter would be useful for reading DL0 telescope event data and processing it up to dl1 in parallel, where we don't need stereo information and to read the anticipated separate streams for shower, calibration and muon events.

E.g. calib pipe would probably only ever use a `DL0TelescopeEventSource` on calibration and muon stream files and dl1 could processing could also run using that on the shower stream.

A possible `SubarrayEventSource` would then read the single files and merge the separate telescope data into the `SubarrayEvent` structure.

kosack:
Will merge into "proposed" and we will open a new PR moving it to "Accepted" for external review

Tobychev (dismissed):

Tobychev (approved):

Tobychev (approved):

maxnoe:
Ok, merging now to have it under proposed rendered in the docs so we can collect comments.

PR #2305, opened 2023-04-06
YAIB: Yet another ImPACT branch

Tobychev:
As requested by @kosack I open a PR with my version of @ParsonsRD ctapipe ImPACT implementation, where I've introduced changes to make the code able to merge with the current HEAD. 

This branch also has a modification to some reco tests, and in particular `test_reconstruction_changes.py` expects that you have a local copy of `LSTCam.template.gz` that I've generated and is not on the testserver yet. This file, along with the notebook I used to generate it based on a much larger template provided by @ParsonsRD is found [here](https://filesender.renater.fr/?s=download&token=85267e59-cafc-4e9c-9767-3bcf8c0eecce)


gschwefer:
I added an implementation of the energy seed. Since this is not strictly necessary to run ImPACT, it is a bit different from the geometry seed. Also, the telescope impact distances were not properly saved in case the reconstruction ran through, so I added the line for that.

gschwefer:
After discussion with @ParsonsRD , I removed the option for energy preminimization in case no energy seed is present for a given event. An energy seed is now required to run ImPACT.

gschwefer:
This update mostly includes the changes and fixes currently under discussion in #2388 adapted for the ImPACT pixel likelihood interface. As an aside, keeping constants in the likelihood can actually be important here, namely when likelihoods from the image and time gradient fits are added to get the correct relative statistical weights.

ctapipe/reco/impact.py
```diff
@@ -85,51 +88,70 @@ class ImPACTReconstructor(Component):
 
     """
 
+    minimiser = traits.CaselessStrEnum(
+        ["minuit", "l-BFGS"],
+        default_value="minuit",
+        help="name minimiser to use in the fit",
+    ).tag(config=True)
+
+    atmosphere_profile_name = traits.CaselessStrEnum(
```
kosack, line 97: This should be replaced with the atmosphere profile provided by the EventSource. That will ensure it cannot be out of sync.

ctapipe/reco/impact.py
```diff
@@ -726,53 +878,81 @@ def predict(self, shower_seed, energy_seed):
...
-            is_valid=True,
-            alt_uncert=np.nan,
-            az_uncert=np.nan,
-            core_uncert=np.nan,
-            # Copy reconstructed hmax
-            h_max=h_max,
-            h_max_uncert=errors[5] * h_max,
-            goodness_of_fit=like,
-            prefix=self.__class__.__name__,
+        h_max = self.altitude_profile(x_max)
```
kosack, line 921: I think this is likely incorrect: the old atmosphere profile functions only worked for vertical heights above sea level, and not distances thought the atmosphere. In other words, this only works correctly at zenith=0 degrees.   This should be replaced with the conversion functions provided by `ctapipe.atmosphere.AtmosphereDenistyProfile`.   However, currently we only have a `line_of_sight_integral(distance, zenith_angle)` function that converts *h_max* to *x_max*, so we will need to implement the inverse transform as well.  I'll open an issue and try to fix that. 

However, I notice that here x_max is defined in terms of cos(z), so it's a hight, not a distance thought the atmosphere.  We should also check elsewhere that we are consistent in it's definition: is it supposed to be the atmospheric depth along the shower's path (slant depth), or vertically to the shower-max (atmosphere depth)? I would think the former is more useful as it is the actual amount of atmosphere traversed by the shower's progenitor.  Or is the latter what the model expects?   In HillasReconstructor, I think we reconstruct the slant depth.

ctapipe/reco/impact.py
```diff
@@ -783,232 +963,68 @@ def minimise(self, params, step, limits, minimiser_name="minuit", max_calls=0):
...
-                centre_y + (diff_y * shift),
-                centre_tel_x + (diff_tel_x * shift),
-                centre_tel_y + (diff_tel_y * shift),
-                energy,
-            )
+        energy = params[4]
+        xmax_scale = 1
+
+        # Now do the minimisation proper
+        self.min = Minuit(
```
kosack, line 976: rename `self.min` → `self.minimizer`  to be more clear

kosack (changes requested):

ctapipe/reco/impact.py
maxnoe: Why assign to self at all?

ctapipe/reco/impact.py
kosack: I would think so as well.  But clearly here the hight only is being reconstructed.  Of course you can always compute the distance from the hight and core position, but I think it's better that we always assume h_max/x_max are slant depths.  

@gschwefer maybe check everywhere that x_max/h_max are used here that they are really using slant depth and not height, and fix places where they are not.  That could easily lead to poor behavior at larger zenith angles.  Maybe we could also add a unit test to reconstruct a dummy shower at small and large zenith angle (say 10º, 60°) and see if the results both come out correctly.

ctapipe/reco/impact.py
kosack: good point, it's not used elsewhere. I guess it was done that way for interactive debugging purposes or something, but clearly not needed.

ctapipe/reco/impact.py
ParsonsRD: Whenever x_max is used here it should be the slant depth at it is the only physically meaningful thing to use. I think this is just an error and instead we should use:
` h_max = self.altitude_profile(x_max*cos(zenith_angle))`
But also do we really want to store h_max in our output and not x_max?

kosack: > But also do we really want to store h_max in our output and not x_max?

We can store both. h_max is already there because that is what the HillasReconstructor reconstructs geometrically;  x_max requires applying the atmosphere profile, so it was not included by default.  But we should always have a quite good estimate of the current atmosphere density, so no good reason not to just reconstruct x_max always.

gschwefer: Ok, so I looked at this issue and as I see it everything in the branch is mathematically correct and consistent, though the naming is not as clear as it could be. Impact indeed intrinsically reconstructs slant depth. The value of h_max that is returned by the class is just the vertical height of the shower maximum, obtained through the altitude_profile function, which as an input correctly gets cos(zenith)*slant_depth.
The get_shower_max function correctly returns the guess of the slant depth of the shower maximum, by geometrically calculating the height above the telescope in the tilted frame, converting this to vertical height over ground (and sea level), using the thickness_profile function to get the column density from this and then divides by cos(zenith) to get slant depth.

I've renamed variables now so x_max is exclusively used for slant depth, the vertical depth is called column_density now. Any suggestions on what to name the h_max or is the name fine?

ctapipe/reco/impact.py
```diff
@@ -783,232 +963,68 @@ def minimise(self, params, step, limits, minimiser_name="minuit", max_calls=0):
...
+            goodness_of_fit=False,
+        )
+        # This time leave everything free
+        self.min.fixed = [False, False, False, False, False, False, True]
+        self.min.errors = step
+        self.min.limits = limits
+        self.min.errordef = 1.0
+
+        # Tighter fit tolerances
+        self.min.tol *= 1000
```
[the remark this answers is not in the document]
gschwefer: Done

ctapipe/containers.py
```diff
@@ -399,6 +399,21 @@ class ImageParametersContainer(Container):
...
+    """
+    Concentrations are ratios between light amount
+    in certain areas of the image and the full image.
+    """
+
+    default_prefix = "ImPACTLikelihood"
+    log_likelihood = Field(
+        nan, "Per telescope likelihood of the ImPACT best fit location"
+    )
+    impact_distance = Field(
```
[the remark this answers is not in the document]
gschwefer: I have a question here: Because both the `ReconstructedEnergyContainer` and `ReconstructedGeometryContainer` have the fields `is_valid`, `goodness_of_fit` and `telescopes`, we get them twice as `_1` and `_2` respectively. There is no difference in these fields between energy and geometry reconstruction, so should one just be left empty or should they both just give the same values?

ctapipe/reco/impact.py
kosack: `column_density` is actually just another name for slant depth - it implies a line-of-site integration, so that is somewhat confusing; I think from the previous discussion about this in another issue that the correct term is  "mass overburden". However,  I would avoid renaming or redefining anything in this pull request.  We can do that in a separate PR where we are careful to be consistent. 

Here, just keep with the definition of x_max and h_max as both representing the height/mass-overburden, and avoid using slant depth anywhere.

gschwefer: Ok, I agree `column_density` is maybe not the ideal name. I also should have been clearer above: I only renamed internal variables used inside the `impact.py` file because the names were internally inconsistent, I didn't change anything globally or in the containers.

ctapipe/containers.py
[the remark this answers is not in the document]
Tobychev: I'd sat that if you reconstruct it, you should set the appropriate values for its  container.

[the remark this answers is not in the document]
kosack: yes, it's ok that they are the same for this reconstructor.

ctapipe/reco/impact.py
kosack: In astrophysics it's the line-of-sight integral through a medium, for example in X-ray analysis through molecular clouds

ctapipe/containers.py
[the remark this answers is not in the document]
gschwefer: All fields that appear twice are now also filled twice

docs/changes/2305.feature.rst
```diff
@@ -0,0 +1,2 @@
+Implements the Image Pixel-wise fit for Atmospheric Cherenkov Telescopes reconstruction method
```
maxnoe, line 1: Is this really true?

Shouldn't this be more like a bugfix release like:

> Fix the implementation of ImPACT 

?

gschwefer: I'd say "Large overhaul of the implementation of ImPACT", but you're right in principle

Tobychev: Good point, should have used what @ParsonsRD put in his PR #2052 that I based this PR on, I updated with some of the notes from there. Left some stuff out because I think it is no longer relevant.

maxnoe:
Some minor issue in the lint, please fix

docs/changes/2305.feature.rst
```diff
@@ -0,0 +1,8 @@
+Large updates to the Image Pixel-wise fit for Atmospheric Cherenkov Telescopes reconstruction method (https://doi.org/10.48550/arXiv.1403.2993)
+
+* ImPACT - General code clean up and optimisation. Now updated to work similarly to other reconstructors using the standardised interface, such that it can be used ctapipe-process. Significant improvements to tests too
+* ImPACT_utilities - Created new file to hold general usage functions, numba used in some areas for speedup
+* template_network_interpolator - Now works with templates with different zenith and azimuth angles
+* unstructured_interpolator - Significant speed improvements
+* pixel_likelihood - Constants added back to neg_log_likelihood_approx, these are quite important to obtaining a well normalised goodness of fit.
+* hillas_intersection - Fixed bug in core position being incorrectly calculated, fixed tests too
```
maxnoe, line 8: Could you add a sentence about `atmosphere_profile` now being an argument to `Reconstructor()`?

src/ctapipe/reco/impact.py
```diff
@@ -2,68 +2,73 @@
 """
 
 """
```
maxnoe, line 4: Please add a docstring here

src/ctapipe/reco/impact.py
```diff
@@ -85,51 +90,59 @@ class ImPACTReconstructor(Component):
 
     """
 
+    minimiser = traits.CaselessStrEnum(
+        ["minuit", "l-BFGS"],
```
maxnoe, line 94: Any point in supporting the scipy minimize?

src/ctapipe/reco/impact.py
```diff
@@ -85,51 +90,59 @@ class ImPACTReconstructor(Component):
...
+        default_value="minuit",
+        help="name minimiser to use in the fit",
+    ).tag(config=True)
+
+    use_time_gradient = traits.Bool(
+        default_value=False, help="Use time gradient in ImPACT reconstruction"
```
maxnoe, line 100: Does this option have any additional requirements on the templates? I.e. do they need to contain time information?

src/ctapipe/reco/impact.py
```diff
@@ -85,51 +90,59 @@ class ImPACTReconstructor(Component):
...
+    ).tag(config=True)
+
+    root_dir = traits.Unicode(
```
maxnoe, line 103: Better name `template_directory`?

src/ctapipe/reco/impact.py
```diff
@@ -85,51 +90,59 @@ class ImPACTReconstructor(Component):
...
+        default_value=".", help="Directory containing ImPACT tables"
+    ).tag(config=True)
+
     # For likelihood calculation we need the with of the
     # pedestal distribution for each pixel
     # currently this is not available from the calibration,
     # so for now lets hard code it in a dict
     ped_table = {
```
maxnoe, line 111: This should at least be a `FloatTelescopeParameter`

src/ctapipe/reco/impact.py
```diff
@@ -85,51 +90,59 @@ class ImPACTReconstructor(Component):
...
 
     def __init__(
-        self,
-        root_dir=".",
-        minimiser="minuit",
-        prior="",
-        template_scale=1.0,
-        xmax_offset=0,
-        use_time_gradient=False,
+        self, subarray, atmosphere_profile, dummy_reconstructor=False, **kwargs
```
maxnoe, line 126: What is dummy_reconstructor?

src/ctapipe/reco/impact.py
```diff
@@ -85,51 +90,59 @@ class ImPACTReconstructor(Component):
...
-        # We also need a conversion function from height above ground to
-        # depth of maximum To do this we need the conversion table from CORSIKA
-        (
-            self.thickness_profile,
-            self.altitude_profile,
-        ) = get_atmosphere_profile_functions("paranal", with_units=False)
+        # self.priors = prior
+
+        # String templates for loading ImPACT templates
+        self.amplitude_template = Template("${base}/${camera}.template.gz")
```
maxnoe, line 144: Is templating only by camera name what we want? Shouldn't we at least use the full `str(telescope_description)`? 

E.g. FlashCam on MST vs. FlashCam on HESS II

src/ctapipe/reco/impact.py
gschwefer: No, I think that can be removed

src/ctapipe/reco/impact.py
```diff
@@ -150,13 +163,90 @@ def __init__(
...
+        # Finally get the telescope images and and the selection masks
+        mask_dict, image_dict, time_dict = {}, {}, {}
+        for tel_id in hillas_dict.keys():
+            image = event.dl1.tel[tel_id].image
+            image_dict[tel_id] = image
+            time_dict[tel_id] = event.dl1.tel[tel_id].peak_time
+            mask = event.dl1.tel[tel_id].image_mask
+
+            # Dilate the images around the original cleaning to help the fit
+            mask = dilate(self.subarray.tel[tel_id].camera.geometry, mask)
```
maxnoe, line 207:
```
for _ in range(3):
    mask = dilate(self.subarray.tel[tel_id].camera.geometry, mask)
```

src/ctapipe/reco/impact.py
```diff
@@ -150,13 +163,90 @@ def __init__(
...
+            mask_dict[tel_id] = mask
+
+        # Next, we look for geometry and energy seeds from previously applied reconstructors.
+        # Both need to be present at elast once for ImPACT to run.
+
+        reconstructor_geometry_prediction = event.dl2.stereo.geometry
+
+        valid_seed = False
+        for reco in reconstructor_geometry_prediction:
```
maxnoe, line 218:
```
for reconstructor, geometry in `reconstructor_geometry_prediction.items():
    if geometry.is_valud:
        valid_seed = True
        breal
```

src/ctapipe/reco/impact.py
```diff
@@ -150,13 +163,90 @@ def __init__(
...
+            if reconstructor_geometry_prediction[reco].is_valid:
+                valid_seed = True
+
+        reconstructor_energy_prediction = event.dl2.stereo.energy
```
maxnoe, line 222: variable name is longer than looking it up...

src/ctapipe/reco/impact.py
```diff
@@ -150,13 +163,90 @@ def __init__(
...
+
+        valid_energy_seed = False
+        for reco in reconstructor_energy_prediction:
```
maxnoe, line 225: same as above

src/ctapipe/reco/impact.py
```diff
@@ -150,13 +163,90 @@ def __init__(
...
+            mask = dilate(self.subarray.tel[tel_id].camera.geometry, mask)
+            mask = dilate(self.subarray.tel[tel_id].camera.geometry, mask)
+            mask_dict[tel_id] = mask
+
+        # Next, we look for geometry and energy seeds from previously applied reconstructors.
+        # Both need to be present at elast once for ImPACT to run.
+
+        reconstructor_geometry_prediction = event.dl2.stereo.geometry
+
+        valid_seed = False
```
maxnoe, line 217: valid_geometry_seed?

src/ctapipe/reco/impact.py
gschwefer: Yes. Maybe even removing the hard coded names for the templates is worth a look

src/ctapipe/reco/impact.py
```diff
@@ -238,76 +348,46 @@ def get_shower_max(self, source_x, source_y, core_x, core_y, zen):
...
-        mean_height += 2150
+        mean_height_asl = (
+            mean_height_above_ground
+            + self.subarray.reference_location.height.to_value(u.m)
+        )
 
-        if mean_height > 100000 or np.isnan(mean_height):
-            mean_height = 100000
+        if mean_height_asl > 100000 or np.isnan(mean_height_asl):
+            mean_height_asl = 100000
```
maxnoe, line 379: why replace nan or nonsensical values with a valid looking value? That might be misleading. Also a magic number.

src/ctapipe/reco/impact.py
gschwefer: I think it's relevant for a couple of tests

maxnoe (approved):
I'll approve here besides the inline comments I make.

I think it will be much easier to address these comments in follow up, smaller PRs one by one.

src/ctapipe/reco/impact.py
maxnoe: That could be a `TelescopeParameter(Path)`

src/ctapipe/reco/impact.py
gschwefer: Yes

src/ctapipe/reco/impact.py
```diff
@@ -390,122 +472,143 @@ def get_likelihood(
...
-                weight.sum(axis=1) * sxy.sum(axis=1) - sx.sum(axis=1) * sy.sum(axis=1)
-            ) / d
-            time_fit /= -1 * (180 / math.pi)
-            chi2 = -2 * np.log(
-                rv.pdf((time_fit - time_gradients.T[0]) / time_gradients.T[1])
-            )
+            time_gradients_uncertainty[time_gradients_uncertainty == 0] = 1e-6
+
+            chi2 = 0
+            for telescope_index in range(self.time.shape[0]):
```
maxnoe, line 558: `for time_mask, image in zip(...)` to avoid the repeated by index lookups.

src/ctapipe/reco/impact_utilities.py
```diff
@@ -0,0 +1,197 @@
...
+    -------
+    tuple of seed, steps size and fit limits
+    """
+    lower_en_limit = energy * 0.1
+    en_seed = energy
+
+    # If our energy estimate falls outside of the range of our templates set it to
+    # the edge
+    if lower_en_limit < 0.02:
+        lower_en_limit = 0.02
```
maxnoe, line 111: Shouldn't this be somehow stored in the template instead of being hardcoded?

src/ctapipe/reco/impact_utilities.py
```diff
@@ -0,0 +1,197 @@
...
+        )
+        for dist in dist_range:
+            for xmax in xmax_range:
+                key = (0, 0, en, dist, xmax)
+                template, x, y = generate_fake_template(-1.5, 0.5)
+                template *= scale
+
+                template_dict[key] = template.T * pe
+
+    filehandler = gzip.open(output_file, "wb")
```
maxnoe, line 195: use a context manager instead of `open()` followed by `close`

src/ctapipe/reco/impact.py
kosack: Still missing a docstring. Note that this appears in the documentation, so it should at least say what this module is about (as simple  as:
```py
"""Implementation of the ImPACT reconstruction algorithm
"""
```

src/ctapipe/reco/impact.py
```diff
@@ -2,68 +2,73 @@
...
+INVALID_ENERGY = ReconstructedEnergyContainer(
+    prefix="ImPACTReconstructor",
+    telescopes=[],
+)
 
-def xmax_prior(energy, xmax, width=100):
-    x_max_exp = guess_shower_depth(energy)
-    diff = xmax - x_max_exp
-    return -2 * np.log(norm.pdf(diff / width))
+# These are settings for the iminuit minimizer
```
kosack, line 63: I assume these are all fixed settings, i.e. things you don't ever expect the user to adjust?

If that is not the case, then you should add them to `__all__` and change the comments to use the form `#: Comment` so they become documented.

src/ctapipe/reco/impact.py
kosack: It seems to be a bit misnamed, as it replaces the `TemplateNetworkInterpolator` with a dummy one with a dummy template, so maybe call it "no_template_mode" or "test_mode" or something like that.

src/ctapipe/reco/impact.py
```diff
@@ -85,51 +90,59 @@ class ImPACTReconstructor(Component):
...
-        minimiser="minuit",
-        prior="",
-        template_scale=1.0,
-        xmax_offset=0,
-        use_time_gradient=False,
+        self, subarray, atmosphere_profile, dummy_reconstructor=False, **kwargs
     ):
         """
         Create a new instance of ImPACTReconstructor
         """
```
kosack, line 130: Missing documentation for the parameters.

kosack (approved):

src/ctapipe/reco/impact.py
maxnoe: The docstring should be completely removed and the parameters of `__init__` documented in the class docstring

maxnoe:
Hi all,

I am going to merge this now, thanks to everyone who put in the hard work to make this function again! @gschwefer @Tobychev @ParsonsRD 

There are a lot of comments for improvements here, but these are very hard to review, so I will merge here and let's continue in easier to review follow-up PRs.

PR #2312, opened 2023-04-20
Fix activity start order

maxnoe:
By not starting the activity before calling initialize,
an extra activity was started only containing the config files
and the config files were not included in the provenance of the actual
tool run.


kosack (dismissed):

maxnoe:
Provenance is not thread safe... there is only a single `Provenance` global...

LukasNickel (approved):

kosack (approved):

PR #2308, opened 2023-04-18
Add possibility to compute multiplicities using `FeatureGenerator`

maxnoe:
Add `multiplicity` method to `SubarrayDescription` and use a `ChainMap` to allow additional information to be passed to the `FeatureGenerator` other than the table to allow calling `SubarrayDescription.multiplicity` in `FeatureGenerator`.

PR #2315, opened 2023-04-21
Irf tool

Tobychev:
Here is a first draft of an tool to generate irf-files from a set of DL2 files. 

It is primarily a port of the protopipe functionality, but I've tried to also incorporate changes from lstpipe and pyirf example scripts. In practice I think most of the differences is about the handling of offset binning. For expediency I followed the protopipe example on that.


ctapipe/irf/irf_classes.py
```diff
@@ -0,0 +1,404 @@
...
+    ).tag(config=True)
+
+    obs_time = Float(default_value=50.0, help="Observation time").tag(config=True)
+    obs_time_unit = Unicode(
+        default_value="hour",
+        help="Unit used to specify observation time as an astropy unit string.",
+    ).tag(config=True)
+
+    alpha = Float(
+        default_value=5.0, help="Ratio between size of on and off regions"
```
maxnoe, line 64: Alpha is usually size of on / size of off, and the default we use is 0.2, not 5.0, so larger off region than on region

ctapipe/irf/irf_classes.py
```diff
@@ -0,0 +1,404 @@
...
+        help="Prefix of the `_alt` and `_az` reco geometry columns",
+    ).tag(config=True)
+    gammaness_classifier = Unicode(
+        default_value="RandomForestClassifier",
+        help="Prefix of the classifier `_prediction` column",
+    ).tag(config=True)
+
+    preselect_criteria = List(
+        default_value=[
+            ("multiplicity 4", "np.count_nonzero(tels,axis=1) >= 4"),
```
maxnoe, line 97: Not yet in master, but see https://github.com/cta-observatory/ctapipe/pull/2308

ctapipe/irf/irf_classes.py
```diff
@@ -0,0 +1,404 @@
...
+            ("valid classifier", "valid_classer"),
+            ("valid geom reco", "valid_geom"),
+            ("valid energy reco", "valid_energy"),
+        ],
+        help=QualityQuery.quality_criteria.help,
+    ).tag(config=True)
+
+    rename_columns = List(
```
maxnoe, line 105: Why rename those if the actual column names could just be used in the above quality query?

ctapipe/tools/make_irf.py
```diff
@@ -0,0 +1,342 @@
...
+    "IRFDOC_PROTON_SPECTRUM": IRFDOC_PROTON_SPECTRUM,
+}
+
+
+class IrfTool(IrfToolBase, DataBinning):
+    name = "ctapipe-make-irfs"
+    description = "Tool to create IRF files in GAD format"
+
+    def make_derived_columns(self, events, spectrum, target_spectrum):
+        events["pointing_az"] = 0 * u.deg
```
maxnoe, line 57: This can be obtained nowadays from the observation block information (and is loaded if you pass `load_observation_info` to `TableLoader`.

ctapipe/irf/irf_classes.py
Tobychev: As these columns will be carried along for further giving them more general names makes it all a bit neater. But there's no real reason really, I think this is mostly a remnant of an older way to try to structure things for the config.

kosack (changes requested):

ctapipe/irf/irf_classes.py
```diff
@@ -0,0 +1,404 @@
...
+        help="How many subarray events to load at once for making predictions.",
+    ).tag(config=True)
+
+    output_path = traits.Path(
+        default_value=None,
+        allow_none=False,
+        directory_ok=False,
+        help="Output file",
+    ).tag(config=True)
+    output_file = Unicode(
```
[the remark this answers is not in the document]
kosack: note that a "path" here doesn't mean "directory", it is really the path to a file.

ctapipe/tools/make_irf.py
```diff
@@ -0,0 +1,342 @@
...
+        opts = dict(load_dl2=True, load_simulated=True, load_dl1_parameters=False)
+        reduced_events = dict()
+        for kind, file, target_spectrum in [
+            ("gamma", self.gamma_file, PYIRF_SPECTRA[self.gamma_sim_spectrum]),
+            ("proton", self.proton_file, PYIRF_SPECTRA[self.proton_sim_spectrum]),
+            ("electron", self.electron_file, PYIRF_SPECTRA[self.electron_sim_spectrum]),
+        ]:
+            with TableLoader(file, **opts) as load:
+                Provenance().add_input_file(file)
+                table = self._make_empty_table()
```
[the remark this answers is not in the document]
Tobychev: The purpose of `_make_empty_table` is to specify the expectation of what columns need to be present for downstream code, so that it is easier to add readers from other formats later and not depend on the structure of a particular test file.

ctapipe/irf/irf_classes.py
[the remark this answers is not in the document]
Tobychev: Left over since some earlier iteration I think, removed output file as suggested

ctapipe/irf/irf_classes.py
```diff
@@ -90,9 +92,10 @@ class IrfToolBase(Tool):
         help="Prefix of the classifier `_prediction` column",
     ).tag(config=True)
 
+class EventPreSelector(Component):
     preselect_criteria = List(
```
kosack, line 96: This should simply use a QualityQuery component, rather than specifying it separately here. That way if we fix bugs or change the way QualityQuery works, this code will still work.

kosack: In fact, you could just inherit from QualityQuery here and add the renaming options (or simply remove them as I don't think you really need to rename columns as long as you keep track of them nicely). Then there is no need for this component at all ,just replace it with a QualityQuery as we do elsewhere

kosack (changes requested):

ctapipe/irf/irf_classes.py
Tobychev: Do you have a particular example of inheriting from quality query in mind?

kosack: yes, you can see in for example ImageProcessor (ImageQualityQuery) for example,  or HillasReconstructor (StereoQualityQuery)

ctapipe/irf/irf_classes.py
```diff
@@ -115,6 +116,7 @@ class EventPreSelector(Component):
     )
 
     def _preselect_events(self, events):
+        tc = self.parent.tc
```
maxnoe: Why is here an explicit usage of the parent needed? We use this nowhere else and it prevents standalone usage (and testing) of this component.

ctapipe/tools/make_irf.py
```diff
@@ -102,7 +102,7 @@ def get_sim_info_and_spectrum(self, loader):
     def setup(self):
         self.tc = ToolConfig(parent=self)
```
maxnoe, line 103: Having a separate class for the config is not needed, you can just add the traits directly to the tool itself.

ctapipe/irf/irf_classes.py
```diff
@@ -0,0 +1,407 @@
...
+import astropy.units as u
+import numpy as np
+from astropy.table import QTable
+from pyirf.binning import create_bins_per_decade
+
+from ..core import Component, QualityQuery, traits
+from ..core.traits import Bool, Float, Integer, List, Unicode
+
+
+class ToolConfig(Component):
```
maxnoe, line 13: Tools can have their own config. There is now point in introducing a `Component` just to hold configuration for a tool. This also makes config files awkward, since you would have to write:

```yaml
IRFTool:
    ToolConfig:
        gamma_file: "path/to/gamma_file"
```

Instead of just:
```yaml
IRFTool:
    gamma_file: "path/to/gamma_file"
```

ctapipe/irf/irf_classes.py
```diff
@@ -0,0 +1,404 @@
...
+        default_value=None, directory_ok=False, help="Gamma input filename and path"
+    ).tag(config=True)
+    gamma_sim_spectrum = traits.Unicode(
+        default_value="CRAB_HEGRA",
+        help="Name of the pyrif spectra used for the simulated gamma spectrum",
+    ).tag(config=True)
+    proton_file = traits.Path(
+        default_value=None, directory_ok=False, help="Gamma input filename and path"
+    ).tag(config=True)
+    proton_sim_spectrum = traits.Unicode(
```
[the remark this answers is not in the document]
kosack: at minimum, use a `CaselessStrEnum` trait, but using a `UseEnum` one and a separately defined `Enum` class is probably even better.

ctapipe/irf/irf_classes.py
Tobychev: Changed to base on QualityQuery.

ctapipe/irf/irf_classes.py
```diff
@@ -0,0 +1,389 @@
...
+
+    alpha = Float(
+        default_value=0.2, help="Ratio between size of on and off regions"
+    ).tag(config=True)
+    ON_radius = Float(default_value=1.0, help="Radius of ON region in degrees").tag(
+        config=True
+    )
+
+    max_bg_radius = Float(
+        default_value=5.0, help="Radius used to calculate background rate in degrees"
```
maxnoe, line 69: This default is rather large

ctapipe/irf/irf_classes.py
```diff
@@ -0,0 +1,389 @@
...
+    ).tag(config=True)
+
+    max_gh_cut_efficiency = Float(
+        default_value=0.8, help="Maximum gamma purity requested"
```
maxnoe, line 73: It's not "purity", it's "efficiency" or "recall". I.e. the ratio of surviving true gammas.

ctapipe/irf/irf_classes.py
```diff
@@ -0,0 +1,389 @@
...
+            ("multiplicity 4", "subarray.multiplicity(tels_with_trigger) >= 4"),
+            ("valid classifier", "RandomForestClassifier_is_valid"),
+            ("valid geom reco", "HillasReconstructor_is_valid"),
+            ("valid energy reco", "RandomForestRegressor_is_valid"),
+        ],
+        help=QualityQuery.quality_criteria.help,
+    ).tag(config=True)
+
+    rename_columns = List(
+        help="List containing translation pairs new and old column names"
```
maxnoe, line 109: So it's a mapping? then why not use a `Dict(value_trait=Unicode, key_trait=Unicode)`?

ctapipe/irf/irf_classes.py
```diff
@@ -0,0 +1,389 @@
...
+            "reco_alt",
+            "gh_score",
+            "pointing_az",
+            "pointing_alt",
+            "theta",
+            "true_source_fov_offset",
+            "reco_source_fov_offset",
+            "weight",
+        ]
+        units = [
```
maxnoe, line 160: You can also provide units as a dictionary, seems less error prone:

```
units = {"true_energy": u.TeV, "reco_energy": u.TeV, ...}
```

ctapipe/irf/irf_classes.py
```diff
@@ -0,0 +1,389 @@
...
+            u.deg,
+            u.deg,
+            u.deg,
+            None,
+        ]
+
+        return QTable(names=columns, units=units)
+
+
+class ThetaSettings(Component):
```
maxnoe, line 181: I'd say this would also belong to the `CutOptimization` component when we introduce that.

ctapipe/irf/irf_classes.py
```diff
@@ -0,0 +1,389 @@
...
+    min_counts = Integer(
+        default_value=10,
+        help="Minimum number of events in a bin to attempt to find a cut value",
+    ).tag(config=True)
+    fill_value = Float(
+        default_value=0.32, help="Angular cut value used for bins with too few events"
+    ).tag(config=True)
+
+
+class DataBinning(Component):
```
maxnoe, line 198: I'd rather split this into classes per IRF. We also will need to support different parametrizations of the different IRFs, which will need their own configurations.

ctapipe/tools/make_irf.py
```diff
@@ -0,0 +1,367 @@
...
+                reduced_events[kind] = table
+
+        select_ON = reduced_events["gamma"]["theta"] <= self.tc.ON_radius * u.deg
+        self.signal = reduced_events["gamma"][select_ON]
+        self.background = vstack([reduced_events["proton"], reduced_events["electron"]])
+
+    def setup(self):
+        self.tc = ToolConfig(parent=self)
+        self.bins = DataBinning(parent=self)
+        self.eps = EventPreProcessor(parent=self)
```
maxnoe, line 150: eps seems a bad name for this

ctapipe/irf/irf_classes.py
Tobychev: I'll go with efficiency then.

ctapipe/irf/irf_classes.py
Tobychev: This way it stays consistent with how you specify quality criteria, plus I don't have to rewrite the code :)

ctapipe/irf/irf_classes.py
Tobychev: fixed

ctapipe/irf/irf_classes.py
Tobychev: implemented

ctapipe/irf/irf_classes.py
```diff
@@ -0,0 +1,389 @@
...
+    max_gh_cut_efficiency = Float(
+        default_value=0.8, help="Maximum gamma purity requested"
+    ).tag(config=True)
+    gh_cut_efficiency_step = Float(
+        default_value=0.1,
+        help="Stepsize used for scanning after optimal gammaness cut",
+    ).tag(config=True)
+    initial_gh_cut_efficency = Float(
+        default_value=0.4, help="Start value of gamma purity before optimisation"
+    ).tag(config=True)
```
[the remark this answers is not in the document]
Tobychev: implemented

ctapipe/irf/irf_classes.py
Tobychev: implemented

ctapipe/tools/make_irf.py
Tobychev: fixed

ctapipe/irf/irf_classes.py
Tobychev: implemented

ctapipe/irf/irf_classes.py
Tobychev: I just split out the energy binning for now, as I wasn't sure how to group the rest.

ctapipe/irf/irf_classes.py
```diff
@@ -0,0 +1,335 @@
...
+import astropy.units as u
+import numpy as np
+from astropy.table import QTable
+from pyirf.binning import create_bins_per_decade
+
+from ..core import Component, QualityQuery
+from ..core.traits import Float, Integer, List, Unicode
+
+
+class CutOptimising(Component):
```
maxnoe, line 13: We don't use components to just store configuration. 

They should *actually do* the thing they are created for. 

E.g. I think it's an anti-pattern to have a class `CutOptimisation` that only holds the config but then the actual optimization happens in the tool.

The class should have  a method to do the cut optimization using it's own traitlets and an input (e.g. signal and background tables)

ctapipe/irf/irf_classes.py
Tobychev: This class is actually just dead code, removed it.

ctapipe/irf/irf_classes.py
Tobychev: Ok, moved code into it. But the pyirf api looks written for the case when everything is in a shared global scope, which is the opposite of putting things into separate containers.

ctapipe/irf/irf_classes.py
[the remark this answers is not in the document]
Tobychev: implemented

environment.yml
```diff
@@ -25,6 +25,7 @@ dependencies:
   - pandas
   - pre-commit
   - psutil
+  - pyirf
```
kosack, line 28: Can you also add this to `ctapipe/tools/info.py` in the list of dependency version to print? (`_dependencies`)

ctapipe/irf/irf_classes.py
```diff
@@ -0,0 +1,364 @@
...
+    gh_cut_efficiency_step = Float(
+        default_value=0.1,
+        help="Stepsize used for scanning after optimal gammaness cut",
+    ).tag(config=True)
+    initial_gh_cut_efficency = Float(
+        default_value=0.4, help="Start value of gamma efficiency before optimisation"
+    ).tag(config=True)
+
+    def optimise_gh_cut(
+        self, signal, background, Et_bins, Er_bins, bins, alpha, max_bg_radius
```
maxnoe, line 32: is Et_bins, Er_bins for true / reconstructed energy?  what is "bins"?

environment.yml
Tobychev: done

ctapipe/irf/irf_classes.py
```diff
@@ -0,0 +1,364 @@
...
+        self.log.info(
+            f"Using fixed G/H cut of {INITIAL_GH_CUT} to calculate theta cuts"
+        )
+
+        mask_theta_cuts = signal["gh_score"] >= INITIAL_GH_CUT
+
+        theta_cuts = calculate_percentile_cut(
+            signal["theta"][mask_theta_cuts],
+            signal["reco_energy"][mask_theta_cuts],
+            bins=Et_bins,
```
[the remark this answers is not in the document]
Tobychev: fixed

ctapipe/tools/make_irf.py
```diff
@@ -0,0 +1,385 @@
...
+                self.fov_offset_bins,
+            )
+        )
+
+        self.log.info("Writing outputfile '%s'" % self.output_path)
+        fits.HDUList(hdus).writeto(
+            self.output_path,
+            overwrite=self.overwrite,
+        )
+
```
kosack, line 377: Add here:
```py3
Provenance().add_output_file(self.output_path, role="IRF") 
```
That will tell the provenance system that you created a file successfully

kosack (changes requested):

ctapipe/irf/irf_classes.py
kosack: This should be renamed `CutOptimizer` to be consistent with other Components (they are named as nouns by the thing they do, e.g. "EnergyRegressor", "ImageCleaner")

ctapipe/tools/make_irf.py
```diff
@@ -0,0 +1,385 @@
...
+            alpha=self.alpha,
+            fov_offset_min=self.bins.fov_offset_min * u.deg,
+            fov_offset_max=self.bins.fov_offset_max * u.deg,
+        )
+        self.sensitivity = calculate_sensitivity(
+            signal_hist, background_hist, alpha=self.alpha
+        )
+
+        # scale relative sensitivity by Crab flux to get the flux sensitivity
+        for s in (sens2, self.sensitivity):
```
kosack, line 265: Not sure why you also scale `sens2`, which is never used.  What is the difference between self.sensitivity computed above, and sens2, which is computed during the optimization? Are they the same thing?

ctapipe/tools/make_irf.py
```diff
@@ -0,0 +1,385 @@
...
+                        kind, selected, spectrum, target_spectrum, obs_conf
+                    )
+                    bits.append(selected)
+
+                table = vstack(bits, join_type="exact")
+                reduced_events[kind] = table
+
+        select_ON = reduced_events["gamma"]["theta"] <= self.ON_radius * u.deg
+        self.signal = reduced_events["gamma"][select_ON]
+        self.background = vstack([reduced_events["proton"], reduced_events["electron"]])
```
kosack, line 212: To avoid some confusion, I would rename these  to `self.signal_events` and `self.background_events`, since there is also later background that is a histogram

ctapipe/tools/make_irf.py
```diff
@@ -0,0 +1,385 @@
...
+        self.signal["selected"] = (
+            self.signal["selected_theta"] & self.signal["selected_gh"]
+        )
+
+        # calculate sensitivity
+        signal_hist = create_histogram_table(
+            self.signal[self.signal["selected"]], bins=self.reco_energy_bins
+        )
+
+        background_hist = estimate_background(
```
kosack, line 252: What is the difference between this background hist and the one generated in `finish()`? One uses `estimate_background` and one uses `background_2d`.  Are these the same computations? I worry that things are computed multiple times, and it's not clear which ones are in the output vs used in the sensitivity

kosack: In fact, I don't find `pyirf.sensitivity.estimate_background` in the PyIRF documentation, so could it be incorrect?

ctapipe/tools/make_irf.py
```diff
@@ -0,0 +1,385 @@
...
+        return events
+
+    def get_metadata(self, loader):
+        obs = loader.read_observation_information()
+        sim = loader.read_simulation_configuration()
+
+        # These sims better have the same viewcone!
+        assert sim["max_viewcone_radius"].std() == 0
+        sim_info = SimulatedEventsInfo(
+            n_showers=sum(sim["n_showers"] * sim["shower_reuse"]),
```
kosack, line 161: probably should used np.sum here

kosack, line 161: Here there is an implicit assumption that all of the runs contained in the file use the same distribution of thrown events (in energy and impact).  However, that is not necessary to assume.  We can easily support having mixed ranges, and indeed that will likely happen in the future (e.g. to fill in statistics at high energy).  

So at the very least this assumption should be made explicit and you should have a sanity check here that they are all the same, and raise an exception if not.  At the very least, add some more asserts, but better raise a more useful exception explaining to the user that the inputs don't agree and this tool assumes they do.

Later, we should really use the actual thrown energy histograms (`/simulation/service/shower_distribution`)! Those are in the file and can also be summed up to produce the final histogram. In that case, no PowerLaws need to be created, as the shape can be anything.

ctapipe/tools/make_irf.py
```diff
@@ -0,0 +1,385 @@
...
+        )
+
+        hdus.append(
+            create_rad_max_hdu(
+                self.theta_cuts_opt["cut"].reshape(-1, 1),
+                self.true_energy_bins,
+                self.fov_offset_bins,
+            )
+        )
+
```
kosack, line 371: For debugging and benchmarking, it might be nice to add a few more historgrams in the output files (all vs energy and offset):
* N_on and N_off counts (maybe even N_excess and Significance), which can help to see where  limiting factors of background and minimum counts are taken into account.
* background rate _per particle species_ (same as the background2d already there, but for electrons and protons separately)
* Effective area for protons and electrons separately

ctapipe/tools/make_irf.py
Tobychev: fixed

ctapipe/tools/make_irf.py
Tobychev: This is a legacy of whatever script it I converted into the tool, I've just added saving sens2 to the output to check if there is any difference at all. For example, it might be that historically there was some difference but over time pyirf was refactored to use the same sensitivity function everywhere and now there is no difference.

ctapipe/tools/make_irf.py
Tobychev: added

ctapipe/tools/make_irf.py
maxnoe: > Here there is an implicit assumption that all of the runs contained in the file use the same distribution of thrown events (in energy and impact). However, that is not necessary to assume. We can easily support having mixed ranges, and indeed that will likely happen in the future (e.g. to fill in statistics at high energy).

This is at the moment not supported by pyirf effective area computation, you need to pass a single of these `SimulatedEventsInfo` objects.

Agree that a check should be done though that `e_min`, `e_max` and `spectral_index` are indentical.

maxnoe: > Later, we should really use the actual thrown energy histograms (/simulation/service/shower_distribution)! 

This is what we should already do now! The total number of events should come from the shower histograms, not the header information.

ctapipe/tools/make_irf.py
Tobychev: That function is what pyIRF uses [internally](https://github.com/cta-observatory/pyirf/blob/12a609566230d01a68a822aad38f080b60c473ce/pyirf/cut_optimization.py#L119) so probably this is a problem with the docs. 

To be honest I'm not sure what the difference is. It looks like [estimate_background](https://github.com/cta-observatory/pyirf/blob/main/pyirf/sensitivity.py#L268) only does one offset bin but also rescales from the size of the Off region to the On region. Meanwhile [background_2d](https://github.com/cta-observatory/pyirf/blob/main/pyirf/irf/background.py) calculates background rates in each offset bin, but does not rescale to the On region size...

maxnoe: `pyirf.sensitivity.estimate_background`

this function estimates the number of background *counts*  (n_off) for the given observation time and given on region size in the binning used for the IRF computation. 

 
`pyirf.irf.background.background_2d` is the background IRF component as defined by GADF.

ctapipe/tools/make_irf.py
maxnoe: Which script are you referring to?

maxnoe: The current example script in pyirf only computes sensitivity once using `optimize_gh_cuts`

ctapipe/irf/irf_classes.py
Tobychev: renamed

ctapipe/tools/make_irf.py
Tobychev: renamed

ctapipe/tools/make_irf.py
Tobychev: @maxnoe Probably this [protopipe](https://github.com/cta-observatory/protopipe/blob/master/protopipe/scripts/make_performance_EventDisplay.py) script. Is [this](https://github.com/cta-observatory/pyirf/blob/9c7aecc2e0491170adb83c55ddb54de4295714f8/examples/calculate_eventdisplay_irfs.py) the blessed script in pyIRF? The "Event display" in the name always makes me hesitant to conclude it is the appropriate script for `ctapipe` files.

ctapipe/tools/make_irf.py
Tobychev: Added reading of the shower distribution and using that for the number of showers, changed the max viewcone assert for three if statments checking e_min, e_max and spectral_index.

ctapipe/tools/make_irf.py
```diff
@@ -154,11 +154,25 @@ def make_derived_columns(self, kind, events, spectrum, target_spectrum, obs_conf
     def get_metadata(self, loader):
         obs = loader.read_observation_information()
         sim = loader.read_simulation_configuration()
+        show = loader.read_shower_distribution()
 
         # These sims better have the same viewcone!
+        if not np.diff(sim["energy_range_max"]).sum() == 0:
```
maxnoe, line 160: To make intent clearer here, this could be `len(np.unique(...)) > 0` for each of these.

You can also test them in a loop over the keys.

Tobychev: implemented.

ctapipe/tools/make_irf.py
kosack: Great. Even if we don't support using the actual thrown histogram yet in pyirf, using them to get the counts is still probably safer as we do not write those histograms unless the full file is processed (_max-events_ is not set). Though I suppose in that case we would then potentially undercount instead of overcount when summing them...

kosack:
It occurs to me that it would be very useful to add a unit test that reads the generated IRFs using gammapy.  You would have to add gammapy the test requirements in pyproject.toml, but that is fine - it's not a ctapipe dependency, but still could be a test dependency.  That way you test explicitly that the output is readable, which currently it is not due to a mismatch in the HDU bins for background.

ctapipe/irf/irf_classes.py
```diff
@@ -117,28 +97,75 @@ def optimise_gh_cut(self, signal, background, alpha, max_bg_radius):
...
             self.reco_energy_bins(),
-            percentile=68,
-            min_value=self.theta_min_angle * u.deg,
-            max_value=self.theta_max_angle * u.deg,
+        )
+
+        return gh_cuts, theta_cuts, sens2
+
+
+class PointSpreadFunction(Component):
```
maxnoe, line 121: I'd not call this "PointSpreadFunction", since it is not a PointSpreadFunction, which is component of the IRF, it computes the theta cuts, which of course depend on the PSF, but it's not the same.

maxnoe:
Follow up here #2473

PR #2319, opened 2023-04-26
Do not use pathlib for hdf5 datasets, fixes #2318

### LINKED ISSUE #2318: PurePath screws up HDF5 table construction on Windows systems
**Describe the bug**
On Windows systems, code such as
```
with HDF5TableWriter(some_filename) as writer:
    writer.write(table_name="test", containers=some_container)
```
crashes with error `NameError: ``where`` must start with a slash ('/')`

The reason is that `HDF5TableWriter._setup_new_table()` makes use of `pathlib.PurePath` to set up HDF5 table name and path. The HDF5 format expects slashes ("/") as a separator for the internal data organisation.

On Windows systems, however, `PurePath` automatically evaluates to `PureWindowsPath` which expands any slash to a back-slash ("\\"), screwing up the internal HDF5 path description.

**Proposed solution**

The proposed solution is to replace `PurePath` by `PurePosixPath` in `HDF5TableWriter._setup_new_table()`, specifically in
```
table_path = PurePath(self._group) / PurePath(table_name)
```

(However, I didn't test this, as I don't have access to a Windows system myself).

**To Reproduce**
Minimal example for what happens in `HDF5TableWriter._setup_new_table()`:
```
group = "/"
table_name = "example_table"

table_path = PurePath(group) / PurePath(table_name)
table_group = str(table_path.parent)
table_basename = table_path.stem
table_path = str(table_path)

print(table_path)
print(table_group)
```

This results on Windows systems in
```
\example_table
\
```

**Expected behavior**

For the HDF5 internal data structure, I would expect that the result should look as follows:
```
/example_table
/
``` 
This is what is produced on Linux systems.

**Supporting information**
If applicable, add screenshots, logs, or a traceback to help illustrate the bug.

**Additional context**

maxnoe:
fixes #2318

### DIFF
--- a/ctapipe/io/hdf5tableio.py
+++ b/ctapipe/io/hdf5tableio.py
@@ -20,7 +20,7 @@ from .tableio import (
     TimeColumnTransform,
 )
 
-__all__ = ["HDF5TableWriter", "HDF5TableReader"]
+__all__ = ["HDF5TableWriter", "HDF5TableReader", "split_h5path"]
 
 PYTABLES_TYPE_MAP = {
     "float": tables.Float64Col,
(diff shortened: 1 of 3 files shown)


kosack (approved):

ctapipe/io/hdf5tableio.py
Tobychev, line 1: Is this use of paths inside of h5 files tested somewhere? I took a look at `io/tests/test_hdf5.py` but I nothing stands out as testing the access of paths _inside_ of a file.

maxnoe:
> Was there a reason to implement this by hand rather than just switching to using `PurePosixPath` as stated in the issue?

I only used (and moved) the already existing functionality to fix this specific issue.

I agree that we probably should cleanup all our path handling in the hdf5 modules and `PurePosixPath` might be a good solution, but I'd keep that for another PR.

Tobychev (approved):

PR #2328, opened 2023-05-11
Make provenance information in log optional

Tobychev:
The inclusion of all the packages in the environment into the provenance information makes it so long it usually completely dominates the line count of any logfile. This change adds a option that defaults to not including provenance in the log (in addition to it being written to a file).

### DIFF
--- a/ctapipe/core/tool.py
+++ b/ctapipe/core/tool.py
@@ -173,6 +173,13 @@ class Tool(Application):
         help="Logging Level for File Logging",
     ).tag(config=True)
 
+    log_include_provenance = Bool(
+        default_value=False,
+        help="Control if provenance information"
+        "is printed in log in addition to the"
+        "provenance file."
+    ).tag(config=True)
+
     quiet = Bool(default_value=False).tag(config=True)
     overwrite = Bool(default_value=False).tag(config=True)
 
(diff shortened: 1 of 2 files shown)


kosack:
Another option here is simply to never write the provenance to the stdout logfile.  I only had added that for debugging before I had added the `provenance.log` output. But now that there is a provenance log, it's probably redundant to include it at the DEBUG level.

ctapipe/core/tool.py
```diff
@@ -173,6 +173,13 @@ def main():
         help="Logging Level for File Logging",
     ).tag(config=True)
 
+    log_include_provenance = Bool(
```
kosack, line 176: Don't add an option for this (it's somewhat confusing to have control options affect only debugging output).  Better to just decrease the log level of that print out (could set it to 5, which is below debug), or as I mentioned below, simply remove the printout altogether.  I think I prefer the latter option to just drop it, as we already print it in the provenance file.

kosack (changes requested):

ctapipe/core/tool.py
maxnoe: We could also add use a specific provenance sublogger so it's configurable on its own.

```
self.log.getChild("provenance").debug(...)
```

Tobychev:
@kosack @maxnoe I decided for the "remove it" option, with a note in the logs where for where the interested reader might find the provenance.

maxnoe (approved):

kosack (approved):

Issue #2326, opened 2023-05-10
Minor change to tailcuts cleaning  to improve the detection of signal pixels


clara-escanuela:
The tailcut cleaner could be modified in order to add pixels above the (low) boundary threshold if they continuously connect to a pixel above the (high) picture threshold, rather than only the first neighbour. This would require a very small change in the code. It should be equivalent to the hysteresis thresholding implemented in scikit-image (https://scikit-image.org/docs/dev/api/skimage.filters.html#skimage.filters.apply_hysteresis_threshold, reference https://ieeexplore.ieee.org/document/4767851) .

![image](https://github.com/cta-observatory/ctapipe/assets/53437218/e650eee1-1914-4b79-b9f0-00796b474604)
![image](https://github.com/cta-observatory/ctapipe/assets/53437218/074c911b-c285-45c2-888a-204a16bf9523)
![image](https://github.com/cta-observatory/ctapipe/assets/53437218/e3666296-9204-4922-979f-39674279b391)

maxnoe:
That's a nice idea but I'd say not a *small* change to tailcuts.

THat rather seems like it's own cleaning method as an adaption of the normal tailcuts.

I'd be in favor of adding it though (or as an option to the existing tailcuts)

clara-escanuela:
I thought it could be added either as an option in tailcuts or a new cleaning method (although it is very similar to tailcuts)

PR #2329, opened 2023-05-11
Issue #2326 

clara-escanuela:
This would be my approach for Issue #2326


Hckjs:
Maybe also have a look at [this part](https://github.com/cta-observatory/ctapipe/blob/f28298667b7dffea5da87b582bd646362c4ef909/ctapipe/image/reducer.py#L202-L207) of the VolumeReducer, i think its quite similar. Maybe you can write a function used in both classes?

clara-escanuela:
> Maybe also have a look at [this part](https://github.com/cta-observatory/ctapipe/blob/f28298667b7dffea5da87b582bd646362c4ef909/ctapipe/image/reducer.py#L202-L207) of the VolumeReducer, i think its quite similar. Maybe you can write a function used in both classes?

That would make sense I think. Or can we somehow alter the "dilate" function so that it has the option to only dilate pixels above certain threshold?

clara-escanuela:
By the way, what is the reason for CI / docs to fail? The error message is not so clear to me and I am curious as this is also happening in my other PR

ctapipe/image/cleaning.py
```diff
@@ -374,6 +375,90 @@ def fact_image_cleaning(
...
+            ) | pixels_in_picture
+        else:
+            pixels_with_boundary_neighbors = geom.neighbor_matrix_sparse.dot(
+                pixels_above_boundary
+            )
+            pixels_in_picture = (
+                pixels_above_boundary & pixels_with_picture_neighbors
+            ) | (pixels_in_picture & pixels_with_boundary_neighbors)
+
+        iteration = ~np.all(pixels_in_picture == pixels_in_picture_previous)
```
maxnoe, line 454: `~np.all(a == b)` is `np.any(a != b)`, right?

ctapipe/image/cleaning.py
```diff
@@ -374,6 +375,90 @@ def fact_image_cleaning(
...
+        )
+        pixels_in_picture = pixels_above_picture & (
+            number_of_neighbors_above_picture >= min_number_picture_neighbors
+        )
+
+    # by broadcasting together pixels_in_picture (1d) with the neighbor
+    # matrix (2d), we find all pixels that are above the boundary threshold
+    # AND have any neighbor that is in the picture
+    pixels_above_boundary = image >= boundary_thresh
+    iteration = True
```
maxnoe, line 435: I think a for `for count in range(max_iter):` with a `break` when convergence is reached would be easier to read

ctapipe/image/cleaning.py
```diff
@@ -374,6 +375,91 @@ def fact_image_cleaning(
...
+            ) | pixels_in_picture
+        else:
+            pixels_with_boundary_neighbors = geom.neighbor_matrix_sparse.dot(
+                pixels_above_boundary
+            )
+            pixels_in_picture = (
+                pixels_above_boundary & pixels_with_picture_neighbors
+            ) | (pixels_in_picture & pixels_with_boundary_neighbors)
+
+        iteration = np.all(pixels_in_picture == pixels_in_picture_previous)
```
kosack, line 455: `iteration` is maybe not the best name here (unclear, as usually that would be the count of what iteration you are on), and even if it is intended to be "iteration_needed",  the meaning is the opposite: True means nothing changed from the previous iteration, so you can stop iterating.  So maybe call it something like `is_unchanged`, then the later if statement also makes more sense `if is_unchanged: break` .

ctapipe/image/cleaning.py
```diff
@@ -374,6 +375,91 @@ def fact_image_cleaning(
...
+        pixels_in_picture = pixels_above_picture & (
+            number_of_neighbors_above_picture >= min_number_picture_neighbors
+        )
+
+    # by broadcasting together pixels_in_picture (1d) with the neighbor
+    # matrix (2d), we find all pixels that are above the boundary threshold
+    # AND have any neighbor that is in the picture
+    pixels_above_boundary = image >= boundary_thresh
+
+    for count in range(max_iter):
```
kosack, line 438: I know we both complained that this is a new cleaning method, not just tailcuts,but in the end the code is nearly 99% copied from tailcuts.   So maybe:

1. we reconsider and just add an iteration option to tailcuts_clean (even if it's a bit ugly and probably breaks processing all events at once)
2. factor out the common code in both functions to a separate function that can be called by `tailcuts_clean` and `tailcuts_hysteresis_clean`
3. keep as is, but have lots of code duplication and the very real posisbility of forgetting to fix bugs in multiple places.

@maxnoe what do you think?

ctapipe/image/cleaning.py
```diff
@@ -374,6 +375,91 @@ def fact_image_cleaning(
...
+        pixels_in_picture_previous = pixels_in_picture.copy()
+        pixels_with_picture_neighbors = geom.neighbor_matrix_sparse.dot(
+            pixels_in_picture
+        )
+        if keep_isolated_pixels:
+            pixels_in_picture = (
+                pixels_above_boundary & pixels_with_picture_neighbors
+            ) | pixels_in_picture
+        else:
+            pixels_with_boundary_neighbors = geom.neighbor_matrix_sparse.dot(
```
Tobychev, line 448: You can move this out of the loop because it is a constant

kosack: yes this will gain you a bit of speed as well, since pixels_above_boundary is not changing, there is no need to do the dot-product each time

Tobychev (approved):

PR #2333, opened 2023-05-15
Fix peak_time estimation in FlashCamExtractor

clara-escanuela:
After the extractor optimization, I did not check timing performance again as I only made a few changes in the code. But one of these very small changes caused a problem in the time reconstruction. The unit tests did not identify it because they  were not sensitive enough. By adding a new toymodel with a time gradient we can find timing bugs better. I checked the test with the previous implementation (with that wrong line of code) and the test fails (as expected). In fact, the same test with the standard toymodel provided (random true times) passes even when time reconstruction is wrong. I think a further test can also be added to other extractors, to avoid similar problems.

### DIFF
--- a/ctapipe/image/extractor.py
+++ b/ctapipe/image/extractor.py
@@ -1679,8 +1679,9 @@ class FlashCamExtractor(ImageExtractor):
 
         if leading_edge_timing:
             d_waveforms = deconvolve(waveforms, 0.0, upsampling, 1)
+
             # correct the offset between leading edge peak and deconvolved peak
-            peak_index = np.round(peak_index - pz2d).astype(int)  
+            peak_index = np.round(peak_index - pz2d).astype(int)
             n_samples = d_waveforms.shape[-1]
             np.clip(peak_index, 0, n_samples - 1, out=peak_index)
             peak_time = adaptive_centroid(
--- a/ctapipe/image/extractor.py
+++ b/ctapipe/image/extractor.py
@@ -1679,9 +1679,8 @@ class FlashCamExtractor(ImageExtractor):
 
         if leading_edge_timing:
             d_waveforms = deconvolve(waveforms, 0.0, upsampling, 1)
-            peak_index = np.round(peak_index - pz2d).astype(
-                int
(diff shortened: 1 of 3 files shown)


ctapipe/image/extractor.py
```diff
@@ -1681,7 +1681,9 @@ def __call__(
             d_waveforms = deconvolve(waveforms, 0.0, upsampling, 1)
             peak_time = adaptive_centroid(
                 d_waveforms,
-                max(0, min(np.round(peak_index - pz2d).astype(int))),
+                np.clip(
+                    np.round(peak_index - pz2d).astype(int), 0, np.shape(d_waveforms)[1]
```
maxnoe, line 1685: `waveforms.shape[-1]`

Also, clip is inclusive, so it should probably have a -1?

to make the intent clear, it would be more readable to split this into multiple assignments.

Maybe also add a comment why you shift the peak_index by `pz2d`, this is not clear to me. 

You can also avoid a copy by using the `out` arg of `clip`
something like:

```
peak_index = np.round(peak_index - pz2d).astype(int)
n_samples = d_waveforms.shape[-1]
np.clip(peak_index, 0, n_samples - 1, out=peak_index) 
peak_time = adaptive_centroid(d_waveforms, peak_index, leading_edge_rel_descend_limit)
```

maxnoe:
Before:
flashcam_image png

After:
flashcam_image_fixed

docs/changes/2333.bug.rst
```diff
@@ -0,0 +1 @@
+Solve bug in time reconstruction with the FlashCamExtractor
```
maxnoe, line 1:
```suggestion
Fix a bug in the peak_time estimation of ``FlashCamExtractor`` (See issue `#2332 <https://github.com/cta-observatory/ctapipe/issues/2332>`_) 
```

maxnoe (approved):

maxnoe:
@Tobychev @kosack 

Could we get a second quick review here? We need to have this release to restart processing on the grid

Tobychev (approved):

PR #2338, opened 2023-05-26
Add fields defined in CTAO DL0 and R1 data models 

maxnoe:


ctapipe/containers.py
```diff
@@ -165,6 +165,67 @@ class EventType(enum.Enum):
...
+            0 = pixel broken / disabled
+            1 = only high gain read out
+            2 = only low gain read out
+            3 = both gains read out
+        """
+        gain_bits = PixelStatus.HIGH_GAIN_STORED | PixelStatus.LOW_GAIN_STORED
+        return (pixel_status & gain_bits) >> 2
+
+    @staticmethod
+    def is_broken(pixel_status):
```
kosack, line 223: Since people often complain about the term "broken" here, maybe rename this `is_excluded()` to be more clear that the pixel may be excluded for reasons other than being physically broken (e.g. disabled, noisy, etc).

ctapipe/containers.py
```diff
@@ -503,17 +564,51 @@ class R1CameraContainer(Container):
...
+        dtype=np.uint8,
+    )
+
+    first_cell_id = Field(
+        None,
+        "Array of first cell ids of the readout chips. Only used by LST and SST.",
+        dtype=np.uint32,
+    )
+
+    module_hires_local_clock_counter = Field(
```
kosack, line 590: not a great name, but it's also bad in the document...  Not totally clear why the word "local" is there. I guess local is that the time reference is module-dependent? Maybe at least mention what this means in the help string so we know how to use it later.  `hires` is also just about the storage format, so could be dropped.  This could just be `module_clock_counter`, with a help string saying something like "count of clock ticks since a locally defined reference within the camera module" (if that is indeed what it is...)

kosack (changes requested):

ctapipe/containers.py
maxnoe: Is it worth it to deviate from the term in the data model document? I'd say we need a very strong reason to do that and then we'd have to add references like "note: this is called module_hires_local_clock_counter in the data model" every where. 

I also don't much like the chosen name, but at least in this case I preferred to just follow the name in the data model instead of trying to invent a better one.

ctapipe/containers.py
maxnoe: Then I'd say disabled or missing is the correct term, as at R1 / DL0 it means whether the pixel was read out at all.

kosack: Actually, in the R1 model, all pixels are always read out, but may have bogus values in them (the readout is fixed-length).   In the discussions about the interface, it seems this field may also be used for noisy pixels (e.g. those that are read out, but maybe not good for analysis due to a star or something), so _excluded_ is perhaps the better term.  It's different from the DVR flag, which says if this data are stored or not at the DL0 level. 

So this flag means one of two things:
* the pixel is physically broken/missing (i.e. it has an undefined signal in the readout that may be 0 or a junk number)
* the pixel has a high pedestal variance, meaning that it has a hardware fault, or has excess unexpected external light hitting it for some reason (bright star where the HV wasn't adjusted, etc)
* the pixel has a high _predicted_ light level and the HV was turned off or reduced due to an expected star. In that case, there may be some signal, but it is not properly calibrated and should be excluded.

kosack: Note that you would get these values even if the pixel is not stored due to DVR (the pixels_status is always reported)

ctapipe/containers.py
kosack: yeah, maybe stick with the name and just add a better help string for now. We can always update the name when the model evolves.

ctapipe/containers.py
kosack: Invalid is fine, and also nice as we use that term for the parameters as well.

kosack (approved):

kosack (approved):

ctapipe/calib/camera/calibrator.py
```diff
@@ -178,19 +178,35 @@ def _check_dl0_empty(self, waveforms):
...
+        dl0_waveform = r1.waveform.copy()
+        dl0_waveform[~signal_pixels] = 0
+
+        dl0_pixel_status = r1.pixel_status.copy()
+        # set dvr pixel bit in pixel_status for pixels kept by DVR
+        dl0_pixel_status[signal_pixels] |= PixelStatus.DVR_STORED_AS_SIGNAL
+        # unset dvr bits for removed pixels
+        dl0_pixel_status[~signal_pixels] &= ~np.uint8(PixelStatus.DVR_STATUS)
+
+        event.dl0.tel[tel_id] = DL0CameraContainer(
```
Tobychev, line 201: I'm surprised the task of adding fields to the existing storage classes involves the creation of an entirely new Container object in this method.

ctapipe/containers.py
```diff
@@ -165,6 +165,70 @@ class EventType(enum.Enum):
     UNKNOWN = 255
 
 
+class PixelStatus(enum.IntFlag):
+    """
+    Pixel status information
+
+    See DL0 Data Model specification:
+    https://redmine.cta-observatory.org/dmsf/files/17552/view
```
Tobychev, line 173: I don't think this document URL is publicly accessible, so not that helpful to include it.

maxnoe: Most people developing ctapipe will have access to it and it is the official source for these definitions. I think not including it for reference is worse.

maxnoe: It would be great if these would be publicly available though, there was talk about publishing them on e.g. zenodo.

@kosack

ctapipe/io/simteleventsource.py
```diff
@@ -819,9 +823,16 @@ def _generate_events(self):
                     self.calib_scale,
                     self.calib_shift,
                 )
+
+                pixel_status = self._get_r1_pixel_status(
+                    tel_id=tel_id,
+                    selected_gain_channel=selected_gain_channel,
+                )
                 data.r1.tel[tel_id] = R1CameraContainer(
+                    event_type=trigger.event_type,
```
Tobychev, line 832: How can you add extra inputs to the `R1CameraContainer` without updating the definition of the container?

ctapipe/containers.py
Tobychev, line 173: I don't think this document URL is publicly accessible, so not that helpful to include it.

ctapipe/calib/camera/calibrator.py
Tobychev, line 201: I'm surprised the task of adding fields to the existing storage classes involves the creation of an entirely new Container object in this method. What change required this addition?

ctapipe/io/simteleventsource.py
maxnoe: It is updated? Lines 570ff (new line numbers)

Tobychev: Oh, yeah! sorry. It was folded by github and when I tried to check the definition it took me to the old version.

Tobychev (approved):

PR #2343, opened 2023-06-02
Remove SimulationConfigContainer.injection_height, add SimulatedShowercontainer.starting_grammage

### LINKED ISSUE #2335: Changes to simulation containers

In a discussion with @kbernloehr in the gitlab, it came up that the `injection_height` in the `mcrunheader` object, which we store in the `SimulationConfigContainer` is not used by simtel array and exists only for historical reasons, it is always -1.

The real injection height, is stored for each shower in the `MCShower` object, but currently we don't store this.

Remove `injection_height` from the `SimulationConfigContainer` (or at least set it to nan or None in `SimtelEventSource`) and add `injection_height` to the `SimulatedShowerContainer`.

maxnoe:
Fixes #2335

### DIFF
--- a/ctapipe/containers.py
+++ b/ctapipe/containers.py
@@ -586,8 +586,11 @@ class SimulatedShowerContainer(Container):
     core_x = Field(nan * u.m, "Simulated core position (x)", unit=u.m)
     core_y = Field(nan * u.m, "Simulated core position (y)", unit=u.m)
     h_first_int = Field(nan * u.m, "Height of first interaction", unit=u.m)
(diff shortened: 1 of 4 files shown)


kosack (approved):

kosack:
Does this merit a bump to the data model? It's not a really required part of the data, so maybe not (I mean it doesn't break reading old files, and up to now we don't use this column anywhere).   So my feeling would be no, or at least a minor version change rather than major.

Tobychev (approved):

PR #2347, opened 2023-06-13
Add support for hillas parameters in TelescopeFrame to CameraDisplay

maxnoe:
- [x] test

PR #2349, opened 2023-06-13
Add support for using toymodel in telescope frame

maxnoe:
missing tests

### DIFF
--- a/ctapipe/image/toymodel.py
+++ b/ctapipe/image/toymodel.py
@@ -41,13 +41,6 @@ __all__ = [
 TOYMODEL_RNG = default_rng(0)
 
 
-def _obtain_image_no_units(
-    x, y, centroid_x, centroid_y, psi, time_gradient, time_intercept
-):
-    longitudinal, _ = camera_to_shower_coordinates(x, y, centroid_x, centroid_y, psi)
-    return longitudinal * time_gradient + time_intercept
-
-
 def obtain_time_image(x, y, centroid_x, centroid_y, psi, time_gradient, time_intercept):
     """Create a pulse time image for a toymodel shower. Assumes the time development
     occurs only along the longitudinal (major) axis of the shower, and scales
@@ -79,15 +72,16 @@ def obtain_time_image(x, y, centroid_x, centroid_y, psi, time_gradient, time_int
 
     """
     unit = x.unit
-    return _obtain_image_no_units(
-        x=x.to_value(unit),
-        y=y.to_value(unit),
-        centroid_x=centroid_x.to_value(unit),
-        centroid_y=centroid_y.to_value(unit),
-        psi=psi.to_value(u.rad),
-        time_gradient=time_gradient.to_value(u.ns / unit),
-        time_intercept=time_intercept.to_value(u.ns),
-    )
+    x = x.to_value(unit)
+    y = y.to_value(unit)
+    centroid_x = centroid_x.to_value(unit)
+    centroid_y = centroid_y.to_value(unit)
+    psi = psi.to_value(u.rad)
+    time_gradient = time_gradient.to_value(u.ns / unit)
+    time_intercept = time_intercept.to_value(u.ns)
+
+    longitudinal, _ = camera_to_shower_coordinates(x, y, centroid_x, centroid_y, psi)
+    return longitudinal * time_gradient + time_intercept
 
 
 class WaveformModel:
--- a/ctapipe/image/toymodel.py
+++ b/ctapipe/image/toymodel.py
@@ -55,15 +55,15 @@ def obtain_time_image(x, y, centroid_x, centroid_y, psi, time_gradient, time_int
 
     Parameters
     ----------
-    x : u.Quantity[angle]
+    x : u.Quantity[angle|length]
         X camera coordinate to evaluate the time at.
         Usually the array of pixel X positions
-    y : u.Quantity[angle]
+    y : u.Quantity[angle|length]
         Y camera coordinate to evaluate the time at.
         Usually the array of pixel Y positions
-    centroid_x : u.Quantity[angle]
+    centroid_x : u.Quantity[angle|length]
         X camera coordinate for the centroid of the shower
-    centroid_y : u.Quantity[angle]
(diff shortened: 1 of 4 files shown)


maxnoe:
Just because I had the script already from #2347 


https://github.com/cta-observatory/ctapipe/assets/5488440/ad475013-53d6-4e23-a765-84db08397f69

ctapipe/image/toymodel.py
```diff
@@ -79,11 +78,17 @@ def obtain_time_image(x, y, centroid_x, centroid_y, psi, time_gradient, time_int
         Pulse time in nanoseconds at (x, y)
 
     """
-    longitudinal, _ = camera_to_shower_coordinates(x, y, centroid_x, centroid_y, psi)
-    longitudinal_m = longitudinal.to_value(u.m)
-    time_gradient_ns_m = time_gradient.to_value(u.ns / u.m)
-    time_intercept_ns = time_intercept.to_value(u.ns)
-    return longitudinal_m * time_gradient_ns_m + time_intercept_ns
+    unit = x.unit
+    # camera frame
```
LukasNickel, line 82: That comment can be deleted, right?

ctapipe/image/toymodel.py
```diff
@@ -40,35 +41,33 @@
...
+
 def obtain_time_image(x, y, centroid_x, centroid_y, psi, time_gradient, time_intercept):
     """Create a pulse time image for a toymodel shower. Assumes the time development
     occurs only along the longitudinal (major) axis of the shower, and scales
     linearly with distance along the axis.
 
     Parameters
     ----------
-    x : u.Quantity[length]
+    x : u.Quantity[angle]
```
LukasNickel, line 58: Is the description [angle] helpful?
This can now take angles or "lengths"

ctapipe/image/toymodel.py
```diff
@@ -183,11 +188,9 @@ def from_camera_readout(cls, readout, gain_channel=0):
 
 
 class ImageModel(metaclass=ABCMeta):
-    @u.quantity_input(x=u.m, y=u.m)
```
Tobychev, line 186: What is the background to all these Quantity wrappers being deleted?

maxnoe: To support both telescope and camera frame, things need to support degrees and meters, unfortunately, `u.quantity_input` does not allow something like `x=(u.deg, u.m)` to say "either meters or degrees").

So instead now, we take the unit of `x` and check for consistency between the arguments by excplicitly calling `to_value(x.unit)` or comparing to the unit of the geometry

ctapipe/image/toymodel.py
```diff
@@ -40,35 +41,33 @@
...
-    centroid_x : u.Quantity[length]
+    centroid_x : u.Quantity[angle|length]
         X camera coordinate for the centroid of the shower
-    centroid_y : u.Quantity[length]
+    centroid_y : u.Quantity[angle|length]
         Y camera coordinate for the centroid of the shower
     psi : convertible to `astropy.coordinates.Angle`
         rotation angle about the centroid (0=x-axis)
-    time_gradient : u.Quantity[time/length]
+    time_gradient : u.Quantity[time/angle]
```
LukasNickel, line 70: What exactly does this mean? Time divided by length|angle? That would be smth like `time/[angle|length]` maybe? 
Or can we get away with just writing `length` everywhere and in the docstring note that the unit of a length can be that of a distance or an angle?

ctapipe/image/toymodel.py
```diff
@@ -40,35 +41,33 @@
...
-@u.quantity_input(
-    x=u.m,
-    y=u.m,
-    centroid_x=u.m,
-    centroid_y=u.m,
-    psi=u.deg,
-    time_gradient=u.ns / u.m,
-    time_intercept=u.ns,
-)
+def _obtain_image_no_units(
```
LukasNickel, line 44: Is it really helpful to have this as an extra function? It only ever gets called once

ctapipe/image/toymodel.py
```diff
@@ -79,11 +78,16 @@ def obtain_time_image(x, y, centroid_x, centroid_y, psi, time_gradient, time_int
         Pulse time in nanoseconds at (x, y)
 
     """
-    longitudinal, _ = camera_to_shower_coordinates(x, y, centroid_x, centroid_y, psi)
-    longitudinal_m = longitudinal.to_value(u.m)
-    time_gradient_ns_m = time_gradient.to_value(u.ns / u.m)
-    time_intercept_ns = time_intercept.to_value(u.ns)
-    return longitudinal_m * time_gradient_ns_m + time_intercept_ns
+    unit = x.unit
```
LukasNickel, line 81: I think I prefer the old code structure (without the no_units function). It is easier to understand imo without the overhead of the function call.

ctapipe/image/toymodel.py
maxnoe: that might be the better way, yes.

ctapipe/image/toymodel.py
maxnoe: This is a leftover from me trying around with `u.quantity_input`

kosack:
Maybe we should just finish removing the CameraFrame versions of all of this... Then we only support TelescopeFrame, and you can transform into CameraFrame to see the result.

LukasNickel (approved):

Tobychev:
Hi @maxnoe can you explain what bits are actually about the new functionality? Most of this seems to be more style refactoring so I'm not sure what bit is where it is important to look closely...

maxnoe:
> Most of this seems to be more style refactoring so I'm not sure what bit is where it is important to look closely...

Actually, none of this is style refactoring, it's all directly related to support the two frames we can have image parameters in.

In the current main, we use `u.quantity_input(x=u.m, ..)` to make sure units are valid, however, image parameters can be in either `CameraFrame` which uses length units on the sensor or `TelescopeFrame` which uses angles on the sky (in degree).

The units checks prevented the toymodel to be used in the `TelescopeFrame`, now we support both.

kosack (approved):

PR #2342, opened 2023-06-02
Drop support for python 3.8, update env file

maxnoe:
In accordance with [NEP 29](https://numpy.org/neps/nep-0029-deprecation_policy.html):

> On Apr 14, 2023 drop support for Python 3.8 (initially released on Oct 14, 2019)

Setting to draft for now, since we might do another 0.19.x

PR #2353, opened 2023-06-19
Update environment.yml

maxnoe:


Tobychev (commented):
Why this new PR instead of promoting the [old one](https://github.com/cta-observatory/ctapipe/pull/2342) into a full PR? 

It seems to reproduce the env file from that draft but not contain any of the version gating code?

maxnoe:
I just forgot...

PR #2357, opened 2023-06-20
Fix removal of telescope events in software trigger, fixes #2356

### LINKED ISSUE #2356: dl1/event/telescope/trigger written for events that should have been removed by the SoftwareTrigger
**Describe the bug**

Reported by @arrabito for the processing of the Prod6 divergent pointing data on the GRID,
which due to a bug in the simulation config does not contain the hardware stereo trigger.

grid lfns:
```
/vo.cta.in2p3.fr/MC/PROD6/Paranal/gamma-diffuse/ctapipe-merge/4111/Data/-10xxx/gamma_20deg_0deg___cta-prod6-paranal-2147m-Paranal-divergent-div0.0043-dark_cone10.beta4lst_train_en_merged.DL2.h5
```

The telescope trigger table contains more events than the parameters table:
```
path = "gamma_20deg_0deg___cta-prod5b-lapalma_desert-2158m-LaPalma-dark_cone10.alpha_train_cl_applied.DL2.h5"

trigger = read_table(path, "/dl1/event/telescope/trigger", condition="tel_id==1")
parameters = read_table(path, "/dl1/event/telescope/parameters/tel_001")

print(len(trigger))
print(len(parameters))
```

```
145709
145565
```

All additional rows contain the default / nan values in the trigger data:

```
obs_id event_id tel_id time n_trigger_pixels
------ -------- ------ ---- ----------------
    15     2813      1  0.0               -1
    28   187615      1  0.0               -1
    47    32003      1  0.0               -1
    54    36405      1  0.0               -1
   113    96111      1  0.0               -1
   119    40813      1  0.0               -1
   140    93814      1  0.0               -1
   146   225116      1  0.0               -1
   152   140105      1  0.0               -1
   154    15407      1  0.0               -1
   161   510707      1  0.0               -1
   168    97907      1  0.0               -1
   172   140117      1  0.0               -1
   178    54914      1  0.0               -1
   186    10413      1  0.0               -1
   194    72809      1  0.0               -1
   239   201409      1  0.0               -1
   327   359406      1  0.0               -1
   363    27602      1  0.0               -1
   368    59907      1  0.0               -1
   ...      ...    ...  ...              ...
  1716   120812      1  0.0               -1
  1729   383419      1  0.0               -1
  1742    59812      1  0.0               -1
  1747    13010      1  0.0               -1
  1761    43808      1  0.0               -1
  1763    38610      1  0.0               -1
  1765    59410      1  0.0               -1
  1769   109210      1  0.0               -1
  1782    88509      1  0.0               -1
  1784    68917      1  0.0               -1
  1789   184011      1  0.0               -1
  1790   101702      1  0.0               -1
  1823     4814      1  0.0               -1
  1831      310      1  0.0               -1
  1832    55903      1  0.0               -1
  1851   188704      1  0.0               -1
  1863    77111      1  0.0               -1
  1870    89010      1  0.0               -1
  1905     2214      1  0.0               -1
  1932    21305      1  0.0               -1
```

So this might be an issue somewhere else, since the `SoftwareTrigger` seems to clear the trigger data correctly, but then the `DataWriter`? for some reason writes the default values.

**To Reproduce**

Tried to reproduce using the prod3 divergent file in the test data, but no success.

**Expected behavior**

Tables have same length and removed telescopes do not show up in the trigger table.

maxnoe:
fixes #2356

### DIFF
--- a/ctapipe/instrument/tests/test_trigger.py
+++ b/ctapipe/instrument/tests/test_trigger.py
@@ -1,9 +1,22 @@
+import json
+
 import numpy as np
 import pytest
 from numpy.testing import assert_equal
 
 from ctapipe.containers import ArrayEventContainer
 from ctapipe.io import EventSource
+
+
+def assert_all_tel_keys(event, expected):
+    expected = tuple(expected)
+    for name, container in event.items():
+        if hasattr(container, "tel"):
+            actual = tuple(container.tel.keys())
+            if len(actual) > 0 and actual != expected:
+                raise AssertionError(
+                    f"Unexpected tel_ids in container {name}:" f"{actual} != {expected}"
+                )
 
 
 @pytest.mark.parametrize("data_type", (list, np.array))
@@ -102,3 +115,43 @@ def test_software_trigger_simtel(allowed_tels):
         for e, expected_tels in zip(source, expected):
             trigger(e)
             assert_equal(e.trigger.tels_with_trigger, expected_tels)
+            assert_all_tel_keys(e, expected_tels)
+
+
+def test_software_trigger_simtel_process(tmp_path):
+    from ctapipe.core import run_tool
+    from ctapipe.io import TableLoader
+    from ctapipe.tools.process import ProcessorTool
+
+    path = "dataset://gamma_divergent_LaPalma_baseline_20Zd_180Az_prod3_test.simtel.gz"
+    config = dict(
+        ProcessorTool=dict(
+            EventSource=dict(
+                focal_length_choice="EQUIVALENT",
+            ),
+            SoftwareTrigger=dict(
+                min_telescopes=2,
+                min_telescopes_of_type=[
+                    ("type", "*", 0),
+                    ("type", "LST*", 2),
+                ],
+            ),
+        )
+    )
+
+    output_path = tmp_path / "software_trigger.dl1.h5"
+    config_path = tmp_path / "config.json"
+    config_path.write_text(json.dumps(config))
+
+    run_tool(
+        ProcessorTool(),
+        [f"--input={path}", f"--output={output_path}", f"--config={config_path}"],
+    )
+
+    with TableLoader(
+        output_path,
+        load_simulated=True,
+        load_dl1_parameters=True,
+        focal_length_choice="EQUIVALENT",
+    ) as loader:
+        loader.read_telescope_events("LST_LST_LSTCam")
--- a/ctapipe/instrument/tests/test_trigger.py
+++ b/ctapipe/instrument/tests/test_trigger.py
@@ -128,6 +128,8 @@ def test_software_trigger_simtel_process(tmp_path):
         ProcessorTool=dict(
             EventSource=dict(
                 focal_length_choice="EQUIVALENT",
+                # remove 3 LSTs, so that we trigger the 1-LST condition
+                allowed_tels=(1, 5, 6, 7, 8, 9, 10, 11, 12, 13, 14, 15, 16, 17, 18, 19),
             ),
             SoftwareTrigger=dict(
                 min_telescopes=2,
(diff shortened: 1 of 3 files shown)


ctapipe/io/datawriter.py
```diff
@@ -647,6 +647,11 @@ def _write_dl1_telescope_events(
         """
 
         # write the telescope tables
+        # trigger info
```
LukasNickel, line 650: is there any functional change in moving this block before the pointing loop?

maxnoe: This PR basically has to independent fixes for the issue #2356, each fix it on its own.

Writing the trigger data first prevents the possible mutation in the loop below for the pointing to affect the file written and the change in the software trigger now removes all telescope-event related data for telescopes that are removed.

Strictly speaking this change wouldn't be necessary anymore, but I also found the order more logical.

ctapipe/instrument/tests/test_trigger.py
```diff
@@ -6,6 +8,17 @@
 from ctapipe.io import EventSource
 
 
+def assert_all_tel_keys(event, expected):
+    expected = tuple(expected)
+    for name, container in event.items():
+        if hasattr(container, "tel"):
+            actual = tuple(container.tel.keys())
+            if len(actual) > 0 and actual != expected:
```
Tobychev, line 16: Not sure what the `len(actual)`  test is for? Empty tuples seem to evaluate as identical, so what are you testing for here?

ctapipe/instrument/tests/test_trigger.py
```diff
@@ -102,3 +115,118 @@ def test_software_trigger_simtel(allowed_tels):
...
+
+        for e, expected_tels in zip(source, expected):
+            print(e.trigger.tels_with_trigger)
+            trigger(e)
+            print(e.trigger.tels_with_trigger, expected_tels)
+            assert_equal(e.trigger.tels_with_trigger, expected_tels)
+            assert_all_tel_keys(e, expected_tels)
+
+
+def test_software_trigger_simtel_process(tmp_path):
```
Tobychev, line 170: Should probably be `test_process_software_trigger_simtel` to be consistent with the naming of the other tests outside of `test_process.py` that invokes the full `ctapipe-process` machinery.

I also wonder if it isn't more appropriate to put the generation of the two files into `conftest.py` since that place  is already full of `run_tool(ProcessTool)` invocations, whereas they don't seem to happen elsewhere except inside `tool/tests/` (where you expect it to show up).

LukasNickel (approved):

ctapipe/instrument/tests/test_trigger.py
```diff
@@ -102,3 +115,118 @@ def test_software_trigger_simtel(allowed_tels):
...
+                min_telescopes=2,
+                min_telescopes_of_type=[
+                    ("type", "*", 0),
+                    ("type", "LST*", 2),
+                ],
+            ),
+        )
+    )
+
+    output_path = tmp_path / "software_trigger.dl1.h5"
```
Tobychev, line 193: Where does this `tmp_path` come from? Is this some global setup thing from pytest?

ctapipe/instrument/tests/test_trigger.py
maxnoe: conftest is meant for fixtures that are used by multiple tests. Here we actually just want to look at these two specific files and compare them.

ctapipe/instrument/tests/test_trigger.py
maxnoe: data levels that are not filled at all (e.g. r0, dl2) would compare false to a list of expected telescopes

Tobychev: If you want to skip this test for certain data levels then demanding a certain length seems like a very non-obvious way to test for that. Can't you put the test for if a DL is filled into the `hasattr` conditional?

ctapipe/instrument/tests/test_trigger.py
Tobychev: Sure, I guess we can just move it once we invent a second test that could use one of these files.

ctapipe/instrument/tests/test_trigger.py
maxnoe: I added a keyword to explicitly ignore certain sub-containers instead of checking length 0

LukasNickel (approved):

Tobychev (approved):

PR #2369, opened 2023-06-26
Allow passing of ax to Subarray.peek, do not always call tight_layout

maxnoe:


kosack:
What is the use case here?  Really, you should just use an ArrayDisplay directly to do any more fancy plotting, the peek() function is for debugging or a quick look at what is there, not to make useful figures.  One of the design considerations early on in ctapipe was to avoid having plotting functions inside classes (after lessons learned about deep circular dependencies), which I agree is sort of violated already with this and the atmosphere module...

maxnoe:
The main thing I wanted to address was that I got a large warning from matplotlib from the `.peek()` methd in a notebook where I had constrained layout enabled by default for all figures.

Then, I am of the strong opinion that any matplotlib related function should allow overriding in which axes the thing is plotted, so I added that.

kosack (approved):

Issue #656, opened 2018-02-22
Methods run and start counterintuitive


MaxNoe:
The methods `run` and `start` of `ctapipe.core.tool.Tool` should be switched in my opinion.

They are the direct opposite of python's `threading` API for example, in a thread, the `run` method does the work and is invoked by `start`. 

It's the other way around for `Tool`

kosack:
That's fine with me  - not really sure why I did it the other way. It may have been a conflict with how `traitlets.config.Application` names it's methods.

maxnoe:
Looking at the traitlets aplication demo, we have it wrong also with respect to how traitlets defines it

PR #2371, opened 2023-07-04
Start working on main ctapipe cli tool

### LINKED ISSUE #850: Make a top-level Tool called ctapipe
so you can run ctapipe sub-commands with syntax like  `ctapipe produce-dl2` or `ctapipe
download examples` similar to what is in *gammapy*

maxnoe:
Fixes #850

I will also update the tools documentation in this PR, leaving as draft for now

### DIFF
--- a/ctapipe/__main__.py
+++ b/ctapipe/__main__.py
@@ -32,6 +32,10 @@ class MainTool(Application):
             "ctapipe.tools.train_disp_regressor.TrainDispRegressor",
             "Train telescope-type-wise disp regression models for direction reconstruction",
         ),
+        "merge": (
+            "ctapipe.tools.merge.MergeTool",
+            "Merge multiple ctapipe output files into one",
+        ),
         "fileinfo": (
             "ctapipe.tools.fileinfo.FileInfoTool",
             "Obtain metadata and other information from ctapipe output files",
--- a/ctapipe/__main__.py
+++ b/ctapipe/__main__.py
@@ -30,7 +30,8 @@ class MainTool(Application):
         ),
         "train-disp-regressor": (
             "ctapipe.tools.train_disp_regressor.TrainDispRegressor",
-            "Train telescope-type-wise disp regression models for direction reconstruction",
+            "Train telescope-type-wise disp regression models\n"
+            "for direction reconstruction",
         ),
         "merge": (
             "ctapipe.tools.merge.MergeTool",
@@ -47,6 +48,11 @@ class MainTool(Application):
         "quickstart": (
             "ctapipe.tools.quickstart.QuickStartTool",
             "Create a directory with example configuration files",
+        ),
+        "dump-instrument": (
+            "ctapipe.tools.dump_instrument.DumpInstrumentTool",
+            "read the subarray description from any file with an EventSource\nand write "
+            "it to a set of output files.",
         ),
     }
 
--- a/ctapipe/__main__.py
+++ b/ctapipe/__main__.py
@@ -1,0 +1,56 @@
+from traitlets.config import Application
+
+
+class MainTool(Application):
+    """
+    Main entry point for ctapipe, provides other tools as subcommands
+    """
+
+    name = "ctapipe"
+
+    subcommands = {
+        "process": (
+            "ctapipe.tools.process.ProcessorTool",
+            "ctapipe event-wise data processing",
+        ),
+        "apply-models": (
+            "ctapipe.tools.apply_models.ApplyModels",
+            "Apply trained machine learning models",
+        ),
+        "train-energy-regressor": (
+            "ctapipe.tools.train_energy_regressor.TrainEnergyRegressor",
+            "Train telescope-type-wise energy regression models",
+        ),
+        "train-particle-classifier": (
+            "ctapipe.tools.train_particle_classifier.TrainParticleClassifier",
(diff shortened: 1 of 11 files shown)


setup.cfg
```diff
@@ -87,6 +87,7 @@ exclude =
 
 [options.entry_points]
 console_scripts =
+    ctapipe = ctapipe.__main__:main
```
LukasNickel, line 90: There is no `main()` right now, which is why this fails to launch

maxnoe: Ah, right, I locally only tested `python -m ctapipe`

kosack:
I remember trying this in the past and there was something in our Tool base class that broke sub-commands.  Does it work now? I don't recall what was missing/broken though.

maxnoe:
At least I didn't find anything immediately wrong:

```
ctapipe process -i dataset://gamma_prod5.simtel.zst -o /tmp/foo.dl1.h5 --progress
```

for example works correctly here.

```
❯ ctapipe --help
Main entry point for ctapipe, provides other tools as subcommands

Subcommands
===========
Subcommands are launched as `ctapipe cmd [args]`. For information on using
subcommand 'cmd', do: `ctapipe cmd -h`.

process
    ctapipe event-wise data processing
apply-models
    Apply trained machine learning models
train-energy-regressor
    Train telescope-type-wise energy regression models
train-particle-classifier
    Train telescope-type-wise binary classficiation models
train-disp-regressor
    Train telescope-type-wise disp regression models for direction reconstruction
info
    Print information about ctapipe and the current installation
quickstart
    Create a directory with example configuration files
```

kosack:
Should make go through all the tool subclasses and substitute:
```
name = "ctapipe-
```
with 
```
name = "ctapipe 
```
So the help text comes out more correctl

ctapipe/__main__.py
```diff
@@ -0,0 +1,63 @@
...
+class MainTool(Application):
+    """
+    Main entry point for ctapipe, provides other tools as subcommands
+    """
+
+    name = "ctapipe"
+    version = __version__
+
+    subcommands = {
+        "process": (
```
kosack, line 15: I wonder if this could be built on-the fly the same way that `ctapipe info --tools` does it?  Or otherwise, maybe just drop the `--tools` option of ctapipe-info if it is no longer useful and replaced with this list. The only advantage is that it is constructed automatically.

ctapipe/tools/info.py
```diff
@@ -38,48 +40,87 @@
...
-
-    info(**vars(args))
+    reconstructors = Bool(
+        False, help="Print available Reconstructor implementations"
+    ).tag(config=True)
+
+    all = Bool(False, help="show all info").tag(config=True)
+
+    flags = {
+        "tools": ({"InfoTool": {"tools": True}}, "show available command-line tools"),
```
kosack, line 70: Oddly, `ctapipe-info --tools` (or really any option other than `--help`) returns nothing, but e.g. `ctapipe info --tools` does what one expects.  Is there something broken with the `ctapipe-info` entry point maybe?

kosack (changes requested):

ctapipe/__main__.py
```diff
@@ -0,0 +1,63 @@
...
+        ),
+        "train-particle-classifier": (
+            "ctapipe.tools.train_particle_classifier.TrainParticleClassifier",
+            "Train telescope-type-wise binary classficiation models",
+        ),
+        "train-disp-regressor": (
+            "ctapipe.tools.train_disp_regressor.TrainDispRegressor",
+            "Train telescope-type-wise disp regression models for direction reconstruction",
+        ),
+        "info": (
```
kosack, line 35: missing `"fileinfo"` tool

ctapipe/__main__.py
maxnoe: can reproduce, will check

maxnoe: It has to do with https://github.com/cta-observatory/ctapipe/issues/656

I used the "documented" approach  in traitlets, the class method `launch_instance`. But since that calls `start`, not our custom `run`, the code does not execute correctly. Fixed by doing `InfoTool().run()`, but it should really work with `InfoTool.launch_instance()`

maxnoe: Scratch that. It was the other way around. I implemented overrided "run" instead of "start"

ctapipe/__main__.py
kosack: I just pushed a few commits to convert `ctapipe-fileinfo` into a `Tool` and include it in this list. So this should be fixed now.

kosack:
A few things still missing here:
* [ ] we lose tab completion. Not a huge problem, but this could be solved by adding a bash/fish completion script or by also including the explicit command-line tools as well as the common tool (like git, where you can use both `git changelog` or `git-changelog`)
* [ ] ctapipe-info --tools lists the "old-style" tool names (`ctapipe-process` instead of `ctapipe process`). In fact, this functionality could actually be dropped, as the main tool already shows you this info.

kosack:
One more issue to solve: commands that use positional arguments don't seem to work in sub-commands. For example:

Works:
```
ctapipe-fileinfo f1.h5 f2.h5
```

Fails:
```
ctapipe fileinfo f1.h5 f2.h5
# the main program fails, as it tries to interpret the last two files as commands.
```

PR #2373, opened 2023-07-04
Update docs

aknierim:
Docs are now built with Sphinx 6.2.1 and the PyData theme

### DIFF
--- a/docs/_static/switcher.json
+++ b/docs/_static/switcher.json
@@ -5,7 +5,7 @@
     "url": "https://ctapipe.readthedocs.io/en/latest/"
   },
   {
-    "name": "v0.19.2 (stable)",
+    "name": "stable",
     "version": "stable",
     "url": "https://ctapipe.readthedocs.io/en/stable/"
   },
--- a/docs/_static/ctapipe.css
+++ b/docs/_static/ctapipe.css
@@ -12,6 +12,27 @@
    }
 }
 
+.version-switcher__container a[data-version-name*="stable"] {
+  background-color: #E9F6EC;
+  color: #28A745;
+}
+
+.version-switcher__container a[data-version-name*="dev"] {
+  background-color: #FDF4EB;
+  color: #EE9040;
+}
+
+html[data-theme="dark"] .version-switcher__container a[data-version-name*="stable"] {
+  background-color: #222924;
+  color: #28A745;
+}
+
+html[data-theme="dark"] .version-switcher__container a[data-version-name*="dev"] {
+  background-color: #332A21;
+  color: #EE9040;
+}
+
+
 /* sphinx-design */
 .sd-card {
   border-radius: 5px;
@@ -44,4 +65,3 @@
 html[data-theme="dark"] .sd-shadow-sm {
   --sd-color-shadow: #6e6e6e;
 }
-


docs/conf.py
```diff
@@ -128,10 +128,15 @@ def setup(app):
     ("py:class", "ctapipe.compat.StrEnum"),
 ]
 
-# temporary fixes to ignore reference warnings and ensure build
+# temporary workaround to ignore reference warnings and ensure build
 nitpick_ignore += [
     ("py:obj", "ctapipe.calib.CameraCalibrator"),
     ("py:obj", "ctapipe.calib.GainSelector"),
+    ("py:obj", "CameraGeometry"),
```
maxnoe, line 135: This is likely just somewhere in the docs missing the FQN. So instead of ignoring a broken ref to `CameraGeometry` here, the reference should be fixed to something like `~ctapipe.instrument.CameraGeometry`

docs/ctapipe_api/calib/index.rst
```diff
@@ -44,6 +48,19 @@ Submodules
...
+   calling automodapi on the camera package.
+
+ctapipe.calib Package
+---------------------
+
+Calibration
+
+.. raw:: html
+
+    <h3>Classes</h3>
```
kosack, line 62: Why use HTML here? You can just use a third-level Restructured Text heading.  That will keep it working for output other than HTML

aknierim: I wanted to keep it consistent with the way `automodapi` outputs the headings. When using a third-level reST heading the color is the primary color and not the normal font color. I can change it to the third-level heading though, if you like.

aknierim: Okay, never mind, I fixed it with a `subsubsection` heading. Sorry, my mistake.

kosack:
For review: here is the [RTD link for this PR](https://ctapipe--2373.org.readthedocs.build/en/2373/)

kosack (changes requested):
A few style comments:
* on the main page can you hide the left sidebar, as it has no information in it? I think there is an option for this
* Do the CEPs belong under "Coding Guidelines?"  The don't really make sense there. Maybe grouped with the changelog?
* Not for this PR (as it's more structure than technical), but we now have too many top-level headings, so most of the things end up in the *More* menu, which is not nice for user-experience or discoverablility. So we should consider restructuring so we have at most 4-5 top-level headings 
image
* Tutorials and examples should come before API docs

maxnoe:
>  but we now have too many top-level headings,

Proposal:

* User Guide
    * Installation
    * Tutorials
    * Examples
* Changelog
    *  CEPs   
* API Docs
* Developer Guide
    *  Installation
    * Coding Guidelines

aknierim:
> on the main page can you hide the left sidebar, as it has no information in it? I think there is an option for this

Yes, that's actually something I'm working on. The problem is, that the left sidebar on the main page only appears when on RTD, thus making it difficult to see how changes are actually affecting it.

> Do the CEPs belong under "Coding Guidelines?"  The don't really make sense there. Maybe grouped with the changelog?

I have not touched that part yet, but I see your point. We could place it below the changelog... See [@maxnoe's comment ](https://github.com/cta-observatory/ctapipe/pull/2373#issuecomment-1628688261).

> Not for this PR (as it's more structure than technical), but we now have too many top-level headings, so most of the things end up in the _More_ menu, which is not nice for user-experience or discoverablility. So we should consider restructuring so we have at most 4-5 top-level headings

Agree. Will open a new PR for that, once I get to work on it.

> Tutorials and examples should come before API docs

Good point, I'll just move the API docs.

maxnoe (changes requested):
The `docs/examples/Tools.ipynb` needs to be cleared of outputs.

Otherwise, this looks fine and I think we should merge this technical part of the docs update and go on with content changes in subsequent PRs.

kosack (approved):

aknierim:
Removed the version switcher from the navbar per request by @maxnoe as it has to be updated by hand every time a new version is released.

docs/conf.py
```diff
@@ -182,32 +189,53 @@ def setup(app):
...
+        {
+            "name": "CTA Observatory",
+            "url": "https://www.cta-observatory.org/",
+            "type": "url",
+            "icon": "https://www.cta-observatory.org/wp-content/themes/ctao/favicon.ico",
+        },
+    ],
+}
+
+html_sidebars = {"**": ["sidebar-nav-bs.html", "sidebar-ethical-ads.html"]}
```
kosack, line 223: This is a bit of a waste of space: there is no content in the sidebar for most of the pages

maxnoe: The version switcher is there. We had an alternative version, but that would require manually curating a json file of versions.

aknierim: Yeah, you're right, but this is currently the only way to get the version switcher from RTD. The way I see it, we currently have three options:

1. We go back and reimplement the dropdown menu I previously added. This however would mean that we have to add any new version to a JSON file by hand.

2. We use the flyout menu from RTD, thus making it necessary to have the sidebar on all pages.

3. We use the flyout menu only on specific pages like the front page and for example the API docs.

I'm not sure which option is the best as I guess all of them have their drawbacks...

kosack: Hmm, this works fine in the vodf pages I created.. Not sure how I configured it, but there the switcher is floating on the lower-right corner: https://vodf.readthedocs.io/en/latest/  (and also in the top-bar)

aknierim: I tried to figure out why the behaviour is so different from your vodf pages today but couldn't find anything. My only guess would be different behaviour when using Sphinx 4.5 and PyData 0.10 for the vodf pages versus the ctapipe docs with Sphinx 6.2.1 and PyData 0.13.3...

aknierim: The version switcher in the navbar on the vodf pages is of course the same one that uses the manually curated json file. But the version switcher from RTD is what boggles my mind. 

At least from what I gathered from the [PyData theme dev docs ](https://pydata-sphinx-theme.readthedocs.io/en/latest/user_guide/readthedocs.html#version-switcher) and the and an [open issue](https://github.com/pydata/pydata-sphinx-theme/issues/705), they are currently working on a way to make the RTD version switcher more customizable, or at least movable from the sidebar. So for now it should actually only appear in the primary sidebar (if not hidden) on the left.

kosack: Who knows? I can try updating the VODF docs and see if it disappears... You're right that it's not quite the same version.

maxnoe:
Okay, maybe we just bite the bullet and have the manual switcher. Sorry for going back and forth.

It looks much better, doesn't have the sidebar space issue and we already have manual steps in the release process like rendering the changelog.

So let's maybe go for that (and add a section on "how to make a release" to the developer documentation).

I can do that as part of #2383 , after we merge here.

docs/_static/switcher.json
```diff
@@ -0,0 +1,72 @@
+[
+  {
+    "name": "dev",
+    "version": "latest",
+    "url": "https://ctapipe.readthedocs.io/en/latest/"
+  },
+  {
+    "name": "v0.19.2 (stable)",
```
maxnoe, line 8: 0.19.2 is repeated, once with (stable) once without. Why?

Stable should be the latest release (0.19.3).  And probably not have a version tag, since we'd need to always update that...

kosack (approved):

kosack:
The only thing I would suggest is that maybe the top-level admonition about the instability is a bit much: it takes up a lot of page space. Just having it as a disclaimer on the main page is probably fine.  Most people understand that a 0.X release is unstable

maxnoe:
> The only thing I would suggest is that maybe the top-level admonition about the instability is a bit much: it takes up a lot of page space.

I said the same, but we discussed that most people probably don't enter through the main page. I don't have a strong opinion on this.

Let's merge here so we have an up-to-date sphinx and a better design and do more content changes in other PRs.

maxnoe (approved):

PR #2375, opened 2023-07-05
Remove dump-triggers, enable ctapipe-process to store only trigger and simulation data

### LINKED ISSUE #2372: Remove ctapipe-dump-triggers
This just writes timestamps of events into a table that is not compatible with the rest of ctapipe. 

If there is a need for such a tool, we should adapt the `ctapipe-process` tool to just write out the trigger information in the datawriter without doing any actual processing, but I doubt anyone really needs this.

maxnoe:
fixes #2372 

Running `ctapipe-process -i <input> -o <output> --no-write-parameters` will result in an output file only containing trigger and simulation information, no any actual processing will happen.


kosack (dismissed):

kosack:
i didn't check, but I guess the merge tool will work with these files, since this info was already merged?

maxnoe:
@kosack Yes, merging works fine

ctapipe/io/datawriter.py
```diff
@@ -616,6 +613,20 @@ def table_name(self, tel_id):
...
+        self._writer.write(
+            table_name="dl1/event/subarray/trigger",
+            containers=[event.index, event.trigger],
+        )
+
+        for tel_id, trigger in event.trigger.tel.items():
+            writer.write(
+                "dl1/event/telescope/trigger", (_get_tel_index(event, tel_id), trigger)
+            )
+
```
StFroese, line 629: Quick question for my understanding.
`self._write_trigger(self._writer, event)` is called in L304, which means z
`writer=self._writer`.
But `self._writer` is already available inside the the function.
In which case do you want to use a different writer with this function?

kosack: It could have been because DataWriter was meant to support other types of TableWriter in the future as well (only HDF5 is currently implemented, but it would be pretty easy to have other formats). But even then, we could assume the TableWriter itself supports handling of N tables.

maxnoe: This was a copy paste mistake, I moved the subarray writing from __call__ where it directly accessed self._writer

Will change.

Since this is all internal anyway, I could also just remove the wrote argument though

mdpunch (approved):
Looks fine, thanks!

maxnoe:
Removed the superflous `writer=` argument from the internal functions, please review agan @kosack @mdpunch

StFroese (approved):

kosack (approved):

Issue #1278, opened 2020-04-14
Do not use astropy sky coordinates for ground and tilted ground frame


maxnoe:
These are not sky coordinates in any sense, so they shouldn't be using astropy frames and `SkyCoords`

maxnoe:
I think we should use 3d coordinates in a well defined earth frame for this, e.g. WGS84 / ITRS, e.g. like astropy does for obsgeoloc, see

kosack:
I think we can close this now that we have the ability to transform GroundFrame to EarthLocation  #2167

maxnoe:
One of the unintended, surprising side effects was reported by @mdpunch  in #2478

mdpunch:
Hi,

I don't know much about the esthetics of the different containers (though I'd have thought that Cartesian relative to the centre of the array would be simplest, with then the centre of the array being defined in a fancier way... whatever).

But if keeping with SkyCoords, it would be nice to give an error if the user does inapplicable things, like using `separation` with the frame being `GroundFrame` (which gives an angle in reponse)?

I have my answer, I'd just like others not to fall in the trap...

// Michael.

maxnoe:
> The main issue is that having a nice container like SkyCoord that aggregates frames and attributes of those frames is nice (even if we currently misuse or fail to use it in many places).

I think we get more weird behavior by using something that has the explicit assumption of describing a fixed point in space to describe something fixed on Earth's surface, as can be seen by the astropy issue linked above.

We can re-use the representations and EarthLocation to get something better, with the behavior we need an none of the weird inherited behavior from skycoord without much effort I think.

mdpunch:
Found this (and the my linked issue) again, while googling to search for how to get the actual time delta between ground and tilted frames.  I had forgotten the use of `.cartesian`, but using it does indeed work:
```
pointing = AltAz(az=az, alt=alt)
subarray_tilted = subarray.tel_coords.transform_to(TiltedGroundFrame(pointing_direction=pointing))
subarray_tilted_distance = subarray_tilted.cartesian-sub_4LSTs_4MSTs.tel_coords.cartesian
subarray_tilted_norm = subarray_tilted_distance.norm()
from astropy.constants import c
sub_4LSTs_4MSTs_tilted_norm/c
```
[...]

PR #2376, opened 2023-07-05
Use int minval instead -1 for invalid telescope ids in SubarrayDescription.tel_index_array, improve docstrings

maxnoe:
-1 is a dangerous index for "invalid" since it will silently result in using the last element when used as index.

Replaced by using int minval, which will result in an IndexError and can be easily filtered by checking `index >= 0`.

### DIFF
--- a/ctapipe/instrument/subarray.py
+++ b/ctapipe/instrument/subarray.py
@@ -173,13 +173,17 @@ class SubarrayDescription:
     @lazyproperty
     def tel_index_array(self):
         """
-        returns an expanded array that maps tel_id to tel_index. I.e. for a given
-        telescope, this array maps the tel_id to a flat index starting at 0 for
-        the first telescope. ``tel_index = tel_id_to_index_array[tel_id]``
-        If the tel_ids are not contiguous, gaps will be filled in by -1.
+        Array of telescope indices that can be indexed by telescope id
+
+        I.e. for a given telescope, this array maps the tel_id to a flat index
+        starting at 0 for the first telescope.
+        ``tel_index = subarray.tel_index_array[tel_id]``
+
+        If the tel_ids are not contiguous, gaps will be filled in by int minval.
         For a more compact representation use the `tel_indices`
         """
-        idx = np.zeros(np.max(self.tel_ids) + 1, dtype=int) - 1  # start with -1
+        invalid = np.iinfo(int).min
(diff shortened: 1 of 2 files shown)


ctapipe/instrument/subarray.py
```diff
@@ -196,8 +201,9 @@ def tel_ids_to_indices(self, tel_ids):
         np.array:
             array of corresponding tel indices
         """
-        tel_ids = np.array(tel_ids, dtype=int, copy=False).ravel()
-        return self.tel_index_array[tel_ids]
+        tel_ids = np.array(tel_ids, dtype=int, copy=False)
+        shape = tel_ids.shape
+        return self.tel_index_array[tel_ids.ravel()].reshape(shape)
```
Tobychev, line 206: Is this reshape fixing some other bug?

Tobychev (commented):
A bit surprised you can change the sentinel value for missing data and not have to update any tests...

ctapipe/instrument/subarray.py
maxnoe: Ah, no, I didn't want to add this here, was by mistake

Tobychev:
> @Tobychev the tests are failing currently....

Yeah, but it doesn't look like the failures are related to the changes? At least many look to be of the type "column 'ExtraTreesRegressor_energy' not found", but that looks unrelated to your changes?

ctapipe/instrument/subarray.py
```diff
@@ -56,13 +56,15 @@ class SubarrayDescription:
     ----------
     name: str
        name of subarray
-    tel_coords: astropy.coordinates.SkyCoord
+    tel_coords: ctapipe.coordinates.GroundFrame
```
Tobychev, line 59: Is this a pure documentation update? Actual code changes in this PR seems to not react to this change, so was the input always supposed to be in `GroundFrame`?

kosack (approved):

ctapipe/instrument/subarray.py
maxnoe: Both a SkyCoord in GroundFrame and a GroundFrame directly work, however internally, we use GroundFrame as 

a) it is slightly faster
b) Calling something on the ground a SkyCoord doesn't really make sense

Also see #1278

Tobychev (approved):

PR #2378, opened 2023-07-06
Add method to write default config to Tool and Component.

ccossou:
After discussions wit Karl and Tomas, I worked on a way to extract default configuration .yaml file, recursively, directly from the class, without having to instanciate anything. 

This is not yet ready for merging, I create the pull request more for discussing what are your thoughts and requests to make this usefull. 

Current issues are:
* We use the default value of EVERY traits in every class in the hierarchy of a given Tool or Component. 
* When default value is undefined, the default config is then non valid and make it difficult to understand what value we're supposed to use, on top of making the yaml file invalid too, because Undefined can't be a valid input. 

Knowing that:
* Will this functionality ever be useful?
* If so, how do we deal with undefined, does that mean that, apart some key parameters such as input file, we force a valid parameter as default?

### DIFF
--- a/ctapipe/core/component.py
+++ b/ctapipe/core/component.py
@@ -9,6 +9,7 @@ from docutils.core import publish_parts
 from traitlets import TraitError
 from traitlets.config import Configurable
 
+from . import config_writer
 from .plugins import detect_and_import_plugins
 
 __all__ = ["non_abstract_children", "Component"]
@@ -273,3 +274,29 @@ class Component(Configurable, metaclass=AbstractConfigurableMeta):
         state["_trait_values"]["parent"] = None
         state["_trait_notifiers"] = {}
         return state
+
+    @classmethod
+    def _get_default_config(cls):
+        """
+
+        :return:
+        """
+        conf = {cls.__name__: cls.traits(cls, config=True)}
+
+        return conf
+
+    @classmethod
+    def write_default_config(cls, outname=None):
+        """return the current configuration as a dict (e.g. the values
+        of all traits, even if they were not set during configuration)
+        """
+
+        if outname is None:
+            outname = f"{cls.__name__}.yml"
+
+        conf = cls._get_default_config()
+
+        conf_repr = config_writer.trait_dict_to_yaml(conf)
+
+        with open(outname, "w") as obj:
+            obj.write(conf_repr)
--- a/ctapipe/core/config_writer.py
+++ b/ctapipe/core/config_writer.py
@@ -1,0 +1,62 @@
+import logging
+import textwrap
+
+log = logging.getLogger(__name__)
+
+
+def trait_dict_to_yaml(conf, conf_repr="", indent_level=0):
+    """
+    Using a dictionnary of traits, will write this configuration to file. Each value is either a subsection or a
+    trait object so that we can extract value, default value and help message
+
+    :param conf: Dictionnary of traits. Architecture reflect what is needed in the yaml file.
+    :param str conf_repr: internal variable used for recursivity. You shouldn't use that parameter
+
+    :return: str representation of conf, ready to store in a .yaml file
+    """
+    indent_str = "  "
+
+    for k, v in conf.items():
+        if isinstance(v, dict):
+            conf_repr += f"{indent_str * indent_level}{k}:\n"
+            conf_repr = trait_dict_to_yaml(v, conf_repr, indent_level=indent_level + 1)
+        else:
+            conf_repr += _trait_to_str(v, indent_level=indent_level)
+
+    return conf_repr
+
+
+def _trait_to_str(trait, help=True, indent_level=0):
+    """
+    Represent a trait in a futur yaml file, given prior information on its position.
+
+    :param key:
+    :param trait:
+    :param help:
+    :param indent_level:
+    :return:
+    """
+    indent_str = "  "
+
+    def commented(text, indent_level=indent_level, width=144):
+        """return a commented, wrapped block."""
+        return textwrap.fill(
+            text,
+            width=width,
+            initial_indent=indent_str * indent_level + "# ",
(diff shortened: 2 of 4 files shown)


ccossou:
In the new commits, I commented every parameter that is "traitlets.Undefined" by default. So this gives the user and idea of the possibilities, without having to changes too many parameters for no added value.

kosack:
can you add a space before each class?  Or better yet, each class has help info as well (a docstring), so you could add that as a comment?

ccossou:
What do you mean by "space"? In the recent commits, I added a newline before every class to make sure it's not close to the previous parameter. 

As for the class comment, this is certainly possible but a bit more difficult to do. I'll add that to the todo-list.

kosack:
Another thing missing is that the options for on-the-fly constructed classes do not appear, e.g. there is an option `image_cleaner_type` which is defined like:
```py
  image_cleaner_type = ComponentName(
        ImageCleaner, default_value="TailcutsImageCleaner"
    ).tag(config=True)
```

You should then show the options for ImageCleaner and all of it's non-abstract subclasses (e.g. `[ImageCleaner, non_abstract_subclasses(ImageCleaner)]`)

ccossou:
I've updated the example file at the top. 

Now:
* Newline before classes
* summary line of docstring is now help message of class section
* help string (either docstring or trait.help is now cleaned of indentation and newline for better formatting of the yml file).

maxnoe:
The traitlets application already has a method `generate_config_file` that seems to do the same, why do we need custom code here?

maxnoe:
See:
https://github.com/ipython/traitlets/blob/c2d04ba8d24f6303b99c9e8e01a1ab16a1551964/traitlets/config/application.py#L1005-L1015

ccossou:
This is a good question. 

Because:
1. This function from traitlets generate a string that correspond to a Python script, not a yaml file. To my knowledge, there's no config generator in .yaml in traitlets.
Example: 
```python
# Configuration file for frac-eff-plots.

c = get_config()  #noqa

#------------------------------------------------------------------------------
# Application(SingletonConfigurable) configuration
#------------------------------------------------------------------------------
## This is an application.

## The date format used by logging formatters for %(asctime)s
#  Default: '%Y-%m-%d %H:%M:%S'
# c.Application.log_datefmt = '%Y-%m-%d %H:%M:%S'

## The Logging format template
#  Default: '[%(name)s]%(highlevel)s %(message)s'
# c.Application.log_format = '[%(name)s]%(highlevel)s %(message)s'
```
2. To a lesser degree, because the function you cited need an instance, while this shouldn't be necessary, a class should be enough.

ccossou:
Discussed today at the ctapipe meeting, the workaround with dynamic default is to detect it's dynamic, then write "it's dynamic, check documentation". 

I need to dig into this again because I've mostly forgotten where I was with this.

ccossou:
For now, I have an issue that I don't know how to solve. 

I comment traits when the default values is invalid. Problem is, for EventSource for instance, all traits are invalid, and I end up with an empty Section that fail on read with this error:
> ValueError: values whose keys begin with an uppercase char must be Config instances: 'EventSource', None

I tried to parse the string after creating it, but not only is this ugly design, this is also trickier than expected, because you need to check per indentation level, and that means the function need to be recursive to be doable. 

But this is also tricky to do when reading the traits, because you need to check ALL traits for that sections before even writing the section, to know if you need to comment the section or not. 

I'm open to suggestions on this one. 

PS: I already solved the dynamic default apparently, months ago, I didn't even recall doing it.

maxnoe:
I think something went wrong in your last push...

maxnoe:
> I comment traits when the default values is invalid. Problem is, for EventSource for instance,

While I understand the general issue, I don't understand why it happens for `EventSource`.
The defaults for `max_events` and `allowed_tels` should be perfectly valid.

ccossou:
@maxnoe What went wrong? What is the issue that you noticed?

ccossou:
In EventSource:
* default for max_events is None. While this is allowed, this is not a value accepted in the .yaml (or else I'm missing something)
* allowed_tels has as a default CInts(). It should be 0, but I'm seeing None (and I have to understand why)

maxnoe:
> *   default for max_events is None. While this is allowed, this is not a value accepted in the .yaml (or else I'm missing something)
> *   allowed_tels also has a default set to None in EventSource (ctapipe.io.eventsource.py line 104)


The yaml representation of `None` is `null`:

```
In [1]: import yaml
In [2]: config = {"EventSource": {"max_events": None, "allowed_tels": None}}
In [3]: print(yaml.dump(config))
EventSource:
  allowed_tels: null
  max_events: null
```

maxnoe:
Checking the ruamel.yaml docs again, this is how you can programatically create a commented yaml document:

```
from ruamel.yaml import YAML, CommentedMap
import sys

yaml = YAML()

doc = CommentedMap()
doc["EventSource"] = CommentedMap()
doc.yaml_set_comment_before_after_key("EventSource", before="Config of EventSources")

doc["EventSource"]["max_events"] = None
doc["EventSource"].yaml_add_eol_comment("Maximum number of events", key="max_events")
doc["EventSource"]["allowed_tels"] = None
doc["EventSource"].yaml_add_eol_comment("Allowed telescopes", key="allowed_tels")
```

Output:

```
# Config of EventSources
EventSource:
  max_events:  # Maximum number of events
  allowed_tels: # Allowed telescopes
```

ccossou:
Thanks. It took me time to understand that what I needed was CommentedMap, because I couldn't find that information from the documentation alone. 

But I've used yaml_set_comment_before_after_key even for the traits, as sometimes the comments are big, and assumed it would be problematic at the end of the line. 

I'm still in the debug phase, but even after that, comments are not indented as the key they are attached to, so that will be my next item on the list, once I can get it to run at all.

PR #2379, opened 2023-07-06
Add mamba build for python 3.11

kosack:
I've been using python 3.11 with mamba for a month or so now, so it should be stable enough to test with (currently we only test 3.11 with pip)


maxnoe:
I made the same change in #2342

kosack:
This was so we can test the docs build for #2377 but I think now I realize the reason the env failed for me was just the numba dependency

PR #2383, opened 2023-07-17
Update doc structure and dev guide

maxnoe:


docs/getting_started/index.rst
```diff
@@ -57,16 +79,35 @@ ctapipe repo (make sure you put in your own username there):
...
+
+.. code-block:: console
+
+   $ git fetch upstream
+
+Then, update a local branch using
+
+.. code-block:: console
+
+   $ git merge upstream/main
```
kosack, line 101: Why go for fetch+merge here, rather than just `git pull upstream main`? Should we have them set the pull policy  to _merge_ for example?

maxnoe: I think `git pull upstream main` does not fetch new tags, which is e.g. why @Tobychev had issues with the automatic versioning in his fork.

maxnoe: Just verified. `git fetch upstream` fetches new tags, `git merge upstream main` does not.

kosack: `git pull` has a `--tags` option, which according to the docs is on by default, though testing it it seems not to be the case. Anyhow, I agree that's confusing that `pull`, which should be `fetch + merge` acts slightly differently, so just asking to do it in two steps is probably fine.  

```man
       --no-tags
           By default, tags that point at objects that are downloaded from the remote repository are fetched
           and stored locally. This option disables this automatic tag following. The default behavior for a
           remote may be specified with the remote.<name>.tagOpt setting. See git-config(1).
     -t, --tags
           Fetch all tags from the remote (i.e., fetch remote tags refs/tags/* into local tags with the same
           name), in addition to whatever else would otherwise be fetched. Using this option alone does not
           subject tags to pruning, even if --prune is used (though tags may be pruned anyway if they are also
           the destination of an explicit refspec; see --prune).
```

docs/getting_started/index.rst
```diff
@@ -57,16 +79,35 @@ ctapipe repo (make sure you put in your own username there):
     $ cd ctapipe
 
 
-You now need to tell Git that this repo where the master CTA version is:
+You now need to tell Git that this repo where the main ctapipe repository is:
```
LukasNickel, line 82: Sounds like the _that this repo_ is superflous here.

docs/getting_started/index.rst
```diff
@@ -78,42 +119,36 @@ Change to the directory where you cloned ``ctapipe``, and type:
 .. code-block:: console
 
 
-    $ conda env create -n cta-dev -f environment.yml
+    $ mamba env create -f environment.yml
```
LukasNickel, line 122: Isn't it just `mamba create` with mamba?

docs/getting_started/index.rst
```diff
@@ -78,42 +119,36 @@ Change to the directory where you cloned ``ctapipe``, and type:
...
 +++++++++++++++++++++++++++++++++++++
 
 Now setup this cloned version for development. The following command
-will make symlinks in your python library directory to your ctapipe
-installation (it creates a ``.pth`` file, there is no need to set
-PYTHONPATH, in fact it should be blank to avoid other problems). From
-then on, all the ctapipe binaries and the library itself will be
+will make use the editable installation feature of python packages.
+From then on, all the ctapipe executables and the library itself will be
 usable from anywhere.
```
LukasNickel, line 144: as long as you have the `cta-dev` enviroment activated.

docs/getting_started/index.rst
```diff
@@ -144,42 +179,50 @@ try running some command line tools:
...
-You should always create a branch when developing some new code (unless it is
-a very small change).  Generally make a new branch for each new feature, so
-that you can make pull-requests for each one separately and not mix code
-from each.  Remember that ``git switch <name>`` switches between branches,
+You should always create a new branch when developing some new code.
+Make a new branch for each new feature, so that you can make pull-requests
+for each one separately and not mix code from each.
+Remember that ``git switch <name>`` switches between branches,
 ``git switch -c <name>`` creates a new branch, and ``git branch`` on it's own
 will tell you which branches are available and which one you are currently on.
```
LukasNickel, line 208: Should we note that a newish git is required for that and link to `checkout` just in case?

docs/getting_started/index.rst
```diff
@@ -16,6 +16,28 @@ We strongly recommend using the `mambaforge conda distribution <https://github.c
...
 You can use |python_requires| or above.
 
+
+------------------------------------------
+Forking vs. Working in the Main Repository
+------------------------------------------
+If you are a member of CTA (Consortium or Observatory), or
+otherwise a regular contributor to ctapipe, the maintainers can give you
+access too the main repository at ``cta-observatory/ctapipe``.
+Working directly in the main repository has two main advantages over working
```
kosack, line 26: omit *working*

docs/getting_started/index.rst
kosack: Maybe better: *If you are using a fork, the default remote repository will be the one in your user account on GitHub. To make sure you are in sync with the official copy in the cta-observatory account, you need to add an upstream remote as follows:*

docs/getting_started/index.rst
kosack: Could add a [footnote](https://www.sphinx-doc.org/en/master/usage/restructuredtext/basics.html#footnotes)

docs/getting_started/index.rst
```diff
@@ -227,12 +270,19 @@ sub-module), check the style, and make sure the docs render correctly
...
 
-(sometimes refered to as
-"publishing" since it becomes public, but only in your fork) by running
+The first time you push a new branch, you need to specify to which remote the branch
+should be pushed. Normally this will be ``origin``:
 
-.. code-block:: sh
+.. code-block:: console
+
+   $ git push -u origin implement_feature_1
```
kosack, line 278: could also say to set 
```
git config --global push.autoSetupRemote true
```
so this is done automatically

kosack (changes requested):

docs/getting_started/index.rst
maxnoe: That's a good addtion, I adapted this slightly to not imply that by merely adding the remote, things are synchronized.

docs/getting_started/index.rst
maxnoe: Added a footnote about this

docs/getting_started/index.rst
```diff
@@ -144,42 +186,50 @@ try running some command line tools:
...
 ---------------------------------------
 
-We adhere to the PEP8 coding style (see our :doc:`/development/style-guide`).
-To enforce this, setup the
+We are using the ``black`` and ``isort`` auto-formatters for automatic 
+adherence to the code style (see our :doc:`/development/style-guide`).
+To enforce running these tools whenever you make a commit, setup the
 `pre-commit hook <https://pre-commit.com/>`_::
 
     $ pre-commit install
```
Tobychev, line 208: Suggested addition about how the pre-commit hook will block your commit. Something like:

> The pre-commit hook will then execute the tools with the same settings as when the a pull request is checked on github, and if any problems are reported the commit will be rejected. You then have to fix the reported issues before tying to commit again. 
> Note that a common problem is forgetting to run the `black` code formatter, and that whenever this was the only problem found, simply adding the changes resulting from the pre-commit hook to the commit will result in your changes being accepted.

maxnoe:
I update the structure now to follow what we discussed in #2373, specifically the comment here

https://github.com/cta-observatory/ctapipe/pull/2373#issuecomment-1628688261

Please have a look.

The developer guide now has syncing tabs for the commands depending on whether you are on a fork or work in the main repository.

kosack (approved):
This structure looks much better. Thanks

docs/developer-guide/getting-started.rst
```diff
@@ -0,0 +1,458 @@
...
+
+We are using the ``black`` and ``isort`` auto-formatters for automatic 
+adherence to the code style (see our :doc:`/developer-guide/style-guide`).
+To enforce running these tools whenever you make a commit, setup the
+`pre-commit hook <https://pre-commit.com/>`_:
+
+.. code-block:: console
+
+    $ pre-commit install
+
```
Tobychev, line 126: Suggested addition about how the pre-commit hook will block your commit. Something like:

> The pre-commit hook will then execute the tools with the same settings as when the a pull request is checked on github, and if any problems are reported the commit will be rejected. You then have to fix the reported issues before tying to commit again.
>     Note that a common problem is forgetting to run the black code formatter, and that whenever this was the only problem found, simply adding the changes resulting from the pre-commit hook to the commit will result in your changes being accepted.
>

maxnoe: Done!

Tobychev (approved):

kosack (approved):

PR #2388, opened 2023-08-24
Bug fix in full numeric pixel likelihood calculation

gschwefer:
This fixes a bug in the implementation of the full numeric pixel likelihood from [de Naurois 2009 ](https://arxiv.org/pdf/0907.2610.pdf), eq 21, in `neg_log_likelihood_numeric`. The approximate likelihood in `neg_log_likelihood_approx` is adjusted for consistency, which is important given they are combined in the `neg_log_likelihood` function. This was not caught in the tests because these also had a bug, which is now fixed too.

I will also implement the fix in the ImPACT branch where the overall pixel likelihood interface changes.

### DIFF
--- a/ctapipe/image/pixel_likelihood.py
+++ b/ctapipe/image/pixel_likelihood.py
@@ -30,6 +30,7 @@ TODO:
 import numpy as np
 from scipy.integrate import quad
 from scipy.stats import poisson
+from scipy.special import factorial
 
 __all__ = [
     "neg_log_likelihood_approx",
@@ -95,13 +96,14 @@ def neg_log_likelihood_approx(image, prediction, spe_width, pedestal):
     """
     theta = pedestal**2 + prediction * (1 + spe_width**2)
 
-    neg_log_l = np.log(theta + EPSILON) + (image - prediction) ** 2 / theta
+    # This is really 2 times the full log likelihood
+    neg_log_l = np.log(2 * np.pi * theta + EPSILON) + (image - prediction) ** 2 / theta
 
     return np.sum(neg_log_l)
 
 
 def neg_log_likelihood_numeric(
-    image, prediction, spe_width, pedestal, confidence=(0.001, 0.999)
+    image, prediction, spe_width, pedestal, confidence=0.999
 ):
     """
     Calculate likelihood of prediction given the measured signal,
@@ -131,7 +133,7 @@ def neg_log_likelihood_numeric(
 
     likelihood = epsilon
 
-    ns = np.arange(*poisson(np.max(prediction)).ppf(confidence))
+    ns = np.arange(poisson(np.max(prediction)).ppf(confidence) + 1)
 
     ns = ns[ns >= 0]
 
@@ -140,7 +142,8 @@ def neg_log_likelihood_numeric(
         _l = (
             prediction**n
             * np.exp(-prediction)
-            / theta
+            / np.sqrt(2 * np.pi * theta)
+            / factorial(n)
             * np.exp(-((image - n) ** 2) / (2 * theta))
         )
         likelihood += _l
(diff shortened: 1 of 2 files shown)


ctapipe/image/pixel_likelihood.py
```diff
@@ -131,7 +135,7 @@ def neg_log_likelihood_numeric(
 
     likelihood = epsilon
 
-    ns = np.arange(*poisson(np.max(prediction)).ppf(confidence))
+    ns = np.arange(poisson(np.max(prediction)).ppf(confidence) + 1)
```
Tobychev, line 138: What is this change doing? It is not obviously related to the typos in the expressions you are fixing elsewhere.

gschwefer: I removed the `*` and changed the `confidence` argument to only give the upper end of the confidence range (Will change docstring btw.). That is because `ns`  needs to always start at 0, and if `np.max(prediction)` is too large and a lower bound is set, that is not the case. the `+1` is just an extra safety margin.

gschwefer:
What exactly is isort complaining about here?

Tobychev:
> What exactly is isort complaining about here?

Probably you haven't sorted your imports according to the standard scriptures of `isort`, the easiest fix is just to run `isort` locally and commit the changes it produces. 

Do note that it's not actually `isort` itself that complains, it just modifies the files as part of the CI, but the files being modified by `isort` counts as a failure, so you never get to actually hear what it did. Thus the advice to just run it yourself on your own copy.

ctapipe/image/pixel_likelihood.py
```diff
@@ -121,8 +120,9 @@ def neg_log_likelihood_numeric(
         Width of single p.e. peak (:math:`σ_γ`).
     pedestal: ndarray
         Width of pedestal (:math:`σ_p`).
-    confidence: tuple(float, float), 0 < x < 1
-        Confidence interval of poisson integration.
+    confidence: loat, 0 < x < 1
```
Tobychev, line 123: probably mean `float`

gschwefer: Yes

maxnoe:
@gschwefer If you enable the `pre-commit` hooks, this will automatically run when you commit, see:

https://ctapipe.readthedocs.io/en/latest/getting_started/index.html#developing-a-new-feature-or-code-change

maxnoe:
> This fixes a bug in the implementation of the full numeric pixel likelihood 

What is the actual bug here? 

I see that you re-introduce terms that are constant under minimization of the likelihood, which is probably the reason why they were left out.

I guess using factorial is pretty slow, here is the optimized implementation of the poisson related functions in numba-stats:

gschwefer:
Ok, so these are a couple of different things: The terms in the `neg_log_likelihood_numeric` that I reintroduce are actually not simple constants to the log likelihood because they are summed over in the `n_signal` summation and depend on `n`. Therefore, when the log is taken, they aren't just separate additive terms and therefore make an actual difference. This was not caught in the unit tests because an absolute value was missing in the tests of the relative differences that allowed large negative values to pass.

ctapipe/image/pixel_likelihood.py
gschwefer: Fixed it

maxnoe:
> The factorial I'm happy to switch out.

Measure before :)

gschwefer:
> > The factorial I'm happy to switch out.
> 
> Measure before :)

Good hint...the explicit implementation is indeed faster 
image

maxnoe:
For arrays of k or lambda, `numba_stats.poisson.logpmf` was a lot faster (factor of 2 or more), maybe worth introducing this additional dependency.

maxnoe:
Indeed, I only tested arrays for k, I'll open an issue

Tobychev (approved):

maxnoe (approved):

PR #2393, opened 2023-09-05
Test eventio iter_all_events branch

maxnoe:
This might become eventio 2.0 due to the breaking changes but enables us to optionally also include non-triggered events in the `SimTelEventSource`.


maxnoe:
Closing, will test locally and we can update here once released in eventio

PR #2396, opened 2023-09-15
Drop camera frame

StFroese:
As you may remember, I already started this here #2141. A lot happened since the last time (i.e. moving the muon processing), so I decided to give this another try on a fresh branch to avoid rebasing

### DIFF
--- a/docs/changes/2396.api.rst
+++ b/docs/changes/2396.api.rst
@@ -1,0 +1,7 @@
+Drop support for computing image parameters in ``CameraFrame`` (meters on the camera plane).
+Going forward, ctapipe only implements computing image parameters in telescope frame (degrees on the sky).
+
+* ``ctapipe-process`` and ``ImageProcessor`` no longer have options to chose between ``CameraFrame`` and ``TelescopeFrame```
+* All image parameter functions do not support inputs in length units anymore
+* If using these low-level functions directly in your code, you will need to transform your camera geometries into ``TelescopeFrame``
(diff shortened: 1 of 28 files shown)


StFroese:
still one issue in tutorials/ctapipe_handson.py, seems like I'm zooming into the wrong region because of the transformation to `TelescopeFrame`, I'll fix it later

maxnoe:
New traitlets version seems to come with more broken sphinx stuff in their docs, need to add to the ignore rules

kosack:
Great! This reminds me I should look back at #2279  as well once this is finished.

maxnoe:
I added the traitlets docs fixes and the changelog

kosack (dismissed):

StFroese:
Fixed a notebook

maxnoe:
We discussed today in the meeting that this is as much a breaking change or more than the container restructuring discussed in CEP-2, so we decided that we will also write a CEP for this change.

Technically, this looks already fine and is ready to merge.

maxnoe:
@StFroese could you update this fixing the conflict and getting the fixed codecov in?

maxnoe (dismissed):

kosack (dismissed):
This is ready, but seem to need to be rebased on main

kosack (dismissed):

maxnoe (dismissed):

maxnoe:
Note that we wrote CEP 3 for this change and this is still only in "proposed" state.

I think we should just move it to "accepted" by now, but the procedure is not completely clear

kosack (approved):

maxnoe (approved):

maxnoe:
I'd like to make a bugfix 0.25.1 for #2741 but then it's good to go in.

PR #2397, opened 2023-09-19
Use intmin for invalid pixel positions, allow empty arguments

### LINKED ISSUE #1828: CameraGeometry position_to_pix_index returns -1 for nan input
https://github.com/cta-observatory/ctapipe/blob/39762693534f7dc9b884684c3f82e153bd9da560/ctapipe/instrument/camera/geometry.py#L859

For example, of one uses it on image centroid coordinates to make a "pixelized" 2d histogram, all non-reconstructed images end up in the highest-index pixel (=-1). It would be better to return np.nan.

maxnoe:
Fixes #1828

### DIFF
--- a/docs/changes/2397.api.rst
+++ b/docs/changes/2397.api.rst
@@ -1,0 +1,8 @@
+``CameraGeometry.position_to_pix_index`` will now return the minimum integer value for invalid
+pixel coordinates instead of -1 due to the danger of using -1 as an index in python accessing
+the last element of a data array for invalid pixels.
+The function will now also no longer raise an error if the arguments are empty arrays and instead
+just return an empty index array.
+The function will also no longer log a warning in case of coordinates that do not match a camera pixel.
+The function is very low-level and if not finding a pixel at the tested position warrants a warning or
+is expected will depend on the calling code.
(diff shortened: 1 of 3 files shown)


docs/changes/2397.api.rst
```diff
@@ -0,0 +1,8 @@
+``CameraGeometry.position_to_pix_index`` will no return the minimum integer value for invalid
```
Tobychev, line 1: should it be "now" instead of no?

docs/changes/2397.api.rst
```diff
@@ -0,0 +1,8 @@
...
+pixel coordinates instead of -1 due to the danger of using -1 as an index in python accessing
+the last element of a data array for invalid pixels.
+The function will no also no longer raise an error if the arguments are empty arrays and instead
```
Tobychev, line 4: should the first no be "now" instead?

docs/changes/2397.api.rst
```diff
@@ -0,0 +1,8 @@
...
+also just return an empty index array.
```
Tobychev, line 5: not sure if the "also" should be here, the function always returns one variable, not the the optional extra value implied by "also".

Tobychev:
Also, this seems to have encountered that same problem with traitlets and the docs that you fixed for #2396

maxnoe:
@Tobychev yes, thanks, I opened a PR fixing that so we can get tht in fast: #2399

Tobychev (approved):

LukasNickel (approved):

PR #2399, opened 2023-09-20
Fix docs by adding more traitlets nitpick ignores

maxnoe:
Cherry picked the commit from #2396 so we can get this in faster for all open PRs.


Tobychev (approved):

kosack (approved):

maxnoe:
Test failure was due to a new buggy and now fixed setuptools scm release, rerun should fix it

PR #2401, opened 2023-09-20
A time clustering algorithm for image cleaning

clara-escanuela:


ctapipe/image/cleaning.py
```diff
@@ -111,6 +113,67 @@ def tailcuts_clean(
...
+    time,
+    pedestal=1.6,
+    n_noise=2.5,
+    minpts=5,
+    eps=1.0,
+    t_scale=4.0,
+    d_scale=0.25,
+):
+    """
+    Clean an image by selecting pixels which pass a time clustering algorithm using DBSCAN.
```
maxnoe, line 128: There needs to be an empty line after the initial title line here.

We should add a pre-commit linter for this...

ctapipe/image/cleaning.py
```diff
@@ -111,6 +113,67 @@ def tailcuts_clean(
...
+    Previously used for HESS [timecleaning]_.
```
maxnoe, line 129: Empty line before section title

ctapipe/image/cleaning.py
```diff
@@ -111,6 +113,67 @@ def tailcuts_clean(
...
+
+    arr = np.zeros(len(image), dtype=float)
+    arr[~precut_mask] = -1
+
+    pix_x = geom.pix_x.value[precut_mask] / d_scale
+    pix_y = geom.pix_y.value[precut_mask] / d_scale
+
+    X = np.column_stack((time[precut_mask] / t_scale, pix_x, pix_y))
+
+    db = DBSCAN(eps=eps, min_samples=minpts).fit(X)
```
maxnoe, line 165: slightly shorter: `labels = DBSCAN(...).fit_predict(X)`.

maxnoe: Shouldn't this also use `sample_weights=image`? Otherwise you are using image only for the initial cut

clara-escanuela: I tried `sample_weight`  before. I used this weighting:
$\frac{1}{1+exp\left(-\frac{image+s}{d}\right)}$
It introduces two new parameters but I can add weights as an option

maxnoe:

ctapipe/image/cleaning.py
clara-escanuela: I have been thinking about this and I think we could introduce a fourth dimension in the cluster, which is 1/SNR, SNR is the signal-to-noise ratio, image/pedestal. Using just image does not work as the algorithm finds clusters with similar data points but if you plot 1/image, we find clusters:
image

clara-escanuela: It does not seem to make any difference though

kosack:
Not sure if it's interesting, but they just introduced an [improved DBSCAN algorithm](https://scikit-learn.org/stable/auto_examples/release_highlights/plot_release_highlights_1_3_0.html#hdbscan-hierarchical-density-based-clustering) in scikit-learn

ctapipe/image/cleaning.py
```diff
@@ -532,6 +596,49 @@ def __call__(
...
+        """Apply HESS image cleaning. see ImageCleaner.__call__()"""
+        return time_clustering(
+            geom=self.subarray.tel[tel_id].camera.geometry,
+            image=image,
+            time=arrival_times,
+            eps=self.eps.tel[tel_id],
+            n_noise=self.noise_cut.tel[tel_id],
+            d_scale=self.d_scale.tel[tel_id],
+            t_scale=self.t_scale.tel[tel_id],
+            pedestal=self.pedestal.tel[tel_id],
```
kosack, line 637: I think here you should really use the correct pedestal, not some hard-coded value. There is for example

```py3
ped = event.mon.tel[7].calibration.pedestal_per_sample
```
That is per channel, so you need to use the high/low gain switch mask.  We probably should properly process simulated pedestals into DL1 information though, so you don't have to do this.

ctapipe/image/cleaning.py
```diff
@@ -111,6 +113,68 @@ def tailcuts_clean(
...
+        pedestal width for each pixel. This is just one number for now.
+    n_noise: array
+        Noise cut
+    minpts: int
+        Minimum number of points to consider a cluster
+    eps: float
+        Distance in dbscan
+    t_scale: float
+        Time scale in ns
+    d_scale: float
```
kosack, line 150: I'd rename these scale factors `time_scale_ns` and `space_scale_m` (or temporal/spatial if you prefer) to be more clear.  Also, might describe why they are there to begin with in the docs (I guess to get the values as close to 1.0 as possible? Does that help the DBSCAN algorithm?).

kosack (changes requested):

ctapipe/image/cleaning.py
kosack: Could also use astropy units here

ctapipe/image/cleaning.py
clara-escanuela: Maybe I should change the name but what I need is the reconstructed noise per pixel in units of pe. Is that available?

kosack: Yeah, i realized that just after posting... the pedestals themselves are not what you need.  In the simulations currently we don't have the variances, but this is something we really need to fix.  Either we should properly compute pedestals and not rely on the ones given in the sims (which are technically inputs, not outputs), or see if somewhere the variances are already included or ask for them to be included.  That's not just or this PR, but for any "realistic" cleaning method, e.g. the Whipple-10m-style cleaning that is just tailcuts but with the thresholds expressed in pedestal variance units.

kosack: You should definitely change the name, since "pedestal" is the mean value of the noise, not the dispersion.   Here, you are assuming it's Gaussian, right? And you are using it as a standard deviation, not variance, so maybe `pedestal_std`

clara-escanuela:
This is the basic implementation of the method. I think we could finish it soon and then if we need to add some more details, which is possible, we could do so. I am just wondering about unit tests

kosack (approved):

ctapipe/image/cleaning.py
```diff
@@ -111,6 +113,74 @@ def tailcuts_clean(
...
+    pix_x = geom.pix_x.value[precut_mask] / space_scale_m
+    pix_y = geom.pix_y.value[precut_mask] / space_scale_m
+
+    X = np.column_stack((time[precut_mask] / time_scale_ns, pix_x, pix_y))
+
+    labels = DBSCAN(eps=eps, min_samples=minpts).fit_predict(X)
+
+    # no_clusters = len(np.unique(labels))-1  # Could be used for gh separation
+
+    y = np.array(arr[(arr == 0)])
```
Tobychev, line 177: consider removing `y` and `arr` entirely and instead using:
```
mask = np.zeros(len(image), dtype=bool)
mask[precut_mask] = (labels > -1)
```

ctapipe/image/cleaning.py
```diff
@@ -111,6 +113,74 @@ def tailcuts_clean(
...
+    image,
+    time,
+    minpts=5,
+    eps=1.0,
+    time_scale_ns=4.0,
+    space_scale_m=0.25,
+    hard_cut_pe=4,
+):
+    """
+
```
maxnoe, line 127: No empty line here

ctapipe/image/cleaning.py
```diff
@@ -111,6 +113,74 @@ def tailcuts_clean(
...
+    Returns
+    -------
+    A boolean mask of *clean* pixels.
+    """
+    precut_mask = image > hard_cut_pe
+
+    arr = np.zeros(len(image), dtype=float)
+    arr[~precut_mask] = -1
+
+    pix_x = geom.pix_x.value[precut_mask] / space_scale_m
```
maxnoe, line 168: space_scale_m seems to imply that `CameraGeometry` is in `CameraFrame`, however we are moving away from that.

Let's either call it just `space_scale` or actually use the camera geometry / camera readout information to define sensible values regardless of the frame.

ctapipe/image/cleaning.py
```diff
@@ -532,6 +602,43 @@ def __call__(
         )
 
 
+class TimeCleaner(ImageCleaner):
+    """
+    Clean images using the time clustering cleaning method
+    """
+
+    space_scale_m = FloatTelescopeParameter(
+        default_value=0.25, help="Pixel space scaling parameter in m"
```
maxnoe, line 611: Can we determine the scale parameters from the informatino in `CameraDescription`?

clara-escanuela: I am not sure. That parameter needs optimization, it could 2x or 3x the pixel spacing of the camera and teh optimized number depends on the telescope

Tobychev:
@kosack @maxnoe This PR seems to have picked up quite a few unrelated changes, presumably from an attempt to rebase on `main`, what is the preferred way to deal with this?

maxnoe:
I am not sure what happened here. 

@clara-escanuela What git commands did you run?

I see issues like this one from time to time but never could figure out what mistake leads to these.

The proper way of updating a branch is (assuming upstream points to cta-observatory/ctapipe and origin points to clara-escanuela/ctapipe)

```
$ git fetch upstream # get latest changes from main repo
$ git rebase upstream/main
# fix any conflicts that might occure by editing and then running git add / git rebase --continue
```

I think could locally salvage this branch here by cherry-picking the commits that seem to be originally from @clara-escanuela on this branch after resetting the branch hard to `upstream/main`.

```
$ git reset --hard upstream/main
$ git cherry-pick 37e5ea864 55cf8bafb 7e8daf8d4 895ff9774 a9f81e99d 357d02b51 722dfcf17 3c799aaee b9ab65e67 afc170480
```

clara-escanuela:
I will close this PR and start it again. It is too messy.

PR #2402, opened 2023-09-21
Fix empty reference location

kosack:
This adds a fallback dummy location to subarray.reference_location if the metadata in the simtel file doesn't exist.  The dummy location is on "null island" (lat=0, lon=0), but at the correct height as read from the simulation_config.

This was a problem for pre-Prod6 files, which didn't contain the metadata necessary to determine the reference location.


Tobychev (dismissed):

Tobychev:
Would it be possible to add a test that checks "proper" initialisation when metadata is missing?

Tobychev (dismissed):

ctapipe/io/simteleventsource.py
```diff
@@ -513,15 +511,14 @@ def __init__(self, input_url=Undefined, config=None, parent=None, **kwargs):
         )
         if self.back_seekable and self.is_stream:
             raise IOError("back seekable was required but not possible for inputfile")
-
-        self._subarray_info = self.prepare_subarray_info(
-            self.file_.telescope_descriptions, self.file_.header
-        )
```
maxnoe, line 519: Any particular reason for this move?

maxnoe:

ctapipe/io/simteleventsource.py
kosack: Yes, it needs to come after parse_simulation_header, otherwise I can't use the SimulationConfigurationContainer

Tobychev (approved):

maxnoe:
One of the example notebooks no fails because it tries to print a subarray without a reference location:

```
WARNING: /home/runner/work/ctapipe/ctapipe/examples/core/InstrumentDescription.py failed to execute correctly: Traceback (most recent call last):
  File "/home/runner/work/ctapipe/ctapipe/examples/core/InstrumentDescription.py", line 49, in <module>
    newsub.info()
  File "/home/runner/work/ctapipe/ctapipe/ctapipe/instrument/subarray.py", line 129, in info
    printer(f"Height   : {self.reference_location.geodetic.height:.2f}")
AttributeError: 'NoneType' object has no attribute 'geodetic'
```

maxnoe (dismissed):

kosack:
> One of the example notebooks no fails because it tries to print a subarray without a reference location:

Seems it was `select_subarray()` not propegating the reference location. Fixed.  Also, I now changed SubarrayDescription's constructor to require the reference location (and telescope info), to avoid that it is ever set incorrectly.

Tobychev (approved):

Tobychev (approved):

maxnoe (approved):

PR #2403, opened 2023-09-21
Fix h_max definition

### LINKED ISSUE #2395: h_max is inconsistent between HillasIntersection and HillasReconstructor
Both HillasReconstructor and HillasIntersection calculate a quantity h_max and return it in the ReconstructedGeometryContainer, but they are inconistent in the physical meaning of h_max: For HillasIntersection, it is the vertical height of the shower maximum above the array, for HillasReconstructor it is the distance of the shower maximum location from the array center, i.e. kind of a "slant height" 

Whatever h_max really should be, it should of course be the same everywhere.

kosack:
Fixes #2395 

* In HillasReconstructor, Define h_max as the vertical height *above sea level* (including the observatory height) of the shower-max point, and not the distance from the array which was there before. 
* In HillasIntersection, use the correct observatory height value

When #2402  is merged, I will remove the checks for missing reference_location

### DIFF
--- a/ctapipe/containers.py
+++ b/ctapipe/containers.py
@@ -922,7 +922,7 @@ class ReconstructedGeometryContainer(Container):
     )
     h_max = Field(
         nan * u.m,
-        "reconstructed vertical height (above sea level) of the shower maximum",
+        "reconstructed vertical height above sea level of the shower maximum",
         unit=u.m,
     )
     h_max_uncert = Field(nan * u.m, "uncertainty of h_max", unit=u.m)
@@ -930,7 +930,7 @@ class ReconstructedGeometryContainer(Container):
     is_valid = Field(
         False,
         (
-            "direction validity flag. True if the shower direction"
+            "Geometry validity flag. True if the shower geometry"
             "was properly reconstructed by the algorithm"
         ),
     )
--- a/ctapipe/containers.py
+++ b/ctapipe/containers.py
@@ -920,8 +920,13 @@ class ReconstructedGeometryContainer(Container):
         "reconstructed core position uncertainty along tilted frame Y axis",
         unit=u.m,
     )
-    h_max = Field(nan * u.m, "reconstructed height of the shower maximum", unit=u.m)
+    h_max = Field(
+        nan * u.m,
+        "reconstructed vertical height (above sea level) of the shower maximum",
+        unit=u.m,
+    )
     h_max_uncert = Field(nan * u.m, "uncertainty of h_max", unit=u.m)
+
     is_valid = Field(
         False,
         (
--- a/ctapipe/reco/hillas_intersection.py
+++ b/ctapipe/reco/hillas_intersection.py
@@ -261,19 +261,16 @@ class HillasIntersection(HillasGeometryReconstructor):
         tilt = SkyCoord(x=core_x * u.m, y=core_y * u.m, z=0 * u.m, frame=tilted_frame)
         grd = project_to_ground(tilt)
 
-        if self.subarray.reference_location is not None:
-            h_max = self.reconstruct_h_max(
-                nom.fov_lon,
-                nom.fov_lat,
-                tilt.x,
-                tilt.y,
-                hillas_dict_mod,
-                tel_x,
-                tel_y,
-                90 * u.deg - array_pointing.alt,
-            )
-        else:
-            h_max = u.Quantity(np.nan, u.m)
+        h_max = self.reconstruct_h_max(
+            nom.fov_lon,
+            nom.fov_lat,
+            tilt.x,
+            tilt.y,
+            hillas_dict_mod,
+            tel_x,
+            tel_y,
+            90 * u.deg - array_pointing.alt,
+        )
 
         src_error = np.sqrt(err_fov_lon**2 + err_fov_lat**2)
 
--- a/ctapipe/reco/hillas_intersection.py
+++ b/ctapipe/reco/hillas_intersection.py
@@ -506,15 +506,7 @@ class HillasIntersection(HillasGeometryReconstructor):
         mean_height = mean_distance * np.cos(zen.to_value(u.rad))
 
         # Add on the height of the detector above sea level
-        if self.subarray.reference_location is not None:
-            mean_height += self.subarray.reference_location.geodetic.height.to_value(
-                u.m
-            )
-        else:
-            # FIXME: Can remoev this check once we ensure the reference_location is always loaded
-            warnings.warn(
-                "Computing h_max with no reference location. Height will be wrong."
-            )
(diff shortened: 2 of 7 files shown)


ctapipe/reco/hillas_intersection.py
```diff
@@ -495,25 +499,27 @@ def reconstruct_xmax(
...
-
-        mean_height *= u.m
-        # Lookup this height in the depth tables, the convert Hmax to Xmax
-        # x_max = self.thickness_profile(mean_height.to(u.km))
-        # Convert to slant depth
-        # x_max /= np.cos(zen)
+            mean_height = np.nan
 
-        return mean_height
+        return u.Quantity(mean_height, u.m)
```
Tobychev, line 522: This is probably a good opportunity to add returning an `h_max_uncert` from this function, even if just giving NaN for now rather than manually setting it when the `ReconstructedGeometryContainer`  is created

ctapipe/reco/hillas_reconstructor.py
```diff
@@ -411,19 +420,22 @@ def estimate_core_position(array_pointing, psi, positions):
...
         positions = positions.cartesian.xyz.T.to_value(u.m)
-        # not sure if its better to return the length of the vector of the z component
-        return u.Quantity(
-            np.linalg.norm(line_line_intersection_3d(cog_vectors, positions)),
+        shower_max = u.Quantity(
+            line_line_intersection_3d(cog_vectors, positions),
             u.m,
         )
+
+        return shower_max[2]  # the z-coordinate only
```
Tobychev, line 441: Could return a dummy `h_max_uncert` here as well.

maxnoe:

kosack:
I think I will still add "slant" versions of the two quantities to the data model:
* `shower_distance`: the distance from the impact point of the shower projected to sea-level  to the shower-max point
* `shower_slant_depth`: the line integral over the atmosphere from infinity to the shower-max point along the shower axis

The question now is if we keep `h_max` and `x_max` or rename them `shower_height` and `shower_depth` (breaking data model change, but as clearly nobody has used these up to now, maybe it's not a worry)

maxnoe:
Something went wrong here with the merge/rebase?

kosack:
> Something went wrong here with the merge/rebase?

not sure what it was, but I rebased again and did a force-with-lease, and seems ok now

ctapipe/containers.py
```diff
@@ -920,8 +920,13 @@ class ReconstructedGeometryContainer(Container):
         "reconstructed core position uncertainty along tilted frame Y axis",
         unit=u.m,
     )
-    h_max = Field(nan * u.m, "reconstructed height of the shower maximum", unit=u.m)
+    h_max = Field(
+        nan * u.m,
+        "reconstructed vertical height (above sea level) of the shower maximum",
```
maxnoe, line 925: remove parentheses?

kosack: done

ctapipe/reco/hillas_reconstructor.py
kosack: It may need to be computed separately, not sure yet what is the right way to compute it, so so far I'll just leave it.

ctapipe/reco/hillas_intersection.py
```diff
@@ -309,7 +310,7 @@ def reconstruct_nominal(self, hillas_parameters):
 
         """
         if len(hillas_parameters) < 2:
-            return None  # Throw away events with < 2 images
+            return (np.nan, np.nan, np.nan, np.nan)  # Throw away events with < 2 images
```
maxnoe, line 313: I see changes in the return values here but no changes that accept these changes...

What is the effect of changing the simple `None` to a tuple of 4 `np.nan`?

ctapipe/reco/hillas_intersection.py
```diff
@@ -381,7 +382,8 @@ def reconstruct_tilted(self, hillas_parameters, tel_x, tel_y):
             core uncertainty X
         """
         if len(hillas_parameters) < 2:
-            return None  # Throw away events with < 2 images
+            return (np.nan, np.nan, np.nan, np.nan)  # Throw away events with < 2 images
```
maxnoe, line 385: Same question as above.

ctapipe/reco/hillas_intersection.py
```diff
@@ -495,25 +497,19 @@ def reconstruct_xmax(
             np.array(ty),
         )
         weight = np.array(amp)
-        mean_height = np.sum(height * weight) / np.sum(weight)
+        mean_distance = np.sum(height * weight) / np.sum(weight)
```
maxnoe, line 500: This is `np.average(height, weights=weight)`

ctapipe/reco/hillas_intersection.py
```diff
@@ -495,25 +497,19 @@ def reconstruct_xmax(
...
         # This value is height above telescope in the tilted system,
         # we should convert to height above ground
-        mean_height *= np.cos(zen)
+        mean_height = mean_distance * np.cos(zen.to_value(u.rad))
 
         # Add on the height of the detector above sea level
-        mean_height += 2100  # TODO: replace with instrument info
+        mean_height += self.subarray.reference_location.geodetic.height.to_value(u.m)
 
         if mean_height > 100000 or np.isnan(mean_height):
```
maxnoe, line 509: Checking for nan doesn't make sense anymore if you set to nan.

maxnoe: Replace magic number with something like "H_MAX_UPPER_LIMIT = 100000"?

ctapipe/reco/hillas_intersection.py
Tobychev: Looking at the unit test, I think the effect is to strictly increase the correctness of the code. Or more plainly: no code could actually handle the `None` in the first place, the two places where `reconstruct_nominal` is actually called both expect and indexable object :)

kosack: yes, sorry I should have commented here: returning None there was actually  a bug that i encountered. The useage expects a 4-tuple, and the code would have just crashed if None was returned, so I fixed it (it's not strictly part of this PR though)

ctapipe/reco/hillas_intersection.py
kosack: done

ctapipe/reco/hillas_intersection.py
kosack: done

ctapipe/reco/hillas_intersection.py
kosack: same as above, the function returns a 4-tuple that gets unpacked on usage, so None is not a valid return value

Tobychev (approved):

PR #2409, opened 2023-10-15
Interpolate pointing in TableLoader and HDF5EventSource

maxnoe:
This adds the possibility to interpolate pointing information from a not yet used by ctapipe itself `/dl0/monitoring/telescope/pointing/<tel_id>` group.

This is critical for analysis of e.g.  LST observations.

At the moment, I add this table to the DL1 files output by ctapipe-process with a custom script, but eventually we should also add APIs for the datawriter / ctapipe-process / event sources to read this from monitoring files and put it into the outputs.

### DIFF
--- a/docs/changes/2409.feature.rst
+++ b/docs/changes/2409.feature.rst
@@ -1,5 +1,5 @@
 Add support for interpolating a monitoring pointing table
 in ``TableLoader``. The corresponding table is not yet written by ``ctapipe``,
 but can be written by external tools.
-This is to enable analysis of real obsversations, where the pointing changes over time in
+This is to enable analysis of real observations, where the pointing changes over time in
 alt/az.
--- a/docs/changes/2409.feature.rst
+++ b/docs/changes/2409.feature.rst
@@ -1,0 +1,5 @@
+Add support for interpolating a monitoring pointing table
+in ``TableLoader``. The corresponding table is not yet written by ``ctapipe``,
+but can be written by external tools.
+This is to enable analysis of real obsversations, where the pointing changes over time in
+alt/az.
--- a/src/ctapipe/conftest.py
+++ b/src/ctapipe/conftest.py
@@ -7,6 +7,7 @@ from copy import deepcopy
 import astropy.units as u
 import numpy as np
 import pytest
+import tables
 from astropy.coordinates import EarthLocation
 from astropy.table import Table
 from pytest_astropy_header.display import PYTEST_HEADER_MODULES
@@ -677,5 +678,9 @@ def dl1_mon_pointing_file(dl1_file, dl1_tmp_path):
 
     for tel_id in subarray.tel:
         write_table(table, path, f"/dl0/monitoring/telescope/pointing/tel_{tel_id:03d}")
+
+    # remove static pointing table
+    with tables.open_file(path, "r+") as f:
+        f.remove_node("/configuration/telescope/pointing", recursive=True)
 
     return path
--- a/src/ctapipe/conftest.py
+++ b/src/ctapipe/conftest.py
@@ -1,12 +1,14 @@
 """
 common pytest fixtures for tests in ctapipe
 """
-
+import shutil
 from copy import deepcopy
 
 import astropy.units as u
+import numpy as np
 import pytest
 from astropy.coordinates import EarthLocation
+from astropy.table import Table
 from pytest_astropy_header.display import PYTEST_HEADER_MODULES
 
 from ctapipe.core import run_tool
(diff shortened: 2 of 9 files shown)


kosack:
This can also be used eventually to fill in the event-wise monitoring info in the event loop as well not just in TableLoader.  

And just to note, this eventually must also interpolate other mis-pointing parameters like camera-z-axis rotation and any other deformation that needs to be applied during the DL1 step.  LST doesn't compute those yet I guess, but it's good not to forget they will be necessary eventually!

ctapipe/io/pointing.py
```diff
@@ -0,0 +1,81 @@
...
+        # sort first, so it's not done twice for each interpolator
+        pointing_table.sort("time")
+        mjd = pointing_table["time"].tai.mjd
+
+        az = pointing_table["azimuth"].quantity.to_value(u.rad)
+        # prepare azimuth for interpolation "unwrapping", i.e. turning 359, 1 into 359, 361
+        az = np.unwrap(az)
+        alt = pointing_table["altitude"].quantity.to_value(u.rad)
+
+        self._az_interpolators[tel_id] = interp1d(mjd, az, **self.interp_options)
```
maxnoe, line 52: scipy docs:

>  [interp1d](https://docs.scipy.org/doc/scipy/reference/generated/scipy.interpolate.interp1d.html#scipy.interpolate.interp1d) is considered legacy API and is not recommended for use in new code. Consider using more specific interpolators instead.

maxnoe: But I don't see a direct "new" replacement for the case of linear interpolation.

I'd stick with it for now.

maxnoe: Yes, but that is lacking all the options for how to deal with values outside the interpolation range and it requires passing the data in everytime you call it.

docs/changes/2409.feature.rst
```diff
@@ -0,0 +1,5 @@
+Add support for interpolating a monitoring pointing table
+in ``TableLoader``. The corresponding table is not yet written by ``ctapipe``,
+but can be written by external tools.
+This is to enable analysis of real obsversations, where the pointing changes over time in
```
Tobychev, line 4: typo: obsversations -> observations

ctapipe/io/tableloader.py
```diff
@@ -529,6 +541,11 @@ def _read_telescope_events_for_id(self, tel_id, start=None, stop=None):
             )
             table = _join_telescope_events(table, impacts)
 
+        if self.interpolate_pointing and POINTING_GROUP in self.h5file.root:
```
Tobychev, line 544: This looks like it will silently skip doing the requested interpolation if `POINTING_GROUP` is not in the file being processed, but if the user has requested it they would probably like to be at least warned it can't be done.

ctapipe/io/pointing.py
```diff
@@ -0,0 +1,81 @@
...
+from scipy.interpolate import interp1d
+
+from ctapipe.core import Component, traits
+
+from .astropy_helpers import read_table
+
+
+class PointingInterpolator(Component):
+    bounds_error = traits.Bool(default_value=True).tag(config=True)
+    extrapolate = traits.Bool(default_value=False).tag(config=True)
```
Tobychev, line 15: These traits need help strings, especially as the value of one will silently override the value of the other. Ie you if you set both `bounds_error` and `extrapolate` to True then `bounds_error` will silently win the conflict.

morcuended (approved):
Looks good to me

ctapipe/io/pointing.py
```diff
@@ -0,0 +1,131 @@
...
+        # and TAI is contiguous, so no issues with leap seconds.
+        mjd = pointing_table["time"].tai.mjd
+
+        az = pointing_table["azimuth"].quantity.to_value(u.rad)
+        # prepare azimuth for interpolation by "unwrapping": i.e. turning
+        # [359, 1] into [359, 361]. This assumes that if we get values like
+        # [359, 1] the telescope moved 2 degrees through 0, not 358 degrees
+        # the other way around. This should be true for all telescopes given
+        # the sampling speed of pointing values and their maximum movement speed.
+        # No telescope can turn more than 180° in 2 seconds.
```
Tobychev, line 90: I think the more relevant remark is that during tracking of a source the telescopes only move quickly close to the poles, and the worst case for those motions are a straight overhead track that flip the Azimuth 180 degrees. But maybe I misunderstand this comment.

I expect how this works in practice depends on how the telescopes handle the very high zenith tracks, do they stop data taking as they swing around?

Tobychev (approved):

ctapipe/io/pointing.py
kosack: I think there is a telescope requirement that we do not observe very close to a poll (though I have to check exactly how far is allowed). So probably it's not a major issue, at least for CTA. For HESS, for example CT5 can in principle move to altitudes greater than 90°, but in practice due to the effects it has on pointing accuracy, it's no longer used (and the fact that the HESS software also doesn't do this transform correctly).  

It may be worth adding some of this comment's text to the docstring of the class, however, as it is important to remember and should show up in the documentation.

ctapipe/io/pointing.py
```diff
@@ -0,0 +1,131 @@
...
+        help="If true, raises an exception when trying to extrapolate out of the given table",
+    ).tag(config=True)
+
+    extrapolate = traits.Bool(
+        help="If bounds_error is False, this flag will specify whether values outside"
+        "the available values are filled with nan (False) or extrapolated (True).",
+        default_value=False,
+    ).tag(config=True)
+
+    def __init__(self, h5file=None, **kwargs):
```
kosack, line 33: please document the h5file argument (what is expected, what happens if it's None). E.g. if h5file is specified, PointingInterpolator will look for the tables in [path] and load all that are available for each tel_id.  If it is not specified, the user should call `add_table()` to add tables for relevant tel_ids.

kosack (changes requested):
only minor comments (docstring fixes). Otherwise looks ok.

ctapipe/io/pointing.py
maxnoe: > I think the more relevant remark is that during tracking of a source the telescopes only move quickly close to the poles, 

What is implemented and tested here is really the normal case of moving, not related to any movement at the pole, given the sequence of azimuth values and assuming  you interpolate exactly in between, there are 2 solutions for the sequence:
```
[357, 359, 1, 3]
```

Interpolating naïvely:

```
[358, 180, 2]
```

or with the numpy function applied above, the sequence is first turned into:
```
[357, 359, 361, 363]
```
which is then interpolated and wrapped to:

```
[358, 0, 2]
```

kosack (approved):

PR #2411, opened 2023-10-19
Add option to override obs_id in SimTelEventSource

maxnoe:
Simulations were performed that re-use the same CORSIKA run numbers, e.g. different pointing positions for the same particles in prod5 and also different particles for the same pointing.

To still enable unique obs_ids in merged files for these productions, an "override_obs_id" option is added to the ``SimTelEventSource`` that can be used to assign new, globally unique ids.

The original run number is added to the ``SimulationConfigContainer``.

We have discussed before to decouple simulation obs_ids from CORSIKA / simtel run numbers.

This is a first step in enabling this. This was triggered also because the LST MC production reuses run numbers in a way that completely prevented merging files for model training.

### DIFF
--- a/ctapipe/io/simteleventsource.py
+++ b/ctapipe/io/simteleventsource.py
@@ -49,7 +49,7 @@ from ..containers import (
 from ..coordinates import CameraFrame, shower_impact_distance
 from ..core import Map
 from ..core.provenance import Provenance
-from ..core.traits import Bool, ComponentName, Float, Undefined, UseEnum
+from ..core.traits import Bool, ComponentName, Float, Integer, Undefined, UseEnum
 from ..instrument import (
     CameraDescription,
     CameraGeometry,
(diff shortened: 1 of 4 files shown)


kosack (commented):
yes, this fits well with the external discussion you mentioned that obs_ids should be decoupled from simulation run_ids.  

is a config option enough? It means computing the unique obs_id outside the tool, in a script for example.  Another option is to have the obs_id by default be in the agreed ACADA format (run_number + 4000000 or whatever), and then override_obs_id can be used only in special cases where you don't want that behavior

maxnoe:
I need this now for really the case where I need to assign custom obs_ids. The LST simulations restart run numbers for each pointing and I need to merge ~10 pointing directions for model training.

I wouldn't just add the `4`, since `ctapipe` is used on not just `SDMC-DPPS` produced simtel files, in fact, no such files really exist as of now.

ctapipe/io/simteleventsource.py
```diff
@@ -1056,11 +1061,20 @@ def _fill_scheduling_and_observation_blocks(self):
...
-        obs_id = self.file_.header["run"]
+
+        # this event source always contains only a single OB, so we can
+        # also assign a single obs_id
+        if self.override_obs_id is not None:
+            self.obs_id = self.override_obs_id
+        else:
+            self.obs_id = self.file_.header["run"]
+
+        # simulations have at the moment do not have SBs, use OB id
```
LukasNickel, line 1072: remove first have

kosack (approved):

LukasNickel (approved):

PR #2414, opened 2023-10-20
Include example config for ctapipe-train-disp-reconstructor in quickstart tool

LukasBeiske:


ctapipe/tools/quickstart.py
```diff
@@ -55,13 +56,14 @@
 ctapipe-process --help-all
 ```
 
-## ctapipe-train-energy-regressor / ctapipe-train-particle-classifier configs
+## ctapipe-train-energy-regressor / ctapipe-train-particle-classifier / ctapipe-train-disp-reconstructor configs
 
 Included here are also base configurations for training machine learning (ML)
-models for energy regression and gamma/hadron separation.
+models for energy regression, gamma/hadron separation and disp origin reconstruction.
```
maxnoe, line 62: I think we should add proper configuration files for use with ctapipe-quickstart, not the unit test files with slimmed down simplified configurations as we have them here.

Ok for now, but maybe add a note that these files are optimized for very fast training for demos and unit tests and will not result in good performance?

kosack:
Is the actual config file missing here, or am I missing something?

LukasBeiske:
> Is the actual config file missing here, or am I missing something?

It is there at `ctapipe/resources/`, but `ctapipe-quickstart` just does not include it when creating the working directory.

Tobychev (approved):

maxnoe (approved):

PR #2418, opened 2023-10-23
Do not use hidden attribute `reconstructor._models` in `ctapipe-apply-models`

LukasBeiske:
Hidden attributes should not be used outside the class they are defined in. Catching the KeyError is better here.


docs/changes/2418.bugfix.rst
```diff
@@ -0,0 +1,2 @@
+Do not use the hidden attribute `SKLearnReconstructor._models` in
```
maxnoe, line 1: Only use public API ...

ctapipe/tools/apply_models.py
```diff
@@ -177,7 +177,13 @@ def _apply(self, reconstructor, tel_tables, start, stop):
         for tel_id, table in tel_tables.items():
             tel = self.loader.subarray.tel[tel_id]
 
-            if tel not in reconstructor._models:
+            if len(table) == 0:
```
maxnoe, line 180: Why is this needed? 

If it is needed, please add a test.

maxnoe: Ah, just moved, sorry

kosack (dismissed):

docs/changes/2418.bugfix.rst
```diff
@@ -1,2 +1 @@
-Do not use the hidden attribute `SKLearnReconstructor._models` in
-`ctapipe-apply-models`.
+Do not use a hidden attribute of `SKLearnReconstructor` in `ctapipe-apply-models`.
```
maxnoe, line 1: You need double backticks here, single back ticks are references for sphinx, which we don't want in changelogs, since they might get invalidated in future updates.

Tobychev (approved):

maxnoe (approved):

PR #2419, opened 2023-10-23
Fix docstring of EventSource

maxnoe:
The example code was not properly rendered due to a missing blank line before.

### DIFF
--- a/ctapipe/core/component.py
+++ b/ctapipe/core/component.py
@@ -201,8 +201,16 @@ class Component(Configurable, metaclass=AbstractConfigurableMeta):
     @classmethod
     def non_abstract_subclasses(cls):
         """
-        get dict{name: cls} of non abstract subclasses,
-        subclasses can possibly be defined in plugins
+        Get a dict of all non-abstract subclasses of this class.
+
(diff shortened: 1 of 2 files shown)


ctapipe/core/component.py
```diff
@@ -202,7 +202,7 @@ def from_name(cls, name, config=None, parent=None, **kwargs):
     def non_abstract_subclasses(cls):
         """
         get dict{name: cls} of non abstract subclasses,
-        subclasses can possibly be definded in plugins
+        subclasses can possibly be defined in plugins
```
Tobychev, line 205: I am not sure what this line wants to say: Do you mean to say "It is possible to define subclasses in plugins"? Or "A subclass might have been defined in a plugin"? 

Is the intention to say that this function will examine plugins for subclasses too?

maxnoe: > Is the intention to say that this function will examine plugins for subclasses too?

Yes that was the intention. The function imports plugins if the requsted base class defines a plugin entry point.

Here I was just fixing a typo... but I can of course improve the docstring itself.

Tobychev (approved):

PR #2420, opened 2023-10-24
Add docstring for ctapipe-train-disp-reconstructor

### LINKED ISSUE #2283: ctapipe-train-disp-reconstructor missing docstring
```
% ctapipe-info --tools

...
ctapipe-train-disp-reconstructor  - [no documentation. Please add a docstring]
...
```

LukasBeiske:
Fixes #2283

### DIFF
--- a/ctapipe/tools/train_disp_reconstructor.py
+++ b/ctapipe/tools/train_disp_reconstructor.py
@@ -11,6 +11,10 @@ from ctapipe.io import TableLoader
 from ctapipe.reco import CrossValidator, DispReconstructor
 from ctapipe.reco.preprocessing import check_valid_rows, horizontal_to_telescope
 
+__all__ = [
+    "TrainDispReconstructor",
+]
+
 
 class TrainDispReconstructor(Tool):
     """
--- a/ctapipe/tools/train_disp_reconstructor.py
+++ b/ctapipe/tools/train_disp_reconstructor.py
@@ -1,3 +1,6 @@
+"""
+Tool for training the DispReconstructor
+"""
 import astropy.units as u
 import numpy as np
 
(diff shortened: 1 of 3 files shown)


Tobychev (dismissed):

ccossou:
I'm trying to review this, @LukasBeiske are you available now? I have a few questions.

maxnoe:
It's also missing from the sphinx documentation, could you add it please?

maxnoe:
https://github.com/cta-observatory/ctapipe/blob/main/docs/api-reference/tools/index.rst

ccossou:
As far as I understand, this solves the issue, but I don't understand why the corresponding docstring has to be at the top of the file, instead of just using the one that already exist in the tool class **TrainDispReconstructor**.

maxnoe:
@ccossou I agree, the ctapipe-info tool should use the class docstring, not the module docstring

ccossou:
Am I supposed to approve it now, even though we still need to change the documentation (cf comment by max)? 

I assume our comments about using the class docstring are more for a future pull request about a refactoring of ctapipe-info and shouldn't block this one, but that's my opinion only.

maxnoe:
> I assume our comments about using the class docstring are more for a future pull request about a refactoring of ctapipe-info and shouldn't block this one, but that's my opinion only.

Yes, let's do this separately.

maxnoe:
You need to replace some single ticks with double ticks in the docstrings:

https://github.com/cta-observatory/ctapipe/actions/runs/6669605856/job/18127742839?pr=2420#step:6:124

LukasBeiske:
> You need to replace some single ticks with double ticks in the docstrings:
> 
> https://github.com/cta-observatory/ctapipe/actions/runs/6669605856/job/18127742839?pr=2420#step:6:124

I'm not finding any and the docs build fine for me locally.

LukasBeiske:
I noticed that the default target column of the `DispReconstructor` is still called `true_norm`. This target column never gets written out anywhere, since it gets calculated just before training the models, so its name does not really matter, but by now the target column that gets calculated is actually `true_norm * true_sign`. 
To avoid unnecessary confusion, I would suggest just naming this target column `true_disp`. If you agree, should I just add that change here, since it changes only a single line of code?

ccossou (approved):

maxnoe (approved):

PR #2421, opened 2023-10-25
Remove python warnings

### LINKED ISSUE #2407: CameraCalibrator warns in case no R1 data is present
**Describe the bug**

When running ctapipe-process on DL0 input, the `CameraCalibrator` throws a warning (python warning, not log message):

```
ctapipe/camera/calibrator.py:159: UserWarning: Encountered an event with no R1 data. DL0 is unchanged in this circumstance.
```

However, this is the expected case for almost anything (after all, ctapipe really starts at DL0), there shouldn't be  a warning for the expected case.

LukasNickel:
- Use logging of the component
- No R1 is expected, so its just debug level
- No DL0 is not expected, so keep it as warning
- Fixes #2407

### DIFF
--- a/docs/changes/2421.bugfix.rst
+++ b/docs/changes/2421.bugfix.rst
@@ -1,5 +1,5 @@
 Remove warnings about missing R1 or DL0 data when using the CameraCalibrator.
-These were previously emitted directly as pytho warnings and did not use the 
+These were previously emitted directly as python warnings and did not use the 
 component logging system, which they now do.
 As we do not actually expect R1 to be present it was also moved down to
 debug level.
--- a/docs/changes/2421.bugfix.rst
+++ b/docs/changes/2421.bugfix.rst
@@ -1,0 +1,5 @@
+Remove warnings about missing R1 or DL0 data when using the CameraCalibrator.
+These were previously emitted directly as pytho warnings and did not use the 
+component logging system, which they now do.
+As we do not actually expect R1 to be present it was also moved down to
+debug level.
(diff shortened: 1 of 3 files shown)


LukasNickel:
Ah sorry, apparently I did not have my pre-commit set up

docs/changes/2421.bugfix.rst
```diff
@@ -0,0 +1,5 @@
+Remove warnings about missing R1 or DL0 data when using the CameraCalibrator.
+These were previously emitted directly as pytho warnings and did not use the
```
morcuended, line 2: typo in python

morcuended (commented):
thanks @LukasNickel, just a typo in the changelog

Tobychev (approved):

kosack:
Maybe it's not for this PR, but the overall behavior here should be updated to be more in line with how we do the DL1 and DL2 steps. In ctapipe-process, we have methods like `should_compute_dl1()` that determine if the DL0→DL1 transition should happen or not depending on the existence of pre-existing DL1 or the user specifying it to be recomputed, and same with `should_compute_dl2()`.    It would make more sense to have the same thing for DL0:  if DL0 already exists, there should be no need for a warning, but the user should have the option to force re-computing it if and only if R1 also exists. 

That way all transitions are handled in the same way, and you only need to deal with the case if the user asks explicitly for a transformation to be re-computed, but the previous data level doesn't exist in the input.  And in that case it's not even a warning, but an error that should stop processing.

morcuended (approved):

PR #2422, opened 2023-10-25
Overburden-to-height conversion functions

gschwefer:
Implements the overburden to height conversion functions in the atmosphere module

### DIFF
--- a/src/ctapipe/atmosphere.py
+++ b/src/ctapipe/atmosphere.py
@@ -107,7 +107,7 @@ class AtmosphereDensityProfile(abc.ABC):
 
         """
 
-        if zenith_angle > 70 * u.deg:
+        if np.any(zenith_angle > 70 * u.deg):
             warnings.warn(
                 "Zenith angle too high for accurate slant depth",
                 SlantDepthZenithRangeWarning,
@@ -134,6 +134,12 @@ class AtmosphereDensityProfile(abc.ABC):
            unit to output (must be convertible to m)
 
         """
+
+        if np.any(zenith_angle > 70 * u.deg):
+            warnings.warn(
+                "Zenith angle too high for accurate slant depth",
+                SlantDepthZenithRangeWarning,
+            )
 
         return (self.height_from_overburden(slant_depth * np.cos(zenith_angle))).to(
             output_units
@@ -169,7 +175,7 @@ class AtmosphereDensityProfile(abc.ABC):
         axis[1].set_yscale("log")
         axis[1].grid(True)
 
-        zenith_angle = np.linspace(0, 80, 20) * u.deg
+        zenith_angle = np.linspace(0, 70, 20) * u.deg
         for height in [0, 5, 10, 20] * u.km:
             column_density = self.slant_depth_from_height(height, zenith_angle)
             axis[2].plot(zenith_angle, column_density, label=f"Height={height}")
--- a/src/ctapipe/atmosphere.py
+++ b/src/ctapipe/atmosphere.py
@@ -66,8 +66,6 @@ class AtmosphereDensityProfile(abc.ABC):
         r"""Get the height a.s.l. from the mass overburden in the atmosphere.
             Inverse of the integral function
 
-        ..
-
         Returns
         -------
         u.Quantity["m"]:
--- a/src/ctapipe/atmosphere.py
+++ b/src/ctapipe/atmosphere.py
@@ -79,6 +79,7 @@ class AtmosphereDensityProfile(abc.ABC):
         self,
         distance: u.Quantity,
         zenith_angle=0 * u.deg,
+        observer_altitude=0 * u.m,
         output_units=GRAMMAGE_UNIT,
     ):
         r"""Line-of-sight integral from the shower distance to infinity, along
@@ -86,7 +87,7 @@ class AtmosphereDensityProfile(abc.ABC):
         the *slant depth*. The atmosphere here is assumed to be Cartesian, the
         curvature of the Earth is not taken into account.
 
-        .. math:: X(h, \Psi) = \int_{h}^{\infty} \rho(h' \cos{\Psi}) dh'
+        .. math:: X(d, \Psi) = \int_{h_{obs} + d\cos{\Psi}}^{\infty} \rho(h') dh' / \cos{\Psi}
 
         Parameters
         ----------
@@ -94,13 +95,16 @@ class AtmosphereDensityProfile(abc.ABC):
            line-of-site distance from observer to point
         zenith_angle: u.Quantity["angle"]
            zenith angle of observation
+        observer_altitude: u.Quantity["length]
+           Height a.s.l. of the observer
         output_units: u.Unit
            unit to output (must be convertible to g/cm2)
 
         """
 
         return (
-            self.integral(distance * np.cos(zenith_angle)) / np.cos(zenith_angle)
+            self.integral(distance * np.cos(zenith_angle) + observer_altitude)
+            / np.cos(zenith_angle)
         ).to(output_units)
 
     def height_from_slant_depth(
--- a/src/ctapipe/atmosphere.py
+++ b/src/ctapipe/atmosphere.py
@@ -15,7 +15,7 @@ from functools import partial
 
 import numpy as np
 from astropy import units as u
-from astropy.table import Table
+from astropy.table import QTable
 from scipy.interpolate import interp1d
 
 __all__ = [
@@ -28,6 +28,9 @@ __all__ = [
 SUPPORTED_TABLE_VERSIONS = {
     1,
 }
+
+GRAMMAGE_UNIT = u.g / u.cm**2
+DENSITY_UNIT = u.g / u.cm**3
 
 
 class AtmosphereDensityProfile(abc.ABC):
@@ -73,7 +76,10 @@ class AtmosphereDensityProfile(abc.ABC):
         """
 
     def line_of_sight_integral(
-        self, distance: u.Quantity, zenith_angle=0 * u.deg, output_units=u.g / u.cm**2
+        self,
+        distance: u.Quantity,
+        zenith_angle=0 * u.deg,
+        output_units=GRAMMAGE_UNIT,
     ):
         r"""Line-of-sight integral from the shower distance to infinity, along
         the direction specified by the zenith angle. This is sometimes called
@@ -144,7 +150,7 @@ class AtmosphereDensityProfile(abc.ABC):
         return fig, axis
 
     @classmethod
-    def from_table(cls, table: Table):
+    def from_table(cls, table: QTable):
         """return a subclass of AtmosphereDensityProfile from a serialized
         table"""
 
@@ -191,7 +197,7 @@ class ExponentialAtmosphereDensityProfile(AtmosphereDensityProfile):
     """
 
     scale_height: u.Quantity = 8 * u.km
-    scale_density: u.Quantity = 0.00125 * u.g / (u.cm**3)
+    scale_density: u.Quantity = 0.00125 * u.g / u.cm**3
 
     @u.quantity_input(height=u.m)
     def __call__(self, height) -> u.Quantity:
@@ -206,7 +212,7 @@ class ExponentialAtmosphereDensityProfile(AtmosphereDensityProfile):
             self.scale_density * self.scale_height * np.exp(-height / self.scale_height)
         )
 
-    @u.quantity_input(overburden=u.g / (u.cm * u.cm))
+    @u.quantity_input(overburden=u.g / u.cm**2)
     def height_from_overburden(self, overburden) -> u.Quantity:
         return -self.scale_height * np.log(
             overburden / (self.scale_height * self.scale_density)
@@ -247,11 +253,11 @@ class TableAtmosphereDensityProfile(AtmosphereDensityProfile):
         load a TableAtmosphereDensityProfile from a supported EventSource
     """
 
-    def __init__(self, table: Table):
+    def __init__(self, table: QTable):
         """
         Parameters
         ----------
-        table: Table
+        table : QTable
             Table with columns `height`, `density`, and `column_density`
         """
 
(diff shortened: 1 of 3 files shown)


ctapipe/atmosphere.py
```diff
@@ -284,12 +316,18 @@ def __repr__(self):
...
 
+
 # pylint: disable=invalid-name,unused-argument
 def _exponential(h, a, b, c):
     """exponential atmosphere"""
     return a + b * np.exp(-h / c)
 
 
+def _inv_exponential(T, a, b, c):
+    "inverse function for eponetial atmosphere"
```
LukasNickel, line 327: typo: eponential

ctapipe/atmosphere.py
```diff
@@ -300,6 +338,11 @@ def _linear(h, a, b, c):
     return a - b * h / c
 
 
+def _inv_linear(T, a, b, c):
+    "inverse function for linear atmosphere"
+    return -b / c * (T - a)
```
Tobychev, line 343: Should be "- c / b", also a unit test should have caught this!!!! :)

maxnoe:
I wanted to make a simple plot of height vs. grammage using this function to check for the edge behavior for the cubic interpolation, but it flat out failed:

```python
from astropy.coordinates import AltAz
import astropy.units as u

with EventSource("dataset://gamma_prod5.simtel.zst") as source:
    camera_geometry = source.subarray.tel[1].camera.geometry

    pointing = AltAz(alt=70 * u.deg, az=0 * u.deg)
    frame_with_pointing = camera_geometry.frame.replicate_without_data(telescope_pointing=pointing)
    print(frame_with_pointing.telescope_pointing)
```

```
/home/maxnoe/Uni/CTA/ctapipe/ctapipe/atmosphere.py:307: RuntimeWarning: divide by zero encountered in log10
  self._height_interp(np.log10(overburden.to("g cm-2").value)), u.km
  File "/home/maxnoe/Uni/CTA/ctapipe/invalid_values.py", line 7, in <module>
    source.atmosphere_density_profile.height_from_overburden(grammage)
  File "/home/maxnoe/.local/conda/envs/cta-dev/lib/python3.11/site-packages/astropy/units/decorators.py", line 313, in wrapper
    return_ = wrapped_function(*func_args, **func_kwargs)
              ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/maxnoe/Uni/CTA/ctapipe/ctapipe/atmosphere.py", line 307, in height_from_overburden
    self._height_interp(np.log10(overburden.to("g cm-2").value)), u.km
    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/maxnoe/.local/conda/envs/cta-dev/lib/python3.11/site-packages/scipy/interpolate/_polyint.py", line 80, in __call__
    y = self._evaluate(x)
        ^^^^^^^^^^^^^^^^^
  File "/home/maxnoe/.local/conda/envs/cta-dev/lib/python3.11/site-packages/scipy/interpolate/_interpolate.py", line 755, in _evaluate
    below_bounds, above_bounds = self._check_bounds(x_new)
                                 ^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/maxnoe/.local/conda/envs/cta-dev/lib/python3.11/site-packages/scipy/interpolate/_interpolate.py", line 784, in _check_bounds
    raise ValueError("A value ({}) in x_new is below "
ValueError: A value (-inf) in x_new is below the interpolation range's minimum value (-4.864756005337933).
```

maxnoe:
The error is mostly misleading, because it reports the transformed by log-values, so the user cannot really tell what is wrong easily

maxnoe:
Please also rebase vs. current main to get the docs build working

ctapipe/tests/test_atmosphere.py
```diff
@@ -133,3 +133,46 @@ def test_against_reference():
...
+            [9.75 * 100000, -18.4377, 1265.08, 708915, 0],
+            [19.00 * 100000, 0.217565, 1349.22, 636143, 0],
+            [46.00 * 100000, -0.000201796, 703.745, 721128, 0],
+            [106.00 * 100000, 0.000763128, 1, 1.57247e10, 0],
+        ]
+    )
+
+    profile_5 = atmo.FiveLayerAtmosphereDensityProfile.from_array(fit_reference)
+
+    circle_height_5_layer = profile_5.height_from_overburden(
```
Tobychev, line 157: Need to add tests for each of the five atmosphere levels

Tobychev (approved):

ctapipe/atmosphere.py
```diff
@@ -259,6 +279,12 @@ def __init__(self, table: Table):
             kind="cubic",
         )
 
+        self._height_interp = interp1d(
+            np.log10(self.table["column_density"].to("g cm-2").value),
```
maxnoe, line 283: This should be `to_value(u.g / u.cm**-2)` and maybe define the unit just ones.

In general, don't use strings for units if they are easily available as module constants in astropy.

maxnoe: We could also add settings for the interpolator to output nan for invalid input. I think this makes a lot of sense here, and then the calling code can decide whether having nans in the output is an error condition or not.

maxnoe:
@gschwefer 

I played around a bit locally and wanted to open a PR against your branch but accidentally pushed here directly.

Could you have a look at [51dbae7](https://github.com/cta-observatory/ctapipe/pull/2422/commits/51dbae795fd144346d1d43f294181ea4f14cbc24) and see if this ok with you? If not I'll revert

sorry

gschwefer:
The commit looks alright. Should there be a test similar to the one you added for the other profiles as well, i.e. should we explicitly guard against inputing a negative height to the exponential model or is this a potentially relevant case?

maxnoe:
@gschwefer Yes, I think the three classes should behave the same regarding invalid inputs, otherwise they are not interchangeable

ctapipe/atmosphere.py
```diff
@@ -131,7 +151,7 @@ def peek(self):
         return fig, axis
 
     @classmethod
-    def from_table(cls, table: Table):
+    def from_table(cls, table: QTable):
```
kosack, line 154: I didn't use QTables here explicitly because they drop some metadata when writing. I haven't checked if that behavior is fixed in the latest astropy, however.  Due to that, I generally prefer to use bare Tables for internal variables, and only cast to a QTable when needed  for simplicity.

kosack (changes requested):
Looks good, but we're still missing the inverse of `AtmosphereProfile.line_of_site_integral` where the zenith angle is taken into account (i.e. `profile.height_from_slant_depth(h, zenith_angle=60*u.deg)`)

gschwefer:
ImPACT reconstructs slant depth, but what that is converted to as an output is up to whatever we decide it should. Currently, we would need a function that is `height_from_slant_depth`.

gschwefer:
Another question about the functions currently there: Is the line_of_sight_integral function really correct? Because the distance you put into it is a distance from the observer, but the integral functions uses heights a.s.l.

ctapipe/atmosphere.py
gschwefer: So should I revert everything back to Tables?

maxnoe:
It seems you ran an older version of black that formats the powers slightly differently.

Running black via pre-commit should use the same version as in the CI, e.g. copy the exact command of the CI:

```
$ pre-commit run --show-diff-on-failure --color=always --files $(git diff origin/main --name-only)
```

gschwefer:
Yep thanks for the tip, I committed from a different ctapipe installation than normal where the pre-commit hooks are not setup properly

ctapipe/tests/test_atmosphere.py
```diff
@@ -133,3 +133,116 @@ def test_against_reference():
...
+    for height in layer_5_heights:
+        circle_height_5_layer = profile_5.height_from_overburden(
+            profile_5.integral(height)
+        )
+
+        assert np.allclose(circle_height_5_layer, height, rtol=0.005)
+
+    # Exponential atmosphere
+    density_model = atmo.ExponentialAtmosphereDensityProfile(
+        scale_height=10 * u.km, scale_density=0.00125 * u.g / u.cm**3
```
Tobychev, line 182: I think you can import the `DENSITY_UNIT` constant and use it here.

ctapipe/tests/test_atmosphere.py
```diff
@@ -133,3 +133,116 @@ def test_against_reference():
...
+    )
+
+    assert np.allclose(circle_height_table, 47 * u.km, rtol=0.0001)
+
+
+def test_out_of_range_table(table_profile):
+    with pytest.warns(RuntimeWarning, match="divide by zero"):
+        assert np.isposinf(table_profile.height_from_overburden(0 * u.g / u.cm**2))
+
+    assert np.isnan(table_profile.height_from_overburden(2000 * u.g / u.cm**2))
```
Tobychev, line 202: I think you can import the `GRAMMAGE_UNIT` constant and use it here.

ctapipe/tests/test_atmosphere.py
```diff
@@ -133,3 +133,116 @@ def test_against_reference():
         0,
         atol=1e-5,
     )
+
+
+def test_line_of_sight_integral(table_profile):
+    """
+    Test observer altitude and zenith angle dependence of LoS integral function
+    """
+
```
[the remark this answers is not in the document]
gschwefer: Implemented a corresponding test

ctapipe/tests/test_atmosphere.py
gschwefer: All the unit conversions are done

Tobychev (dismissed):

ctapipe/atmosphere.py
maxnoe: Maybe better, I noticed that the ecsv output of a QTable is much more complex and we use mostly `Table` everywhere else.

So if you want to use QTable internally because that saves a couple of `.quantity`, you could do `table = QTable(table, copy=False)` as first line.

maxnoe:
Hi @gschwefer Could you fix the conflicts?

gschwefer:
I'll have a look later today

ctapipe/atmosphere.py
gschwefer: I've done exactly this now

maxnoe:
docs build failure is unrelated, I am investigating

maxnoe:
The docs failure is fixed in main, please rebase

src/ctapipe/atmosphere.py
```diff
@@ -59,30 +62,70 @@ def integral(self, height: u.Quantity) -> u.Quantity:
...
         """
 
-    def line_of_sight_integral(
-        self, distance: u.Quantity, zenith_angle=0 * u.deg, output_units=u.g / u.cm**2
+    @abc.abstractmethod
+    def height_from_overburden(self, overburden: u.Quantity) -> u.Quantity:
+        r"""Get the height a.s.l. from the mass overburden in the atmosphere.
+            Inverse of the integral function
+
+        ..
```
maxnoe, line 70: What are these two dots doing here?

gschwefer: Nothing, will remove them

maxnoe (dismissed):

Tobychev (dismissed):

maxnoe:
@kosack Have your change requests been addressed?

kosack:
> @kosack Have your change requests been addressed?

Will review today. I want to check in detail that everything works as expected

maxnoe:
@kosack I think we are at the point where we have accumulated enough significant changes to warrant a new release, it would be great to get this in, did you finish your review?

kosack (changes requested):

kosack (changes requested):
Other than minor changes (see comments), this looks good and even fixes a major bug in the original that I didn't notice.

src/ctapipe/atmosphere.py
```diff
@@ -232,31 +292,40 @@ def __init__(self, table: Table):
...
+            self.table["column_density"].to_value(GRAMMAGE_UNIT)
+        )
+        height_km = self.table["height"].to_value(u.km)
 
-        self._density_interp = interp1d(
-            self.table["height"].to("km").value,
-            np.log10(self.table["density"].to("g cm-3").value),
+        interp_kwargs = dict(
             kind="cubic",
+            bounds_error=False,
```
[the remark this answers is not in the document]
kosack: another option to solving max's problem above would be to catch the extrapolation exception and raise a nicer one with the correct (non-log) value.

src/ctapipe/atmosphere.py
```diff
@@ -59,30 +62,68 @@ def integral(self, height: u.Quantity) -> u.Quantity:
...
+        zenith_angle=0 * u.deg,
+        output_units=GRAMMAGE_UNIT,
     ):
-        r"""Line-of-sight integral from the shower distance to infinity, along
+        r"""Line-of-sight integral from height to infinity, along
         the direction specified by the zenith angle. This is sometimes called
         the *slant depth*. The atmosphere here is assumed to be Cartesian, the
-        curvature of the Earth is not taken into account.
+        curvature of the Earth is not taken into account. Inverse of
+        height_from_slant_depth function.
```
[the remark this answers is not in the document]
gschwefer: I'm fine with raising a warning here, and the docstring is easy of course. Should it be a custom warning type?

src/ctapipe/tests/test_atmosphere.py
```diff
@@ -129,7 +129,210 @@ def test_against_reference():
...
+
+    assert np.allclose(
+        los_integral_z_20,
+        table_profile.table["column_density"].to("g cm-2")[5:15]
+        / np.cos(np.deg2rad(20)),
+        rtol=1e-3,
+    )
+
+
+def test_height_overburden_circle(table_profile):
```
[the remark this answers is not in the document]
gschwefer: Done

src/ctapipe/atmosphere.py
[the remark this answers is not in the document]
gschwefer: I don't know whether it needs the exception or not. The problem with the log-value only applies for one way of the roundtrip (height(overburden)), right?

src/ctapipe/atmosphere.py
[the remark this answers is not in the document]
kosack: I think you could create e.g.:
```py3
class SlantDepthZenithRangeWarning(UserWarning):
    """
    Issued when the zenith angle of an event is beyond
    the range where the slant depth computation is correct
    """ 

# then later:
warnings.warn("zenith angle too high for accurate slant depth", SlantDepthZenithRangeWarning)
```

I think the default warning filter is to issue the warning only once (as long as the message is static), so that would avoid the case where for very-large-zenith-angle observations it would trigger for many or all events (which would clutter the logs)

I also prefer a warning here to an exception the more I think about it, since there will be random events where this could happen due to mis-reconstruction, or as mentioned earlier for high zenith observations where the bottom of the FOV is below 70 deg.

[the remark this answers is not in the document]
kosack: Really there is no much the user can do, other than to be aware of this and potentially filter events if they are relying on an accurate x_max... it should be a rare case, though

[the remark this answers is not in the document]
maxnoe: Disappointed that it's not `FlatEarthWarning`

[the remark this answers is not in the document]
gschwefer: Done

maxnoe:
@gschwefer Looks good besides the couple of minor issues the linter has found:

https://github.com/cta-observatory/ctapipe/actions/runs/8566542976/job/23476496256?pr=2422

gschwefer:
Did the pre-commit hooks change somehow? Or why is this not caught by them?

maxnoe:
The CI is running the pre-commit hooks

gschwefer:
Running the same thing, it doesn't complain for me...but I'll just fix it manually

maxnoe:
The PR runs a merge request build: it runs on the merged result of your branch into main.

It could be that this branch is not up to date with main and the config changed in main.

maxnoe:
> This branch is [23 commits ahead of](https://github.com/gschwefer/ctapipe/compare/cta-observatory%3Actapipe%3Amain...grammage-to-altitude-conversion), [45 commits behind](https://github.com/gschwefer/ctapipe/compare/grammage-to-altitude-conversion...cta-observatory%3Actapipe%3Amain) cta-observatory/ctapipe:main.

Yes, that seems to be the case. Update your branch and you should get the same errors as in the CI:

```
git fetch upstream
git rebase upstream/main
```

gschwefer:
@maxnoe Looks like this tests fail is a github issue to me

maxnoe:
I am a bit confused: https://github.com/actions/runner-images?tab=readme-ov-file#available-images

Says the `macos-latest` should be a macos 14 arm based image, but after I restarted the job, it says it's macos 12 on x86...

Maybe the rollout is not yet complete? And the failed run was running on arm / macos 14

maxnoe:
Ok, was flaky, the latest run passed

gschwefer:
I'll have a look

kosack (approved):

maxnoe (approved):

PR #2423, opened 2023-10-25
Chunked loading of training data

### LINKED ISSUE #2413: Use chunked loading in `ctapipe-train-*` tools

At the moment, the `ctapipe-train-...` tools use `TableLoader.read_telescope_events` to load all telescope events for a given telescope type in one go.

This potentially uses large amounts of memory given that we 

* Apply quality criteria that will throw away a significant percentage of the events
* Only use a subset of the available columns
* Sub-sample events if `n_events` or `n_signal` / `n_background` are configured.

Load data in smaller chunks, applying the event selection and column selection for each chunk and then merge chunks into the needed big training table to reduce overall memory usage.

LukasBeiske:
This will fix #2413.

- [X] `ctapipe-train-energy-regressor`
- [X] `ctapipe-train-particle-classifier`
- [X] `ctapipe-train-disp-reconstructor`

### DIFF
--- a/ctapipe/tools/train_disp_reconstructor.py
+++ b/ctapipe/tools/train_disp_reconstructor.py
@@ -3,15 +3,14 @@ Tool for training the DispReconstructor
 """
 import astropy.units as u
 import numpy as np
-from astropy.table import vstack
 
-from ctapipe.containers import CoordinateFrameType
 from ctapipe.core import Tool
 from ctapipe.core.traits import Bool, Int, IntTelescopeParameter, Path
-from ctapipe.exceptions import TooFewEvents
 from ctapipe.io import TableLoader
 from ctapipe.reco import CrossValidator, DispReconstructor
-from ctapipe.reco.preprocessing import check_valid_rows, horizontal_to_telescope
+from ctapipe.reco.preprocessing import horizontal_to_telescope
+
+from .utils import read_training_events
 
 __all__ = [
     "TrainDispReconstructor",
@@ -118,7 +117,28 @@ class TrainDispReconstructor(Tool):
         self.log.info("Training models for %d types", len(types))
         for tel_type in types:
             self.log.info("Loading events for %s", tel_type)
-            table = self._read_table(tel_type)
+            feature_names = self.models.features + [
+                "true_energy",
+                "subarray_pointing_lat",
+                "subarray_pointing_lon",
+                "true_alt",
+                "true_az",
+                "hillas_fov_lat",
+                "hillas_fov_lon",
+                "hillas_psi",
+            ]
+            table = read_training_events(
+                self.loader,
+                self.chunk_size,
+                tel_type,
+                self.models,
+                feature_names,
+                self.log,
+                self.rng,
+                self.n_events.tel[tel_type],
+            )
+            table[self.models.target] = self._get_true_disp(table)
+            table = table[self.models.features + [self.models.target, "true_energy"]]
 
             self.log.info("Train models on %s events", len(table))
             self.cross_validate(tel_type, table)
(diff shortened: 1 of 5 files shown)


ctapipe/tools/train_energy_regressor.py
```diff
@@ -113,31 +122,56 @@ def start(self):
...
+                )
+                n_valid_events_in_file += len(table_chunk)
+
+                table_chunk = self.regressor.feature_generator(
+                    table_chunk, subarray=self.loader.subarray
+                )
+                feature_names = self.regressor.features + [self.regressor.target]
+                table_chunk = table_chunk[feature_names]
+
+                valid = check_valid_rows(table_chunk)
```
maxnoe, line 159: Keep track of how many and only report once at the end?

ctapipe/tools/train_energy_regressor.py
```diff
@@ -113,31 +122,56 @@ def start(self):
...
+                if np.any(~valid):
```
maxnoe, line 160: It should be cheaper to evaluate `if not np.all(valid)` since that doesn't require inverting the whole array first.

ctapipe/tools/train_energy_regressor.py
LukasBeiske: I think doing the validity check itself after merging the chunks, is even simpler.

LukasBeiske: The quality criteria and the column selection would still happen within the loop, but in case there are other non-predictable events left, it would be better to also drop those within the loop, I guess. I will change it.

Tobychev:
Is there any test that checks that the `_read_tabel` functions are correct?

maxnoe:
This looks extremely similar between the three tools. Can we refactor the code into a single common function?

LukasBeiske:
> This looks extremely similar between the three tools. Can we refactor the code into a single common function?

For the disp tool, the calculation of `true_disp` needs columns which get dropped right after this calculation is done. But we could try to keep these needed columns if all of them are loaded, which only happens in the disp tool, and calculate `true_disp` outside of `_read_table`.

ctapipe/tools/train_disp_reconstructor.py
```diff
@@ -121,40 +128,62 @@ def start(self):
...
 
-        if not np.all(
-            table["subarray_pointing_frame"] == CoordinateFrameType.ALTAZ.value
-        ):
-            raise ValueError(
-                "Pointing information for training data has to be provided in horizontal coordinates"
-            )
+            valid = check_valid_rows(table_chunk)
+            if not np.all(valid):
+                n_non_predictable += np.sum(valid)
```
Tobychev, line 171: What does `n_non_predicatable` mean? From this line it looks like it stores the number of events that could be used for the ml reconstruction, so the "non" is a bit strange. 

Did you mean to say `np.sum(~valid)`?

LukasBeiske: Oh, yes you are right.

Tobychev (dismissed):

ctapipe/tools/utils.py
```diff
@@ -4,14 +4,18 @@
...
 from astropy.table import vstack
 
 from ctapipe.containers import CoordinateFrameType
+from ctapipe.core.traits import Int
 from ctapipe.exceptions import TooFewEvents
+from ctapipe.instrument.telescope import TelescopeDescription
+from ctapipe.io import TableLoader
 from ctapipe.reco.preprocessing import check_valid_rows
-from ctapipe.reco.sklearn import DispReconstructor
+from ctapipe.reco.sklearn import DispReconstructor, SKLearnReconstructor
```
maxnoe, line 18: Please use relative imports here.

ctapipe/tools/train_particle_classifier.py
```diff
@@ -181,8 +181,8 @@ def _read_input_data(self, tel_type):
             tel_type,
             self.classifier,
             feature_names,
-            self.log,
             self.rng,
+            self.log,
```
maxnoe, line 185: use keyword arguments to be safe against such errors

Tobychev (dismissed):

maxnoe (approved):

maxnoe (dismissed):

maxnoe (approved):

Tobychev (approved):

PR #2424, opened 2023-10-25
Support telescope position from `EarthLocation`s for `SubarrayDescription`

### LINKED ISSUE #2162: SubarrayDescription should support taking telescope positions as `EarthLocation`

To support actual telescopes, it would be better to support getting `EarthLocation` instances for the `SubarrayDescription` telescope positions.

StFroese:
fixes #2162


StFroese:
Ignore the test for now, I'll write a new one as a regression test...

ctapipe/instrument/subarray.py
```diff
@@ -81,8 +81,8 @@ def __init__(
         ----------
         name : str
             name of this subarray
-        tel_positions : Dict[int, np.ndarray]
-            dict of x,y,z telescope positions on the ground by tel_id. These are
+        tel_positions : Dict[int, Union[np.ndarray, EarthLocation]]
+            dict of x,y,z telescope positions on the ground by tel_id or EarthLocation. These are
```
maxnoe, line 85: `by tel_id or EarthLocation` sounds like `EarthLocation` could be the key of the dictionary, not another format for the value.

ctapipe/instrument/subarray.py
```diff
@@ -91,7 +91,9 @@ def __init__(
             coordinate system used for `tel_positions`.
         """
         self.name = name
-        self.positions = tel_positions or dict()
+        self.positions: Dict[int, Union[np.ndarray, EarthLocation]] = (
```
maxnoe, line 94: We shouldn't use mixed types for internal storage. We should accept `EarthLocation` or `GroundFrame` coordinates for the inputs, but then convert in `__init__` to one common format.

I think for now, that should be the previous, so ground frame coordinas.

ctapipe/instrument/subarray.py
```diff
@@ -157,9 +159,17 @@ def info(self, printer=print):
     def tel_coords(self):
         """Telescope positions in `~ctapipe.coordinates.GroundFrame`"""
 
-        pos_x = [p[0].to_value(u.m) for p in self.positions.values()]
-        pos_y = [p[1].to_value(u.m) for p in self.positions.values()]
-        pos_z = [p[2].to_value(u.m) for p in self.positions.values()]
+        positions = self.positions.values()
```
maxnoe, line 162: See above, internal representation should not be mixed.

ctapipe/instrument/subarray.py
Tobychev: True, better with something like 

dict of telescope positions by tel_id, given as x,y,z position or `EarthLocation`.

Issue #2415, opened 2023-10-21
Test for muon intensitity fitter fails on some platforms / computers


maxnoe:
**Describe the bug**

The test of the muon intensity fitter seems to fail on some platforms / versions with the fit not converging successfully.

We should make the test more robust so this false positive doesn't happen.

PR #2425, opened 2023-10-25
Stabilize muon intensity fit

RuneDominik:
See #2415

The cause of this issue might be that the likelihood values grow quite large and thus minuit does not converge properly as the values leave the [0, 1]-range by several orders of magnitude.

@maxnoe suggested to divide the likelihood used by the dof to lower the value. The likelihood is constructed to asymptotically resemble a chi2 so this changes the approach to chi2 per dof which does not change the minimum.


maxnoe:
> This does, ofc., not make the test more robust in any way.

If this fixes the numerical issue that was the root cause, why not?

ctapipe/image/pixel_likelihood.py
```diff
@@ -93,11 +93,11 @@ def neg_log_likelihood_approx(image, prediction, spe_width, pedestal):
...
 
-    neg_log_l = 0.5 * (
-        np.log(2 * np.pi * theta + EPSILON) + (image - prediction) ** 2 / theta
-    )
+    neg_log_l = np.log(theta + EPSILON) + (image - prediction) ** 2 / theta
 
-    return np.sum(neg_log_l)
+    # neg_log_l provides the variable term, add constants here to only
+    # compute them once
+    return 0.5 * (len(neg_log_l) * np.log(2 * np.pi) + np.sum(neg_log_l))
```
maxnoe, line 100: `log(2π)` could be a global constant.

maxnoe: Also, do we gain something by putting an `njit` on this?

maxnoe:
Docs builds failure is fixed in main, please rebase

PR #2430, opened 2023-10-26
Allow setting n_jobs on tool invocation

### LINKED ISSUE #2307: Allow overriding number of cores used by reconstructors on the CLI

Useful for GRID processing / other job systems with resource management.

Add a CLI option that allows overriding the number of cores used by reconstructors e.g. SKLearn based reconstructors to the train and apply tools.

This is needed to be able to set different n_cores for training and application.

**Additional context**

LukasNickel:
- It can be useful to set n_jobs indepently from the config file
- For the sklearn reconstructors this requires setting n_jobs on every model, that is attached to the reconstructor
- Fixes #2307

### DIFF
--- a/ctapipe/reco/reconstructor.py
+++ b/ctapipe/reco/reconstructor.py
@@ -70,7 +70,7 @@ class Reconstructor(TelescopeComponent):
     algorithms should inherit from
     """
 
-    #: ctapipe_rco entry points may provide Reconstructor implementations
+    #: ctapipe_reco entry points may provide Reconstructor implementations
     plugin_entry_point = "ctapipe_reco"
 
     def __init__(self, subarray, **kwargs):
@@ -140,6 +140,18 @@ class Reconstructor(TelescopeComponent):
 
         Provenance().add_input_file(path, role="reconstructor")
         return instance
+
+    def set_n_jobs(self, n_jobs):
+        """
+        Set n_jobs if applicable for the reconsructor.
+        This is not just a traits option of the class, because
+        for example in the case of the SKLearnReconstructors you need
+        to set the property of all of the sklearn models for it to be
+        applied in the fit and predict steps.
+        """
+        self.log.warning(
+            f"Trying to set n_jobs to {n_jobs}, but the reconstructor does not make any use of it."
+        )
 
 
 class HillasGeometryReconstructor(Reconstructor):
--- a/ctapipe/reco/reconstructor.py
+++ b/ctapipe/reco/reconstructor.py
@@ -8,7 +8,7 @@ from astropy.coordinates import AltAz, SkyCoord
 
 from ctapipe.containers import ArrayEventContainer, TelescopeImpactParameterContainer
 from ctapipe.core import Provenance, QualityQuery, TelescopeComponent
-from ctapipe.core.traits import List
+from ctapipe.core.traits import Integer, List
 
 from ..compat import StrEnum
 from ..coordinates import shower_impact_distance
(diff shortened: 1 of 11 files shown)


Tobychev (approved):

Tobychev (dismissed):

ctapipe/reco/sklearn.py
```diff
@@ -222,6 +222,13 @@ def fit(self, key, table):
         y = self._table_to_y(table, mask=valid)
         self._models[key].fit(X, y)
 
+    def set_n_jobs(self, n_jobs):
+        """
+        Update n_jobs of all associated models.
+        """
+        for model in self._models.values():
+            setattr(model, "n_jobs", n_jobs)
```
maxnoe, line 230: nonsensical use of setattr... why not use `mode.n_jobs = n_jobs`?

LukasNickel: True, this is not really necessary.
I started out with a general `set_model_attrs` function, but really we don't want to change anything else (at least I can't think of anything right now)

ctapipe/reco/reconstructor.py
```diff
@@ -141,6 +141,18 @@ def read(cls, path, parent=None, subarray=None, **kwargs):
...
 
+    def set_n_jobs(self, n_jobs):
+        """
+        Set n_jobs if applicable for the reconsructor.
+        This is not just a traits option of the class, because
+        for example in the case of the SKLearnReconstructors you need
+        to set the property of all of the sklearn models for it to be
+        applied in the fit and predict steps.
+        """
+        self.log.warning(
```
maxnoe, line 152: No warning please for thing that are expected.

LukasNickel: Are they expected?
This should only happen if a user specifically wants to overwrite a value and they expect that to work and not be ignored.

maxnoe: I'd expect that we add this option to all tools or via global configuration. I.e. the number of cores ctapipe as a whole is allowed to use in parallel.

Anything that can only use one core should just ignore it.

LukasNickel: Hm, if we extend this to every step of ctapipe via a global setting I agree.
I was more concerned about the specific usage of `ctapipe-apply-models` and was assuming
the lower datalevels would not be parallelized on the level of a single file anyway.

ctapipe/reco/reconstructor.py
```diff
@@ -141,6 +141,18 @@ def read(cls, path, parent=None, subarray=None, **kwargs):
         Provenance().add_input_file(path, role="reconstructor")
         return instance
 
+    def set_n_jobs(self, n_jobs):
+        """
+        Set n_jobs if applicable for the reconsructor.
+        This is not just a traits option of the class, because
```
maxnoe, line 147: This is an implementation detail, not documentation. It makes sense as a public API, so it is.

Also, you can do this with traitlets by using the on change mechanism of traitlets.

LukasNickel: Ah, nice, I wasn't aware traitlets can do that.

ctapipe/tools/train_energy_regressor.py
```diff
@@ -57,10 +57,17 @@ class TrainEnergyRegressor(Tool):
         default_value=0, help="Random seed for sampling and cross validation"
     ).tag(config=True)
 
+    n_jobs = Int(
+        default_value=None,
+        allow_none=True,
+        help="Number of threads to use for the reconstruction. This overwrites the values in the config",
```
maxnoe, line 63: `In the config of each reconstructor`?

This value can also be set via the config

maxnoe:
Maybe it also makes sense to set `n_jobs=None` for the models *after* training *before* pickling? So that the pickled models have `n_jobs=None`?

maxnoe (changes requested):
Please update the config files so that they now pass `n_jobs` to the Reconstructor classes and not in the model configs.

Tobychev (dismissed):

maxnoe (approved):

kosack (approved):

PR #2431, opened 2023-10-26
Check pointing coordinate frame in DispReconstructor

### LINKED ISSUE #2427: DispRegressor doesn't check in which coordinate system the array pointing is given
**Describe the bug**

DispRegressor reads the subarray pointing from the observation block data, assumes it's alt az, but for observed data like LST-1, this will not be true, as the tracking is constant in ICRS (ra/dec), not horizontal coordinates.

LukasBeiske:
Fixes #2427.

### DIFF
--- a/ctapipe/reco/sklearn.py
+++ b/ctapipe/reco/sklearn.py
@@ -778,7 +778,7 @@ class DispReconstructor(Reconstructor):
         fov_lon = table["hillas_fov_lon"].quantity + disp * np.cos(psi)
         fov_lat = table["hillas_fov_lat"].quantity + disp * np.sin(psi)
 
-        self.log.warning("Assuming constant and parallel pointing for each run")
+        self.log.warning("FIXME: Assuming constant and parallel pointing for each run")
         if np.all(table["subarray_pointing_frame"] is CoordinateFrameType.ALTAZ):
             pointing_alt = table["subarray_pointing_lat"]
             pointing_az = table["subarray_pointing_lon"]
--- a/ctapipe/reco/sklearn.py
+++ b/ctapipe/reco/sklearn.py
@@ -9,7 +9,7 @@ from typing import Dict
 import astropy.units as u
 import joblib
 import numpy as np
-from astropy.coordinates import AltAz, SkyCoord
+from astropy.coordinates import AltAz
 from astropy.table import QTable, Table, vstack
 from astropy.utils.decorators import lazyproperty
 from sklearn.metrics import accuracy_score, r2_score, roc_auc_score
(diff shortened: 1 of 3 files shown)


Tobychev (approved):

ctapipe/reco/sklearn.py
```diff
@@ -777,13 +778,34 @@ def predict_table(self, key, table: Table) -> Dict[ReconstructionProperty, Table
         fov_lon = table["hillas_fov_lon"].quantity + disp * np.cos(psi)
         fov_lat = table["hillas_fov_lat"].quantity + disp * np.sin(psi)
 
-        # FIXME: Assume constant and parallel pointing for each run
-        self.log.warning("Assuming constant and parallel pointing for each run")
+        self.log.warning("FIXME: Assuming constant and parallel pointing for each run")
+        if np.all(table["subarray_pointing_frame"] is CoordinateFrameType.ALTAZ):
```
maxnoe, line 782: this won't work. `is` cannot be applied element wise and these checks will always be False.

You need to check `np.all(table["subarray_pointing_frame"] == CoordinateFrameType.ALTAZ.value)`

ctapipe/reco/sklearn.py
```diff
@@ -777,13 +778,34 @@ def predict_table(self, key, table: Table) -> Dict[ReconstructionProperty, Table
...
+            pointing_alt = table["subarray_pointing_lat"]
+            pointing_az = table["subarray_pointing_lon"]
+        elif np.all(table["subarray_pointing_frame"] is CoordinateFrameType.ICRS):
+            pointing_altaz = SkyCoord(
+                ra=table["subarray_pointing_lon"],
+                dec=table["subarray_pointing_lat"],
+                frame="icrs",
+            ).transform_to(AltAz())
```
maxnoe, line 790: This also won't work. Transformation from ICRS to the AltAz system needs location and observation time

ctapipe/reco/sklearn.py
```diff
@@ -777,13 +778,34 @@ def predict_table(self, key, table: Table) -> Dict[ReconstructionProperty, Table
...
+            pointing_alt = pointing_altaz.alt
+            pointing_az = pointing_altaz.az
+        elif np.all(table["subarray_pointing_frame"] is CoordinateFrameType.GALACTIC):
+            pointing_altaz = SkyCoord(
+                l=table["subarray_pointing_lon"],
+                b=table["subarray_pointing_lat"],
+                frame="galactic",
+            ).transform_to(AltAz())
```
maxnoe, line 798: Same as above, needs obstime and location

maxnoe (changes requested):

maxnoe:
I think this is partially already addressed (for the application  case) in #2409 and you should just throw an error if the training data is not fixed altaz

Tobychev (approved):

PR #2433, opened 2023-10-27
Flatfielding factor instead of absolute & relative factor for amplitude correction

### LINKED ISSUE #1397: DL1 calibration's relative_factor / absolute_factor are confusingly named, documented and applied
The documentation of these factors makes no mention that in fact, that the `absolute_factor` is used not as factor but as divisor.

they are applied as :

``
charge *= relative / absolute
```

But the documentation of these values never mentions that and I think the naming is very confusing.

Also, why do we have these two factors in the same step? Why are these different calibration constants?

### PR DESCRIPTION
**Pull request by StFroese**
closes #1397

### DIFF
--- a/ctapipe/calib/camera/tests/test_calibrator.py
+++ b/ctapipe/calib/camera/tests/test_calibrator.py
@@ -173,6 +173,9 @@ def test_dl1_charge_calib(example_subarray):
     relative = random.normal(1, 0.01, n_pixels)
     y /= relative[:, np.newaxis]
 
+    # Define flatfielding coefficients
+    flatfielding = relative / absolute
+
     # Define pedestal
     pedestal = random.uniform(-4, 4, n_pixels)
     y += pedestal[:, np.newaxis]
@@ -191,8 +194,7 @@ def test_dl1_charge_calib(example_subarray):
     np.testing.assert_allclose(event.dl1.tel[tel_id].image, y.sum(1), rtol=1e-4)
 
     event.calibration.tel[tel_id].dl1.pedestal_offset = pedestal
-    event.calibration.tel[tel_id].dl1.absolute_factor = absolute
-    event.calibration.tel[tel_id].dl1.relative_factor = relative
+    event.calibration.tel[tel_id].dl1.flatfielding_factor = flatfielding
 
     # Test without timing corrections
     calibrator(event)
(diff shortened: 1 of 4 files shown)

### CODE REVIEW

File: ctapipe/calib/camera/tests/test_calibrator.py [hunk 1]
```diff
@@ -173,6 +173,9 @@ def test_dl1_charge_calib(example_subarray):
     relative = random.normal(1, 0.01, n_pixels)
     y /= relative[:, np.newaxis]
 
+    # Define flatiflieding coefficients
```
maxnoe, line 176: typo

ctapipe/containers.py
```diff
@@ -523,16 +523,12 @@ class DL1CameraCalibrationContainer(Container):
...
         1,
-        "Multiplicative coefficients for the absolute calibration of extracted charge into "
-        "physical units (e.g. photoelectrons or photons) for each pixel",
-    )
-    relative_factor = Field(
-        1,
-        "Multiplicative Coefficients for the relative correction between pixels to achieve a "
-        "uniform charge response (post absolute calibration) from a "
-        "uniform illumination.",
+        "Multiplicative coefficients for the absolute calibration of extracted charge "
```
maxnoe, line 528: That's not really true. We assume that r1 is already pre-calibrated to pe, so this is only an additional relative correction, not the conversion from adc counts to pe.

Tobychev (approved):

Hckjs:
Was it on purpose to set this PR to draft again? @maxnoe

kosack (commented):
I added some comments to the linked issue, but I disagree that these should be separated.  Flat-fielding and gain correction are not the same thing: even if you can combine them, you are losing important information that is needed for data quality monitoring as well as for better reconstruction techniques that do not use the flatfielding information but do use the absolute gains.

maxnoe:
> : even if you can combine them, you are losing important information that is needed for data quality monitoring as well as for better reconstruction techniques that do not use the flatfielding information but do use the absolute gains.

The change here and discussed in #1397 is not about the monitored quantities, it's about what is filled into the `event` so that it can be applied in the `CameraCalibrator`.

I argued that the event should only contain a single correction coefficient, that is computed from all the different monitoring inputs we have for that specific event.

PR #2436, opened 2023-10-30
Add option single_ob to HDF5Merger for merging chunks of the same ob

maxnoe:
Needed for merging of observations that are stored in multiple chunks into one final file for the OB.

TODO:

- [x] Add test

### DIFF
--- a/docs/changes/2436.feature.rst
+++ b/docs/changes/2436.feature.rst
@@ -1,0 +1,2 @@
+Add option ``single_ob`` to the ``HDF5Merger`` to support merging
+chunks of the same observation block into the same.
--- a/src/ctapipe/io/hdf5merger.py
+++ b/src/ctapipe/io/hdf5merger.py
@@ -158,6 +158,15 @@ class HDF5Merger(Component):
         True, help="Whether to include processing statistics in merged output"
     ).tag(config=True)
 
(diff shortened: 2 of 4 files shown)


kosack:
Is an option really needed for this? I would think the default behavior should just be updated to not duplicate an OB if it already exists.

maxnoe:
> Is an option really needed for this? I would think the default behavior should just be updated to not duplicate an OB if it already exists.

That could be done, yes. Depends a bit on the use case, e.g. we might want to check for uniqueness of obs_id and error out if `not single_ob`

kosack (approved):

maxnoe:
@kosack sorry, when reviewing my past self, I realized there was a bug when using the `append` mode. Added a test and fixed.

maxnoe:
That test exists, but it's not visible in the diff, it's above in the file:

https://github.com/cta-observatory/ctapipe/blob/e8eecfaede85e81728446f83856f7673c6ffd083/src/ctapipe/tools/tests/test_merge.py#L193-L200

maxnoe:
@mdebony ah, sorry, you are right, this tests another case.

kosack (approved):

mdebony:
Sorry I missed it.

Issue #2228, opened 2023-01-20
Split "geometry" into "direction" and "impact"


maxnoe:
**Please describe the use case that requires this feature.**

Some algorithms might not be able to predict both impact and direction (e.g. disp).
Some algorithms might only be able to predict impact (e.g. for monoscopic observations, two independent machine learning models).

Should we split the data model for "geometry" into a "direction" and an "impact" component?

PR #2443, opened 2023-10-31
Improve column names for disp

LukasBeiske:
- This removes the `disp_parameter_is_valid` column, since it is identical to the `disp_tel_is_valid` column.
- Doing this makes it possible to use the same prefix for both the `DispContainer` and the `ReconstructedGeometryContainer`, as filled by `DispReconstructor`. 
- As the `DispContainer` contains the output values of both the regressor (norm) and the classifier (sign) as a single value (sign * norm), I renamed this value to "parameter" to avoid confusion.

**HOWEVER:** `DispReconstructor` still creates all the geometry-related columns contained in `ReconstructedGeometryContainer` even though it only computes altitude and azimuth. But since the creation of such empty columns is not limited to the disp reconstructor, I think it would be better to address this in a different PR.

### DIFF
--- a/ctapipe/io/hdf5eventsource.py
+++ b/ctapipe/io/hdf5eventsource.py
@@ -70,6 +70,7 @@ DL2_CONTAINERS = {
 COMPATIBLE_DATA_MODEL_VERSIONS = [
     "v4.0.0",
     "v5.0.0",
+    "v6.0.0",
 ]
 
 
--- a/ctapipe/io/hdf5eventsource.py
+++ b/ctapipe/io/hdf5eventsource.py
@@ -16,6 +16,7 @@ from ..containers import (
     CameraHillasParametersContainer,
     CameraTimingParametersContainer,
     ConcentrationContainer,
+    DispContainer,
     DL1CameraContainer,
     EventIndexContainer,
     HillasParametersContainer,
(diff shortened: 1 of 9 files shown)


maxnoe:
Related: https://github.com/cta-observatory/ctapipe/issues/2228

We could have a `ReconstructedDirectionContainer`, a `ReconstructedImpactContainer` and a `ReconstructedGeometryContainer` that combines the first two for algorithms that do all on one go.

maxnoe (changes requested):
The changes look good to me, but this requires a bump in the data model version,
an entry in the data model changelog and probably dropping support for reading older files.

maxnoe:
And some tests and more code has to be adapted, see the test failures.

ctapipe/io/hdf5eventsource.py
```diff
@@ -70,6 +70,7 @@
 COMPATIBLE_DATA_MODEL_VERSIONS = [
     "v4.0.0",
     "v5.0.0",
+    "v6.0.0",
```
maxnoe, line 73: Is this true? Can we read the old files still?

I guess we should raise a blank error when detecting an older version, but only raise a warning when someone tries to read the old disp format?

We shouldn't remove the option to read DL1 files for this relatively minor change in DL2 that probably noone is using together with the HDF5EventSource.

Tobychev (dismissed):

ctapipe/io/hdf5eventsource.py
LukasBeiske: Until now, the DispContainer was not loaded by HDF5EventSource. I will add it there now and there should be now issues regarding compatibility.

maxnoe (approved):

Tobychev (approved):

Issue #2366, opened 2023-06-22
Change default values of `progress` bar option to `None`


maxnoe:
**Please describe the use case that requires this feature.**

tqdm can be smart with `disable=None`, in which case it will show the progress bar on a tty but not if redirected to e.g. a file.

This would be a good default I think for the tools.

**Describe the solution you'd like**

Change the defaults of all tools that have a progress bar option to `None` instead of `False`.

PR #2444, opened 2023-11-02
Set disable=None by default in ctapipe tools

morcuended:
Sets `tqdm(..., disable=None)` in process and merge ctapipe tools.

By default the progress bar is displayed when running the tools in interactive mode. If they are executed in non-tty mode the progress bar is disabled. There is also a command line option `--disable-progress` to disable it.

I added some tests to check the expected behavior in both tty and non-tty modes.

Solves #2366

### DIFF
--- a/ctapipe/tools/merge.py
+++ b/ctapipe/tools/merge.py
@@ -28,11 +28,11 @@ class MergeTool(Tool):
     examples = """
     To merge several files in the current directory:
 
-    > ctapipe-merge file1.h5 file2.h5 file3.h5 --output=/path/output_file.h5 --progress
+    > ctapipe-merge file1.h5 file2.h5 file3.h5 --output=/path/output_file.h5
 
     For merging files from a specific directory with a specific pattern, use:
 
-    > ctapipe-merge --input-dir=/input/dir/ --output=/path/output_file.h5 --progress
+    > ctapipe-merge --input-dir=/input/dir/ --output=/path/output_file.h5
     --pattern='*.dl1.h5'
 
     If no pattern is given, all .h5 files in the given directory will be taken as input.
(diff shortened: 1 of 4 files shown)


kosack (changes requested):
You need to remove the `--progress` option in a few other tests in `test_process.py`, I think. 

Another potential suggestion (optional): since many people may still have this option in their scripts, maybe deprecating it more gracefully would be nice?  e.g. add a `--progress` option that does nothing except emit a warning that the `--progress` option is deprecated and no longer needed (and to use --disable-progress to remove the automatic progress bar if necessary).    We might want to always do that on changes to top-level command-line options.

morcuended:
> You need to remove the `--progress` option in a few other tests in `test_process.py`, I think.

Yes, done. Thanks.

> Another potential suggestion (optional): since many people may still have this option in their scripts, maybe deprecating it more gracefully would be nice? e.g. add a `--progress` option that does nothing except emit a warning that the `--progress` option is deprecated and no longer needed (and to use --disable-progress to remove the automatic progress bar if necessary). We might want to always do that on changes to top-level command-line options.

Thanks, Good idea. I will do so.

ctapipe/tools/apply_models.py
```diff
@@ -118,6 +113,10 @@ class ApplyModels(Tool):
             },
             "Overwrite output file if it exists",
         ),
+        "disable-progress": (
```
maxnoe, line 116: I think the flag definition should be left as before,  but the action needs to changed accordingly.

I think the desired behavior is:

* Flag not given sets `disable=None`, so tqdm checks if it runs in an interactive environment or not
* `--progress` forces the progress bar -> `disable=False`
* `--no-progress` forces the progress bar to be disabled -> `disable=True`.

This would be the case for:


```
"progress": ({"ApplyModels": {"disable_progress_bar": False}}, "Always enable progress bar."),
"no-progress": ({"ApplyModels": {"disable_progress_bar": True}}, "Always disable progress bar."),
```

morcuended: Thanks for the suggestion, indeed that would not change the current implementation. It is actually what we briefly discussed in the meeting, I think.

maxnoe: Looking at all this duplicated code now, we could move these options to the base `Tool` class with a method `progress_bar` so we have this only in one place and the implementations can just do:

```
for foo in self.progress_bar(iterable, **options):
    ...
```

PR #2447, opened 2023-11-02
Bump setuptools and setuptools scm versions

### LINKED ISSUE #2445: Bump setuptools version
I got this error in CI

```
ERROR: setuptools==59.6.0 is used in combination with setuptools_scm>=8.x

Your build configuration is incomplete and previously worked by accident!
setuptools_scm requires setuptools>=61
```

current settings in pyproject.toml are:

```toml
requires = ["setuptools >= 40.6.0", "setuptools_scm[toml]>=3.4"]
```
Can we upgrade to more recent versions?
```toml
requires = ["setuptools>=64", "setuptools_scm>=8"]
```

morcuended:
Versions in the current build settings seem not compatible. Resolves #2445

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -108,8 +108,10 @@ jobs:
 
       - name: Install dependencies
         if: contains(github.event.pull_request.labels.*.name, 'documentation-only') == false
+        env:
+          PYTHON_VERSION: ${{ matrix.python-version }}
         run: |
-          python --version
+          python --version | grep "Python ${PYTHON_VERSION}"
           pip install pytest-cov restructuredtext-lint pytest-xdist 'coverage!=6.3.0'
           pip install .[all]
           pip install ./test_plugin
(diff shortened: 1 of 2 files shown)


maxnoe:
@morcuende, that doesn't seem to fix the build issue. The same error now pops up for eventio.

I will fix it there as well.

maxnoe:
Looking carefully at the log, there seems to be something more amiss here.

THe mamba build shouldn't install eventio via pip. 

It also reports a python version of 3.10 in the 3.11 build.

I think the step "fix PATH for macos" that is also executed on ubuntu runners is the issue

maxnoe:
The pr introducing the "documentation-only" fixes broke this with a wrong logic condition.

maxnoe (approved):

morcuended:
thanks for the fixes @maxnoe

kosack (approved):

PR #2448, opened 2023-11-06
Removed stray : from ctapipe.io page

Tobychev:
Removed stray :'s from the ctapipe.io page, something leftover from the long history of reformatting I guess.

### DIFF
--- a/docs/api-reference/io/index.rst
+++ b/docs/api-reference/io/index.rst
@@ -14,7 +14,7 @@ in-memory storage of event data
 
 
 Reading Event Data
-===================
+==================
 
 This module provides a set of *event sources* that are python
 generators that loop through an input file or stream and fill in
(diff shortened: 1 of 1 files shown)


docs/api-reference/io/index.rst
```diff
@@ -114,7 +114,7 @@ for new data:
    column of an output table.
 
 
-Serialization of Containers:
+Serialization of Containers
 ============================
```
maxnoe, line 118: Need to also fix the underlining

kosack (dismissed):

docs/api-reference/io/index.rst
Tobychev: Done, that affected one further section that was also too long.

maxnoe (approved):

LukasNickel (approved):

PR #2456, opened 2023-11-16
Update and add missing docstrings related to ML functionalities

LukasBeiske:


Tobychev:
Looks like your doc-tests got hit by the scipy devs letting their ssl cert expire :(

Tobychev (approved):

maxnoe:
> Looks like your doc-tests got hit by the scipy devs letting their ssl cert expire :(

Fixed now, restarted the CI

ctapipe/reco/sklearn.py
```diff
@@ -437,6 +436,9 @@ class ParticleClassifier(SKLearnClassificationReconstructor):
     property = ReconstructionProperty.PARTICLE_TYPE
 
     def __call__(self, event: ArrayEventContainer) -> None:
```
maxnoe, line 438: subclasses should inherit docstrings from their parents automatically in sphinx I think. Was this needed?

LukasBeiske: Oh yes, you are right. I was confused since it got overwritten everywhere else, but I think it would be better to just use the docstring from the parent class for `__call__` and `predict_table` in all subclasses here. 👍

ctapipe/reco/sklearn.py
```diff
@@ -292,7 +295,7 @@ class SKLearnClassificationReconstructor(SKLearnReconstructor):
     ).tag(config=True)
 
     invalid_class = traits.Integer(
-        default_value=-1, help="The label to fill in case no prediction could be made"
+        default_value=-1, help="The label to fill in case no prediction could be made."
```
Tobychev, line 298: To me "label" sounds very string-like, might be better to say "value" when talking about an integer.

LukasBeiske: How about "label value" like for `positive_class`?

maxnoe: > To me "label" sounds very string-like

This is the "label" in the machine learning sense: i.e. the true / target quantity.

Tobychev: Label value sounds fine @LukasBeiske

Tobychev (approved):

maxnoe (approved):

PR #2460, opened 2023-11-17
Use editable install to fix coverage

maxnoe:


maxnoe:
Using the editable install fixed coverage.py being able to associate the coverage report with the source files, so now we have correctly working coverage reports again.

Please review @Tobychev @LukasNickel 

The muon fit fix is temporary, a proper fix is under investigation  in #2425

Tobychev:
I don't understand what the problem being solved is, nor why doing an editable install would matter so I don't think I can provide a meaningful review.

maxnoe:
> I don't understand what the problem being solved is

Codecov reports coverage of 60% on 30 lines, because it doesn't recognize the file paths being tested because we run the tests in a temporary directory using the package installed into the prefix of the python installation.

See: https://app.codecov.io/gh/cta-observatory/ctapipe/

>  why doing an editable install would matter. 

Doing an editable install means the installation is just creating symlinks to the checked out repository, which means that codecov can link the executed code to the source files correctly.

>  I don't even understand why the muon test needs to be patched in this PR

To be honest, me neither. The test started failing after I added the `-e` and I fixed it to see if the coverage was ok after I got the tests to run using the editable install, which it does now, see:

https://app.codecov.io/gh/cta-observatory/ctapipe/pull/2460/indirect-changes

Tobychev (approved):

LukasNickel (approved):

PR #2465, opened 2023-11-20
Use signal_fraction for training particle classifier 

LukasBeiske:
This removes the `n_signal` and `n_background` options of `ctapipe-train-particle-classifier`. Instead the total number of training events `n_events` and the `signal_fraction` can be chosen, where
$$\texttt{signal\\_fraction}= \frac{n_s}{n_s + n_b}.$$
If `n_events` is not specified, as many events as possible will be used considering the given `signal_fraction`.

- [x]  If #2455 is merged before this, I will change the default config file accordingly in this PR.

### DIFF
--- a/docs/changes/2465.api.rst
+++ b/docs/changes/2465.api.rst
@@ -1,0 +1,2 @@
+Replace ``n_signal`` and ``n_background`` options in ``ctapipe-train-particle-classifier``
+with ``n_events`` and ``signal_fraction``, where ``signal_fraction`` = n_signal / (n_signal + n_background).
--- a/src/ctapipe/resources/train_particle_classifier.yaml
+++ b/src/ctapipe/resources/train_particle_classifier.yaml
@@ -6,13 +6,11 @@
 # ======================================================================
 
 TrainParticleClassifier:
-  random_seed: 0              # Seed used for sampling n_* events for training.
-  n_signal:                   # The number of signal events used for training that can be provided
+  random_seed: 0              # Seed used for sampling n_events for training.
+  n_events:                   # The number of events used for training that can be provided
     # - [type, "LST*", 1000]  # independently for each telescope type (e.g. "LST_LST_LSTCam").
-    # - [type, "MST*", 1000]  # If not specified, all events in the file are used.
-  n_background:               # Same as above, but for background events.
-    # - [type, "LST*", 1000]
-    # - [type, "MST*", 1000]
+    # - [type, "MST*", 1000]  # If not specified, as many events as possible are used.
+  signal_fraction: 0.5        # signal_fraction = n_signal / n_events
 
   CrossValidator:
     n_cross_validations: 5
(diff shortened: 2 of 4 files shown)


ctapipe/tools/train_particle_classifier.py
```diff
@@ -193,8 +207,26 @@ def _read_input_data(self, tel_type):
             feature_names=feature_names,
             rng=self.rng,
             log=self.log,
-            n_events=self.n_background.tel[tel_type],
+            n_events=n_background,
         )
+        if n_events is None:  # use as many events as possible (keeping signal_fraction)
+            if len(signal) < len(background) / (1 / self.signal_fraction - 1):
```
Tobychev, line 213: I think using the condition written in the form `N_tot*signal_fraction > N_signal` is much clearer than the formulation you are using here, so 
```
len(signal) < (len(signal) + len(background))*self.signal_fraction 
```
but preferably you define the constants `N_s` and `N_b` as the lengths of the signal and background input samples, and use those instead of `len(signal)` and `len(background)` in the above expression.

Tobychev (changes requested):

LukasBeiske:
I'm not sure why this provenance test is failing right now. It works on my machine...

Tobychev (dismissed):

Tobychev (approved):

Tobychev (approved):

ctapipe/resources/train_particle_classifier.yaml
```diff
@@ -6,13 +6,11 @@
...
-  n_signal:                   # The number of signal events used for training that can be provided
+  random_seed: 0              # Seed used for sampling n_events for training.
+  n_events:                   # The number of events used for training that can be provided
     # - [type, "LST*", 1000]  # independently for each telescope type (e.g. "LST_LST_LSTCam").
-    # - [type, "MST*", 1000]  # If not specified, all events in the file are used.
-  n_background:               # Same as above, but for background events.
-    # - [type, "LST*", 1000]
-    # - [type, "MST*", 1000]
+    # - [type, "MST*", 1000]  # If not specified, as many events as possible are used.
+  signal_fraction: 0.5        # signal_fraction = n_signal / n_events
```
maxnoe, line 13: I would find it more intuitive I think to define the ratio? 

I.e. 1.0 means use as much signal as background?

Tobychev: I think signal fraction being how much signal is in the total is pretty intuitive?

LukasBeiske: I think there are merits to both options:
- signal fraction in combination with `n_events` makes it a little bit more intuitive how many signal and how many background events are used, but we write out this information in the logs anyway.
- signal ratio is maybe a bit closer to actual use cases, e.g. "Let's use twice as many signal events than background events".

I don't have a personal preference.

maxnoe (approved):

Tobychev (approved):

Issue #1843, opened 2022-02-17
Definition of TelescopeFrame is incompatible with GADF FoV Coordinate Definition


maxnoe:
**Describe the bug**

From the GADF FoV coordinates section:

>The other FOV coordinates depend on the alignment and orientation of a second coordinate systems (OTHER, either ALTAZ or RADEC).

>* FOV PHI is counterclockwise from OTHER north, i.e. PHI=0 deg pointing to OTHER LAT, and PHI=270 deg pointing to OTHER LON
>* FOV LON should increase with decreasing OTHER LON
>* FOV LAT should increase with increasing OTHER LAT
```


We have LON increasing with OTHER LON (azimuth) in `TelescopeFrame`. We should invert the sign in the transform graph.

That should introduce a couple of signs in the `AltAz` ↔ `TelescopeFrame` and `TelescopeFrame` ↔ `CameraFrame` transformations.

PR #2473, opened 2023-11-24
Irf maker and cut optimiser

Tobychev:
As we are planning to produce IRFs on events that were not used to select the GH cut, it makes sense to split the procedure generating IRFs into two tools. Because this splitting results in a large restructuring of the  the previous PR on generating IRFs (#2315)  that was basically working properly, I open a new PR to preserve the discussion on the old attempt and let its more or less functional code serve as easy reference.


ctapipe/irf/binning.py
```diff
@@ -0,0 +1,176 @@
+import astropy.units as u
+import numpy as np
+from pyirf.binning import create_bins_per_decade
+
+from ..core import Component
+from ..core.traits import Float, Integer
+
+
+class OutputEnergyBinning(Component):
```
maxnoe, line 9: I'd better like this to be closer to where the things are actually used and split up into components.

We might not want to use the same energy binning for effective area and higher dimensional IRFs for example.

ctapipe/irf/irf_classes.py
```diff
@@ -0,0 +1,74 @@
+"""
+Define a parent IrfTool class to hold all the options
```
maxnoe, line 2: Wrong module

ctapipe/irf/irf_classes.py
```diff
@@ -0,0 +1,74 @@
...
+from ..core.traits import Float, Integer
+
+
+class Spectra(Enum):
+    CRAB_HEGRA = 1
+    IRFDOC_ELECTRON_SPECTRUM = 2
+    IRFDOC_PROTON_SPECTRUM = 3
+
+
+PYIRF_SPECTRA = {
```
maxnoe, line 20: I'd just call this "spectra" because we might define our own using the base classes from pyirf

ctapipe/irf/irf_classes.py
```diff
@@ -0,0 +1,74 @@
...
+        default_value=10,
+        help="Minimum number of events in a bin to attempt to find a cut value",
+    ).tag(config=True)
+
+    theta_fill_value = Float(
+        default_value=0.32, help="Angular cut value used for bins with too few events"
+    ).tag(config=True)
+
+    theta_smoothing = Float(
+        default_value=-1,
```
maxnoe, line 46: What does `-1` mean here? If that is supposed to mean "option not given, do not apply smoothing", please use `default_value=None, allow_none=True`.

ctapipe/irf/irf_classes.py
```diff
@@ -0,0 +1,74 @@
...
+        help="When given, the width (in units of bins) of gaussian smoothing applied (-1)",
+    ).tag(config=True)
+
+    target_percentile = Float(
+        default_value=68,
+        help="Percent of events in each energy bin keep after the theta cut",
+    ).tag(config=True)
+
+    def calculate_theta_cuts(self, theta, reco_energy, energy_bins):
```
maxnoe, line 55: Classes that basically are configurable functions (i.e. have traitlets and then a single method) should use `__call__` I think.

ctapipe/irf/binning.py
```diff
@@ -0,0 +1,176 @@
...
+                self.fov_offset_min,
+                self.fov_offset_max,
+                self.fov_offset_n_edges,
+            )
+            * u.deg
+        )
+        return fov_offset
+
+
+class SourceOffsetBinning(Component):
```
maxnoe, line 127: This is specific to the PSF format, so it should only be configured there

ctapipe/irf/optimise.py
```diff
@@ -0,0 +1,149 @@
...
+        self.gh_cuts = gh_cuts
+        if gh_cuts:
+            self.gh_cuts.meta["extname"] = "GH_CUTS"
+        if offset_lim and isinstance(offset_lim[0], list):
+            self.offset_lim = offset_lim
+        else:
+            self.offset_lim = [offset_lim]
+
+    def write(self, out_name, precuts, overwrite=False):
+        if isinstance(precuts, QualityQuery):
```
maxnoe, line 25: I don't understand the branching here. `precuts` is required, but you check its type and if it is anything but a `QualityQuery` you ignore whatever the user passed in and create a new empty quality query...

ctapipe/irf/optimise.py
```diff
@@ -0,0 +1,149 @@
...
+
+from ..core import Component, QualityQuery
+from ..core.traits import Float
+
+
+class OptimisationResult:
+    def __init__(self, gh_cuts=None, offset_lim=None):
+        self.gh_cuts = gh_cuts
+        if gh_cuts:
+            self.gh_cuts.meta["extname"] = "GH_CUTS"
```
maxnoe, line 18: extname is the name of a FITS extension. You should't set this in the metadata of the table, but only when writing out, it's the `name` argument of `write`.

ctapipe/irf/optimise.py
```diff
@@ -0,0 +1,149 @@
...
+from astropy.table import QTable, Table
+from pyirf.binning import create_bins_per_decade
+from pyirf.cut_optimization import optimize_gh_cut
+from pyirf.cuts import calculate_percentile_cut, evaluate_binned_cut
+
+from ..core import Component, QualityQuery
+from ..core.traits import Float
+
+
+class OptimisationResult:
```
maxnoe, line 14: theta cuts missing here

ctapipe/irf/select.py
```diff
@@ -0,0 +1,205 @@
...
+from ..core import Component, Provenance, QualityQuery
+from ..core.traits import List, Unicode
+from ..io import TableLoader
+
+
+class EventSelector(Component):
+    def __init__(self, event_pre_processor, kind, file, target_spectrum, **kwargs):
+        super().__init__(**kwargs)
+
+        self.epp = event_pre_processor
```
maxnoe, line 17: Why is `EventPreprocessor` given as positional argument here? It should be a subcomponent configured via the config system.

Also the naming here is a bit strange, `EventPreprocessor` inherits from the quality query, so it makes selections and this `EventSelector` uses a `TableLoader` to read things...

ctapipe/irf/select.py
```diff
@@ -0,0 +1,205 @@
...
+        self.target_spectrum = target_spectrum
+        self.kind = kind
+        self.file = file
+
+    def load_preselected_events(self, chunk_size, obs_time, fov_bins):
+        opts = dict(load_dl2=True, load_simulated=True, load_dl1_parameters=False)
+        with TableLoader(self.file, **opts) as load:
```
maxnoe, line 24: parent=self missing to correctly setup logging and configuration.

ctapipe/irf/select.py
```diff
@@ -0,0 +1,205 @@
...
+            Provenance().add_input_file(self.file)
```
maxnoe, line 25: That should be in tableloader, so not needed.

ctapipe/irf/select.py
```diff
@@ -0,0 +1,205 @@
...
+        help="Prefix of the `_alt` and `_az` reco geometry columns",
+    ).tag(config=True)
+    gammaness_classifier = Unicode(
+        default_value="RandomForestClassifier",
+        help="Prefix of the classifier `_prediction` column",
+    ).tag(config=True)
+
+    quality_criteria = List(
+        default_value=[
+            ("multiplicity 4", "np.count_nonzero(tels_with_trigger,axis=1) >= 4"),
```
maxnoe, line 131: Do we really want a multiplicity cut of 4 by default? I would remove that and only require the valid reconstruction.

ctapipe/tools/make_irf.py
```diff
@@ -0,0 +1,353 @@
...
+    OptimisationResult,
+    OutputEnergyBinning,
+    SourceOffsetBinning,
+    Spectra,
+    ThetaCutsCalculator,
+)
+
+
+class IrfTool(Tool):
+    name = "ctapipe-make-irfs"
```
maxnoe, line 42: Our terminology is a bit imprecises, but according to the DL3 data model document, the `IRF` is complete parametrization composed of *IRF components* (effective area, edisp, ...).

This tool computes thus a single IRF. Long story short, I'd call this `ctapipe-compute-irf` without the s.

ctapipe/tools/make_irf.py
```diff
@@ -0,0 +1,353 @@
...
+        self.opt_result.read(self.cuts_file)
+        self.epp.quality_criteria = self.opt_result.precuts
+        self.reco_energy_bins = self.e_bins.reco_energy_bins()
+        self.true_energy_bins = self.e_bins.true_energy_bins()
+        self.energy_migration_bins = self.e_bins.energy_migration_bins()
+
+        self.source_offset_bins = self.bins.source_offset_bins()
+        self.fov_offset_bins = self.opt_result.offset_lim
+
+        self.particles = [
```
maxnoe, line 125: Protons and electrons should probably be optional. You only need them to compute a background model, which we maybe should split into a separate tool (`ctapipe-compute-background-from-simulations`) or similar, since we won't use simulations for computing the background component later.

ctapipe/tools/make_irf.py
```diff
@@ -0,0 +1,353 @@
...
+            self.opt_result.gh_cuts,
+            operator.ge,
+        )
+        self.theta_cuts_opt = self.theta.calculate_theta_cuts(
+            self.signal_events[self.signal_events["selected_gh"]]["theta"],
+            self.signal_events[self.signal_events["selected_gh"]]["reco_energy"],
+            self.reco_energy_bins(),
+        )
+
+        self.signal_events["selected_theta"] = evaluate_binned_cut(
```
maxnoe, line 184: Applying the theta cut should be optional, we don't need it for computing general (not point-like) IRFs.

ctapipe/tools/make_irf.py
```diff
@@ -0,0 +1,353 @@
...
+        # species
+        self.log.debug(
+            "Keeping %d signal, %d backgrond events"
+            % (
+                sum(self.signal_events["selected"]),
+                sum(self.background_events["selected"]),
+            )
+        )
+
+        # calculate sensitivity
```
maxnoe, line 211: computing sensitivity should be separate from computing IRFs I think.

ctapipe/tools/make_irf.py
```diff
@@ -0,0 +1,353 @@
...
+        self.sensitivity = calculate_sensitivity(
+            signal_hist, background_hist, alpha=self.alpha
+        )
+
+        # scale relative sensitivity by Crab flux to get the flux sensitivity
+        self.sensitivity["flux_sensitivity"] = self.sensitivity[
+            "relative_sensitivity"
+        ] * self.spectrum(self.sensitivity["reco_energy_center"])
+
+    def finish(self):
```
maxnoe, line 234: It seems weird to me to do the actual IRF computations in `finish`, I'd move them to the main `run` method.

ctapipe/irf/optimise.py
Tobychev: Yeah, this part is very rushed so I suspect something simply got lost.

ctapipe/tools/make_irf.py
```diff
@@ -0,0 +1,353 @@
...
+        masks = {
+            "": self.signal_events["selected"],
+            "_NO_CUTS": slice(None),
```
maxnoe, line 237: The IRF tool should probably only compute a single effective area, depending on user options it should be either the point-like (apply gammaness and theta cuts) or the full-enclosure one (only apply gammaness cuts)

ctapipe/tools/make_irf.py
```diff
@@ -0,0 +1,353 @@
...
+                    edisp,
+                    true_energy_bins=self.true_energy_bins,
+                    migration_bins=self.energy_migration_bins,
+                    fov_offset_bins=self.fov_offset_bins,
+                    extname="ENERGY_DISPERSION" + label,
+                )
+            )
+        # Here we use reconstructed energy instead of true energy for the sake of
+        # current pipelines comparisons
+        bias_resolution = energy_bias_resolution(
```
maxnoe, line 282: bias_resolution are benchmarks that can be computed from events (like here) or from the IRFs (functions in pyirf or methods of the irf objects in gammapy), I'd not compute them here in the IRF tool.

ctapipe/tools/make_irf.py
```diff
@@ -0,0 +1,353 @@
...
+            self.signal_events[self.signal_events["selected"]],
+            self.true_energy_bins,
+            bias_function=np.mean,
+            energy_type="true",
+        )
+        hdus.append(fits.BinTableHDU(bias_resolution, name="ENERGY_BIAS_RESOLUTION"))
+
+        # Here we use reconstructed energy instead of true energy for the sake of
+        # current pipelines comparisons
+        ang_res = angular_resolution(
```
maxnoe, line 292: same here, it's a benchmark, not an IRF, so I'd say it doesn't belong into the IRF tool

ctapipe/tools/make_irf.py
```diff
@@ -0,0 +1,353 @@
...
+        )
+        hdus.append(
+            create_background_2d_hdu(
+                background_rate,
+                self.reco_energy_bins,
+                fov_offset_bins=self.fov_offset_bins,
+            )
+        )
+
+        psf = psf_table(
```
maxnoe, line 316: Compute PSF only if not pointlike.

Should be a configurable class so we can later extend to computing different variants.

ctapipe/irf/select.py
Tobychev: fixed

ctapipe/irf/optimise.py
Tobychev: I got
```
TypeError: write_table_fits() got an unexpected keyword argument 'name'
```
when I tried `self.gh_cuts.write(out_name, name="GH_CUTS", format="fits", overwrite=overwrite)` and I don't find anything about a `name` argument to `table.write`, do you mean I should manage the fits file more directly and just cast the tables to hdus?

ctapipe/irf/irf_classes.py
Tobychev: fixed

ctapipe/irf/irfs.py
```diff
@@ -0,0 +1,157 @@
...
+        help="Maximum value for Source offset for PSF IRF",
+        default_value=1,
+    ).tag(config=True)
+
+    source_offset_n_edges = Integer(
+        help="Number of edges for Source offset for PSF IRF",
+        default_value=101,
+    ).tag(config=True)
+
+    def __init__(self, parent, energy_bins, valid_offset):
```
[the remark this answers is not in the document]
Tobychev: fixed

ctapipe/irf/irf_classes.py
Tobychev: After recently having tried to explain how `__call__` means you can do object(args) to a beginner, I think that syntax has too much magic and explicit is better.

ctapipe/irf/binning.py
Tobychev: Made a Psf class.

ctapipe/irf/irf_classes.py
maxnoe: I think it's more important to be consistent. If that's a decision we should make, we should make it everywhere and not just here. However, regarding this particular point, I have a different opinion.

We have many instances in ctapipe now that implement `__call__` and I think it's the correct thing to do. 

Beginners have problems with many concepts, that shouldn't mean that we optimize our code for being understandable by beginners vs. writing idiomatic python.

ctapipe/tools/make_irf.py
Tobychev, line 1: I suspect this is incorrect because it applies a theta cut we don't want for the background estimation, could you confirm if the `backgrond_2d` function expects a theta cut applied @maxnoe ?

LukasBeiske: I just noticed, that you are right. A theta cut does not make sense here. But I am currently working making both point like and full enclosure IRFs possible and the background will only be done for full enclosure, so it would no difference in the end, but I will fix it now :+1:

ctapipe/irf/binning.py
```diff
@@ -0,0 +1,104 @@
...
+        default_value=0.015,
+    ).tag(config=True)
+
+    reco_energy_max = Float(
+        help="Maximum value for Reco Energy bins in TeV units",
+        default_value=200,
+    ).tag(config=True)
+
+    reco_energy_n_bins_per_decade = Float(
+        help="Number of edges per decade for Reco Energy bins",
```
maxnoe, line 47: Name is `n_bins`, help says `n_edges`

maxnoe: Value means it should be `n_bins` like the variable name, so the help needs to be updated.

LukasBeiske: :+1:  Thats a problem at multiple places. Will fix.

ctapipe/tools/make_irf.py
maxnoe: No, no theta cut must be applied to the background.

You can't really compute `theta` for background as you would have to define an assumed source position, using the true origin of the background particle is not correct.

The background rate is per solid angle, so it gets scaled to the on region size by the science tools (or by the sensitivity computation code).

src/ctapipe/irf/optimize.py
```diff
@@ -156,18 +145,248 @@ class GridOptimizer(Component):
...
+            valid_energy=[self.reco_energy_min, self.reco_energy_max],
+            valid_offset=[min_fov_radius, max_fov_radius],
+            clf_prefix=clf_prefix,
+            theta_cuts=theta_cuts if point_like else None,
+        )
+
+        return result_saver
+
+
+class GridOptimizer(CutOptimizerBase):
```
maxnoe, line 354: I'd call this something like "PointSourceSensitivityOptimizer"

src/ctapipe/irf/irfs.py
```diff
@@ -0,0 +1,385 @@
...
+            np.linspace(
+                self.fov_offset_min.to_value(u.deg),
+                self.fov_offset_max.to_value(u.deg),
+                self.fov_offset_n_bins + 1,
+            ),
+            u.deg,
+        )
+
+
+class PsfIrfBase(IrfTrueEnergyBase):
```
maxnoe, line 118: I'd not call these classes "PsfIrf" or similar. They don't full fill the *is a* criterion.... They *are* not IRFs, they *compute* IRFs. 

So I'd call them something like `PSF3DMaker`

docs/changes/2315.irf-maker.rst
```diff
@@ -0,0 +1 @@
+Add a `make-irf tool` able to produce irfs given a gamma, proton and electron DL2 input files.
```
maxnoe, line 1: This file needs to be removed

docs/changes/2473.feature.rst
```diff
@@ -0,0 +1 @@
+Add a `make-irf tool` able to produce irfs given a gamma, proton and electron DL2 input files.
```
maxnoe, line 1: You need double ticks in rst and it should be : `ctapipe-make-irf`

docs/changes/2315.irf-maker.rst
maxnoe: I only see two change log entries here. The other one uses the correct PR number and topic

src/ctapipe/irf/vis_utils.py
```diff
@@ -0,0 +1,64 @@
+import numpy as np
```
maxnoe, line 1: The commit message says "Fixed some quality issues in sonar" but it also adds a couple of new files. So I think that was probably by mistake.

Tobychev: No, part of the things sonar raised was code duplication, so collected various repeated snippets into new functions and put those into a separate files.

src/ctapipe/tools/make_irf.py
```diff
@@ -279,20 +279,19 @@ def setup(self):
             "Aeff fov offset",
             raise_error=self.range_check_error,
         )
-        if self.full_enclosure:
```
LukasBeiske: I don't think it makes sense to calculate the PSF for point-like irfs, since point-like irfs assume you know the position of your source anyway and the theta cut also truncates the PSF, no?

Tobychev: The PSF simply measures how far away from the true position photons are being reconstructed, so it can always be defined. 

What matters for the PSF is the difference between diffuse simulations and point-like, since for the latter we only known of the behaviour for photons coming from a single position, so we can't talk about what happens when the source is placed in other parts of the field of view. Note that this is an entirely separate issue from the full_enclosure request. 

It is true that for point-like irfs there is a cut placed on how far away from the source position photons can be while being included, but you still need to know the spatial distribution of the the photons that are kept, so you need the PSF.

maxnoe: No, the PDF is not used in point like analyses.

maxnoe: See e.g.: https://github.com/gammapy/gammapy/blob/011cefef8916ebc9584fa06f52d999ca01af78ff/gammapy/data/data_store.py#L23

maxnoe: It is even stronger than "we don't use it". The whole reason we do point-like IRFs is to simplify the IRF to have one less component: the PSF:

See section A.2 of https://redmine.cta-observatory.org/dmsf/files/17785/view

src/ctapipe/irf/__init__.py
```diff
@@ -0,0 +1,72 @@
+"""Top level module for the irf functionality"""
+from .benchmarks import (
+    AngularResolution2dMaker,
+    AngularResolutionMakerBase,
```
kosack, line 4: You should probably avoid exposing the base classes here, as those are internal to the package

src/ctapipe/irf/binning.py
```diff
@@ -0,0 +1,156 @@
...
+import astropy.units as u
+import numpy as np
+
+from ..core import Component
+from ..core.traits import AstroQuantity, Integer
+
+logger = logging.getLogger(__name__)
+
+
+def check_bins_in_range(bins, valid_range, source="result", raise_error=True):
```
kosack, line 14: missing docstring

kosack:
I tried testing this and get the following error when running `ctapipe-make-irf` when using all default options (also for generating the cuts file):

```
ValueError: Valid range for background reco energy is 0.15 to 150., got [1.5000e-02 2.3773e-02 ... 9.4644e+01 1.5000e+02]
ValueError: Valid range for Edisp true energy is 0.15 to 150., got [1.5000e-02 1.8884e-02 ... 1.1915e+02 1.5000e+02]
```
From the error message, it looks like the low-energy range is off by a factor of 10, e.g. the bins go from (0.015, 150) instead of (0.15,150). Is that some unit error somewhere (GeV instead of TeV)?

maxnoe:
Why don't we allow energies above 150 TeV? or is that value taken from the CORSIKA headers?

kosack:
> Why don't we allow energies above 150 TeV? or is that value taken from the CORSIKA headers?

It think that's just the current default (See below, I don't think they are read from anywhere). It's just that currently when using defaults, they don't seem to agree for the minimum value. I would expect the tool to work correctly with default values.  


image

LukasBeiske:
> > Why don't we allow energies above 150 TeV? or is that value taken from the CORSIKA headers?
> 
> It think that's just the current default (See below, I don't think they are read from anywhere). It's just that currently when using defaults, they don't seem to agree for the minimum value. I would expect the tool to work correctly with default values.
> image

Yes, exactly. This are just the current default values. I'll take a look why it doesn't work.

LukasBeiske:
> I tried testing this and get the following error when running `ctapipe-make-irf` when using all default options (also for generating the cuts file):
> 
> ```
> ValueError: Valid range for background reco energy is 0.15 to 150., got [1.5000e-02 2.3773e-02 ... 9.4644e+01 1.5000e+02]
> ValueError: Valid range for Edisp true energy is 0.15 to 150., got [1.5000e-02 1.8884e-02 ... 1.1915e+02 1.5000e+02]
> ```
> 
> From the error message, it looks like the low-energy range is off by a factor of 10, e.g. the bins go from (0.015, 150) instead of (0.15,150). Is that some unit error somewhere (GeV instead of TeV)?

I'm not able to reproduce this locally on my end.  The default values for `ctapipe.irf.binning.{True,Reco}EnergyBinsBase.{true,reco}_energy_min` and `ctapipe.irf.optimize.CutOptimizerBase.reco_energy_min` also seem to be correct at 0.015 TeV, which are the only places where the default values for any energy ranges are defined.

kosack:
What I ran was:
```sh
ctapipe-optimize-event-selection  --log-level=INFO \
    --gamma-file=gamma_60deg_180deg_run10000___cta-prod5b-paranal_desert-2147m-Paranal-dark_cone10.alpha_test_applied.DL2.h5  \
     --proton-file=proton_60deg_180deg_run10000___cta-prod5b-paranal_desert-2147m-Paranal-dark.alpha_test_alpha_test_applied.DL2.h5 \
    --electron-file electron_60deg_180deg_run1000___cta-prod5b-paranal_desert-2147m-Paranal-dark.alpha_test_applied.DL2.h5

ctapipe-make-irf  --log-level=INFO \
    --gamma-file=gamma_60deg_180deg_run10000___cta-prod5b-paranal_desert-2147m-Paranal-dark_cone10.alpha_test_applied.DL2.h5   \
    --proton-file=proton_60deg_180deg_run10000___cta-prod5b-paranal_desert-2147m-Paranal-dark.alpha_test_alpha_test_applied.DL2.h5 \
    --electron-file electron_60deg_180deg_run1000___cta-prod5b-paranal_desert-2147m-Paranal-dark.alpha_test_applied.DL2.h5 \
    --cuts Selection_Cuts.fits
```
The output is:
```
2024-06-14 17:35:29,052 ERROR [ctapipe.ctapipe-make-irf] (tool.run): Caught unexpected exception: Valid range for background reco energy is 0.15 to 150., got [1.5000e-02 2.3773e-02 ... 9.4644e+01 1.5000e+02]
  File "/Users/kkosack/Projects/CTA/Working/ctapipe/src/ctapipe/core/tool.py", line 413, in run
    self.setup()
  File "/Users/kkosack/Projects/CTA/Working/ctapipe/src/ctapipe/tools/make_irf.py", line 258, in setup
    check_e_bins(
  File "/Users/kkosack/Projects/CTA/Working/ctapipe/src/ctapipe/irf/binning.py", line 24, in check_bins_in_range
    raise ValueError(
ValueError: Valid range for background reco energy is 0.15 to 150., got [1.5000e-02 2.3773e-02 ... 9.4644e+01 1.5000e+02]
2024-06-14 17:35:29,062 INFO [ctapipe.ctapipe-make-irf] (tool.write_provenance): Output:
```

So somewhere it is getting 0.15 as a lower energy bin.  If I add `--no-do-background`, I get the same error, but for the edisp instead of the background

LukasBeiske:
> What I ran was:
> 
> ```shell
> ctapipe-optimize-event-selection  --log-level=INFO \
>     --gamma-file=gamma_60deg_180deg_run10000___cta-prod5b-paranal_desert-2147m-Paranal-dark_cone10.alpha_test_applied.DL2.h5  \
>     --proton-file=proton_60deg_180deg_run10000___cta-prod5b-paranal_desert-2147m-Paranal-dark.alpha_test_alpha_test_applied.DL2.h5 \
>     --electron-file=electron_60deg_180deg_run1000___cta-prod5b-paranal_desert-2147m-Paranal-dark.alpha_test_applied.DL2.h5
> 
> ctapipe-make-irf  --log-level=INFO \
>     --gamma-file=gamma_60deg_180deg_run10000___cta-prod5b-paranal_desert-2147m-Paranal-dark_cone10.alpha_test_applied.DL2.h5   \
>     --proton-file=proton_60deg_180deg_run10000___cta-prod5b-paranal_desert-2147m-Paranal-dark.alpha_test_alpha_test_applied.DL2.h5 \
>     --electron-file electron_60deg_180deg_run1000___cta-prod5b-paranal_desert-2147m-Paranal-dark.alpha_test_applied.DL2.h5 \
>     --cuts Selection_Cuts.fits
> ```
> 
> The output is:
> 
> ```
> 2024-06-14 17:35:29,052 ERROR [ctapipe.ctapipe-make-irf] (tool.run): Caught unexpected exception: Valid range for background reco energy is 0.15 to 150., got [1.5000e-02 2.3773e-02 ... 9.4644e+01 1.5000e+02]
> Traceback (most recent call last):
>   File "/Users/kkosack/Projects/CTA/Working/ctapipe/src/ctapipe/core/tool.py", line 413, in run
>     self.setup()
>   File "/Users/kkosack/Projects/CTA/Working/ctapipe/src/ctapipe/tools/make_irf.py", line 258, in setup
>     check_e_bins(
>   File "/Users/kkosack/Projects/CTA/Working/ctapipe/src/ctapipe/irf/binning.py", line 24, in check_bins_in_range
>     raise ValueError(
> ValueError: Valid range for background reco energy is 0.15 to 150., got [1.5000e-02 2.3773e-02 ... 9.4644e+01 1.5000e+02]
> 2024-06-14 17:35:29,062 INFO [ctapipe.ctapipe-make-irf] (tool.write_provenance): Output:
> ```
> 
> So somewhere it is getting 0.15 as a lower energy bin. If I add `--no-do-background`, I get the same error, but for the edisp instead of the background.
> 
> Edit: seems to be in the valid energy stored in the Selection_Cuts.fits:
> 
> ```
> % showtable Selection_Cuts.fits --hdu VALID_ENERGY                                         
>      energy_min         energy_max
>         TeV                TeV
> ------------------- ------------------
> 0.14999999999999997 150.00000000000003
> 
> % showtable Selection_Cuts.fits --hdu GH_CUTS                                              
>         low                 center                high                 cut
>         TeV                  TeV                  TeV
> -------------------- -------------------- -------------------- -------------------
> 0.014999999999999996 0.019386698943458347 0.023773397886916695  0.9754451779370514
> 0.023773397886916695 0.030725847179780198   0.0376782964726437  0.9754451779370514
>   0.0376782964726437  0.04869718602783414  0.05971607558302459  0.9754451779370514
>  0.05971607558302459  0.07717983862752678  0.09464360167202897  0.9754451779370514
>  0.09464360167202897  0.12232180083601446  0.14999999999999997  0.9754451779370514
>  0.14999999999999997  0.19386698943458353  0.23773397886916706  0.4092728996725426
>  0.23773397886916706    0.307258471797802    0.376782964726437  0.7170864798156927
> ```
> 
> I looked through the code, and so far cannot find anywhere where a factor of 10 (or wrong units) gets applied in the valid energy range... It's a bit of a mystery
> 
> ```
> 462
> 463  	        result_saver = OptimizationResultStore(precuts)
> 464  ->	        result_saver.set_result(
> 465  	            gh_cuts=gh_cuts,
> 466  	            valid_energy=valid_energy,
> 467  	            valid_offset=[self.min_fov_offset, self.max_fov_offset],
> 468  	            clf_prefix=clf_prefix,
> 469  	            theta_cuts=theta_cuts_opt if point_like else None,
> (Pdb) valid_energy
> [<Quantity 0.15 TeV>, <Quantity 150. TeV>]
> (Pdb) self.reco_energy_min
> <Quantity 0.015 TeV>
> ```
> 
> The value 0.15 TeV seems to come from the function `_get_valid_energy_range()`, so it's the minimum finite value of the sensitivity, which should be larger than the reco binning range. So I guess the question is why is there a check that they are the same? I would just guess you want to check if the reco range is larger or equal to the valid range or something.

Thanks for tracking this down. As far as I understood it, the optimization of the G/H cut only gives a valid result if a finite sensitivity could be computed. So for bins where this was not possible, the calculated G/H cut is not valid. 
And this is the main idea behind this check in `ctapipe-make-irf`: To see if the provided G/H (and theta) cuts are valid for the whole energy range for which the IRF should be computed.

So to avoid this problem while testing this code, it might be better to not use this optimization for maximum sensitivity by default, but instead only do percentile cuts by default (`ctapipe.irf.optimize.PercentileCuts`), where the valid energy range is always the one defined during configuration.
Right now, you could run `ctapipe-make-irf` with `--IrfEventSelector.optimization_algorithm="PercentileCuts"` and everything should work.

kosack:
> Thanks for tracking this down. As far as I understood it, the optimization of the G/H cut only gives a valid result if a finite sensitivity could be computed. So for bins where this was not possible, the calculated G/H cut is not valid.
And this is the main idea behind this check in ctapipe-make-irf: To see if the provided G/H (and theta) cuts are valid for the whole energy range for which the IRF should be computed.

I think the issues is maybe that the chosen binning shouldn't matter: if the requested bin range is larger than what can be computed, the IRFs should just have NaN (until we have a better way to store invalid masks) in the bins that are outside the valid energy range.  Required the user to know in advance what range will be valid is less useful, particularly if we want to automate this for a large number of analyses.

It's the same as if we have low statistics that prevent an IRF from being computed, we would want to ensure that those bins are still stored, but marked properly as unusable.

LukasBeiske:
> I think the issues is maybe that the chosen binning shouldn't matter: if the requested bin range is larger than what can be computed, the IRFs should just have NaN (until we have a better way to store invalid masks) in the bins that are outside the valid energy range.  Required the user to know in advance what range will be valid is less useful, particularly if we want to automate this for a large number of analyses.
> 
> It's the same as if we have low statistics that prevent an IRF from being computed, we would want to ensure that those bins are still stored, but marked properly as unusable.

Good point. I'm at the CTAO Summer School at the moment, but once I'm back I'll change it accordingly.

LukasBeiske:
When looking at it again, I noticed that most of these bin-in-range checks did not make sense anyway. They compared the true energy bins of the irfs with the reco energy range for which the cuts are calculated. All the the fov offset checks also did not make sense, since we do not calculate the cuts in fov bins yet and instead we compared it with the fov offset range used for the background events in the sensitivity estimation in `PointSourceSensitivityOptimizer`.

For the remaining checks, I switched the default to only give a warning if the requested irf binning is outside the valid range of the cuts.
The reco energy bins for the background which are actually outside the range of the cuts are currently filled with 0 in the corresponding pyirf functions. I'll open a PR there to instead have them be NaN. This is also the case for empty bins in other irfs.

src/ctapipe/irf/optimize.py
```diff
@@ -0,0 +1,483 @@
...
+
+
+class OptimizationResultStore:
+    def __init__(self, precuts=None):
+        self._init_precuts(precuts)
+        self._results = None
+
+    def _init_precuts(self, precuts):
+        if precuts:
+            if isinstance(precuts, QualityQuery):
```
maxnoe, line 52: Why do we need case distinctions here? Why is it not always `QualityQuery` or `None`?

src/ctapipe/irf/optimize.py
```diff
@@ -0,0 +1,483 @@
...
+        else:
+            return (
+                f"<OptimizationResult with {len(self.gh_cuts)} G/H bins valid "
+                f"between {self.valid_offset.min} to {self.valid_offset.max} "
+                f"and {self.valid_energy.min} to {self.valid_energy.max} "
+                f"with {len(self.precuts.quality_criteria)} precuts>"
+            )
+
+
+class OptimizationResultStore:
```
maxnoe, line 45: Why two classes for the data itself and methods to just read and write them?

Why not add the io related methods to the class itself or have simple, free functions to read / store it

src/ctapipe/irf/select.py
```diff
@@ -0,0 +1,298 @@
...
+        events = QTable(events[keep_columns], copy=False)
+        events.rename_columns(rename_from, rename_to)
+        return events
+
+    def make_empty_table(self) -> QTable:
+        """
+        This function defines the columns later functions expect to be present
+        in the event table.
+        """
+        columns = [
```
maxnoe, line 108: this has much duplication and has the danger of missing something / changing it only in one place.
it also results in the wrong dtypes for many columns (as float64 is the default and that's wrong e.g. for obs_id and event_id).


Better directly define the columns:
```
from astropy.table import Column

columns = [
    Column(name="obs_id", dtype=np.uint64, description="..."),
    Column(name="true_energy", unit=u.TeV, description="..."),
   ...
]
```

src/ctapipe/irf/select.py
```diff
@@ -0,0 +1,298 @@
...
+            "theta": "Reconstructed angular offset from source position",
+            "true_source_fov_offset": "Simulated angular offset from pointing direction",
+            "reco_source_fov_offset": "Reconstructed angular offset from pointing direction",
+            "gh_score": "prediction of the classifier, defined between [0,1], where values close to 1 mean that the positive class (e.g. gamma in gamma-ray analysis) is more likely",
+        }
+
+        return QTable(names=columns, units=units, descriptions=descriptions)
+
+
+class EventsLoader(Component):
```
maxnoe, line 163: `EventLoader` fits better I would say and is consistent with `EventPreProcessor`

src/ctapipe/irf/select.py
```diff
@@ -0,0 +1,298 @@
...
+    classes = [EventPreProcessor]
+
+    def __init__(self, kind: str, file: Path, target_spectrum: Spectra, **kwargs):
```
maxnoe, line 166: why are `kind`, `file` and `target_spectrum` keyword arguments here and not done via the configuration system?

src/ctapipe/irf/select.py
```diff
@@ -0,0 +1,298 @@
...
+                selected = self.epp.normalise_column_names(selected)
+                selected = self.make_derived_columns(selected, obs_conf)
+                bits.append(selected)
+                n_raw_events += len(events)
+
+            bits.append(header)  # Putting it last ensures the correct metadata is used
+            table = vstack(bits, join_type="exact", metadata_conflicts="silent")
+            return table, n_raw_events, meta
+
+    def get_metadata(
```
maxnoe, line 195: I'd call this `get_simulation_information` instead of `get_metadata`.

src/ctapipe/irf/select.py
```diff
@@ -0,0 +1,298 @@
...
+        return (
+            sim_info,
+            PowerLaw.from_simulation(sim_info, obstime=obs_time),
+            obs,
+        )
+
+    def make_derived_columns(self, events: QTable, obs_conf: Table) -> QTable:
+        if obs_conf["subarray_pointing_lat"].std() < 1e-3:
+            assert all(obs_conf["subarray_pointing_frame"] == 0)
+            # Lets suppose 0 means ALTAZ
```
maxnoe, line 227: You don't need  to suppose: 


https://github.com/cta-observatory/ctapipe/blob/80914fb888e09cf1bf145c330c813a5be3e6efda/src/ctapipe/containers.py#L134-L143

src/ctapipe/irf/select.py
```diff
@@ -0,0 +1,298 @@
...
+            alt=events["pointing_alt"], az=events["pointing_az"], frame=altaz
+        )
+        reco = SkyCoord(
+            alt=events["reco_alt"],
+            az=events["reco_az"],
+            frame=altaz,
+        )
+        nominal = NominalFrame(origin=pointing)
+        reco_nominal = reco.transform_to(nominal)
+        events["reco_fov_lon"] = -reco_nominal.fov_lon  # minus for GADF
```
maxnoe, line 258: I think we should add a new frame in `ctapipe.coordinates` called `GADFFovCoordinatesAltAz` or similar instead of doing this here by hand.

maxnoe: Or just make the `NominalFrame` compatible with `GADF`, see #1843

src/ctapipe/irf/select.py
```diff
@@ -0,0 +1,298 @@
...
+        if (
+            self.kind == "gammas"
+            and self.target_spectrum.normalization.unit.is_equivalent(
+                spectrum.normalization.unit * u.sr
+            )
+        ):
+            if fov_offset_bins is None:
+                raise ValueError(
+                    "gamma_target_spectrum is point-like, but no fov offset bins "
+                    "for the integration of the simulated diffuse spectrum was given."
```
maxnoe, line 277: were given (plural bins)

src/ctapipe/irf/select.py
```diff
@@ -0,0 +1,298 @@
...
+    def make_event_weights(
+        self,
+        events: QTable,
+        spectrum: PowerLaw,
+        fov_offset_bins: u.Quantity | None = None,
+    ) -> QTable:
+        if (
+            self.kind == "gammas"
+            and self.target_spectrum.normalization.unit.is_equivalent(
+                spectrum.normalization.unit * u.sr
```
maxnoe, line 271: could just compare to `DIFFUSE_FLUX_UNIT` from pyirf.

src/ctapipe/irf/spectra.py
```diff
@@ -0,0 +1,17 @@
+"""Definition of spectra to be used to calculate event weights for irf computation"""
+from enum import Enum
+
+from pyirf.spectral import CRAB_HEGRA, IRFDOC_ELECTRON_SPECTRUM, IRFDOC_PROTON_SPECTRUM
+
+
+class Spectra(Enum):
```
maxnoe, line 7: Missing docstrings.

src/ctapipe/irf/vis_utils.py
```diff
@@ -0,0 +1,64 @@
...
+
+    if isinstance(column, str):
+        mat_vals = np.squeeze(table[column])
+    else:
+        mat_vals = column
+
+    return mat_vals, xbins, ybins
+
+
+def get_bin_centers(bins):
```
maxnoe, line 51: function is already in `pyirf.binning`.

src/ctapipe/irf/visualisation.py
```diff
@@ -0,0 +1,224 @@
...
+from pyirf.binning import join_bin_lo_hi
+
+from .vis_utils import (
+    find_columnwise_stats,
+    get_2d_hist_from_table,
+    get_bin_centers,
+    get_x_bin_values_with_rebinning,
+)
+
+quantity_support()
```
maxnoe, line 15: Please don't call globally mutating functions in ctapipe on import.

Use as a context manager around the plots where needed.

src/ctapipe/irf/visualisation.py
```diff
@@ -0,0 +1,224 @@
...
+    mat_vals, xbins, ybins = get_2d_hist_from_table(x_prefix, y_prefix, table, column)
+
+    if not x_label:
+        x_label = x_prefix
+    if not y_label:
+        y_label = y_prefix
+    plot = plot_hist2d(
+        ax, mat_vals, xbins, ybins, xlabel=x_label, ylabel=y_label, **mpl_args
+    )
+    plt.colorbar(plot)
```
maxnoe, line 30: don't mix object oriented mpl syntax with pyplot style syntax.

`plt.colorbar(plot)` -> `axes.figure.colorbar(plot, ax=ax)`

src/ctapipe/tools/make_irf.py
```diff
@@ -0,0 +1,615 @@
...
+    BackgroundRateMakerBase,
+    EffectiveAreaMakerBase,
+    EnergyMigrationMakerBase,
+    PsfMakerBase,
+)
+
+
+class IrfTool(Tool):
+    name = "ctapipe-make-irf"
+    description = "Tool to create IRF files in GAD format"
```
maxnoe, line 37: GADF

maxnoe: (I know the F in GADF is for format, still, we shouldn't destroy the acronym, it's confusing)

src/ctapipe/tools/make_irf.py
```diff
@@ -0,0 +1,615 @@
...
+
+    do_background = Bool(
+        True,
+        help="Compute background rate IRF using supplied files",
```
maxnoe, line 41: we should be clear with naming here and follow the data model terms.

 *The IRF* is the whole thing, i.e. effective area + energy dispersion + psf + background. 

One of these is an *IRF Component*.

src/ctapipe/tools/make_irf.py
```diff
@@ -0,0 +1,615 @@
...
+        default_value="./IRF.fits.gz",
+        allow_none=False,
+        directory_ok=False,
+        help="Output file",
+    ).tag(config=True)
+
+    obs_time = AstroQuantity(
+        default_value=u.Quantity(50, u.hour),
+        physical_type=u.physical.time,
+        help="Observation time in the form ``<value> <unit>``",
```
maxnoe, line 110: explain for what this is needed

src/ctapipe/tools/make_irf.py
```diff
@@ -0,0 +1,615 @@
...
+    ).tag(config=True)
+
+    edisp_parameterization = traits.ComponentName(
+        EnergyMigrationMakerBase,
+        default_value="EnergyMigration2dMaker",
+        help="The parameterization of the energy migration to be used.",
+    ).tag(config=True)
+
+    aeff_parameterization = traits.ComponentName(
```
maxnoe, line 119: It's a bit weird to have these option be called "<component>_parametrization" but then the class names all end on "Maker".

Can we make this consistent? Either call the options `..._maker` or the classes `...Parametrization`?

src/ctapipe/tools/make_irf.py
```diff
@@ -0,0 +1,615 @@
...
+        ),
+    ).tag(config=True)
+
+    ang_res_parameterization = traits.ComponentName(
+        AngularResolutionMakerBase,
+        default_value="AngularResolution2dMaker",
+        help="The parameterization of the angular resolution benchmark.",
+    ).tag(config=True)
+
+    sens_parameterization = traits.ComponentName(
```
maxnoe, line 152: please don't use needless shortenings of words.

src/ctapipe/tools/make_irf.py
```diff
@@ -0,0 +1,615 @@
...
+        help="The parameterization of the point spread function to be used.",
+    ).tag(config=True)
+
+    bkg_parameterization = traits.ComponentName(
+        BackgroundRateMakerBase,
+        default_value="BackgroundRate2dMaker",
+        help="The parameterization of the background rate to be used.",
+    ).tag(config=True)
+
+    energy_bias_res_parameterization = traits.ComponentName(
```
maxnoe, line 137: resolution

src/ctapipe/tools/make_irf.py
```diff
@@ -0,0 +1,615 @@
...
+        EnergyBiasResolutionMakerBase,
+        default_value="EnergyBiasResolution2dMaker",
+        help=(
+            "The parameterization of the bias and resolution benchmark "
+            "for the energy prediction."
+        ),
+    ).tag(config=True)
+
+    ang_res_parameterization = traits.ComponentName(
```
maxnoe, line 146: angular_resolution

src/ctapipe/tools/make_irf.py
```diff
@@ -0,0 +1,615 @@
...
+    ).tag(config=True)
+
+    sens_parameterization = traits.ComponentName(
+        SensitivityMakerBase,
+        default_value="Sensitivity2dMaker",
+        help="The parameterization of the point source sensitivity benchmark.",
+    ).tag(config=True)
+
+    full_enclosure = Bool(
+        False,
```
maxnoe, line 159: I think the option should be `point_like` with default `False` so that the default are non-point-like / full-enclosure IRFs.

src/ctapipe/tools/make_irf.py
```diff
@@ -0,0 +1,615 @@
...
+        + classes_with_traits(EffectiveAreaMakerBase)
+        + classes_with_traits(EnergyMigrationMakerBase)
+        + classes_with_traits(PsfMakerBase)
+        + classes_with_traits(AngularResolutionMakerBase)
+        + classes_with_traits(EnergyBiasResolutionMakerBase)
+        + classes_with_traits(SensitivityMakerBase)
+    )
+
+    def setup(self):
+        self.opt_result = OptimizationResultStore().read(self.cuts_file)
```
maxnoe, line 210: This is a clear sign that the comment I made above is correct: `OptimizationResultStore` should not be a class.

Its `read` should either be a function or a method on the `OptimizationResult` itself.

src/ctapipe/tools/make_irf.py
```diff
@@ -0,0 +1,615 @@
...
+                    EventsLoader(
+                        parent=self,
+                        kind="electrons",
+                        file=self.electron_file,
+                        target_spectrum=self.electron_target_spectrum,
+                    )
+                )
+            if len(self.particles) == 1:
+                raise RuntimeError(
+                    "At least one electron or proton file required when specifying `do_background`."
```
maxnoe, line 251: I'd require the proton file actually. Only using electrons for background doesn't make sense.

src/ctapipe/tools/make_irf.py
```diff
@@ -0,0 +1,615 @@
...
+        )
+        self.aeff = EffectiveAreaMakerBase.from_name(
+            self.aeff_parameterization, parent=self
+        )
+
+        if self.full_enclosure:
+            self.psf = PsfMakerBase.from_name(self.psf_parameterization, parent=self)
+
+        if self.do_benchmarks:
+            self.b_output = self.output_path.with_name(
```
maxnoe, line 272: please don't automatically create output files with names not under user control.

A good solution would be to replace the boolean `do_bechmarks` with an optional `benchmarks_output` Path and if it is given, compute the benchmarks and put them there.

src/ctapipe/tools/make_irf.py
```diff
@@ -0,0 +1,615 @@
...
+                "Keeping %d signal, %d proton events, and %d electron events"
+                % (
+                    sum(reduced_events["gammas"]["selected"]),
+                    sum(reduced_events["protons"]["selected_gh"]),
+                    sum(reduced_events["electrons"]["selected_gh"]),
+                )
+            )
+        else:
+            self.log.debug(
+                "Keeping %d signal events" % (sum(reduced_events["gammas"]["selected"]))
```
maxnoe, line 350: np.count_nonzero instead of sum

also, I think this is more interesting than "debug"

src/ctapipe/tools/make_irf.py
```diff
@@ -0,0 +1,615 @@
...
+            )
+        return reduced_events
+
+    def _stack_background(self, reduced_events):
```
maxnoe, line 354: this function seems needlessly complex.

How about:

```
if self.proton_file and self.electron_file:
    return vstack([reduced_events["protons"], reduced_events["electrons"])

if self.proton_file:
    return reduced_events["proton"]

return reduced_events["electron"]
```

but the same comment as above also applies here: i think we shouldn't support only running with electrons as background.

src/ctapipe/tools/make_irf.py
```diff
@@ -0,0 +1,615 @@
...
+                        ],
+                    )
+                )
+                sel.epp = EventPreProcessor(
+                    parent=sel,
+                    quality_criteria=self.opt_result.precuts.quality_criteria,
+                )
+
+            if sel.epp.gammaness_classifier != self.opt_result.gh_cuts.meta["CLFNAME"]:
+                self.log.warning(
```
maxnoe, line 474: why is this a warning and not an error?

src/ctapipe/tools/make_irf.py
```diff
@@ -0,0 +1,615 @@
...
+            reduced_events[f"{sel.kind}_meta"] = meta
+            self.log.debug(
+                "Loaded %d %s events" % (reduced_events[f"{sel.kind}_count"], sel.kind)
+            )
+            if sel.kind == "gammas":
+                self.signal_is_point_like = (
+                    meta["sim_info"].viewcone_max - meta["sim_info"].viewcone_min
+                ).value == 0
+
+        if self.signal_is_point_like:
```
maxnoe, line 510: I think this should also be a plain error: the configuration should match the passed data.

src/ctapipe/tools/optimize_event_selection.py
```diff
@@ -0,0 +1,216 @@
...
+    description = "Tool to create optimized cuts for IRF generation"
+
+    gamma_file = traits.Path(
+        default_value=None, directory_ok=False, help="Gamma input filename and path"
+    ).tag(config=True)
+
+    gamma_sim_spectrum = traits.UseEnum(
+        Spectra,
+        default_value=Spectra.CRAB_HEGRA,
+        help="Name of the pyirf spectra used for the simulated gamma spectrum",
```
maxnoe, line 23: This is not the spectrum "used" it's the spectrum that is going to be used for weighting, right?

maxnoe (changes requested):

src/ctapipe/irf/benchmarks.py
```diff
@@ -0,0 +1,274 @@
...
+        )
+
+        for i in range(len(self.fov_offset_bins) - 1):
+            ang_res = angular_resolution(
+                events=events[fov_bin_idx == i],
+                energy_bins=e_bins,
+                energy_type=energy_type,
+            )
+            result["N_EVENTS"][:, i, :] = ang_res["n_events"]
+            result["ANG_RES"][:, i, :] = ang_res["angular_resolution"]
```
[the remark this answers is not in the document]
LukasBeiske: Atm, I don't think this is supported by pyirf, but it should be a simple change. I'll look into it.

src/ctapipe/irf/binning.py
```diff
@@ -0,0 +1,174 @@
...
+    unit = e_min.unit
+    log_lower = np.log10(e_min.to_value(unit))
+    log_upper = np.log10(e_max.to_value(unit))
+
+    n_bins = int(np.ceil((log_upper - log_lower) * n_bins_per_decade))
+
+    return u.Quantity(np.logspace(log_lower, log_upper, n_bins + 1), unit, copy=False)
+
+
+class ResultValidRange:
```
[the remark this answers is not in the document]
LukasBeiske: I think, the main reason for this was to support the `.{min,max}` syntax and make the code a bit more readable like this.

src/ctapipe/irf/irfs.py
```diff
@@ -0,0 +1,316 @@
...
+    ).tag(config=True)
+
+    energy_migration_n_bins = Integer(
+        help="Number of bins in log scale for Energy Migration matrix",
+        default_value=30,
+    ).tag(config=True)
+
+    def __init__(self, parent=None, **kwargs):
+        super().__init__(parent=parent, **kwargs)
+        self.migration_bins = np.linspace(
```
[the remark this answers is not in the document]
LukasBeiske: I'll fix it. And I also think renaming it to energy dispersion would be good to avoid confusion with the naming in pyirf (e.g. the `energy_dispersion_to_migration` function).

src/ctapipe/irf/irfs.py
```diff
@@ -0,0 +1,316 @@
...
+    """Base class for calculating the effective area."""
+
+    def __init__(self, parent=None, **kwargs):
+        super().__init__(parent=parent, **kwargs)
+
+    @abstractmethod
+    def make_aeff_hdu(
+        self,
+        events: QTable,
+        point_like: bool,
```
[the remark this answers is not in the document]
LukasBeiske: Atm this is configured once for all irf components when running the `make-irf` tool via the `--full-enclosure` flag. Is there a use-case where some irf components are calculated point-like and others not?

src/ctapipe/irf/optimize.py
LukasBeiske: I think, this should cover cases where this component is used on it's own outside the tools.

src/ctapipe/irf/optimize.py
LukasBeiske: If I remember correctly, the idea was to separate the definition of the output product of the optimization tool from the code generating it.
@Tobychev Do you think it is worth keeping it this way?

src/ctapipe/irf/select.py
LukasBeiske: Since multiple files are loaded when running the tools, these configuration options would have to be overwritten within the tool. This is not intended is it?

src/ctapipe/tools/make_irf.py
LukasBeiske: It should be an error, I'll fix it :+1:

src/ctapipe/irf/irfs.py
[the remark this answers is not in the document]
maxnoe: No, this variable is called migration also in pyirf, e.g. here:
https://pyirf.readthedocs.io/en/latest/api/pyirf.irf.energy_dispersion.html#pyirf.irf.energy_dispersion

(migration bins)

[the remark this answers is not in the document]
LukasBeiske: Ok, but I meant renaming `EnergyMigrationMaker` into `EnergyDispersionMaker` since it returns the dispersion matrix (true energy vs (true - reco) / true) and not the migration matrix (true energy vs reco energy).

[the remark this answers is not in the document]
maxnoe: Ah, yes! I thought you meant renaming the variable

src/ctapipe/irf/select.py
LukasBeiske: I would be in favor of making the existing frames compatible with GADF, but probably in a separate PR.

src/ctapipe/irf/select.py
LukasBeiske: `obs_id` and `event_id` are the only columns not using float64, or am I forgetting something?

src/ctapipe/tools/make_irf.py
LukasBeiske: I just removed the "IRF" here.

src/ctapipe/tools/make_irf.py
LukasBeiske: When requiring the proton file, this turned into only 4 lines, so I removed this helper function and did it inline instead.

src/ctapipe/tools/optimize_event_selection.py
LukasBeiske: Yes, I fixed the help strings in both tools accordingly.

src/ctapipe/irf/select.py
maxnoe: The _valid are bool

Tobychev: Yeah, good call. Added explicit dimensionless units to the gh score and weight too.

src/ctapipe/tools/make_irf.py
LukasBeiske: I think I changed that already twice in the past. I would agree that doing it this way would make the most sense, if only to be consistent with pyirf. @Tobychev Are you fine with changing it like this?

LukasBeiske:
The tests are passing after https://github.com/cta-observatory/pyirf/pull/289.

src/ctapipe/irf/benchmarks.py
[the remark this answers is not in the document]
LukasBeiske: This is possible after https://github.com/cta-observatory/pyirf/pull/290.

maxnoe:
In #2641, we made some of the ctapipe dependencies optional. I think this should also be the case for the code here (`pyirf`).

src/ctapipe/tools/make_irf.py
kosack: It's not used for analysis in point-like IRFs, but it is used for visualization. E.g.if you produce a point-like excess map, you need to overplot the PSF to show how the map is correlated. Therefore it is always useful to include the PSF in even the point-like IRFs.  It's also useful just for cross-checking that your RAD_MAX looks ok.

src/ctapipe/irf/spectra.py
```diff
@@ -0,0 +1,26 @@
+"""Definition of spectra to be used to calculate event weights for irf computation"""
+
+from enum import Enum
+
+import astropy.units as u
+from pyirf.spectral import CRAB_HEGRA, IRFDOC_ELECTRON_SPECTRUM, IRFDOC_PROTON_SPECTRUM
```
kosack, line 6: wasn't this moved to a new package? Or is that just handled inside pyirf?

src/ctapipe/tools/optimize_event_selection.py
```diff
@@ -0,0 +1,233 @@
...
+        )
+        self.particles = [
+            EventLoader(
+                parent=self,
+                kind="gammas",
+                file=self.gamma_file,
+                target_spectrum=self.gamma_target_spectrum,
+            )
+        ]
+        if self.optimization_algorithm != "PercentileCuts":
```
kosack, line 122: Might be safer to use `isinstance()` here and elsewhere instead of a string compare to the class name, so that if we ever re-factor or rename classes, it gets updated correctly. But that would mean importing those sub-classes or the module, so not clear if that makes it messier.

src/ctapipe/tools/tests/test_compute_irf.py
```diff
@@ -0,0 +1,284 @@
...
+        assert isinstance(hdul["EFFECTIVE AREA"], fits.BinTableHDU)
+        if point_like:
+            assert isinstance(hdul["RAD_MAX"], fits.BinTableHDU)
+        else:
+            assert isinstance(hdul["PSF"], fits.BinTableHDU)
+
+        if include_bkg:
+            assert isinstance(hdul["BACKGROUND"], fits.BinTableHDU)
+
+    os.remove(output_path)  # Delete output file
```
kosack, line 86: use `output_path.unlink()` instead, so no need to `import os`.

kosack (changes requested):
I did a more detailed by-eye code review and have a few suggested changes. I will now do some more detailed testing by running the code, so may submit a second review shortly.

LukasBeiske:
> I did a more detailed by-eye code review and have a few suggested changes. I will now do some more detailed testing by running the code, so may submit a second review shortly.

Thanks! I can also add an example config file (maybe containing all config options and there default values?) at `ctapipe.resources`, if you think that would make sense.

src/ctapipe/irf/irfs.py
```diff
@@ -0,0 +1,336 @@
...
+            bin_func = np.linspace
+        self.migration_bins = bin_func(
+            self.energy_migration_min,
+            self.energy_migration_max,
+            self.energy_migration_n_bins + 1,
+        )
+
+    @abstractmethod
+    def make_edisp_hdu(
+        self, events: QTable, point_like: bool, extname: str = "ENERGY MIGRATION"
```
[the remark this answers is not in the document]
LukasBeiske: I will change it to `spatial_selection_applied`. I guess we should also change it accordingly in pyirf then, right?

src/ctapipe/irf/benchmarks.py
```diff
@@ -0,0 +1,293 @@
...
+
+    def __init__(self, parent=None, **kwargs):
+        super().__init__(parent=parent, **kwargs)
+
+    @abstractmethod
+    def make_sensitivity_hdu(
+        self,
+        signal_events: QTable,
+        background_events: QTable,
+        theta_cut: QTable,
```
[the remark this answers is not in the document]
LukasBeiske: Then we should probably not use "theta cut" anywhere anymore, right? So also rename it in the cut calculation components etc.

src/ctapipe/irf/benchmarks.py
```diff
@@ -0,0 +1,293 @@
...
+    mat_shape = (len(fov_bins) - 1, len(e_bins) - 1)
+    return result, fov_bin_index, mat_shape
+
+
+class EnergyBiasResolutionMakerBase(DefaultTrueEnergyBins):
+    """
+    Base class for calculating the bias and resolution of the energy prediciton.
+    """
+
+    def __init__(self, parent=None, **kwargs):
```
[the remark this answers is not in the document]
LukasBeiske: The traitlets docs say something similar. I'll fix it :+1:

src/ctapipe/irf/benchmarks.py
```diff
@@ -0,0 +1,293 @@
...
+        super().__init__(parent=parent, **kwargs)
+
+    @abstractmethod
+    def make_bias_resolution_hdu(
```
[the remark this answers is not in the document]
LukasBeiske: I'll also change `self.psf` in `compute-irf` to `self.psf_maker` and the corresponding config traits from `self.psf_maker` to `self.psf_maker_name` to make it a bit more clear what these attributes are.

src/ctapipe/irf/spectra.py
LukasBeiske: I don't see any recent change regarding this in pyirf. Am I missing something?

src/ctapipe/irf/preprocessing.py
```diff
@@ -0,0 +1,321 @@
...
+from ..coordinates import NominalFrame
+from ..core import Component, QualityQuery
+from ..core.traits import List, Tuple, Unicode
+from ..io import TableLoader
+from .spectra import SPECTRA, Spectra
+
+__all__ = ["EventLoader", "EventPreProcessor"]
+
+
+class EventPreProcessor(QualityQuery):
```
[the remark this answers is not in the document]
LukasBeiske: I'll fix this tomorrow

src/ctapipe/irf/irfs.py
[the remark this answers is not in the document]
kosack: yes, but that can be something we do later. Maybe just open an issue there, so it doesn't block this PR.  It's only really important for the API here

src/ctapipe/irf/spectra.py
kosack: maybe it's just a proposal. I seem to remember somebody making a common package for the spectra, but I can't recall who or where it is! (maybe @maxnoe  remembers), but in any case, it's not a problem here if it's still in PyIRF

kosack:
> Thanks! I can also add an example config file (maybe containing all config options and their default values?) at `ctapipe.resources`, if you think that would make sense.

Sure, and make sure it gets outputted by `ctapipe-quickstart`. That would make it a bit friendlier. .Could also do that in a future PR if you prefer.

src/ctapipe/irf/spectra.py
morcuended: Is it https://gitlab.cta-observatory.org/cta-computing/dpps/cosmic-ray-spectra?

src/ctapipe/irf/benchmarks.py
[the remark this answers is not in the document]
kosack: I also had trouble (also in pyirf) understanding what this option was. It sounds like a single value (`theta_cut=0.02`), but it is in fact a table per energy bin, so renaming will help with that. 

not for this PR, but at some point it would be nice to find a way to have "type-checked" astropy Tables, i.e. be able to define a subclass that ensures that certain columns exist and have the expected units, so that we can type check properly. That's a general problem for _PyIRF_ and _gammapy_ as well.  But not something easily solved... I.e. I'd prefer to have the signature be `spatial_selection_table: SpatialSelectionTable`

src/ctapipe/irf/irfs.py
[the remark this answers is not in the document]
LukasBeiske: See https://github.com/cta-observatory/pyirf/issues/297

LukasBeiske:
The default values of many config options should probably be changed, but this can also be done in a follow-up PR.

kosack (dismissed):

src/ctapipe/tools/compute_irf.py
```diff
@@ -0,0 +1,637 @@
...
+        Initialize components from config and load g/h (and theta) cuts.
+        """
+        self.opt_result = OptimizationResult.read(self.cuts_file)
+        if (
+            self.spatial_selection_applied
+            and self.opt_result.spatial_selection_table is None
+        ):
+            raise ToolConfigurationError(
+                f"{self.cuts_file} does not contain any direction cut, "
+                "but --spatial-selection-applied was given."
```
maxnoe, line 241: I am really not sure if it is worth changing the terminology here... we will constantly tell people "We call point-like IRFs now "spatial-selection-applied IRFs". And it won't ever catch on because it is really unweildy compared to just saying "point-like".

cc @kosack

src/ctapipe/tools/compute_irf.py
```diff
@@ -0,0 +1,637 @@
...
+            )
+
+        check_e_bins = partial(
+            check_bins_in_range,
+            valid_range=self.opt_result.valid_energy,
+            raise_error=self.range_check_error,
+        )
+        self.particles = [
```
maxnoe, line 249: call this `event_loaders`? I find `particles` a surprising name for a list of IO objects.

Maybe make it a `dict[particle] -> EventLoader`?

src/ctapipe/tools/compute_irf.py
```diff
@@ -0,0 +1,637 @@
...
+                        self.opt_result.clf_prefix,
+                    )
+                )
+
+            if (
+                sel.epp.quality_query.quality_criteria
+                != self.opt_result.precuts.quality_criteria
+            ):
+                self.log.warning(
+                    "Precuts are different from precuts used for calculating "
```
maxnoe, line 485: Why are we calling this now "precuts" when we chose the term "QualityQuery" precisely to avoid the term "precuts"...

src/ctapipe/tools/compute_irf.py
```diff
@@ -0,0 +1,637 @@
...
+                )
+                sel.epp.quality_query = EventQualityQuery(
+                    parent=sel,
+                    quality_criteria=self.opt_result.precuts.quality_criteria,
+                )
+
+            self.log.debug(
+                "%s Precuts: %s" % (sel.kind, sel.epp.quality_query.quality_criteria)
+            )
+            evs, cnt, meta = sel.load_preselected_events(self.chunk_size, self.obs_time)
```
maxnoe, line 505: General comment concerning the code base is to not use these unnecessary abbreviations

src/ctapipe/tools/optimize_event_selection.py
```diff
@@ -0,0 +1,234 @@
...
+                self.log.warning("Optimizing cuts without electron file.")
+
+    def start(self):
+        """
+        Load events and optimize g/h (and theta) cuts.
+        """
+        reduced_events = dict()
+        for sel in self.particles:
+            evs, cnt, meta = sel.load_preselected_events(self.chunk_size, self.obs_time)
+            if isinstance(self.optimizer, PointSourceSensitivityOptimizer):
```
maxnoe, line 158: It seems kind of an anti-pattern to define an abstract API, but then to do things specifically for one particular implementation. This essentially means that the interface is missing parts.

E.g. here there could be a property `self.optimizer.needs_event_weights`  which would then be a general property any `Optimizer` could set

src/ctapipe/tools/optimize_event_selection.py
```diff
@@ -0,0 +1,234 @@
...
+
+            reduced_events[sel.kind] = evs
+            reduced_events[f"{sel.kind}_count"] = cnt
+            if sel.kind == "gammas":
+                self.sim_info = meta["sim_info"]
+                self.gamma_spectrum = meta["spectrum"]
+
+        self.signal_events = reduced_events["gammas"]
+
+        if isinstance(self.optimizer, PercentileCuts):
```
maxnoe, line 176: Same comment here. Could be implemented via a `requires_background` property or similar.

src/ctapipe/irf/binning.py
```diff
@@ -0,0 +1,187 @@
...
+                    f"Valid range for {source} is {min_val} to {max_val}, got {bins}"
+                )
+            else:
+                logger.warning(
+                    f"Valid range for {source} is {min_val} to {max_val}, got {bins}",
+                )
+
+
+@u.quantity_input(e_min=u.TeV, e_max=u.TeV)
+def make_bins_per_decade(e_min, e_max, n_bins_per_decade=5):
```
maxnoe, line 63: What's the reason this is here and we are not just using:
https://github.com/cta-observatory/pyirf/blob/5e8b8c9dd54d632fcb759df90a35d4029119085a/pyirf/binning.py#L100 ?

src/ctapipe/tools/compute_irf.py
LukasBeiske: Good point, fixed it.

src/ctapipe/tools/compute_irf.py
LukasBeiske: I removed the worst offenders I noticed in here, for now

src/ctapipe/tools/optimize_event_selection.py
LukasBeiske: I combined both these properties into one

src/ctapipe/irf/benchmarks.py
```diff
@@ -0,0 +1,306 @@
...
+        return BinTableHDU(result, name=extname)
+
+
+class AngularResolutionMakerBase(DefaultTrueEnergyBins, DefaultRecoEnergyBins):
+    """
+    Base class for calculating the angular resolution.
+    """
+
+    # Use reconstructed energy by default for the sake of current pipeline comparisons
+    use_true_energy = Bool(
```
maxnoe, line 111: I'd invert the option. `use_reco_energy` with default false. The default should be the "good" option.

src/ctapipe/irf/binning.py
maxnoe: ah, I see. This is a slightly different version that uses "roughly" n_bins_per_decade and uses exact lower, upper limits.

The pyirf version should be used for sensitivity and uses __exactly__ n_bins_per_decade.

Essentially this is linspace vs. arange

maxnoe: This could be a bit clearer in the docstring

src/ctapipe/irf/binning.py
```diff
@@ -0,0 +1,187 @@
...
+    """
+    Create energy bins with at least ``n_bins_per_decade`` bins per decade.
```
maxnoe, line 65: You could add "if you want to get exactly ``n_bins_per_decade`` use `pyirf.binning.create_bins_per_decade`

src/ctapipe/irf/benchmarks.py
```diff
@@ -0,0 +1,306 @@
...
+        -------
+        BinTableHDU
+        """
+
+
+class Sensitivity2dMaker(SensitivityMakerBase, DefaultFoVOffsetBins):
+    """
+    Calculates the point source sensitivity in bins of reconstructed energy
+    and fov offset.
+    """
```
maxnoe, line 240: It's not obvious from the code, but does this class use the create_bins_per_decade function create here or the one from pyirf?

To comply with the cta requirements, we should use the one that produces _exactly_ five bins per decade.

src/ctapipe/irf/irfs.py
```diff
@@ -0,0 +1,357 @@
...
+            true_energy_bins=self.true_energy_bins,
+            fov_offset_bins=self.fov_offset_bins,
+            source_offset_bins=self.source_offset_bins,
+            extname=extname,
+        )
+        # We also calculate a psf for IRFs including a spatial selection
+        # ("point-like IRF") to enable RAD_MAX cross-checks downstream.
+        # In that case, we have to change the header accordingly.
+        if spatial_selection_applied:
+            hdu.header["HDUCLAS3"] = "POINT-LIKE"
```
maxnoe, line 355: point-like PSF does not make sense. Even if you include it in the output, which I still don't think is right, the PSF itself is not "point-like" if you didn't apply the theta cut to compute it, which you shouldn't.

maxnoe (approved):

LukasBeiske (approved):

kosack (approved):

PR #2483, opened 2023-12-14
Write out cross-validation results immediately

### LINKED ISSUE #2480: Write cross-validation results immediately after cross validation
Currently, the `CrossValidator` keeps the results of the cross-validation, including the full result table in memory and writes it in `finish` method of the tools.

We should write the results immediately in the cross validation loop using `write_table(..., append=True)` to avoid keeping this information in memory unnecessarily.

Write cross validation results out immediately after each step, not in `finish`.

LukasBeiske:
Fixes #2480

### DIFF
--- a/ctapipe/reco/sklearn.py
+++ b/ctapipe/reco/sklearn.py
@@ -967,26 +967,19 @@ class CrossValidator(Component):
 
             cv_result, metrics = self.cross_validate(telescope_type, train, test)
             if self.output_path:
-                results = hstack(
-                    [
-                        Table(
-                            data={
-                                "cv_fold": np.full(
-                                    len(cv_result), fold, dtype=np.uint8
-                                ),
-                                "tel_type": [str(telescope_type)] * len(cv_result),
-                                "true_energy": test["true_energy"],
-                                "true_impact_distance": test["true_impact_distance"],
-                            },
-                            descriptions={
-                                "cv_fold": "Cross validation iteration",
-                                "tel_type": "Telescope type",
-                            },
-                        ),
-                        cv_result,
-                    ],
-                    join_type="exact",
+                results = Table(
+                    data={
+                        "cv_fold": np.full(len(cv_result), fold, dtype=np.uint8),
+                        "tel_type": [str(telescope_type)] * len(cv_result),
+                        "true_energy": test["true_energy"],
+                        "true_impact_distance": test["true_impact_distance"],
+                    },
+                    descriptions={
+                        "cv_fold": "Cross validation iteration",
+                        "tel_type": "Telescope type",
+                    },
                 )
+                results = hstack([results, cv_result], join_type="exact")
                 write_table(
                     results,
                     self.h5file,
--- a/ctapipe/reco/sklearn.py
+++ b/ctapipe/reco/sklearn.py
@@ -10,11 +10,12 @@ import astropy.units as u
 import joblib
 import numpy as np
 from astropy.coordinates import AltAz
-from astropy.table import QTable, Table, vstack
+from astropy.table import QTable, Table
 from astropy.utils.decorators import lazyproperty
 from sklearn.metrics import accuracy_score, r2_score, roc_auc_score
 from sklearn.model_selection import KFold, StratifiedKFold
 from sklearn.utils import all_estimators
+from tables import open_file
 from tqdm import tqdm
 from traitlets import TraitError, observe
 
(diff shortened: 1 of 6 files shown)


ctapipe/reco/sklearn.py
```diff
@@ -878,9 +878,8 @@ class CrossValidator(Component):
         default_value=1337, help="Random seed for splitting the training data."
     ).tag(config=True)
 
-    def __init__(self, model_component, **kwargs):
+    def __init__(self, model_component, overwrite=False, **kwargs):
```
maxnoe, line 881: You don't use overwrite

ctapipe/reco/sklearn.py
```diff
@@ -965,7 +967,14 @@ def __call__(self, telescope_type, table):
...
                     cv_values.std(),
                 )
-        self.cv_predictions[telescope_type] = vstack(predictions)
+
+        if self.output_path:
+            write_table(
+                vstack(predictions),
+                self.output_path,
+                f"/cv_predictions_{telescope_type}",
+                overwrite=True,
```
maxnoe, line 976: And here you set it to true immediately

ctapipe/reco/sklearn.py
LukasBeiske: I forgot to remove that when removing the redundant test 👍

ctapipe/reco/sklearn.py
maxnoe: You can move the writing into the loop (further reducing memory) by adding `append=True` here after overwriting the file in `__init__`

maxnoe (changes requested):
Good start, but you should

a) Check overwrite conditions in `__init__`: if output file exists and not overwrite, raise an error, if output file exists and overwrite=true truncate it (tables.open_file(path, mode='w'). 

b) Move the writing into the cv loop using `write_table(..., append=True)` to further reduce memory usage

c) Add writing of metadata to the output file

LukasBeiske:
> Good start, but you should
> 
> a) Check overwrite conditions in `__init__`: if output file exists and not overwrite, raise an error, if output file exists and overwrite=true truncate it (tables.open_file(path, mode='w').
> 
> b) Move the writing into the cv loop using `write_table(..., append=True)` to further reduce memory usage
> 
> c) Add writing of metadata to the output file

a) The error if file exists and not overwrite is already raised by the train tools via `Tool.check_output`.

a)/b) Using `write_table(..., append=True, mode='w')` does not work, does it? Isn't this basically the same as `write_table(..., append=True, overwrite=True)`?

c) Is there any other metadata requiered besides some column descriptions?

maxnoe:
> a) The error if file exists and not overwrite is already raised by the train tools via Tool.check_output.

Also for the cv output file and not just the model?

maxnoe:
> a)/b) Using write_table(..., append=True, mode='w') does not work, does it? Isn't this basically the same as write_table(..., append=True, overwrite=True)?

You need to truncate the file once in `__init__` and then only use `append=True`

LukasBeiske:
> > a) The error if file exists and not overwrite is already raised by the train tools via Tool.check_output.
> 
> Also for the cv output file and not just the model?

Yes, both output paths are passed and `CrossValidator.ouput_path` gets ignored if it is None

ctapipe/reco/sklearn.py
```diff
@@ -898,6 +905,29 @@ def __init__(self, model_component, **kwargs):
                 "Unsupported Model of type %s supplied", self.model_component
             )
 
+        if self.output_path:
```
Tobychev, line 908: You can combine these two into `if self.output_path and self.output_path.exists():` without any danger, if `output_path` is none the second bit is not executed.

LukasBeiske: But if I do it like this, the lines below the checks for the correct configuration regarding `overwrite` do not get executed if the output path doesn't already exist. So I would have to put those in a separate if statement which would result in the same number of if statements as right now.

ctapipe/reco/sklearn.py
```diff
@@ -937,21 +965,34 @@ def __call__(self, telescope_type, table):
...
-                    {
-                        "cv_fold": np.full(len(truth), fold, dtype=np.uint8),
-                        "tel_type": [str(telescope_type)] * len(truth),
-                        "truth": truth,
-                        "true_energy": test["true_energy"],
-                        "true_impact_distance": test["true_impact_distance"],
-                        **cv_prediction,
-                    }
+            cv_result, metrics = self.cross_validate(telescope_type, train, test)
+            if self.output_path:
```
maxnoe, line 969: This is very unreadable... introducing some intermediate variables might be better, i.e:
```
results = Table(
     ...
)
results = hstack([results, cv_predictions], joint_type='exact')
```

Tobychev (approved):

PR #2486, opened 2024-01-08
Test on python 3.12

maxnoe:


maxnoe:
wohoo :fireworks:

kosack (approved):

.github/workflows/ci.yml
```diff
@@ -43,19 +43,23 @@ jobs:
 
           - os: ubuntu-latest
             python-version: "3.10"
```
Tobychev, line 45: Is there a reason why phyton 3.10 only uses pip while 3.9, 3.11, and 3.12 also tries mamba?

maxnoe: No particular reason, we at some point in the past decided not to test on the full product matrix of `[python versions] x [pip, mamba]` but to only test some combinations that cover all python versions and both installation methods.

maxnoe: To limit the number of parallel builds that run in the CI

Tobychev (approved):

kosack (approved):

PR #2497, opened 2024-01-19
Add functionality for array-event-wise aggregation of dl1 image parameters

LukasBeiske:
- This adds a new `Component` and a `Tool` for aggregating the telescope-wise dl1 image parameters for each array event.
- `BaseStatisticsContainer` is introduced which does not contain the higher order moments present in `StatisticsContainer`
- The helper functions for vectorizing numpy calculations in `ctapipe.reco.stereo_combination` are refactored and expanded in a new module called `ctapipe.vectorization`  

TODO:
- [ ] Add/ update unit tests

### DIFF
--- a/ctapipe/containers.py
+++ b/ctapipe/containers.py
@@ -410,13 +410,16 @@ class MorphologyContainer(Container):
     n_large_islands = Field(-1, "Number of > 50 pixel islands")
 
 
-class StatisticsContainer(Container):
+class BaseStatisticsContainer(Container):
     """Store descriptive statistics"""
 
     max = Field(np.float32(nan), "value of pixel with maximum intensity")
     min = Field(np.float32(nan), "value of pixel with minimum intensity")
     mean = Field(np.float32(nan), "mean intensity")
     std = Field(np.float32(nan), "standard deviation of intensity")
+
+
+class StatisticsContainer(BaseStatisticsContainer):
     skewness = Field(nan, "skewness of intensity")
     kurtosis = Field(nan, "kurtosis of intensity")
 
@@ -521,6 +524,10 @@ class DL1Container(Container):
     tel = Field(
         default_factory=partial(Map, DL1CameraContainer),
         description="map of tel_id to DL1CameraContainer",
+    )
+    aggregate = Field(
+        default_factory=partial(Map, BaseStatisticsContainer),
+        description="map of image parameter to aggregation statistics",
     )
 
 
--- a/ctapipe/image/__init__.py
+++ b/ctapipe/image/__init__.py
@@ -60,7 +60,7 @@ from .pixel_likelihood import (
     neg_log_likelihood_numeric,
 )
 from .reducer import DataVolumeReducer, NullDataVolumeReducer, TailCutsDataVolumeReducer
-from .statistics import descriptive_statistics
+from .statistics import FeatureAggregator, descriptive_statistics
 from .timing import timing_parameters
 
 __all__ = [
@@ -119,4 +119,5 @@ __all__ = [
     "TailCutsDataVolumeReducer",
     "InvalidPixelHandler",
     "NeighborAverage",
+    "FeatureAggregator",
 ]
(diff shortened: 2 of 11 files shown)


kosack:
Can you explain better what this is and why is needed in the description? What does it mean to "Aggregate DL1 parameters"?

Tobychev:
> What does it mean to "Aggregate DL1 parameters"? and what is being aggregated over (over telescopes in an event, or over time, both?)
It looks like it groups things by trigger, so I guess it is for getting average intensity and width and such, maybe?

kosack:
I don't quite understand how this will be used, but perhaps I need to see an example.  I imagine when you say "mean", you really mean "weighted mean", weighted by the impact parameter and intensity, so basically the same as the "mean reduced scaled {width, length}" used in standard non-machine-learning gamma-hadron separation methods in HESS and VERITAS (but not MARS, if I understand correctly)?  If that's true, we've basically come full-circle and re-invented those.

LukasBeiske:
> which can then be merged in a weighted average like we do for energy and gammaness. So are we now undoing that design decision?

AFAIK these averaged features are only to be used for the model estimating the error of the direction reconstruction. While the event type based on the background contamination will be using the averaged gammaness scores predicted by the telescope models. 
Therefore, it would also be possible to only compute these averaged features on-the-fly when training and applying this direction reconstruction error estimator, similar to the existing `FeatureGenerator`.

The separate tool to calculate and write these averaged features into files is only meant to keep this separate from the implementation of the direction reconstruction error estimator itself. If there really is no other use for these averaged features, the on-the-fly solution might be better.

maxnoe:
@kosack this is part of the implementation of the event types based approach developed by @TarekHC and @JBernete.

It works by training a machine learning regressor on feature table for subarray events. Part of that feature table are multiplicities, the reconstructed values themselves but also these aggregated image features over the telescopes.

kosack:
> AFAIK these averaged features are only to be used as input for the model estimating the error of the direction reconstruction. While the event type based on the background contamination will be using the averaged gammaness scores predicted by the telescope models.

Yes, that I understand, the question is more why is being done using mean-scaled array parameters rather than using per-telescope estimators that are then combined, which is how we do the other machine learning methods.  At least then all ML methods would be symmetric and there would be no need for this extra complexity. 

In fact, I really expected to see something that leverages how the reconstruction algorithm actually works, for example computing some aggregate parameters not on the telescopes, but on the telescope **pairs** used in reconstruction.  For example, computing the mean/std/min/max of the _delta-phi angle_ between image pairs.  Wouldn't that have much more power to estimate the uncertainty?  To first order, the reconstruction error for a 2-telescope pair is just related to this angle: small is poorly reconstructed, and large is better reconstructed, which is why we weight by cos(delta_phi)--implicitly via the cross-product--in the algorithm itself,  which leads to lower reconstruction uncertainty.

src/ctapipe/image/statistics.py
```diff
@@ -92,3 +108,109 @@ def descriptive_statistics(
...
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+        self.feature_generator = FeatureGenerator(parent=self)
+        self.quality_query = QualityQuery(parent=self)
+
+    def __call__(self, event: ArrayEventContainer) -> None:
+        """Fill event container with aggregated image parameters."""
+        table = None
+        for tel_id in event.trigger.tels_with_trigger:
+            t = collect_features(event, tel_id)
```
kosack, line 134: It's possible I missed it, but you are collecting features from all telescope in the event, but shouldn't it include only those used in reconstruction? In other words, instead of `tels_with_trigger`, you should have something like `event.dl2.stereo.geometry.telescopes`

LukasBeiske: Right now this is done via a quality query in `aggregate_table`, which should use the same quality criteria as the reconstruction for which the error will be estimated.  But if this is integrated into the train/apply tool for the error estimator, doing it this way would be better, I agree.

LukasBeiske:
codecov fails because of the new numba jitted code and the already existing numba functions in `image.statistics`

LukasBeiske:
flake8 fails now because of lines being to long which I did not write in this PR. Is this something I should fix here, or would it be better to fix this everywhere in a separate PR?

PR #2499, opened 2024-01-22
Fix docstring of ImPACT predict method

HealthyPear:
Just a small fix I found while roaming around the code...

The name of the geometry container should be`ReconstructedGeometryContainer` and not `ReconstructedShowerContainer`.

### DIFF
--- a/ctapipe/reco/impact.py
+++ b/ctapipe/reco/impact.py
@@ -671,14 +671,14 @@ class ImPACTReconstructor(Component):
 
         Parameters
         ----------
-        shower_seed: ReconstructedShowerContainer
+        shower_seed: ReconstructedGeometryContainer
             Seed shower geometry to be used in the fit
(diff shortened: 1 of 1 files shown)


HealthyPear:
The linting failing job is caused by the fact that I used the onine `github.dev` web-based editor.

I can of course squash a commit from my local installation to fix stuff like isort changes, but what if we use ruff with the `--fix` option? Applying formatting or basic linting fixes manually post-commit can be disruptive to the development workflow.

The documentation job is instead failing for a reason which I believe is not related to my changes...

Tobychev:
Could you make the changes directly on #2305 ? That's the "active" ImPACT pr and it has probably already touched these names.

HealthyPear:
> Could you make the changes directly on #2305 ? That's the "active" ImPACT pr and it has probably already touched these names.

sure no problem

PR #2503, opened 2024-01-25
Angular error regressor

JBernete:
Adding tools to reconstruct the angular error at the DL2 level. Which will allow the definition of event types based on the angular reconstruction quality.


kosack:
Can you define "angular error"?

kosack:
This PR also seems to include #2497, maybe it should be parented to that branch?

JBernete:
With angular error I mean the angular difference between the true direction and the reconstructed direction of an event. This is useful to define event types in terms of the quality of the reconstruction.

maxnoe:
> This PR also seems to include https://github.com/cta-observatory/ctapipe/pull/2497, maybe it should be parented to that branch?

This PR needs #2497 to create the input features for the machine learning. I think we should first merge 2497 and then update here, not make 2497 bigger by including this branch there.

kosack:
> With angular error I mean the angular difference between the true direction and the reconstructed direction of an event. This is useful to define event types in terms of the quality of the reconstruction.

I suggest naming it "DirectionReconstructionError"  or something more clear, since AngularError could refer to other angles. Or maybe better yet "DirectionReconstructionUncertainty" to avoid the class being confused with RuntimeErrors...

ctapipe/reco/direction_uncertainty.py
```diff
@@ -62,13 +62,13 @@ def predict_subarray_table(self, table):
...
-        ang_error[valid] = self._model.predict(X)
-        ang_error = u.Quantity(ang_error, self.unit, copy=False)
+        dir_uncert = np.full(n_rows, np.nan)
+        dir_uncert[valid] = self._model.predict(X)
+        dir_uncert = u.Quantity(dir_uncert, self.unit, copy=False)
 
         result = Table(
             {
-                f"{self.reconstructor_prefix}_ang_distance_uncert": ang_error,
+                f"{self.reconstructor_prefix}_direction_uncertainty": dir_uncert,
```
maxnoe, line 71: This column name actually needs to stay the same, as it comes from the definitions in `ctapipe.containers`.

kosack: could also just rename it to `direction_uncertainty`, since we have not really used that column up to now. It's technically a data model change, but again, that column was never filled, so shouldn't cause a problem.

LukasBeiske: The stereo combiner fills this column when combining alt/az predictions by disp. But I don't think this was used for anything up until now, so changing it shouldn't be a problem (and `direction_uncertainty` is a better name imo).

PR #2511, opened 2024-02-16
Change `ImageCleaner` API

### LINKED ISSUE #2015: Change API of ImageCleaner to be `__call__(self, event: ArrayEvent)`
Currently, the `ImageCleaner` has the API of `__call__(self, tel_id, image, peak_time)`.
For more advanced cleaning methods that e.g. require monitoring information about the current NSB / pedestal variation, it is impossible to use this information.

Also switch to `__call__(self, event)` so the cleaning algorithms can access all relevant information.

**Additional context**

E.g. LSTs cleaning is using a factor of the pedestal standard deviation as a lower bound for the cleaning threshold.

Hckjs:
Change API of `ImageCleaner` to be `__call__(self, tel_id: int,  event: ArrayEventContainer)`. Closes #2015.

@aleberti

### DIFF
--- a/src/ctapipe/image/cleaning.py
+++ b/src/ctapipe/image/cleaning.py
@@ -28,6 +28,7 @@ from abc import abstractmethod
 
 import numpy as np
 
+from ..containers import ArrayEventContainer
 from ..core import TelescopeComponent
 from ..core.traits import (
     BoolTelescopeParameter,
@@ -465,9 +466,7 @@ class ImageCleaner(TelescopeComponent):
     """
 
     @abstractmethod
-    def __call__(
-        self, tel_id: int, image: np.ndarray, arrival_times: np.ndarray = None
-    ) -> np.ndarray:
+    def __call__(self, tel_id: int, event: ArrayEventContainer) -> np.ndarray:
         """
         Identify pixels with signal, and reject those with pure noise.
 
@@ -476,10 +475,7 @@ class ImageCleaner(TelescopeComponent):
         tel_id: int
             which telescope id in the subarray is being used (determines
             which cut is used)
-        image : np.ndarray
-            image pixel data corresponding to the camera geometry
-        arrival_times: np.ndarray
-            image of arrival time (not used in this method)
+        event: `ctapipe.containers.ArrayEventContainer`
 
         Returns
         -------
@@ -513,16 +509,14 @@ class TailcutsImageCleaner(ImageCleaner):
         "removed.",
     ).tag(config=True)
 
-    def __call__(
-        self, tel_id: int, image: np.ndarray, arrival_times=None
-    ) -> np.ndarray:
+    def __call__(self, tel_id: int, event: ArrayEventContainer) -> np.ndarray:
         """
         Apply standard picture-boundary cleaning. See `ImageCleaner.__call__()`
         """
 
         return tailcuts_clean(
             self.subarray.tel[tel_id].camera.geometry,
-            image,
+            event.dl1.tel[tel_id].image,
             picture_thresh=self.picture_threshold_pe.tel[tel_id],
             boundary_thresh=self.boundary_threshold_pe.tel[tel_id],
             min_number_picture_neighbors=self.min_picture_neighbors.tel[tel_id],
@@ -535,16 +529,14 @@ class MARSImageCleaner(TailcutsImageCleaner):
     1st-pass MARS-like Image cleaner (See `ctapipe.image.mars_cleaning_1st_pass`)
     """
 
-    def __call__(
-        self, tel_id: int, image: np.ndarray, arrival_times=None
-    ) -> np.ndarray:
+    def __call__(self, tel_id: int, event: ArrayEventContainer) -> np.ndarray:
         """
         Apply MARS-style image cleaning. See `ImageCleaner.__call__()`
         """
 
         return mars_cleaning_1st_pass(
             self.subarray.tel[tel_id].camera.geometry,
-            image,
+            event.dl1.tel[tel_id].image,
             picture_thresh=self.picture_threshold_pe.tel[tel_id],
             boundary_thresh=self.boundary_threshold_pe.tel[tel_id],
             min_number_picture_neighbors=self.min_picture_neighbors.tel[tel_id],
@@ -562,14 +554,12 @@ class FACTImageCleaner(TailcutsImageCleaner):
         default_value=5.0, help="arrival time limit for neighboring " "pixels, in ns"
     ).tag(config=True)
 
-    def __call__(
-        self, tel_id: int, image: np.ndarray, arrival_times=None
-    ) -> np.ndarray:
+    def __call__(self, tel_id: int, event: ArrayEventContainer) -> np.ndarray:
         """Apply FACT-style image cleaning. see ImageCleaner.__call__()"""
         return fact_image_cleaning(
             geom=self.subarray.tel[tel_id].camera.geometry,
-            image=image,
-            arrival_times=arrival_times,
+            image=event.dl1.tel[tel_id].image,
+            arrival_times=event.dl1.tel[tel_id].peak_time,
             picture_threshold=self.picture_threshold_pe.tel[tel_id],
             boundary_threshold=self.boundary_threshold_pe.tel[tel_id],
             min_number_neighbors=self.min_picture_neighbors.tel[tel_id],
@@ -591,17 +581,15 @@ class TimeConstrainedImageCleaner(TailcutsImageCleaner):
         help="arrival time limit for neighboring " "boundary pixels, in ns",
     ).tag(config=True)
 
-    def __call__(
-        self, tel_id: int, image: np.ndarray, arrival_times=None
-    ) -> np.ndarray:
+    def __call__(self, tel_id: int, event: ArrayEventContainer) -> np.ndarray:
         """
         Apply MAGIC-like image cleaning with timing information. See `ImageCleaner.__call__()`
         """
 
         return time_constrained_clean(
             self.subarray.tel[tel_id].camera.geometry,
-            image,
-            arrival_times=arrival_times,
+            event.dl1.tel[tel_id].image,
+            arrival_times=event.dl1.tel[tel_id].peak_time,
             picture_thresh=self.picture_threshold_pe.tel[tel_id],
             boundary_thresh=self.boundary_threshold_pe.tel[tel_id],
             min_number_picture_neighbors=self.min_picture_neighbors.tel[tel_id],
--- a/src/ctapipe/image/cleaning.py
+++ b/src/ctapipe/image/cleaning.py
@@ -28,7 +28,7 @@ from abc import abstractmethod
 
 import numpy as np
 
-from ctapipe.containers import ArrayEventContainer
+from ctapipe.containers import MonitoringCameraContainer
 
 from ..core import TelescopeComponent
 from ..core.traits import (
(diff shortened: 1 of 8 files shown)


kosack (changes requested):
Can you explain why TailCutsDataVolumeReducer was changed as well here? This change seems to remove the ability to change which image cleaner implementation is used inside it,  but that is not mentioned at all in the changelog, and seems to be outside of the scope of the PR

maxnoe:
> This change seems to remove the ability to change which image cleaner implementation is used inside it,

Why would something called `TailCutsImageCleaner` allow to use a different cleaning method than `TailCuts`?

kosack:
> Why would something called TailCutsImageCleaner allow to use a different cleaning method than TailCuts?

I agree and mentioned that in the request,  but that reducer was configurable, presumably for a reason. I suspect more that it was just misnamed, rather than that it should be restricted to one type of cleaning.  Nevertheless, that should be a different PR if you want to change how it works, and the ACADA people should be involved.

maxnoe:
Ok...

I think we discussed this already elsewhere: keep the old signature in a method like: `clean_low_level` (please invent a better name), and have `__call__` use that internally.

Then the `TailCutsImageReducer` (and the TwoStepExtractor) can call the low-level API with minimal changes to them.

Hckjs:
Adapting it now to have a `__call__(self, tel_id: int, event: ArrayEventContainer)` which internally uses the abstract `clean_image(self, tel_id, image, arrival_times)` method defined in each `ImageCleaner`-subclass.

The TwoStepExtractor is already using the low level `tailcuts_clean` method.

src/ctapipe/image/cleaning.py
```diff
@@ -466,12 +466,36 @@ class ImageCleaner(TelescopeComponent):
     ``ImageCleaner.from_name()`` to construct an instance of a particular algorithm
     """
 
+    def __call__(self, tel_id: int, event: ArrayEventContainer) -> np.ndarray:
```
maxnoe, line 469: Should be an `@abstractmethod`

src/ctapipe/image/cleaning.py
```diff
@@ -466,12 +466,36 @@ class ImageCleaner(TelescopeComponent):
...
+            boolean mask of pixels passing cleaning
+        """
+        mask = self.clean_image(
+            tel_id=tel_id,
+            image=event.dl1.tel[tel_id].image,
+            arrival_times=event.dl1.tel[tel_id].peak_time,
+        )
+        return mask
+
     @abstractmethod
```
maxnoe, line 493: Not sure if this should now be an abstractmethod, since we explicitly changed the API to support more inputs. I.e. we shouldn't force people to implement the restricted API.

Hckjs: If people not implement the low-level cleaning method, the `VolumeReducer` can't use it, can it?

maxnoe: Yes.

Which is I think for the moment ok, we don't take anything away from `VolueReducer` that's currently supported.

In a future PR, we could update the VolumeReducer though to use the `__call__` API.

maxnoe:
Maybe a better change would be to for now just add the event as new optional parameter?
````
def __call__(self, tel_id, image, peak_time, *, event=None):
```

That is

a) backwards compatible
b) allows to clean other images than what is in `event.dl1.tel[tel_id].image` / `event.dl1.tel[tel_id].peak_time`
c) still allows accessing additional information from the event

src/ctapipe/image/cleaning.py
```diff
@@ -468,7 +469,12 @@ class ImageCleaner(TelescopeComponent):
 
     @abstractmethod
     def __call__(
-        self, tel_id: int, image: np.ndarray, arrival_times: np.ndarray = None
+        self,
+        tel_id: int,
+        image: np.ndarray,
+        arrival_times: np.ndarray = None,
+        *,
+        event: ArrayEventContainer = None,
```
kosack, line 477: I have just a general question:  In this PR you are simply adding the `event` as an extra parameter so the algorithms can access things like the interpolated pedestal variances.  Would it be better to simply turn this into a high-level API, simliar to other "processors", so that the call signature is just `cleaner(tel_id, event)` (and eventually after the event restructuring `cleaner(televent)`)?  Or is there some advantage to keeping `image` and `arrival_times` separate?

maxnoe: I am very confused now. That was the original proposal and you didn't like it.

https://github.com/cta-observatory/ctapipe/issues/2015#issuecomment-1549592270

Hckjs: I think the approach Max suggested [at this point](https://github.com/cta-observatory/ctapipe/pull/2511#discussion_r1494831073) and [here](https://github.com/cta-observatory/ctapipe/issues/2015#issuecomment-1558866965)  would be a good compromise for now, wouldn't it? Having an abstract `__call__(teld_id, event)`-function which internally uses a low-level `clean_image(self, tel_id, image, peak_time)` method. That wouldn't mix the implicit/explicit concepts and provides a simple "high level API" call signature like in other processors. It additionally ensures that people implementing new `ImageCleaner` are unchained by using different event/monitoring information to e.g. adapt cleaning thresholds/parameters. And it would also work for the `VolumeReducer`. 

I could adapt the `VolumeReducer` API (which is the only reason why we need the low-level `clean_image(..)` API, isn't it?) in a next PR and delete the low-level `clean_image(...)` afterwards leaving just the simple call signature.

kosack: > I am very confused now. That was the original proposal and you didn't like it.

It's not quite the same criticism: what I tried to avoid (as mentioned above) is having too many places where data was used implicitly, which just makes debugging and understanding of the code difficult. However, once you pass in the full event, you basically break that, and as we have done that elsewhere, I think it is ok.  I also agree that keeping he low-level functions explicit makes that better in any case.  

What I'm confused by is that In this current implementation,  the `__call__` function has both _explicit_ parameters (`image` and `peak_time`), but also potentially anything that is inside the event _implicily_.  So if we go that way, it makes no sense to keep the explicit parameters.   So just change to `__call__(self, tel_id, event)`

However, I still did not understand why the TailCutsDataVolumeReducer had to be drastically refactored: just change how it calls he cleaner inside to pass the full event, which is a one-line change.  Then you still keep the feature of being able to change cleaning algorithms, and nothing is broken.

kosack: The alternate scenario is just to pass in more explicit parameters, but that could get messy if there are many. 
 Can you remind me also what is the main problem this is trying to solve?  It was just to be able to use  pedestal-variance in the cleaning algorithms, right? Or are there other parameters that are needed?     I.e. would an alternate scenario where you don't pass the event, but rather use `__call__(self, tel_id, image, arrival_time, ped_variance)` be sufficient? If that is the case, maybe passing the event is overkill, but it may limit fancier algorithms...

kosack: Maybe the simpler way of saying it is: just pick one way or the other. Either pass all possible parameters explicitly, or just the event.   But mixing both ways makes no sense other than adding confusion.

Hckjs: > However, I still did not understand why the TailCutsDataVolumeReducer had to be drastically refactored: just change how it calls he cleaner inside to pass the full event, which is a one-line change. Then you still keep the feature of being able to change cleaning algorithms, and nothing is broken.

How can i pass the full event to the cleaner call inside the VolumeReducer, when its own call just uses `(waveforms, tel_id=None, selected_gain_channel=None)` ? We have do adapt its API for using the full event first, don't we? Thats why we came up with this different approaches to keep the functionality of the VolumeReducer.

kosack: Good point, yes I think it will have to also get the full event as well, and indeed fancier algorithms might need it. For now, having the old API function separate from `__call__`  is probably sufficient as you said above until the full refactoring can be done.

kosack: Yeah, I did worry about the use case where you pass some other image. Then it does get messy.

kosack: Ok, let's stick with just the optional Event for now then.   Or just pass in an extra optional pedestal_variances image

kosack: Actually, it occurs to me that the source of the pixel noise level (pedestal widths) is also not something that can be assumed to be in a single place in the event, much like the issue you raised with the `image` above.  If `TailCutsCleaner` is used in the DataVolumeReducer for example, it would need to get the pedestal widths from either the R1 pedestals, or the even the DL0 event-wise pedestals (for cameras that generate those), whereas for normal image cleaning, they should come from the DL1 pedestals.   So I would recommend that we just :

```py3
def __call__(self, 
    tel_id: int,  
    image: ndarray, arrival_time: Optional[NDArray]=None, 
    noise_level: Optional[NDArray]=None
):
    """
    ...
    noise_level:  Optional[NDArray]
         The level (width, assuming 0 bias) of noise in each pixel, required for noise-based threshold cleaners. 
         This is typically the pedestal variance.
    """
```

Would that be satisfactory for implementing the noise cleaning used in real-data processing, or is there still some information missing?

Hckjs: Is that applicable for e.g. [lstchains pedestal cleaning](https://github.com/cta-observatory/cta-lstchain/blob/5ea3010a6169d0640d4af2640c3b0abbb3de5b9c/lstchain/scripts/lstchain_dl1ab.py#L197), where they are adapting the cleaning thresholds also based on pedestal_mean (i.e. pedestal_mean + sigma * pedestal_std)? Where do we correct for the bias then? 
In addition @aleberti wants to use an ImageCleaner subclass to implement their cleaning in magicctapipe using the pixel_status for interpolating between unusable pixel if understood correctly.
There are certainly more ways of using additional information for the cleaning, so sticking with the optional event could be less strict. Wouldn't there be cleaners that the VolumeReducer can't use anyway (when considering all event information)? So maybe restrict the VolumeReducers in the usable ImageCleaners?

i just noticed that we should also add the arrival_times to the [cleaners call in the DataVolumeReducer](https://github.com/cta-observatory/ctapipe/blob/d73970186d3b43d48c7e53e42d5746508155f4ab/src/ctapipe/image/reducer.py#L194) to allow different cleaners using arrival_times.

maxnoe: >          This is typically the pedestal variance.

Pedestal standard deviation, right? We should be clear what we give here.

> Is that applicable for e.g. [lstchains pedestal cleaning](https://github.com/cta-observatory/cta-lstchain/blob/5ea3010a6169d0640d4af2640c3b0abbb3de5b9c/lstchain/scripts/lstchain_dl1ab.py#L197), where they are adapting the cleaning thresholds also based on pedestal_mean (i.e. pedestal_mean + sigma * pedestal_std)?

The pedestal mean (bias) should already have been removed by the calibration code and should not be needed anymore at the level of image cleaning.

kosack: > that applicable for e.g. [lstchains pedestal cleaning](https://github.com/cta-observatory/cta-lstchain/blob/5ea3010a6169d0640d4af2640c3b0abbb3de5b9c/lstchain/scripts/lstchain_dl1ab.py#L197), where they are adapting the cleaning thresholds also based on pedestal_mean (i.e. pedestal_mean + sigma * pedestal_std)?

If it's really needed, we could simply pass in a PedestalContainer that contains camera images of pedestal  mean and std.

maxnoe: How about passing the whole "MonitoringContainer"?

That also contains the broken pixels requested by aleberti (although that might not be needed after #2518 )

maxnoe:
Just to summarize from the discussions that are now in collapsed threads: we decided to not pass the full event, but just the `CameraMonitoringContainer`, which should have all information necessary for advanced cleaning algorithms.

Tobychev (approved):

PR #2516, opened 2024-02-23
Fix entry point specification

maxnoe:
Noticed because in the docs build, we had  a lot of errors in the log:

 ModuleNotFoundError: No module named 'ctapipe.reco.sklearn.ParticleClassifier'; 'ctapipe.reco.sklearn' is not a package
Error loading ctapipe_reco plugin: 'ctapipe.reco.sklearn.DispReconstructor'

The last . should be a :


kosack (approved):

maxnoe:
I am going to merge as it is not a code change and blocks all other PRs.

Issue #1836, opened 2022-02-01
Re-introduce possibility of running `ImageExtractor` on waveforms with multiple gains


maxnoe:
**Please describe the use case that requires this feature.**

Calibration events won't be gain selected, as we need to have calibration coefficients for all pixels for all gains.

**Describe the solution you'd like**

Make ImageExtractors work with non-gain-selected input again

**Describe alternatives you've considered**

Using a python loop over the gain channels with `ManualGainSelector`, but that could be much less efficient for `ImageExtractors` that could easily be vectorized over multiple gain channels.

maxnoe:
We discussed this during today's developer call.


The conclusion was that we:

a) Want to reintroduce the feature that extractors can work on two gain channels and that should result in two images, one per gain channel

b) R1 and DL0 waveforms should always have the shape (n_channels, n_pixels, n_samples), even if n_channels is 1. 

Together with the event type, pixel status and the camera readout, it is then clear whether gain selection has happened or the camera has just one gain.

Issue #1226, opened 2020-03-07
Proposal: Refactor FlatFieldCalculator and PedestalCalculator


watsonjj:
The `FlatFieldCalculator` and the `PedestalCalculator` that are currently located in the `calib` submodule are not directly used for calibration. Instead they are used for monitoring and analysis of the charge statistics. From these values it is possible to extract the calibration coefficients, but this is secondary.

Additionally, these classes are fairly large and complicated as they instance their own `ImageExtractor`, something I think should be discouraged. To ensure a common utilisation of `ImageExtractor` and other calibration, I would suggest that a correctly configured `CameraCalibrator` should be provided to such classes, as I describe in #1147.

As part of #1026, and in order to achieve the second pass described in https://github.com/cta-observatory/ctapipe/issues/1147#issuecomment-546264078 I intended to implement the calculators which directly calculate and fill the calibration coefficients created in #1160. However, I am finding that these current "Calculator" classes are obstructing and would cause confusion with the calibration coefficient calculators I intend to contribute.

I propose that the calculators that exist in the `calib` submodule should only directly calculate the calibration coefficients. The classes should also not instance their own `ImageExtractor`. The classes as they currently exist, to monitor the charge/pedestal statistics, should be moved elsewhere.

kosack:
Are you talking about on-the-fly computation here of pedestals and FF? Or just to ensure that the current off-line FF and Pedestal calculators use the same CameraCalibrator internally?

watsonjj:
My main concern is the `FlatFieldCalculator` and `PedestalCalculator` don't calculate calibration coefficients. Their purpose appears to be more towards "charge statistics monitoring". Further analysis or "calculation" has to be done on the return values to extract the calibration coefficients.

FrancaCassol:
Hi @watsonjj ,
these modules are the base Components to fill the monitoring containers used to calculate the calibration coefficients (which are then finalised in a tool based on them). They are presently in use and represent the core components for the LST1 calibration (which works successfully). Are you really sure you can not propose a development without touching them? At least till you do not assure an equivalent calibration level ;-)

watsonjj:
This is a problem I was conscious of, and is why I am scared of touching these classes. I don't think progress in development of ctapipe should be stalled because it would break specific cameras pipelines. With this approach ctapipe will never be able to progress. It is clear from the open issues that there are lots of areas that may change about ctapipe, which will break current api.

[...]

Issue #1147, opened 2019-10-23
Proposals for charge calibration flow


watsonjj:
I have prepared some proposals for the flow of charge calibration. I am mainly concerned with the charge calibration at the DL1 level.

Certain aspects of the charge calibration are common in all of these proposals, and I think they are mostly agreed upon:
- The Calculators attach calibration coefficients to the MON container after they are calculated
- The CameraCalibrator takes the calibration coefficients from the MON container and applies them to the extracted charge
- Events are labelled with a trigger type, which dictates if it is used in a calibration calculator
- The three points above allow an on-the-fly calibration to occur in an intuitive way
- A Tool can be created that iterates over a file, and stores the resulting MON container to a file
- A MON source can be setup to read MON information into the container (therefore providing either an initial calibration, or an alternative to simultaneously running a Calculator in the event loop for on-the-fly calibration)

Connected previous discussions: #1037 #1041 

I will post the proposals I currently have below, so they can be discussed.

watsonjj:
![image](https://user-images.githubusercontent.com/17825673/67409419-d6eba380-f5ba-11e9-99fb-fbef358a9026.png)

watsonjj:
![image](https://user-images.githubusercontent.com/17825673/67409462-e79c1980-f5ba-11e9-830d-7c2ee8d0ebda.png)

watsonjj:
![image](https://user-images.githubusercontent.com/17825673/67409508-f256ae80-f5ba-11e9-850f-4445580442e5.png)

watsonjj:
![image](https://user-images.githubusercontent.com/17825673/67409526-fbe01680-f5ba-11e9-896a-6cb74a5fbfc0.png)

watsonjj:
At the moment my preferred option is **Approach 3**

watsonjj:
EDIT for approach 3:

![image](https://user-images.githubusercontent.com/17825673/67410069-d0116080-f5bb-11e9-93ae-cc930a3f5105.png)

kosack:
Let's separate calibration *calculation* from calibration *application*. 

Let's start with a simple 2-pass method:

- Pass 0: 
  - process DC/PE data and generate coefficients per pixel (special run).  This should include the overall calib_scale for MC data. 

- pass 1 : **calculate calibration**:  loop through data (perhaps only calib events), output a DL0/Monitoring files with:
   - pedestal/pixel (every second or so)
   - flatfield correction /pixel

[...]

PR #2519, opened 2024-02-27
Migration of flatfield and pedestal components from lstchain

TjarkMiener:
This PR draft adds a migration proposal for the flatfield and pedestal components from lstchain into ctapipe. Shared functionality is moved into a parent component, while the different subcomponents hold the _process() function for the conventional processing of flatfield and pedestal events. Additional subcomponents with complex processing routines of flatfield and pedestal events can be added in the future.


src/ctapipe/calib/camera/extractor.py
```diff
@@ -0,0 +1,426 @@
...
+            < self.charge_std_cut_outliers[0] * std_of_pixel_std[:, np.newaxis],
+            deviation
+            > self.charge_std_cut_outliers[1] * std_of_pixel_std[:, np.newaxis],
+        )
+
+        return {
+            "sample_time": (
+                trigger_times[0] + (trigger_times[-1] - trigger_times[0]) / 2
+            ).unix
+            * u.s,
```
Hckjs, line 271: I think you need at least one of sample_time_{min, max} as timestamps to match the current event (trigger_times) with the related set of calibration coefficients.

maxnoe:
Hi @TjarkMiener thanks, this is much needed.

However, I think we should first discuss a bit how we want the general API of these classes to look like.

I think we certainly don't want to keep the current one.

* Do we need an abstract base class for the calibration like we have currently?

    This would depend on whether we can use the same general structure and algorithm per telescope. Will all telescopes use the F-Factor method?

* Do we use a custom tool to compute these calibration coefficients or do we integrate into `ctapipe-process`.

    I think specific tool is the way to go.

* What structure do we need? 

    Currently we have three classes, one for pedestal events, one for flatfield events and one tying these two together to compute the final calibration coefficients. That makes sense I think, but there is also a lot of duplication between the flatfield and the pedestal base class in lstchain currently. This should be refactored.

* How do we store the output?

* How do we read it back in for calibration?

There are a couple of related open issues:

* https://github.com/cta-observatory/ctapipe/issues/1836 (as  calibration events come with two gains)
* https://github.com/cta-observatory/ctapipe/issues/1397
* https://github.com/cta-observatory/ctapipe/issues/1226
* https://github.com/cta-observatory/ctapipe/issues/1147
* https://github.com/cta-observatory/ctapipe/issues/856

src/ctapipe/calib/camera/extractor.py
```diff
@@ -0,0 +1,426 @@
...
+    # TODO:
+        1) Discuss whether to express the outlier cuts based on quantiles.
+        2) Discuss whether to calculate the number of iterations for sigma clipping on the fly.
+
+    Attributes
+    ----------
+    sample_size : int
+        Minimal number of calibration events requested for the statistics
+    update_frequency : int
+        Calculation frequency of the statistics
```
maxnoe, line 51: This value is not clear to me from the explanation given.

src/ctapipe/calib/camera/extractor.py
```diff
@@ -0,0 +1,426 @@
...
+            )
+        # set up the buffer
+        self.n_events_in_buffer = 0  # number of events in sample
+        self.trigger_time = []  # trigger time of event in sample
+        self.charges = []  # charge per event in sample
+        self.arrival_times = []  # arrival time per event in sample
+        self.broken_pixels = []  # masked pixels per event in sample
+
+        # load the waveform charge extractor
+        self.extractor = ImageExtractor.from_name(
```
maxnoe, line 122: Since you have `image_extractor` as a `TelescopeParameter`, you need to create all types requested, not just the one type.

See e.g.:
https://github.com/cta-observatory/ctapipe/blob/eb43957c838bfe30fd1a6403601b724b984a25e4/src/ctapipe/calib/camera/calibrator.py#L130

src/ctapipe/calib/camera/extractor.py
```diff
@@ -0,0 +1,426 @@
...
+            deviation
+            < self.charge_std_cut_outliers[0] * std_of_pixel_std[:, np.newaxis],
+            deviation
+            > self.charge_std_cut_outliers[1] * std_of_pixel_std[:, np.newaxis],
+        )
+
+        return {
+            "sample_time": (
+                trigger_times[0] + (trigger_times[-1] - trigger_times[0]) / 2
+            ).unix
```
maxnoe, line 270: Please don't convert to unix here, just keep it as astropy time object

src/ctapipe/calib/camera/extractor.py
```diff
@@ -0,0 +1,426 @@
...
+        Clean up the buffer by removing the events that will not participate
+        in the caluclation of the next sample.
+        """
+        self.n_events_in_buffer -= self.update_frequency
+        self.trigger_time = self.trigger_time[self.update_frequency :]
+        self.charges = self.charges[self.update_frequency :]
+        self.arrival_times = self.arrival_times[self.update_frequency :]
+        self.broken_pixels = self.broken_pixels[self.update_frequency :]
+
+    @abstractmethod
```
maxnoe, line 339: private methods should not be abstract

src/ctapipe/calib/camera/extractor.py
```diff
@@ -0,0 +1,426 @@
...
+    def __init__(self, **kwargs):
+        super().__init__(**kwargs)
+
+    def _process(self, event: ArrayEventContainer):
+        """
+        Process the sample statistics of flat-field events and fill the
+        mon.tel[self.tel_id].flatfield container
+        """
+
+        container = event.mon.tel[self.tel_id].flatfield
```
maxnoe, line 365: I don't think we should set the new container on the event.

Instead, let the class keep a list of computed calibrations.

src/ctapipe/calib/camera/extractor.py
```diff
@@ -0,0 +1,426 @@
...
+            self.charges,
+            self.sample_masked_pixels,
+            self.trigger_times,
+        )
+        arrivaltime_stats = self._calculate_arrival_time_stats(
+            self.arrival_times,
+            self.sample_masked_pixels,
+        )
+
+        flatfield_container = {
```
maxnoe, line 377: We made very bad experiences with modifying containers in place. You should create and fill a fresh container.

You can also directly pass the fields to the container itself.

src/ctapipe/calib/camera/extractor.py
```diff
@@ -0,0 +1,426 @@
...
+            **arrivaltime_stats,
+        }
+        for key, value in flatfield_container.items():
+            setattr(container, key, value)
+
+        # update the flatfield mask
+        ff_charge_failing_pixels = np.logical_or(
+            container.charge_median_outliers, container.charge_std_outliers
+        )
+        event.mon.tel[self.tel_id].pixel_status.flatfield_failing_pixels = (
```
maxnoe, line 389: Again, we shouldn't modify the input event. That's for the code that reads back the calibration coefficients and interpolates / matches them to events when applying the calibration.

src/ctapipe/calib/camera/extractor.py
```diff
@@ -0,0 +1,426 @@
...
+                 False otherwise
+        """
+
+        # extract the charge of the event and
+        # the peak position (assumed as time for the moment)
+        dl1: DL1CameraContainer = self._extract_charge_and_time(event)
+
+        # append valid dl1 events
+        self.n_events_in_buffer += 1
+        self.trigger_times.append(event.trigger.tel[self.tel_id].time)
```
maxnoe, line 146: `self.tel_id` seems not to be defined anymore.

Also: these components should be able to handle multiple telescopes I think, so we keep e.g. a dict from tel_id to the properties.

maxnoe, line 146: It also looks like you are implementing a deque with the logic in clean:
https://docs.python.org/3/library/collections.html#collections.deque

E.g. if you always want to keep the most recent N events, use a `deque(maxsize=N)` and just `append` to it.

Once you reach maxsize, the oldest event will be discarded.

src/ctapipe/calib/camera/extractor.py
FrancaCassol: I agree, this is not clear

maxnoe:
@FrancaCassol I think it doesn't make much sense to apply any optimization to the existing code, as we now we need to replace it completely relatively soon.

maxnoe:
The logic is already somewhat different to what is implemented in LST currently, so clearly this at least has some new development.

kosack (commented):
I have a general question:  these classes are for computing running pedestals and flatfields, but they process each event at at time and have mechanisms for bunching events, and keeping running statistics.    Now that we know all calibration events will be in their own file and that for offline processing we can do a first pass on that file to compute coefficents, is the current design with an event-loop really the best one?  The only use case I can see for that is for real-time analysis, but that is not what we have, we can compute all calibration after the run has finished and can even loop many times if needed.

I would think it is much more efficient and easy to implement the computation of calibration coefficients using a procedure that first pre-processes and writes out the events to an HDF5 file, and then using TableLoader, apply functions to all events at once, similar to how `ctapipe-apply-models` works.    Should we maybe have a design discussion about this?  We can still of course integrate this method for the time being, but I suspect it would be much less code and allow fancier algorithms if we process *column-wise* rather than *row-wise*(per event). For example doing a boxcar average for the pedestals for example would be trivial to implement.

TjarkMiener:
This PR draft was meant to refactor the current lstchain components for flatfield and pedestal to remove repeated code and start a discussion on the migration. In addition, I added this config parameter, which allows the calculation of the stats after a given iteration and not once the sample buffer is filled. Originally I had in mind to follow what @kosack suggested (pre-process/integrate charges and arrival times, dump dl1-like events, and then table read plus the calculation of coefficients). I dropped this idea because a presumed 2x 100Hz calibration rate and an obs block duration of 20-25 mins would exceed the RAM limitation of 2GB per job when reading the whole table. However, as we discussed in a separate thread, the calibration rate will be likely reduced in the near future. So I guess we could reconsider the suggestion of integrating and temporarily storing dl1-like calibration data to read them afterward using the TableLoader and apply functions to all events at once.

The idea I followed in this PR for the design of the components was to be able to flexibly add subcomponents that could host the "fancy algorithms" via the `_process()` function. You could in principle increase the `sample_size` and then apply the complex functions on the `self.charges` and `self.arrival_times` arrays as you would do for all the events read with the TableLoader. Those complex functions could even be applied on a pixel-wise level, i.e. we might want to treat pixels that are affected by stars differently. The current subcomponents `FlatFieldExtractor` and `PedestalExtractor` were targeting the cat-B processing, while fancy subcomponents can be added in the future for cat-C processing.

I like the idea proposed in #1226 for the on-the-fly computation of flatfield and pedestals, even though we might want to store the charge statistics monitoring of the calibration events for the data quality monitoring.

maxnoe:
The table loader also supports chunked loading, which allows you to iterate through chunks and keeping consecutive chunks in memory, you only need roughly the amount of events of your windows sizes.

TjarkMiener:
True, good point! My initial plan was to follow a workflow like: integrating charges and extracting arrival times via the `process` tool  `($ ctapipe-process ... --write_images --no-write_parameters)`. And then code a tool for the calculation of the camera calibration coefficients based on the TableLoader. Therefore, I would need to make the `process` tool compatible with calibration events, i.e. allowing two gains integration. By using the `process` tool we would directly ensure to follow the standard `ctapipe` DL1 data model, which then probably needs some adjustments for the allowance of two gains images and peak_times. I think the dl1-like files for the calibration events can be stored temporarily. One advantage would also be that those dl1-like calibration files could be sent directly to the off-site data center after production rather than the waveforms of the calibration events. A naive guess from my side would be that we don't want to repeat the charge integration and arrival time extraction for the calibration events each time we are calculating camera calibration coefficients (on- or off-site).

I'm closing this draft for now and working on the suggestion above in a separate branch. In case you think we should the design of this PR feel free to comment and I will reopen this draft. Thanks for the discussion and suggestion!

PR #2522, opened 2024-03-12
Enable processing of interleaved calibration events via the process tool

TjarkMiener:
This commit enables the processing of flatfield and sky pedestal events up to dl1-like using the `ctapipe-process` tool. Code changes are tested with interleaved calibration events from LST-1. I needed to tweak the pointing frame in `hdf5eventsource` in order to make it work with the standard `lstchain` R1 interleaved files (not produced with the latest `ctapipe` version).

Currently only tested with the ImageExtractors (LocalPeakWindowSum and FixedWindowSum), which are the default ones.

### DIFF
--- a/src/ctapipe/calib/camera/calibrator.py
+++ b/src/ctapipe/calib/camera/calibrator.py
@@ -294,15 +294,6 @@ class CameraCalibrator(TelescopeComponent):
         if self._check_r1_empty(waveforms):
             return
 
-        if self.image_extractor_type.tel[tel_id] not in [
-            "LocalPeakWindowSum",
-            "FixedWindowSum",
-        ]:
-            raise ValueError(
-                f"Invalid ImageExtractor '{self.image_extractor_type.tel[tel_id]}'."
-                "Valid options: LocalPeakWindowSum FixedWindowSum."
-            )
-
         n_channels, n_pixels, n_samples = waveforms.shape
 
         broken_pixels = event.mon.tel[tel_id].pixel_status.hardware_failing_pixels
--- a/src/ctapipe/calib/camera/calibrator.py
+++ b/src/ctapipe/calib/camera/calibrator.py
@@ -7,7 +7,13 @@ import astropy.units as u
 import numpy as np
 from numba import float32, float64, guvectorize, int64
 
-from ctapipe.containers import DL0CameraContainer, DL1CameraContainer, PixelStatus
+from ctapipe.containers import (
+    DL0CameraContainer,
+    DL1CameraContainer,
+    PixelStatus,
+    EventType,
+)
+
 from ctapipe.core import TelescopeComponent
 from ctapipe.core.traits import (
     BoolTelescopeParameter,
@@ -346,8 +352,8 @@ class CameraCalibrator(TelescopeComponent):
         tel = event.r1.tel or event.dl0.tel or event.dl1.tel
         for tel_id in tel.keys():
             if (
-                self.process_flatfield_events
-                or self.process_pedestal_events
+                event.trigger.event_type == EventType.SKY_PEDESTAL
+                or event.trigger.event_type == EventType.FLATFIELD
             ):
                 self._process_calib(event, tel_id)
             else:
--- a/src/ctapipe/calib/camera/calibrator.py
+++ b/src/ctapipe/calib/camera/calibrator.py
@@ -11,6 +11,7 @@ from ctapipe.containers import DL0CameraContainer, DL1CameraContainer, PixelStat
 from ctapipe.core import TelescopeComponent
 from ctapipe.core.traits import (
     BoolTelescopeParameter,
+    Bool,
     ComponentName,
     TelescopeParameter,
 )
@@ -87,12 +88,12 @@ class CameraCalibrator(TelescopeComponent):
         ),
     ).tag(config=True)
 
-    process_flatfield_events = BoolTelescopeParameter(
+    process_flatfield_events = Bool(
         default_value=False,
         help="Process flatfield events from a calibration data stream.",
     ).tag(config=True)
 
-    process_pedestal_events = BoolTelescopeParameter(
+    process_pedestal_events = Bool(
         default_value=False,
         help="Process pedestal events from a calibration data stream.",
     ).tag(config=True)
@@ -345,8 +346,8 @@ class CameraCalibrator(TelescopeComponent):
         tel = event.r1.tel or event.dl0.tel or event.dl1.tel
         for tel_id in tel.keys():
             if (
-                self.process_flatfield_events.tel[tel_id]
-                or self.process_pedestal_events.tel[tel_id]
+                self.process_flatfield_events
+                or self.process_pedestal_events
             ):
                 self._process_calib(event, tel_id)
             else:
--- a/src/ctapipe/calib/camera/calibrator.py
+++ b/src/ctapipe/calib/camera/calibrator.py
@@ -87,6 +87,16 @@ class CameraCalibrator(TelescopeComponent):
         ),
     ).tag(config=True)
 
+    process_flatfield_events = BoolTelescopeParameter(
+        default_value=False,
+        help="Process flatfield events from a calibration data stream.",
+    ).tag(config=True)
+
+    process_pedestal_events = BoolTelescopeParameter(
+        default_value=False,
+        help="Process pedestal events from a calibration data stream.",
+    ).tag(config=True)
+
     def __init__(
         self,
         subarray,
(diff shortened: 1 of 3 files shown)


src/ctapipe/tools/process.py
```diff
@@ -307,6 +368,12 @@ def start(self):
                 )
                 continue
 
+            if not self.should_process(event):
```
mexanick, line 371: Do we really want this? It is a huge load of debug printouts. If you really want to have it, please consider changing log level to TRACE

src/ctapipe/calib/camera/calibrator.py
```diff
@@ -282,6 +292,44 @@ def _calibrate_dl1(self, event, tel_id):
...
+        if n_channels == 2:
+            no_gain_selection[1] = 1
+
+        # Extract charge and arrival time and return the DL1CameraContainer
+        extractor = self.image_extractors[self.image_extractor_type.tel[tel_id]]
+
+        dl1 = extractor(
+            waveforms,
+            tel_id=tel_id,
+            selected_gain_channel=no_gain_selection,
```
mexanick, line 327: Can you please explain the logic behind this? I see that if n_channels is 1, you pass 1d array of zeros, and if n_channel is 2 it is a 2d array of zeros and ones...

src/ctapipe/calib/camera/calibrator.py
```diff
@@ -282,6 +293,44 @@ def _calibrate_dl1(self, event, tel_id):
...
+        n_channels, n_pixels, n_samples = waveforms.shape
+
+        broken_pixels = event.mon.tel[tel_id].pixel_status.hardware_failing_pixels
+        if broken_pixels is None:
+            broken_pixels = np.zeros(n_pixels, dtype=bool)
+
+        # TODO: See #1836 (process all provided gains in the ImageExtractor)
+        no_gain_selection = np.zeros((n_channels, n_pixels), dtype=np.int64)
+
+        if n_channels == 2:
```
Hckjs, line 319: should we maybe force to have two channels here since we need both for calibrations? i.e. raising an error if `n_channels != 2` ?

FrancaCassol:
Hi @TjarkMiener, 
I am not so sure to understand the  goal of this code, why do you need a special process for the calibration and flatfield events? (perhaps I missed something ...)

src/ctapipe/containers.py
```diff
@@ -483,16 +483,12 @@ class DL1CameraContainer(Container):
 
     image = Field(
         None,
-        "Numpy array of camera image, after waveform extraction." "Shape: (n_pixel)",
-        dtype=np.float32,
```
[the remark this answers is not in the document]
TjarkMiener: Since `ndim` is removed, the container complains when two gain images are filled:
```
ctapipe.core.container.FieldValidationError: DL1CameraContainer Field 'image': the value '[[76.30804  76.080025 76.71846  ... 84.10605  67.9628   84.71408 ]
 [65.03804  65.34207  74.57242  ... 82.70825  62.36258  73.0401  ]]' (<class 'numpy.ndarray'>) is invalid:  Should have numpy dtype float32
```

src/ctapipe/tools/process.py
TjarkMiener: No strong opinion about the log level from my side.

src/ctapipe/calib/camera/calibrator.py
TjarkMiener: I think it's just a workaround because the ImageExtractors are currently designed to work with gainselected images. Therefore this issue should be addressed in a separate PR.

src/ctapipe/calib/camera/calibrator.py
TjarkMiener: But then it will not work with a single gain camera, which is currently supported (I guess)

TjarkMiener:
Ciao @FrancaCassol, following the suggestion from @kosack in #2519, this PR enables the charge integration and arrival time extraction of the calibration data stream. The DL1-like images and peak times are stored following the current `ctapipe` data model. Those files, which will most likely be stored temporarily, are supposed to be the input of the (TBD - to be developed) tool extracting the cat-B camera calibration coefficients. The advantage would be that we can make use of `ctapipe`'s `TableLoader`.

FrancaCassol:
> Ciao @FrancaCassol, following the suggestion from @kosack in #2519, this PR enables the charge integration and arrival time extraction of the calibration data stream. The DL1-like images and peak times are stored following the current `ctapipe` data model. Those files, which will most likely be stored temporarily, are supposed to be the input of the (TBD - to be developed) tool extracting the cat-B camera calibration coefficients. The advantage would be that we can make use of `ctapipe`'s `TableLoader`.

Hi @TjarkMiener, I see. However, in the case the flatfield and pedestal events are in a file separated by cosmics (which is our case) and if we solve [https://github.com/cta-observatory/ctapipe/issues/1836](https://github.com/cta-observatory/ctapipe/issues/1836), flatfield and pedestal events could not just be processed with the standard _calibrate_dl1 function? We need just setting the correct configuration cards (as we do for the other cases). The only thing to provide is to select the correct charge extractor for each event_type (which, by the way it seems to me it is missing in this PR?), this is necessary only in the case flatfields and pedestals are in the same file and we want to process them together, otherwise there is no problem.

mexanick:
> Hi @TjarkMiener, I see. However, in the case the flatfield and pedestal events are in a file separated by cosmics (which is our case)

1. ACADA will provide us separate streams per telescope and per calibration kind 
2. cta-process will work with only particular event types, so even if the calibration events are truly interleaved (or FF and pedestal combined), it won't be a problem.

> and if we solve #1836, flatfield and pedestal events could not just be processed with the standard _calibrate_dl1 function? 

We want to actually change it a bit and use the TableLoader.

>We need just setting the correct configuration cards (as we do for the other cases). The only thing to provide is to select the correct charge extractor for each event_type (which, by the way it seems to me it is missing in this PR?),

The charge extractor type is an input parameter and so far is restricted to two options - LocalPeakWindowSum and FixedWindowSum. The door is open for other extractors in case other telescopes/cameras will need them.

> this is necessary only in the case flat fields and pedestals are in the same file and we want to process them together, otherwise there is no problem.

See the part of the answer above.

src/ctapipe/containers.py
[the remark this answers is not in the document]
kosack: Hmm. Perhaps we then need to break this into a parent contain and two childred that inherit from it, adding the `image`  field.  That way it's clear that DL1CameraContainer for Calibration events look different than calibrated ones.   Maybe DL1CameraContainer  (contains all common fields) → CalibratedDL1CameraContainer (contains float DL1 image with ndims=1), UncalibratedDL1CameraContainer (contains int dl1 image with ndims=2?)

Or something along those lines...

src/ctapipe/calib/camera/calibrator.py
```diff
@@ -87,6 +88,16 @@ class CameraCalibrator(TelescopeComponent):
         ),
     ).tag(config=True)
 
+    process_flatfield_events = Bool(
```
maxnoe, line 91: We already have a `EventTypeFilter` in the process tool for this, no new options should be needed.

src/ctapipe/calib/camera/calibrator.py
```diff
@@ -282,6 +293,44 @@ def _calibrate_dl1(self, event, tel_id):
         # store the results in the event structure
         event.dl1.tel[tel_id] = dl1
 
+    def _process_calib(self, event, tel_id):
+        waveforms = event.r1.tel[tel_id].waveform
+        if self._check_r1_empty(waveforms):
+            return
+
+        if self.image_extractor_type.tel[tel_id] not in [
```
maxnoe, line 301: Why do you only allow these two extractors? We have an abstract API for a reason to enable users to chose.

src/ctapipe/calib/camera/calibrator.py
maxnoe: I think we should add an API back to the ImageExtractor for non-gain-selected events instead of special casing two event types here.

See https://github.com/cta-observatory/ctapipe/issues/1836

src/ctapipe/tools/process.py
```diff
@@ -187,6 +201,19 @@ def setup(self):
         self.write = self.enter_context(
             DataWriter(event_source=self.event_source, parent=self)
         )
+
+        if (
+            self.calibrate.process_flatfield_events
+            or self.calibrate.process_pedestal_events
+        ):
+            self.log.warning(
```
maxnoe, line 209: Please don't mess with the configuration system like this.

src/ctapipe/tools/process.py
```diff
@@ -249,8 +294,24 @@ def should_compute_muon_parameters(self):
...
 
+        if (
+            self.calibrate.process_flatfield_events
+            or self.calibrate.process_pedestal_events
+        ):
+            return False
+
         return False
 
+    def should_process(self, event):
```
maxnoe, line 305: There is already an event type filter in the tool that basically does the same and should be used for this.

src/ctapipe/tools/process.py
TjarkMiener: Okay, removed!

src/ctapipe/tools/process.py
TjarkMiener: Good point!

src/ctapipe/calib/camera/calibrator.py
TjarkMiener: The check has been removed.

src/ctapipe/tools/process.py
TjarkMiener: This log is removed because the event type filter is now used.

src/ctapipe/calib/camera/calibrator.py
maxnoe: I think we should just add an API to the extractors for not gain selected waveforms. 

I.e. allow `selected_gain_channel=None` which should mean no gain selection happened and that the waveform has shape `n_channels, n_pixels, n_samples`

TjarkMiener: yes, I'm trying to solve it right now. I will let you know.

Hckjs: Im also currently working on #1836 which would solve it - ill create a draft PR today or tomorrow, maybe we can discuss it there.

TjarkMiener: great! thank you @Hckjs I have also contacted you on the CTA North slack channel with the changes I had in mind so far. Maybe they can be helpful.

src/ctapipe/containers.py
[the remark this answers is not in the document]
TjarkMiener: What is the drawback of having it flexible as it is in the PR?
One problem with having two inheritance containers would be that if we want to store the calibration dl1 images for a single gain camera, we can not store them in `UncalibratedDL1CameraContainer` because of the dimension restrictions. Though, it would also not belong in `CalibratedDL1CameraContainer`. A semantic hardcoding structure, i.e. `OneGainDL1CameraContainer`; `TwoGainDL1CameraContainer`, would also not be ideal since then we don't have information if the `OneGainDL1CameraContainer` is gain_selected/calibrated or not.

TjarkMiener:
Redundant due to #2529

PR #2524, opened 2024-03-20
Add a trait for astropy quantities

LukasBeiske:
Having a configurable trait containing an `astropy.units.Quantity` is useful for the irf tool and will probably be useful for other high-level tools, too.

### DIFF
--- a/src/ctapipe/core/traits.py
+++ b/src/ctapipe/core/traits.py
@@ -79,14 +79,26 @@ class AstroQuantity(TraitType):
 
     def __init__(self, physical_type=None, **kwargs):
         super().__init__(**kwargs)
-        self.physical_type = physical_type
+        if physical_type is not None:
+            if isinstance(physical_type, u.PhysicalType):
+                self.physical_type = physical_type
+            elif isinstance(physical_type, u.UnitBase):
+                self.physical_type = u.get_physical_type(physical_type)
+            else:
+                raise TraitError(
+                    "Given physical type must be either of type"
+                    " astropy.units.PhysicalType or a subclass of"
+                    f" astropy.units.UnitBase, was {type(physical_type)}."
+                )
+        else:
+            self.physical_type = physical_type
 
         if self.default_value is not Undefined and self.physical_type is not None:
             default_type = u.get_physical_type(self.default_value)
             if default_type != self.physical_type:
                 raise TraitError(
                     f"Given physical type {self.physical_type} does not match"
-                    f" physical type of the default value {default_type}."
+                    f" physical type of the default value, {default_type}."
                 )
 
     def info(self):
--- a/src/ctapipe/core/traits.py
+++ b/src/ctapipe/core/traits.py
@@ -5,6 +5,7 @@ import os
 import pathlib
 from urllib.parse import urlparse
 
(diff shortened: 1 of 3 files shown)


kosack (changes requested):
Nice! One  suggestion would be to add the ability to set a _physical type_ or default unit, and have it fail if the user tries to set the value to a quantity that is not convertable.  

I.e. if my trait is called `energy_min` and the user tries to set it to "12 m" it should fail.

maxnoe:
Also missing are tests for setting quantities from the command line and in config files. For the CLI, you have to implement `from_string` I think, but probably just passing to `u.Quantity` is enough, maybe that's already the default.

LukasBeiske:
> Also missing are tests for setting quantities from the command line and in config files. For the CLI, you have to implement `from_string` I think, but probably just passing to `u.Quantity` is enough, maybe that's already the default.

I implemented `from_string` at first, but after testing it, it seems that everything works even without it. The test checks, if [setting via a string works](https://github.com/cta-observatory/ctapipe/pull/2524/files#diff-3e5cc46060e228b978f773f2a4be9135e4025fe0100a2832047754d44272137dR294-R297)
Do you think additional test for config files and cmd are still needed? I think this is what happens in both cases, no?

maxnoe:
Would be nice to make sure, can be as simple as:


```
class MyTool(Tool):
    energy = AstroQuantity(physical_type=u.physical.energy).tag(config=True)

tool = MyTool()
run_tool(tool, ["--MyTool.energy='5 TeV'"])
assert tool.energy == 5 * u.TeV
```

and maybe add a test that the error message is nice on invalid input

maxnoe (approved):

src/ctapipe/core/traits.py
```diff
@@ -72,8 +74,48 @@
 flag = traitlets.config.boolean_flag
 
 
+class AstroQuantity(TraitType):
+    """A trait containing an ``astropy.units`` quantity."""
+
+    def __init__(self, physical_type=None, **kwargs):
+        super().__init__(**kwargs)
+        self.physical_type = physical_type
```
Tobychev, line 82: This should probably do something more nuanced than simply trusting that user input is of type `astropy.unit`...

like 
```
if physical_type and isinstance(physical_type,u.unit):
     self.physical_type = physical_type
elif not physical_type:
     self.physical_type = physical_type
else:
    raise ValueError("if given physical_type must be of type astropy.unit, was {type(physical_type)}")
```

Tobychev (changes requested):

src/ctapipe/core/traits.py
maxnoe: The expected type is u.PhysiclType not u.Unit.

However, I guess it would be nice to allow unit input and call unit.physical_type in that case

LukasBeiske: Good point. I added support for both `units.PhysicalType` and `units.Unit`.

maxnoe:
@kosack I think this is ready, you requested changes so merger is currently blocked. Could you please check that your requested changes were implemented?

kosack (approved):

PR #2526, opened 2024-03-25
Drop python 3.9

maxnoe:
The drop schedule according to NEP 29 says to drop 3.9 in the first week of April and the newer SPEC https://scientific-python.org/specs/spec-0000/ already dropped it.

So I think we should follow and for 0.21, also drop 3.9.

I used `ruff --select=UP` to check what can be changed and opted for now not converting every `.format` to an `f-string` and also didn't remove explicit `encoding='utf-8'` arguments as "explicit is better than implicit":

```
ruff --select=UP --ignore=UP032,UP012
```

### DIFF
--- a/src/ctapipe/instrument/camera/geometry.py
+++ b/src/ctapipe/instrument/camera/geometry.py
@@ -672,7 +672,7 @@ class CameraGeometry:
 
         version = tab.meta.get("TAB_VER")
         if version not in cls.SUPPORTED_TAB_VERSIONS:
-            raise IOError(f"Unsupported camera geometry table version: {version}")
+            raise OSError(f"Unsupported camera geometry table version: {version}")
 
         return cls(
             name=tab.meta.get("CAM_ID", "Unknown"),
--- a/src/ctapipe/tools/tests/test_train.py
+++ b/src/ctapipe/tools/tests/test_train.py
@@ -104,12 +104,12 @@ def test_signal_fraction(tmp_path, gamma_train_clf, proton_train_clf):
             ],
         )
 
-        with open(log_file, "r") as f:
+        with open(log_file) as f:
             log = f.readlines()
 
         for line in log[::-1]:
             if "Train on" in line:
-                n_signal, n_background = [int(line.split(" ")[i]) for i in (7, 10)]
+                n_signal, n_background = (int(line.split(" ")[i]) for i in (7, 10))
                 break
 
         assert np.allclose(n_signal / (n_signal + n_background), frac, atol=1e-4)
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -24,7 +24,7 @@ jobs:
           fetch-depth: 0
       - uses: actions/setup-python@v5
         with:
-          python-version: "3.9"
+          python-version: "3.10"
       - name: Check README syntax
         run: |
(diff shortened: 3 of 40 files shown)


src/ctapipe/instrument/camera/geometry.py
```diff
@@ -672,7 +672,7 @@ def from_table(cls, url_or_table, **kwargs):
 
         version = tab.meta.get("TAB_VER")
         if version not in cls.SUPPORTED_TAB_VERSIONS:
-            raise IOError(f"Unsupported camera geometry table version: {version}")
+            raise OSError(f"Unsupported camera geometry table version: {version}")
```
kosack, line 675: I think this is probably not the correct exception type here (and same in readout.py).  _IOError_ was also probably also misused.

maxnoe: True, good catch.

maxnoe: We throw IOError everywhere where we cannot load an input file due to an unsupported version.

We could instead throw:

```
class UnsupportedFileVersion(IOError):
    pass
```

?

I'd do that in another PR though

maxnoe:
Checking the uncovered lines, it is all error conditions which we cannot really check in the tests reasonably.

I deactivated the hard "wait=True" which makes the whole CI fail in case of sonarqube not being happy.

src/ctapipe/instrument/camera/geometry.py
maxnoe: Checking the documentation, `IOError` is indeed a deprecated alias of `OSError` and one should use `OSError` directly, even:

```
>>> IOError("foo")
OSError('foo')
```

src/ctapipe/tools/tests/test_train.py
```diff
@@ -104,12 +104,12 @@ def test_signal_fraction(tmp_path, gamma_train_clf, proton_train_clf):
...
         )
 
-        with open(log_file, "r") as f:
+        with open(log_file) as f:
             log = f.readlines()
 
         for line in log[::-1]:
             if "Train on" in line:
-                n_signal, n_background = [int(line.split(" ")[i]) for i in (7, 10)]
+                n_signal, n_background = (int(line.split(" ")[i]) for i in (7, 10))
```
Tobychev, line 112: changing from list to tuple, not sure that's desired

src/ctapipe/tools/tests/test_train.py
```diff
@@ -104,12 +104,12 @@ def test_signal_fraction(tmp_path, gamma_train_clf, proton_train_clf):
             ],
         )
 
-        with open(log_file, "r") as f:
+        with open(log_file) as f:
```
Tobychev, line 107: prefer it when the `open` is explicit about being a read.

src/ctapipe/tools/tests/test_train.py
maxnoe: That's not changing from a list to a tuple, it's changing from a list comprehension to a generator expression.

I didn't do this manually, it was done by `ruff --select=UP --fix`, and it avoids the intermediate list that is directly unpacked into the two variables

kosack (approved):

Tobychev (approved):

PR #2527, opened 2024-03-25
Fix sonar complaining about statements in notebooks

maxnoe:
Sonarqube complained about statements that "have no effect" in the examples, but these are the output cell values for sphinx-gallery, so not actually a bug


kosack:
> Shall I just include the examples for everything?

probably.. Those aren't really subject to code quality

maxnoe:
@kosack with excluding the `examples/**`, codecov and sonarqube now agree on the coverage value

Tobychev (approved):

kosack (approved):

PR #2529, opened 2024-03-28
Change R1- and DL0-waveforms shape to (n_channels, n_pixels, n_samples)

### LINKED ISSUE #1836: Re-introduce possibility of running `ImageExtractor` on waveforms with multiple gains

Hckjs:
This PR basically adds three (somehow dependent) features:

1) Changing the R1 and DL0 `waveforms` shape  to be always `(n_channels, n_pixels, n_samples)`.

2) Implementing 1) in such a general way (and two lines of code) allows the `ImageExtractors` to handle not gain selected data (by passing `selected_gain_channel=None`)

3) To test these i've adapted the `WaveformToymodel` to be able to create 2 gain waveforms. All extractors are now tested for different camera types (including different number of gain channels)

Things that probably still need to be discussed:

- I don't think it will be needed, but for now the TwoPassWindowSum does not support for 2 gain channels
- There was already some [discussion](https://github.com/cta-observatory/ctapipe/pull/2522#pullrequestreview-1930456074) about splitting the `DL1CameraContainer` considering the two different ndims of `image` and `peak_time`
- With the current API of the `GainSelector` it cannot distinguish between already gain selected data and data that originally consist of just one gain channel. We may pass something like `camera.readout.n_channels`?

Closes #1836 

Thanks already to @TjarkMiener for the discussion and input!


maxnoe:
I guess with this we break the reading-in of R0, R1 and DL0 data written into HDF5 files? I.e. this is a data model change that requires changes in the compatibility table of the `HDF5EventSource`?

TjarkMiener:
> I guess with this we break the reading-in of R0, R1 and DL0 data written into HDF5 files? I.e. this is a data model change that requires changes in the compatibility table of the `HDF5EventSource`?

At the moment (lst-cluster down), I can only test it with the LST-1 HDF5 files of R1 calibration data, which comes with two gains. This is working fine for me with the `HDF5EventSource`. I only have an (unrelated) issue with the coordinate frame when filling the array pointing in the `HDF5EventSource`. I will have a look at the DL0 LST-1 ACADA data once the lst-cluster is running.  
 ```
File "/Users/tjarkmiener/calibration/2529/ctapipe/src/ctapipe/io/hdf5eventsource.py", line 632, in _generator
  self._fill_array_pointing(data)
 File "/Users/tjarkmiener/calibration/2529/ctapipe/src/ctapipe/io/hdf5eventsource.py", line 716, in _fill_array_pointing
  raise ValueError(f"Unsupported pointing frame: {frame}")
ValueError: Unsupported pointing frame: CoordinateFrameType.UNKNOWN
```

**Comment by maxnoe** [comment]
> At the moment (lst-cluster down), I can only test it with the LST-1 HDF5 files of R1 calibration data, which comes with two gains. This is working fine for me with the HDF5EventSource. 

That case will work fine, the issue will be reading single-gain data written with older versions which will result in the 2d shape that is written into the file not the 3d shape expected now

**Comment by maxnoe** [comment]
> I will have a look at the DL0 LST-1 ACADA data once the lst-cluster is running.

I don't understand the relevance of DL0 ACADA (zfits) data in the discussion about the HDF5 event source...

**Comment by TjarkMiener** [comment]
> That case will work fine, the issue will be reading single-gain data written with older versions which will result in the 2d shape that is written into the file not the 3d shape expected now

Yes, sorry! This is what I meant. I accidentally thought that this applies to DL0 ACADA data. However, as you said it's stored (of course) in zfits and not HDF5!

**Comment by mexanick** [comment]
> That case will work fine, the issue will be reading single-gain data written with older versions which will result in the 2d shape that is written into the file not the 3d shape expected now

Is there a good reason to care about backward compatibility here? I mean, do you want a workaround to be implemented to remain compatible with files written by previous versions or just update the `is_compatible` method of `HDF5EventSource` to provide a correct compatibility info?

**Comment by TjarkMiener** [comment]
I have tested it with an old LST-1 sim file in hdf5, where the R1 waveform data is stored with 2 dims, i.e. shape of waveform: (1855,40). In case you would like to have backward compatibility, I did the following:
I expanded the dimension [here](https://github.com/cta-observatory/ctapipe/blob/a2edf56d12a25eef4641b25e832873b9ea7edab0/src/ctapipe/io/hdf5eventsource.py#L645) of the waveforms:
```
                    data.r1.tel[tel_id] = next(waveform_readers[key])
                    if data.r1.tel[tel_id]["waveform"].ndim < 3:
                        data.r1.tel[tel_id]["waveform"] = np.expand_dims(data.r1.tel[tel_id]["waveform"], axis=0)   
```
Since the `r1.pixel_status` is None, I introduced a check [here](https://github.com/cta-observatory/ctapipe/blob/a2edf56d12a25eef4641b25e832873b9ea7edab0/src/ctapipe/calib/camera/calibrator.py#L193-L197)
```
         dl0_pixel_status = None
         if r1.pixel_status is not None:
             dl0_pixel_status = r1.pixel_status.copy()
             # set dvr pixel bit in pixel_status for pixels kept by DVR
             dl0_pixel_status[signal_pixels] |= PixelStatus.DVR_STORED_AS_SIGNAL
             # unset dvr bits for removed pixels
             dl0_pixel_status[~signal_pixels] &= ~np.uint8(PixelStatus.DVR_STATUS)
```
I'm having some trouble with `self.reference_location` (this is "None" with my file) in `SubarrayDescription` which I don't know how to solve. Any ideas?
I hope it can be useful!

**Comment by maxnoe** [comment]
> Is there a good reason to care about backward compatibility here? I mean, do you want a workaround to be implemented to remain compatible with files written by previous versions or just update the is_compatible method of HDF5EventSource to provide a correct compatibility info?

I don't have a strong opinion here, but we need to do one of two things:

* Raise a nice error in case we encounter old data we do not support anymore
* Add a little bit of code to reshape the 2d data into the now expected 3d shape

**Comment by TjarkMiener** [comment]
> * Raise a nice error in case we encounter old data we do not support anymore

I'm in favor of raising a nice error. I don't think there are many (and crucial) R1 or DL0 datasets produced in HDF5 following the 2d waveform shape. Most users are only storing DL1b and/or DL1a. (I actually have a R1 production of LST-1 sims for DL studies. However, it can be easily reprocessed with this PR!)

**Comment by maxnoe** [comment]
The docs failure is due to an unclosed tables file, I think the one that is opened on line 222 of `examples/core/table_writer_reader.py`.

Either you close the file, or you add an exception for that warning into the docs conf

File: src/ctapipe/image/extractor.py [hunk 1]
```diff
@@ -561,26 +576,34 @@ def __call__(
...
-            charge *= self._calculate_correction(tel_id=tel_id)[selected_gain_channel]
-        return DL1CameraContainer(image=charge, peak_time=peak_time, is_valid=True)
+            correction = self._calculate_correction(tel_id=tel_id)
+            if selected_gain_channel is None:
+                charge *= correction[:, np.newaxis]
+            else:
+                charge *= correction[selected_gain_channel]
+        return DL1CameraContainer(
+            image=np.squeeze(charge), peak_time=np.squeeze(peak_time), is_valid=True
+        )
```
kosack, line 606: _Sonarqube_ is complaining that this block is duplicated verbatim in about 4 places. Could we maybe move it into the parent class? Or maybe refactor it into a common a helper function?  That would reduce potential errors if somebody fixes a bug later.

kosack (changes requested):

src/ctapipe/io/hdf5eventsource.py
```diff
@@ -70,9 +70,7 @@
 
 
 COMPATIBLE_DATA_MODEL_VERSIONS = [
-    "v4.0.0",
-    "v5.0.0",
-    "v5.1.0",
+    "v6.0.0",
```
kosack, line 73: Since this only affects R1 and DL0 data stored in hdf5 files, I wonder if it is too harsh to make old versions completely unreadable here? Writing that level of data is more of an exception than the default case.   Of course we will likely re-process everything soon, so maybe it's not a huge issue.  However, maybe an interim solution would be to not drop support for old versions, but simply raise an exception if somebody tries to read files that have R0 or DL0 data in from an old version number, but proceed as usual for DL1 and beyond? 

I.e. right now we have a strict data version that is checked, but maybe we should check the version per data level or at least for R0/DL0 vs higher data levels?

maxnoe: Yes, I think especially in this case, since the compatibility layer is extremely simple (just check ndim==2 and then do `waveform[np.newaxis, ...]` we should make the effort and keep compatibility to the older DL1+ versions at least.

src/ctapipe/calib/camera/calibrator.py
```diff
@@ -31,7 +31,10 @@ def _get_invalid_pixels(n_pixels, pixel_status, selected_gain_channel):
     )
     for mask in masks:
         if mask is not None:
-            broken_pixels |= mask[selected_gain_channel, index]
+            if selected_gain_channel is not None:
+                broken_pixels |= mask[selected_gain_channel, index]
+            else:
+                broken_pixels |= mask
```
maxnoe, line 37: You also need to change the shape of `broken_pixels` to be `(n_channels, n_pixels)`, since pixels might be broken only in one channel (e.g. LST channels of the same pixel are on different DAQ hardware boards)

Hckjs: Should the `InvalidPixelHandler` then also be able to handle multiple gain channels?

src/ctapipe/calib/camera/calibrator.py
```diff
@@ -265,10 +271,12 @@ def _calibrate_dl1(self, event, tel_id):
 
             # correct non-integer remainder of the shift if given
             if self.apply_peak_time_shift.tel[tel_id] and time_shift is not None:
-                dl1.peak_time -= remaining_shift
+                dl1.peak_time = (dl1.peak_time.T - remaining_shift).T
```
maxnoe, line 274: Why was this necessary? 

This also looks like it could be better with a `np.newaxis` instead of transpose and transposing back

src/ctapipe/image/invalid_pixels.py
```diff
@@ -75,31 +78,35 @@ def __call__(self, tel_id, image, peak_time, pixel_mask):
...
-        peak_time = peak_time.copy()
-        peak_time[pixel_mask] = peak_time_sum
-
-        image = image.copy()
-        image[pixel_mask] = image_sum
-
-        return image, peak_time
+        image = np.atleast_2d(image)
+        peak_time = np.atleast_2d(peak_time)
+        for ichannel in range(image.shape[-2]):
```
maxnoe, line 87: ichannel -> channel

src/ctapipe/calib/camera/calibrator.py
```diff
@@ -227,8 +231,10 @@ def _calibrate_dl1(self, event, tel_id):
         # subtract any remaining pedestal before extraction
         if dl1_calib.pedestal_offset is not None:
             # this copies intentionally, we don't want to modify the dl0 data
-            # waveforms have shape (n_pixel, n_samples), pedestals (n_pixels, )
-            waveforms = waveforms - dl1_calib.pedestal_offset[:, np.newaxis]
```
maxnoe, line 231: Shouldn't the pedestal have shape `(n_channels, n_pixels)`? We certainly don't expect the pedestal to be the same for different signal paths.

src/ctapipe/image/tests/test_sliding_window_correction.py
```diff
@@ -1,62 +0,0 @@
...
-from numpy.testing import assert_allclose
-from traitlets.config.loader import Config
-
-from ctapipe.containers import DL1CameraContainer
-from ctapipe.image.extractor import ImageExtractor, SlidingWindowMaxSum
-from ctapipe.image.toymodel import WaveformModel
-from ctapipe.instrument import SubarrayDescription
-
-
-def test_sw_pulse_lst(prod5_lst, reference_location):
```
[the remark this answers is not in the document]
Hckjs: All usecases are (were) already covered by the tests in `test_extractors`.  I dont really know why there was a seperate file for just this specific extractor

src/ctapipe/calib/camera/calibrator.py
Hckjs: I cant actually remember why i used this, i think it was due to a not needed usecase  - peak_time and remaining_shift should have the same shapes so the change isnt needed. I updated it.

maxnoe:
The duplicated lines are in code we just "touched" not actually new duplicated code. 

Since it is test code, I'd rather not do a large refactoring here that might change the tests inadvertently

src/ctapipe/image/extractor.py
```diff
@@ -351,6 +359,15 @@ def integration_correction(
     return correction
 
 
+def _apply_correction(charge, correction, selected_gain_channel):
+    """
+    Helper function for applying the integration correction for certain `ImageExtractor`s.
+    """
+    if selected_gain_channel is None:
+        return (charge * correction[:, np.newaxis]).astype(charge.dtype)
+    return (charge * correction[selected_gain_channel]).astype(charge.dtype)
```
maxnoe, line 368: Doesn't this line need the `pixel_index`?

maxnoe (approved):
I'll approve here, noting that I contributed to the last couple of commits, so please
also have a look @kosack @Tobychev

src/ctapipe/calib/camera/calibrator.py
```diff
@@ -211,24 +215,27 @@ def _calibrate_dl1(self, event, tel_id):
...
             n_pixels,
             event.mon.tel[tel_id].pixel_status,
             selected_gain_channel,
         )
+        pixel_index = np.arange(n_pixels)
 
         dl1_calib = event.calibration.tel[tel_id].dl1
-        time_shift = event.calibration.tel[tel_id].dl1.time_shift
+        ped_offset = dl1_calib.pedestal_offset
+        time_shift = dl1_calib.time_shift
```
TjarkMiener, line 231: I'd stay consistent on how to access the data of the dl1 calibration container, i.e. always use `dl1_calib.pedestal_offset` and `dl1_calib.time_shift` as you do it later the relative and absolute factor.

maxnoe: Using the full name everywhere is running into line length limitations of the formatter att several points and results in awkward multi-line statements.

I think the current version is more readable.

src/ctapipe/image/tests/test_extractor.py
```diff
@@ -603,7 +678,7 @@ def test_waveform_extractor_factory_args(subarray):
 
 def test_extractor_tel_param(toymodel):
     waveforms, subarray, _, _, _, _ = toymodel
-    _, n_samples = waveforms.shape
+    n_samples = waveforms.shape[-1]
```
Tobychev, line 681: Prefer the form `_, _, n_samples =` for consistency with L190, L693, and L714 (among others).

src/ctapipe/calib/camera/calibrator.py
```diff
@@ -264,11 +282,22 @@ def _calibrate_dl1(self, event, tel_id):
             )
 
             # correct non-integer remainder of the shift if given
-            if self.apply_peak_time_shift.tel[tel_id] and time_shift is not None:
+            if self.apply_peak_time_shift.tel[tel_id] and remaining_shift is not None:
```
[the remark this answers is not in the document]
maxnoe: It's update in the `if self.waveform_timeshift` branch. It is the remaining, non-integer time difference after shifting the waveform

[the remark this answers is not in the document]
maxnoe: Line 269

src/ctapipe/image/extractor.py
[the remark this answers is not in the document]
maxnoe: Because in case of gain selection already happend, charge will be 1d and we only need to apply the correction for the gain channel that was used.

src/ctapipe/image/extractor.py
```diff
@@ -424,6 +444,12 @@ def __call__(
         charge, peak_time = extract_around_peak(
             waveforms, 0, waveforms.shape[-1], 0, self.sampling_rate_ghz[tel_id]
         )
+
+        # reduce dimensions for gain selected data to (n_pixels, )
+        if selected_gain_channel is not None:
```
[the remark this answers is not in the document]
maxnoe: In case of gain selected data, you only have a single image, but due to the changes made here to support also non gain selected data, it will have shape `(1, n_pixels)`, we throw away that outer dimension to stay backwards compatible (1d array for dl1 images in case of gain selection).

src/ctapipe/calib/camera/tests/test_gainselection.py
```diff
@@ -23,13 +24,14 @@ def test_gain_selector():
     np.testing.assert_equal(selected_gain_channel, 0)
 
 
-def test_pre_selected():
-    shape = (2048, 128)
-    waveforms = np.zeros(shape)
+def test_ndim():
+    shape = (5, 2, 2048, 128)
+    waveforms = np.indices(shape)[1]
+    waveforms[1] *= 2
```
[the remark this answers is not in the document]
Hckjs: Just copied from the test above but you are right, ill change it

src/ctapipe/image/tests/test_extractor.py
```diff
@@ -386,18 +433,29 @@ def test_integration_correction_off(Extractor, toymodel):
...
+
+    dl1 = extractor(waveforms, tel_id, selected_gain_channel, broken_pixels)
     assert dl1.is_valid
 
     # peak time should stay the same
-    assert_allclose(dl1.peak_time, true_time, rtol=0.1)
-
-    # charge should be too small without correction
-    assert np.all(dl1.image <= true_charge)
+    # charge should be too small without correction for the used reference pulse
```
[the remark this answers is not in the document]
Hckjs: Why? it also checks this condition in  Line 457 (just also for multiple gain images). Below the comment it checks 1) that charge should be too small and 2) peak_time should stay the same - just for the cases that image has different shapes. So it references the whole if-clause, doesnt it?

src/ctapipe/image/tests/test_extractor.py
```diff
@@ -386,18 +433,29 @@ def test_integration_correction_off(Extractor, toymodel):
...
 
+    if Extractor is TwoPassWindowSum and waveforms.shape[-3] != 1:
+        with pytest.raises(AttributeError):
+            extractor(waveforms, tel_id, selected_gain_channel, broken_pixels)
+        return
+
+    dl1 = extractor(waveforms, tel_id, selected_gain_channel, broken_pixels)
     assert dl1.is_valid
 
     # peak time should stay the same
```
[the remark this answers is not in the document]
Hckjs: same as above

src/ctapipe/image/tests/test_extractor.py
maxnoe: Why unpack the tuple if you only need just one element?

mexanick (approved):

mexanick:
@kosack @maxnoe @Tobychev is there anything apart from the issue with the documentation that holds this PR?

maxnoe:
Not from my side, docs are fixed now on main, please rebase

kosack (approved):

maxnoe (approved):

PR #2534, opened 2024-04-16
Use macos 14 (apple arm) in CI

maxnoe:
The macos-latest (currently macos 12 on x86_64 for us, but being transitioned to macos 14) is giving transient failures relatively often that seem unrelated to anything we do here.

Trying to switch to the macos 14 image that's using apple arm.

See: https://github.com/actions/runner-images?tab=readme-ov-file#available-images

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -59,11 +59,11 @@ jobs:
             python-version: "3.12"
             install-method: pip
 
-          - os: macos-latest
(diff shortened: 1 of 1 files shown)


maxnoe:
pytables doesn't seem to offer an arm wheel, installation from source is failing due to missing system dependencies.

Tobychev (approved):

kosack (approved):

PR #2539, opened 2024-04-23
Adding a Reconstruction Plotter

kosack:
Added a simple HillasRecoPlotter Component for debugging or making nice diagrams.

image
So far it only plots the ellipses.   

Todo:
- [x] optionally retain previous points 
- [x] display true points of origin and impact
- [ ] display reco points of origin and impact
- [ ] mark which telescopes were actually used in reconstruction (gray out others) 
- [ ] overplot  image axes on sky plot
- [ ] optionally display images on one or both plots (only really works well with small subarrays)

```py3
  with EventSource(filename, max_events=1000) as source:
        calib = CameraCalibrator(subarray=source.subarray)
        process_images = ImageProcessor(subarray=source.subarray)
        process_shower = ShowerProcessor(subarray=source.subarray)
        display = HillasRecoDisplay(subarray=source.subarray, trace_points=True, figsize=(10,6))

        for event in source:
            calib(event)
            process_images(event)
            process_shower(event)
            display(event)
```

### DIFF
--- a/src/ctapipe/visualization/__init__.py
+++ b/src/ctapipe/visualization/__init__.py
@@ -4,10 +4,11 @@ Visualization: Methods for displaying data
 """
 
 try:
+    from .hillas_reco_display import HillasRecoDisplay
     from .mpl_array import ArrayDisplay
     from .mpl_camera import CameraDisplay
 except ImportError:
     pass
 
 
-__all__ = ["CameraDisplay", "ArrayDisplay"]
+__all__ = ["CameraDisplay", "ArrayDisplay", "HillasRecoDisplay"]
--- a/src/ctapipe/visualization/__init__.py
+++ b/src/ctapipe/visualization/__init__.py
@@ -7,7 +7,8 @@ try:
     from .hillas_reco_display import HillasRecoDisplay
     from .mpl_array import ArrayDisplay
     from .mpl_camera import CameraDisplay
-except ImportError:
+except ImportError as err:
+    print(err)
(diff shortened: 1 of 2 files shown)


maxnoe:
Seems the arrows are in the wrong direction on the ground. What do you use to plot them?

maxnoe:
I also often though about a general display component that can be run inside the tool.

This could be one tab of it. Others for lower level data.

maxnoe:
E.g. instead of `ctapipe-display-dl1` we could just use `ctapipe-process --display`

kosack:
> I also often though about a general display component that can be run inside the tool.
> This could be one tab of it. Others for lower level data.

yes, this was my original idea as well. WOuld need to use Qt or something for the UI, or perhaps just going for bokeh would be nicer, as that is quite a bit faster.

kosack:
> Seems the arrows are in the wrong direction on the ground. What do you use to plot them?

It's using the `set_hillas_vector()` function, which is a bit nicer than the line method added later as it can be updated simply by calling a second time.  I may just have the wrong sign for the asymmetry, I didn't check: though it seems a bit random.  I also don't really lke how it looks, and will probably modify ArrayDisplay to have the vectors in extend in both directions, with an arrow pointing along the gradient.  That  way it works even when the assymmetry is incorrect.

kosack:
This is very much a work in progress, or more of a proof-of-concept that I was using to check some algorithms

PR #2541, opened 2024-04-23
Implement lstchains image cleaning

Hckjs:
Changing the `ImageCleaner`s API in #2511 now allows to make use of monitoring information in the cleaning process.

This enables implementing lstchains image cleaning procedure from its [r0_to_dl1 script](https://github.com/cta-observatory/cta-lstchain/blob/main/lstchain/reco/r0_to_dl1.py#L219) and [dl1ab script](https://github.com/cta-observatory/cta-lstchain/blob/main/lstchain/scripts/lstchain_dl1ab.py) including the 'pedestal cleaning' method.


src/ctapipe/image/cleaning.py
```diff
@@ -112,6 +112,41 @@ def tailcuts_clean(
         )
 
 
+def bright_cleaning(image, threshold=267, fraction=0.03):
```
maxnoe, line 115: this is a weirdly precise number for a default argument

src/ctapipe/image/cleaning.py
```diff
@@ -112,6 +112,41 @@ def tailcuts_clean(
...
+    fraction: `float`
+        Pixels below fraction * (average charge in the 3 brightest pixels)
+        will be removed from the cleaned image
+
+    Returns
+    -------
+    A boolean mask of *clean* pixels.
+
+    """
+    max_3_value_index = np.argsort(image)[-3:]
```
maxnoe, line 138: Sorting the whole array to find the n largest values is not very efficient. 

Maybe not for now, but it could be implemented e.g. by a numba function using a priority queue with a maxsize.

maxnoe: that function could also be used in the GlobalPeakWindowSum which also looks for "brightest n pixels"

src/ctapipe/image/cleaning.py
Hckjs: for now i just used the default values from lstchain

src/ctapipe/image/cleaning.py
```diff
@@ -112,6 +112,41 @@ def tailcuts_clean(
         )
 
 
+def bright_cleaning(image, threshold, fraction):
+    """
+    Clean an image by removing pixels below
```
maxnoe, line 117: First line should be a one-sentence summary.

src/ctapipe/image/cleaning.py
```diff
@@ -459,6 +494,137 @@ def time_constrained_clean(
...
+        Pedestal standard deviation for each pixel. See
+        `ctapipe.containers.PedestalContainer`
+
+    Returns
+    -------
+    A boolean mask of *clean* pixels.
+
+    """
+    # Step 1
+    if apply_pedestal_cleaning and pedestal_std is not None:
```
maxnoe, line 587: Why do we need two arguments if `None` for `pedestal_std` is used to activate/deactivate this?

src/ctapipe/image/cleaning.py
```diff
@@ -459,6 +494,137 @@ def time_constrained_clean(
...
+        boundary_thresh=boundary_thresh,
+        min_number_picture_neighbors=min_number_picture_neighbors,
+        keep_isolated_pixels=keep_isolated_pixels,
+    )
+    # Check that at least one pixel survives tailcuts_clean
+    if np.count_nonzero(mask) == 0:
+        return mask
+
+    # Step 3
+    if apply_time_cleaning:
```
maxnoe, line 605: Same here, could just set `time_limit` to `None`

src/ctapipe/image/cleaning.py
maxnoe: Actually, it's two lines of code, sinde numba supports `heapq.nlargest`:

```
In [17]: from heapq import nlargest
    ...: from numba import njit
    ...: @njit
    ...: def n_largest(n, image):
    ...:     return nlargest(n, image)
    ...: 


In [18]: %timeit image[np.argsort(image)[-3:]]
82.5 µs ± 2.83 µs per loop (mean ± std. dev. of 7 runs, 10,000 loops each)

In [20]: %timeit n_largest(3, image)
3 µs ± 89.5 ns per loop (mean ± std. dev. of 7 runs, 100,000 loops each)
```

maxnoe: I'd say we should directly go for it given that speed up ratio for this amount of code...

src/ctapipe/image/cleaning.py
Hckjs: Can we also set the FloatTelescopeParameter to None?

maxnoe: Yes. pass `allow_none=True`

src/ctapipe/image/cleaning.py
Hckjs: should the function be put in `morphology` ?

maxnoe: This is very generic, so I put it in statistics or utils

src/ctapipe/image/cleaning.py
```diff
@@ -459,6 +494,129 @@ def time_constrained_clean(
     return mask_core | mask_boundary
 
 
+def lst_image_cleaning(
```
kosack, line 497: General comment: can we name this something more descriptive than "LST" image cleaning? I would assume it is applicable to non-LST telescopes as well, so this name is confusing.

src/ctapipe/image/cleaning.py
```diff
@@ -544,6 +702,85 @@ def __call__(
         )
 
 
+class LSTImageCleaner(TailcutsImageCleaner):
```
kosack, line 705: Same comment here: a less generic name would be useful, and also please add a refererence to the LST prototype performance paper in the docstring so it's clear where it came from.

src/ctapipe/image/cleaning.py
maxnoe: I would have said so as well but all names I came up with  were extremely unwieldy, like (one of the shorter ones)

```
pedestal_tailcuts_with_time_cleaning
```

src/ctapipe/image/cleaning.py
```diff
@@ -459,6 +494,129 @@ def time_constrained_clean(
...
+    geom,
+    image,
+    arrival_times,
+    picture_thresh=8,
+    boundary_thresh=4,
+    min_number_picture_neighbors=2,
+    keep_isolated_pixels=False,
+    time_limit=2,
+    time_num_neighbors=1,
+    bright_cleaning_fraction=0.03,
```
maxnoe, line 507: I'd set these "advanced" options to `None` by default

src/ctapipe/image/cleaning.py
Hckjs: time for an [acronym](https://github.com/bacook17/acronym)?
```
=====================================================
(   50) PELTATED  : PEdestaL TAilcuts with TimE Delta
(   43) PATASHTE  : PedestAl TAilcutS witH Time dElta
(   41) PALAMEDEA : PedestAL tAilcuts with tiME DEltA
(   40) PLASTID   : PedestaL tAilcutS with TIme Delta
(   40) PELTATE   : PEdestaL TAilcuts with Time dElta
(   39) PESTICIDE : PEdeSTal taIlCuts with tIme DElta
(   38) PLAITED   : PedestaL tAIlcuts with TimE Delta
(   38) PALATED   : PedestAL tAilcuts with TimE Delta
(   38) PLATTED   : PedestaL tAilcuTs with TimE Delta
(   36) PEETWEET  : PEdEsTal tailcuts With timE dElTa
(   36) PEDATED   : PEDestAl tailcuts with TimE Delta
(   36) PETALITE  : PEdesTAl taiLcuts wIth Time dElta
(   36) PETALLED  : PEdesTAL taiLcuts with timE Delta
(   35) PETTED    : PEdesTal tailcuts with TimE Delta
(   35) PLATED    : PedestaL tAilcuts with TimE Delta
(   33) PASTED    : PedestAl tailcutS with TimE Delta
(   33) PALSIED   : PedestAL tailcutS with tImE Delta
(   33) PATTEE    : PedestAl Tailcuts with TimE dElta
(   33) PETTISH   : PEdesTal TaIlcutS witH time delta
(   33) PELTAST   : PEdestaL TAilcutS with time delTa
(   33) PLUSHED   : PedestaL tailcUtS witH timE Delta
(   33) PEELITE   : PEdEstaL tailcuts wIth Time dElta
(   33) PALAITE   : PedestAL tAilcuts wIth Time dElta
(   33) PATHED    : PedestAl Tailcuts witH timE Delta
(   31) PITTED    : Pedestal taIlcuTs with TimE Delta
(   31) PETALED   : PEdesTAl taiLcuts with timE Delta
(   31) PITSIDE   : Pedestal taIlcuTS with tIme DElta
(   31) PALACED   : PedestAL tAilCuts with timE Delta
(   30) PETITE    : PEdesTal tailcuts wIth Time dElta
(   30) PELITE    : PEdestaL tailcuts wIth Time dElta
```
pesticide_cleaning?

kosack: I'm not so sure you can pick letters in the middle pf a word for acronyms... 

But I'm ok with a longish name.  If I understand correctly this is tailcuts but using pedestal variance thresholds rather than fixed PE thresholds (which is nearly identical to what was used for the Whipple-10m since the early 1990s, though called _picture-boundary cleaning_ there) + it adds the time delta information.   I think in HESS we called this "NSB Cleaning" since it adapts to the the NSB level, but maybe there's a nicer name.

src/ctapipe/image/cleaning.py
Hckjs: I cant find a good way (easy to read and w/o loop) to implement this for the `GlobalPeakWindowSum`, since it needs the indices of the brightest pixels for each channel. `numpy.argpatition` would make it faster anyway

src/ctapipe/image/cleaning.py
Hckjs: i'll name it `NSBImageCleaner` and `nsb_image_cleaning` for now - i really have no idea for a better name. Maybe someone will come up with a better one.

src/ctapipe/image/cleaning.py
Hckjs:
```
n_pixels = 10
def test_argsort(waveforms, broken_pixels):
    brightest = np.argsort(
        waveforms.max(
            axis=-1, where=~broken_pixels[..., np.newaxis], initial=-np.inf
        )
    )[..., -n_pixels:]

    peak_index = (
        waveforms[:, brightest][:, 0, ...].mean(axis=-2).argmax(axis=-1)
    )
    return peak_index

def test_n_largest(waveforms, broken_pixels):
    n_channels = waveforms.shape[0]
    brightest = np.zeros((n_channels, n_pixels), dtype=np.int32)
    wv_max = waveforms.max(
                axis=-1, where=~broken_pixels[:, np.newaxis], initial=-np.inf
            )
    for ichannel in range(n_channels):
        _, brightest[ichannel] = zip(*n_largest(n_pixels, list(zip(wv_max[ichannel], range(wv_max.shape[-1])))))

    peak_index = (
        waveforms[:, brightest][:, 0, ...].mean(axis=-2).argmax(axis=-1)
    )
    return peak_index

%timeit test_argsort(waveforms, broken_pixels)
341 µs ± 796 ns per loop (mean ± std. dev. of 7 runs, 1,000 loops each)

%timeit test_n_largest(waveforms, broken_pixels)
47.6 ms ± 516 µs per loop (mean ± std. dev. of 7 runs, 10 loops each)
```

src/ctapipe/image/tests/test_statistics.py
```diff
@@ -60,3 +60,13 @@ def test_return_type():
 
     stats = descriptive_statistics(data, container_class=PeakTimeStatisticsContainer)
     assert isinstance(stats, PeakTimeStatisticsContainer)
+
+
+def test_nlargest():
+    from ctapipe.image.statistics import n_largest
+
+    image = np.random.rand(1855)
```
maxnoe, line 68: please use the new form of numpy rng :+1: 
```
image = rng.random(1855)
```

src/ctapipe/image/cleaning.py
```diff
@@ -459,6 +496,129 @@ def time_constrained_clean(
...
+        `threshold` parameter for `bright_cleaning`. Minimum average charge
+        in the 3 brightest pixels to apply the cleaning. Set to None if no
+        `bright_cleaning` should be applied.
+    largest_island_only: `bool`
+        Set to true to get only largest island.
+    pedestal_factor: `float`
+        Factor for interleaved pedestal cleaning. It is multiplied by the
+        pedestal standard deviation for each pixel to calculate pixelwise picture
+        threshold parameters for `tailcuts_clean` considering the current background.
+    pedestal_std: `np.ndarray`
```
kosack, line 571: should e `np.ndarray | None`, since it seems None is an acceptable value, and you should mention that if None, the value of `picture_thresh` is used instead.

src/ctapipe/image/cleaning.py
```diff
@@ -459,6 +496,129 @@ def time_constrained_clean(
...
+    Parameters
+    ----------
+    geom: `ctapipe.instrument.CameraGeometry`
+        Camera geometry information
+    image: `np.ndarray`
+        pixel charges
+    arrival_times: `np.ndarray`
+        Pixel timing information
+    picture_thresh: `float` or `np.ndarray`
+        threshold above which all pixels are retained. Used for `tailcuts_clean`.
```
kosack, line 537: You should note that this value is overridden in some cases. Later, if `pedestal_factor` is set, it sets the picture_thresh based on that value and this one.  That shoudld be explained, and maybe change the name so it's clear it's the nominal picture threshold?

src/ctapipe/image/cleaning.py
```diff
@@ -459,6 +496,129 @@ def time_constrained_clean(
...
+
+    Returns
+    -------
+    A boolean mask of *clean* pixels.
+
+    """
+    # Step 1
+    if pedestal_std is not None:
+        pedestal_threshold = pedestal_std * pedestal_factor
+        picture_thresh = np.clip(pedestal_threshold, picture_thresh, None)
```
kosack, line 583: overwriting a parameter of the function is confusing, especially since you use the previous value iin the call to clip... I suggest introducing a new variable

kosack (changes requested):

src/ctapipe/image/cleaning.py
maxnoe: It also shouldn't be in ticks and there needs to be a space before the colon I think...

This reminds me we need to expand the ruff checks to run the documentation checks.

src/ctapipe/image/cleaning.py
kosack: maybe calling the function parameter as follows (or min_picture_threshold if you prefer):
```
picture_threshold_min : float | np.ndarray
    minimum picture threshold. Defines the minimum value used for the picture threshold for tailcust cleaning. 
    The threshold used will be at least this value, or higher if pedestal_std and pedestal_thresh are set.
```

src/ctapipe/image/cleaning.py
```diff
@@ -112,6 +113,41 @@ def tailcuts_clean(
         )
 
 
+def bright_cleaning(image, threshold, fraction):
+    """
+    Clean an image by removing pixels below a fraction of the mean charge
+    in the 3 brightest pixels.
```
maxnoe, line 119: I guess this should be a parameter `n_pixels=3`

src/ctapipe/image/cleaning.py
```diff
@@ -112,6 +113,41 @@ def tailcuts_clean(
...
+
+    Select no pixels instead if the mean charge of the brightest pixels
```
maxnoe, line 121: This is ambiguous. What does "select no pixels" mean? Is the mask all false? Or are no pixels removed?

src/ctapipe/image/cleaning.py
```diff
@@ -112,6 +113,41 @@ def tailcuts_clean(
...
+
+    Returns
+    -------
+    A boolean mask of *clean* pixels.
+
+    """
+    mean_3_max_signal = np.mean(n_largest(3, image))
+
+    if mean_3_max_signal < threshold:
+        return np.zeros(image.shape, dtype=bool)
```
maxnoe, line 143: Shouldn't this return `np.ones` ?  I.e. if we the bright pixels are not bright enough, we don't remove any pixels?

src/ctapipe/image/cleaning.py
```diff
@@ -459,6 +496,129 @@ def time_constrained_clean(
...
+            geom,
+            mask,
+            arrival_times,
+            min_number_neighbors=time_num_neighbors,
+            time_limit=time_limit,
+        )
+
+    # Step 4
+    if bright_cleaning_threshold is not None:
+        mask &= bright_cleaning(
```
maxnoe, line 610: You do an `&` here after returning all zeros in the case of "mean brightness below threshold". Shouldn't it rather not modify the mask in that case?

src/ctapipe/image/cleaning.py
Hckjs: yes you are right

kosack (approved):

TjarkMiener:
Would there be the possibility of patching this lstchain cleaning algo to ctapipe v.19.x in order to already use it on LST-1 data with `ctapipe_io_lst v0.23.2`?

maxnoe:
We don't maintain patch releases at the moment, and starting to do one for new features instead of for security fixes or critical bugs is not something I intend to do.

We should update `ctapipe_io_lst` instead to be compatible with the latest ctapipe.

Issue #2542, opened 2024-04-23
Camera and pointing calibration API


TjarkMiener:
This issue summarizes the suggestion for camera and pointing calibration API discussed in the [ctapipe-dev call (5.4.2024)](https://indico.cta-observatory.org/event/5513/contributions/44982/attachments/25599/37472/20240404_CamCalibAPI_TjarkMiener.pdf) and aims to divide the implementation of the camera and pointing calibration API into sub-tasks, which will be addressed in dedicated PRs.

- [x] Processing (non-gainselected) calibration data using the `process-tool`. #2529 by @Hckjs 
- [x] Add an image extractor for the variance of the waveforms. #2543 @ctoennis 
- [x] Add API for the statistical calculation of pixel-wise values over a given time interval (mean, median, variance, std …), i.e. `StatisticsAggregator`. Different aggregators can be based on numpy/scipy; [astropy sigma clipping](https://docs.astropy.org/en/stable/api/astropy.stats.sigma_clipped_stats.html) #2554 @TjarkMiener 
- [x] Add OutlierDetector components #2604 @TjarkMiener 
- [x] Adjust `StatisticsAggregator` for star tracking purposes @ctoennis 
- [x] Add API for the `CalibrationCalculator`, which holds different subcomponents for the calculation of camera coefficients, e.g. FFactor method. #2609 @TjarkMiener
- [x] Add a pointing correction component using star tracking @ctoennis 
- [x] Implementation of orchestrating tools and workflows (telescope specific tools and routines  in `calibpipe` directly; generic tools in `ctapipe`) #2628 @TjarkMiener @ctoennis 

Related issues:  #1041 #1147 #1836 #1037 #2439

Please comment in case something is missing or we should follow a different path forward @Hckjs @FrancaCassol @mexanick. Discussion on specific sub-tasks can be moved to a dedicated issue.

mexanick:
All planned tasks have been implemented.

Issue #1243, opened 2020-03-24
Extracting pulse width


watsonjj:
The pulse width is potentially a useful parameter to extract from the waveform. A longer pulse width may suggest a time-extended Cherenkov emission, potentially from a shower with heavy progenitors (e.g. iron showers). We may therefore wish to include the extraction of the pulse width as part of the ImageExtractor (much in the same way as charge and pulse time is extracted), and include it in the DL1 container.

The difficulty with this exercise is to do this in an efficient way, so as to not heavily impact the runtime of the ImageExtractor algorithms. This is likely to need the use of the numba package.

kosack:
Yes, this is a good idea. The pulse can be fully defined by peak, width, risetime, and falltime, so it could be useful to compute all of them here (we did this in VERITAS, at least for testing).

note that the width can be defined in many ways as well - is it the width at 50% of the peak?  A few time measurements are useful:

* t_thresh: time the pulse first crosses a user-defined threshold for peak-detection (useful to measuring time jitter)
* t_50 [leading, trailing]: time position at 50% of the pulse on the leading-edge or trailing edge 
* t_90 [leading, trailing]: time position at 90% of pulse for leading and trailing edges
* t_10 [leading, trailing]: time position at 10% of pulse for leading and trailing edges

Then
* width = t_50[1] - t_50[0]. (1=trailing edge, 0=leading edge)
* rise_time = t_90[0] - t_10[0]
* fall_time = t_10[1] - t_90[1]

kosack:
so the main function needed is one that finds the time in the waveform where the value of the waveform crosses a threshold, either forward (from start of search window) or backward (from end of search window). That allows you to calculate all of these times. 

Something like:

```py3
def find_waveform_time(waveform, threshold, forward=True):
   ...

t_50_0 = find_waveform_time(waveform, peak/2.0, forward=True)
t_50_1 = find_waveform_time(waveform, peak/2.0, forward=False)
```
where the return values are the time for each pixel in the image.

watsonjj:
I had a much-less-than-perfect algorithm to extract the FWHM and rise time of the pulse implemented a while ago in CHECLabPy. It uses numba, but it still impacted the event processing speed enough that I avoided using it.

https://github.com/sstcam/CHECLabPy/blob/master/CHECLabPy/waveform_reducers/timing.py

It could be a useful resource though.

kosack:
The simplest method is just this, but not sure how efficient it is:

```py3
[...]

PR #2543, opened 2024-04-24
Variance extractor

ctoennis:
This pull request is to add a new ImageExtractor subclass to generate images from the variance of waveforms that will be needed for the pointing calibration using stars. This request is related to issue #2542 .

### DIFF
--- a/docs/changes/2543.feature.rst
+++ b/docs/changes/2543.feature.rst
@@ -1,2 +1,2 @@
-A new ImageExtractor called ``VarianceExtractor`` was added with a corresponding container
-called ``DL1PedestalVarianceContainer``. These will be used for the startracker code later.
+A new ImageExtractor called ``VarianceExtractor`` was added
+An Enum class was added to containers.py that is used in the metadata of the VarianceExtractor output
--- a/docs/changes/2543.feature.rst
+++ b/docs/changes/2543.feature.rst
@@ -1,0 +1,2 @@
+A new ImageExtractor called ``VarianceExtractor`` was added with a corresponding container
+called ``DL1PedestalVarianceContainer``. These will be used for the startracker code later.
--- a/src/ctapipe/containers.py
+++ b/src/ctapipe/containers.py
@@ -19,6 +19,7 @@ __all__ = [
     "DL0Container",
     "DL1CameraCalibrationContainer",
     "DL1CameraContainer",
+    "DL1PedestalVarianceContainer",
     "DL1Container",
     "DL2Container",
     "EventCalibrationContainer",
@@ -164,6 +165,16 @@ class EventType(enum.Enum):
     SUBARRAY = 32
 
     UNKNOWN = 255
+
+
+class VarianceType(enum.Enum):
+    """Enum of variance types used for the DL1PedestalVarianceContainer
+    """
+
+    #Simple variance of waveform
+    SIMPLE=0
+    #Variance of intgrated samples of a waveform
+    SAMPLE=1
 
 
 class PixelStatus(enum.IntFlag):
@@ -513,6 +524,28 @@ class DL1CameraContainer(Container):
 
     parameters = Field(
         None, description="Image parameters", type=ImageParametersContainer
+    )
+
+
+class DL1PedestalVarianceContainer(Container):
+    """
+    Storage of output of camera variance image e.g. 
+    the variance of each pixel composed as an image.
+    """
+
+    image = Field(
+        None,
+        "Numpy array of camera variance image"
+        "Shape: (n_pixel) if n_channels is 1 or data is gain selected"
+        "else: (n_channels, n_pixel)",
+    )
+
+    VarMethod = Field(
+        VarianceType.SIMPLE,
+        "Method by which the variance was calculated"
+        "This can either be a plain variance"
+        "or a variance of integrated samples",
+        type=VarianceType,
     )
 
 
--- a/src/ctapipe/containers.py
+++ b/src/ctapipe/containers.py
@@ -168,13 +168,12 @@ class EventType(enum.Enum):
 
 
 class VarianceType(enum.Enum):
-    """Enum of variance types used for the DL1PedestalVarianceContainer
-    """
+    """Enum of variance types used for the DL1PedestalVarianceContainer"""
 
-    #Simple variance of waveform
-    SIMPLE=0
-    #Variance of intgrated samples of a waveform
-    SAMPLE=1
+    # Simple variance of waveform
+    SIMPLE = 0
+    # Variance of integrated samples of a waveform
+    SAMPLE = 1
 
 
 class PixelStatus(enum.IntFlag):
@@ -529,7 +528,7 @@ class DL1CameraContainer(Container):
 
 class DL1PedestalVarianceContainer(Container):
     """
-    Storage of output of camera variance image e.g. 
+    Storage of output of camera variance image e.g.
     the variance of each pixel composed as an image.
     """
 
@@ -539,24 +538,24 @@ class DL1PedestalVarianceContainer(Container):
         "Shape: (n_pixel) if n_channels is 1 or data is gain selected"
         "else: (n_channels, n_pixel)",
     )
-
     trigger_time = Field(
         None,
-        "Trigger time for this image"
-        "Will be needed by the startracker code later to determine ",
+        "Trigger time for this variance image" "Value is a float",
     )
-
-    pointing = Field(default_factory=TelescopePointingContainer,
-        description="Telescope pointing for the startracker code",
-        )
-
-
     VarMethod = Field(
         VarianceType.SIMPLE,
         "Method by which the variance was calculated"
         "This can either be a plain variance"
         "or a variance of integrated samples",
         type=VarianceType,
+    )
+    is_valid = Field(
+        False,
+        (
+            "True if image extraction succeeded, False if failed "
+            "or in the case of TwoPass methods, that the first "
+            "pass only was returned."
+        ),
     )
 
 
--- a/src/ctapipe/containers.py
+++ b/src/ctapipe/containers.py
@@ -486,6 +486,17 @@ class ImageParametersContainer(Container):
 
 
 class DL1CameraContainer(Container):
+    """
+    Storage of output of camera calibration e.g the final calibrated
+    image in intensity units and the pulse time.
+    """
+
+    image = Field(
+        None,
+        "Numpy array of camera image, after waveform extraction."
+        "Shape: (n_pixel) if n_channels is 1 or data is gain selected"
+        "else: (n_channels, n_pixel)",
+    )
     peak_time = Field(
         None,
         "Numpy array containing position of the peak of the pulse as determined by "
--- a/src/ctapipe/containers.py
+++ b/src/ctapipe/containers.py
@@ -540,6 +540,17 @@ class DL1PedestalVarianceContainer(Container):
         "else: (n_channels, n_pixel)",
     )
 
+    trigger_time = Field(
+        None,
+        "Trigger time for this image"
+        "Will be needed by the startracker code later to determine ",
+    )
+
+    pointing = Field(default_factory=TelescopePointingContainer,
+        description="Telescope pointing for the startracker code",
+        )
+
+
     VarMethod = Field(
         VarianceType.SIMPLE,
         "Method by which the variance was calculated"
--- a/src/ctapipe/containers.py
+++ b/src/ctapipe/containers.py
@@ -19,7 +19,6 @@ __all__ = [
     "DL0Container",
     "DL1CameraCalibrationContainer",
     "DL1CameraContainer",
-    "DL1PedestalVarianceContainer",
     "DL1Container",
     "DL2Container",
     "EventCalibrationContainer",
@@ -168,12 +167,12 @@ class EventType(enum.Enum):
 
 
 class VarianceType(enum.Enum):
-    """Enum of variance types used for the DL1PedestalVarianceContainer"""
+    """Enum of variance types used for the VarianceContainer"""
 
     # Simple variance of waveform
-    SIMPLE = 0
+    WAVEFORM = 0
     # Variance of integrated samples of a waveform
-    SAMPLE = 1
+    INTEGRATED = 1
 
 
 class PixelStatus(enum.IntFlag):
@@ -487,17 +486,6 @@ class ImageParametersContainer(Container):
 
 
 class DL1CameraContainer(Container):
-    """
-    Storage of output of camera calibration e.g the final calibrated
-    image in intensity units and the pulse time.
-    """
-
-    image = Field(
-        None,
-        "Numpy array of camera image, after waveform extraction."
-        "Shape: (n_pixel) if n_channels is 1 or data is gain selected"
-        "else: (n_channels, n_pixel)",
-    )
     peak_time = Field(
         None,
         "Numpy array containing position of the peak of the pulse as determined by "
@@ -520,42 +508,8 @@ class DL1CameraContainer(Container):
             "pass only was returned."
         ),
     )
-
     parameters = Field(
         None, description="Image parameters", type=ImageParametersContainer
-    )
-
-
-class DL1PedestalVarianceContainer(Container):
-    """
-    Storage of output of camera variance image e.g.
-    the variance of each pixel composed as an image.
-    """
-
-    image = Field(
-        None,
-        "Numpy array of camera variance image"
-        "Shape: (n_pixel) if n_channels is 1 or data is gain selected"
-        "else: (n_channels, n_pixel)",
-    )
-    time = Field(
-        None,
-        "Trigger time for this variance image" "Value is a float",
-    )
-    VarMethod = Field(
-        VarianceType.SIMPLE,
-        "Method by which the variance was calculated"
-        "This can either be a plain variance"
-        "or a variance of integrated samples",
-        type=VarianceType,
-    )
-    is_valid = Field(
-        False,
-        (
-            "True if image extraction succeeded, False if failed "
-            "or in the case of TwoPass methods, that the first "
-            "pass only was returned."
-        ),
     )
 
 
--- a/src/ctapipe/containers.py
+++ b/src/ctapipe/containers.py
@@ -538,7 +538,7 @@ class DL1PedestalVarianceContainer(Container):
         "Shape: (n_pixel) if n_channels is 1 or data is gain selected"
         "else: (n_channels, n_pixel)",
     )
-    trigger_time = Field(
+    time = Field(
         None,
         "Trigger time for this variance image" "Value is a float",
     )
--- a/src/ctapipe/image/extractor.py
+++ b/src/ctapipe/image/extractor.py
@@ -1315,7 +1315,7 @@ class VarianceExtractor(ImageExtractor):
             image=variance,
             VarMethod=VarianceType.SIMPLE,
             is_valid=True,
-            trigger_time=np.float32(trigger_time),
+            time=np.float32(trigger_time),
         )
 
 
--- a/src/ctapipe/image/extractor.py
+++ b/src/ctapipe/image/extractor.py
@@ -1302,8 +1302,7 @@ class VarianceExtractor(ImageExtractor):
 
     """
     Extractor that calculates
-    the variance of each waveform and
-    corrects for the relative gain.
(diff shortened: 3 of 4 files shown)


src/ctapipe/image/extractor.py
```diff
@@ -1295,6 +1296,20 @@ def __call__(
...
+    """
+    Extractor that calculates
+    the variance of each waveform
+    """
+
+    def __call__(
+        self, waveforms, tel_id
+    ) -> DL1CameraContainer:
+        variance = np.nanvar(waveforms,axis=2)
+        return DL1CameraContainer(image=variance, peak_time=0, is_valid=True)
```
TjarkMiener, line 1310: I think the DL1CameraContainer should be adjusted rather than storing the variances where the charge images belong and setting peak_time to zero. This will introduce breaking changes ...

maxnoe: peak_time should be `None`, not 0, which will e.g. make it so that this field is skipped when writing out.

If you need to give a value, it needs to have the correct shape

src/ctapipe/image/extractor.py
```diff
@@ -1295,6 +1296,20 @@ def __call__(
...
+
+    """
+    Extractor that calculates
+    the variance of each waveform
+    """
+
+    def __call__(
+        self, waveforms, tel_id
+    ) -> DL1CameraContainer:
+        variance = np.nanvar(waveforms,axis=2)
```
maxnoe, line 1309: Do we really just want to compute the variance of the hole waveform?

I know pedvar as the the variance over an integration window. 

You place as many integration windows as you can in the waveform and then compute the variance over the integrals.

src/ctapipe/image/extractor.py
mexanick: Do we want to add a metadata field that describes the meaning of the image? Could be an enum of "charge", "variance" etc. with the default value being "charge" for backwards compatibility

src/ctapipe/image/extractor.py
mexanick: that might be interesting, however, I didn't test it... A simple variance like here works for sure. 
@ctoennis I think the idea of Max is nice and we can implement it alongside and compare later

maxnoe: Just to be clear, that's not my idea but comes from the original ped var paper, I think by VERITAS, let me dig it up

kosack:
If I understand correctly, what you are basically doing here is computing an event-wise pedestal variance. In the data model (see the original top-level data model document), there are a few places you can put pedestals:

- _dl1.event.telescope.pedestal_: pedestals measured for each event (usually in a time window as @maxnoe mentioned, since you can for example just use time slices outside the trace integration window, even for Cherenkov events). This is also computed by some cameras directly in hardware as dl0 information.
- _dl1.monitoring.telescope.pedestal_: is the pedestal computed from an ensemble of pedestal events, or from Cherenkov events with the shower cut out  (this is also in the DL0.monitoring.telescope.pedestal, computed by the cameras) computed every few seconds.
- dl1.service.telescope.pedestal: pedestals measured from some standard run that are applied for e.g. RTA

So I think what you are computing is the first one, and you should store it as pedestal information:  It is in fact a second measurement of the pedestal and its variance, and could be compared to the other methods (even if the use case you are describing here is for doing star tracking, star tracking is computed from the pedestal variances, which can be over many time scales)

fwerner:
In addition to what has been said I'd suggest to clarify the purpose of this extractor: it is calculating the waveform _sample_ variance, which should be reflected in the name/docstring to communicate that it has a complicated relationship to charge.

src/ctapipe/image/extractor.py
```diff
@@ -1295,6 +1296,21 @@ def __call__(
         )
 
 
+class VarianceExtractor(ImageExtractor):
+
+    """
+    Extractor that calculates
```
maxnoe, line 1302: First line needs to be a single line summary followed by an empty line, then you can give more details.

I would propose:

```
Compute variance over samples for each waveform.
```

src/ctapipe/image/extractor.py
ctoennis: I will see how well this works compared to just looking at the variance directly.

ctoennis:
The thing we want to calculate is the variance of the waveforms specifically of interleaved pedestal events and not the variance of the pedestal in units of charge. 

@kosack : What we get is a sort of image based on the variance of the waveform in each pixel. What we get is not a pedestal value, so putting it in a pedestal container is not the best option i think. 

I think it is best to write a container appropriate for these variance images similar to the DL1CameraContainer

src/ctapipe/image/extractor.py
mexanick: since the amplitude of the signal in bin N+1 does not depend on the amplitude of the signal in bin N, then the variance of a sample and the variance of charge are equivalent (cov(Ai, Aj) == 0). The only differences are statistical error on the calculated variance and the normalization.

fwerner: I'm not sure I understand the context, so apologies if I misunderstood what you've written - just wanted to say that if by "bin" you mean "time slice" for a given channel: consecutive samples are _not_ independent because the single-p.e. pulse shapes are several nanoseconds long (tens of nanoseconds for some cameras). Also electronics noise is usually not white.

mexanick: You understand it correctly and I was thinking of what you wrote. However we consider NSB radiation, which is commonly modeled as a Poisson process, i.e. the numbers of photons hitting the pixel in two consecutive time intervals are independent. Now, for the PMT-equipped camera the conversion of photons to p.e. will also be uncorrelated. This is not correct for SiPM cameras, where generalized Poisson shall be applied.

ctoennis:
I just added a custom continer for the variance images. I also compared the variance calculation using just a plain variance and a variance of integrated chunks of the waveforms using the FixedWindowSum extractor in the plot below.
comparemethod

The x axis shows the number of images used in calculating an average variance value that is shown on the y axis. 

Though the methods are generally statistically equivalent the average variance becomes stable with fewer images when using a simple variance over the variance from integrated chunks.

maxnoe:
> Though the methods are generally statistically equivalent the average variance becomes stable with fewer images when using a simple variance over the variance from integrated chunks.

That's not surprising, since you have just more samples. But you change *of what you compute the variance*

ctoennis:
Hi all, can we restart the discussion here? I made a small update to the extractor to contain timing information that will later be needed in the corresponding subclass of StatisticsExtractor that i am working on with Tjark. 

There was a decision needed as to where the variance images will be attached to. Should this find a place somewhere in dl1.event.telescope.pedestal or something similar?

src/ctapipe/image/extractor.py
```diff
@@ -1300,8 +1300,7 @@ class VarianceExtractor(ImageExtractor):
 
     """
     Extractor that calculates
-    the variance of each waveform and
-    corrects for the relative gain.
+    the variance of each waveform
```
mexanick: This has to be a _single_ line, really concise. A longer description can follow after one blank line.

ctoennis: sorry about that. i shortened it.

maxnoe:
Hi @ctoennis the lint step in the CI is failing due to some typos and wrong formatting.

To check locallly, make sure to install the pre-commit hooks:
https://ctapipe.readthedocs.io/en/latest/developer-guide/getting-started.html#installing-ctapipe-in-development-mode

ctoennis:
i fixed all the isses with the spelling, code checks etc.

src/ctapipe/containers.py
```diff
@@ -166,6 +167,15 @@ class EventType(enum.Enum):
     UNKNOWN = 255
 
 
+class VarianceType(enum.Enum):
+    """Enum of variance types used for the DL1PedestalVarianceContainer"""
+
+    # Simple variance of waveform
+    SIMPLE = 0
+    # Variance of integrated samples of a waveform
+    SAMPLE = 1
```
maxnoe, line 176: SIMPLE and SAMPLE are extremely similar and `SAMPLE` could also mean "variance over the waveform samples" so what you refer to here as "SIMPLE".

Maybe better `SAMPLE` vs. `INTEGRATED` ?

Also please add a color before the comment: `#: ` which is the special syntax for docstrings for member attributes.

src/ctapipe/containers.py
```diff
@@ -516,6 +526,39 @@ class DL1CameraContainer(Container):
     )
 
 
+class DL1PedestalVarianceContainer(Container):
+    """
+    Storage of output of camera variance image e.g.
+    the variance of each pixel composed as an image.
+    """
+
+    image = Field(
```
maxnoe, line 535: I'd call this `variance`, not `image` if it is allways the variance

src/ctapipe/containers.py
```diff
@@ -516,6 +526,39 @@ class DL1CameraContainer(Container):
...
+        None,
+        "Numpy array of camera variance image"
+        "Shape: (n_pixel) if n_channels is 1 or data is gain selected"
+        "else: (n_channels, n_pixel)",
+    )
+    trigger_time = Field(
```
maxnoe, line 541: Why do we need to store the time here in this container?

src/ctapipe/image/extractor.py
```diff
@@ -1295,6 +1300,23 @@ def __call__(
...
 
 
+class VarianceExtractor(ImageExtractor):
+
+    """
+    Calculates the variance of each sample waveform
+
+    """
+
+    def __call__(self, waveforms, tel_id, trigger_time) -> DL1PedestalVarianceContainer:
```
maxnoe, line 1310: Is this really an image extractor? It doesn't follow the abstract API of `ImageExtractor`, it returns a different data type and it's arguments are also different.

For it to *be* an image extractor, it must be usable if configured through the configuration system.

This fails for example:

```
❯ ctapipe-process -i dataset://gamma_prod5.simtel.zst -o /tmp/test.dl1.h5 --CameraCalibrator.image_extractor_type=VarianceExtractor

2024-07-10 15:28:47,988 WARNING [ctapipe.core.telescope_component] (telescope_component.attach_subarray): TelescopeParameter type argument 'SST_1M_*' did not match any known telescope types
2024-07-10 15:28:49,210 ERROR [ctapipe.ctapipe-process] (tool.run): Caught unexpected exception: VarianceExtractor.__call__() got an unexpected keyword argument 'selected_gain_channel'
  File "/home/mnoethe/Uni/CTA/ctapipe/src/ctapipe/core/tool.py", line 426, in run
    self.start()
  File "/home/mnoethe/Uni/CTA/ctapipe/src/ctapipe/tools/process.py", line 311, in start
    self.calibrate(event)
  File "/home/mnoethe/Uni/CTA/ctapipe/src/ctapipe/calib/camera/calibrator.py", line 329, in __call__
    self._calibrate_dl1(event, tel_id)
  File "/home/mnoethe/Uni/CTA/ctapipe/src/ctapipe/calib/camera/calibrator.py", line 277, in _calibrate_dl1
    dl1 = extractor(
          ^^^^^^^^^^
TypeError: VarianceExtractor.__call__() got an unexpected keyword argument 'selected_gain_channel'
```

ctoennis:
i can remove trigger time. But i am not sure how to proceed on the difference to the ImageExtractor API. I could make it that the variance image is stored in a DL1ImageContainer and that it complies with the API or make it a separate class that doesn't inherit from the ImageExtractor. 

I think what i generate here is a type of image as i try to look for stars in the camera FOV with them. But if it makes more sense to you it can be separate from the ImageExtractors.

maxnoe:
The question is if it is possible to adhere to the API that is expected from the image extractors: if yes, no problem, let it be an image extractor. But that means for example the call above should work and correctly write out the data to the HDF5 file.

If you need the different API, it should be its own thing and we need to think about how to integrate it into the process tool for example.

kosack:
> The question is if it is possible to adhere to the API that is expected from the image extractors: if yes, no problem, let it be an image extractor. But that means for example the call above should work and correctly write out the data to the HDF5 file.

I think what this class does is not the same as an `ImageExtractor`, or am I missing something? `ImageExtractors` take a "movie" (a set of waveforms per pixel for one event) for a single event and convert them to  set of time-independent dependent images that must include the  _peak time_ and _integral_ over a window, but can also optionally inlcude any other parameters of the waveform like the pulse width or variabce.    Generally, there is only one ImageExtractor that is run at a time.    That doesn't seem to be what you have here.

maxnoe:
> I think what this class does is not the same as an ImageExtractor, or am I missing something?

You could think of this as an `ImageExtractor` that uses the variance of each waveform to produce the image and doesn't produce a `peak_time`.

mexanick:
I believe, it is an ImageExtractor, the `__call__()` signature is of course wrong here. It should copy the one of the base extractor:
```
@abstractmethod
    def __call__(
        self, waveforms, tel_id, selected_gain_channel, broken_pixels
    ) -> DL1CameraContainer:
```

Only `waveforms` are then used, the other parameters are simply ignored.

The only issue, that has to be resolved, is the return type. `DL1CameraContainer` in its current shape contains a number of fields that are not needed in this case (actually, everything except `image`). 

I'd propose to add an intermediate container:

```python
class CameraImageContainer(Container):
    """
    Storage of output of generic camera images (any reduction of waveforms).
    """

    image = Field(
        None,
        "Numpy array of camera image, after waveform extraction."
        "Shape: (n_pixel) if n_channels is 1 or data is gain selected"
        "else: (n_channels, n_pixel)",
    )

class DL1CameraContainer(CameraImageContainer):
    """
    Storage of output of camera calibration e.g the final calibrated
    image in intensity units and the pulse time.
    """
    peak_time = Field(
        None,
        "Numpy array containing position of the peak of the pulse as determined by "
        "the extractor."
        "Shape: (n_pixel) if n_channels is 1 or data is gain selected"
        "else: (n_channels, n_pixel)",
    )
    # rest of the DL1CameraContainer fields
    
class VarianceContainer(CameraImageContainer):
    """
    Storage of output of camera variance image.

    Stores the variance of waveform in each pixel, composed as an image.
    """
    variance_method = Field(
        VarianceType.SIMPLE,
        "Method by which the variance was calculated"
        "This can either be a plain variance"
        "or a variance of integrated samples",
        type=VarianceType,
    )
    # I don't think you need anything else here. The trigger time you extract from elsewhere in the event tree.
```
and change the return type of base `ImageExtractor` to `CameraImageContainer`

maxnoe:
If it's just that the `VarianceExtractor` doesn't provide a `peak_time`, I think that's fine, just leave it `None` (the default value).

Adding an additional field to the `DL1CameraContainer` that's an "image_type" sounds like a good idea.

I would avoid having extractors that have different return types.

mexanick:
Following a private discussion with @maxnoe , the new Container class is not needed. 
@ctoennis just use the `DL1CameraContainer` as a return type, setting the extractor type as a container meta keyword. You only need to fill `Image` field and leave all others as defaults (`None`). In this case, the serialization will be done correctly and only the image that you need will be stored. I.e. something like this (assuming you change VarianceType enum to WAVEFORM):

```python
def __call__(
        self, waveforms, tel_id, selected_gain_channel, broken_pixels
    ) -> DL1CameraContainer:
    container = DL1CameraContainer(image=np.nanvar(waveforms, dtype="float32", axis=2))
    container.meta["ExtractionMethod"] = str(VarianceType.WAVEFORM)
    return container
```

maxnoe:
Since this is also somewhat connected to #1243 

I.e. we could introduce a new class e.g. `WaveformProperties` that runs after the `CameraCalibrator` and computes stuff like variance, FWHM, etc.

src/ctapipe/containers.py
```diff
@@ -477,17 +486,6 @@ class ImageParametersContainer(Container):
 
 
 class DL1CameraContainer(Container):
-    """
```
mexanick, line 480: Why this is removed?

docs/changes/2543.feature.rst
```diff
@@ -0,0 +1,2 @@
+A new ImageExtractor called ``VarianceExtractor`` was added with a corresponding container
+called ``DL1PedestalVarianceContainer``. These will be used for the startracker code later.
```
mexanick, line 2: This needs an update

src/ctapipe/containers.py
ctoennis: sorry, that was by accident

src/ctapipe/image/tests/test_extractor.py
```diff
@@ -423,7 +427,7 @@ def test_extractors(Extractor, toymodels, request):
 @pytest.mark.parametrize("Extractor", extractors)
 def test_integration_correction_off(Extractor, toymodels, request):
     # full waveform extractor does not have an integration correction
-    if Extractor is FullWaveformSum:
+    if Extractor is FullWaveformSum or Extractor is VarianceExtractor:
```
maxnoe, line 430:
```suggestion
    if Extractor is in (FullWaveformSum, VarianceExtractor):
```

src/ctapipe/image/tests/test_extractor.py
```diff
@@ -713,7 +717,13 @@ def test_dtype(Extractor, subarray):
     extractor = Extractor(subarray=subarray)
     n_channels, n_pixels, _ = waveforms.shape
     broken_pixels = np.zeros((n_channels, n_pixels), dtype=bool)
+    if Extractor is VarianceExtractor:
```
maxnoe, line 720: I'd remove this here and just skip the `dl1.image.dtype` check in case it's the `VarianceExtractor`

maxnoe (dismissed):

mexanick:
> Since this is also somewhat connected to #1243
> 
> I.e. we could introduce a new class e.g. `WaveformProperties` that runs after the `CameraCalibrator` and computes stuff like variance, FWHM, etc.

While this might be a good idea, I'd save it for a separate refactoring PR and merge the current (small) addition as is.

mexanick (dismissed):

src/ctapipe/image/extractor.py
```diff
@@ -1297,6 +1301,23 @@ def __call__(
         )
 
 
+class VarianceExtractor(ImageExtractor):
+
+    """
+    Calculates the variance of each sample waveform
+
+    """
```
maxnoe, line 1309:
```suggestion
    """Calculate the variance over samples in each waveform."""
```

maxnoe (approved):

maxnoe (dismissed):

ctoennis:
So, is this request then good to merge?

maxnoe:
@ctoennis From my side yes, but we require two approvals and @kosack is on vacation right now. Since he had comments / requests for changes before, I'd like to wait for his approval here.

kosack:
I'll take a look tomorrow or today hopefully! Might take a bit of time since there are a lot of changes.

kosack (changes requested):
Looks good. but I would like to have a test that checks for correctness.  It is a simple algorithm, but the test will also ensure that nobody breaks it in the future. E.g. a test that runs the extractor on some dummy data with a known variance.

ctoennis:
i added a test to check that an actual variance is being calculated.

src/ctapipe/image/tests/test_extractor.py
```diff
@@ -281,6 +282,17 @@ def test_extract_around_peak_charge_expected():
     assert_equal(charge, n_samples)
 
 
+def test_variance_extractor(toymodel):
+    _, subarray, _, _, _, _ = toymodel
+    # make dummy data with known variance
+    rng = np.random.default_rng(0)
+    var_data = rng.normal(2.0, 5.0, size=(5000, 2, 1855))
```
maxnoe, line 289: The shape here is a bit weird for testing an image extractor. Of course for the test this doesn't really matter, but what you are doing here means

n_channels = 5000
n_pixels = 2 
n_samples = 1855

src/ctapipe/image/tests/test_extractor.py
```diff
@@ -281,6 +282,17 @@ def test_extract_around_peak_charge_expected():
...
+    extractor = ImageExtractor.from_name("VarianceExtractor", subarray=subarray)
+
+    variance = extractor(var_data, 0, None, None).image
+    assert np.isclose(variance, np.var(var_data, axis=2)).all()
```
maxnoe, line 293: Use `np.testing.assert_all_close(variance, np.var(...))` for better error messages in case of failure

maxnoe (approved):

PR #2544, opened 2024-04-24
Fix possible out of bounds error in FlashCamExtractor

maxnoe:
The code had a possible out of bounds array access in case of the peak not being fully contained in the waveform.

Should fix a segfault reported by @gschwefer

### DIFF
--- a/src/ctapipe/image/extractor.py
+++ b/src/ctapipe/image/extractor.py
@@ -1531,7 +1531,7 @@ def adaptive_centroid(waveforms, peak_index, rel_descend_limit, centroids):
         sum_ += waveforms[j]
         jsum += j * waveforms[j]
         j -= 1
-        if waveforms[j] > peak_amplitude:
+        if j >= 0 and waveforms[j] > peak_amplitude:
             descend_limit = rel_descend_limit * peak_amplitude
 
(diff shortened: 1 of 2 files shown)


gschwefer (approved):

maxnoe:
macos test is now failing because macos-latest was updated to be macos 14 on M1 for which no pytables wheel exist, let's merge: #2534

src/ctapipe/image/extractor.py
```diff
@@ -1531,15 +1531,15 @@ def adaptive_centroid(waveforms, peak_index, rel_descend_limit, centroids):
         sum_ += waveforms[j]
         jsum += j * waveforms[j]
         j -= 1
-        if waveforms[j] > peak_amplitude:
+        if j > 0 and waveforms[j] > peak_amplitude:
```
maxnoe, line 1534: Should be >=

gschwefer: Yes, you're right

maxnoe: That was more of a note to self that I realized after I opened the PR, but thanks for fixing :+1:

gschwefer (approved):

PR #2547, opened 2024-04-25
Fix CI

maxnoe:
Fix CI by slightly adapting impact test to find a minimum on both macos and linux... not a great fix, but the test needs to be improved some other way.

The new pytest version prints out the full stack trace for xfails in the CI log, which makes it harder to spot the actual error that caused the test failure.

Checking the xfails, we only have these once here, which are not really xfails but really expected, so I removed the xfail.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -42,7 +42,7 @@ dependencies = [
     "psutil",
     "pyyaml >=5.1",
     "requests",
-    "scikit-learn <1.4",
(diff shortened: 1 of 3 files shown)


kosack (approved):

maxnoe:
The macos-13 / macos-14 tests are running *much* faster now than the old macos-latest (macos 12) tests. Nice! thanks github

gschwefer (approved):

PR #2550, opened 2024-04-25
Prepare changelog for 0.21.0

maxnoe:


kosack (dismissed):

CHANGES.rst
```diff
@@ -1,3 +1,249 @@
...
+
+Bug Fixes
+---------
+
+- Ensure that ``SubarrayDescription.reference_location`` is always generated by
+  ```SimTelEventSource``, even if the metadata is missing. In that case, construct a
+  dummy location with the correct observatory height and latitude and longitude
+  equal to zero ("Null Island").
+
+  From now one require that ``SubarrayDescriptions`` contain a reference_location
```
Tobychev, line 80: This reads like a duplicate  of the above point, but looks like it was supposed to expand on it. Unfortunately I'm not sure what you were trying to add.

maxnoe: The first point is regarding what the `SimTelEventSource` now does, the second point is a general requirement for the `SubarrayDescription`

Tobychev: Ok, I think this would be clearer if the statements are in the reverse order, because the causal order, right?

As it is written I find the second paragraph to be mostly a confusing restatement of the first, because  "require that" there is a `reference_location` is not very different from "Ensure that" it "is always generated".

maxnoe: Good point. I reformulated a bit and moved the general point to the "API changes" section

Tobychev (approved):

PR #2551, opened 2024-04-26
Impact code cleanup

gschwefer:
In this PR I've collected all the implementations of the comments left on #2305 that I could just implement on the spot. This is mostly documentation updates, renaming of variables and such minor things. For the other comments, a bit more work or discussion might be required. I'm fine with either appending some of the things to this PR or opening new ones, depending on how large the things turn out to be

### DIFF
--- a/src/ctapipe/reco/impact.py
+++ b/src/ctapipe/reco/impact.py
@@ -100,12 +100,6 @@ class ImPACTReconstructor(HillasGeometryReconstructor):
 
     """
 
-    minimiser = traits.CaselessStrEnum(
-        ["minuit", "l-BFGS"],
-        default_value="minuit",
-        help="name minimiser to use in the fit",
-    ).tag(config=True)
-
     use_time_gradient = traits.Bool(
         default_value=False, help="Use time gradient in ImPACT reconstruction"
     ).tag(config=True)
@@ -617,24 +611,6 @@ class ImPACTReconstructor(HillasGeometryReconstructor):
             final_sum += chi2
 
         return final_sum
-
-    def get_likelihood_min(self, x):
-        """Wrapper class around likelihood function for use with scipy
-        minimisers
-
-        Parameters
-        ----------
-        x: ndarray
-            Array of minimisation parameters
-
-        Returns
-        -------
-        float: Likelihood value of test position
-
-        """
-        val = self.get_likelihood(x[0], x[1], x[2], x[3], x[4], x[5])
-
-        return val
 
     def set_event_properties(
         self,
--- a/src/ctapipe/reco/impact.py
+++ b/src/ctapipe/reco/impact.py
@@ -91,7 +91,7 @@ class ImPACTReconstructor(HillasGeometryReconstructor):
     ----------
     subarray : ctapipe.instrument.SubarrayDescription
         The telescope subarray to use for reconstruction
-    atmosphere_profile : ctapipe.AtmosphereDensityProfile
+    atmosphere_profile : ctapipe.atmosphere.AtmosphereDensityProfile
         Density vs. altitude profile of the local atmosphere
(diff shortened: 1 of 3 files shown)


src/ctapipe/reco/impact.py
```diff
@@ -84,20 +87,25 @@ class ImPACTReconstructor(HillasGeometryReconstructor):
     - Distance units in metres
     - Energy units in TeV
 
+    Parameters
+    ----------
+    subarray : ctapipe.instrument.SubarrayDescription
+        The telescope subarray to use for reconstruction
+    atmosphere_profile : ctapipe.AtmosphereDensityProfile
```
maxnoe, line 94: missing `.atmosphere`

src/ctapipe/reco/impact.py
```diff
@@ -84,20 +87,25 @@ class ImPACTReconstructor(HillasGeometryReconstructor):
...
+        Density vs. altitude profile of the local atmosphere
+    dummy_reconstructor : bool, optional
```
maxnoe, line 96: I think we should remove this option and find another way. A class under test should not know it's being tested. We should be able to mock the templates in the unit tests without having to have specific code for testing inside the class.

src/ctapipe/reco/impact.py
```diff
@@ -204,29 +208,31 @@ def __call__(self, event):
...
-            mask = dilate(self.subarray.tel[tel_id].camera.geometry, mask)
-            mask = dilate(self.subarray.tel[tel_id].camera.geometry, mask)
-            mask = dilate(self.subarray.tel[tel_id].camera.geometry, mask)
+            for _ in range(3):
+                mask = dilate(self.subarray.tel[tel_id].camera.geometry, mask)
+
             mask_dict[tel_id] = mask
 
         # Next, we look for geometry and energy seeds from previously applied reconstructors.
         # Both need to be present at elast once for ImPACT to run.
```
maxnoe, line 217: Should be configurable I think, not just take any?

i.e. `geometry_seed = Unicode(default="HillasReconstructor")`?

src/ctapipe/reco/impact.py
gschwefer: Ok, but I would make that a separate PR because it would likely also change the tests

src/ctapipe/reco/impact.py
gschwefer: Yep

maxnoe (dismissed):

src/ctapipe/reco/impact_utilities.py
```diff
@@ -192,6 +191,5 @@ def create_dummy_templates(
 
                 template_dict[key] = template.T * pe
 
-    filehandler = gzip.open(output_file, "wb")
-    pickle.dump(template_dict, filehandler)
-    filehandler.close()
+    with open(output_file, "wb") as filehandler:
```
maxnoe, line 194: needs to be `gzip.open` not `open`, the reader expects a gzipped file

maxnoe: Tests are failing with "not a gzipped file"

gschwefer: Yep, saw it

maxnoe (approved):

Tobychev (approved):

src/ctapipe/reco/impact.py
```diff
@@ -100,12 +100,6 @@ class ImPACTReconstructor(HillasGeometryReconstructor):
 
     """
 
-    minimiser = traits.CaselessStrEnum(
```
maxnoe, line 93: So this traitlets wasn't used anymore anyway? Just the config was still there?

gschwefer: Yep...

PR #2553, opened 2024-04-30
Remove upload to pages ci, replaced by using readthedocs

### LINKED ISSUE #2533: Remove deploy to GitHub pages action
https://github.com/cta-observatory/ctapipe/blob/bd7da6a5a00d52a5795afc306bb2debe66bded5e/.github/workflows/ci.yml#L179

I think this relic step can be removed

maxnoe:
Fixes #2533

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -179,18 +179,3 @@ jobs:
 
       - name: Build docs
         run: make doc
-
-      - name: Deploy to github pages
-        # only run on push to master
-        if: ${{ github.event_name == 'push' && github.ref == 'refs/heads/master' }}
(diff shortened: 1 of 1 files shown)


Tobychev:
I'm a bit surprised that switching documentation host doesn't require more changes than just cleaning out the config. I realise this is partially because the readthedocs support was added some time ago, but it still feels like there should be a corner of the developer documentation that had to be updated. 

I guess the maintainer info about the online docs where written at the same time as readthedocs was added?

maxnoe:
This was really just a leftover, it was not used anymore at all. It was behind a "if branch == master" check, but we are on main already for a long time

kosack (approved):

PR #2554, opened 2024-05-02
Add StatisticsAggregator API

TjarkMiener:
This PR adds a skeleton API for the statistical calculation of pixel-wise values over a given time interval. The current implementation features two different extraction methods, i.e. a `PlainExtractor` based on numpy functions and `SigmaClippingExtractor` based on astropy functions, which are also used in `lstchain`. 

I'm marking this PR as a draft because of the following discussion points:
1. How to implement the unit tests? I would propose to create random distributions with a given mean and std to assert the correctness of the extraction components.
2. How to deal with the last data chunk that is calculated with fewer samples as requested? Skip it or take the last samples fulfilling the `sample_size`?
3. Allow overlapping chunks of samples? How to define the validity range (using start-stop or only start)?

Related to #2542.


src/ctapipe/calib/camera/extractor.py
```diff
@@ -0,0 +1,246 @@
...
+    def __call__(
+        self, dl1_table, masked_pixels_of_sample=None, col_name="image"
+    ) -> list:
+
+        # in python 3.12 itertools.batched can be used
+        image_chunks = (
+            dl1_table[col_name].data[i : i + self.sample_size]
+            for i in range(0, len(dl1_table[col_name].data), self.sample_size)
+        )
+        time_chunks = (
```
mexanick, line 93: This will duplicate the above code block. Please refactor

src/ctapipe/calib/camera/extractor.py
```diff
@@ -0,0 +1,246 @@
...
+        )
+
+
+class SigmaClippingExtractor(StatisticsExtractor):
+    """
+    Extractor the statistics from a sequence of images
+    using astropy's sigma clipping functions
+    """
+
+    sigma_clipping_max_sigma = Int(
```
mexanick, line 147: naming question: do we want/need to carry "sigma_clipping_" prefix here and on the next line? Since this is an implementation that won't have generalization, I would recommend to remove it for conciseness

src/ctapipe/calib/camera/extractor.py
```diff
@@ -0,0 +1,246 @@
...
+        unused_values = np.abs(masked_images - pixel_mean) > (max_sigma * pixel_std)
+
+        # only warn for values discard in the sigma clipping, not those from before
+        outliers = unused_values & (~masked_images.mask)
+
+        # add outliers identified by sigma clipping for following operations
+        masked_images.mask |= unused_values
+
+        # median of the median over the camera
+        median_of_pixel_median = np.ma.median(pixel_median, axis=1)
```
mexanick, line 212: do we want a medial of pixels median or a _mean_ of pixels median?

src/ctapipe/calib/camera/extractor.py
```diff
@@ -0,0 +1,246 @@
...
+
+        # median of the std over the camera
+        median_of_pixel_std = np.ma.median(pixel_std, axis=1)
+
+        # std of the std over camera
+        std_of_pixel_std = np.ma.std(pixel_std, axis=1)
+
+        # outliers from median
+        image_deviation = pixel_median - median_of_pixel_median[:, np.newaxis]
```
mexanick, line 221: this is absolute deviation, in the docstring you say relative. Please clarify and update accordingly

src/ctapipe/calib/camera/tests/test_extractors.py
```diff
@@ -0,0 +1,51 @@
...
+    assert np.any(np.abs(sigmaclipping_stats_list[0].mean - 77.0) > 1.5) == False
+    
+    assert np.any(np.abs(plain_stats_list[1].median - 2.0) > 1.5) == False
+    assert np.any(np.abs(sigmaclipping_stats_list[1].median - 77.0) > 1.5) == False
+    
+    assert np.any(np.abs(plain_stats_list[0].std - 5.0) > 1.5) == False
+    assert np.any(np.abs(sigmaclipping_stats_list[0].std - 10.0) > 1.5) == False
+    
+def test_check_outliers(example_subarray):
+    sigmaclipping_extractor = SigmaClippingExtractor(subarray=example_subarray, sample_size=2500)
```
mexanick, line 38: This and line 14 can be combined together in one setup function

src/ctapipe/containers.py
```diff
@@ -412,7 +413,18 @@ class MorphologyContainer(Container):
...
-    """Store descriptive statistics"""
+    """Store descriptive statistics of a sequence of images"""
+
+    validity_start = Field(np.float32(nan), "start of the validity range")
+    validity_stop = Field(np.float32(nan), "stop of the validity range")
+    mean = Field(np.float32(nan), "Channel-wise and pixel-wise mean")
+    median = Field(np.float32(nan), "Channel-wise and pixel-wise median")
+    median_outliers = Field(np.float32(nan), "Channel-wise and pixel-wise median outliers")
+    std = Field(np.float32(nan), "Channel-wise and pixel-wise standard deviation")
+    std_outliers = Field(np.float32(nan), "Channel-wise and pixel-wise standard deviation outliers")
```
mexanick, line 424: Not sure what this means, please update the docstring

mexanick (changes requested):

src/ctapipe/calib/camera/extractor.py
```diff
@@ -0,0 +1,246 @@
...
+        self, dl1_table, masked_pixels_of_sample=None, col_name="image"
+    ) -> list:
+        """
+        Call the relevant functions to extract the statistics
+        for the particular extractor.
+
+        Parameters
+        ----------
+        dl1_table : ndarray
+            dl1 table with images and times stored in a numpy array of shape
```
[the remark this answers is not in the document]
TjarkMiener: it's the trigger time here in order to give the validity range. I modify to "timestamps".

src/ctapipe/calib/camera/extractor.py
TjarkMiener: pinging @FrancaCassol here since this part is from the lstchain routines

src/ctapipe/calib/camera/extractor.py
TjarkMiener: same @FrancaCassol

src/ctapipe/calib/camera/tests/test_extractors.py
TjarkMiener: How to do this? I would need the `example_subarray` to be passed from the `test_function`.

TjarkMiener:
I'm having some troubles with ruff, which makes pylint fail. Can someone please help? Thanks in advance!

src/ctapipe/calib/camera/extractor.py
```diff
@@ -0,0 +1,220 @@
...
+        Returns
+        -------
+        List StatisticsContainer:
+            List of extracted statistics and validity ranges
+        """
+
+        # in python 3.12 itertools.batched can be used
+        image_chunks = (
+            dl1_table[col_name].data[i : i + self.sample_size]
+            for i in range(0, len(dl1_table[col_name].data), self.sample_size)
```
TjarkMiener, line 78: @mexanick about the overlapping sample windows. In principle, we could introduce a `self.sample_shift` here:
```
        image_chunks = (
            dl1_table[col_name].data[i : i + self.sample_size]
            for i in range(0, len(dl1_table[col_name].data), self.sample_shift)
        )
```
same for the time_chunks. What do you think?

mexanick (dismissed):

src/ctapipe/calib/camera/extractor.py
FrancaCassol: we generally prefer median as figure of merit which is more robust in the case of outliers

maxnoe:
For future reference: the issue was that the pre-commit hook wasn't yet installed for some of the commits. 

To run it on all files affected by this PR, the solution was:

```
pre-commit run --files $(git diff --name-only origin/main)
```

src/ctapipe/calib/camera/extractor.py
FrancaCassol: sorry, of which docstring are we speaking? I can not find it. In any case, the point is that the deviation is evaluated with respect to the median value , not as absolute value

mexanick (dismissed):

ctoennis (approved):

src/ctapipe/calib/camera/extractor.py
```diff
@@ -0,0 +1,231 @@
...
+        chunk_shift=None,
+        col_name="image",
+    ) -> list:
+        """
+        Call the relevant functions to extract the statistics
+        for the particular extractor.
+
+        Parameters
+        ----------
+        dl1_table : ndarray
```
Hckjs, line 66: its a `Qtable`-object, isnt it?

src/ctapipe/calib/camera/extractor.py
```diff
@@ -0,0 +1,231 @@
...
+        # Function to split table data into appropriated chunks
+        def _get_chunks(col_name):
+            return [
+                (
+                    dl1_table[col_name].data[i : i + self.chunk_size]
+                    if i + self.chunk_size <= len(dl1_table[col_name])
+                    else dl1_table[col_name].data[
+                        len(dl1_table[col_name].data) - self.chunk_size : len(
+                            dl1_table[col_name].data
+                        )
```
Hckjs, line 96: Do we maybe need a condition somewhere that `len(dl1_table) >= self.chunk_size`? Othwerwise it will silently select a weird chunk here (`len(dl1_table[col_name].data) - self.chunk_size` will get negative)

TjarkMiener: Good point. I think we should catch it at the beginning of this `__call__` and break.

src/ctapipe/calib/camera/extractor.py
```diff
@@ -0,0 +1,231 @@
...
+    ) -> list:
+        """
+        Call the relevant functions to extract the statistics
+        for the particular extractor.
+
+        Parameters
+        ----------
+        dl1_table : astropy.table.QTable
+            dl1 table with images and timestamps stored in a numpy array of shape
+            (n_images, n_channels, n_pix).
```
Hckjs, line 68: explanation needs an update too

src/ctapipe/calib/camera/extractor.py
```diff
@@ -0,0 +1,231 @@
...
+        return stats_list
+
+    @abstractmethod
+    def _extract(self, images, times, masked_pixels_of_sample) -> StatisticsContainer:
+        pass
+
+
+class PlainExtractor(StatisticsExtractor):
+    """
+    Extractor the statistics from a chunk of images
```
Hckjs, line 119: typo -> Extract

Hckjs (dismissed):

mexanick (approved):

TjarkMiener:
Thanks @Hckjs and @mexanick for the comments and the approvals of the PR. Can you please merge in case you are happy with the PR, @maxnoe @kosack? Thanks in advance!

mexanick (dismissed):

Hckjs (dismissed):

src/ctapipe/calib/camera/extractor.py
```diff
@@ -0,0 +1,228 @@
...
+                            dl1_table[col_name].data
+                        )
+                    ]
+                )
+                for i in range(0, len(dl1_table[col_name].data), chunk_shift)
+            ]
+
+        # Get the chunks for the timestamps and selected column name
+        time_chunks = _get_chunks("time")
+        image_chunks = _get_chunks(col_name)
```
TjarkMiener, line 108: Actually, the operation `dl1_table[col_name].data` works only for `<class 'astropy.table.column.Column'>` objects. However, `dl1_table["time"]` is a `<class 'astropy.time.core.Time'>` object. My suggestion would be:
```
        # Function to split table data into appropriated chunks
        def _get_chunks(dl1_table_data):
            return [
                (
                    dl1_table_data[i : i + self.chunk_size]
                    if i + self.chunk_size <= len(dl1_table_data)
                    else dl1_table_data[
                        len(dl1_table_data) - self.chunk_size : len(
                        dl1_table_data
                        )
                    ]
                )
                for i in range(0, len(dl1_table_data), chunk_shift)
            ]

        # Get the chunks for the timestamps and selected column name
        time_chunks = _get_chunks(dl1_table["time"])
        image_chunks = _get_chunks(dl1_table[col_name].data)
```
What do you think, @mexanick @Hckjs?

mexanick: looks aok, but please also add relevant tests

src/ctapipe/calib/camera/extractor.py
```diff
@@ -0,0 +1,228 @@
...
+                        len(dl1_table[col_name].data) - self.chunk_size : len(
+                            dl1_table[col_name].data
+                        )
+                    ]
+                )
+                for i in range(0, len(dl1_table[col_name].data), chunk_shift)
+            ]
+
+        # Get the chunks for the timestamps and selected column name
+        time_chunks = _get_chunks("time")
```
TjarkMiener, line 107: We should probably read the column "`time_mono`"  instead of "`time`". When reading the dl1 table from a file using the `TableLoader` with the `read_telescope_events_by_id` function, we have both columns (for mono data telescope and subarray trigger is identical). However, the correct timestamps should be taking from the telescope trigger, since the calibration events are monoscopic. Can you please confirm that "`time_mono`" should be used?@maxnoe @Hckjs

src/ctapipe/calib/camera/extractor.py
```diff
@@ -0,0 +1,228 @@
...
+        # outliers from median
+        image_deviation = pixel_median - median_of_pixel_median[:, np.newaxis]
+        image_median_outliers = np.logical_or(
+            image_deviation
+            < self.image_median_cut_outliers[0]  # pylint: disable=unsubscriptable-object
+            * median_of_pixel_median[:, np.newaxis],
+            image_deviation
+            > self.image_median_cut_outliers[1]  # pylint: disable=unsubscriptable-object
+            * median_of_pixel_median[:, np.newaxis],
+        )
```
TjarkMiener, line 207: After some offline discussions with @FrancaCassol, we noticed that here we need to distinguish between checking for outliers for signal (flatfield) and noise (pedestal) using `median_of_pixel_median` and `std_of_pixel_std`, respectively.

src/ctapipe/calib/camera/extractor.py
TjarkMiener: Resolved by [e8497fd](https://github.com/cta-observatory/ctapipe/pull/2554/commits/e8497fde9386d17644bdf8bc02deba736658cdf3).

src/ctapipe/calib/camera/extractor.py
TjarkMiener: Resolved by [e8497fd](https://github.com/cta-observatory/ctapipe/pull/2554/commits/e8497fde9386d17644bdf8bc02deba736658cdf3).

src/ctapipe/calib/camera/extractor.py
TjarkMiener: Resolved by [e8497fd](https://github.com/cta-observatory/ctapipe/pull/2554/commits/e8497fde9386d17644bdf8bc02deba736658cdf3).

src/ctapipe/calib/camera/extractor.py
```diff
@@ -0,0 +1,261 @@
...
+
+        return StatisticsContainer(
+            extraction_start=times[0],
+            extraction_stop=times[-1],
+            mean=pixel_mean.filled(np.nan),
+            median=pixel_median.filled(np.nan),
+            median_outliers=image_median_outliers.filled(True),
+            std=pixel_std.filled(np.nan),
+        )
```
TjarkMiener, line 174: The `PlainExtractor` is used for the peak arrival time correction in the [lstchain code](https://github.com/cta-observatory/cta-lstchain/blob/main/lstchain/calib/camera/flatfield.py#L269-L309). The outliers are only identified based on the medians and not on the stds. Therefore `std_outliers` is empty. We should still fill the `StatisticsContainer` by setting `std_outliers` to `False`, i.e. adding:
`std_outliers=np.full(np.shape(image_median_outliers), False),`

maxnoe (commented):
Sorry, I just realized that some comments I made were still "pending", I am posting them now as they were.

src/ctapipe/calib/camera/extractor.py
```diff
@@ -0,0 +1,228 @@
...
+
+
+class StatisticsExtractor(TelescopeComponent):
+    """Base StatisticsExtractor component"""
+
+    chunk_size = Int(
+        2500,
+        help="Size of the chunk used for the calculation of the statistical values",
+    ).tag(config=True)
+    image_median_cut_outliers = List(
```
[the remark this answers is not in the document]
TjarkMiener: `median_cut_outliers` is used in both extractors, but `std_cut_outliers` is not.

src/ctapipe/calib/camera/extractor.py
```diff
@@ -0,0 +1,267 @@
...
+    """Base StatisticsExtractor component"""
+
+    chunk_size = Int(
+        2500,
+        help="Size of the chunk used for the calculation of the statistical values",
+    ).tag(config=True)
+    median_cut_outliers = List(
+        [-0.3, 0.3],
+        help="""Interval of accepted values \\
+                (fraction with respect to camera median value)""",
```
maxnoe, line 34: From the values and the description here, it is not clear to me how an outlier will be defined.

src/ctapipe/calib/camera/extractor.py
```diff
@@ -0,0 +1,267 @@
...
+    def __init__(self, subarray, config=None, parent=None, **kwargs):
+        """
+        Base component to handle the extraction of the statistics
+        from a chunk of charges and pulse times (images).
+
+        Parameters
+        ----------
+        kwargs
+        """
+        super().__init__(subarray=subarray, config=config, parent=parent, **kwargs)
```
maxnoe, line 46: This method is not needed if it only contains the `super()` call

src/ctapipe/calib/camera/extractor.py
```diff
@@ -0,0 +1,267 @@
...
+    ).tag(config=True)
+    median_cut_outliers = List(
+        [-0.3, 0.3],
+        help="""Interval of accepted values \\
+                (fraction with respect to camera median value)""",
+    ).tag(config=True)
+
+    def __init__(self, subarray, config=None, parent=None, **kwargs):
+        """
+        Base component to handle the extraction of the statistics
```
maxnoe, line 39: doc of `__init__` should be empty and everything documented in the cls docstring

src/ctapipe/calib/camera/extractor.py
```diff
@@ -0,0 +1,267 @@
...
+        col_name : string
+            column name in the DL1 table
+
+        Returns
+        -------
+        List StatisticsContainer:
+            List of extracted statistics and extraction chunks
+        """
+
+        # Check if the provided data contains an unique type of calibration events
```
maxnoe, line 78: a unique (starts with a consonant sound)

src/ctapipe/calib/camera/extractor.py
```diff
@@ -0,0 +1,261 @@
...
+)
+
+
+class StatisticsExtractor(TelescopeComponent):
+    """
+    Base component to handle the extraction of the statistics from a dl1 table
+    containing charges, peak times and/or charge variances (images).
+    """
+
+    chunk_size = Int(
```
maxnoe, line 30: Since you were talking also about other approaches to extract these values (rolling mean for example),
I think the `chunk_size` argument shouldn't be here in the base class.

The baseclass should only define the API that is common for all extractors.

src/ctapipe/calib/camera/extractor.py
```diff
@@ -0,0 +1,261 @@
...
+        2500,
+        help="Size of the chunk used for the calculation of the statistical values",
+    ).tag(config=True)
+
+    def __call__(
+        self,
+        dl1_table,
+        masked_pixels_of_sample=None,
+        chunk_shift=None,
```
maxnoe, line 39: chunk_shift sounds like it should be a configuration value and not an argument to `__call__`? It also seems to be too specific for the base class API.

src/ctapipe/calib/camera/extractor.py
```diff
@@ -0,0 +1,261 @@
...
+    Base component to handle the extraction of the statistics from a dl1 table
+    containing charges, peak times and/or charge variances (images).
+    """
+
+    chunk_size = Int(
+        2500,
+        help="Size of the chunk used for the calculation of the statistical values",
+    ).tag(config=True)
+
+    def __call__(
```
maxnoe, line 35: I think for the base class, `__call__` should be abstract and the arguments of `__call__` need to be the same for all subclasses to be exchangeble via the configuration system without adapting the call site.

src/ctapipe/calib/camera/tests/test_extractors.py
```diff
@@ -0,0 +1,155 @@
...
+import numpy as np
+import pytest
+from astropy.table import Table
+from astropy.time import Time
+
+from ctapipe.calib.camera.extractor import PlainExtractor, SigmaClippingExtractor
+
+
+@pytest.fixture(name="test_plainextractor")
+def fixture_test_plainextractor(example_subarray):
```
maxnoe, line 14: Instantiating a class is fine in the test itself, I would create a fixture for this.

Use fixtures to setup more complex test data for example.

src/ctapipe/calib/camera/tests/test_extractors.py
```diff
@@ -0,0 +1,155 @@
...
+
+import numpy as np
+import pytest
+from astropy.table import Table
+from astropy.time import Time
+
+from ctapipe.calib.camera.extractor import PlainExtractor, SigmaClippingExtractor
+
+
+@pytest.fixture(name="test_plainextractor")
```
maxnoe, line 13: name is not needed if you just directly name the function appropriately

src/ctapipe/calib/camera/extractor.py
TjarkMiener: I'm not sure if we really need any rolling mean, since one could in principle set the `chunk_shift` to 1. Currently, I can't think of more extractors that should be implemented. Do you have some in mind that we should consider?

src/ctapipe/calib/camera/extractor.py
```diff
@@ -0,0 +1,261 @@
...
+        stats_list = []
+        for images, times in zip(image_chunks, time_chunks):
+            stats_list.append(
+                self.extract(images, times, masked_pixels_of_sample, outlier_check)
+            )
+        return stats_list
+
+    @abstractmethod
+    def extract(
+        self, images, times, masked_pixels_of_sample, outlier_check
```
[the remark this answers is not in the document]
TjarkMiener: Sounds good to me that will avoid checking the data for the specific event type. Even though, we should still check and break if the provided data not contains a unique type of calibration events.

src/ctapipe/calib/camera/extractor.py
TjarkMiener: Maybe I'm missing something, but the `__call__` is designed to be identical for all the subclasses. Simply the function `extract()` is implemented in each subclass.

src/ctapipe/calib/camera/extractor.py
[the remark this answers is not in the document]
TjarkMiener: Done

src/ctapipe/calib/camera/tests/test_extractors.py
TjarkMiener: Okay

mexanick (dismissed):
Looks good to me. I believe, after the last round of reviews from @maxnoe the documentation and semantics have been improved. @maxnoe , considering the API-related comments, I believe the remaining ones would depend on the way how this API will be used. Therefore, I suggest to merge the current implementation and proceed with the calibration calculators (primary users of this API) and fix remaining issues there or even in a separate refactoring PR if needed, just to avoid a never-ending and constantly growing PR here.

src/ctapipe/calib/camera/extractor.py
```diff
@@ -0,0 +1,256 @@
...
+            column name in the DL1 table
+
+        Returns
+        -------
+        List StatisticsContainer:
+            List of extracted statistics and extraction chunks
+        """
+
+        # Check if the length of the dl1 table is greater or equal than the size of the chunk.
+        if len(dl1_table[col_name]) < self.chunk_size:
```
maxnoe, line 66: Why not just compute a single value on the available rows?

src/ctapipe/calib/camera/extractor.py
```diff
@@ -0,0 +1,256 @@
...
+            raise ValueError(
+                f"The length of the DL1 table ({len(dl1_table[col_name])}) must be greater or equal than the size of the chunk ({self.chunk_size})."
+            )
+
+        # Function to split table data into appropriated chunks
+        def _get_chunks(dl1_table_data, chunk_shift):
+            if chunk_shift is None:
```
maxnoe, line 73: I think this could just use the same code if you use `chunk_shift=0` as default instead of `chunk_shift=None`:

src/ctapipe/calib/camera/extractor.py
```diff
@@ -0,0 +1,256 @@
...
+            else:
+                return (
+                    dl1_table_data[i : i + self.chunk_size]
+                    for i in range(
+                        0, len(dl1_table_data) - self.chunk_size, chunk_shift
+                    )
+                )
+
+        # Get the chunks for the timestamps and selected column name
+        time_chunks = _get_chunks(dl1_table["time_mono"], chunk_shift)
```
maxnoe, line 93: Why do we loop over two chunk iterators instead of just selecting the rows of the whole table?

TjarkMiener: Good point!

src/ctapipe/calib/camera/extractor.py
TjarkMiener: Actually, `chunk_shift=self.chunk_size` once `chunk_shift is None`. This is the first pass, where we do not allow overlapping chunks except in the border case at the end of the sequence. The last extraction chunk of the first pass will be performed with the last available elements of the sequence in order to fill a complete chunk. For the second pass, we detect faulty chunks and perform a detail scan of the region of trouble with a sliced dl1 table (around that region). Therefore, we utilize the `chunk_shift` argument. If the detail scan is performed we do not want to include the border case as before in order to avoid repetitions of the last complete chunk every time the chunk is shifted.

src/ctapipe/calib/camera/extractor.py
TjarkMiener: Relates a bit to the reply to the comment below. I think we should never allow an extraction with less stats than required.

src/ctapipe/calib/camera/extractor.py
maxnoe: That's not what I meant. I meant that the `if / else` block here with very similar code could be avoided if "no shift" would be `chunk_shift=0` instead of `chunk_shift=None`

TjarkMiener: We bypassed the `if/else` bock by `yield`ing non-overflow chunks. If requested the last complete chunk will be `yield` as well. Can you please check if the function is clearer now? Thanks!

mexanick (dismissed):

src/ctapipe/calib/camera/extractor.py
```diff
@@ -0,0 +1,250 @@
...
+
+        Parameters
+        ----------
+        dl1_table : astropy.table.Table
+            DL1 table with images of shape (n_images, n_channels, n_pix)
+            and timestamps of shape (n_images, ) stored in an astropy Table
+        masked_pixels_of_sample : ndarray
+            boolean array of masked pixels of shape (n_pix, ) that are not available for processing
+        chunk_shift : int
+            number of samples to shift the extraction chunk
```
maxnoe, line 55: From the description here, I didn't understand that chunk_shift shift by which the window is advanced and that `chunk_shift < chunk_size` implies overlapping chunks.

Maybe the description here could be clearer.

maxnoe: I also think this should be a traitlet, so people can configure this through the config system

src/ctapipe/calib/camera/extractor.py
```diff
@@ -0,0 +1,250 @@
...
+            )
+
+        # Function to split the dl1 table into appropriated chunks
+        def _get_chunks(dl1_table, chunk_shift):
+            # Calculate the range step: Use chunk_shift if provided, otherwise use chunk_size
+            step = chunk_shift if chunk_shift is not None else self.chunk_size
+
+            # Generate chunks that do not overflow
+            for i in range(0, len(dl1_table), step):
+                if i + self.chunk_size <= len(dl1_table):
```
maxnoe, line 78: Couldn't this if be integrated into the range?

maxnoe: Maybe  `for i in range(0, len(dl1_table) - self.chunk_size, step)` ? (I didn't check..)

mexanick: I think an integer division with reminder may help here.

maxnoe: I think the range for above is enough... tested it with a couple of examples

src/ctapipe/calib/camera/extractor.py
```diff
@@ -0,0 +1,250 @@
...
+            if chunk_shift is None and len(dl1_table) % self.chunk_size != 0:
+                yield dl1_table[-self.chunk_size :]
+
+        # Get the chunks of the dl1 table
+        dl1_chunks = _get_chunks(dl1_table, chunk_shift)
+
+        # Calculate the statistics from a chunk of images
+        stats_list = [
+            self.extract(
+                chunk[col_name].data, chunk["time_mono"], masked_pixels_of_sample
```
maxnoe, line 91: This looses unit information on the column and I don't see where this is re-attached.

src/ctapipe/calib/camera/extractor.py
```diff
@@ -0,0 +1,250 @@
...
+        pass
+
+
+class PlainExtractor(StatisticsExtractor):
+    """
+    Extract the statistics from a chunk of peak time images
+    using numpy and scipy functions
+    """
+
+    time_median_cut_outliers = List(
```
maxnoe, line 109: Is it necessary to define the outliers here in this extractor class?

The outliers are defined from information that is also stored in the output. Outliers can be defined in many different ways.

I think I'd prefer if the `StatsExtractor` class only did that: extract statistics.

The definitions of outliers in the extracted statistics should be done by something else, maybe an `OutlierDetector`

src/ctapipe/calib/camera/extractor.py
```diff
@@ -0,0 +1,250 @@
...
+        cam_median_of_pix_median = np.ma.median(pix_median, axis=1)
+
+        # Camera median of the pixel stds
+        cam_median_of_pix_std = np.ma.median(pix_std, axis=1)
+
+        # Camera std of the pixel stds
+        cam_std_of_pix_std = np.ma.std(pix_std, axis=1)
+
+        # Detect outliers based on the deviation of the median distribution
+        median_deviation = pix_median - cam_median_of_pix_median[:, np.newaxis]
```
maxnoe, line 211: Why is the outlier detection different to the other class?
The computation of the statistics is differente, because its a different extraction method, but why is the outlier detection different?

I think we really should split computation of statistics and deciding which pixels are "suspicious" into two separate steps.

That also makes different methods of statistics extraction and different methods of outlier detection composable

src/ctapipe/calib/camera/extractor.py
TjarkMiener: The idea of making it not a trailet was that we want to pass twice over the data with only one extractor instance. Maybe this pseudo code helps (justify that to make `chunk_shift` a trailet of the component using the `StatisticsExtractor` API):

```python
        ...
        # Get the extractor
        extractor = self.stats_extractor[self.stats_extractor_type.tel[tel_id]]

        # First pass through the whole provided dl1 data
        stats_list_firstpass = extractor(dl1_table=dl1_table[tel_id], col_name=col_name)

        for chunk_nr, stats in enumerate(stats_list_firstpass):
            ...
            # Detect faulty chunks by calculating the fraction of faulty pixels over the camera and checking if the threshold is exceeded.
            outliers_to_faultless_pixels_ratio = detect_outliers(...)
            # Check if chunk is faulty 
            if outliers_to_faultless_pixels_ratio > self.faulty_pixels_threshold:
                   # Get the ranges for slicing the region of troubles
                   slice_start, slice_stop = self._get_slice_range(...)
                   # Slice the dl1 table according to the previously caluclated start and stop.
                   dl1_table_sliced = dl1_table[tel_id][slice_start:slice_stop]
                   # Run the stats extractor on the sliced dl1 table with a chunk_shift
                   # to remove the period of trouble (carflashes etc.) as effectively as possible.
                   stats_list_secondpass = extractor(
                        dl1_table=dl1_table_sliced, chunk_shift=self.chunk_shift
                    )
                    # Do something
                    do_somthing(stats_list_secondpass)

```

src/ctapipe/calib/camera/extractor.py
TjarkMiener: Those columns are image, peak_time or variance. I don't see units for them when printing the Table. Do they have units attached before?

src/ctapipe/calib/camera/extractor.py
TjarkMiener: Sounds reasonable to us. The outlier detection is remove from the PR and will be treated outside the stats extraction.

src/ctapipe/calib/camera/extractor.py
TjarkMiener: See comment reply above.

src/ctapipe/calib/camera/extractor.py
```diff
@@ -0,0 +1,250 @@
...
+            # Calculate the range step: Use chunk_shift if provided, otherwise use chunk_size
+            step = chunk_shift if chunk_shift is not None else self.chunk_size
+
+            # Generate chunks that do not overflow
+            for i in range(0, len(dl1_table), step):
+                if i + self.chunk_size <= len(dl1_table):
+                    yield dl1_table[i : i + self.chunk_size]
+
+            # If chunk_shift is None, ensure the last chunk is of size chunk_size, if needed
+            if chunk_shift is None and len(dl1_table) % self.chunk_size != 0:
```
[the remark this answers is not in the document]
TjarkMiener: Updated the doc string describing the reason why we need to have it here.

src/ctapipe/calib/camera/extractor.py
TjarkMiener: if is redundant when adjusting the end of the range as suggested by @maxnoe offline. New commit includes this change.

src/ctapipe/calib/camera/extractor.py
```diff
@@ -41,18 +37,24 @@ def __call__(
         col_name="image",
     ) -> list:
         """
-        Prepare the extraction chunks and call the relevant function of the particular extractor
-        to extract the statistical values.
+        Divides the input DL1 table into overlapping or non-overlapping chunks of size `chunk_size`
```
maxnoe, line 40: First line of docstring should be a one sentence summary:
```
"""
Divide table into chunks.

More details on method follow here.
"""
```

src/ctapipe/calib/camera/extractor.py
```diff
@@ -41,18 +37,24 @@ def __call__(
...
+        and call the relevant function of the particular extractor to extract the statistical values.
+        The chunks are generated in a way that ensures they do not overflow the bounds of the table.
+        - If `chunk_shift` is None, extraction chunks will not overlap, but the last chunk is ensured to be
+        of size `chunk_size`, even if it means the last two chunks will overlap.
+        - If `chunk_shift` is provided, it will determine the number of samples to shift between the start
```
maxnoe, line 45: Do we allow `chunk_shift > chunk_size` resulting in non-contiguous chunks?

TjarkMiener: good point. We should not allow this scenario.

src/ctapipe/calib/camera/extractor.py
maxnoe: This method is very general, I can pass any table I like and it will perform mean/std/median computation.

maxnoe: It should behave correctly when the column passed has units.

TjarkMiener: Okay, I see. I generalized it in the last commit. Please check if you agree with the changes. Thanks!

TjarkMiener: Now, seeing the following comment regarding the units. I'm not sure how to solve it. Can you please help? Thanks

src/ctapipe/calib/camera/tests/test_extractor.py
```diff
@@ -0,0 +1,92 @@
...
+        np.linspace(60117.911, 60117.9258, num=5500), scale="tai", format="mjd"
+    )
+    charge_data = np.random.normal(77.0, 10.0, size=(5500, 2, 1855))
+    # Create table
+    charge_table = Table(
+        [times, charge_data],
+        names=("time_mono", "image"),
+    )
+    # Initialize the extractor
+    extractor = SigmaClippingExtractor(subarray=example_subarray, chunk_size=2500)
```
maxnoe, line 79: I'd like to see here a test that actually contains some outliers and checks that the expected values are in the output.

So add some very large outliers to `charge_data`, that would skew the mean beyond the expected uncertainties and check the sigma clipping works as intended.

maxnoe (changes requested):
Aside from missing tests for the actually computed values, this looks good to me now!

src/ctapipe/calib/camera/tests/test_extractor.py
TjarkMiener: Good point. I added a separate test for it.

src/ctapipe/calib/camera/extractor.py
```diff
@@ -0,0 +1,164 @@
...
+        help="Size of the chunk used for the calculation of the statistical values",
+    ).tag(config=True)
+
+    def __call__(
+        self,
+        table,
+        masked_pixels_of_sample=None,
+        chunk_shift=None,
+        col_name="image",
+    ) -> dict:
```
maxnoe, line 38: Wouldn't a `Table` be a more suitable return value?

maxnoe: I think instead of timestamp, it should be `timestamp_start`, `timestamp_stop`, and maybe also add the event_ids of the first / last events of the chunk?

TjarkMiener: Before we had `extraction_start` and `extraction_stop` which is more descriptive I think. Only the `event_ids` enough or do we need also the`obs_ids` (and `tel_ids`)?

TjarkMiener: Resolved with the latest commit. I choose `time_start`, `time_end` and `event_id_start`,  `event_id_end`.

src/ctapipe/calib/camera/extractor.py
```diff
@@ -0,0 +1,167 @@
...
+
+            # Generate chunks that do not overflow
+            for i in range(0, len(table) - self.chunk_size + 1, step):
+                yield table[i : i + self.chunk_size]
+
+            # If chunk_shift is None, ensure the last chunk is of size chunk_size, if needed
+            if chunk_shift is None and len(table) % self.chunk_size != 0:
+                yield table[-self.chunk_size :]
+
+        # Calculate the statistics for each chunk of images
```
maxnoe, line 96: `units = {colname: table[colname].unit}`

src/ctapipe/calib/camera/extractor.py
```diff
@@ -0,0 +1,167 @@
...
+            stats = self.extract(chunk[col_name].data, masked_pixels_of_sample)
+            data["time_start"].append(chunk["time_mono"][0])
+            data["time_end"].append(chunk["time_mono"][-1])
+            data["event_id_start"].append(chunk["event_id"][0])
+            data["event_id_end"].append(chunk["event_id"][-1])
+            data["mean"].append(stats.mean)
+            data["median"].append(stats.median)
+            data["std"].append(stats.std)
+
+        return Table(data)
```
maxnoe, line 108:
```suggestion
        return Table(data, units=units)
```

src/ctapipe/calib/camera/extractor.py
TjarkMiener: Would this also be working when `table[col_name].unit` is `None`? The tests are failing because I didn't set any units in there, but how to set a unit `None` to match table["image"].unit and table["peak_time"].unit, which are `None`s in the dl1 file?

TjarkMiener: Thanks!

maxnoe (approved):

src/ctapipe/calib/camera/tests/test_extractor.py
```diff
@@ -0,0 +1,136 @@
...
+
+    # Check if the start and end values are properly set for the timestamps and event IDs
+    assert ped_stats[0]["time_start"] == times[0]
+    assert time_stats[0]["event_id_start"] == event_ids[0]
+    assert ped_stats[1]["time_end"] == times[-1]
+    assert time_stats[1]["event_id_end"] == event_ids[-1]
+
+    # Check if the calculated statistical values are reasonable
+    # for a camera with two gain channels
+    assert not np.any(np.abs(ped_stats[0]["mean"] - 2.0) > 1.5)
```
maxnoe, line 60: These tests are not very helpful if they fail.

Using the numpy test functions yields much better error messages, e.g. in this case, I would test:

```
np.testing.assert_allclose(ped_stats[0]["mean"], 2.0, atol=..., rtol=...)
```

src/ctapipe/calib/camera/tests/test_extractor.py
```diff
@@ -0,0 +1,136 @@
...
+
+def test_extractors(example_subarray):
+    """test basic functionality of the StatisticsExtractors"""
+
+    # Create dummy data for testing
+    times = Time(
+        np.linspace(60117.911, 60117.9258, num=5000), scale="tai", format="mjd"
+    )
+    event_ids = np.linspace(35, 725000, num=5000, dtype=int)
+    ped_data = np.random.normal(2.0, 5.0, size=(5000, 2, 1855))
```
maxnoe, line 21: Please use a fixed random seed so we don't get spurious test failures and use the new numpy random API, not the deprecated one:

```
ped_data = rng.normal(...)
charge_data = rng.normal(...)
```

maxnoe (changes requested):
The tests should use a fixed random seed and the modern numpy random api.

Test asserts should use the numpy assert functions

maxnoe (approved):

mexanick (approved):

TjarkMiener:
Thanks, @maxnoe @mexanick! Anything missing or can this PR get merged? The treatment of outlier detection has been moved to a separate PR #2604.

src/ctapipe/calib/camera/extractor.py
```diff
@@ -0,0 +1,168 @@
...
+
+from ctapipe.containers import StatisticsContainer
+from ctapipe.core import TelescopeComponent
+from ctapipe.core.traits import Int
+
+
+class StatisticsExtractor(TelescopeComponent):
+    """
+    Base component to handle the extraction of the statistics from a table
+    containing e.g. charges, peak times and/or charge variances (images).
```
kosack, line 26: I think a better docstring is needed here. It's very unclear what the output is.  Is it a single set of statistics, or a series of statistics aggreggated over (potentially overlapping) time bins?  It's more clear later in the `__call__` docstring, but could be better here. Also this is **pixel-wise** statistics, which is maybe important to mention.

src/ctapipe/calib/camera/tests/test_extractor.py
```diff
@@ -0,0 +1,141 @@
+"""
+Tests for StatisticsExtractor and related functions
+"""
+
+import numpy as np
+import pytest
+from astropy.table import Table
+from astropy.time import Time
+
+from ctapipe.calib.camera.extractor import PlainExtractor, SigmaClippingExtractor
```
kosack, line 10: in real use (outside of tests) do we really want to import this from such a deep module? Should we maybe expose it at the `ctapipe.calib` level in `__init__.py`?

kosack (commented):
I'm still not a huge fan of the name "StatisticsExtractor" is a bit generic and is easily confused with the fact that we use "Extractor" to mean deriving images from waveforms elsewhere...  But I can live it with.  What this class does is take a event table and converts it to a summary table with statistics computed in time intervals. Basically, it's similar to a Pandas operation like:
```py3
df.group_by(time_slice).aggregate((np.mean, np.median, np.std))
```
It's also not clear that it is for images, and not any table column, so maybe `ImageStatisticsAggregator`? 

By the way, we need a similar functionality for _all table columns_, not just images to be able to generate monitoring information for QualPipe and for the DL3.monitoring tables!  e.g. to get the mean/meadian/std/min/max of the weather info (temperature, humidity), or the HV settings, or pointing, etc.  So maybe this could be generalized this into such a component?  I.e. you could just have a "column_to_aggregate" option that is "image" by default, but could be set to any column (or better yet, set of columns, which would be nice for monitoring quantities that are measured together), and you would just let broadcasting do the work of doing it pixel-wise vs scalars.   EDIT: that would also mean we would have to build the Container on the fly probably, since we if the columns to aggregate are `("temperature", "pressure")` then you would want the output table to have columns like `temperature_mean`, `temperature_std`, ...

That brings me to one more comment: right now this is in `calib.camera`, but if it was more general, it could be in a module called `monitoring`. Anyhow those last 2 points could be a future refactoring in  another PR.

src/ctapipe/calib/camera/extractor.py
```diff
@@ -0,0 +1,168 @@
...
+
+        # Calculate the mean, median, and std over the chunk per pixel
+        pixel_mean = np.ma.mean(masked_images, axis=0).filled(np.nan)
+        pixel_median = np.ma.median(masked_images, axis=0).filled(np.nan)
+        pixel_std = np.ma.std(masked_images, axis=0).filled(np.nan)
+
+        return StatisticsContainer(
+            mean=pixel_mean,
+            median=pixel_median,
+            std=pixel_std,
```
kosack, line 133: You are missing a `count` value here, which makes `std` more useful.  We need to know how many events were aggregated.

kosack (changes requested):

maxnoe:
> It's also not clear that it is for images, and not any table column, so maybe ImageStatisticsAggregator?

It should support any column now

maxnoe:
Good point, while reviewing the code itself I didn't pay close enough attention to the module structure.

That certainly needs to be changed.

kosack (approved):

kosack:
We can generalize it more later, but for now this is fine

maxnoe (approved):

maxnoe (changes requested):
`monitoring/extractor.py` is a new module, it needs to be added to the API docs.

Please add it in `docs/api-reference/monitoring/` .

You can check the other files there

maxnoe:
The error is not in the new docs page (these are fine!) but in the docstring of one of the new classes:

```
 /home/runner/work/ctapipe/ctapipe/src/ctapipe/monitoring/extractor.py:docstring of ctapipe.monitoring.extractor.StatisticsExtractor.__call__:4: WARNING: py:obj reference target not found: chunk_shift
/home/runner/work/ctapipe/ctapipe/src/ctapipe/monitoring/extractor.py:docstring of ctapipe.monitoring.extractor.StatisticsExtractor.__call__:4: WARNING: py:obj reference target not found: chunk_shift
```

which are now included in the docs, which is why you didn't see them before.

You need double back-ticks in rst for code: 
```
``chunk_shift``
```

maxnoe (dismissed):

src/ctapipe/monitoring/extractor.py
```diff
@@ -0,0 +1,171 @@
+"""
+Extraction algorithms to compute the statistics from a chunk of images
```
kosack, line 2: Please add a bit more detail so the documentation is useful (this will appear in the SatisticsExtractor page of the API documentation). A new user would have no idea what this class actually does based on the current docstring.

I suggest at minimum:
```py3
"""
Algorithms to compute aggregated time-series statistics from columns of an event table.

These classes take as input an events table, divide it into time chunks, which
may optionally overlap, and computes various aggregated statistics for each
chunk, including the count, mean, median, and standard deviation. The result is
a monitoring table with columns describing the start and stop time of the chunk
and the aggregated statistic values.
"""
```

and maybe also add an example or tutorial.

kosack (changes requested):

src/ctapipe/monitoring/extractor.py
TjarkMiener: Thanks for the suggestion! While updating this, should I also consider/include the suggested renaming into `StatisticsAggregator`, `aggregator.py`, and accordingly?

docs/api-reference/monitoring/index.rst
```diff
@@ -0,0 +1,28 @@
+.. _monitoring:
+
+**********************************
+Monitoring (`~ctapipe.monitoring`)
+**********************************
+
+.. currentmodule:: ctapipe.monitoring
+
+This module include all the functions and classes needed for the Monitoring of CTA data.
```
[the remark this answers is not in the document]
mexanick: shall we then reference a computing glossary here? RST allows neat referencing and in this way we keep the definitions in one place.

src/ctapipe/calib/camera/extractor.py
kosack: as long as you are recording the count (`n_events`), it should be fine. It just changes the error bars. Of course then in any final analysis, you might want to filter on that column to remove very low values.

src/ctapipe/monitoring/extractor.py
kosack: Yes, I think that would be good to avoid confusing it with ImageExtractor.

docs/api-reference/monitoring/index.rst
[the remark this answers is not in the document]
kosack: Sure. Though that document is not in a public place yet (we don't have permission yet to make it open), so at least having the current def copied here is still useful.

src/ctapipe/monitoring/extractor.py
TjarkMiener: I updated the docs with your suggestions and renamed `extractor` to `aggregator `. For the tutorial, I planned to put together one once we have implemented the "`TwoPassCalculator`" and the "`FFactor component `".

kosack (approved):

maxnoe (approved):

PR #2566, opened 2024-06-19
Feature/custom exit codes in tools

mexanick:
Following a need in custom exit codes in `ctapipe.core.Tool`, I've implemented an extra catch of `SystemExit` that can be generated during the tool's execution phase that also performs a clean up in the same fashion as in general Exception  interception and propagates the exit code further down.

### DIFF
--- a/src/ctapipe/core/tool.py
+++ b/src/ctapipe/core/tool.py
@@ -461,7 +461,7 @@ class Tool(Application):
                     )
                     Provenance().finish_activity(
                         activity_name=self.name,
-                        status="partial_success",
+                        status="error",
                         exit_code=exit_status,
                     )
             finally:
--- a/src/ctapipe/core/tool.py
+++ b/src/ctapipe/core/tool.py
@@ -452,13 +452,18 @@ class Tool(Application):
                 if raises:
                     raise
             except SystemExit as err:
-                exit_status = err.code
-                self.log.exception("Caught SystemExit with exit code %s", exit_status)
-                Provenance().finish_activity(
-                    activity_name=self.name,
-                    status="partial_success",
-                    exit_code=exit_status,
-                )
+                if raises:
+                    raise  # do not re-intercept in tests
+                else:
+                    exit_status = err.code
+                    self.log.exception(
+                        "Caught SystemExit with exit code %s", exit_status
+                    )
+                    Provenance().finish_activity(
+                        activity_name=self.name,
+                        status="partial_success",
+                        exit_code=exit_status,
+                    )
             finally:
                 if not {"-h", "--help", "--help-all"}.intersection(self.argv):
                     self.write_provenance()
--- a/src/ctapipe/core/tool.py
+++ b/src/ctapipe/core/tool.py
@@ -1,4 +1,5 @@
 """Classes to handle configurable command-line user interfaces."""
+
 import html
 import logging
 import logging.config
@@ -445,6 +446,12 @@ class Tool(Application):
                 exit_status = 1  # any other error
                 if raises:
                     raise
+            except SystemExit as err:
+                exit_status = err.code
+                self.log.exception("Caught SystemExit with exit code %s", exit_status)
+                Provenance().finish_activity(
+                    activity_name=self.name, status="SystemExit"
+                )
             finally:
                 if not {"-h", "--help", "--help-all"}.intersection(self.argv):
                     self.write_provenance()
--- a/src/ctapipe/core/tool.py
+++ b/src/ctapipe/core/tool.py
@@ -432,25 +432,32 @@ class Tool(Application):
                 self.log.error("%s", err)
                 self.log.error("Use --help for more info")
                 exit_status = 2  # wrong cmd line parameter
+                Provenance().finish_activity(
+                    activity_name=self.name, status="error", exit_code=exit_status
+                )
                 if raises:
                     raise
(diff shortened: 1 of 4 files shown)


src/ctapipe/core/tool.py
```diff
@@ -440,11 +441,17 @@ def run(self, argv=None, raises=False):
...
+                self.log.exception("Caught unexpected exception: %s", err)
                 Provenance().finish_activity(activity_name=self.name, status="error")
                 exit_status = 1  # any other error
                 if raises:
                     raise
+            except SystemExit as err:
+                exit_status = err.code
+                self.log.exception("Caught SystemExit with exit code %s", exit_status)
+                Provenance().finish_activity(
+                    activity_name=self.name, status="SystemExit"
```
kosack, line 453: the original plan was to not allow any random exit code, only a pre-defined set.  If we need to have arbitrary exit codes (I guess the use case here is for workflow handling?), at the very least the code that was used should be stored in the provenance. Just recording "SystemExit" loses the reason for the exit happening. So we should either add a field to the provenance, or include the code in the status.  The latter is less ideal since it would require parsing, and means the sets of possible status values is not finite

mexanick: Yes, the use case here is workflow handling. And the exit codes should not be completely randomly, but rather controlled by the corresponding ICD in our case. However, given that `ctapipe` is a community product, I'm a bit hesitant to introduce such a dependency here. On the other hand, I fully agree with your point that the exit code should be stored in provenance. Here before making a move, I'd like to discuss options.

So far, the possible statuses are `["completed", "interrupted", "error"]` plus "SystemExit" I've added. I'm not sure what the subsequent purpose of this field. Is it just a human-readable label? If so, I would add another field, `exit_code` or just `code` where an actual exit code will be stored and change the list of possible statuses to `["success", "interrupted", "error", "partial_success"]` or similar. On the other hand, if the `status` field is supposed to be machine-readable, shall it just keep the exit code value?

mexanick: Perhaps, this should be default in the `_ActivityProvenance` helper class [constructor](https://github.com/cta-observatory/ctapipe/blob/2b702076effad4bcb2b9cd468b9784d7dc722ce2/src/ctapipe/core/provenance.py#L220-L230)?
```python
    def __init__(self, activity_name=sys.executable):
        self._prov = {
            "activity_name": activity_name,
            "activity_uuid": str(uuid.uuid4()),
            "start": {},
            "stop": {},
            "system": {},
            "input": [],
            "output": [],
        }
        self.name = activity_name
```

maxnoe: > sys.executable

That's always the python interpreter, right? It won't be the cli tool name even if used via an entry point installed script

mexanick: this is the default value in the current code. It is overridden in the `Tool.run()` function:
https://github.com/cta-observatory/ctapipe/blob/2b702076effad4bcb2b9cd468b9784d7dc722ce2/src/ctapipe/core/tool.py#L409

src/ctapipe/core/tool.py
```diff
@@ -425,26 +426,44 @@ def run(self, argv=None, raises=False):
...
+                if raises:
+                    raise  # do not re-intercept in tests
+                else:
+                    exit_status = err.code
+                    self.log.exception(
+                        "Caught SystemExit with exit code %s", exit_status
+                    )
+                    Provenance().finish_activity(
+                        activity_name=self.name,
+                        status="partial_success",
```
maxnoe, line 464: Why is catching a non-zero exit code here a "partial_success" and not just "error"

I don't really understand why we here need "partial_success" and it's very ambiguous. I think we should just use error here. Any semantic meaning can be then assigned downstream to the individual error codes.

mexanick: I agree that "partial_success" is ambiguous, on the other hand, my point was to differentiate from the UnknownError condition. However, I tend to agree with you that all three conditions outlined in exceptions (TraitError, UnknownError and SystemExit) are all "errors" in general meaning and the further disentanglement can (and probably should) be done by the error codes.

kosack (approved):

src/ctapipe/core/tool.py
```diff
@@ -425,26 +426,44 @@ def run(self, argv=None, raises=False):
...
-                self.log.exception(f"Caught unexpected exception: {err}")
-                Provenance().finish_activity(activity_name=self.name, status="error")
+                self.log.exception("Caught unexpected exception: %s", err)
                 exit_status = 1  # any other error
+                Provenance().finish_activity(
+                    activity_name=self.name, status="error", exit_code=exit_status
+                )
                 if raises:
                     raise
+            except SystemExit as err:
```
maxnoe, line 454: Do we need to check for 0 as a special case of success here?

mexanick: hm... why would someone call a `sys.exit(0)` inside a tool? a normal termination is done later... I think such behavior should be banned.

maxnoe: There could be an early exit success. But that would also be possible with just `return`

mexanick: yes, exactly. And I believe, this shall be the proper way to go.

maxnoe (approved):

PR #2573, opened 2024-07-03
Remove some obsolete TODOs

kosack:


src/ctapipe/calib/camera/calibrator.py
```diff
@@ -244,12 +244,6 @@ def _calibrate_dl1(self, event, tel_id):
             waveforms -= np.atleast_2d(dl1_calib.pedestal_offset)[..., np.newaxis]
 
         if n_samples == 1:
-            # To handle ASTRI and dst
```
maxnoe, line 247: I think this one still applies (see #1777 )

src/ctapipe/io/simteleventsource.py
```diff
@@ -637,10 +637,7 @@ def prepare_subarray_info(self, telescope_descriptions, header):
                     )
                 except ValueError:
                     telescope = unknown_telescope(mirror_area, n_pixels)
-
-                # TODO: switch to warning or even an exception once
-                # we start relying on this.
-                self.log.debug(
+                self.log.warning(
```
maxnoe, line 640: This will create log.warning for each telescope when analyzing prod5 data. Do we really want to enable this?

kosack: It should maybe just be at the `info` level, not warning. It is useful to show, since it makes it clear when the user is using an old data file with is lacking metadata.

src/ctapipe/calib/camera/calibrator.py
kosack: Do we still want to support DSTs? if so, should that be written here as a TODO? I would expect it to be an issue, but nothing in the code.

Tobychev (approved):

maxnoe:
@kosack the CI run fails because the job from your fork doesn't have access to `SONAR_TOKEN`.

Which is really not good and we should have noticed earlier... how to we proceed here? Convert the `SONAR_TOKEN` from a secret to a simple env variable? Doesn't seem like a good idea.

kosack:
> Which is really not good and we should have noticed earlier... how to we proceed here? Convert the SONAR_TOKEN from a secret to a simple env variable? Doesn't seem like a good idea. 

I could also just push it to upstream and re-open the PR.  But it is a general problem if people develop in their own forks. Not sure how that is usually dealt with? In larger open-source projects, I can't imagine everyone sets up sonar tokens for their own forks

maxnoe:
See the issue I created: #2577  and an attempt at a fix I found in the sonarqube forums: #2578

maxnoe:
You could leave this open here so we can test after merging #2578

maxnoe:
sonarqube reporting should work again now, let's see

maxnoe:
FINALLY

Tobychev:
@maxnoe @kosack  Bump

PR #2578, opened 2024-07-09
Run sonar in separate step to enable PRs from forks again

maxnoe:


maxnoe:
Unfortunately, we need to merge to be able to see if this works:

> Note: This event will only trigger a workflow run if the workflow file is on the default branch.

Tobychev (approved):

PR #2580, opened 2024-07-09
Add support for numpy 2.0

maxnoe:


maxnoe:
Core dependency tables does not yet support numpy 2.0, see https://github.com/PyTables/PyTables/issues/1172

maxnoe:
The remaining test failures are all due to the changed behavior of `copy=False`, which with numpy 2.0 now errors when the copy cannot be avoided.

This needs to be replaced with `copy=None`, astropy has a `COPY_AS_NEEDED` that is `None` if running under numpy 2.0 and `False` for earlier numpy versions, I'll add this here as well.

Tobychev (approved):

kosack (approved):

Issue #1881, opened 2022-04-19
Error message for non-existing config-file is suboptimal


maxnoe:
**Describe the bug**
Due to the new support for multiple config files, the error message for non-existing files is a bit verbose / confusing:

```
❯  ctapipe-process --config this/does/not/exist
2022-04-19 11:08:18,094 CRITICAL [ctapipe.ctapipe-process] (application.inner): Bad config encountered during initialization: The 'config_files' trait of a ProcessorTool instance contains  which expected <traitlets.traitlets.List object at 0x7f395f9dcd30>, not the str 'Path "/home/maxnoe/this/does/not/exist" does not exist'.
```
**Expected behavior**

Clear error message like `configuration file "<file>" does not exist`

kosack:
I guess this is because the `config_files` `Path` trait fails first, as it checks the existence before any config loading happens.  We could make a sub-class called `ConfigPath` perhaps that outputs a more useful error message?

kosack:
Ah, but it's also complaining that it is not a list... So it should somehow accept both a list and a single item.  But I thought that should work, as the latest traitlets allows lists to be specified using multiple of the same options?

kosack:
I guess it's also the fact that this is a `List(Path)` trait, not just `Path` - not sure why it complains that it is not a List though, since that should not be a problem as it should accept it as a single-item list.

maxnoe:
Just ran into this again, really, really misleading and annoying. Need to fix

maxnoe:
```
from ctapipe.core import Tool, run_tool


class TestTool(Tool):

    def setup(self):
        pass

    def start(self):
        print(self.config)



if __name__ == '__main__':
    run_tool(TestTool())
```


results in:

```
❯ python test_tool_config.py -c i_do_not_exist.yaml
{}
```

no error, just an empty config.

maxnoe:
I have a hard time debugging this, since it never actually seems to enter ctapipe code, the values seem to be lost before in traitlets (e.g. `Path.validate` is never called on the given path)

PR #2594, opened 2024-07-17
Update exception handling in tools

mexanick:
Following some real-world experience with my own changes proposed before, I believe this would be a better way to handle custom runtime exceptions with the custom exit codes.
- ~~Remove SystemExit handling in `Tool.run()`~~
  - decided to keep
- Add a possibility to register custom exceptions and handle them in `Tool.run()` with the preservation of the exit code.

### DIFF
--- a/src/ctapipe/core/tests/test_tool.py
+++ b/src/ctapipe/core/tests/test_tool.py
@@ -381,12 +381,41 @@ def test_tool_raises():
         def start(self):
             raise ValueError("1 does not equal 0.")
 
+    class CustomErrorNoExitCode(Exception):
+        pass
+
+    class CustomErrorWithExitCode(Exception):
+        exit_code = 42
+
+    class ToolCustomExceptionNoExitCode(Tool):
+        name = "CustomException"
+        description = "This tool raises a custom exception without an exit code."
+        custom_exceptions = (CustomErrorNoExitCode,)
+
+        def start(self):
+            raise CustomErrorNoExitCode("This is a custom exception.")
+
+    class ToolCustomExceptionWithExitCode(Tool):
+        name = "CustomException"
+        description = "This tool raises a custom exception with a custom exit code."
+        custom_exceptions = (CustomErrorWithExitCode,)
+
+        def start(self):
+            raise CustomErrorWithExitCode("This is a custom exception.")
+
     assert run_tool(ToolGood(), raises=True) == 0
 
     assert run_tool(ToolBad(), raises=False) == 1
 
+    assert run_tool(ToolCustomExceptionNoExitCode(), raises=False) == 1
+
+    assert run_tool(ToolCustomExceptionWithExitCode(), raises=False) == 42
+
     with pytest.raises(ValueError):
         run_tool(ToolBad(), raises=True)
+
+    with pytest.raises(CustomErrorNoExitCode):
+        run_tool(ToolCustomExceptionNoExitCode(), raises=True)
 
 
 def test_exit_stack():
--- a/src/ctapipe/core/tests/test_tool.py
+++ b/src/ctapipe/core/tests/test_tool.py
@@ -390,7 +390,6 @@ def test_tool_raises():
     class ToolCustomExceptionNoExitCode(Tool):
         name = "CustomException"
         description = "This tool raises a custom exception without an exit code."
-        custom_exceptions = (CustomErrorNoExitCode,)
 
         def start(self):
             raise CustomErrorNoExitCode("This is a custom exception.")
@@ -398,7 +397,6 @@ def test_tool_raises():
     class ToolCustomExceptionWithExitCode(Tool):
         name = "CustomException"
         description = "This tool raises a custom exception with a custom exit code."
-        custom_exceptions = (CustomErrorWithExitCode,)
 
         def start(self):
             raise CustomErrorWithExitCode("This is a custom exception.")
--- a/src/ctapipe/core/tool.py
+++ b/src/ctapipe/core/tool.py
@@ -87,6 +87,9 @@ class Tool(Application):
     executing. This happens after the ``finish`` method has run or
     in case of exceptions.
 
+    User-defined code can raise custom exceptions both in the components
+    or in the tool methods. If these custom exceptions have an `exit_code` attribute,
+    it will be propagated to the final exit code of the tool.
 
     .. code:: python
 
--- a/src/ctapipe/core/tool.py
+++ b/src/ctapipe/core/tool.py
@@ -446,10 +446,7 @@ class Tool(Application):
             except Exception as err:
                 current_exception = err
                 exit_status = getattr(err, "exit_code", 1)
-                if exit_status == 1:
-                    self.log.exception("Caught unexpected exception: %s", err)
-                else:
-                    self.log.error("Caught exception: %s", err)
+                self.log.exception("Caught unexpected exception: %s", err)
                 Provenance().finish_activity(
                     activity_name=self.name, status="error", exit_code=exit_status
                 )
--- a/src/ctapipe/core/tool.py
+++ b/src/ctapipe/core/tool.py
@@ -460,6 +460,8 @@ class Tool(Application):
                     # Finish normally
                     Provenance().finish_activity(activity_name=self.name)
                 else:
+                    if raises:
+                        raise
                     # Finish with error
                     self.log.critical(
                         "Caught SystemExit with exit code %s", exit_status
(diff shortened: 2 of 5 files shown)


src/ctapipe/core/tool.py
```diff
@@ -443,6 +449,19 @@ def run(self, argv=None, raises=False):
                 Provenance().finish_activity(
                     activity_name=self.name, status="interrupted", exit_code=exit_status
                 )
+            except self.custom_exceptions as err:
```
maxnoe, line 452: I think we should still catch the `SystemExit` with error code != 0 case here, I think this was the cause of the issue for #1881

src/ctapipe/core/tests/test_tool.py
```diff
@@ -381,13 +381,42 @@ class ToolBad(Tool):
...
+    class CustomErrorNoExitCode(Exception):
+        pass
+
+    class CustomErrorWithExitCode(Exception):
+        exit_code = 42
+
+    class ToolCustomExceptionNoExitCode(Tool):
+        name = "CustomException"
+        description = "This tool raises a custom exception without an exit code."
+        custom_exceptions = (CustomErrorNoExitCode,)
```
maxnoe, line 393: Why do we need to "register" these custom exceptions here? And if we do, why not use a mapping:

E.g.:
```
exception_code_mapping = {
    ValueError: 1,
    TypeError: 5,
}

```

src/ctapipe/core/tool.py
maxnoe, line 452: Why do we need a custom exception here? Couldn't we just do:

```
exit_status =  getattr(err, "exit_code", 1):
```
in the existing exception path?

src/ctapipe/core/tests/test_tool.py
mexanick: Exception registration is to distinguish between our own (means expected in certain circumstances) exceptions.
Considering the mapping, I believe exit codes shall be defined in the custom exceptions themselves, so writing them also here would be redundant.

src/ctapipe/core/tool.py
mexanick: To observe the difference is between "expected" (known, with specific exit code defined), and "unexpected" exceptions. If this is too narrow use case, we can follow your suggestion, but I'd prefer to distinguish between them.

mexanick: ok, SystemExit treatment is added back

maxnoe: But the handling is the exact same

maxnoe: I am really reluctant to add more complex logic to this already very complex tool that also has interactions with the traitlets base class.

maxnoe:
btw.: I simplified the `SystemExit` part here:
https://github.com/cta-observatory/ctapipe/pull/2591/files#diff-73435d23e062b16d391308f70039eb8d6bc9c40c1b12d03e9737eb4ec91b36f3R456
after realizing that `raise SystemExit` is now different than `sys.exit()`, so `raise ` doesn't make sense for it

maxnoe:
How about introducing a member variable `exit_code` on the `Tool`, so that the need for custom exceptions is not there, but tools can just do:

```
self.exit_code = 5
raise ValueError(...)
```

However, I don't really see why this is different to just doing this in the tool:

```
log.critical("Error condition foo")
self.exit(5)
```

mexanick:
The exceptions can be raised in the components that tool uses, in this case what you propose is still possible, but requires additional boiler plate try-except code in every tool. Here, for the price of a bit more complex exception handling, we can properly catch all the exceptions raised in the Tool itself or in the components it uses, and also keep the definition of exit codes tightly coupled to exceptions themselves. Adding an `exit_code` attribute to the tool seems to be redundant to me.

maxnoe:
That means a component down the line would already specify the exit code?

What if a component is used in more than one tool?

maxnoe:
I think it would be much more straight forward to catch the exception where it appears in your tool:

```
try:
    self.comp.foo()
except Exception:
     self.exception("comp failed doing foo")   
     self.exit(5)
```

mexanick:
> That means a component down the line would already specify the exit code?

I'm in favor of defining custom exceptions specifying the exit code, i.e. one exception can be reused between the components or tools if needed and the exit code is tied to this particular exception.

> What if a component is used in more than one tool?

I don't see a problem here... can you elaborate?

> I think it would be much more straight forward to catch the exception where it appears in your tool:
> 
```
try:
    self.comp.foo()
except Exception:
     self.exception("comp failed doing foo")   
     self.exit(5)
```

This is an option, but as I wrote above, would require more boiler plate code above.  Since I'm still in favor of having exit code being tied to the custom exception, this should become like this:
```
try:
    self.comp.foo()
except MyCustomException as exc:
    self.log.exception(".{}", exc)
    self.exit(exc.exit_code)\
```

which is basically the same that's added to the `Tool`'s exception handling in the PR.

mexanick:
CI failed because of broken provenance, will rebase as soon as #2595 is merged

src/ctapipe/core/tool.py
```diff
@@ -429,46 +430,50 @@ def run(self, argv=None, raises=False):
...
                 Provenance().finish_activity(activity_name=self.name)
             except (ToolConfigurationError, TraitError) as err:
+                current_exception = err
                 self.log.error("%s", err)
                 self.log.error("Use --help for more info")
                 exit_status = 2  # wrong cmd line parameter
                 Provenance().finish_activity(
                     activity_name=self.name, status="error", exit_code=exit_status
                 )
-                if raises:
```
maxnoe, line 438: Why was this raises removed?

maxnoe: Ah, I see

src/ctapipe/core/tool.py
```diff
@@ -429,46 +430,50 @@ def run(self, argv=None, raises=False):
...
+                    Provenance().finish_activity(
+                        activity_name=self.name,
+                        status="error",
+                        exit_code=exit_status,
+                    )
             finally:
                 if not {"-h", "--help", "--help-all"}.intersection(self.argv):
                     self.write_provenance()
+                if raises and current_exception:
+                    raise current_exception
```
maxnoe, line 476: Doesn't this destroy the traceback? A simple `raise` in the except block keeps the traceback chain intact, I am not sure, but I think raising the specific exception here in the finally block will destroy the traceback.

maxnoe: Ok, nevermind, seems the traceback stays intact

maxnoe: It has been a long time since a ran into a python 3 vs python 2 thing... in python 2, what I said was true.

src/ctapipe/core/tool.py
```diff
@@ -429,46 +430,50 @@ def run(self, argv=None, raises=False):
...
                 )
             except Exception as err:
-                self.log.exception("Caught unexpected exception: %s", err)
-                exit_status = 1  # any other error
+                current_exception = err
+                exit_status = getattr(err, "exit_code", 1)
+                if exit_status == 1:
+                    self.log.exception("Caught unexpected exception: %s", err)
+                else:
+                    self.log.error("Caught exception: %s", err)
```
maxnoe, line 452: Why hide the traceback here?

maxnoe: And even if it is an exception with an exit code, wouldn't it always be "unexpected" by definition?

mexanick: well... I actually have a regular case of "expected" exceptions... Not that it is a nice thing to have, but e.g. GDAS server is not responsive from time to time on a regular basis. However, since this is just a message in the log, and exception is actually known as will be reflected in its message, I will factor this out.

src/ctapipe/core/tool.py
mexanick: Yeah, it keeps the traceback, and moreover, a simple `raise` in the `finally` block will result in an error, as there will be no active exception to raise, that's why you can either have an `if raises: raise` clause in each leg, or carry the exception like this.

src/ctapipe/core/tool.py
maxnoe: Wouldn't this better be solved by something like exponential back-off inside the script?

E.g. https://pypi.org/project/retrying/

mexanick: No, as the server is sometimes unresponsive for a long-ish time and since it is not very critical, we just proceed differently.

mexanick:
@maxnoe @Tobychev can you please take a look?

Tobychev (approved):

maxnoe (changes requested):
Please add some documentation on this to the docstring of `Tool`.

maxnoe (approved):

Tobychev (approved):

kosack (approved):

Issue #2565, opened 2024-06-19
Generalise PointingInterpolator for calibration parameters


ctoennis:
**Please describe the use case that requires this feature.**
For the use cases DPPS-UC-120-2.21 (calculating and calibrating pedestals), DPPS-UC-120-2.22  (calculating and calibrating gain), and DPPS-UC-120-2.28 (Pointing corrections) calibration parameters will be generated that will need some form of interpolation similar to the basic pointing information (linear interpolation for the pointing correction and a sort of step function-like interpolation for gain and pedestals). 

**Describe the solution you'd like**
I would like to modify the PointingInterpolator into a more general interpolator class that is told which data to interpolate and which interpolation method is to be used.

**Describe alternatives you've considered**
Alternatively we can have a new interpolator class for the pedestals and gain as they need a different method of interpolation and simply just modify the PointingInterpolator to also handle pointing corrections.

**Additional context**
Links to the use cases:
[https://gitlab.cta-observatory.org/cta-computing/dpps/calibrationpipeline/documentation/use-case-registry/-/blob/main/TelescopeCalibration/CameraCalibration/DPPS-UC-120-2.21.md?ref_type=heads](Pedestals)
[https://gitlab.cta-observatory.org/cta-computing/dpps/calibrationpipeline/documentation/use-case-registry/-/blob/main/TelescopeCalibration/CameraCalibration/DPPS-UC-120-2.22.md?ref_type=heads](gain)
[https://gitlab.cta-observatory.org/cta-computing/dpps/calibrationpipeline/documentation/use-case-registry/-/blob/usecase/PointingCalibration/TelescopeCalibration/CameraCalibration/DPPS-UC-120-2.28.md?ref_type=heads](Pointing)

PR #2600, opened 2024-07-23
Interpolator

ctoennis:
This PR is for extending the pointing interpolator to also handle pedestal and gain. This request is related to issue [#2565](https://github.com/cta-observatory/ctapipe/issues/2565). 

A new interpolator is added based on the searchsorted function in numpy and set up to accept the same options as interp1d class from numpy that is already in use for the pointing interpolation. This interpolator is creating a step function that applies a pedestal or gain calibration after a corresponding timestamp until the next pedestal or gain value is available. 

The new interpolator is able to interpolate array-like objects to handle 

It is currently set up to automatically recognise what data is interpolated and picks an appropriate interpolator based on that determination. This can be changed to require a choice when the data that is to be interpolated is added.

The pointing interpolator has a location for the pointing data hardcoded into it that is used by default for now to avoid having to change code elsewhere. This scheme can be extended to have the place of pointing corrections, pedestals or gain at hand as well.

Tests have been included that check that the step interpolation works as expected.

### DIFF
--- a/src/ctapipe/io/interpolating.py
+++ b/src/ctapipe/io/interpolating.py
@@ -106,10 +106,8 @@ class Interpolator(Component):
             self.interp_options["bounds_error"] = False
             self.interp_options["fill_value"] = np.nan
 
-        self._alt_interpolators = {}
-        self._az_interpolators = {}
-        self._ped_interpolators = {}
-        self._gain_interpolators = {}
+        self._interpolators = {}
+        self._secondary_interpolators = {}
 
     def add_table(self, tel_id, input_table):
         """
@@ -167,15 +165,17 @@ class Interpolator(Component):
             az = np.unwrap(az)
             alt = input_table["altitude"].quantity.to_value(u.rad)
             mjd = input_table["time"].tai.mjd
-            self._az_interpolators[tel_id] = interp1d(mjd, az, **self.interp_options)
-            self._alt_interpolators[tel_id] = interp1d(mjd, alt, **self.interp_options)
+            self._interpolators[tel_id] = interp1d(mjd, az, **self.interp_options)
+            self._secondary_interpolators[tel_id] = interp1d(
+                mjd, alt, **self.interp_options
+            )
 
         elif self.parameter_type == "pedestal":
             time = input_table["time"]
 
             ped = input_table["pedestal"]
 
-            self._ped_interpolators[tel_id] = StepInterpolator(
+            self._interpolators[tel_id] = StepInterpolator(
                 time, ped, **self.interp_options
             )
 
@@ -184,7 +184,7 @@ class Interpolator(Component):
 
             gain = input_table["gain"]
 
-            self._gain_interpolators[tel_id] = StepInterpolator(
+            self._interpolators[tel_id] = StepInterpolator(
                 time, gain, **self.interp_options
             )
 
@@ -192,11 +192,11 @@ class Interpolator(Component):
         if table_location is None:
             table_location = f"/dl0/monitoring/telescope/pointing/tel_{tel_id:03d}"
 
-        pointing_table = read_table(
+        input_table = read_table(
             self.h5file,
             table_location,
         )
-        self.add_table(tel_id, pointing_table)
+        self.add_table(tel_id, input_table)
 
     def __call__(self, tel_id, time):
         """
@@ -217,16 +217,7 @@ class Interpolator(Component):
             interpolated azimuth angle
         """
 
-        if not any(
-            [
-                tel_id in x
-                for x in (
-                    self._az_interpolators,
-                    self._ped_interpolators,
-                    self._gain_interpolators,
-                )
-            ]
-        ):
+        if tel_id not in self._interpolators:
             if self.h5file is not None:
                 self._read_parameter_table(tel_id)
             else:
@@ -234,14 +225,12 @@ class Interpolator(Component):
 
         if self.parameter_type == "pointing":
             mjd = time.tai.mjd
-            az = u.Quantity(self._az_interpolators[tel_id](mjd), u.rad, copy=False)
-            alt = u.Quantity(self._alt_interpolators[tel_id](mjd), u.rad, copy=False)
+            az = u.Quantity(self._interpolators[tel_id](mjd), u.rad, copy=False)
+            alt = u.Quantity(
+                self._secondary_interpolators[tel_id](mjd), u.rad, copy=False
+            )
             return alt, az
 
-        elif self.parameter_type == "pedestal":
-            ped = self._ped_interpolators[tel_id](time)
-            return ped
-
-        elif self.parameter_type == "gain":
-            gain = self._gain_interpolators[tel_id](time)
-            return gain
+        elif self.parameter_type in ("pedestal", "gain"):
+            cal = self._interpolators[tel_id](time)
+            return cal
--- a/src/ctapipe/io/interpolating.py
+++ b/src/ctapipe/io/interpolating.py
@@ -117,10 +117,12 @@ class Interpolator(Component):
         ----------
         tel_id : int
             Telescope id
-        pointing_table : astropy.table.Table
+        input_table : astropy.table.Table
             Table of pointing values, expected columns
             are ``time`` as ``Time`` column, ``azimuth`` and ``altitude``
-            as quantity columns.
+            as quantity columns for pointing and pointing correction data.
+            For pedestal data the quantity column is expected as
+            ``pedestal`` and for gain data as ``gain``.
         """
 
         if "gain" in set(input_table.colnames):
@@ -188,8 +190,8 @@ class Interpolator(Component):
                 time, gain, **self.interp_options
             )
 
-    def _read_parameter_table(self, tel_id, table_location=None):
-        if table_location is None:
+    def _read_parameter_table(self, tel_id, table_location="pointing"):
+        if table_location == "pointing":
             table_location = f"/dl0/monitoring/telescope/pointing/tel_{tel_id:03d}"
 
         input_table = read_table(
--- a/src/ctapipe/io/interpolating.py
+++ b/src/ctapipe/io/interpolating.py
@@ -1,0 +1,247 @@
+from typing import Any
+
+import astropy.units as u
+import numpy as np
+import tables
+from astropy.time import Time
+from scipy.interpolate import interp1d
+
+from ctapipe.core import Component, traits
+
+from .astropy_helpers import read_table
+
+
+class StepInterpolator:
+
+    """
+    Step function Interpolator for the gain and pedestals
+
+    Parameters
+    ----------
+    values : None | np.array
+        Numpy array of the data that is to be interpolated.
+        The first dimension needs to be an index over time
+    times : None | np.array
+        Time values over which data are to be interpolated
+        need to be sorted and have same length as first dimension of values
+    """
+
+    def __init__(
+        self,
+        times,
+        values,
+        bounds_error=True,
+        fill_value="extrapolate",
+        assume_sorted=True,
+        copy=False,
+    ):
+        self.values = values
+        self.times = times
+        self.bounds_error = bounds_error
+        self.fill_value = fill_value
+
+    def __call__(self, point):
+        if point < self.times[0]:
+            if self.bounds_error:
+                raise ValueError("below the interpolation range")
+
+            if self.fill_value == "extrapolate":
+                return self.values[0]
+
+            else:
+                a = np.empty(self.values[0].shape)
+                a[:] = np.nan
+                return a
+
+        else:
+            i = np.searchsorted(self.times, point, side="left")
+            return self.values[i - 1]
+
+
+class Interpolator(Component):
+    """
+    Interpolate pointing and calibration parameters from a monitoring table to a given timestamp.
+
+    Parameters
+    ----------
+    h5file : None | tables.File
+        A open hdf5 file with read access.
+    table_location: | str
+        location where the monitoring data is expected to be stored in that file
+    interpolation_method: | str
+        method of interpolation to be used
+    """
+
+    bounds_error = traits.Bool(
+        default_value=True,
+        help="If true, raises an exception when trying to extrapolate out of the given table",
+    ).tag(config=True)
+
+    extrapolate = traits.Bool(
+        help="If bounds_error is False, this flag will specify whether values outside"
+        "the available values are filled with nan (False) or extrapolated (True).",
+        default_value=False,
+    ).tag(config=True)
+
+    def __init__(
+        self, h5file=None, table_location=None, interpolation_method=None, **kwargs
+    ):
+        super().__init__(**kwargs)
+
+        if h5file is not None and not isinstance(h5file, tables.File):
+            raise TypeError("h5file must be a tables.File")
+        self.h5file = h5file
+
+        if table_location is not None and not isinstance(table_location, str):
+            raise TypeError("table_location must be a string")
+        self.table_location = table_location
+
+        self.interp_options: dict[str, Any] = dict(assume_sorted=True, copy=False)
+        if self.bounds_error:
+            self.interp_options["bounds_error"] = True
+        elif self.extrapolate:
+            self.interp_options["bounds_error"] = False
+            self.interp_options["fill_value"] = "extrapolate"
+        else:
+            self.interp_options["bounds_error"] = False
+            self.interp_options["fill_value"] = np.nan
+
+        self._alt_interpolators = {}
+        self._az_interpolators = {}
+        self._ped_interpolators = {}
+        self._gain_interpolators = {}
+
+    def add_table(self, tel_id, input_table):
+        """
+        Add a table to this interpolator
+
+        Parameters
+        ----------
+        tel_id : int
+            Telescope id
+        pointing_table : astropy.table.Table
+            Table of pointing values, expected columns
+            are ``time`` as ``Time`` column, ``azimuth`` and ``altitude``
+            as quantity columns.
+        """
+
+        if "gain" in set(input_table.colnames):
+            self.parameter_type = "gain"
+            # here i want to detect if the table contains pointing, gain or pedestal information
+            # can this be a loop? Should this be a function?
+            missing = {"time", "gain"} - set(input_table.colnames)
+            if len(missing) > 0:
+                raise ValueError(f"Table is missing required column(s): {missing}")
+
+        elif "pedestal" in set(input_table.colnames):
+            self.parameter_type = "pedestal"
+            missing = {"time", "pedestal"} - set(input_table.colnames)
+            if len(missing) > 0:
+                raise ValueError(f"Table is missing required column(s): {missing}")
+
+        elif "azimuth" in set(input_table.colnames):
+            self.parameter_type = "pointing"
+            missing = {"time", "azimuth", "altitude"} - set(input_table.colnames)
+            if len(missing) > 0:
+                raise ValueError(f"Table is missing required column(s): {missing}")
+
+            for col in ("azimuth", "altitude"):
+                unit = input_table[col].unit
+                if unit is None or not u.rad.is_equivalent(unit):
+                    raise ValueError(f"{col} must have units compatible with 'rad'")
+
+            if not isinstance(input_table["time"], Time):
+                raise TypeError(
+                    "'time' column of pointing table must be astropy.time.Time"
+                )
+        # sort first, so it's not done twice for each interpolator
+        input_table.sort("time")
+        if self.parameter_type == "pointing":
+            az = input_table["azimuth"].quantity.to_value(u.rad)
+            # prepare azimuth for interpolation by "unwrapping": i.e. turning
+            # [359, 1] into [359, 361]. This assumes that if we get values like
+            # [359, 1] the telescope moved 2 degrees through 0, not 358 degrees
+            # the other way around. This should be true for all telescopes given
+            # the sampling speed of pointing values and their maximum movement speed.
+            # No telescope can turn more than 180° in 2 seconds.
+            az = np.unwrap(az)
+            alt = input_table["altitude"].quantity.to_value(u.rad)
+            mjd = input_table["time"].tai.mjd
+            self._az_interpolators[tel_id] = interp1d(mjd, az, **self.interp_options)
+            self._alt_interpolators[tel_id] = interp1d(mjd, alt, **self.interp_options)
+
+        elif self.parameter_type == "pedestal":
+            time = input_table["time"]
+
+            ped = input_table["pedestal"]
+
+            self._ped_interpolators[tel_id] = StepInterpolator(
+                time, ped, **self.interp_options
+            )
+
+        elif self.parameter_type == "gain":
+            time = input_table["time"]
+
+            gain = input_table["gain"]
+
+            self._gain_interpolators[tel_id] = StepInterpolator(
+                time, gain, **self.interp_options
+            )
+
+    def _read_parameter_table(self, tel_id, table_location=None):
+        if table_location is None:
+            table_location = f"/dl0/monitoring/telescope/pointing/tel_{tel_id:03d}"
+
+        pointing_table = read_table(
+            self.h5file,
+            table_location,
+        )
+        self.add_table(tel_id, pointing_table)
+
+    def __call__(self, tel_id, time):
+        """
+        Interpolate alt/az for given time and tel_id.
+
+        Parameters
+        ----------
+        tel_id : int
+            telescope id
+        time : astropy.time.Time
+            time for which to interpolate the pointing
+
+        Returns
+        -------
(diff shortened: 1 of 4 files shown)


Tobychev:
Why do you put everything into one big constructor instead of having a base class and with several more specialised children? Is there some specific technical reason for why you want to entirely eliminate the `PointingInterpolator` instead of adding more classes for the case you want to cover?

maxnoe:
@ctoennis I think we need separate classes here, per thing that can be interpolated, putting common code into free function or a common base class.

This has to do with our configuration system: it configures things by class name. So to be able to have two interpolators with different configurations (e.g. for input file from which it reads the monitoring table), we need different class names.

ctoennis:
@maxnoe im doing that right now

ctoennis:
I split it up into a parent and 2 child classes. I am now updating the docustrings.

src/ctapipe/io/interpolating.py
```diff
@@ -0,0 +1,242 @@
...
+                a = np.empty(self.values[0].shape)
+                a[:] = np.nan
+                return a
+
+        else:
+            i = np.searchsorted(self.times, point, side="left")
+            return self.values[i - 1]
+
+
+class Interpolator(Component):
```
maxnoe, line 61: This should probably be an abstract base class

src/ctapipe/io/interpolating.py
```diff
@@ -0,0 +1,242 @@
...
+        else:
+            self.interp_options["bounds_error"] = False
+            self.interp_options["fill_value"] = np.nan
+
+        self._interpolators = {}
+        self._secondary_interpolators = {}
+
+    def add_table(self, tel_id, input_table):
+        # sort first, so it's not done twice for each interpolator
+        input_table.sort("time")
```
maxnoe, line 104: This doesn't do anything here. Remove all code from this method and make it an `@abstractmethod`

src/ctapipe/io/interpolating.py
```diff
@@ -0,0 +1,242 @@
...
+        # No telescope can turn more than 180° in 2 seconds.
+        az = np.unwrap(az)
+        alt = input_table["altitude"].quantity.to_value(u.rad)
+        mjd = input_table["time"].tai.mjd
+        self._interpolators[tel_id] = interp1d(mjd, az, **self.interp_options)
+        self._secondary_interpolators[tel_id] = interp1d(
+            mjd, alt, **self.interp_options
+        )
+
+    def _read_parameter_table(self, tel_id):
```
maxnoe, line 184: This could probably be put into the base class, if we add a class attribute `table_path` this:

```
class PointingInterpolator(Interpolator):
    table_path = "/dl0/monitoring/telescope/pointing"

class PedestalInterpolator(Interpolator):
    table_path = "/dl0/monitoring/telescope/pedestal"
```

src/ctapipe/io/interpolating.py
```diff
@@ -0,0 +1,242 @@
...
+        time : astropy.time.Time
+            time for which to interpolate the pointing
+
+        Returns
+        -------
+        cal : array [float]
+            interpolated calibration quantity
+        """
+
+        if tel_id not in self._interpolators:
```
maxnoe, line 210: This is probably the same for all subclasses, so it could be put in the base class.

src/ctapipe/io/interpolating.py
maxnoe: which means adding `meta=ABCMeta` as second argument here.

src/ctapipe/io/interpolating.py
```diff
@@ -0,0 +1,253 @@
...
+            raise ValueError(f"Table is missing required column(s): {missing}")
+        for col in ("azimuth", "altitude"):
+            unit = input_table[col].unit
+            if unit is None or not u.rad.is_equivalent(unit):
+                raise ValueError(f"{col} must have units compatible with 'rad'")
+
+        if not isinstance(input_table["time"], Time):
+            raise TypeError("'time' column of pointing table must be astropy.time.Time")
+
+        input_table.sort("time")
```
kosack, line 172: This is a side effect that I guess was there before this PR, but should be documented:  the input table gets sorted in-place when this function is called, which may be unexpected.  I guess at least we should add a note in the docstring that this function modifies the inputs.

kosack (changes requested):
general comment: I would name the python file `interpolation.py` instead of `interpolating` to keep the grammar consistent. Then it's also similar to `scipy.interpolation`

src/ctapipe/io/interpolating.py
```diff
@@ -0,0 +1,253 @@
...
+        assume_sorted=True,
+        copy=False,
+    ):
+        self.values = values
+        self.times = times
+        self.bounds_error = bounds_error
+        self.fill_value = fill_value
+
+    def __call__(self, point):
+        if point < self.times[0]:
```
maxnoe, line 45: It seems you don't check for `point > self.times[-1]` here, i.e. the other out of bounds condition.

src/ctapipe/io/interpolating.py
```diff
@@ -0,0 +1,253 @@
...
+
+    def __call__(self, tel_id, time):
+        if tel_id not in self._interpolators:
+            if self.h5file is not None:
+                self._read_parameter_table(tel_id)  # might need to be removed
+            else:
+                raise KeyError(f"No table available for tel_id {tel_id}")
+
+    @abstractmethod
+    def _read_parameter_table(self, tel_id):
```
maxnoe, line 115: This is not what I meant when I wrote: "move this method to the base class".

I meant actually moving it here, with the code. It should be common between subclasses (no need to further specialize) if we define which table to load as a class variable.

src/ctapipe/io/interpolating.py
```diff
@@ -0,0 +1,253 @@
...
+
+        Returns
+        -------
+        altitude : astropy.units.Quantity[deg]
+            interpolated altitude angle
+        azimuth : astropy.units.Quantity[deg]
+            interpolated azimuth angle
+        """
+
+        super().__call__(tel_id, time)
```
maxnoe, line 140: This is not really how inheritance is supposed to work. 

It's a bit weird to have `super().__call__` with a different API (not returning anything), doing only part of the work instead of a different variant of the work.

This would better be in a differently named method used here than by invoking the same method of the parent class.

src/ctapipe/io/interpolating.py
```diff
@@ -0,0 +1,253 @@
...
+        if self.bounds_error:
+            self.interp_options["bounds_error"] = True
+        elif self.extrapolate:
+            self.interp_options["bounds_error"] = False
+            self.interp_options["fill_value"] = "extrapolate"
+        else:
+            self.interp_options["bounds_error"] = False
+            self.interp_options["fill_value"] = np.nan
+
+        self._interpolators = {}
```
maxnoe, line 100: I think this is bad design. This forces that we can only ever interpolate two things and we loose the descriptive names.

Also, alt and az are equally important, so it's weird to have one be "secondary"

src/ctapipe/io/interpolating.py
maxnoe: If we need this on the base class, then we should just use a dict of dicts, e.g.:

```
self.interpolators[tel_id]["azimuth"]
self.interpolators[tel_id]["altitude"]
```

src/ctapipe/io/interpolating.py
maxnoe: Probably best just add a `.copy()` so that we actually own the data and don't mutate someone else's table

src/ctapipe/io/interpolating.py
ctoennis: The tables tjarks Statistics extractors will produce do not keep track of the end at which the data is to be used, so it is not clear how to define an upper bound here. An upper bound could be parsed separately as the last point is not where the interpolation should end.

maxnoe: If you are talking about #2554 , the latest version does include start and end time in each chunk

src/ctapipe/io/interpolating.py
maxnoe: If we need this on the base class, then we should just use a dict of dicts, e.g.:

```
self.interpolators[tel_id]["azimuth"]
self.interpolators[tel_id]["altitude"]
```

src/ctapipe/io/interpolation.py
```diff
@@ -83,6 +74,15 @@ class Interpolator(Component):
     table_location = None
 
     def __init__(self, h5file=None, **kwargs):
```
maxnoe, line 76: The docstring should be add the class, not at `__init__`

src/ctapipe/io/interpolation.py
```diff
@@ -0,0 +1,259 @@
...
+                a = np.empty(self.values[0].shape)
+                a[:] = np.nan
+                return a
+
+        else:
+            i = np.searchsorted(self.times, point, side="left")
+            return self.values[i - 1]
+
+
+class Interpolator(Component):
```
maxnoe, line 62: add `metaclass=ABCMeta`

src/ctapipe/io/interpolation.py
```diff
@@ -0,0 +1,259 @@
...
+            self.interp_options["bounds_error"] = False
+            self.interp_options["fill_value"] = "extrapolate"
+        else:
+            self.interp_options["bounds_error"] = False
+            self.interp_options["fill_value"] = np.nan
+
+        self._interpolators = {}
+
+    @abstractmethod
+    def add_table(self, tel_id, input_table):
```
maxnoe, line 105: This needs a docstring telling developers what needs to be implemented in the child classes.

src/ctapipe/io/interpolation.py
```diff
@@ -0,0 +1,259 @@
...
+                self._read_parameter_table(tel_id)  # might need to be removed
+            else:
+                raise KeyError(f"No table available for tel_id {tel_id}")
+
+    def _read_parameter_table(self, tel_id):
+        input_table = read_table(
+            self.h5file,
+            self.table_location + f"/tel_{tel_id:03d}",
+        )
+        print(self.table_location + f"/tel_{tel_id:03d}")
```
maxnoe, line 120: remove print

src/ctapipe/io/interpolation.py
```diff
@@ -0,0 +1,259 @@
...
+        if tel_id not in self._interpolators:
+            if self.h5file is not None:
+                self._read_parameter_table(tel_id)  # might need to be removed
+            else:
+                raise KeyError(f"No table available for tel_id {tel_id}")
+
+    def _read_parameter_table(self, tel_id):
+        input_table = read_table(
+            self.h5file,
+            self.table_location + f"/tel_{tel_id:03d}",
```
maxnoe, line 118: Better just use one f-string: `f"{self.table_location}/tel_{tel_id:03d}"`

src/ctapipe/io/interpolation.py
```diff
@@ -0,0 +1,259 @@
...
+        ----------
+        tel_id : int
+            Telescope id
+        input_table : astropy.table.Table
+            Table of pointing values, expected columns
+            are ``time`` as ``Time`` column, ``azimuth`` and ``altitude``
+            as quantity columns for pointing and pointing correction data.
+        """
+
+        missing = {"time", "azimuth", "altitude"} - set(input_table.colnames)
```
maxnoe, line 171: I think we can de-duplicate a bit of code here by adding a `required_columns` and `expected_units` argument on the classes and create a `_check_table` function on the base class.

ctoennis: the calibration data will not have units, so when interpolating them we don't expect them to have a specific unit

maxnoe: Sure, then just leave it empty:

```
expected_units = {}
```

the idea would be that the `_check_table` function checks for  each column in `expected_units` that the units are compatible.

You could even make sure to check that there are in fact no units, by doing:

```
expected_units = {"gain": None}
```

src/ctapipe/io/interpolation.py
```diff
@@ -0,0 +1,283 @@
...
+        # No telescope can turn more than 180° in 2 seconds.
+        az = np.unwrap(az)
+        alt = input_table["altitude"].quantity.to_value(u.rad)
+        mjd = input_table["time"].tai.mjd
+        self._interpolators[tel_id] = {}
+        self._interpolators[tel_id]["az"] = interp1d(mjd, az, **self.interp_options)
+        self._interpolators[tel_id]["alt"] = interp1d(mjd, alt, **self.interp_options)
+
+
+class CalibrationInterpolator(Interpolator):
```
kosack, line 223: General question: I would expect the flat-fielding corrections to vary on a slower rate than the pedetals, which can vary with NSB level and therefore zenith angle.  Is it really the  best idea to have a single table with both of those? That would imply they are measured and vary at the same rate.   Wouldn't it be better to have a `PedestalInterpolator` and a `FlatfieldInterolator` separately?

src/ctapipe/io/interpolation.py
```diff
@@ -0,0 +1,283 @@
...
+
+from ctapipe.core import Component, traits
+
+from .astropy_helpers import read_table
+
+
+class StepFunction:
+
+    """
+    Step function Interpolator for the gain and pedestals
```
kosack, line 18: Some more description would be useful here.  What is meant by "step function"? I guess you mean it is similar to the "nearest" option in `interp1d`, but only backward in time? (e.g. only the nearest previous measurement?)

src/ctapipe/io/interpolation.py
```diff
@@ -0,0 +1,283 @@
...
+        help="If true, raises an exception when trying to extrapolate out of the given table",
+    ).tag(config=True)
+
+    extrapolate = traits.Bool(
+        help="If bounds_error is False, this flag will specify whether values outside"
+        "the available values are filled with nan (False) or extrapolated (True).",
+        default_value=False,
+    ).tag(config=True)
+
+    table_location = None
```
kosack, line 83: Should add a docstring or comment that these *class variables* should be set by sub-classes. You could instead _use instance variables_ and  require them to be set in the `__init__` instead of using class variables, and then the sub-classes would look like:

```py3
class SomeOtherInterpolator(Interpolator):
    def __init__(self):
         super().__init__(self, table_location="/dl0/something/table")
         ...
```
Not sure which is more clear, but using class variables is less common.   @maxnoe do you have a guideline here on when we should use class variables for defaults vs instance?

src/ctapipe/io/interpolation.py
```diff
@@ -0,0 +1,283 @@
...
+        if tel_id not in self._interpolators:
+            if self.h5file is not None:
+                self._read_parameter_table(tel_id)  # might need to be removed
+            else:
+                raise KeyError(f"No table available for tel_id {tel_id}")
+
+    def _read_parameter_table(self, tel_id):
+        input_table = read_table(
+            self.h5file,
+            f"{self.table_location}/tel_{tel_id:03d}",
```
kosack, line 147: Somewhat confusing as now `table_location` is not the location of the table, but rather the location of the **group** that contains the set of tables.  Maybe we should use a better name?

src/ctapipe/io/interpolation.py
```diff
@@ -0,0 +1,283 @@
...
+        par_name : str
+            Name of the parameter that is to be interpolated
+            ``pedestal`` is used for pedestals, ``gain`` for gain
+            can also be the name of statistical parameters to
+            interpolate the content of StatisticsContainers
+        """
+
+        self.par_name = par_name
+
+        self.required_columns = {"time", par_name}
```
kosack, line 273: this shadows the class variable `Interpolator.required_columns` with an instance variable of the same name.  Is that intended? I would expect it to cause unexpected behavior/bugs in the future...  (maybe more argument for not using class variables for this)

kosack (changes requested):

src/ctapipe/io/interpolation.py
maxnoe: I think this should be a class variable and not changeable by instance.

E.g. the path in the h5file is fixed for `PointingInterpolator`, `PedestalInterpolator`, whatever we come up with.

maxnoe: It's a question of when this data changes, in the design I had in mind, each of the different interpolators looks to a fixed table path → class variable.

src/ctapipe/io/interpolation.py
maxnoe: `telescope_data_group`?

src/ctapipe/io/interpolation.py
maxnoe: No, certainly not. 

At minimum, it should extend the class variable, not shadow it.

I also don't understand how this "parname" is supposed to work at all.

In the pointing interpolator, `add_table` sets up all necessary interpolators. Here, it seems, it only sets up one column per call of `add_table`, which I think is not how it should be.


`add_table` should setup all interpolators relevant for this class.

ctoennis:
So, i split the Calibration Interpolator into a pedestal and gain interpolator. This way the variable par_name is not needed anymore.

src/ctapipe/io/interpolation.py
kosack: Ok.  Then just have to fix the `add_table` method to not step on this

src/ctapipe/io/interpolation.py
```diff
@@ -267,25 +269,74 @@ def add_table(self, tel_id, input_table, par_name="pedestal"):
...
-        par_name : str
-            Name of the parameter that is to be interpolated
-            ``pedestal`` is used for pedestals, ``gain`` for gain
-            can also be the name of statistical parameters to
-            interpolate the content of StatisticsContainers
+            are ``time`` as ``Time`` column and "gain"
+            for the relative gain data
+        """
+
+        self.required_columns = {"time", "gain"}
```
maxnoe, line 276: This is wrong here

src/ctapipe/io/interpolation.py
```diff
@@ -267,25 +269,74 @@ def add_table(self, tel_id, input_table, par_name="pedestal"):
...
+        tel_id : int
+            Telescope id
+        input_table : astropy.table.Table
+            Table of pointing values, expected columns
+            are ``time`` as ``Time`` column and "pedestal"
+            for the pedestal data
+        """
 
-        self.required_columns = {"time", par_name}
+        self.required_columns = {"time", "pedestal"}
```
maxnoe, line 332: and here as well

src/ctapipe/io/interpolating.py
kosack: Maybe some memory overhead with that though... the  input tables could be large.  I guess there is no way make a "sorted view" of the table.

src/ctapipe/io/interpolation.py
```diff
@@ -228,12 +230,12 @@ def add_table(self, tel_id, input_table):
         self._interpolators[tel_id]["alt"] = interp1d(mjd, alt, **self.interp_options)
 
 
-class CalibrationInterpolator(Interpolator):
+class GainInterpolator(Interpolator):
```
kosack, line 233: I don't like the name GainInterpolator... There are many "gains" we deal with, and the names are often used incorrectly as "any multiplicative factor":
-  the HV gain setting (the only real "gain")
-  the "gain channel" (high or low)
- the DC/PE conversion factor (sometimes also called relative gain correction, which ensures a PE in one camera is a PE in another)
- the inter-pixel flat-field correction (which ensures the pixel response is flat across the camera as measured using a uniform light source).  

If this class is interpolating the flat-fielding values, please call it `FlatFieldInterpolator` to avoid confusion.   We also may in the future want to interpolate these other gain-correction values.  That also means that the name in the ctapipe output file should not be `gain` either, but `/dl1/calibration/flatfield/tel_x`.

kosack (changes requested):

src/ctapipe/io/interpolation.py
mexanick: should this be a list?

maxnoe: Since the check is an intersection, a set is best.

You can use a `frozenset`

src/ctapipe/io/interpolation.py
mexanick: I think we shall use dimensionless units instead of None. This will allow to simplify the validation a little bit too.
```
>>> from astropy.units import Unit
>>> dimensionless_unit = Unit()
>>> dimensionless_unit
Unit(dimensionless)
```

src/ctapipe/io/interpolation.py
```diff
@@ -0,0 +1,342 @@
...
+        ----------
+        tel_id : int
+            Telescope id
+        input_table : astropy.table.Table
+            Table of pointing values, expected columns
+            are ``time`` as ``Time`` column and "gain"
+            for the flatfield data
+        """
+
+        self.required_columns = frozenset(["time", "gain"])  # TBD
```
kosack, line 276: this still shadows the class variable with an instance variable. I think you should not ever modify the `required_columns` outside of the class definition.  If you need another set of columns, you should create a new variable for that.  But here, I think you can just remove this line, right? It does nothing that coudn't be set in the class.

kosack (changes requested):

ctoennis:
sorry for th oversight with the class variable. I moved it to the definition and made it so the other calibration data is required to have a dimensionless unit.

src/ctapipe/io/interpolation.py
```diff
@@ -0,0 +1,344 @@
...
+            are ``time`` as ``Time`` column, ``azimuth`` and ``altitude``
+            as quantity columns for pointing and pointing correction data.
+        """
+
+        self._check_tables(input_table)
+
+        if not isinstance(input_table["time"], Time):
+            raise TypeError("'time' column of pointing table must be astropy.time.Time")
+
+        input_table.sort("time")
```
kosack, line 218: As we mentioned earlier, you should probably make a copy of the table here to avoid the unexpected side-effect that calling this function modifies the input table. So just change it to:
```python
input_table = input_table.copy().sort()
```
That may increase memory usage a small amount, but as the copy goes out of scope, I guess it should only retain the numpy arrays used in interp1d, so it shouldn't be a problem.

kosack (changes requested):
one more minor change (sorry), but otherwise, looks ready!

kosack (dismissed):

kosack:
Looks like all comments from @maxnoe are addressed, but he is on vacation, so maybe @mexanick you can do a final review and we can merge?

src/ctapipe/io/interpolation.py
```diff
@@ -11,6 +11,8 @@
 
 from .astropy_helpers import read_table
 
+dimensionless_unit = u.Unit()
```
maxnoe, line 14: There e is no need to define this.

There is `u.one` as short  alias or `u.dimensionless_unscaled` if you want to be explicit.

maxnoe (changes requested):
It seems that the Flat field interpolator is not clear / final. It has some TBD comments, but most importantly, it doesn't match the definition of the dl1 calibration container.

I think this should match, right? The interpolator fills the values to be used for the calibration.

ctoennis:
> It seems that the Flat field interpolator is not clear / final. It has some TBD comments, but most importantly, it doesn't match the definition of the dl1 calibration container.
> 
> I think this should match, right? The interpolator fills the values to be used for the calibration.

So, the pedestal and flatfield calibration data will in the end have different time values and the interpolator will have to load each type of data from a different table. To me it makes sense to therefore have a different interpolator for those fields in the DL1CameraCalibrationContainer. 

The TBD comments relate to the location where the input groups for the different types of the calibration data will be stored. This would be settled when we add the CalibrationCalculators. Some of these, like the PointingCalculator will need to have the interpolator there to interpolate gain or pedestal data for their calculations. I think it would make sense to leave these fields with the TBD comments and fix this when we implement the CalibrationCalculators.

maxnoe:
I think it would make more sense then to only introduce the base class in this PR here and change the pointing interpolator.

The other interpolators can be added once the TBDs are clear. It doesn't make much sense to have interpolators in main for which no input data exists.

ctoennis:
i will move the other interpolators into the branch me and tjark have for the CalibrationCalculators then and remove them here. Does that make sense?

kosack (dismissed):

docs/changes/2600.feature.rst
```diff
@@ -0,0 +1,2 @@
+Add Interpolator class to generalize the PointingInterpolator in the io collection.
+Add new interpolator class called StepInterpolator to be used for calibration data.
```
kosack, line 2: maybe remove the mention of StepInterpolator now that it has moved (otherwise this will end up in the release notes).

ctoennis: yes, i will do that

kosack (approved):

ctoennis:
@maxnoe can you have look again? i think it should be good to merge now.

maxnoe (approved):

PR #2604, opened 2024-07-26
OutlierDetector components

TjarkMiener:
This PR added outlier detection approaches. The sub classes can be applied to any astropy column of shape (n_entries, n_channels, n_pix) returning an outlier mask of the same shape. The code utilizing this components later need to handle multiple instances and therefore combine the returning outlier masks.

–––
Related: #2542


src/ctapipe/calib/camera/outlier.py
```diff
@@ -0,0 +1,112 @@
+"""
+Outlier detection algorithms to identify faulty pixels
+"""
+
+__all__ = [
+    "OutlierDetection",
```
mexanick, line 6: Semantically, I'd use `*Detector` instead of `*Detection`

src/ctapipe/calib/camera/outlier.py
```diff
@@ -0,0 +1,112 @@
...
+        -------
+        boolean mask
+            mask of outliers of shape (n_entries, n_channels, n_pix)
+        """
+        pass
+
+
+class RangeBasedOutlierDetection(OutlierDetection):
+    """
+    Remove outliers based on a valid range.
```
mexanick, line 59: This doesn't remove anything, it detects (as correctly mentioned in other classes)

src/ctapipe/calib/camera/outlier.py
```diff
@@ -0,0 +1,112 @@
...
+
+    The interval `outliers_interval` corresponds to a range of valid statistical values.
+    """
+
+    def __call__(self, column):
+        # Remove outliers is statistical values out a given range
```
mexanick, line 65: Remove --> detect

mexanick (commented):
Looks nice, a couple of naming/doc suggestions, but otherwise is good to me.

mexanick (approved):

src/ctapipe/calib/camera/outlier.py
```diff
@@ -0,0 +1,112 @@
...
+from ctapipe.core import TelescopeComponent
+from ctapipe.core.traits import List
+
+
+class OutlierDetector(TelescopeComponent):
+    """
+    Base class for outlier detection algorithms.
+    """
+
+    outliers_interval = List(
```
kosack, line 25: In one case, this is the actually clipping interval (in units of the image value), but in others it is a multiplicative factor (unitless). Is it really needed in the base class then? Is there a reason not just just have explicit config variables for each implementation?

TjarkMiener: They follow this approach in `lstchain`, where they have config variables named `charge_median_cut_outliers`, `charge_std_cut_outliers`, `time_cut_outliers`. I aimed here to remove replications and the endless underscore naming conventions (which aren't really precise). The config variable would be the same for the sub-components `MedianBasedOutlierDetector` and `StdBasedOutlierDetector`, so should a parent of those components (maybe named `CameraBasedOutlierDetector`?) be introduced that holds the config variable of the multiplicative factor?

TjarkMiener:
This Pr needs to be rebased and "outlier files" have to moved to the `monitoring` module once #2554 is merged.

src/ctapipe/calib/camera/outlier.py
maxnoe: This is mainly about properly documenting the options I think. Which is much easier, if you have the option at each class with a fitting help=

kosack: Yes, It's ok to have it called outliers_interval in each class, but as the meaning is different for each sub-class, it's better to not put it in the base class, but explicitly in each subclass.  And then you can even be more explicit in the names if you prefer, or keep the same name. 

Especiallyy since the configuration for each must be different for each case (i.e. you cannot configure the base class values and then have them work for all sub-classes).  I.e. this would not work:

```yaml
OutlierDetector:
     time_cut_outliers: [3, 5] 
```
And then selecting `outlier_detector_method=MedianOutlierDetector` vs `outlier_detector_method=StdOutlierDetector`, would not work correctly. You'd want to explicitly configure each one.

TjarkMiener: I included the changes. @kosack, can you please have a look if it's now more clear and sounding? Thanks in advance!

src/ctapipe/monitoring/outlier.py
```diff
@@ -0,0 +1,131 @@
...
+    Detect outliers based on the deviation from the camera standard deviation.
+
+    The clipping interval to set the thresholds for detecting outliers is computed by multiplying
+    the configurable factors and the camera standard deviation of the statistic values.
+    """
+
+    std_range_factors = List(
+        [-1.0, 1.0],
+        help=(
+            "Multiplicative factors (unitless) applied to the camera standard deviation"
```
kosack, line 114: missing a space after `deviation "` and similar on the following lines.

Otherwise the docs end up like: 

>  Multiplicative factors (unitless) applied to the camera standard deviationof the provided statistic values to define a valid range based on thedeviation of the values to its camera median.

src/ctapipe/monitoring/outlier.py
```diff
@@ -0,0 +1,131 @@
...
+    the configurable factors and the camera median of the statistic values.
+    """
+
+    median_range_factors = List(
+        [-1.0, 1.0],
+        help=(
+            "Multiplicative factors (unitless) applied to the camera median"
+            "of the provided statistic values to define a valid range based on the"
+            "deviation of the values to its camera median."
+        ),
```
kosack, line 88: I missed this before, but you should set `minlen=2, maxlen=2` in all of the range factor config options.  Otherwise, incorrect list values like `[-1.0,]` validate correctly.   See the [traitlets docs](https://traitlets.readthedocs.io/en/stable/trait_types.html#containers).  Maybe also set `trait=traits.Float` to ensure the two values are floats.

kosack (changes requested):
Just some minor changes to make. .

src/ctapipe/monitoring/outlier.py
kosack: Also it may be clearer to write this as:

> Defines the range of acceptable values (lower, upper) in units of standard deviations. Values outside of this range will be flagged as outliers.

TjarkMiener: I think the help message needs to include that the config parameters one passed are actually factors which are multiplied by either the camera median or std. Regarding the second sentences, it's values **whose deviation to the camera median** lays outside this range are flagged as outliers.

Suggestions:
StdOutlierDetector
```python
        help=(
            "Multiplicative factors (unitless) applied to the camera standard deviation "
            "of the provided statistic values to define the range of acceptable values (lower, upper) "
            "in units of standard deviations. Values ​​whose deviation to the camera median lays outside "
            "of this range will be flagged as outliers."
        ),
```
MedianOutlierDetector
```python
        help=(
            "Multiplicative factors (unitless) applied to the camera median "
            "of the provided statistic values to define the range of acceptable values (lower, upper) "
            "in units of medians. Values ​​whose deviation to the camera median lays outside "
            "of this range will be flagged as outliers."
        ),
```

mexanick: I'd prefer Karl's formulation. It is simpler, more concise and easier to read imo.

mexanick (approved):

mexanick (approved):

src/ctapipe/monitoring/outlier.py
kosack: Yes, and for the mean, it would just be:

> Defines the range of acceptable values (lower, upper) as fractional deviations from the mean. Values outside of this range will be flagged as outliers.

kosack (approved):

mexanick (approved):

PR #2609, opened 2024-08-26
Add calibration calculators

TjarkMiener:
This PR contains the ``StatisticsCalculator``, which aggregates statistics, detects outliers, handles faulty data chunks. The ``StatisticsCalculator`` holds two functions to conduct two different passes over the data with and without overlapping chunks. The first pass is conducted with non-overlapping, while overlapping chunks can be set by the ``chunk_shift`` parameter in the second pass. The second pass over the data is only conducted in regions of trouble with a high percentage of faulty pixels exceeding the threshold ``faulty_pixels_threshold``.

related to issue #2542


src/ctapipe/monitoring/calculator.py
```diff
@@ -0,0 +1,337 @@
...
+
+
+        Parameters
+        ----------
+        table : astropy.table.Table
+            DL1-like table with images of shape (n_images, n_channels, n_pix), event IDs and
+            timestamps of shape (n_images, )
+        tel_id : int
+            Telescope ID for which the calibration is being performed
+        masked_pixels_of_sample : ndarray, optional
```
mexanick, line 176: Perhaps, you can consider using `MaskedColumn` columns in the table?

docs/api-reference/monitoring/index.rst
```diff
@@ -10,7 +10,7 @@ Monitoring data are time-series used to monitor the status or quality of hardwar
 
 This module provides some code to help to generate monitoring data from processed event data, particularly for the purposes of calibration and data quality assessment.
 
-Currently, only code related to :ref:`stats_aggregator` and :ref:`outlier_detector` is implemented here.
+Currently, code related to :ref:`stats_aggregator`, :ref:`calibration_calculator`, and :ref:`outlier_detector` is implemented here.
```
mexanick, line 13: Do we intend to add some more stuff here in the foreseeable future? If not, remove "currently" and rephrase accordingly.

src/ctapipe/monitoring/calculator.py
```diff
@@ -0,0 +1,337 @@
...
+    calculating camera-related monitoring data. It is designed
+    to be extended by specific calibration calculators that implement
+    the required methods for different types of calibration.
+
+    Attributes
+    ----------
+    stats_aggregator_type : ctapipe.core.traits.TelescopeParameter
+        The type of StatisticsAggregator to be used for aggregating statistics.
+    outlier_detector_list : list of dict
+        List of dictionaries containing the apply to, the name of the OutlierDetector subclass to be used, and the validity range of the detector.
```
mexanick, line 41: containing the apply to --> a bit unclear. I believe "apply_to" is a dict key. I think you can just provide an example `dict` here.

src/ctapipe/monitoring/calculator.py
```diff
@@ -0,0 +1,337 @@
...
+        default_value="SigmaClippingAggregator",
+        help="Name of the StatisticsAggregator subclass to be used.",
+    ).tag(config=True)
+
+    outlier_detector_list = List(
+        trait=Dict(),
+        default_value=None,
+        allow_none=True,
+        help=(
+            "List of dicts containing the name of the OutlierDetector subclass to be used, "
```
mexanick, line 57: Same here, provide an example dict.

src/ctapipe/monitoring/calculator.py
```diff
@@ -0,0 +1,337 @@
...
+
+
+class StatisticsCalculator(CalibrationCalculator):
+    """
+    Component to calculate statistics from calibration events.
+
+    This class inherits from ``CalibrationCalculator`` and is responsible for
+    calculating various statistics from calibration events, such as pedestal
+    and flat-field data. It aggregates statistics, detects outliers,
+    handles faulty data chunks.
```
mexanick, line 125: "Chunk" definition is missing, it should be added somewhere.

src/ctapipe/monitoring/calculator.py
```diff
@@ -0,0 +1,337 @@
...
+
+
+class CalibrationCalculator(TelescopeComponent):
+    """
+    Base component for calibration calculators.
+
+    This class provides the foundational methods and attributes for
+    calculating camera-related monitoring data. It is designed
+    to be extended by specific calibration calculators that implement
+    the required methods for different types of calibration.
```
mexanick, line 34: You don't declare any abstract methods, I'd drop "that implement..." part.

src/ctapipe/monitoring/calculator.py
```diff
@@ -0,0 +1,337 @@
...
+                "chunk_shift must be set if second pass over the data is requested"
+            )
+        # Get the aggregator
+        aggregator = self.stats_aggregator[self.stats_aggregator_type.tel[tel_id]]
+        # Conduct a second pass over the data
+        aggregated_stats_secondpass = None
+        faulty_chunks_indices = np.where(~valid_chunks)[0]
+        for index in faulty_chunks_indices:
+            # Log information of the faulty chunks
+            self.log.warning(
```
mexanick, line 255: Shall this be a `warning`? I'd say, rather `info`.

src/ctapipe/monitoring/calculator.py
```diff
@@ -0,0 +1,337 @@
...
+                f"Faulty chunk detected in the first pass at index '{index}'."
```
mexanick, line 256: Don't use f-strings in the log messages. Weird this passed the CI, [G004 should be raised ](https://docs.astral.sh/ruff/rules/logging-f-string/)

@maxnoe FYI

mexanick (commented):
A few comments inline. On top of this, looking at the implemented methods in classes, I'd consider the CalibrationCalculator base class as redundant. Do we foresee other children for this base class? StatisticsCalculator is rather universal (and can be used for calibration or monitoring, or data quality purposes), I'd consider merging them into one class.

TjarkMiener:
> I'd consider the CalibrationCalculator base class as redundant. Do we foresee other children for this base class? StatisticsCalculator is rather universal (and can be used for calibration or monitoring, or data quality purposes), I'd consider merging them into one class.

@ctoennis asked for this for his pointing calibration. Can you please have a look at the PR and see if you can use ``StatisticsCalculator`` for your pointing purposes? Or maybe we need just minor adjustment? Otherwise we should only have one class as suggested by @mexanick

src/ctapipe/monitoring/calculator.py
TjarkMiener: At the moment, I do not have the the info about "bad pixels" in the LST-1 files. Therefore, I used as an optional ndarray and passed `None`. If we change the type here to `MaskedColumn` and one point down in the aggregation we would need to access the data via `masked_pixels_of_sample.data` which would break for `NoneType`.

mexanick (dismissed):

TjarkMiener:
> @ctoennis asked for this for his pointing calibration. Can you please have a look at the PR and see if you can use `StatisticsCalculator` for your pointing purposes? Or maybe we need just minor adjustment? Otherwise we should only have one class as suggested by @mexanick

After some internal discussion, we decided to decouple the pointing from the `StatisticsCalculator`. I already removed the previous base class `CalibrationCalculator` and implemented the changes from @mexanick. This PR would be now ready for review @kosack @maxnoe. Thanks in advance!

TjarkMiener:
While coding the [Tool](https://gitlab.cta-observatory.org/cta-computing/dpps/calibrationpipeline/calibpipe/-/merge_requests/121) utilizing the `StatisticsCalculator`, I noticed the bug, which is now fix. The code was not appending the aggregated stats during the `second_pass`.

CODEOWNERS
```diff
@@ -5,6 +5,8 @@
 ctapipe/calib/camera  @watsonjj
 ctapipe/image/extractor.py  @watsonjj @HealthyPear
 
+ctapipe/monitoring  @TjarkMiener
```
maxnoe, line 8: we don't really make use of this mechanism and the people listed here are not currenly active anymore, so I'd rather just remove it (in another PR) than adding to it here.

TjarkMiener: Okay, I removed my entry.

TjarkMiener:
Dear @maxnoe @kosack. May I kindly ask for a code review on this PR. Many thanks in advance!

mexanick (approved):

src/ctapipe/monitoring/calculator.py
```diff
@@ -0,0 +1,327 @@
...
+)
+from ctapipe.monitoring.aggregator import StatisticsAggregator
+from ctapipe.monitoring.outlier import OutlierDetector
+
+__all__ = [
+    "StatisticsCalculator",
+]
+
+
+class StatisticsCalculator(TelescopeComponent):
```
kosack, line 26: One minor comment/question: since "monitoring" is more general (it's any time-series), this name is a bit generic, and it's not clear that the inputs/outputs of this are _pixel-wise_ monitoring information.  Maybe `PixelStatisticsCalculator` would be more clear?    Or is the intent here to eventually have this work for any monitoring data, not just pixel-wise?

TjarkMiener: I agree with your suggestion. The `TelescopeComponent` is very tailored for _pixel-wise_ monitoring information at the moment. I'm not sure if all the functionality implemented for _pixel-wise_ monitoring data is needed for other regular monitoring data. There is also the [astropy aggregation](https://docs.astropy.org/en/stable/table/operations.html#aggregation) operations, which can be also utilized for rather slow-changing monitoring data.

mexanick (approved):

src/ctapipe/monitoring/calculator.py
```diff
@@ -0,0 +1,327 @@
...
+            this is mutually exclusive with passing ``config``
+        stats_aggregator: ctapipe.monitoring.aggregator.StatisticsAggregator
+            The ``StatisticsAggregator`` to use. If None, the default via the
+            configuration system will be constructed.
+        """
+        super().__init__(subarray=subarray, config=config, parent=parent, **kwargs)
+        self.subarray = subarray
+
+        # Initialize the instances of StatisticsAggregator
+        self.stats_aggregator = {}
```
maxnoe, line 109: should be plural

src/ctapipe/monitoring/calculator.py
```diff
@@ -0,0 +1,327 @@
...
+        if stats_aggregator is None:
+            for _, _, name in self.stats_aggregator_type:
+                self.stats_aggregator[name] = StatisticsAggregator.from_name(
+                    name, subarray=self.subarray, parent=self
+                )
+        else:
```
maxnoe, line 115: I guess you copied this style from `CameraCalibrator` or similar. But I am not sure we want to support this. I think it circumvents somewhat the configuration system and thus the provenance.

src/ctapipe/monitoring/calculator.py
```diff
@@ -0,0 +1,327 @@
...
+            name = stats_aggregator.__class__.__name__
+            self.stats_aggregator_type = [("type", "*", name)]
+            self.stats_aggregator[name] = stats_aggregator
+
+        # Initialize the instances of OutlierDetector
+        self.outlier_detectors = {}
+        if self.outlier_detector_list is not None:
+            for outlier_detector in self.outlier_detector_list:
```
maxnoe, line 123: I think here you should better validate that the configuration is correct (the dict has the right keys) and raise a `TraitConfigurationError` if not, not just the `KeyError` that will result from wrong configuration here.

Also, validate that the `apply_to` is a valid key.

src/ctapipe/monitoring/calculator.py
```diff
@@ -0,0 +1,327 @@
...
+
+        Parameters
+        ----------
+        table : astropy.table.Table
+            DL1-like table with images of shape (n_images, n_channels, n_pix), event IDs and
+            timestamps of shape (n_images, )
+        tel_id : int
+            Telescope ID for which the calibration is being performed
+        masked_pixels_of_sample : ndarray, optional
+            Boolean array of masked pixels of shape (n_pix, ) that are not available for processing
```
maxnoe, line 157: `n_pixels`? Or `(n_channels, n_pixels)`? In principle, high gain and low gain can be independently fail.

maxnoe: Also, please use `n_pixels`

src/ctapipe/monitoring/calculator.py
```diff
@@ -0,0 +1,327 @@
...
+        # Pass through the whole provided dl1 table
+        aggregated_stats = aggregator(
+            table=table,
+            masked_pixels_of_sample=masked_pixels_of_sample,
+            col_name=col_name,
+            chunk_shift=None,
+        )
+        # Detect faulty pixels with multiple instances of ``OutlierDetector``
+        outlier_mask = np.zeros_like(aggregated_stats["mean"], dtype=bool)
+        for aggregated_val, outlier_detector in self.outlier_detectors.items():
```
maxnoe, line 177: Maybe we should store a bitmask / boolean array of which of the detectors rejected the sample as outlier?

Like this, you cannot really understand why an event was rejected.

So maybe, compute all masks for all detectors in an nd array, then apply `or` in the end.

src/ctapipe/monitoring/calculator.py
```diff
@@ -0,0 +1,327 @@
...
+            Table containing the aggregated statistics after the second pass, their outlier masks, and the validity of the chunks.
+        """
+        # Check if the chunk_shift is set for the second pass
+        if self.chunk_shift is None:
+            raise ValueError(
+                "chunk_shift must be set if second pass over the data is requested"
+            )
+        # Check if at least one chunk is faulty
+        if np.all(valid_chunks):
+            raise ValueError(
```
maxnoe, line 230: Is this really an error? Or should it just return as a no-op?

src/ctapipe/monitoring/calculator.py
```diff
@@ -0,0 +1,327 @@
...
+                "All chunks are valid. The second pass over the data is redundant."
+            )
+        # Get the aggregator
+        aggregator = self.stats_aggregator[self.stats_aggregator_type.tel[tel_id]]
+        # Conduct a second pass over the data
+        aggregated_stats_secondpass = []
+        faulty_chunks_indices = np.where(~valid_chunks)[0]
```
maxnoe, line 237: Use `nonzero` instead of single-argument `where`, see: https://numpy.org/doc/stable/reference/generated/numpy.where.html

src/ctapipe/monitoring/calculator.py
```diff
@@ -0,0 +1,327 @@
...
+                        table=table_sliced,
+                        masked_pixels_of_sample=masked_pixels_of_sample,
+                        col_name=col_name,
+                        chunk_shift=self.chunk_shift,
+                    )
+                )
+        # Stack the aggregated statistics of each faulty chunk
+        aggregated_stats_secondpass = vstack(aggregated_stats_secondpass)
+        # Detect faulty pixels with multiple instances of OutlierDetector of the second pass
+        outlier_mask_secondpass = np.zeros_like(
```
maxnoe, line 276: this is the same code as in the `first_pass`. Should be factored out to a `def _find_outliers(...)`

src/ctapipe/monitoring/calculator.py
```diff
@@ -0,0 +1,327 @@
...
+
+        Returns
+        -------
+        numpy.ndarray
+            Boolean array where each element indicates whether the corresponding
+            chunk is valid (True) or faulty (False).
+        """
+        # Check if the camera has two gain channels
+        if outlier_mask.shape[1] == 2:
+            # Combine the outlier mask of both gain channels
```
maxnoe, line 316: Why? The two gains can be independent

src/ctapipe/monitoring/calculator.py
```diff
@@ -0,0 +1,327 @@
...
+            outlier_mask = np.logical_or(
```
maxnoe, line 317: axis=1

src/ctapipe/monitoring/calculator.py
```diff
@@ -0,0 +1,327 @@
...
+                outlier_mask[:, 0, :],
+                outlier_mask[:, 1, :],
+            )
+        # Calculate the fraction of faulty pixels over the camera
+        faulty_pixels_percentage = (
+            np.count_nonzero(outlier_mask, axis=-1) / np.shape(outlier_mask)[-1]
+        ) * 100.0
```
maxnoe, line 324: Do we really want to work in percentages here? I think just using fractions would be less confusing

src/ctapipe/monitoring/outlier.py
```diff
@@ -81,7 +81,7 @@ class MedianOutlierDetector(OutlierDetector):
     the configurable factors and the camera median of the statistic values.
     """
 
-    median_range_factors = List(
+    validity_range = List(
```
maxnoe, line 84: Why this change?

TjarkMiener:
> After this, we are still missing a `Component` and/or `Tool` to perform the actual computations on input data containing the calibration events, correct?

Correct, we have prototype tools ready in `calibpipe-MRs` for the [stats calculation](https://gitlab.cta-observatory.org/cta-computing/dpps/calibrationpipeline/calibpipe/-/merge_requests/121) and the [ffactor method](https://gitlab.cta-observatory.org/cta-computing/dpps/calibrationpipeline/calibpipe/-/merge_requests/124).

> And that will finally replace the two classes in `ctapipe.calib.camera.{flatfield,pedestal}`?

Yes, we would have to make sure that the interface to `camcalib coefficients` is in place before removing them though.

src/ctapipe/monitoring/calculator.py
```diff
@@ -0,0 +1,327 @@
...
+    stats_aggregator_type = TelescopeParameter(
+        trait=ComponentName(
+            StatisticsAggregator, default_value="SigmaClippingAggregator"
+        ),
+        default_value="SigmaClippingAggregator",
+        help="Name of the StatisticsAggregator subclass to be used.",
+    ).tag(config=True)
+
+    outlier_detector_list = List(
+        trait=Dict(),
```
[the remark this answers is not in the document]
TjarkMiener: This component is used to aggregate the stats for the _pedestal_, _flatfield_, _timecalibration_, and _pedestalvariance_.  Each of this aggregation has sensible distinct default configuration. Should we prioritize one of the aggregation over the others and set the default config for that aggregation?

src/ctapipe/monitoring/calculator.py
TjarkMiener: Currently, we only have a `ToolConfigurationError` trait in `ctapipe`. Should be a `TraitConfigurationError` trait be added in the `ctapipe core.traits`?

src/ctapipe/monitoring/calculator.py
TjarkMiener: This boolean mask corresponds to hardware failure, so i think it should be (n_pixels,). We treat high and low gain independently later on software level when aggregating the stats for each type of calibration.

src/ctapipe/monitoring/outlier.py
TjarkMiener: We would need the same generic name in each `OutlierDetector` class in order to init them homogeneously in the `PixelStatisticsCalculator`.

src/ctapipe/monitoring/calculator.py
TjarkMiener: This function only detects whether the chunk is faulty or valid. Therefore, we need to calculate the fraction of faulty pixels in the camera. Here I define a pixel as faulty if one of the gains are faulty. Mind that definition is only used to decide if a chunk is faulty in order to calculate the second pass around that chunk. The outlier mask is stored for two gains, so one would have the information when _camcalib coefficients_ are applied.

src/ctapipe/monitoring/calculator.py
TjarkMiener: `outlier_mask = np.logical_or.reduce(outlier_mask, axis=1)`

src/ctapipe/monitoring/calculator.py
TjarkMiener: We do not reject here on event-level but only on pixel-level. Currently we have for each calibration type (_pedestal_, _flatfield_, _timecalibration_, _pedestalvariance_) an outlier mask with shape of `(n_channels, n_pixels)`. Storing per detectors for each calibration type seems a little bit heavy and would make the downstream code a little bit more complex. I would only add this if that is really needed for the monitoring of the data quality.

src/ctapipe/monitoring/calculator.py
TjarkMiener: `faulty_chunks_indices = np.flatnonzero(~valid_chunks)`

src/ctapipe/monitoring/calculator.py
maxnoe: But the hardware of the two channels is at least partly independent, at least in LST (same PMT but e.g. different DRS4 chips). So they can also fail independently.

src/ctapipe/monitoring/calculator.py
maxnoe: I am not saying that you should *use* that mask for the analysis. Just for understanding *why* a pixel was deemed unusable. Which is very important information e.g. to distinguish a hardware error from just reduced HV due to a star.

src/ctapipe/monitoring/calculator.py
TjarkMiener: Actually the shape is `(n_channels, n_pixels)` as you say. It's propagated to the [aggregator](https://github.com/cta-observatory/ctapipe/blob/main/src/ctapipe/monitoring/aggregator.py#L167C50-L167C73) (`np.ma.array(images, mask=masked_pixels_of_sample)`), where `masked_pixels_of_sample` should have the same shape as `images` (which is of shape `(n_channels, n_pixels)`). I will update all the docstrings accordingly.

With my current setup using [`lstchain-data_r0_to_dl1` script](https://github.com/cta-observatory/cta-lstchain/blob/main/lstchain/scripts/lstchain_data_r0_to_dl1.py) (for writing R1 files for calib events) and `ctapipe-process` (for integrating and extracting DL1a images for calib events), the information about broken pixels due to hardware failure is lost. Once we write in ACADA DL0 files (cosmic and calib events) we should have access to this information.

src/ctapipe/monitoring/calculator.py
TjarkMiener: Okay, makes sense! I just pushed a commit. Does it look good to you?

src/ctapipe/monitoring/calculator.py
TjarkMiener: Done in the new commit!

src/ctapipe/monitoring/calculator.py
```diff
@@ -0,0 +1,336 @@
...
+            "Number of samples to shift the aggregation chunk for the calculation "
+            "of the statistical values. Only used in the second_pass(), since the "
+            "first_pass() is conducted with non-overlapping chunks (chunk_shift=None)."
+        ),
+    ).tag(config=True)
+
+    faulty_pixels_threshold = Float(
+        default_value=0.1,
+        allow_none=True,
+        help="Threshold of faulty pixels over the camera to identify regions of trouble.",
```
kosack, line 74: I don't quite understand from the help what the units are here. Is this a fraction of the camera? If so, maybe rename it to `faulty_pixels_fraction` and say "Minimum fraction of faulty camera pixels...."

TjarkMiener: Sounds good! Resolved in the latest commit.

src/ctapipe/monitoring/calculator.py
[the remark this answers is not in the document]
maxnoe: How do you configure the three separate cases? Our configuration system has the limitation of being able only to configure by class name, not several instances of the same class.

Are these in separate tools? Or are multiple instances of this class created in the same tool?
There are a couple of possibilities depending on that. 

If you only need one instance per application, then leaving the default empty like here is fine.

If you need multiple, differently configured instances of this component, the only way unfortunately is to use inheritance to create several different classes, e.g.:

```python
class FlatFieldStatisticsCalculator(PixelStatisticsCalculator):
    
    @default("outlier_detector_list")
    def default_outlier_config(self):
        return ...
```

Since the plan is to have both types of calibration events in the same stream, I guess you need at least two instances in parallel in the tool.

I think in this case, we can also leave it like this here, and have the two subclasses only inside the tool with the appropriate default configuration.

src/ctapipe/monitoring/calculator.py
```diff
@@ -0,0 +1,336 @@
...
+        config: traitlets.loader.Config
+            Configuration specified by config file or cmdline arguments.
+            Used to set traitlet values.
+            This is mutually exclusive with passing a ``parent``.
+        parent: ctapipe.core.Component or ctapipe.core.Tool
+            Parent of this component in the configuration hierarchy,
+            this is mutually exclusive with passing ``config``
+        """
+        super().__init__(subarray=subarray, config=config, parent=parent, **kwargs)
+        self.subarray = subarray
```
maxnoe, line 100: not needed, done by `super()`

src/ctapipe/monitoring/calculator.py
```diff
@@ -0,0 +1,336 @@
...
+            for d, outlier_detector in enumerate(self.outlier_detector_list):
+                # Check if all required keys are present
+                missing_keys = {
+                    "apply_to",
+                    "name",
+                    "validity_range",
+                } - outlier_detector.keys()
+                if missing_keys:
+                    raise TraitError(
+                        f"Entry '{d}' in the ``outlier_detector_list`` trait is missing required key(s): {', '.join(missing_keys)}"
```
maxnoe, line 121: This line looks too long... how did it escape the pre-commit checks?

src/ctapipe/monitoring/outlier.py
maxnoe: And what if someone comes up with an `OutlierDetector` that works differently?

maxnoe: I also do not understand this argument. You have a full dict config for each of the outlier detectors.

What prevents you from configuring:

```
{
  "outlier_detector_list": [
    {
      "apply_to": "mean",
      "name": "MedianOutlierDetector",
      "median_range_factor": [-2.0, 2.0],
    },
    {
      "apply_to": "mean",
      "name": "StdOutlierDetector",
      "std_range_factor": [-2.0, 2.0],
    },
  ],
}
```

or maybe even better:

```
{
  "outlier_detector_list": [
    {
      "apply_to": "mean",
      "name": "MedianOutlierDetector",
      "config": {
        "median_range_factor": [-2.0, 2.0],
      }
    },
    {
      "apply_to": "mean",
      "name": "StdOutlierDetector",
      "config": {
        "std_range_factor": [-2.0, 2.0],
      },
    },
  ],
}
```

maxnoe: Using the first example, you could do (from memory, not tested...):

```
for outlier_detector in enumerate(self.outlier_detector_list):
    # copy, so we don't mutate the actual configuration
    outlier_detector = outlier_detector.copy()
    cls_name = outlier_detector.pop("name")
    apply_to = outlier_detector.pop("apply_to")
    self.outlier_detectors.append(OutlierDetector.from_name(cls_name, parent=self, **outlier_detector)
```

src/ctapipe/monitoring/calculator.py
```diff
@@ -0,0 +1,336 @@
...
+
+    stats_aggregator_type = TelescopeParameter(
+        trait=ComponentName(
+            StatisticsAggregator, default_value="SigmaClippingAggregator"
+        ),
+        default_value="SigmaClippingAggregator",
+        help="Name of the StatisticsAggregator subclass to be used.",
+    ).tag(config=True)
+
+    outlier_detector_list = List(
```
maxnoe, line 49: I think for the configuration / provenance system to properly track the configuration of each of the detectors, there should be a `List(OutlierDetector).tag(config=True)` on the class. 

It can be empty and populated from this list of configuration dicts as you have it here. Just add the additional traitlet instead of using a plain list in `__init__`.

maxnoe:
> Correct, we have prototype tools ready in calibpipe-MRs for the [stats calculation](https://gitlab.cta-observatory.org/cta-computing/dpps/calibrationpipeline/calibpipe/-/merge_requests/121) and the [ffactor method](https://gitlab.cta-observatory.org/cta-computing/dpps/calibrationpipeline/calibpipe/-/merge_requests/124).

I think (after this is merged here) we should talk about the place of these tools. 

I don't really see why we should add the basic components needed (only) by these tools to ctapipe itself, but then the tools only to calibpipe.

src/ctapipe/monitoring/calculator.py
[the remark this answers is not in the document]
TjarkMiener: > If you only need one instance per application, then leaving the default empty like here is fine.

Yes, this is what we have foreseen in the workflow. Each calibration type should be processed in a separate stream.

src/ctapipe/monitoring/outlier.py
mexanick: For me having the name to appear in the class name and in the config variable is a self-repetition. 
We have an outlier detector, that uses a validity range, this is a generic concept, which is specialized in the derived outlier detectors. Now, if someone develops an outlier detector that doesn't follow this logic, it should not be derived from the same parent (and btw, what would be a use case for this? I can't figure that out).

src/ctapipe/monitoring/calculator.py
mexanick: this is an interesting question... And we've seen some weird skips of pre-commit hook in Christoph's PR too...

TjarkMiener: I don't know. To double-check I run 
```bash
$ pre-commit run --files $(git diff --name-only origin/main)
trim trailing whitespace.................................................Passed
check for added large files..............................................Passed
check for case conflicts.................................................Passed
check for merge conflicts................................................Passed
fix end of files.........................................................Passed
codespell................................................................Passed
ruff.....................................................................Passed
ruff-format..............................................................Passed
```
which `Passed`.

src/ctapipe/monitoring/calculator.py
TjarkMiener: This traitlet is currently a list of dicts since it holds besides the config for the `OutlierDetector` component also the column which the detector should be `apply_to`, i.e. on which column the `call` function of the component is applied to. Is this answer resolving the comment or did I misinterpret the comment?

src/ctapipe/monitoring/outlier.py
maxnoe: Before, it made clear that it is a relative factor, not an absolute range.

> what would be a use case for this? 

There are many difference outlier detection algorithms. Not all of them have a two parameter valid range. 
One example would be sigma clipping.

In general, we don't require that subclasses of components have the same configuration options.

Quite the opposite actually. The should have the same API, so they can be used interchangebly in the code, but the configuration is specific to each one and the configuration system takes care of that.

This is the main use case of the config / component systems: have classes that have the same API in the code, but that can be configured using different options. See e.g. the cleaning or image extraction components.

src/ctapipe/monitoring/calculator.py
maxnoe: Yes, you didn't understand the point. I did not ask to change the current option.

I asked to add a second traitlet for the list of the **instances** of  `OutlierDetector`

maxnoe: I see this is currently the same as in CameraCalibrator. I think we can leave it for now like this and after this is merged, I will have a look if this needs fixing in both.

src/ctapipe/monitoring/outlier.py
mexanick: Ah, I see your point better now. In my opinion, the configuration parameter meaning shall be properly documented in the specific implementation of the OutlierDetector. Also, we do not require two parameter valid range. Perhaps a better name can work, to have it generic (with a proper description in the specific implementation). But I don't have a strong opinion on that.

maxnoe: Above I posted a code snipped that would allow passing any configuration option to the specific outlier detector. 

Independently of what is the best name for the options of the current detectors, I think this is the way to go.

I would prefer if this PR wasn't making API breaking changes in code that is not directly affected by the main goal of this PR.

src/ctapipe/monitoring/calculator.py
```diff
@@ -106,27 +105,30 @@ def __init__(
                 name, subarray=self.subarray, parent=self
             )
 
-        # Initialize the instances of OutlierDetector
-        self.outlier_detectors = {}
+        # Initialize the instances of OutlierDetector from the configuration
+        self.outlier_detectors, self.apply_to_list = [], []
         if self.outlier_detector_list is not None:
             for d, outlier_detector in enumerate(self.outlier_detector_list):
+                outlier_detector = outlier_detector.copy()
```
maxnoe, line 112: Since you now introduced the `config` key and are not actually passing the `**outlier_detector` dict itself to the constructor, the copy and pop are not necessary.

Simply iterating over the dicts and accessing the three keys is fine

src/ctapipe/monitoring/calculator.py
maxnoe: I created an obvious style limitation in the file and it got picked up by the linter... so I am lost why this long line is "ok"...

src/ctapipe/monitoring/calculator.py
maxnoe: What I missed until now: there are only `first_pass` and `second_pass` methods here, but now common one.

What is the intendend use? Can we add a `__call__` along the lines of:

```
def __call__(self, ...):
    result_first = self.first_pass(...)
    second_pass_necessary = ...
    if second_necessary_necessary:
        result_second = self.second_pass(...)
    # merge results
    result = ...
    return result
```
?

TjarkMiener: This is currently done in the tool, i.e. see [here](https://gitlab.cta-observatory.org/cta-computing/dpps/calibrationpipeline/calibpipe/-/blob/fcd10f4ff3f04f5e01af4d77312760d4250a448c/src/calibpipe/tools/stats_calculation.py#L124-151). Should this be moved into the `PixelStatisticsCalculator` component?

maxnoe: Yes, I think this is purely concerning this extractor here.

maxnoe: And without it, the intended use is not obvious, the code is incomplete

mexanick: While I agree that as is the intended use is not obvious, I'm not sure we want to implement the logic of first/second pass calling here. There might be situations when you don't want to call the second pass (e.g. first pass detected too many problematic chunks and you conclude your data is basically trash). I suggest to add to the class docstring the example of use like @maxnoe pointed above, but leave the decisive logic to the tool that calls the component.

src/ctapipe/monitoring/calculator.py
maxnoe: Wouldn't it make sense to process both gains independently?

TjarkMiener: That would complicate things but in principle that would be possible. However, the goal of the functionality is to detect time periods of troubles, mainly car flashes. Therefore, both gains should be highly correlated anyway.

kosack (commented):
I think it looks good now. The only question is if you should add a template/example config file for it that can be generated by `ctapipe-quickstart`, as we do for other tools.

For that you would just need to create a YAML config file in `ctapipe/reasources/` and modify `CONFIGS_TO_WRITE` in `quickstart.py`.

maxnoe:
@kosack there is no tool here

kosack:
>  @kosack there is no tool here

You're right... I was looking at the config in the test case, but that is only a Component config, and the tool is in calibpipe.  

Then everything is ok, and we can open an issue to add a generic tool separately.

src/ctapipe/monitoring/calculator.py
maxnoe: We discussed the boundary of ctapipe / calibpipe in a meeting and decided that the statistics calculators are useful to have in ctapipe.

A generic tool to apply these to data would also be good as a follow up PR. The rest stays / goes into calibpipe

maxnoe (approved):

kosack (approved):

maxnoe:
@TjarkMiener If you have no issue with it, I will squash this PR. 116 commits for these changes seems a bit much.

TjarkMiener:
Sure, I think the huge numbers of commits is because we branched off the stats_extractor branch back in the days, which are included.

Issue #2610, opened 2024-09-04
Check uniqueness of obs_id / event_id when merging files.


maxnoe:
**Please describe the use case that requires this feature.**

Processing on the GRID merged the same file multiple times resulting in later errors when joining tables.

**Describe the solution you'd like**

Check uniqueness of obs_id event combinations when merging files to prevent such issues at the point where the error happens.

Shadi201711:
try this
https://gofile.io/d/yMHabA
select the desired compiler in the installer

PR #2611, opened 2024-09-04
Check uniqueness of input paths and obs_ids in merge tool

maxnoe:
This is a very simple check to partially address #2610 

I will add the check for obs_id in another PR to fully address #2610

### DIFF
--- a/src/ctapipe/io/hdf5merger.py
+++ b/src/ctapipe/io/hdf5merger.py
@@ -188,6 +188,8 @@ class HDF5Merger(Component):
         self.data_model_version = None
         self.subarray = None
         self.meta = None
+        self._merged_obs_ids = set()
+
         # output file existed, so read subarray and data model version to make sure
         # any file given matches what we already have
         if appending:
@@ -201,6 +203,9 @@ class HDF5Merger(Component):
                 focal_length_choice=FocalLengthKind.EQUIVALENT,
             )
             self.required_nodes = _get_required_nodes(self.h5file)
+
+            # this will update _merged_obs_ids from existing input file
+            self._check_obs_ids(self.h5file)
 
     def __call__(self, other: str | Path | tables.File):
         """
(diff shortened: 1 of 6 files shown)


src/ctapipe/io/hdf5merger.py
```diff
@@ -267,7 +283,31 @@ def _check_can_merge(self, other):
...
+        ob_key = "/configuration/observation/observation_block"
+        # fallback for old files without ob config
+        trigger_key = "/dl1/event/subarray/trigger"
+
+        if ob_key in other.root:
+            key = ob_key
+        elif trigger_key in other.root:
+            key = trigger_key
+        else:
+            self.log.warning(
```
Tobychev, line 296: Is a missing obs_id really such an innocent error that throwing a warning should be the default action?

 I feel like a `only_warn` cli option and a default to raising an `AttributeError` or something similar would be safer.

maxnoe: I wasn't totally sure on what kinds of input files the merge tool works and if we really do want to require all files on which it works to have either the observation block configuration or the dl1 subarray trigger table.

However, thinking about it again, either of those should probably always be there.

maxnoe:
The only real unit tests testing the merging thoroughly is merging the only two large files we have: the gamma and the proton training dl2 files.

However, due to the simulations re-using run ids for the different particles, the tool now correctly complains about duplicated obs_ids.

I am not sure how to address this, I think we should keep the test but adding an option to disable the uniqueness check just to test the that the merging works on this invalid input also seems strange.

I could modify the file on disk to have different obs_ids...

maxnoe:
> What happens if we do want to make a file with multiple particles types mixed in it, and they have overlapping obs_ids?

This is something that is not supported now in any case: the data model assumes that we have unique obs_id / event_id combinations and on this assumptions many things are build (TableLoader, StereoPrediction, etc.).

This is why we added the `override_obs_id` option to `SimTelEventSource`, so that we can change the obs_ids at the start to something unique.

kosack:
> This is why we added the override_obs_id option to SimTelEventSource, so that we can change the obs_ids at the start to something unique.

Yes, that's clear. So nothing to do in ctapipe, but the question is if we should update the way obs_ids are re-assigned for simulations so that each particle species is unique. That would  have to just be a convention, not a software fix.

orelgueta:
> @GernotMaier or @orelgueta might confirm, but I think we talked about it and prod6 already has unique obs_ids, at least per site / pointing position

Unique obs_ids are only per site/pointing/**particle**. So we reuse the run numbers also for different primaries in the same pointing/site. 

If I remember correctly our discussions, we realised that anyway you will have to assign a unique obs_id after the fact, so there is no reason for us to change the behaviour (which is done manually when launching the jobs so quite prone to error).

maxnoe:
Ok, I don't really like adding more meaning to obs_ids... but looks like this might be the best option for now.

Including particle id automatically is easy, site probably as well.  Pointing position maybe not so much.

At least for now, all our tools that need multiple species as input are designed to read from multiple files anyway.

Tobychev (approved):

kosack (approved):

PR #2615, opened 2024-09-10
Move interpolator from io to monitoring

maxnoe:
While preparing the changelog for 0.22, I realized that docs were missing for the interpolator and the new `monitoring` module is also the more appropriate place.


TjarkMiener (approved):
Thanks! Looks good to me!

kosack (approved):

PR #2616, opened 2024-09-11
Convert metadata Instrument.site to free string instead of enum

maxnoe:
I was creating the configuration for prod6 and changed the site name from `CTA-North` to `CTAO-North` and it resulted in an error.

I think having a fixed list of sites here is too restrictive and just using a free string is fine.


kosack (approved):
yes, fine with me

Tobychev (approved):

PR #2619, opened 2024-09-13
Add CI run running on oldest supported versions of core dependencies

maxnoe:


kosack (dismissed):

Tobychev (approved):

.github/workflows/ci.yml
```diff
@@ -143,7 +144,11 @@ jobs:
           pip install -e ".[${EXTRAS}]"
           pip install ./test_plugin
           pip freeze
-          pip list
+
+      - if: contains(matrix.extra-args, 'oldest-deps') && contains(github.event.pull_request.labels.*.name, 'documentation-only') == false
+        run: |
+          micromamba install -n cta-dev -y numpy=1.24 numba=0.57 scipy=1.10 astropy=5.3 matplotlib=3.6
```
Tobychev, line 150: Is there really no other way than hardcoding the version numbers in the `ci.yml` config file? Seems easy to have it get out of synch, but may that doesn't lead to any hard to solve problems?

maxnoe: I didn't find an elegant solution to this. astropy for example also hard-codes their oldest supported versions (in tox.ini in this case):
https://github.com/astropy/astropy/blob/fc2ef61cf6dbb3a06a5e9d11c19995960902e63b/tox.ini#L76-L90

Tobychev (dismissed):

kosack (approved):

mdebony (approved):

PR #2623, opened 2024-09-30
add PointingCalculator

ctoennis:
I wanted to post this pull request here for the PointingCalculator.

I have not finished adding unit tests for the calculator and the code does not run as is.

One important thing to discuss is the method of looking up star locations. Currently the code uses astroquery, which would be a new dependency.

Otherwise this is still work in progress.


src/ctapipe/calib/camera/calibrator.py
```diff
@@ -2,6 +2,10 @@
 Definition of the `CameraCalibrator` class, providing all steps needed to apply
 calibration and image extraction, as well as supporting algorithms.
 """
+<<<<<<< HEAD
+
+=======
+>>>>>>> c5f385f0 (I added a basic star fitter)
```
maxnoe, line 8: merge conflict markers

ctoennis:
i removed the merge conflict markers and moved the flatfield and pedestal interpolators over to where the other interpolators are.

kosack:
I think relying on accessing the internet to get star positions is not a good idea if these will run in large batch queues, as it is a point of failure and requires outside access.  Instead, maybe we should think about just making one of configurable inputs be an astropy table (i.e. read the output of astroquery, but not requiring astroquery to be called internally)

ctoennis:
@kosack that sounds good. Since the cameras will often track specific objects we can pre-create such tables for those objects. The StarTracer thant needs to have the star positions can be just given a table of positions instead of using astroquery. 

I will start out with some table for the crab nebula to have something for the unit tests.

maxnoe:
@ctoennis the yale bright star catalog is already included in ctapipe as a static resource:
https://ctapipe.readthedocs.io/en/latest/api/ctapipe.utils.astro.get_bright_stars.html

It is complete to magnitude 6.5, which I think is enough for the pointing estimation, right?

If not, we can also include e.g. the hipparcos catalog down to a lower brightness.

To get a good idea of how you could use astroquery, have a look at this script I recently wrote to find stars suitable for PSF measurement:

https://github.com/maxnoe/find_psf_stars/blob/main/find_psf_stars.py

It queries the HIP catalog via astroquery to Vizier and downloads two files from the HIP webpage providing additionally common names for the stars where they are available.

mexanick:
In general `vizier` creates a cache, so we can make a setup with cached catalogs. However, we probably can get away with Yale bright star catalog too. @maxnoe you selection uses V-mag? We use B-mag for selection, so here we may need a bit of optimization on the range, but in general it should still work quite ok. @ctoennis please try this option

maxnoe:
> your selection uses V-mag? We use B-mag for selection

Yes, the current limit in the utility function is put on `VMag`, but you can simply change `max_magnitude` to have to arguments, `max_vmag` and `max_bmag`.

mexanick:
I think we can try with the V-mag and see... 6.5 magnitude is close to the limit, we need to re-evaluate the star reconstruction performance to see whether we need fainter stars.

maxnoe:
Ah, nevermind, the catalog export I did doesn't include `Vmag`, but you can also replace the fits file in the resoruces with an updated version that includes BMag if you need it.

maxnoe:
After writing the psf stars script, I also realize that the function does not apply space motion, so stars with high proper motions will not be at the correct positions.

ctoennis:
Ok, i'll switch to that tool.

PR #2625, opened 2024-10-14
Add Hipparcos as second star catalog, apply space motion

ctoennis:
This pull request adds a utility to get star positions with proper motion from different catalogues. It allows to cache the catalogue lookup locally and uses cached data before looking up data online again.


src/ctapipe/utils/astro.py
```diff
@@ -4,10 +4,151 @@
...
+
+log = logging.getLogger("main")
+
+__all__ = ["get_bright_stars_with_motion", "get_bright_stars"]
+
+
+CACHE_FILE = Path("~/.psf_stars.ecsv").expanduser()
+
+
+def cut_stars(stars, pointing=None, radius=None, Bmag_cut=None, Vmag_cut=None):
```
mexanick, line 25: I suggest to use "select_stars" as a function name. "cut" is a jargon.

src/ctapipe/utils/astro.py
```diff
@@ -4,10 +4,151 @@
...
-from astropy import units as u
+
+import logging
+from pathlib import Path
+
+import astropy.units as u
 from astropy.coordinates import Angle, SkyCoord
+from astropy.table import Table
+from astropy.time import Time
+from astroquery.vizier import Vizier
```
mexanick, line 15: This is a new dependency. It should be included in pyproject.
@maxnoe is it ok to have it? Or we need to download and stash a cached version of the catalog?

src/ctapipe/utils/astro.py
```diff
@@ -4,10 +4,151 @@
...
+                "provided if radius is given."
+            )
+        separations = stars["ra_dec"].separation(pointing)
+        stars["separation"] = separations
+        stars = stars[separations < radius]
+
+    return stars
+
+
+def get_bright_stars_with_motion(
```
mexanick, line 66: Do we have a reason to keep the getter without proper motion?

src/ctapipe/utils/astro.py
```diff
@@ -4,10 +4,151 @@
...
+    ----------
+    pointing: astropy Skycoord
+       pointing direction in the sky (if none is given, full sky is returned)
+    radius: astropy angular units
+       Radius of the sky region around pointing position. Default: full sky
+    Vmag_cut: float
+        Return only stars above a given apparent magnitude. Default: None (all entries)
+    Bmag_cut: float
+        Return only stars above a given absolute magnitude. Default: None (all entries)
+    catalog: string
```
mexanick, line 84: Create an enum with human-readable catalog names.

src/ctapipe/utils/tests/test_astro.py
```diff
@@ -3,9 +3,28 @@
...
 from astropy import units as u
-from astropy.coordinates import SkyCoord
+from astropy.coordinates import Angle, SkyCoord
 
-from ..astro import get_bright_stars
+from ..astro import get_bright_stars, get_bright_stars_with_motion
+
+
+def test_get_bright_stars_with_motion():
+    """
```
mexanick, line 12: If you add more catalogs, it makes sense to test all of them. Also, it would be nice to test a but more details, or add more comments about the expected outcome of the query.

src/ctapipe/utils/astro.py
maxnoe: No, that's essentially a bug

ctoennis: ok, i'll get rid of it

src/ctapipe/utils/tests/test_astro.py
ctoennis: oka, i will get to that part tomorrow morning

ctoennis:
I extended the tests to check if stars are moving and to check both catalogues. Also i added to the cached metadata which catalog was used. Right now the function will download the catalog data again if you use a different one. 

I was thinking to add to the cached file some part of the filename that marks the catalog used, so that you can cache multiple catalogs.

src/ctapipe/utils/astro.py
maxnoe: I'd prefer having the catalog we need as a resource file, it should be fairly small.

It would be good have the code to create it here in any case, but that can be done without hard-requiring astroquery. it should be kept as an optional dependency.

ctoennis: i made a function to make the catalog file in the last commit and added astroquery to the optional dependencies. 

The function get_bright_stars() should then only look to the resource file with the function get_table_dataset(), no? 
Schould we then have multiple files uploaded for different catalogues?

ctoennis:
For the startracker code each star will need an identifier and not all stars have nice common names available. I suggest we use the HIP number for identification as it is the most available identifier for the objects we can use for pointing and PSF fitting.

docs/changes/2625.features.rst
```diff
@@ -0,0 +1,4 @@
+Add ''get_bright_stars_with_motion'' to ctapipe.utils.astro
```
maxnoe, line 1: Please update the changelog once we have this finalized

src/ctapipe/utils/astro.py
```diff
@@ -4,63 +4,231 @@
...
+from astropy.table import Table
+from astropy.time import Time
+
+log = logging.getLogger("main")
 
 __all__ = ["get_bright_stars"]
 
 
-def get_bright_stars(pointing=None, radius=None, magnitude_cut=None):
+CACHE_FILE = Path("~/.psf_stars.ecsv").expanduser()
```
maxnoe, line 22: This needs to be changed. As discussed in the comments, we want to have the catalogs as resource files in the package and only have the code to create them for developers to update those files if needed.

So this should not be needed at all. *if* we need it, it should be using this function here to obtain an appropriate cache path:
https://github.com/cta-observatory/ctapipe/blob/40faf694867aa570c217c3f9456013684455f6ae/src/ctapipe/utils/download.py#L69

instead of just storing a hidden file in the user's home

src/ctapipe/utils/astro.py
```diff
@@ -4,63 +4,231 @@
...
+    Returns
+    -------
+    Astropy table:
+       List of all stars after cuts with same keys as the input table stars
     """
-    Get an astropy table of bright stars.
 
-    This function is using the Yale bright star catalog, available through ctapipe
-    data downloads.
+    if Bmag_cut is not None and "Bmag" in stars.keys():
```
maxnoe, line 53: Are `Bmag` / `BTmag` and `Vmag` / `VTmag`  really the same?

Why was it needed to add the "T" versions?

src/ctapipe/utils/astro.py
```diff
@@ -4,63 +4,231 @@
...
+            "recno",
+            "RAJ2000",
+            "DEJ2000",
+            "RAICRS",
+            "DEICRS",
+            "pmRA",
+            "pmDE",
+            "Vmag",
+            "Bmag",
+            "BTmag",
```
maxnoe, line 100: Why also the T versions?

src/ctapipe/utils/astro.py
```diff
@@ -4,63 +4,231 @@
...
+    filename: string
+        Name and location of the catalog file to be produced
+    """
+    from astroquery.vizier import Vizier
+
+    vizier = Vizier(
+        catalog=StarCatalogues[catalog].value,
+        columns=[
+            "recno",
+            "RAJ2000",
```
maxnoe, line 92: Please remove the computed "RAJ2000" / "DEJ2000" values. 

These are computed by vizier on the fly and are redundant with the original coordinates and the proper motions.

We anyway need to apply proper motion, so it doesn't matter whether we start from the catalog epoch or j2000

src/ctapipe/utils/astro.py
```diff
@@ -4,63 +4,231 @@
...
+            "RAICRS",
+            "DEICRS",
+            "pmRA",
+            "pmDE",
+            "Vmag",
+            "Bmag",
+            "BTmag",
+            "VTmag",
+        ],
+        row_limit=1000000,
```
maxnoe, line 103: This should be a keyword argument. Or at least there should be a warning if `len(stars) == row_limit`.

src/ctapipe/utils/astro.py
```diff
@@ -4,63 +4,231 @@
...
+    )
+
+    stars = vizier.query_constraints(Vmag="0.0..10.0")[0]
```
maxnoe, line 106: Why limit this to magnitude 0.0? There are a handful of brighters stars. The limit to magnitude 0 is a rule of LST for psf stars, not applicable here.

The constraints should be an option to this function as well.

src/ctapipe/utils/astro.py
```diff
@@ -4,63 +4,231 @@
...
+
+    if "Bmag" not in stars.keys():
+        log.exception("The chosen catalog does not have Bmag data")
```
maxnoe, line 109: This is wrong usage of `log.exception`. `log.exception` is meant only to be used in an `except` block to provide additional information to the traceback:
https://docs.python.org/3/library/logging.html#logging.Logger.exception

This should be an actual exception, so throw a `KeyError`

src/ctapipe/utils/astro.py
```diff
@@ -4,63 +4,231 @@
...
+    if "Vmag" not in stars.keys():
+        log.exception("The chosen catalog does not have Vmag data")
+    stars.meta["Catalog"] = StarCatalogues[catalog].value
+
+    stars.write(CACHE_FILE, overwrite=True)
```
maxnoe, line 114: THis should use the (at the moment unused) `filename` option.

But I would completely remove the `filename` option and just return the table. Querying the catalog and writing it to a file should be separate.

src/ctapipe/utils/astro.py
```diff
@@ -4,63 +4,231 @@
...
+
+
+def get_bright_stars(
+    pointing=None, radius=None, Bmag_cut=None, Vmag_cut=None, catalog="Yale"
```
maxnoe, line 118: `catalog` argument should use the enum

src/ctapipe/utils/astro.py
```diff
@@ -4,63 +4,231 @@
...
+):  # max_magnitude):
```
maxnoe, line 119: remove commented part

maxnoe: For backwards compatibilty, we could think about keeping the `max_magnitude` argument and adding an `magnitude_type=` keyword that can either be `"Vmag"` or `"BMag"`

src/ctapipe/utils/astro.py
```diff
@@ -4,63 +4,231 @@
...
+    stars = None
 
-    starpositions = SkyCoord(
-        ra=Angle(catalog["RAJ2000"], unit=u.deg),
-        dec=Angle(catalog["DEJ2000"], unit=u.deg),
-        frame="icrs",
-        copy=False,
-    )
-    catalog["ra_dec"] = starpositions
+    if CACHE_FILE.exists():
```
maxnoe, line 147: This should use the importlib resources system to load the resources file for the specified catalog

maxnoe (changes requested):
I left some comments in-line.

The main point is that the star catalog(s) should remain a resource file for users and the code to create it is run by the developers when we need to update it.

src/ctapipe/utils/astro.py
ctoennis: some catalogs only have the T versions available

maxnoe: But we only deal with two here, which to my knowledge have Bmag and Vmag

src/ctapipe/utils/astro.py
```diff
@@ -4,63 +4,231 @@
...
 
 __all__ = ["get_bright_stars"]
 
 
-def get_bright_stars(pointing=None, radius=None, magnitude_cut=None):
+CACHE_FILE = Path("~/.psf_stars.ecsv").expanduser()
+
+
+class StarCatalogues(Enum):
+    Yale = "V/50/catalog"  # Yale bright star catalogue
```
maxnoe, line 26: Change the comment to be above and have the form `#: `, this will be picked up by sphinx for the docs

src/ctapipe/utils/astro.py
ctoennis: here is the link to the hippoarcos catalogue: https://vizier.cds.unistra.fr/viz-bin/VizieR-3?-source=I/239/hip_main&-out.max=50&-out.form=HTML%20Table&-out.add=_r&-out.add=_RAJ,_DEJ&-sort=_r&-oc.form=sexa

It has Vmag, but no Bmag. It has BTmag and VTmag available. The Yale catalogue is here:

https://vizier.cds.unistra.fr/viz-bin/VizieR-3?-source=V/50&-out.max=50&-out.form=HTML%20Table&-out.add=_r&-out.add=_RAJ,_DEJ&-sort=_r&-oc.form=sexa

Here it's only Vmag. 

the Nomad catalogue has both Bmag and Vmag:

https://vizier.cds.unistra.fr/viz-bin/VizieR-2

If it you want to not have the T versions i can remove it and add error messages when you try to use Bmag with the Yale and Hippoarcos catalogues.

mexanick: BT and VT are photometry measurements from Tycho catalog, the BT and VT bands are close to the Johnson's UBV, but not exactly the same. See par.2.2 here:
https://www.cosmos.esa.int/documents/532822/553204/sect2_02.pdf/9f2b5393-bc66-4480-89d2-88200aecf5c8

In case Bmag is not provided, but Vmag and B-V is there, you can calculate it. However, in our case, I believe this is not that much important. I suggest to let users decide on the selection and bear the responsibility of a correct use of the fields with an error raised in case the selection can't be calculated.

maxnoe: Sounds good

src/ctapipe/utils/astro.py
maxnoe: Yes. And only if your really need multiple catalogs.

I guess we should keep the yale bright stars for compatibilty. If you require another catalog for your applications, select the most appropriate one and add it.

I.e. after this PR, `get_bright_stars()` should still return the stars of the yale bright star catalog, but with the correct positions after applying the space motion.
Add a second function for the catalog you choose, e.g. `get_hipparcos_stars()`.

src/ctapipe/utils/astro.py
maxnoe: Please decide on which catalog you actually need. Then only add that one.

I don't think we really have the need for a general wrapper around astroquery in ctapipe.

mexanick: I used NOMAD originally, but I think HIPPARCOS is more widely used, so I suggest switching to it.

ctoennis:
I made an update based on all the comments i got. I now need to put the catalog files generated with the functions i wrote to where the importlib is looking for tables.

src/ctapipe/utils/astro.py
```diff
@@ -4,22 +4,113 @@
...
+
+    Parameters
+    ----------
+    stars: astropy table
+        Table of stars, including magnitude and coordinates
+    pointing: astropy Skycoord
+        pointing direction in the sky (if none is given, full sky is returned)
+    radius: astropy angular units
+        Radius of the sky region around pointing position. Default: full sky
+    Vmag_cut: float
```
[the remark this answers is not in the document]
maxnoe: I am not sure I like this, it hides the available options and it is not super obvious that `Vmag=7` actually means `Vmag<7`.

Somewhere else, I proposed keeping the argument as "min_magnitude" and adding a second argument that could be called `band` or `magnitude_column`.

src/ctapipe/utils/astro.py
ctoennis: For the Yale catalog there is no other good coordinates other than RAJ2000 and DEJ2000. I made it the same way as the old function now where the coorinates from the catalog are deleted and you just keep the SkyCoord objects for the star position, this time just with proper motion.

maxnoe (changes requested):
I don't see where the space motion is now applied, seems it is not done in the code, just in one of the tests.

Since astropy currently has an issue with transforming ICRS to AltAz after applying space motion, I think we should add a function here and directly apply space motion in the `get_...` functions, adding an `obstime` argument.

See https://github.com/astropy/astropy/issues/17103 for the issue and workaround

src/ctapipe/utils/astro.py
```diff
@@ -4,63 +4,199 @@
...
     if radius is not None:
         if pointing is None:
             raise ValueError(
                 "Sky pointing, pointing=SkyCoord(), must be "
                 "provided if radius is given."
             )
-        separations = catalog["ra_dec"].separation(pointing)
-        catalog["separation"] = separations
-        catalog = catalog[separations < radius]
+        separations = stars["ra_dec"].separation(pointing)
```
maxnoe, line 75: Selecting stars by separation using ra dec only makes sense after applying space motion, so I think we really need to apply it inside the `get...` functions.

src/ctapipe/utils/astro.py
maxnoe: Ok, then let's use the precomputed RAJ2000 consistently, which also makes it easier since we can use the same epoch for both catalogs. Remove the other RA/DEC then.

src/ctapipe/utils/astro.py
maxnoe: I think we should have no upper limit here, there are very few stars this bright and if the client code as a magnitude limit, it can sort them out.

src/ctapipe/utils/astro.py
```diff
@@ -18,8 +18,18 @@
 
 
 class StarCatalogues(Enum):
-    Yale = "V/50/catalog"  #: Yale bright star catalogue
-    Hipparcos = "I/239/hip_main"  #: hipparcos catalogue
+    Yale = {
```
maxnoe, line 21: Why exactly is this needed?

src/ctapipe/utils/astro.py
```diff
@@ -18,8 +18,18 @@
...
+        "directory": "V/50/catalog",
+        "band": "V",
+        "frame": "J2000",
+        "ID_type": "HR",
+    }  #: Yale bright star catalogue
+    Hipparcos = {
+        "directory": "I/239/hip_main",
+        "band": "V",
```
maxnoe, line 29: Hipparcos has both V and B band, why specify V here?

src/ctapipe/utils/astro.py
mexanick: This loosely follows my suggestion to define applicable catalog metadata: the name/id of the catalog, the columns we are going to query/store (this is missing yet!), and some other metadata like frame. 
Then, the min amplitude cut can be applied on any column (and if attempted to apply to inexisting one, a KeyError should be thrown of course).

maxnoe: Do we really need this, given that we query two catalogs? Again: I don't really see a use case for providing a general high-level wrapper around astroquery here.

I just wanted to have the tools to re-create these resource files so that developers can easily update them if needed.

mexanick: One has to define it somewhere. Right now it is done in the get_catalog functions, here looks more reasonable to me...

src/ctapipe/utils/astro.py
ctoennis: i put that into both get_stars functions. It should now cut on location after motion is applied. The user now needs to specify the time for the lookup

src/ctapipe/utils/astro.py
ctoennis: i made it now so that both B and V can be used in one catalog

src/ctapipe/utils/astro.py
ctoennis: I removed the other coordinate, but the functions for the two catalogues are almost identical now. I can have just one function and bring the catalog name as an argument.

maxnoe: Yes, please factor out common code into a lower-level function.

I expect this to look basically like this:

```
def get_bright_stars(...):
    stars = get_table_dataset("yale_bright_stars.fits.gz")
    return _get_stars(stars, ...)
```
and the same for the other catalog

src/ctapipe/utils/astro.py
```diff
@@ -4,63 +4,218 @@
...
-    This function is using the Yale bright star catalog, available through ctapipe
-    data downloads.
+    catalog_dict = StarCatalogues[catalog].value
 
-    The included Yale bright star catalog contains all 9096 stars, excluding the
-    Nova objects present in the original catalog and is complete down to magnitude
-    ~6.5, while the faintest included star has mag=7.96. :cite:p:`bright-star-catalog`
+    columns = ["RAJ2000", "DEJ2000", "pmRA", "pmDE", catalog_dict["ID_type"]]
+    if "B" in catalog_dict["band"]:
+        columns.append("Bmag")
```
maxnoe, line 117: Is it an error to request columns that do not exist? If not, why this logic instead of requesting all possible columns we need for both?

src/ctapipe/utils/astro.py
```diff
@@ -4,63 +4,218 @@
...
+            )
+        separations = stars["ra_dec"].separation(pointing)
+        stars["separation"] = separations
+        stars = stars[separations < radius]
 
+    return stars
 
-def get_bright_stars(pointing=None, radius=None, magnitude_cut=None):
+
+def get_star_catalog(catalog, min_magnitude=0.0, max_magnitude=10.0, row_limit=1000000):
```
maxnoe, line 90: As commented elsewhere, I don't think we should use a min_magnitude at all. The condition can just be `f"< {max_maginitude}"`

src/ctapipe/utils/astro.py
```diff
@@ -4,63 +4,172 @@
...
+
+    stars["ra_dec"] = SkyCoord(
+        ra=Angle(stars["RAJ2000"], unit=u.deg),
+        dec=Angle(stars["DEJ2000"], unit=u.deg),
+        pm_ra_cosdec=stars["pmRA"].quantity,  # yes, pmRA is already pm_ra_cosdec
+        pm_dec=stars["pmDE"].quantity,
+        frame="icrs",
+        obstime=Time("J2000.0"),
+    )
+    stars["ra_dec"].apply_space_motion(new_obstime=time)
```
maxnoe, line 168: This function returns a new coordinate, you need to assign it:

```
stars["ra_dec"] = stars["ra_dec"].apply_space_motion(new_obstime=time)
``` 

Also, please have a look at the astropy issue I linked for the workaround to transforming these coordinantes then to AltAz.

It would also be good to add a test converting the resulting star coordinates to `AltAz` to make sure this works.

https://docs.astropy.org/en/stable/api/astropy.coordinates.SkyCoord.html#astropy.coordinates.SkyCoord.apply_space_motion

mexanick:
@maxnoe the code is more/less in place, but the tests are yet to be updated. To do so, we need the new catalog caches (existing one from Yale also can't be used as it doesn't have proper motion columns). Can you please add them to the data server?

maxnoe:
@mexanick That's something only @kosack can do

ctoennis:
@maxnoe I'm creating these files right now. give me a few minutes

ctoennis:
@mexanick @maxnoe I made files for the Yale and Hipparcos catalogues. I cut at a magnitude of 8.0 for both. The Yale catalog makes a file of 308 KB, the Hipparcos catalog takes 1.78 MB.

maxnoe:
Thanks, that sounds very reasonable to include as package data in `src/ctapipe/resources/`

you then access them using:

```
from importlib.resources import files, as_file

with as_file(files("ctapipe", "resources/hipparcos_catalog.fits.gz")) as f:
    stars = Table.read(f)
```

ctoennis:
ok, i can change that quickly

src/ctapipe/utils/astro.py
```diff
@@ -4,63 +4,197 @@
...
-        frame="icrs",
-        copy=False,
-    )
-    catalog["ra_dec"] = starpositions
+    stars = get_table_dataset(cat["record"], role="bright star catalog")
 
-    if magnitude_cut is not None:
-        catalog = catalog[catalog["Vmag"] < magnitude_cut]
+    stars["ra_dec"] = SkyCoord(
+        ra=Angle(
```
maxnoe, line 168: I think this can just be `ra=stars["RAJ2000"].quantity`

src/ctapipe/utils/astro.py
```diff
@@ -4,63 +4,197 @@
...
+    }  #: Yale bright star catalogue
+    Hipparcos = {
+        "directory": "I/239/hip_main",
+        "coordinates": {
+            "frame": "icrs",
+            "epoch": "J1991.25",
+            "RA": {"column": "RAICRS", "unit": "deg"},
+            "DE": {"column": "DEICRS", "unit": "deg"},
+            "pmRA": {"column": "pmRA", "unit": "mas/yr"},
+            "pmDE": {"column": "pmDE", "unit": "mas/yr"},
```
maxnoe, line 44: The units should be already included in the table you get from astroquery, no need to hardcore them here

src/ctapipe/utils/astro.py
```diff
@@ -4,63 +4,197 @@
...
+            stars[cat["coordinates"]["RA"]["column"]],
+            unit=u.Unit(cat["coordinates"]["RA"]["unit"]),
+        ),
+        dec=Angle(
+            stars[cat["coordinates"]["DE"]["column"]],
+            unit=u.Unit(cat["coordinates"]["DE"]["unit"]),
+        ),
+        pm_ra_cosdec=stars[
+            cat["coordinates"]["pmRA"]["column"]
+            * u.Unit(cat["coordinates"]["pmRA"]["unit"])
```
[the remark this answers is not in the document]
mexanick: ups, closing bracket in wrong place! thanks!

src/ctapipe/utils/astro.py
mexanick: not for hms (or at least not for Yale BSC):
```
<Column name='pmRA' dtype='float32' unit='arcsec / yr' format='{:6.3f}' description='?Annual proper motion in RA J2000, FK5 system' length=9096>
```

ctoennis:
I switched to having the catalog files in ctapipe/resources. It now works on the Yale catalog. The Hipparcos catalog was too big and the pre-commit hooks raised an error. I also need to change the unit tests tomorrow.

src/ctapipe/utils/astro.py
maxnoe: That's because it's a string, not a value with a unit

maxnoe:
Please make an exception for the pre commit hook.

mexanick:
LGTM. @maxnoe please take a look. Since I contributed to the code myself, an extra eye might be required.

src/ctapipe/utils/astro.py
```diff
@@ -4,63 +4,194 @@
...
-    data downloads.
+class StarCatalogues(Enum):
+    Yale = {
+        "directory": "V/50/catalog",
+        "coordinates": {
+            "frame": "icrs",
+            "epoch": "J2000.0",
+            "RA": {"column": "RAJ2000", "unit": "hourangle"},
+            "DE": {"column": "DEJ2000", "unit": "deg"},
+            "pmRA": {"column": "pmRA", "unit": "arcsec/yr"},
```
maxnoe, line 30: again, units here should not be needed. and the columns are the same in both catalogs.

src/ctapipe/utils/astro.py
```diff
@@ -4,63 +4,194 @@
...
+    return stars_
+
+
+def get_star_catalog(catalog, magnitude_cutoff=8.0, row_limit=1000000):
+    """
+    Utility function to download a star catalog for the get_bright_stars function
+
+    Parameters
+    ----------
+    catalog: string
```
kosack, line 104: Should update this to `catalog: str | StarCatalog` and maybe also add a type hint using the same, since that will make checking nicer. 

```suggestion
def get_star_catalog(catalog: StarCatalog | str, magnitude_cutoff=8.0, row_limit=1000000):
    """
    ...
    catalog: StarCatalog | str
    """
```

Note that fancy language servers like PyCharm actually parse this parameters section and will complain that  "string" is not a python type.

src/ctapipe/utils/tests/test_astro.py
```diff
@@ -2,20 +2,65 @@
...
+from astropy.time import Time
 
 from ..astro import get_bright_stars
 
 
 def test_get_bright_stars():
     """
-    unit test for utils.astro.get_bright_stars(). Tests that only Zeta Tau is
-    returned close to the Crab Nebula as object brighter than mag=3.5.
+    unit test for utils.astro.get_bright_stars_with_motion().
```
maxnoe, line 15: Please add a test transforming the obtained icrs coordinates into altaz.

E.g. `table_yale["ra_dec"].transform_to(AltAz(location=location, obstime=obstime))`

src/ctapipe/utils/astro.py
```diff
@@ -4,63 +4,194 @@
...
+
+def get_bright_stars(
+    time, catalog="Yale", pointing=None, radius=None, magnitude_cut=None
+):
+    """
+    Get an astropy table of bright stars from the Yale bright star catalog
+
+    Parameters
+    ----------
+    time: astropy Time
```
kosack, line 144:
```
time: astropy.time.Time
```

src/ctapipe/utils/astro.py
```diff
@@ -4,63 +4,194 @@
...
+        time to which proper motion is applied
+    catalog : string
```
kosack, line 146: catalog: StarCatalog | str

src/ctapipe/utils/astro.py
```diff
@@ -4,63 +4,194 @@
...
+    stars["ra_dec"] = SkyCoord(
+        ra=Angle(
+            stars[cat["coordinates"]["RA"]["column"]],
+            unit=u.Unit(cat["coordinates"]["RA"]["unit"]),
+        ),
+        dec=Angle(
+            stars[cat["coordinates"]["DE"]["column"]],
+            unit=u.Unit(cat["coordinates"]["DE"]["unit"]),
+        ),
+        pm_ra_cosdec=stars[cat["coordinates"]["pmRA"]["column"]],
```
maxnoe, line 179: simplify to `stars["pmRA"].quantity`

src/ctapipe/utils/astro.py
```diff
@@ -4,63 +4,194 @@
...
+    """
+    Get an astropy table of bright stars from the Yale bright star catalog
+
+    Parameters
+    ----------
+    time: astropy Time
+        time to which proper motion is applied
+    catalog : string
+        name of the catalog to be used available catalogues are 'Yale' and 'Hipparcos'. Default: 'Yale'
+    pointing: astropy Skycoord
```
kosack, line 148: use `astropy.coordinate.SkyCoord`

src/ctapipe/utils/astro.py
```diff
@@ -4,63 +4,194 @@
...
+       pointing direction in the sky (if none is given, full sky is returned)
+    radius: astropy angular units
```
kosack, line 150: `radius: astropy.coordinate.Angle`  ? Is that what you mean by angular units?

src/ctapipe/utils/astro.py
mexanick: yeah, I didn't delete them for consistency with RA/DE, but of course this can be done along with your comment below on `qunatity` conversion. Do you know if this holds for the majority of the catalogs? I don't have much experience with them...

src/ctapipe/utils/astro.py
```diff
@@ -4,63 +4,194 @@
...
 
+__all__ = ["get_star_catalog", "get_bright_stars"]
 
-def get_bright_stars(pointing=None, radius=None, magnitude_cut=None):
-    """
-    Get an astropy table of bright stars.
 
-    This function is using the Yale bright star catalog, available through ctapipe
-    data downloads.
+class StarCatalogues(Enum):
```
maxnoe, line 22: I find it strange to use mutable values as value of an enum.

I'd define the enum like this:

```
class StarCatalog(Enum):
    Yale = auto()
    Hipparcos = auto()
```

and then define the settings as:

```
CATALOG_PROPERTIES = {
    StarCatalog.Hiparcos: {...},
    StarCatalog.Yale: {...},
}
```

I naively tried `get_star_catalog(StarCatalogues.Hipparcos)` which resulted in a very, very unweildy error:

```
KeyError: <StarCatalogues.Hipparcos: {'directory': 'I/239/hip_main', 'coordinates': {'frame': 'icrs', 'epoch': 'J1991.25', 'RA': {'column': 'RAICRS', 'unit': 'deg'}, 'DE': {'column': 'DEICRS', 'unit': 'deg'}, 'pmRA': {'column': 'pmRA', 'unit': 'mas/yr'}, 'pmDE': {'column': 'pmDE', 'unit': 'mas/yr'}}, 'columns': ['RAICRS', 'DEICRS', 'pmRA', 'pmDE', 'Vmag', 'BTmag', 'HIP'], 'record': 'hipparcos_star_catalog'}>
```
```

File: src/ctapipe/utils/astro.py (hunk 29 above)
Comment (maxnoe) [reply src/ctapipe/utils/astro.py:30]: I don't know. But I can only repeat myself, the goal here should not be to define a high-level interface to all catalogs in vizier via astroquery

File: src/ctapipe/utils/astro.py (hunk 30 above)
Comment (maxnoe) [reply src/ctapipe/utils/astro.py:104]: At least currently, giving the enum does not work. You have to give the name of the enum member as string.

Comment (maxnoe) [reply src/ctapipe/utils/astro.py:104]: See https://github.com/cta-observatory/ctapipe/pull/2625/files#r1816939901

File: src/ctapipe/utils/astro.py [hunk 38]
```diff
@@ -4,63 +4,194 @@
...
+
+    vizier = Vizier(
+        catalog=catalog_dict["directory"],
+        columns=catalog_dict["columns"],
+        row_limit=row_limit,
+    )
+
+    stars = vizier.query_constraints(Vmag=f"<{magnitude_cutoff}")[0]
+
+    stars.meta["Catalog"] = StarCatalogues[catalog].value
```
maxnoe, line 131: please also store the magnitude cut in the metadata

src/ctapipe/utils/astro.py
```diff
@@ -4,63 +4,194 @@
...
+            stars[cat["coordinates"]["DE"]["column"]],
+            unit=u.Unit(cat["coordinates"]["DE"]["unit"]),
+        ),
+        pm_ra_cosdec=stars[cat["coordinates"]["pmRA"]["column"]],
+        pm_dec=stars[cat["coordinates"]["pmDE"]["column"]],
+        frame=cat["coordinates"]["frame"],
+        obstime=Time(cat["coordinates"]["epoch"]),
+    )
+    stars["ra_dec"] = stars["ra_dec"].apply_space_motion(new_obstime=time)
+    stars["ra_dec"] = SkyCoord(
```
maxnoe, line 185: Please use the more appropriate workaround given by the astropy maintainers of using the `UnitSphericalRepresentation`

src/ctapipe/utils/astro.py
mexanick: It is not mandatory Angle (i.e. a Quantity with deg unit is not equiv to Angle)

kosack:
One option is to drop astroquery (which queries non-uniform webservces) and use the VO instead, which returns UCDs that describe which columns are the ra/dec, etc.  Something like this (here I used the "new" hipparcos catalog, which also you might want to also try)

```py3
tap = pyvo.dal.TAPService("https://gea.esac.esa.int/tap-server/tap")
result = tap.search("SELECT * from hipparcos_newreduction")

ra_column_name = result.fieldname_with_ucd('pos.eq.ra')
dec_column_name = result.fieldname_with_ucd('pos.eq.dec')
ra_column_name, dec_column_name

catalog = result.to_qtable()  # can then write this to a file, though I notices it loses the UCD metadata for some reason
print(catalog[20:25])

coords = SkyCoord(ra = catalog[ra_column_name], dec=catalog[dec_column_name], frame="icrs")
```

maxnoe:
> Of course the real best thing to do would be to just query Gaia DR3

No, Gaia DR3 is not right for this, it is limited to relatively dark stars:

> with a limiting magnitude of about G ≈ 21 and a bright limit of about G ≈ 3.

mexanick (approved):

mexanick:
I'm not sure what's the reason of the docs failure:
```
home/runner/work/ctapipe/ctapipe/src/ctapipe/utils/astro.py:docstring of ctapipe.utils.astro.get_bright_stars:1: WARNING: py:class reference target not found: ctapipe.utils.astro.StarCatalog [ref.class]
```

any hints?

maxnoe:
> I'm not sure what's the reason of the docs failure:
> 
> ```
> home/runner/work/ctapipe/ctapipe/src/ctapipe/utils/astro.py:docstring of ctapipe.utils.astro.get_bright_stars:1: WARNING: py:class reference target not found: ctapipe.utils.astro.StarCatalog [ref.class]
> ```
> 
> any hints?

The Enum `StarCatalog` is not in `__all__` and thus not documented, which is why it cannot be referenced.

mexanick:
The problem was with the attributes of the named tuple at the end. Now docs build is passing. @maxnoe @kosack you can give it a final review pass.

maxnoe:
> The problem was with the attributes of the named tuple at the end.

The immediate issue was using single-backticks (refernces in sphinx) instead of double-backticks (code), but of course properly documenting the attributes is the best solution.

docs/changes/2625.features.rst
```diff
@@ -0,0 +1,7 @@
+Add possibility to use ``HIPPARCOS`` catalog to get star positions
+
+- add catalogs enumerator to ``ctapipe.utils.astro``
+- update ``get_bright_stars`` in ``ctapipe.utils.astro`` to allow catalog selection
+- ensure application of proper motion
```
maxnoe, line 5: Add note that the API changed, now it requires the `time`  to apply proper motion.

You could also keep backwards compatibility by making this argument optional and applying proper motion only if it is given. But I am also fine with requiring it to make sure users get correct coordinates.

maxnoe:
@ctoennis The files you uploaded now are roughly half the size of what you initially mentioned. Just out of curiosity, did you apply a stricted magnitude selection or did you use uncompressed fits before?

src/ctapipe/utils/astro.py
```diff
@@ -4,63 +4,230 @@
...
+    -------
+    astropy.table.Table
+        List of all stars after cuts with catalog numbers, magnitudes,
+        and coordinates as SkyCoord objects including proper motion.
+    """
+    from astroquery.vizier import Vizier
+
+    if isinstance(catalog, str):
+        catalog = StarCatalog[catalog]
+    catalog_dict = catalog.value
```
maxnoe, line 148: not a dict anymore. Prefer better names to encoding type in variable name. E.g. `catalog_info`

src/ctapipe/utils/astro.py
```diff
@@ -4,63 +4,230 @@
...
+        catalog=catalog_dict.directory,
+        columns=catalog_dict.columns,
+        row_limit=row_limit,
+    )
+
+    stars = vizier.query_constraints(Vmag=f"<{magnitude_cutoff}")[0]
+    meta = catalog_dict._asdict()
+    meta["magnitude_cutoff"] = magnitude_cutoff
+
+    stars.meta["Catalog"] = meta
```
maxnoe, line 160: How are theses nested metadata items handled with the FITS export?

maxnoe: Ah, in an astropy custom way in `COMMENT` cards:
```
COMMENT --BEGIN-ASTROPY-SERIALIZED-COLUMNS--                                    
COMMENT datatype:                                                               
COMMENT - name: RAICRS                                                          
COMMENT   unit: deg                                                             
COMMENT   datatype: float64                                                     
COMMENT   format: '{:12.8f}'                                                    
COMMENT   description: '? alpha, degrees (ICRS, Epoch=J1991.25) (H8)'           
COMMENT   meta: !!omap                                                          
COMMENT   - {ucd: POS_EQ_RA_MAIN}                                               
COMMENT   - {width: 12}                                                         
COMMENT   - {precision: '8'}                                                    
COMMENT   - values: {'null': .nan}                                              
COMMENT - name: DEICRS                                                          
COMMENT   unit: deg                                                             
COMMENT   datatype: float64                                                     
COMMENT   format: '{:12.8f}'                                                    
COMMENT   description: '? delta, degrees (ICRS, Epoch=J1991.25) (H9)'           
COMMENT   meta: !!omap                                                          
COMMENT   - {ucd: POS_EQ_DEC_MAIN}                                              
COMMENT   - {width: 12}                                                         
COMMENT   - {precision: '8'}                                                    
COMMENT   - values: {'null': .nan}                                              
COMMENT - name: pmRA                                                            
COMMENT   unit: mas / yr                                                        
COMMENT   datatype: float64                                                     
COMMENT   format: '{:8.2f}'                                                     
COMMENT   description: '*? Proper motion mu_alpha.cos(delta), ICRS(H12)'        
COMMENT   meta: !!omap                                                          
COMMENT   - {ucd: POS_EQ_PMRA}                                                  
COMMENT   - {width: 8}                                                          
COMMENT   - {precision: '2'}                                                    
COMMENT   - values: {'null': .nan}                                              
COMMENT - name: pmDE                                                            
COMMENT   unit: mas / yr                                                        
COMMENT   datatype: float64                                                     
COMMENT   format: '{:8.2f}'                                                     
COMMENT   description: '*? Proper motion mu_delta, ICRS (H13)'                  
COMMENT   meta: !!omap                                                          
COMMENT   - {ucd: POS_EQ_PMDEC}                                                 
COMMENT   - {width: 8}                                                          
COMMENT   - {precision: '2'}                                                    
COMMENT   - values: {'null': .nan}                                              
COMMENT - name: Vmag                                                            
COMMENT   unit: mag                                                             
COMMENT   datatype: float32                                                     
COMMENT   format: '{:5.2f}'                                                     
COMMENT   description: '? Magnitude in Johnson V (H5)'                          
COMMENT   meta: !!omap                                                          
COMMENT   - {ucd: PHOT_JHN_V}                                                   
COMMENT   - {width: 5}                                                          
COMMENT   - {precision: '2'}                                                    
COMMENT   - values: {'null': .nan}                                              
COMMENT - name: BTmag                                                           
COMMENT   unit: mag                                                             
COMMENT   datatype: float32                                                     
COMMENT   format: '{:6.3f}'                                                     
COMMENT   description: '? Mean BT magnitude (H32)'                              
COMMENT   meta: !!omap                                                          
COMMENT   - {ucd: PHOT_TYCHO_B}                                                 
COMMENT   - {width: 6}                                                          
COMMENT   - {precision: '3'}                                                    
COMMENT   - values: {'null': .nan}                                              
COMMENT - name: HIP                                                             
COMMENT   datatype: int32                                                       
COMMENT   description: Identifier (HIP number) (H1)                             
COMMENT   meta: !!omap                                                          
COMMENT   - {ucd: ID_MAIN}                                                      
COMMENT   - {width: 6}                                                          
COMMENT meta:                                                                   
COMMENT   __serialized_columns__: {}                                            
COMMENT --END-ASTROPY-SERIALIZED-COLUMNS-- 
```

I'd prefer if you store the single metadata fields we need to save using a flat hierarchy that can be directly stored to the FITS header.

src/ctapipe/utils/astro.py
```diff
@@ -4,63 +4,230 @@
...
+        and coordinates as SkyCoord objects including proper motion.
+    """
+    from importlib.resources import files
+
+    if isinstance(catalog, str):
+        catalog = StarCatalog[catalog]
+    cat = catalog.value
+    record = cat.record
+
+    f = files("ctapipe").joinpath(f"resources/{record}.fits.gz")
```
maxnoe, line 201: Proper usage of the importlib API (as shown in the example I posted) is:

```
with as_file(files("ctapipe").joinpath(f"resources/{record}.fits.gz")) as f:
    stars = Table.read(f)
```

maxnoe: This guarantees that the resource file is actually available as a normal file on disk, which isn't the case when the source module is zipped.

maxnoe (changes requested):
This is almost ready to go, some minor last comments inline

src/ctapipe/utils/astro.py
maxnoe: > (i.e. a Quantity with deg unit is not equiv to Angle) 

Angle is a subclass of Quantity that forces angular units and performs additional checks.

maxnoe: I'd write `u.Quantity[angle]`

ctoennis:
> @ctoennis The files you uploaded now are roughly half the size of what you initially mentioned. Just out of curiosity, did you apply a stricted magnitude selection or did you use uncompressed fits before?

i did not apply compression when i gave the estimate.

src/ctapipe/utils/astro.py
ctoennis: I had this before and got an error saying that the output of files and as_files is depricated for use as context. I will reproduce it in a moment

maxnoe: That's strange, there is no mention of any deprecations in the current documentation:
https://docs.python.org/3/library/importlib.resources.html#importlib.resources.as_file

mexanick: I just changed it locally and don't see any deprecation either...

src/ctapipe/utils/astro.py
mexanick: Sorry, I didn't check before whether it is getting in properly.

What if we use json dumps? The result will look like this:

```

> stars = Table.read('./hipparcos.fits.gz')
> stars.meta

OrderedDict([('ID', 'I_239_hip_main'),
             ('NAME', 'I/239/hip_main'),
             ('description', 'The Hipparcos Main Catalogue'),
             ('CATALOG',
              '{"directory": "I/239/hip_main", "coordinates": {"frame": "icrs", "epoch": "J1991.25", "RA": {"column": "RAICRS", "unit": "deg"}, "DE": {"column": "DEICRS", "unit": "deg"}}, "columns": ["RAICRS", "DEICRS", "pmRA", "pmDE", "Vmag", "BTmag", "HIP"], "record": "hipparcos_star_catalog", "magnitude_cutoff": 8.0}')])
```

src/ctapipe/utils/astro.py
ctoennis: re-installed my cta-dev environment and it is now gone. I will implement this

src/ctapipe/utils/astro.py
```diff
@@ -191,15 +191,16 @@ def get_bright_stars(
...
+    from importlib.resources import as_file, files
 
     if isinstance(catalog, str):
         catalog = StarCatalog[catalog]
     cat = catalog.value
     record = cat.record
 
-    f = files("ctapipe").joinpath(f"resources/{record}.fits.gz")
-    stars = Table.read(f)
+    # f = files("ctapipe").joinpath(f"resources/{record}.fits.gz")
```
mexanick, line 201: remove commented out text

ctoennis: sorry, i removed it

ctoennis:
@maxnoe @kosack i think i adressed all comments. Could you have a look again if there is anything left to change?

src/ctapipe/utils/astro.py
maxnoe: I would prefer setting proper FITS headers instead of dumping a json string into a header card that will certainly require using lots of `CONTINUE` cards.

For some of these, there are even reserved appropriate keywords:

* `JEPOCH = <float>` for the epoch
* `RADECSYS = "ICRS"` for the reference frame

mexanick: Now it looks like this:
```
ORIGIN  = 'CTAPIPE '                                                            
JEPOCH  =              1991.25                                                  
RADESYS = 'ICRS    '                                                            
MAGCUT  =                  8.0                                                  
BAND    = 'V       '                                                            
CATALOG = 'Hipparcos'                                                           
REFERENC= 'I/239/hip_main'                                                      
COLUMNS = 'RAICRS_DEICRS_pmRA_pmDE_Vmag_BTmag_HIP' 
```

maxnoe: Great!

maxnoe:
@mexanick Did you update the files after changing how the metadata is stored?

src/ctapipe/utils/astro.py
```diff
@@ -4,63 +4,239 @@
...
+    stars : astropy.table.Table
+        Table of stars, including magnitude and coordinates.
+    pointing : astropy.coordinates.SkyCoord, optional
+        Pointing direction in the sky. If None is given, the full sky is returned.
+    radius : astropy.units.Quantity, optional
+        Radius of the sky region around the pointing position. Default is the full sky.
+    magnitude_cut : float, optional
+        Return only stars above a given apparent magnitude. Default is None (all entries).
+    band : str, optional
+        Wavelength band to use for the magnitude cut. Options are 'Vmag' and 'B'. Default is 'Vmag'.
```
maxnoe, line 94: Really `'B'`? Not `"BTmag"`?

mexanick: This should be changed. Default is Vmag as it is in all or almost all catalogs, but the B-band can be different (just B, Bmag, BTmag...) I will update this.

src/ctapipe/utils/astro.py
```diff
@@ -4,63 +4,241 @@
...
+
+    return stars
+
+
+def get_bright_stars(
+    time: Time,
+    catalog: StarCatalog | str = "Yale",
+    pointing: SkyCoord | None = None,
+    radius: Quantity | None = None,
+    magnitude_cut: float | None = None,
```
maxnoe, line 182: This doesn't include and forward the `band` argument for the `select_stars`, should be added, right?

src/ctapipe/utils/astro.py
```diff
@@ -4,63 +4,241 @@
...
+        stars["ra_dec"]
+        .data.differentials["s"]
+        .represent_as(UnitSphericalCosLatDifferential)
+    )
+    stars.remove_columns(
+        [cat.coordinates["RA"]["column"], cat.coordinates["DE"]["column"]]
+    )
+
+    stars = select_stars(
+        stars, pointing=pointing, radius=radius, magnitude_cut=magnitude_cut
```
maxnoe, line 241: `, band=band`

maxnoe (approved):

docs/changes/2625.features.rst
```diff
@@ -0,0 +1,8 @@
+Add possibility to use ``HIPPARCOS`` catalog to get star positions
```
maxnoe, line 1: The filename here needs to be `feature` or `api` (better, since this PR contains API changes) not `feature*s*`.

The names need to match one of the pre-defined tags in `pyproject.toml`

maxnoe (dismissed):

mexanick:
@kosack can you please give a final pass?

docs/changes/2625.api.rst
```diff
@@ -0,0 +1,8 @@
+Add possibility to use ``HIPPARCOS`` catalog to get star positions
+
+- add catalogs enumerator to ``ctapipe.utils.astro``
+- update ``get_bright_stars`` in ``ctapipe.utils.astro`` to allow catalog selection
+- ensure application of proper motion
+- add possibility to select stars on proximit to a given position and on apparent magnitude
+- move cached star catalogs to ``ctapipe/resources/``
+- API change: ``ctapipe.utils.astro.get_star_catalog`` now requires to give a time to apply proper motion, before there was no time argument
```
kosack, line 8: This doesn't seem to be the case. The function is defined below as:
```py3
def get_star_catalog(
    catalog: str | StarCatalog, magnitude_cutoff: float = 8.0, row_limit: int = 1000000
) -> Table:
```

mexanick: @kosack yes, you're right. Fixed now.

maxnoe (approved):

kosack (approved):

PR #2628, opened 2024-10-28
Stats calc tool

TjarkMiener:
This PR adds a generic stats-calculation tool utilizing the `PixelStatisticsCalculator`.

Related #2542


src/ctapipe/resources/stats_calc_config.yaml
```diff
@@ -0,0 +1,37 @@
+StatisticsCalculatorTool:
```
mexanick, line 1: Since we don't use `yaml` for anything apart from the configurations, I suggest to rename the configuration file to just `tool_name.yaml`, i.e. stripping `_config`.

src/ctapipe/tools/stats_calculation.py
```diff
@@ -0,0 +1,179 @@
...
+        trait=CInt(),
+        default_value=None,
+        allow_none=True,
+        help=(
+            "List of allowed tel_ids, others will be ignored. "
+            "If None, all telescopes in the input stream will be included."
+        ),
+    ).tag(config=True)
+
+    dl1a_column_name = CaselessStrEnum(
```
mexanick, line 60: Is DL1a/b is "official"? Also, I'd perhaps use generic `input_column_name` similar to the output one.

pyproject.toml
```diff
@@ -99,6 +99,7 @@ ctapipe-process = "ctapipe.tools.process:main"
 ctapipe-merge = "ctapipe.tools.merge:main"
 ctapipe-fileinfo = "ctapipe.tools.fileinfo:main"
 ctapipe-quickstart = "ctapipe.tools.quickstart:main"
+ctapipe-stats-calculation = "ctapipe.tools.stats_calculation:main"
```
maxnoe, line 102: I'd prefer a verb here, like the other tools. E.g. `ctapipe-calculate-pixel-statistics`

src/ctapipe/resources/stats_calc_config.yaml
```diff
@@ -0,0 +1,37 @@
...
+  allowed_tels: [1,2,3,4]
+  dl1a_column_name: "image"
+  output_column_name: "statistics"
+
+PixelStatisticsCalculator:
```
mexanick, line 6: I think you want to make this example a bit more complex and apply to the telescopes of a different kind (i.e. with different parameters of the PixelStatisticsCalculator)

src/ctapipe/tools/stats_calculation.py
```diff
@@ -0,0 +1,179 @@
...
+                        [aggregated_stats, aggregated_stats_secondpass]
+                    )
+                    # Sort the stacked aggregated statistic values by starting time
+                    aggregated_stats.sort(["time_start"])
+                else:
+                    self.log.info(
+                        "No faulty chunks found for telescope 'tel_id=%d'. Skipping second pass.",
+                        tel_id,
+                    )
+            # Write the aggregated statistics and their outlier mask to the output file
```
mexanick, line 155: I think logically output writing shall be in the `finish` function.

src/ctapipe/tools/stats_calculation.py
maxnoe: no, in ctapipe we use `DL1_IMAGES` and `DL1_PARAMETERS` to distinguish between things that are per-pixel vs. single quantities per event.

https://ctapipe.readthedocs.io/en/latest/api/ctapipe.io.DataLevel.html

src/ctapipe/resources/stats_calc_config.yaml
```diff
@@ -0,0 +1,37 @@
...
+  stats_aggregator_type: [["type", "*", "SigmaClippingAggregator"]]
+  chunk_shift: 1000
+  faulty_pixels_fraction: 0.1
+  outlier_detector_list: [
```
maxnoe, line 10: Why use the json syntax here? I think using the yaml syntax is much more readable:

```
outlier_detector_list:
   -  name: ...
      apply_to: median
      config: 
       ...
   -  name: ...
      apply_to: median
      config: 
       ...
```

src/ctapipe/tools/stats_calculation.py
maxnoe: I'd also not make this an enum. In the generic tool, users should be able to chose any column that has compatible shape.

maxnoe: This could also be a list of columns, to compute on multiple at the same time.

src/ctapipe/tools/stats_calculation.py
```diff
@@ -0,0 +1,179 @@
...
+        # Initialization of the statistics calculator
+        self.stats_calculator = PixelStatisticsCalculator(
+            parent=self, subarray=subarray
+        )
+        # Read the input data with the 'TableLoader'
+        input_data = TableLoader(input_url=self.input_url)
+        # Get the telescope ids from the input data or use the allowed_tels configuration
+        tel_ids = subarray.tel_ids if self.allowed_tels is None else self.allowed_tels
+        # Read the whole dl1 images
+        self.dl1_tables = input_data.read_telescope_events_by_id(
```
maxnoe, line 111: Why load all data for all telescopes into memory when you then loop over them one by one?

src/ctapipe/tools/stats_calculation.py
maxnoe: That requires keeping all intermediate results for each telescope in memory until `finish`.

I'd write here, so only the data of a single telescope is in RAM at any given time.

src/ctapipe/tools/stats_calculation.py
TjarkMiener: Ok, I changed the column name and also polish the references to DL1a data by using pixel-wise image data which is more descriptive. `ToolConfigurationError` is raised once the column is not found. Having list of columns seems a little bit of an overkill here, which would just make the code more complex. Maybe the aggregation config could be shared between the columns, but especially the outlier detection will be different between the columns.

mexanick (approved):

src/ctapipe/tools/calculate_pixel_stats.py
```diff
@@ -0,0 +1,185 @@
...
+            )
+
+        # Load the subarray description from the input file
+        subarray = SubarrayDescription.from_hdf(self.input_url)
+        # Initialization of the statistics calculator
+        self.stats_calculator = PixelStatisticsCalculator(
+            parent=self, subarray=subarray
+        )
+        # Read the input data with the 'TableLoader'
+        self.input_data = TableLoader(input_url=self.input_url)
```
kosack, line 105: This needs to be configurable, so you need to pass `parent=self` in the constructor.  You should not expicitly set `input_url` here, but rather just make an alias for `TableLoader.input_url`. That will allow other options to be passed, like `focal_length_choice` when needed

src/ctapipe/tools/calculate_pixel_stats.py
```diff
@@ -0,0 +1,185 @@
...
+
+    """
+
+    input_url = Path(
+        help="Input CTA HDF5 files including pixel-wise image data",
+        allow_none=True,
+        exists=True,
+        directory_ok=False,
+        file_ok=True,
+    ).tag(config=True)
```
kosack, line 47: You should remove this option, and instead use the parameter provided by TableLoader.

src/ctapipe/tools/calculate_pixel_stats.py
```diff
@@ -0,0 +1,185 @@
...
+
+    output_path = Path(
+        help="Output filename", default_value=pathlib.Path("monitoring.h5")
+    ).tag(config=True)
+
+    overwrite = Bool(help="Overwrite output file if it exists").tag(config=True)
+
+    aliases = {
+        ("i", "input_url"): "StatisticsCalculatorTool.input_url",
```
kosack, line 78:
```suggestion
        ("i", "input_url"): "TableLoader.input_url",
```

kosack (changes requested):
Currently fails to read prod5 files (which have no EFFECTIVE focal length information).  The tool current fails with a  `focal_length_choice` exception, however it seems there is no way to set the focal length choioce since the `TableLoader` is not set up to be configrable.

src/ctapipe/tools/calculate_pixel_stats.py
```diff
@@ -0,0 +1,185 @@
...
+            {"StatisticsCalculatorTool": {"overwrite": True}},
+            "Overwrite existing files",
+        ),
+    }
+
+    classes = classes_with_traits(PixelStatisticsCalculator)
+
+    def setup(self):
+        # Check that the input and output files are not the same
+        if self.input_url == self.output_path:
```
kosack, line 93: will also need to change all instances of `self.input_url` to be `self.tableloader.input_url`

maxnoe:
> Currently fails to read prod5 files (which have no EFFECTIVE focal length information).

prod 5 files should have effective focal length

src/ctapipe/tools/calculate_pixel_stats.py
```diff
@@ -0,0 +1,185 @@
...
+        # Get the telescope ids from the input data or use the allowed_tels configuration
+        self.tel_ids = (
+            subarray.tel_ids if self.allowed_tels is None else self.allowed_tels
+        )
+
+    def start(self):
+        # Iterate over the telescope ids and calculate the statistics
+        for tel_id in self.tel_ids:
+            # Read the whole dl1 images for one particular telescope
+            dl1_table = self.input_data.read_telescope_events_by_id(
```
kosack, line 115: I get a crash here:

```sh
% ctapipe-calculate-pixel-statistics -i events.dl1.h5 -o  stats.h5

    dl1_table = self.input_data.read_telescope_events_by_id(
                ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/kkosack/Projects/CTA/Working/ctapipe/src/ctapipe/io/tableloader.py", line 1089, in read_telescope_events_by_id
    tel_ids = self.subarray.get_tel_ids(telescopes)
              ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/kkosack/Projects/CTA/Working/ctapipe/src/ctapipe/instrument/subarray.py", line 549, in get_tel_ids
    for telescope in telescopes:
TypeError: 'numpy.int16' object is not iterable
```

Seems to be due to passing an integer instead of a list, which is what is required by `read_telescope_events_by_id`
```suggestion
            dl1_table = self.input_data.read_telescope_events_by_id(
                  telescopes = [tel_id,]
```

How does this work in the tests?

kosack:
image

In the output, how can I tell what column was aggregated? It is always named "statistics" and there is no metadata in the group or tables that contain that information. Wouldn't it be better to name the group like `monitoring/statistics/{input_column_name}`?

src/ctapipe/tools/stats_calculation.py
maxnoe: > ToolConfigurationError is raised once the column is not found. Having list of columns seems a little bit of an overkill here, which would just make the code more complex. Maybe the aggregation config could be shared between the columns, but especially the outlier detection will be different between the columns.

I think the case where you only want to know about a single column is quite rare, you are usually interested in multiple. So having to read all data again to compute metrics on a new column seems very limiting and a loop over columns shouldn't make the code much more complex.

kosack (commented):
A more general comment: with very minor changes, this could be turned into `ctapipe-calculate-stats`, i.e. the ability to compute stats for any column, not just pixel-wise ones. 

- Expose TableLoader as a configurable component (needed anyhow, see above)
- minor modifications to drop assumption on data shape in `calculator.py`.

I would expect e.g. to be able to do:
```
ctapipe-calculate-pixel-statistics -i events-prod5.DL1.h5  --StatisticsAggregator.chunk_size=100 --StatisticsCalculatorTool.input_column_name hillas_length -o length.h5
```
and get the stats on the length parameter.  This is perhaps outside the scope of this PR, but should be kept in mind.  It also relates to @maxnoe's comment that we could change the API to accept a mapping of columns to Aggragators.

kosack (changes requested):
A common error is to have too small a chunk size, but this now results in a  very ugly error and a full trace-back and exception, along with an `UnclosedFileWarning` (bug?)

- THe former should e caught and raises as a `ToolConfigurationError`, so the user gets a nice message. And please explain in the message what parameters controls this (it took me a very long time reading the code to find it), i.e. say `Change --StatisticsAggregator.chunk_size to decrease this`.  
```
2024-11-06 15:14:43,361 ERROR [ctapipe.StatisticsCalculatorTool] (tool.run): Caught unexpected exception: The length of the provided table (853) is insufficient to meet the required statistics for a single chunk of size (2500).
2024-11-06 15:14:43,361 ERROR [ctapipe.StatisticsCalculatorTool] (tool.run): Caught unexpected exception: The length of the provided table (853) is insufficient to meet the required statistics for a single chunk of size (2500).
  File "/Users/kkosack/Projects/CTA/Working/ctapipe/src/ctapipe/core/tool.py", line 431, in run
    self.start()
  File "/Users/kkosack/Projects/CTA/Working/ctapipe/src/ctapipe/tools/calculate_pixel_stats.py", line 134, in start
    aggregated_stats = self.stats_calculator.first_pass(
                       ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/kkosack/Projects/CTA/Working/ctapipe/src/ctapipe/monitoring/calculator.py", line 169, in first_pass
    aggregated_stats = aggregator(
                       ^^^^^^^^^^^
  File "/Users/kkosack/Projects/CTA/Working/ctapipe/src/ctapipe/monitoring/aggregator.py", line 86, in __call__
    raise ValueError(
ValueError: The length of the provided table (853) is insufficient to meet the required statistics for a single chunk of size (2500).
2024-11-06 15:14:43,377 INFO [ctapipe.StatisticsCalculatorTool] (tool.write_provenance): Output:
/Users/kkosack/miniconda3/envs/ctapipe-0.21/lib/python3.12/site-packages/tables/file.py:113: UnclosedFileWarning: Closing remaining open file: /Users/kkosack/Projects/CTA/PipeWork/v0.21.3/events-prod5.DL1.h5
  warnings.warn(UnclosedFileWarning(msg))
```

src/ctapipe/tools/calculate_pixel_stats.py
TjarkMiener: Okay that sounds good!

src/ctapipe/tools/calculate_pixel_stats.py
TjarkMiener: Therefore I'd need to init the `TableLoarder` before performing the check.

TjarkMiener:
> * The former (Unexpected exception) should e caught and raises as a `ToolConfigurationError`, so the user gets a nice message. And please explain in the message what parameters controls this,  i.e. say `Change --StatisticsAggregator.chunk_size to decrease this`.

This is solved now. I added also a test for it.

I do not know why it's raised this `UnclosedFileWarning`. Is this related to my changes or a more general warning?

src/ctapipe/tools/calculate_pixel_stats.py
```diff
@@ -121,6 +121,15 @@ def start(self):
                 instrument=False,
                 pointing=False,
             )[tel_id]
+            # Check if the chunk size does not exceed the table length of the input data
+            if self.stats_calculator.stats_aggregators[
+                self.stats_calculator.stats_aggregator_type.tel[tel_id]
+            ].chunk_size > len(dl1_table):
+                raise ToolConfigurationError(
+                    f"Change --StatisticsAggregator.chunk_size to decrease the chunk size "
+                    f"of the aggregation to at least '{len(dl1_table)}' (table length of the "
```
Hckjs, line 130: Despite the leading 'decrease' -  'at least' could be a bit ambiguous here. Maybe ...' to a maximum of' .. would be more clear.

TjarkMiener:
> In the output, how can I tell what column was aggregated? It is always named "statistics" and there is no metadata in the group or tables that contain that information. Wouldn't it be better to name the group like `monitoring/statistics/{input_column_name}`? (i.e. maybe set the default of output_column_name to be the value of input_column_name? And also add the input_column namein the output table's metadata (`table.meta['input_column_name']=input_column_name`)

This is a good point. I think we need to metadata here. However, in the current base functionality the `input_column_name` is either `image`, `peak_time`, or `variance`. So the input_url  is also needed in the metadata to understand which quantities were actually aggregated. I'm adding this information to the metadata. The `output_column_name` is needed to configurable and organize the different output tables. (Maybe we should rename this config parameter to `output_table_name` to be more correct).
So in the end the user should put in the config e.g.:
TableLoader.input_url: "path/to/pedestal.dl1.h5"
input_column_name: "image"
output_table_name: "pedestal"

Would this sounds good to you?

--
Another point you raised about the tree schema. I think it should be `monitoring/telescope/...` since it is telescope-wise monitoring data.

src/ctapipe/tools/tests/test_calculate_pixel_stats.py
```diff
@@ -100,3 +100,18 @@ def test_tool_config_error(tmp_path, dl1_image_file):
             cwd=tmp_path,
             raises=True,
         )
+    # Check if ToolConfigurationError is raised
+    # when the chunk size is larger than the number of events in the input file
+    with pytest.raises(ToolConfigurationError):
```
mexanick, line 105: the tool will raise `ToolConfigurationError` in a few situations. Please test all of them and use regexp matching to check whether a correct error message is displayed, e.g. 

```suggestion
    with pytest.raises(ToolConfigurationError, match="Change --StatisticsAggregator.chunk_size")):
        ...
```

src/ctapipe/tools/calculate_pixel_stats.py
```diff
@@ -0,0 +1,194 @@
...
+                    # Sort the stacked aggregated statistic values by starting time
+                    aggregated_stats.sort(["time_start"])
+                else:
+                    self.log.info(
+                        "No faulty chunks found for telescope 'tel_id=%d'. Skipping second pass.",
+                        tel_id,
+                    )
+            # Add metadata to the aggregated statistics
+            aggregated_stats.meta["input_url"] = self.input_data.input_url
+            aggregated_stats.meta["input_column_name"] = self.input_column_name
```
maxnoe, line 169: I'd not add this as `input_column_name` but as prefix for the columns. Using prefixes you can also aggregate multiple columns easily.  

See e.g. how we write tables in `ctapipe-apply-models`.

mexanick: @maxnoe , we discussed this with @TjarkMiener offline and we don't see much use cases for this. Also, having the metadata as a prefix for the column name can make queries awkward and schemas inflexible (i.e. set of aggregators will be fixed for each input source). Also, in this case, something like pandas.MultiIndex will be more appropriate in my opinion, but again, I don't see much use. I'd leave this as is for the moment, and modify if a corresponding use case appears.

src/ctapipe/tools/calculate_pixel_stats.py
```diff
@@ -0,0 +1,194 @@
...
+                    )
+                    # Sort the stacked aggregated statistic values by starting time
+                    aggregated_stats.sort(["time_start"])
+                else:
+                    self.log.info(
+                        "No faulty chunks found for telescope 'tel_id=%d'. Skipping second pass.",
+                        tel_id,
+                    )
+            # Add metadata to the aggregated statistics
+            aggregated_stats.meta["input_url"] = self.input_data.input_url
```
mexanick, line 168: Add source event type to the metadata, so we know whether this is calculated from FF, pedestal, or some other type of events.

src/ctapipe/tools/calculate_pixel_stats.py
maxnoe: The usecase is consistency with what we doe elsewhere and loading such data with e.g.`TableLoader`. 

I.e. my use case would be to load these data and look add aggregated statistics for multiple columns.

This has to match waht e.g. is planned for quality pipe, right? Aggregating high-level information for many of the data columns and other things and loading them at the same time, producing control plots etc.

But fine to do in a follow up PR.

I really don't like the limitation to a single column though.

mexanick: To me what you want sounds like a pandas.MultiIndex, which is not supported in the current way of astropy, and I'm not sure whether it will ever be supported like that... Also, aggregating (flattening) and joining can be done elsewhere (e.g. when QualPipe reads it).

maxnoe: I meant that independent of the columns names / data organization I don't like that this tool can only aggregate a single column, and that using it for multiple requires separate runs of the tool with separate output files, performing the expensive data loading again and again.

src/ctapipe/tools/tests/test_calculate_pixel_stats.py
TjarkMiener: Good point. Actually, I found that the first raise pytest wasn't raised on the 'column' check but on the 'chunk_size' check which comes before in the tool.

mexanick (dismissed):

src/ctapipe/tools/calculate_pixel_stats.py
```diff
@@ -0,0 +1,194 @@
...
+from ctapipe.io.tableloader import TableLoader
+from ctapipe.monitoring.calculator import PixelStatisticsCalculator
+
+
+class StatisticsCalculatorTool(Tool):
+    """
+    Perform statistics calculation for pixel-wise image data
+    """
+
+    name = "StatisticsCalculatorTool"
```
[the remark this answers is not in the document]
TjarkMiener: The suggested renaming makes sense since the tool and the used components are tailored to pixel-wise quantities. First, we planned to re-used the components for any monitoring data. However, for most of the monitoring data, the [astropy built-in aggregation](https://docs.astropy.org/en/stable/table/operations.html#aggregation) should be sufficient.

mexanick (approved):

mexanick:
@maxnoe @kosack @FrancaCassol The code is ready for review, please take a look.

kosack (approved):

mexanick:
I'd wait for @maxnoe review, then we can merge.

src/ctapipe/tools/calculate_pixel_stats.py
```diff
@@ -0,0 +1,194 @@
...
+
+    def start(self):
+        # Iterate over the telescope ids and calculate the statistics
+        for tel_id in self.tel_ids:
+            # Read the whole dl1 images for one particular telescope
+            dl1_table = self.input_data.read_telescope_events_by_id(
+                telescopes=[
+                    tel_id,
+                ],
+                dl1_images=True,
```
maxnoe, line 114: These are all configuration options. By hardcoding them here as kwargs, you prevent users from overriding them via the config.

maxnoe (approved):

src/ctapipe/tools/calculate_pixel_stats.py
mexanick: @maxnoe fixed

maxnoe:
Needs update with main to fix docs build

mexanick (approved):

Issue #1721, opened 2021-05-16
backward compatibility of data model


kosack:
When we merge #1426 for example, we will break reading data files produced with a previous version of ctapipe. 

This brings up the question: *should we support some level of backward compatibility in DL1/DL2 outputs and how?*

This has already been an issue when we have updated the *instrument* data model in the past, and something we should be careful about.  Since we store both the table version (TAB_VER) for instrument outputs, and the DATA MODEL VERSION for the DL1/DL2 files themselves, we should be able to support reading old data, at the expense of keeping around some old code (e.g. renaming `from_table()`  to `from_table_v1()` and having `from_table()` dispatch the correct version.

Clearly we want this after v1.0.0 of ctapipe, but since people are using DL1 files now, it could already be critical.

maxnoe:
@kosack No, you can still read them. I build it in a backwards compatible way.

bregeon:
just a note that it's fast enough to reprocess "simulated" reference datasets with a new version of ctapipe at any time, so loosing backwards compatibility at some point should not be an issue from this point of view... however it might be a different story for LST1 data.

kosack:
I think some backward compatibility is necessary, as it's a bad user experience if we cannot repeat a particular analysis because the data model changed (otherwise it can be hard to compare benchmarks between versions)

maxnoe:
We now version the data model and raise clear errors for unsupported versions. We still might to need breaking changes in future, so I think this is all we can do for now.

PR #2629, opened 2024-10-28
Hillas psi uncertainty ruo

ruoyushang:
Adding estimation of image psi uncertainty and uncertainty of image cog along the rotated y-axis in hillas.py.
Also adding a test script to show the validation of psi uncertainty estimation using toy models.

### DIFF
--- a/src/ctapipe/containers.py
+++ b/src/ctapipe/containers.py
@@ -302,6 +302,9 @@ class CameraHillasParametersContainer(BaseHillasParametersContainer):
     width_uncertainty = Field(nan * u.m, "uncertainty of width", unit=u.m)
     psi = Field(nan * u.deg, "rotation angle of ellipse", unit=u.deg)
     psi_uncertainty = Field(nan * u.deg, "uncertainty of psi", unit=u.deg)
+    b_uncertainty = Field(
+        nan * u.m, "rotated centroid y coordinate uncertainty", unit=u.m
+    )
 
 
 class HillasParametersContainer(BaseHillasParametersContainer):
@@ -331,6 +334,9 @@ class HillasParametersContainer(BaseHillasParametersContainer):
     width_uncertainty = Field(nan * u.deg, "uncertainty of width", unit=u.deg)
     psi = Field(nan * u.deg, "rotation angle of ellipse", unit=u.deg)
     psi_uncertainty = Field(nan * u.deg, "uncertainty of psi", unit=u.deg)
+    b_uncertainty = Field(
+        nan * u.m, "rotated centroid y coordinate uncertainty", unit=u.m
+    )
 
 
 class LeakageContainer(Container):
--- a/src/ctapipe/containers.py
+++ b/src/ctapipe/containers.py
@@ -301,6 +301,7 @@ class CameraHillasParametersContainer(BaseHillasParametersContainer):
     width = Field(nan * u.m, "standard spread along the minor-axis", unit=u.m)
     width_uncertainty = Field(nan * u.m, "uncertainty of width", unit=u.m)
     psi = Field(nan * u.deg, "rotation angle of ellipse", unit=u.deg)
+    psi_uncertainty = Field(nan * u.deg, "uncertainty of psi", unit=u.deg)
 
 
 class HillasParametersContainer(BaseHillasParametersContainer):
@@ -329,6 +330,7 @@ class HillasParametersContainer(BaseHillasParametersContainer):
     width = Field(nan * u.deg, "standard spread along the minor-axis", unit=u.deg)
     width_uncertainty = Field(nan * u.deg, "uncertainty of width", unit=u.deg)
     psi = Field(nan * u.deg, "rotation angle of ellipse", unit=u.deg)
+    psi_uncertainty = Field(nan * u.deg, "uncertainty of psi", unit=u.deg)
 
 
 class LeakageContainer(Container):
--- a/src/ctapipe/image/hillas.py
+++ b/src/ctapipe/image/hillas.py
@@ -87,6 +87,7 @@ def hillas_parameters(geom, image):
     unit = geom.pix_x.unit
     pix_x = geom.pix_x.to_value(unit)
     pix_y = geom.pix_y.to_value(unit)
+    pix_area = geom.pix_area.to_value(unit**2)
     image = np.asanyarray(image, dtype=np.float64)
 
     if isinstance(image, np.ma.masked_array):
@@ -131,7 +132,7 @@ def hillas_parameters(geom, image):
     # avoid divide by 0 warnings
     # psi will be consistently defined in the range (-pi/2, pi/2)
     if length == 0:
-        psi = skewness_long = kurtosis_long = np.nan
+        psi = psi_uncert = skewness_long = kurtosis_long = np.nan
     else:
         if vx != 0:
             psi = np.arctan(vy / vx)
@@ -139,13 +140,23 @@ def hillas_parameters(geom, image):
             psi = np.pi / 2
 
         # calculate higher order moments along shower axes
-        longitudinal = delta_x * np.cos(psi) + delta_y * np.sin(psi)
+        cos_psi = np.cos(psi)
+        sin_psi = np.sin(psi)
+        longi = delta_x * cos_psi + delta_y * sin_psi
+        trans = delta_x * -sin_psi + delta_y * cos_psi
 
-        m3_long = np.average(longitudinal**3, weights=image)
+        m3_long = np.average(longi**3, weights=image)
         skewness_long = m3_long / length**3
 
-        m4_long = np.average(longitudinal**4, weights=image)
+        m4_long = np.average(longi**4, weights=image)
         kurtosis_long = m4_long / length**4
+
+        # lsq solution to determine uncertainty on phi
+        W = np.diag(image / pix_area)
+        X = np.column_stack([longi, np.ones_like(longi)])
+        cov = np.linalg.inv(X.T @ W @ X)
+        p = cov @ X.T @ W @ trans
+        psi_uncert = np.sqrt(cov[0, 0] / np.cos(p[0]) ** 2)
 
     # Compute of the Hillas parameters uncertainties.
     # Implementation described in [hillas_uncertainties]_ This is an internal MAGIC document
@@ -204,6 +215,7 @@ def hillas_parameters(geom, image):
         width=u.Quantity(width, unit),
         width_uncertainty=u.Quantity(width_uncertainty, unit),
         psi=Angle(psi, unit=u.rad),
+        psi_uncertainty=Angle(psi_uncert, unit=u.rad),
         skewness=skewness_long,
         kurtosis=kurtosis_long,
     )
--- a/src/ctapipe/image/hillas.py
+++ b/src/ctapipe/image/hillas.py
@@ -132,7 +132,7 @@ def hillas_parameters(geom, image):
     # avoid divide by 0 warnings
     # psi will be consistently defined in the range (-pi/2, pi/2)
     if length == 0:
-        psi = psi_uncert = skewness_long = kurtosis_long = np.nan
+        psi = psi_uncert = b_uncert = skewness_long = kurtosis_long = np.nan
     else:
         if vx != 0:
             psi = np.arctan(vy / vx)
@@ -154,9 +154,10 @@ def hillas_parameters(geom, image):
         # lsq solution to determine uncertainty on phi
         W = np.diag(image / pix_area)
         X = np.column_stack([longi, np.ones_like(longi)])
-        cov = np.linalg.inv(X.T @ W @ X)
-        p = cov @ X.T @ W @ trans
-        psi_uncert = np.sqrt(cov[0, 0] / np.cos(p[0]) ** 2)
+        lsq_cov = np.linalg.inv(X.T @ W @ X)
+        p = lsq_cov @ X.T @ W @ trans
+        psi_uncert = np.sqrt(lsq_cov[0, 0] + p[0] * p[0])
+        b_uncert = np.sqrt(lsq_cov[1, 1] * (1.0 + np.sin(p[0]) * np.sin(p[0])))
 
     # Compute of the Hillas parameters uncertainties.
     # Implementation described in [hillas_uncertainties]_ This is an internal MAGIC document
@@ -201,6 +202,8 @@ def hillas_parameters(geom, image):
             width=u.Quantity(width, unit),
             width_uncertainty=u.Quantity(width_uncertainty, unit),
             psi=Angle(psi, unit=u.rad),
+            psi_uncertainty=Angle(psi_uncert, unit=u.rad),
+            b_uncertainty=u.Quantity(b_uncert, unit),
             skewness=skewness_long,
             kurtosis=kurtosis_long,
         )
@@ -216,6 +219,7 @@ def hillas_parameters(geom, image):
         width_uncertainty=u.Quantity(width_uncertainty, unit),
         psi=Angle(psi, unit=u.rad),
         psi_uncertainty=Angle(psi_uncert, unit=u.rad),
+        b_uncertainty=u.Quantity(b_uncert, unit),
         skewness=skewness_long,
         kurtosis=kurtosis_long,
     )
--- a/src/ctapipe/tests/test_hillas_psi_unc_toy_model.py
+++ b/src/ctapipe/tests/test_hillas_psi_unc_toy_model.py
@@ -1,0 +1,119 @@
+import os
+
+import astropy.units as u
+import matplotlib.pyplot as plt
+import numpy as np
+from scipy.stats import norm
+from tqdm.auto import tqdm
+
+from ctapipe.image import hillas_parameters
+from ctapipe.image.cleaning import tailcuts_clean
+from ctapipe.image.toymodel import Gaussian
+from ctapipe.instrument import CameraGeometry
+
+ctapipe_output = os.environ.get("CTAPIPE_OUTPUT_PATH")
+ctapipe_input = os.environ.get("CTAPIPE_SVC_PATH")
+
+rng = np.random.default_rng(0)
+
+# cam = CameraGeometry.from_name("LSTCam")
+cam = CameraGeometry.from_name("NectarCam")
+# cam = CameraGeometry.from_name("SCTCam")
+
+true_width = 0.05 * u.m
+true_length = 0.3 * u.m
+true_psi = 45 * u.deg
+true_x = 0.5 * u.m
+true_y = -0.2 * u.m
+
+image_intensity = 1000
+test_nsb_level_pe = 3
+# test_nsb_level_pe = 5
+
+# n_sample = 10
+n_sample = 1000
+
+model = Gaussian(true_x, true_y, true_length, true_width, true_psi)
+
+
+def sample_no_noise_no_cleaning():
+    _, signal, _ = model.generate_image(
+        cam, intensity=image_intensity, nsb_level_pe=0, rng=rng
+    )
+    h = hillas_parameters(cam, signal)
+    return h
+
+
+def sample_no_noise_with_cleaning():
+    _, signal, _ = model.generate_image(
+        cam, intensity=image_intensity, nsb_level_pe=0, rng=rng
+    )
+
+    mask = tailcuts_clean(
+        cam,
+        signal,
+        3.0 * test_nsb_level_pe,
+        2.0 * test_nsb_level_pe,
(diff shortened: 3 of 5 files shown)


src/ctapipe/containers.py
```diff
@@ -301,6 +301,10 @@ class CameraHillasParametersContainer(BaseHillasParametersContainer):
     width = Field(nan * u.m, "standard spread along the minor-axis", unit=u.m)
     width_uncertainty = Field(nan * u.m, "uncertainty of width", unit=u.m)
     psi = Field(nan * u.deg, "rotation angle of ellipse", unit=u.deg)
+    psi_uncertainty = Field(nan * u.deg, "uncertainty of psi", unit=u.deg)
+    b_uncertainty = Field(
```
maxnoe, line 305: This needs a better name. Also, for what is this additional uncertainty needed?

ruoyushang: The axis of the image is modeled with the function y = a*x + b. After rotating the image (so the axis is horizontal), the uncertainty of a = psi uncertainty. However, the axis uncertainty along the transverse direction should be the combination of the psi uncertainty and the b uncertainty, i.e. transverse uncertainty sigma_t = pow( pow(d*sigma_psi,2) + pow(sigma_b,2) , 0.5), where d is the longitudinal coordinate of the test point.

ruoyushang: How about "transverse_cog_uncertainty" ?

ruoyushang:
> @ruoyushang to get the CI to work, you must add a file called `docs/changes/2629.feature.rst` with a human-readable description of the change that will go in the final changelog. After that we can merge this.

@kosack Thank you for the information! I just added the file with a description. However, I also see this error when I pushed the commit: 
CI / lint (pull_request) Failing after 25s. --> Is this critical?

ruoyushang:
@Tobychev Thank you for the information. Can you please show me how to find this bug? How do I use this pre-commit hook? I did pre-commit install following this link (https://ctapipe.readthedocs.io/en/latest/developer-guide/getting-started.html), but I don't know what to do next.

maxnoe:
You can run it manually now to fix the past commits:

```
git fetch
pre-commit run --files $(git diff --name-only origin/main)
```

It will make changes, add and commit them. Some checks might also require manual changes, edit the files accordingly and then also add and commit.

ruoyushang:
> You can run it manually now to fix the past commits:
> 
> ```
> git fetch
> pre-commit run --files $(git diff --name-only origin/main)
> ```
> 
> It will make changes, add and commit them. Some checks might also require manual changes, edit the files accordingly and then also add and commit.

Hi @maxnoe , thank you for the feedback. I was able to fix the error. Now there is one remaining failed check that complains about duplicated lines: 6.56% Duplicated Lines (%) on New Code (is greater than 3.00%)

Is this something that requires fixing, or we can proceed to merging?

src/ctapipe/containers.py
```diff
@@ -301,6 +301,10 @@ class CameraHillasParametersContainer(BaseHillasParametersContainer):
     width = Field(nan * u.m, "standard spread along the minor-axis", unit=u.m)
     width_uncertainty = Field(nan * u.m, "uncertainty of width", unit=u.m)
     psi = Field(nan * u.deg, "rotation angle of ellipse", unit=u.deg)
+    psi_uncertainty = Field(nan * u.deg, "uncertainty of psi", unit=u.deg)
+    transverse_cog_uncertainty = Field(
+        nan * u.m, "rotated centroid y coordinate uncertainty", unit=u.m
```
Tobychev, line 306: I have a hard time understanding how "rotated centroid y coordinate uncertainty" and "uncertainty on the center of gravity along the transverse axis of the image" can mean the same thing, and I think the definition you give in the changelog is much clearer.

Tobychev (dismissed):

Tobychev:
> Hi @maxnoe , thank you for the feedback. I was able to fix the error. Now there is one remaining failed check that complains about duplicated lines: 6.56% Duplicated Lines (%) on New Code (is greater than 3.00%)
> 
> Is this something that requires fixing, or we can proceed to merging?

@ruoyushang Looking at what Sonarqube complains about, its actually an old sin that your four additional lines caused to be judged anew by the CI. It also looks a bit annoying to fix elegantly because the two different return Containers have different argument names. 

So you could probably resolve the issue by stuffing all the arguments into a list and feed it into `CameraHillasParametersContainer` or `CameraHillasParametersContainer` depending on the condition, but it'll not be as pretty.

maxnoe:
@Tobychev We can just ignore this duplication for now, we anyway want to remove the `CameraFrame` part and that wil then get rid of the duplication.

maxnoe:
More critical here is to do some performance testing: how do these new computations affect the processing time?

ruoyushang:
Hi @Tobychev @maxnoe ,

I calculated the time needed for the uncertainty estimation with the following code:

`      tic_psi_unc = time.perf_counter()
        # lsq solution to determine uncertainty on phi
        W = np.diag(image / pix_area)
        X = np.column_stack([longi, np.ones_like(longi)])
        lsq_cov = np.linalg.inv(X.T @ W @ X)
        p = lsq_cov @ X.T @ W @ trans
        psi_uncert = np.sqrt(lsq_cov[0, 0] + p[0] * p[0]) * (
            1.0 + pow(np.tan(width / length * np.pi / 2.0), 2)
        )
        transverse_cog_uncert = np.sqrt(
            lsq_cov[1, 1] * (1.0 + np.sin(p[0]) * np.sin(p[0]))
        )
        toc_psi_unc = time.perf_counter()
        print(f"image size = {size:0.1f}, time spent: {toc_psi_unc-tic_psi_unc:0.6f} sec")
`
Here are several images and the time spent for the uncertainty calculation:

image size = 551.4, time spent: 0.000064 sec
image size = 500.3, time spent: 0.000051 sec
image size = 138.6, time spent: 0.000049 sec
image size = 2811.4, time spent: 0.000051 sec
image size = 2788.4, time spent: 0.000051 sec
image size = 2774.8, time spent: 0.000051 sec
image size = 45.7, time spent: 0.004733 sec

ruoyushang:
Hi @maxnoe @Tobychev ,

Is the processing time test sufficient for the approval?

Best,
Ruo

src/ctapipe/containers.py
kosack: Yeah, i would agree here, just change this to be 
```python
"uncertainty on the center of gravity along the transverse axis of the image"
```

src/ctapipe/containers.py
```diff
@@ -329,6 +333,10 @@ class HillasParametersContainer(BaseHillasParametersContainer):
     width = Field(nan * u.deg, "standard spread along the minor-axis", unit=u.deg)
     width_uncertainty = Field(nan * u.deg, "uncertainty of width", unit=u.deg)
     psi = Field(nan * u.deg, "rotation angle of ellipse", unit=u.deg)
+    psi_uncertainty = Field(nan * u.deg, "uncertainty of psi", unit=u.deg)
+    transverse_cog_uncertainty = Field(
+        nan * u.deg, "rotated centroid y coordinate uncertainty", unit=u.deg
```
kosack, line 338: and here as well

kosack (changes requested):
This needs a unit test.  Maybe construct a simple dummy image where the result is something you can compute by hand, to check if the output is what you expect?

src/ctapipe/containers.py
ruoyushang: Hi @kosack ,

Thank you very much for the feedback. I have update the description for the variable transverse_cog_uncertainty.

"This needs a unit" --> Do you mean a dimension analysis? Please see the attach slides:
[website_hillas_psi_uncertainty.pdf](https://github.com/user-attachments/files/19527295/website_hillas_psi_uncertainty.pdf)

From P7 to P10, you can see how the uncertainty of the image is calculated. On P10 you can also find the dimensionality check.

P12 explains how one can further derive the source location uncertainty from the individual image uncertainties.

P13 shows that there is a good correlation between the estimated uncertainty and the observed reconstruction error in simulated events.

P14-P23 show a few selected examples of the estimated uncertainty for individual images and the combined source location uncertainty.

Please let me know if you have further questions. Thank you.

Best regards,
Ruo

kosack: No, I mean it needs a [unit test](https://en.wikipedia.org/wiki/Unit_testing), not scientific units.  Meaning there should be a test in `ctapipe/image/tests/test_hillas.py` that checks that these new parameters give the expected results on some known input event(s).   That will ensure among other things that if anyone changes or refactors this code in the future, that they do not break the algorithm.  That may happen soon for example if we decided to speed it up.     You can see the other tests in that file for some examples.

ruoyushang: Hi @kosack,

Thank s for the explanation. The code that calculates the image psi angle uncertainty and the cog uncertainty is in the function `hillas_parameters`, which has been used in multiple tests in `ctapipe/image/tests/test_hillas.py`.
I just ran `make test` from the top-level ctapipe directory, and I see that `src/ctapipe/image/tests/test_hillas.py` passed the test.
The only failure came from `src/ctapipe/reco/tests/test_HillasReconstructor.py`, but I think this failure is not related. Here is the error message from `test_HillasReconstructor.py`:
`AssertionError: attr alt_uncert not matching, camera: {'alt': <Latitude 1.22031451 rad>, 'alt_uncert': <Quantity 0. deg>, ...`

Since it has been used in multiple tests in `test_hillas.py`, do I still need to add a new test in `test_hillas.py`?

Best,
Ruo

maxnoe: But none of these tests make sure that the values for the new fields you compute are actually correct.

So yes, new tests need to be added (or you can add new checks to the existing tests). But we need tests that actually verify the values that are computed.

maxnoe:
> I think we shouldn't yet worry too much about processing time, as these will be useful parameters scientifically. We can speed it up later if we need to with numba or something, or have an option to disable it if they are not used in some analyses.

I disagree here a bit. We have already spent quite some work on making sure these image parameter functions are efficient. So I think modifying them, we cannot disregard the performance impact.

As I said, I'd like to see a comparison to before for:

- processing a standard simulation file to dl1
- just the pure hillas parameter function

In the past, I used this benchmark here to time the hillas parameters:

```
from ctapipe.image.tests.test_hillas import create_sample_image
from ctapipe.image.hillas import hillas_parameters
import numpy as np
from timeit import repeat
geom, image, clean_mask = create_sample_image("30d")
code = '''
hillas = hillas_parameters(geom[clean_mask], image[clean_mask])
'''
exec(code)
N = 1000
times = np.array(repeat(code, number=N, repeat=10, globals=globals()))
print(f'{np.mean(times / N) * 1e6:.2f} ± {np.std(times) / N * 1e6:.2f} µs')
```

maxnoe:
Running this, I get:

Here: 166.39 ± 1.32 µs
Main: 133.81 ± 1.46 µs

I would call this somewhat acceptable, although the impact is still quite high for just adding two more fields.

kosack:
Probably we should later come back to #1512 or #1721, which got closed during the PR cleanup, but maybe should be resurrected. These functions look like good candidates for jitting, but that can be a separate PR

src/ctapipe/containers.py
kosack: Yes, the lines are *covered* by other tests, but that only means they can execute, not that they give the right answer. We try to always add a test for all new computations that ensures not only that the computation works as expected, but also as a "guard" against other developers accidentally breaking it in the future, which is a big concern when we have many developer and lots of inter-connected code.   For example, if I decide t go in and speed up the computation later, how can I be sure I still get the right answer?

kosack: https://ctapipe.readthedocs.io/en/latest/developer-guide/code-guidelines.html#unit-tests

ruoyushang: Hi @kosack @maxnoe , 

Thank you for explaining the unit test and why it is needed.
I added a test function `test_psi_uncertainty` in `test_hillas.py`. This test function contains three tests:

Test 1: Check if a symmetric image gives a large psi uncertainty (there is no primary axis for symmetric images).
Test 2: Compare two images with identical morphology but different intensities. The second image has an intensity twice larger than that of the first image. The psi uncertainty for the first image should be larger than the psi uncertainty for the second image by a factor of sqrt(2).
Test 3: Check if the `hillas_parameter()`function returns a psi uncertainty that is consistent with the calculation defined within the `test_psi_uncertainty` function. This prevents the future modification of `hillas_parameter` breaks the uncertainty estimation.

A local run of `make test` shows that it passed `test_psi_uncertainty`.
Please check out the new test and let me know if this looks okay to you.

To really check if the uncertainty estimation agrees with the observed error in psi angle, I think we should run multiple simulations and plot uncertainty against error, for example, P13 in [website_hillas_psi_uncertainty.pdf](https://github.com/user-attachments/files/19527295/website_hillas_psi_uncertainty.pdf).

Thank you and best regards,
Ruo

ruoyushang: Hi @kosack @maxnoe ,

I would like to continue this conversation. Do these new tests test_hillas.py look good?

Thank you and best regards,
Ruo

kosack (dismissed):

ruoyushang:
Thank you to @kosack for approving the request!
@maxnoe Could you please have a look at the unit test and approve this request if it looks good to you? Thanks.

maxnoe:
@ruoyushang Sure, I will take a look. Please also update the pull request branch here with main though, this is the reason why the readthedocs build is failing.

docs/changes/2629.feature.rst
```diff
@@ -0,0 +1,15 @@
...
+
+The method that estimates these uncertainties has been presented at these ASWG calls:
+https://indico.cta-observatory.org/event/5847/contributions/47481/attachments/26575/39053/ctapipe_rethink_hillas_20240718.pdf
+https://indico.cta-observatory.org/event/5900/contributions/47899/attachments/26789/39379/Psi_uncertainty_short_summary_20241022.pdf
+
+The affected files are:
+image/hillas.py
+containers.py
+
+Caution:
```
maxnoe, line 13: This does not belong in a changelog but in the actual documentation of the method. Also: how is this information to be used?

docs/changes/2629.feature.rst
```diff
@@ -0,0 +1,15 @@
+This commit adds two new hillas-parameter variables:
+1. psi_uncertainty (uncertainty on the psi angle of the image)
+2. transverse_cog_uncertainty (uncertainty on the center of gravity along the transverse axis of the image)
+
+The method that estimates these uncertainties has been presented at these ASWG calls:
+https://indico.cta-observatory.org/event/5847/contributions/47481/attachments/26575/39053/ctapipe_rethink_hillas_20240718.pdf
+https://indico.cta-observatory.org/event/5900/contributions/47899/attachments/26789/39379/Psi_uncertainty_short_summary_20241022.pdf
+
+The affected files are:
+image/hillas.py
```
maxnoe, line 10: You do not need to list the affected files

docs/changes/2629.feature.rst
```diff
@@ -0,0 +1,15 @@
+This commit adds two new hillas-parameter variables:
```
maxnoe, line 1: The changelog describes a release, not a commit.

Just write something like:

Add two new fields in the Hillas parameters computations:

src/ctapipe/image/hillas.py
```diff
@@ -131,22 +133,42 @@ def hillas_parameters(geom, image):
...
+        lsq_cov = np.linalg.inv(X.T @ W @ X)
+        p = lsq_cov @ X.T @ W @ trans
+        # p[0] is the psi angle in the rotated frame, which should be zero.
+        # Now we add the non-zero residual psi angle in the rotated frame to psi uncertainty
+        # We also add additional uncertainty to account for elongation of the image (i.e. width / length)
+        psi_uncert = np.sqrt(lsq_cov[0, 0] + p[0] * p[0]) * (
+            1.0 + pow(np.tan(width / length * np.pi / 2.0), 2)
+        )
+        transverse_cog_uncert = np.sqrt(
+            lsq_cov[1, 1] * (1.0 + np.sin(p[0]) * np.sin(p[0]))
```
maxnoe, line 169: double computation of `np.sin(p[0])`

src/ctapipe/image/tests/test_hillas.py
```diff
@@ -253,6 +253,108 @@ def test_single_pixel():
...
+    image[0, 2] = 5
+    image[1, 0] = 5
+    image[1, 1] = 10
+    image[1, 2] = 5
+    image[2, 0] = 5
+    image[2, 1] = 5
+    image[2, 2] = 5
+    image = image.ravel()
+    hillas = hillas_parameters(geom, image)
+    assert hillas.psi_uncertainty.value > 1e10
```
maxnoe, line 285: This is a really imprecises test... why do we just check for `> 1e10`?

src/ctapipe/image/tests/test_hillas.py
```diff
@@ -253,6 +253,108 @@ def test_single_pixel():
...
+    image1 = image1.ravel()
+    image2 = image2.ravel()
+    hillas1 = hillas_parameters(geom, image1)
+    hillas2 = hillas_parameters(geom, image2)
+    assert (
+        abs(
+            hillas1.psi_uncertainty.value / hillas2.psi_uncertainty.value
+            - np.sqrt(intensity_scaling)
+        )
+        < 1e-4
```
maxnoe, line 304: Same here, why do we just check for < 1e-4, what if it is larger? What is the unit? Why do you subtract sqrt(2)?

Why do we test on the ratio of two different hillas parameters?

src/ctapipe/image/tests/test_hillas.py
```diff
@@ -253,6 +253,108 @@ def test_single_pixel():
...
+    )
+
+    # check psi uncertainty for a random image with the manual calculation
+    image = np.random.random((3, 3)) * 100.0
+    image = image.ravel()
+    hillas = hillas_parameters(geom, image)
+    # Now calculate the uncertainty manually
```
maxnoe, line 311: This is repeating the code of the function verbatim. I don't think this is a good test. Sure it can be used to make sure any new implementation gives the same result as the previous implementation.

Can we not define test cases with known results?

src/ctapipe/image/tests/test_hillas.py
```diff
@@ -253,6 +253,108 @@ def test_single_pixel():
...
+        lsq_cov = np.linalg.inv(X.T @ W @ X)
+        p = lsq_cov @ X.T @ W @ trans
+        # p[0] is the psi angle in the rotated frame, which should be zero.
+        # Now we add the non-zero residual psi angle in the rotated frame to psi uncertainty
+        # We also add additional uncertainty to account for elongation of the image (i.e. width / length)
+        psi_uncert = np.sqrt(lsq_cov[0, 0] + p[0] * p[0]) * (
+            1.0 + pow(np.tan(width / length * np.pi / 2.0), 2)
+        )
+
+    assert abs(hillas.psi_uncertainty.value - psi_uncert) / psi_uncert < 0.01
```
maxnoe, line 357: `np.isclose` or `u.isclose`

ruoyushang:
Hi @maxnoe @kosack ,

I modified the unit test for the psi uncertainty calculation.
There are three tests: the first and second tests have the same morphology but with different intensities.
The second test image with twice intensity gives smaller uncertainty by a factor of sqrt(2).
The third test image adds off-diagonal perturbation and gives larger uncertainty than the second test image.

I also modified the changelog and hillas.py to remove double calculation of np.sin(p[0]).
Please have a look at the new unit test and approve the request if it looks ok to you.

Best regards,
Ruo

src/ctapipe/image/tests/test_hillas.py
```diff
@@ -253,6 +253,69 @@ def test_single_pixel():
...
+    image[1, 0] = 0
+    image[1, 1] = 5
+    image[1, 2] = 0
+    image[2, 0] = 0
+    image[2, 1] = 0
+    image[2, 2] = 2
+    image = image.ravel()
+    hillas = hillas_parameters(geom, image)
+    assert np.isclose(
+        hillas.psi_uncertainty.to_value(u.deg), 20.25711711353489, rtol=1e-05
```
maxnoe, line 285: putting in this many digits when you have an rtol of 1e-05 is a bit overkill :)

ruoyushang:
Thanks a lot to @maxnoe for approving this request!
@kosack could you please approve this request again? I think your previous approval was canceled because I pushed new commits.

maxnoe:
Note: sonarqube is failing due to duplicated code, which is not an issue of this pull request and will be addressed by #2141

kosack (approved):

PR #2634, opened 2024-10-31
Chunk interpolation to select calibration data 

ctoennis:
I will need a method to select calibration data for the strar tracker. I made some slides to decribe how it is supposed to work:
https://docs.google.com/presentation/d/1oxIcYSQvGnU7IQYy3fGdcv0qXiLpvaXR9YtmnDesj4Y/edit?usp=sharing


src/ctapipe/monitoring/interpolation.py
```diff
@@ -4,15 +4,13 @@
...
 from astropy.time import Time
 from scipy.interpolate import interp1d
 
 from ctapipe.core import Component, traits
 
-__all__ = [
-    "Interpolator",
-    "PointingInterpolator",
-]
+__all__ = ["PointingInterpolator", "ChunkInterpolator"]
```
maxnoe, line 13: Please do not remove the base `Interpolator` class here, it will prevent the interface from being documented in the sphinx docs

ctoennis: Ok, i'll do it while adding the changelog

ctoennis:
@maxnoe @kosack Can you have another look if there is something else to be changed?

mexanick:
@kosack @maxnoe this PR is needed to complete the pointing calibration (for the variance calibration application), can we advance it?

src/ctapipe/monitoring/interpolation.py
```diff
@@ -186,3 +183,108 @@ def add_table(self, tel_id, input_table):
...
+                # If the closest preceding chunk has nan, check the next closest chunk
+                for i in range(preceding_index - 1, -1, -1):
+                    if start_time[i] <= mjd <= end_time[i]:
+                        value = values[i]
+                        if not np.isnan(value):
+                            return value
+
+                return np.nan
+
+            self._interpolators[tel_id][column] = interpolate_chunk
```
maxnoe, line 290: I think it's a bit weird here to store closures in a dict instead of just calling a method with `tel_id` and `column` that then checks if the necessary table exists.

ctoennis: ok, i'll change it to a method

mexanick: The current implementation follows the PointingInterpolator approach, but of course can be changed.

ctoennis: so, ``_check_interpolators`` in the parent class assumes the current implementation. If i change it to what you describe i will need to make changes to the parent class or forgo this check for the ``ChunkInterpolator``

ctoennis: Maybe keeping it like this is better.

ctoennis: Ok, i can make this as a separate class.

The way overlapping chunks are handled is that it tries to return the chunk with the earlies start time and use later chunks if that chunk has no valid data data (for instance there was a car flash during the period that causes the calibration data to be flagged as invalid). This way the amount of time for which valid data is provided is maximized.

ctoennis: Also i want to add that my intent was to use it in a similar way to the ``PointingInterpolator``. When event data are read from some file with the ``EventSource`` or some similar tool then the ``PointingInterpolator`` provides event-by-event data for the pointing. I would like this interpolator to do the same for calibration data. The Calibrators can then just look up event-by-event calibration data provided by this class and apply it to the events.

mexanick: I consider a _new_ data point as a point in (value, timestamp) domain :) Prior to the call, there's no unambiguous pair value-timestamp, so to me it is new.

I agree that it is a bit weird, however, this logically fits in the inheritance tree and in the future we may want to introduce a more complex interpolation (i.e. linear between the nearest measured points) and I would like to preserve the API.

ctoennis: @maxnoe could i suggest that i open an issue for refactoring the pointing interpolator to use a method as you describe. I can implement this change then on this new interpolator and the PointingInterpolator after this request is merged. Switching both over like that would not impact the API.

maxnoe: Wait: what is wrong with the pointing interpolator now?

mexanick: It also keeps the call results in a mutable shared state (interpolators dict).

maxnoe: > the call results

Not the call results. The interpolator instances.

That was not the point... the point was using a closure to store data on a class, instead of the actual datastructure that can be e.g. investigated in a test or debugger.

maxnoe: Is the point that the `Interpolator` baseclass already assumes this? That's the case because it was designed as the base class to share code for things using `interp1d` to interpolate tables.

You have something completely different here now. So it should not use the same base class. 

If we need a common base class to ensure a common API for anything that takes some monitoring tables and returns the correct value for a given time, we should have a purely abstract interface definition and then the current `Interpolator` only as  a baseclass for the things that are based on `interp1d`.

So:

```
class MonitoringInterpolator:
   @abstractmethod 
   def __call__(self, timestamp):
        ...

class LinearMonitoringInterpolator(MonitoringInterpolator):
     ...

class PointingInterpolator(LinearMonitoringInterpolator):
   ...

class TimeRangeFinder(MonitoringInterpolator)
   ...
```

ctoennis: @maxnoe So, i can move the closure and make it just a method of ChunkInterpolator. If i do that i don't need to do anything that impacts the API. Would that be okay?

maxnoe: The `Interpolator` class has traitlets that are directly related to `interp1d` for example. That really doesn't apply here, so I don't think it should be in a shared baseclass.

mexanick: ok. @ctoennis please provide an interface sketch, maintaining one abstract interface-only class, so we can check it before the implementation.

ctoennis: ok, I'm doing that

ctoennis: Here is a sketch for the parent classes:

``
class MonitoringInterpolator(Component, metaclass=ABCMeta):
    """
    MonitoringInterpolator parent class.

    Parameters
    ----------
    h5file : None | tables.File
        An open hdf5 file with read access.
    """

    telescope_data_group = None
    required_columns = set()
    expected_units = {}

    def __init__(self, h5file: None | tables.File = None, **kwargs: Any) -> None:
        super().__init__(**kwargs)
 
        ...

    @abstractmethod
    def add_table(self, tel_id: int, input_table: Table) -> None:
        """
        Parameters
        ----------
        tel_id : int
            Telescope id.
        input_table : astropy.table.Table
            Table of pointing values, expected columns
            are always ``time`` as ``Time`` column and
            other columns for the data that is to be interpolated.
        """
        pass

    def _check_tables(self, input_table: Table) -> None:
        ...

    def _check_interpolators(self, tel_id: int) -> None:
        ...

    def _read_parameter_table(self, tel_id: int) -> None:
        ...

class LinearInterpolator(MonitoringInterpolator):
    """
    LinearInterpolator parent class.

    Parameters
    ----------
    h5file : None | tables.File
        An open hdf5 file with read access.
    """

    bounds_error = traits.Bool(
        default_value=True,
        help="If true, raises an exception when trying to extrapolate out of the given table",
    ).tag(config=True)

    extrapolate = traits.Bool(
        help="If bounds_error is False, this flag will specify whether values outside"
        "the available values are filled with nan (False) or extrapolated (True).",
        default_value=False,
    ).tag(config=True)

    def __init__(self, h5file: None | tables.File = None, **kwargs: Any) -> None:
        super().__init__(**kwargs)

        self.interp_options: dict[str, Any] = dict(assume_sorted=True, copy=False)
        if self.bounds_error:
            self.interp_options["bounds_error"] = True
        elif self.extrapolate:
            self.interp_options["bounds_error"] = False
            self.interp_options["fill_value"] = "extrapolate"
        else:
            self.interp_options["bounds_error"] = False
            self.interp_options["fill_value"] = np.nan

        self._interpolators = {}

``

maxnoe: :+1:  Looks good

I am not sure if putting the private methods you listed here to the base class will work out in the end, but you'll see how much code can be shared between the chunk interpolator and the linear interpolator.

mexanick: Shall we take an opportunity and stop using `interp1d`? It is legacy, and`scipy` suggests this: https://docs.scipy.org/doc/scipy/tutorial/interpolate/1D.html#tutorial-interpolate-1dsection

Also, if we call a base class LinearInterpolator, I guess we should enforce linearity (in the example above, e.g. cubic spline can be applied)

maxnoe: There is no replacement for linear 1d interpolation for interp1d. It is legacy, but not deprecated.

See https://github.com/scipy/scipy/issues/21258

ctoennis:
I am a bit stuck here with one of the tests. test_hdf5 is failing in pytest begause the data from the file is not loaded correctly. When i look in the test what x and y values the interpolators have i get some wrong values. However if i try to do the same outside of pytest it works. I used this code to test by myself:

```
import astropy.units as u
import numpy as np
import tables
from astropy.time import Time

from functools import partial
from ctapipe.core import Component, traits

from ctapipe.monitoring.interpolation import PointingInterpolator

from ctapipe.io import write_table

t0 = Time("2022-01-01T00:00:00")

table = Table(
    {"time": t0 + np.arange(0.0, 10.1, 2.0) * u.s, "azimuth": np.radians(np.linspace(0.0, 10.0, 6)) * u.rad, "altitude": np.radians(>)

path = "pointing.h5"

write_table(table, path, "/dl0/monitoring/telescope/pointing/tel_001")

with tables.open_file(path) as h5file:
    interpolator = PointingInterpolator(h5file)
    t = t0 + 1 * u.s
    alt, az = interpolator(tel_id=1, time=t)
    print(interpolator._interpolators[1]["alt"].y,interpolator._interpolators[1]["alt"].x)
    print(alt,az)
```

Has anyone an idea what is wrong here?

src/ctapipe/monitoring/tests/test_interpolator.py
```diff
@@ -57,7 +173,7 @@ def test_invalid_input():
     wrong_unit = Table(
         {
             "time": Time(1.7e9 + np.arange(3), format="unix"),
-            "azimuth": [1, 2, 3] * u.deg,
+            "azimuth": np.radians([1, 2, 3]) * u.rad,
```
maxnoe, line 176: Why did you change these tests here?

maxnoe: The test is explicitly to test the unit conversion, so please don't change to rad

src/ctapipe/monitoring/tests/test_interpolator.py
```diff
@@ -72,8 +188,8 @@ def test_hdf5(tmp_path):
     table = Table(
         {
             "time": t0 + np.arange(0.0, 10.1, 2.0) * u.s,
-            "azimuth": np.linspace(0.0, 10.0, 6) * u.deg,
```
maxnoe, line 75: same here

src/ctapipe/monitoring/tests/test_interpolator.py
```diff
@@ -82,8 +198,8 @@ def test_hdf5(tmp_path):
     with tables.open_file(path) as h5file:
         interpolator = PointingInterpolator(h5file)
         alt, az = interpolator(tel_id=1, time=t0 + 1 * u.s)
-        assert u.isclose(alt, 69 * u.deg)
-        assert u.isclose(az, 1 * u.deg)
+        assert u.isclose(alt, np.radians(69) * u.rad)
```
maxnoe, line 201: same here

src/ctapipe/monitoring/tests/test_interpolator.py
ctoennis: The pointing_interpolator is supposed to use radians for input. Line:
```
157         expected_units = {"azimuth": u.rad, "altitude": u.rad}
```

mexanick: right, and this test checks that correct exception is raised in case the units are not matching. Please revert back

ctoennis:
@mexanick Thanks, i got the issue fixed.

ctoennis:
@maxnoe Hi Max, could you give this another look?

src/ctapipe/monitoring/interpolation.py
```diff
@@ -184,5 +216,135 @@ def add_table(self, tel_id, input_table):
...
+
+class ChunkInterpolator(MonitoringInterpolator):
+    """
+    Simple interpolator for overlapping chunks of data.
+    """
+
+    required_columns = frozenset(["start_time", "end_time"])
+    start_time = {}
+    end_time = {}
+    values = {}
```
mexanick, line 235: Why do you keep these attributes as class attributes? Is their sharing between the instances is expected? I suggest to define them as object attributes instead.

ctoennis: ok, i'll change it

src/ctapipe/monitoring/interpolation.py
```diff
@@ -1,85 +1,81 @@
...
-
-    telescope_data_group = None
-    required_columns = set()
-    expected_units = {}
-
-    def __init__(self, h5file=None, **kwargs):
+    def __init__(self, h5file: None | tables.File = None, **kwargs: Any) -> None:
         super().__init__(**kwargs)
 
+        self.telescope_data_group = None
```
maxnoe, line 35: Why did you change the code here? Why move the class variables to instance variables? 

I don't think the code here should be changed at all. The only thing was to add one "pure" abstract class in the hierarchy. No changes other than changing to that new class as base class should be required for the existing code.

maxnoe (changes requested):

src/ctapipe/monitoring/interpolation.py
```diff
@@ -116,32 +112,69 @@ def _read_parameter_table(self, tel_id):
...
+            self.interp_options["bounds_error"] = True
+        elif self.extrapolate:
+            self.interp_options["bounds_error"] = False
+            self.interp_options["fill_value"] = "extrapolate"
+        else:
+            self.interp_options["bounds_error"] = False
+            self.interp_options["fill_value"] = np.nan
+
+
+class PointingInterpolator(LinearInterpolator):
```
maxnoe, line 150: Same here: the only change for `PointingInterpolator` should be the base class

src/ctapipe/monitoring/interpolation.py
mexanick: At least the interpolators shall be the instance attributes, otherwise they will be shared between the instances and won't be updated if already exist. @ctoennis bumped into this with the tests before. 
Also, @maxnoe do you want to forbid parallel existence of different interpolators?

ctoennis: telescope_data_group can be a class variable. the rest needed to be instance variables. I'll make some changes

mexanick: the `telescope_data_group` can be a class variable of some of the child classes, but not of the base class, because in general it is not shared between different interpolators. I.e. if you make it a class variable of the base class, two instances of interpolators could not coexist if the have different `telescope_data_group`.

maxnoe: > the rest needed to be instance variables.

Why? It is a class variable now and it works.

maxnoe: > At least the interpolators shall be the instance attributes

Yes of course, but they are also instance attributes now

src/ctapipe/monitoring/interpolation.py
```diff
@@ -184,5 +211,140 @@ def add_table(self, tel_id, input_table):
...
+            mjd, alt, kind="linear", **self.interp_options
+        )
+
+
+class ChunkInterpolator(MonitoringInterpolator):
+    """
+    Simple interpolator for overlapping chunks of data.
+    """
+
+    required_columns = frozenset(["start_time", "end_time"])
```
mexanick, line 227: This is an immutable type that you "mutate" further down in the code. Please don't do this, if you want to extend it with the value column, particular for an object, just make a copy and to the object attribute and modify that one.

src/ctapipe/monitoring/interpolation.py
```diff
@@ -184,5 +211,140 @@ def add_table(self, tel_id, input_table):
...
+    expected_units = {}
```
mexanick, line 228: Expected units can be different for different objects. Please ensure you can create two different ChunkInterpolators objects with different interpolation column and different units to it.

src/ctapipe/monitoring/interpolation.py
```diff
@@ -1,85 +1,78 @@
...
-
-    telescope_data_group = None
-    required_columns = set()
-    expected_units = {}
-
-    def __init__(self, h5file=None, **kwargs):
+    def __init__(self, h5file: None | tables.File = None, **kwargs: Any) -> None:
         super().__init__(**kwargs)
 
+        self._interpolators = {}
```
maxnoe, line 35: As discussed before, the base class should be rather pure. An interface shouldn't prescribe private data layout.

src/ctapipe/monitoring/interpolation.py
```diff
@@ -116,32 +107,72 @@ def _read_parameter_table(self, tel_id):
...
 
     telescope_data_group = "/dl0/monitoring/telescope/pointing"
     required_columns = frozenset(["time", "azimuth", "altitude"])
     expected_units = {"azimuth": u.rad, "altitude": u.rad}
 
-    def __call__(self, tel_id, time):
+    def __init__(self, h5file: None | tables.File = None, **kwargs: Any) -> None:
+        super().__init__(h5file=h5file, **kwargs)
+
+        self._interpolators = {}
```
maxnoe, line 157: Why is this now here and not in the `LinearInterpolator`?

maxnoe:
This looks good now. A remaining question would be if you want to add specific `ChunkInterpolators` (like `PointingInterpolator` for `LinearInterpolator`)?

I.e. `CalibrationInterpolator`, `FlatFieldInterpolator`, `PedestalInterpolator` etc.

maxnoe:
Can also be done in a follow-up PR of course.

mexanick:
> This looks good now. A remaining question would be if you want to add specific `ChunkInterpolators` (like `PointingInterpolator` for `LinearInterpolator`)?
> 
> I.e. `CalibrationInterpolator`, `FlatFieldInterpolator`, `PedestalInterpolator` etc.

I'd consider having a factory then. I think, just "CalibrationInterpolator" won't make much sense, but a specific ones like "FFInterpolator" may. I'd address it in another PR.

maxnoe:
Yes, factory makes sense for this

mexanick (approved):

src/ctapipe/monitoring/interpolation.py
```diff
@@ -184,5 +209,139 @@ def add_table(self, tel_id, input_table):
...
+class ChunkInterpolator(MonitoringInterpolator):
+    """
+    Simple interpolator for overlapping chunks of data.
+    """
+
+    def __init__(self, h5file: None | tables.File = None, **kwargs: Any) -> None:
+        super().__init__(**kwargs)
+        self._interpolators = {}
+        self.required_columns = ["start_time", "end_time"]
+        self.expected_units = {}
```
maxnoe, line 229: Why are these instance variables? And why are they empty?

This is not in line with how the other class works.

It seems this class variable is unused even.

mexanick: Here `required_columns` shall become class attributes, but the interpolators and expected units shall remain instance variable, to allow creation of multiple instances. We have to see in the future PR, whether we want further specialization (e.g. a factory of `VarNameInterpolator`s), that may lead to change this (they will basically become singletons).

mexanick: expected units are unused because the quantity is dimensionless, we may want to actually enforce this through `u.dimensioneless_unscaled`

ctoennis: We had implemented that for collumns that are supposed to not have a unit we set the expected unit to None and check if the actual unit is equivalent. I put it like that.

src/ctapipe/monitoring/interpolation.py
```diff
@@ -184,5 +209,139 @@ def add_table(self, tel_id, input_table):
...
+
+class ChunkInterpolator(MonitoringInterpolator):
+    """
+    Simple interpolator for overlapping chunks of data.
+    """
+
+    def __init__(self, h5file: None | tables.File = None, **kwargs: Any) -> None:
+        super().__init__(**kwargs)
+        self._interpolators = {}
+        self.required_columns = ["start_time", "end_time"]
```
mexanick: required columns as start and stop time shall be the class attributes and shall be frozen, they are *mandatory*. You can copy them to an instance and extend with a value column(s) when you engage a `__call__`.

ctoennis: as the functions that use required_columns are in the parent class and always look to that name it makes more sense to have it as a modifiable instance variable

maxnoe: No. The original design idea was to have the required_columns frozen per "final" class. I.e. `MonitoringInterpolator` requires `altitude` and `azimuth` columns.

This is due to the configuration system in ctapipe, which works on class name basis, not instances.

src/ctapipe/monitoring/interpolation.py
```diff
@@ -184,5 +209,140 @@ def add_table(self, tel_id, input_table):
...
+            ``end_time`` as ``validity end Time`` and the specified columns
+            for the data of the chunks.
+        columns : list of str
+            Names of the columns to interpolate.
+        """
+
+        self.required_columns.update(columns)
+        self.required_columns = set(self.required_columns)
+        for col in columns:
+            self.expected_units[col] = None
```
ctoennis, line 293: Here i set the unit of the new columns to None, which is then in the next line enforced by _check_tables. This way we ensure the values have no unit.

src/ctapipe/monitoring/interpolation.py
```diff
@@ -184,5 +209,136 @@ def add_table(self, tel_id, input_table):
...
+
+        if tel_id not in self._interpolators:
+            self._interpolators[tel_id] = {}
+            self.values[tel_id] = {}
+            self.start_time[tel_id] = {}
+            self.end_time[tel_id] = {}
+
+        for column in self.columns:
+            self.values[tel_id][column] = input_table[column]
+            self.start_time[tel_id][column] = input_table["start_time"].to_value("mjd")
```
maxnoe, line 295: Why store `start_time` and `end_time` per column?

src/ctapipe/monitoring/interpolation.py
```diff
@@ -184,5 +209,136 @@ def add_table(self, tel_id, input_table):
...
+        self._check_interpolators(tel_id)
+
+        result = {}
+        mjd = time.to_value("mjd")
+        for column in self.columns:
+            if column not in self._interpolators[tel_id]:
+                raise ValueError(
+                    f"Column '{column}' not found in interpolators for tel_id {tel_id}"
+                )
+            result[column] = self._interpolators[tel_id][column](mjd)
```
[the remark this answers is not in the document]
ctoennis: `read_table` checks if `_interpolators` has already been set up for the given tel_id. `_check_interpolators` in `MonitoringInterpolator` also does that check and adds data from hdf5file if the interpolator is not set. I can move column to be an argument of `_interpolate_chunk` though.

src/ctapipe/monitoring/interpolation.py
```diff
@@ -184,5 +209,128 @@ def add_table(self, tel_id, input_table):
...
+            if start_time[i] <= mjd <= end_time[i]:
+                value = values[i]
+                if not np.isnan(value):
+                    return value
+
+        return np.nan
+
+
+class FlatFieldInterpolator(ChunkInterpolator):
+    required_columns = frozenset(["start_time", "end_time", "relative_gain"])
```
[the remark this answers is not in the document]
ctoennis: Ah, yes. I will add this right away. Also the Tests should use data that have the actual shape that the calibration data will have.

mexanick (approved):

src/ctapipe/monitoring/__init__.py
```diff
@@ -2,7 +2,12 @@
 Module for handling monitoring data.
 """
 from .aggregator import PlainAggregator, SigmaClippingAggregator, StatisticsAggregator
-from .interpolation import Interpolator, PointingInterpolator
+from .interpolation import (
+    ChunkInterpolator,
```
mexanick, line 6: Add final child classes here

mexanick: @ctoennis please add the child classes, your docs are failing because of this omission I believe:
```
Handler <function process_automodsumm_generation at 0x7f46067e34c0> for event 'builder-inited' threw an exception (exception: module 'ctapipe.monitoring' has no attribute 'FlatFieldInterpolator')
```

kosack (dismissed):

mexanick:
@maxnoe can you please take a look? I believe, all your comments have been addressed.

mexanick (approved):

src/ctapipe/monitoring/interpolation.py
```diff
@@ -184,5 +209,133 @@ def add_table(self, tel_id, input_table):
...
+
+        input_table = input_table.copy()
+        input_table.sort("start_time")
+
+        if tel_id not in self._interpolators:
+            self._interpolators[tel_id] = {}
+            self.values[tel_id] = {}
+            self.start_time[tel_id] = input_table["start_time"].to_value("mjd")
+            self.end_time[tel_id] = input_table["end_time"].to_value("mjd")
+            self._interpolators[tel_id] = partial(self._interpolate_chunk, tel_id)
```
maxnoe, line 288: Why are we still storing a lambda / partial here in `_interpolators`? Why not just call `self._interpolate_chunk` directly in `__call__`?

src/ctapipe/monitoring/interpolation.py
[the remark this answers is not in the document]
maxnoe: That's because you didn't move `_check_interpolators` to the `LinearInterpolator` but kept it in `MonitoringInterpolator`. The `MonitoringInterpolator` still contains more than just the abstract interface.

kosack (dismissed):

src/ctapipe/monitoring/interpolation.py
```diff
@@ -116,32 +99,74 @@ def _read_parameter_table(self, tel_id):
...
+
+    extrapolate = traits.Bool(
+        help="If bounds_error is False, this flag will specify whether values outside"
+        "the available values are filled with nan (False) or extrapolated (True).",
+        default_value=False,
+    ).tag(config=True)
+
+    def __init__(self, h5file: None | tables.File = None, **kwargs: Any) -> None:
+        super().__init__(h5file, **kwargs)
+        self._interpolators = {}
```
maxnoe, line 125: Still here but not used anymore?

src/ctapipe/monitoring/interpolation.py
```diff
@@ -184,5 +208,131 @@ def add_table(self, tel_id, input_table):
...
+
+
+class ChunkInterpolator(MonitoringInterpolator):
+    """
+    Simple interpolator for overlapping chunks of data.
+    """
+
+    def __init__(self, h5file: None | tables.File = None, **kwargs: Any) -> None:
+        super().__init__(**kwargs)
+        self._interpolators = {}
```
maxnoe, line 226: Still here but unused?

mexanick: You're right! I'm wondering why ruff didn't complain? Probably, because it is a "private" attribute...
Fixed now, I also updated the docstring in abstract add_table.

maxnoe (approved):

kosack (approved):

PR #2639, opened 2024-11-07
Support astropy 7.0

maxnoe:


Tobychev:
The changelog says "Add support" but actually nothing was added (because we were already compatible?) which I find a bit surprising given the description. Can't think of a another way to word it that's worth using though...

Tobychev (approved):

maxnoe:
The requirements excluded astropy 7 before

kosack (approved):

PR #2642, opened 2024-11-08
Fix changelog checks

mexanick:
Use bash script instead of file-existence-action, as that action doesn't seem to have an option "check if at least one exists"

Tested manually with local run of github-actions.
Currently add "no-changelog-needed" label, but can remove and add changelog for automated test.
Not sure if this propagate to an actual changelog though.

### DIFF
--- a/.github/workflows/changelog.yml
+++ b/.github/workflows/changelog.yml
@@ -20,16 +20,5 @@ jobs:
       - name: Check for at least one changelog fragment
         if: ${{ !contains(github.event.pull_request.labels.*.name, 'no-changelog-needed') }}
         run: |
-          CHANGE_TYPES=("feature" "bugfix" "api" "datamodel" "optimization" "maintenance")
-          FRAGMENT_FOUND=false
-          for CHANGE_TYPE in "${CHANGE_TYPES[@]}"; do
-            FILE="${{ env.FRAGMENT_NAME_PREFIX }}.${CHANGE_TYPE}.rst"
-            if [ -f "$FILE" ]; then
-              FRAGMENT_FOUND=true
-              break
-            fi
-          done
-          if [ "$FRAGMENT_FOUND" = false ]; then
-            echo "No changelog fragment found. Please add a changelog fragment."
-            exit 1
-          fi
+          find docs/changes -type f | \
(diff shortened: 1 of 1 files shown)


.github/workflows/changelog.yml
```diff
@@ -6,19 +6,30 @@ on:
...
 
-      - name: Check for news fragment
-        if: ${{ ! contains( github.event.pull_request.labels.*.name, 'no-changelog-needed')}}
-        uses: andstor/file-existence-action@v3
-        with:
-          files: ${{ env.FRAGMENT_NAME }}
-          fail: true
+      - name: Check for at least one changelog fragment
+        if: ${{ !contains(github.event.pull_request.labels.*.name, 'no-changelog-needed') }}
+        run: |
```
maxnoe, line 22: Couldn't this just be:

```
find docs/changes -type f | grep -P 'docs/changes/2411\.(feature|bugfix|api|datamodel|optimization|maintenance)\.rst'
```
?

mexanick: automatically did in pure bash, but this is indeed much neater.

maxnoe (approved):

LukasBeiske (approved):

.github/workflows/changelog.yml
```diff
@@ -6,19 +6,19 @@ on:
...
           fetch-depth: 0
 
-      - name: Check for news fragment
-        if: ${{ ! contains( github.event.pull_request.labels.*.name, 'no-changelog-needed')}}
-        uses: andstor/file-existence-action@v3
-        with:
-          files: ${{ env.FRAGMENT_NAME }}
-          fail: true
+      - name: Check for at least one changelog fragment
+        if: ${{ !contains(github.event.pull_request.labels.*.name, 'no-changelog-needed') }}
```
Tobychev, line 21: Are you absolutely sure that no space after "!" works as you want? The evaluation rules for bash are very fickle and specific.

mexanick: yes. This PR has the "no-changelog-needed" label and passes the changelog check despite it has no changelog entry.

maxnoe: This is not bash, it is githubs expression language

PR #2643, opened 2024-11-10
PSF model

ctoennis:
I am adding PSF models to ctapipe.image.psf_model. I made a parent class called PSFModel and a PSF model based on pure coma abberation called ComaModel. 

This will be used in the pointing calculation using star tracking and PSF fitting using stars.


kosack:
Quick question: are these models applicable for both gamma-ray and optical PSF? If not, maybe call it OpticalPSFModel to be more clear where it should be used.

maxnoe:
I think this is purely optical, not IRF. Also, I think the appropriate place would be in `ctapipe.instrument.optics`

ctoennis:
> Quick question: are these models applicable for both gamma-ray and optical PSF? If not, maybe call it OpticalPSFModel to be more clear where it should be used.

The model accounts for coma abbarations and the default parameters i have would be for photons in the optical range. With different parameters this should work with gamma photons, but i am not sure. I could call the ComeModel OpticalComaModel and keep the parent class as is. Then later models specifically for gamma photons can be added.

ctoennis:
> I think this is purely optical, not IRF. Also, I think the appropriate place would be in `ctapipe.instrument.optics`

Makes sense. I will move it to optics.

maxnoe:
> Side note: we should check with what is used in SimPipe's instrument model, to ensure what is output here is compatible with what the simulations expect.

I don't think there is a model like this in SimPipe, but @orelgueta or @GernotMaier can confirm. Rather coma is naturally resulting from the definition of the mirror facets.

kosack:
> I think the appropriate place would be in ctapipe.instrument.optics

The `instrument` module is where we define the `InstrumentDescription` model (which currently is collected under `SubarrayDescription` class here in ctapipe). Would be useful then to integrate it with the rest of the model, i.e. add a place for it in `OpticsDescription`, and add serialization so that it can be written and read.  A  collection of these could be serialized to a new table (one row per telescope), as we do with other OpticsDescription parts (`OpticsDescription.to_table(kind="optical_psf")`) .

src/ctapipe/instrument/optics.py
```diff
@@ -263,3 +270,176 @@ def __repr__(self):
...
+
+
+class PSFModel:
+    def __init__(self, **kwargs):
+        """
+        Base component to describe image distortion due to the optics of the different cameras.
+        """
+
+    @classmethod
+    def from_name(cls, name, **kwargs):
```
kosack, line 282: could be confusing here... In the instrument model, we use `from_name` to load a named model from a file, not to get a sub-class...  this is a general API conflict with the `Component` classes, however, which is a larger issue, but if this is in instrument, it should at least be consistent.

ctoennis:
> > I think the appropriate place would be in ctapipe.instrument.optics
> 
> The `instrument` module is where we define the `InstrumentDescription` model (which currently is collected under `SubarrayDescription` class here in ctapipe). Would be useful then to integrate it with the rest of the model, i.e. add a place for it in `OpticsDescription`, and add serialization so that it can be written and read. A collection of these could be serialized to a new table (one row per telescope), as we do with other OpticsDescription parts (`SubarrayDescription.to_table(kind="optical_psf")`, or else just in OpticsDescription... ) . That could be in another PR if you like, but in that case we should open an issue to make sure this new part of the model is properly integrated
> 
> Edit: to be more clear, I would expect something like this to work:
> 
> ```python
> psf : Optional[PSFModel]  = subarray.tel[tel_id].optics.psf
> 
> if psf: 
>      plot_psf(psf)  # some general function that plots any PSFModel
> ```

if it is okay then i will open an issue for this

src/ctapipe/instrument/optics.py
```diff
@@ -263,3 +270,176 @@ def __repr__(self):
...
+        Returns
+        -------
+        Instance
+            Instance of subclass to this class
+        """
+        requested_subclass = cls.non_abstract_subclasses()[name]
+        return requested_subclass(**kwargs)
+
+    @classmethod
+    def non_abstract_subclasses(cls):
```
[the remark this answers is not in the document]
ctoennis: i am removing it

src/ctapipe/instrument/optics.py
```diff
@@ -263,3 +269,161 @@ def __repr__(self):
...
+        radial_scale_params=[0.01409259, 0.02947208, 0.06000271, -0.02969355],
+        az_scale_params=[0.24271557, 7.5511501, 0.02037972],
+    ):
+        """
+        PSF model, describing purely the effect of coma aberration on the PSF
+
+        Parameters
+        ----------
+        asymmetry_params: list
+            Parameters describing the dependency of the asymmetry of the psf on the distance to the center of the camera
```
maxnoe, line 331: These parameters need better explanations and possibly units.

I suggest to add the relevant formulas to the docstring, see 

https://www.sphinx-doc.org/en/master/usage/restructuredtext/directives.html#directive-math

src/ctapipe/instrument/optics.py
```diff
@@ -263,3 +269,161 @@ def __repr__(self):
...
+        radial_scale_params : list
+            Parameters describing the dependency of the radial scale on the distance to the center of the camera
+        az_scale_params : list
+            Parameters describing the dependency of the azimuthal scale scale on the distance to the center of the camera
```
maxnoe, line 335: double scale

src/ctapipe/instrument/optics.py
```diff
@@ -263,3 +269,161 @@ def __repr__(self):
...
+        """
+
+        self.asymmetry_params = asymmetry_params
```
maxnoe, line 338: some validation should happen here I think

src/ctapipe/instrument/optics.py
```diff
@@ -263,3 +269,161 @@ def __repr__(self):
...
+        self.asymmetry_params = model_params["asymmetry_params"]
+        self.radial_scale_params = model_params["radial_scale_params"]
+        self.az_scale_params = model_params["az_scale_params"]
+        k = self.k_func(self.radial_pdf_params[1])
+        sr = self.sr_func(self.radial_pdf_params[1])
+        sf = self.sf_func(self.radial_pdf_params[1])
+        self.radial_pdf_params = (k, self.radial_pdf_params[1], sr)
+        self.azimuthal_pdf_params = (self.azimuthal_pdf_params[0], sf)
+
+    def update_location(self, r, f):
```
maxnoe, line 413: Why do we need an "update" function? Do we have to pre-compute expensive operations? 

Why is evaluating the PSF not enough?

src/ctapipe/instrument/optics.py
```diff
@@ -263,3 +269,161 @@ def __repr__(self):
...
+
+    def k_func(self, x):
+        return (
+            1
+            - self.asymmetry_params[0] * np.tanh(self.asymmetry_params[1] * x)
+            - self.asymmetry_params[2] * x
+        )
+
+    def sr_func(self, x):
+        return (
```
maxnoe, line 352: This looks like it should use 

https://numpy.org/doc/stable/reference/generated/numpy.polyval.html

src/ctapipe/instrument/optics.py
ctoennis: I'll add units where possible,  but for the most part they won't have them. I will add the documentation when I arrive home.

src/ctapipe/instrument/optics.py
```diff
@@ -263,3 +269,161 @@ def __repr__(self):
...
+    """
+
+    def __init__(
+        self,
+        asymmetry_params=[0.49244797, 9.23573115, 0.15216096],
+        radial_scale_params=[0.01409259, 0.02947208, 0.06000271, -0.02969355],
+        az_scale_params=[0.24271557, 7.5511501, 0.02037972],
+    ):
+        """
+        PSF model, describing purely the effect of coma aberration on the PSF
```
maxnoe, line 326: Is there a citable reference for the math here?

maxnoe (changes requested):

src/ctapipe/instrument/optics.py
```diff
@@ -263,3 +269,161 @@ def __repr__(self):
...
+
+
+class ComaModel(PSFModel):
+    """
+    PSF model, describing pure coma aberrations PSF effect
+    """
+
+    def __init__(
+        self,
+        asymmetry_params=[0.49244797, 9.23573115, 0.15216096],
```
maxnoe, line 321: Where do these magic numbers come from?

How are you expecting we determine these numbers? Is there a tool to compute them from a series of PSF images or pedestal images?

How do you expect them to be fed into ctapipe for the telescopes?

ctoennis: I got these numbers from https://gitlab.cern.ch/cta-unige/startracker (line [162 https://gitlab.cern.ch/cta-unige/startracker/-/blob/master/startracker/analysis/image.py?ref_type=heads]). @mexanick generated these parameters based on some fits. I can remove these as default.

src/ctapipe/instrument/optics.py
ctoennis: Here it is: https://www.aanda.org/articles/aa/full_html/2023/11/aa47128-23/aa47128-23.html
I will add this to the documentation

maxnoe: The bib file is here:
https://github.com/cta-observatory/ctapipe/blob/main/docs/references.bib

you can then cite using:
```
:cite:p:`<bibtex key>`
```

ctoennis: awesome. will do

maxnoe: Which makes me realize we didn't add a reference for the hipparcos catalog

src/ctapipe/instrument/optics.py
ctoennis: I have right now 2 use cases for this PSF model: processing variance images (e.g. finding stars in variance images) and fitting the PSF parameters based on stars in variance images. for the former i need to evaluate it the PSF at different locations for each possible star location, in the latter i also need to vary the psf parameters for fitting. It made sense to me to have these 2 functions available to modify these parameters. 

Also just to clarify, update location is where the "point" in point spread function is, as in if you want the PSF for a point source at r = 1m and f = 1.0rad these coordinates will need to be put in this function.

ctoennis: i will try and make the documentation more clear.

maxnoe: The question is why this point needs to be state on the class and cannot just be evaluated.

src/ctapipe/instrument/optics.py
ctoennis: i can add that.

src/ctapipe/instrument/optics.py
mexanick: I believe this is a remnant of the debugging code I used during the development... :)
@ctoennis, we don't need this step to be factored out like this anymore.

src/ctapipe/instrument/optics.py
mexanick: this is an ansatz recursive parametrization (parametrization of the parameters) :)
Done empirically by fitting the best-fit values of the PSF parameters for simulations of stars with different offset. The formulas should indeed be provided as a part of the documentation.

src/ctapipe/instrument/optics.py
ctoennis: ok, i'll put it all together then. I am having a little issue with the docustrings i'm sorting out now. I'll get to it after that.

src/ctapipe/instrument/optics.py
maxnoe: > So, when i will use this for extracting stars these parameters will not need to be altered

*for LST-1*, right?

ctoennis: Ah, yes, they need to be configured per telescope.

maxnoe: Ok, then I think we should discuss a bit about API.

I think, it's natural then to make this a `TelescopeComponent` and if there is code that can take a list of images and fit the result, we could have that as a 
```
@classmethod
def fit(self, ...) -> Parameters: 
```
here

ctoennis: that makes sense. For the fitting though my idea was to fit the PSF based on images with multiple stars. I have made some draft of a function that would generate a picture based on star locations, magnitudes and psf parameters that would be used as a model for fitting: 

`def get_star_pdf(r0, f0, geometry, psf, n_pdf_bins, pdf_bin_size, focal_length):
    image = np.zeros(len(geometry))

    dr = pdf_bin_size * np.rad2deg(np.arctan(1 / focal_length.to_value(u.m))) / 3600.0
    r = np.linspace(
        r0 - dr * n_pdf_bins / 2.0,
        r0 + dr * n_pdf_bins / 2.0,
        n_pdf_bins,
    )
    df = np.deg2rad(pdf_bin_size / 3600.0) * 100
    f = np.linspace(
        f0 - df * n_pdf_bins / 2.0,
        f0 + df * n_pdf_bins / 2.0,
        n_pdf_bins,
    )

    for r_ in r:
        for f_ in f:
            val = psf.pdf(r_, f_, r0, f0) * dr * df
            x, y, _ = r_ * s2c(f_, 0)
            pixelN = geometry.position_to_pix_index(x * u.m, y * u.m)
            if pixelN > -1 and pixelN < len(image):
                image[pixelN] += val

    return image

def StarImageGenerator(
    radius,
    phi,
    magnitude,
    n_pdf_bins,
    pdf_bin_size,
    psf_model,
    camera_geometry,
    focal_length,
):
    """
    :param list stars: list of star containers, stars to be placed in image
    :param dict psf_model_pars: psf model parameters
    """

    image = np.zeros(len(camera_geometry))
    for r, p, m in zip(radius, phi, magnitude):
        image += np.power(10,-0.4*m) * get_star_pdf(
            r, p, camera_geometry, psf_model, n_pdf_bins, pdf_bin_size, focal_length
        )

    return image`

ctoennis: I made the PSFModel into a `TelescopeComponent`. I will need to see how i will handle the fitting in detail, but i can work that out.

src/ctapipe/instrument/optics.py
```diff
@@ -263,3 +270,141 @@ def __repr__(self):
 
     def __str__(self):
         return self.name
+
+
+class PSFModel(TelescopeComponent):
+    def __init__(self, subarray, **kwargs):
```
maxnoe, line 276: not needed, since identical with baseclass.

Docstring should be at the class definition, not `__init__`.

ctoennis: OK, I'll fix it.

src/ctapipe/instrument/optics.py
```diff
@@ -263,3 +270,141 @@ def __repr__(self):
...
+        """
+        Base component to describe image distortion due to the optics of the different cameras.
+        """
+        super().__init__(subarray=subarray, **kwargs)
+
+    @abstractmethod
+    def pdf(self, *args):
```
maxnoe, line 283: This needs a docstring so that subclass implementers know what is expected.

Also the signature should be the same for all classes, so it should be explicit here in the interface definition.

orelgueta:
> > Side note: we should check with what is used in SimPipe's instrument model, to ensure what is output here is compatible with what the simulations expect.
> 
> I don't think there is a model like this in SimPipe, but @orelgueta or @GernotMaier can confirm. Rather coma is naturally resulting from the definition of the mirror facets.

You are right, it comes out naturally through ray tracing with mirror roughness and misalignment.

ctoennis:
@kosack @maxnoe I am preparing a PR for the tools to process the variance imaes for the StarTracker that will need the code here. Could you give it another look and see what needs to be changed?

docs/changes/2643.feature.rst
```diff
@@ -0,0 +1,4 @@
+Adds psf models to ``ctapipe.instrument.optics`` with the parent class ``PSFModel`` and a psf model based on pure coma abbaration called ``ComaModel``
```
mexanick, line 1: abbaration --> aberration

src/ctapipe/instrument/optics.py
```diff
@@ -263,3 +270,158 @@ def __repr__(self):
...
+    asymmetry_params describe the dependency of the psf on the distance to the center of the camera
+    Used to calculate a pdf asymmetry parameter K of the asymmetric radial laplacian of the psf as
+    a function of the distance r to the optical axis
+
+    .. math:: K(r) = 1 - asym_0 \tanh(asym_1 r) - asym_2 r
+
+    radial_scale_params describes the dependency of the radial scale on the distance to the center of the camera
+    Used to calculate width Sr of the asymmetric radial laplacian in the PSF as a of function the distance r to the optical axis
+
+    .. math:: S_{R}(r) & = b_1 - b_2\,r + b_3\,r^2 + b_4\,r^3
```
mexanick, line 315: Since you changed the implementation to `np.polyval`, you should update also the definition here:
```
.. math:: S_{R}(r) & = b_1 + b_2\,r + b_3\,r^2 + b_4\,r^3
```

src/ctapipe/instrument/tests/test_psf_model.py
```diff
@@ -0,0 +1,32 @@
...
+            az_scale_params=[0.24271557, 7.5511501],
+        )
+
+
+def test_asymptotic_behavior(example_subarray):
+    psf_coma = PSFModel.from_name(
+        "ComaModel",
+        subarray=example_subarray,
+        asymmetry_params=[0.49244797, 9.23573115, 0.15216096],
+        radial_scale_params=[0.01409259, 0.02947208, 0.06000271, -0.02969355],
```
mexanick, line 29: Your parametrization is wrong now, since you changed the signs of coefficients. Please update and add proper tests.

docs/changes/2643.feature.rst
```diff
@@ -0,0 +1,4 @@
...
+- All psf models have a function ``PSFModel.update_location`` that updates the location in the focal plane where the PSF is evaluated from
```
mexanick, line 2: This and the next line are not correct anymore.

src/ctapipe/instrument/tests/test_psf_model.py
```diff
@@ -0,0 +1,32 @@
...
+
+from ctapipe.instrument.optics import PSFModel
+
+
+def test_psf(example_subarray):
+    with pytest.raises(
+        ValueError,
+        match="asymmetry_params and az_scale_params needs to have length 3 and radial_scale_params length 4",
+    ):
+        PSFModel.from_name(
```
mexanick, line 15: use fixtures to avoid code duplications.

mexanick (changes requested):

src/ctapipe/instrument/tests/test_psf_model.py
```diff
@@ -7,26 +7,43 @@
 from ctapipe.instrument.optics import PSFModel
 
 
-def test_psf(example_subarray):
+@pytest.fixture(scope="session")
+def asymmetry_params():
+    return [0.49244797, 9.23573115, 0.15216096]
```
maxnoe, line 12: Fixtures are not really meant for simple data like this. 

You could just enter them in the test itself.

Also: why these highly specific values just for a unit test? Wouldn't nice round values in the right order of magnitude be easier on the eyes?

src/ctapipe/instrument/optics.py
```diff
@@ -263,3 +270,158 @@ def __repr__(self):
...
+    """
+
+    @abstractmethod
+    def pdf(self, r, f, r0, f0, *args):
+        """
+        Calculates the value of the psf at a given location
+
+        Parameters
+        ----------
+        r : float
```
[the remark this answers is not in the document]
ctoennis: i changed the API to use carthesian coordinates and fixed the docustrings to be more clear. I also changed the parameters to round numbers that give reasonable values for some LST-sized camera.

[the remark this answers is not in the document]
maxnoe: Here it still has r,f,r0,f0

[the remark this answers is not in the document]
maxnoe: And if meters are required, then this should probably use `u.quantity_input(..)` with astropy units

src/ctapipe/instrument/optics.py
```diff
@@ -263,3 +270,158 @@ def __repr__(self):
...
+        sr = self.sr_func(r0)
+        sf = self.sf_func(r0)
+        self.radial_pdf_params = (k, r0, sr)
+        self.azimuthal_pdf_params = (f0, sf)
+
+        return laplace_asymmetric.pdf(r, *self.radial_pdf_params) * laplace.pdf(
+            f, *self.azimuthal_pdf_params
+        )
+
+    def check_model_parameters(self):
```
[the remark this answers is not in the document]
ctoennis: Sounds good. I will get to it in about an hour.

mexanick (dismissed):
Looks ok for the moment. I've posted one more suggestion of change, but this can be addressed in the future PR.

src/ctapipe/instrument/optics.py
```diff
@@ -263,3 +272,168 @@ def __repr__(self):
...
+        psf : float
+            value of the PSF at the specified location with the specified position of the point source
+        """
+
+        pass
+
+
+class ComaModel(PSFModel):
+    r"""
+    PSF model, describing pure coma aberrations PSF effect
```
maxnoe, line 308: Need an empty line here after the description.

maxnoe:
I wanted to make a plot of the psf, but this is giving an error:

```python
import astropy.units as u
import numpy as np
from ctapipe.instrument.optics import ComaModel

subarray = SubarrayDescription.read("dataset://gamma_prod5.simtel.zst")

lst1 = subarray.select_subarray([1])
psf_model = ComaModel(
    subarray=lst1,
    asymmetry_params=[0.5, 10, 0.15],
    radial_scale_params=[0.015, -0.1, 0.06, 0.03],
    az_scale_params=[0.25, 7.5, 0.02],
)
edges_x = np.linspace(-0.1, 0.1, 250) * u.m
edges_y = np.linspace(-0.1, 0.1, 250) * u.m
centers_x = 0.5 * (edges_x[:-1] + edges_x[1:])
centers_y = 0.5 * (edges_y[:-1] + edges_y[1:])
x, y = np.meshgrid(centers_x, centers_y)

psf = psf_model.pdf(x, y, 0.0 * u.m, 0.0 * u.m)

fig, ax = plt.subplots()

ax.pcolormesh(psf, edges_x.to_value(u.m), edges_y.to_value(u.m), cmap='inferno')

plt.show()
```

It seems the `ComaModel.pdf` does not support broadcasting, which is I think should be supported.

src/ctapipe/instrument/optics.py
```diff
@@ -263,3 +272,168 @@ def __repr__(self):
...
+        ----------
+        psf : float
+            value of the PSF at the specified location with the specified position of the point source
+        """
+        f, _, r = cartesian_to_spherical((x, y, 0.0))
+        f0, _, r0 = cartesian_to_spherical((x0, y0, 0.0))
+        k = self.k_func(r0)
+        sr = self.sr_func(r0)
+        sf = self.sf_func(r0)
+        self.radial_pdf_params = (k, r0, sr)
```
maxnoe, line 416: Calling the pdf method should not assign parameters to `self`

src/ctapipe/instrument/optics.py
```diff
@@ -263,3 +272,168 @@ def __repr__(self):
...
+
+    def k_func(self, x):
+        return (
+            1
+            - self.asymmetry_params[0] * np.tanh(self.asymmetry_params[1] * x)
+            - self.asymmetry_params[2] * x
+        )
+
+    def sr_func(self, x):
+        return np.polyval(self.radial_scale_params, x)
```
maxnoe, line 385: needs a `[::-1]` on the parameters as `polyval` goes from highest power to lowest power (Horner schema)

mexanick:
@maxnoe , continuing the discussion started in #2689 :

The parametrization in polar coordinates works well under assumption that the radial and polar components are independent, i.e. $\vec{r}\perp\vec{\phi}$. This breaks at low radii, that's why some workarounds were introduced. However, with the Cherenkov camera we are not that much interested in the stars images that close to the center of the camera (we need a lever arm and a larger light spread to obtain reasonable precision), that's why no particular attention was given to this kind of edge cases before. I agree that it looks a bit ugly, and a better parametrization can be considered, but I would propose to move on for the moment with this and add other parameterizations in the future as needed (there'll be need to do this anyway for double-mirror telescopes).

maxnoe:
When evaluating at 0,0 I got a warning: 

```
/home/maxnoe/Uni/CTA/ctapipe/src/ctapipe/instrument/optics.py:430: RuntimeWarning: invalid value encountered in scalar divide
  dphi = np.arcsin(chord_length / (2 * r0))
```

otherwise looks good

maxnoe:
The math does not render properly: https://ctapipe--2643.org.readthedocs.build/en/2643/api/ctapipe.instrument.ComaPSFModel.html#ctapipe.instrument.ComaPSFModel

maxnoe:
>  Also, the init docstring is ignored

Yes, documentation should be in the class docstring, not `__init__`. Add the `Parameters` section to the class docstring

maxnoe (approved):

maxnoe (approved):

mexanick:
@kosack can you please take a look? I shouldn't participate in the review any further as I've contributed to the code.

src/ctapipe/instrument/optics.py
```diff
@@ -263,3 +271,172 @@ def __repr__(self):
...
+
+
+class ComaPSFModel(PSFModel):
+    r"""
+    PSF model describing pure coma aberrations PSF effect.
+
+    The PSF is described by a combination of an asymmetric Laplacian for the radial part and a symmetric Laplacian in the polar direction.
+    It uses an asymmetric Laplacian for the radial part
+
+    .. math:: f_{R}(r, K) = \begin{cases}\frac{1}{S_{R}(K+K^{-1})}e^{-K\frac{r-L}{S_{R}}}, r\ge L\\ \frac{1}{S_{R}(K+K^{-1})}e^{\frac{r-L}{KS_{R}}}, r < L\end{cases}
```
Tobychev, line 319: Probably better to rename L, the breakpoint of the asymmetric Laplacian, to r_0 in line with the terminology of the code?

src/ctapipe/instrument/optics.py
```diff
@@ -263,3 +271,172 @@ def __repr__(self):
...
+
+
+    Attributes
+    ----------
+    asymmetry_params : list
+        Describes the dependency of the PSF on the distance to the center of the camera.
+        Used to calculate a PDF asymmetry parameter K of the asymmetric radial Laplacian
+        of the PSF as a function of the distance r to the optical axis.
+
+        .. math:: K(r) = 1 - \text{asym}_0 \tanh(\text{asym}_1 r) - \text{asym}_2 r
```
Tobychev, line 333: Why do you use the unusual name  `\text{asym}` for these constants when the reference just uses "c", in line with how the constants for the `radial_scale_params` and `phi_scale_params` are named?

src/ctapipe/instrument/optics.py
```diff
@@ -263,3 +271,172 @@ def __repr__(self):
...
+    phi = np.arctan2(y, x)
+    return r, phi
+
+
+class ComaPSFModel(PSFModel):
+    r"""
+    PSF model describing pure coma aberrations PSF effect.
+
+    The PSF is described by a combination of an asymmetric Laplacian for the radial part and a symmetric Laplacian in the polar direction.
+    It uses an asymmetric Laplacian for the radial part
```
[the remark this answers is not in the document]
Tobychev: I also suggest naming all the constants you introduce here, or at least saying they are defined further down in the attributes section.

maxnoe (approved):

maxnoe (changes requested):
@mexanick Could you please fix the sonarqube complaints:

https://sonar-cta-dpps.zeuthen.desy.de/project/issues?resolved=false&inNewCodePeriod=true&types=CODE_SMELL&pullRequest=2643&id=cta-observatory_ctapipe_AY52EYhuvuGcMFidNyUs&open=AZS3XtUOGE3VkZNQVvss

maxnoe (approved):

Tobychev (approved):

PR #2644, opened 2024-11-11
Add property to EventSource for support of CTAO reference metadata

maxnoe:
* Set to false for SimTelEventSource
* Silences a warning for SimTelEventSource, which at least currently cannot support the reference metadata.


Tobychev (approved):

maxnoe:
I think a better solution would be #2648

kosack (approved):

maxnoe:
Closing in favor of #2648

Issue #2571, opened 2024-06-28
`Provenance().add_input_file()` should take product id


maxnoe:
**Please describe the use case that requires this feature.**

Add the moment, the provenance system only stores the path of an input file.

However, for CTAO data products used as input, the most important thing to store would actually be the product id of the reference metadata, not just the path.

**Describe the solution you'd like**
Optionally also store product uuid in provenance.

kosack:
Would be useful to also add a nice helper function to extract the product_id from existing files.  In fact that could just be automatic: `add_input_file(path)` could call `get_product_id(path)`, and if it fails, emit a warning and continue as we do now without recording it.   That way it's more foolproof that the real product_id gets propagated

PR #2648, opened 2024-11-12
Improve reference metadata handling for EventSource

maxnoe:
The recent addition of attaching the reference metadata of input files to the provenance information implemented only that the reference metadata is read directly by ctapipe from the input file. This made it necessary to either support all possible input file types
or impossible for plugin event sources to provide this metadata on their own. It also makes the assumption 1 EventSource = 1 input file. This is not true for some event sources.

This issue is solved by:
* Adding the possibility to directly provide the reference metadata to `add_input_file`
* Move the responsibility of calling `add_input_file` from the EventSource baseclass to the implementation

This is an alternative implementation to https://github.com/cta-observatory/ctapipe/pull/2644

### DIFF
--- a/src/ctapipe/core/provenance.py
+++ b/src/ctapipe/core/provenance.py
@@ -62,6 +62,138 @@ class MissingReferenceMetadata(UserWarning):
     """Warning raised if reference metadata could not be read from input file."""
 
 
+class _ActivityProvenance:
+    """
+    Low-level helper class to collect provenance information for a given
+    *activity*.  Users should use `Provenance` as a top-level API,
+    not this class directly.
+    """
+
+    def __init__(self, activity_name=sys.executable):
+        self._prov = {
+            "activity_name": activity_name,
+            "activity_uuid": str(uuid.uuid4()),
+            "status": "running",
+            "start": {},
+            "stop": {},
+            "system": {},
+            "input": [],
+            "output": [],
+            "exit_code": None,
+        }
+        self.name = activity_name
+
+    def start(self):
+        """begin recording provenance for this activity. Set's up the system
+        and startup provenance data. Generally should be called at start of a
+        program."""
+        self._prov["start"].update(_sample_cpu_and_memory())
+        self._prov["system"].update(_get_system_provenance())
+
+    def _register(self, what, url, role, add_meta, reference_meta):
+        if what not in {"input", "output"}:
+            raise ValueError("what must be 'input' or 'output'")
+
+        reference_meta = self._get_reference_meta(
+            url=url, reference_meta=reference_meta, read_meta=add_meta
+        )
+        self._prov[what].append(dict(url=url, role=role, reference_meta=reference_meta))
+
+    def register_input(self, url, role=None, add_meta=True, reference_meta=None):
+        """
+        Add a URL of a file to the list of inputs (can be a filename or full
+        url, if no URL specifier is given, assume 'file://')
+
+        Parameters
+        ----------
+        url: str
+            filename or url of input file
+        role: str
+            role name that this input satisfies
+        add_meta: bool
+            If true, try to load reference metadata from input file
+            and add to provenance.
+        """
+        self._register(
+            "input", url, role=role, add_meta=add_meta, reference_meta=reference_meta
+        )
+
+    def register_output(self, url, role=None, add_meta=True, reference_meta=None):
+        """
+        Add a URL of a file to the list of outputs (can be a filename or full
+        url, if no URL specifier is given, assume 'file://')
+
+        Should only be called once the file is finalized, so that reference metadata
+        can be read.
+
+        Parameters
+        ----------
+        url: str
+            filename or url of output file
+        role: str
+            role name that this output satisfies
+        add_meta: bool
+            If true, try to load reference metadata from input file
+            and add to provenance.
+        """
+        self._register(
+            "output", url, role=role, add_meta=add_meta, reference_meta=reference_meta
+        )
+
+    def register_config(self, config):
+        """add a dictionary of configuration parameters to this activity"""
+        self._prov["config"] = config
+
+    def finish(self, status="success", exit_code=0):
+        """record final provenance information, normally called at shutdown."""
+        self._prov["stop"].update(_sample_cpu_and_memory())
+
+        # record the duration (wall-clock) for this activity
+        t_start = Time(self._prov["start"]["time_utc"], format="isot")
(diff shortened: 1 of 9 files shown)


kosack:
Looks like a good idea. Probably need some tests, particularly to ensure that anyone implementing an EventSource plugin doesn't forget to add the input file to the provenance.

maxnoe:
>  anyone implementing an EventSource plugin doesn't forget to add the input file to the provenance.

MMh. The only way we could enforce this I think is to add an abstract method to the `EventSource` that subclasses then have to implement.

However, I want to support event sources that have multiple input files, as I know those exist (and we'll have e.g. parallel zfits streams also for the ACADA data).

So the API should be something like this:

```
@abstractmethod
def get_input_reference_meta(self) -> dict[str, ReferenceMeta]:
    pass 
```

But the issue here is that I know that some event source implementations only open files one-by-one. 

So I think the solution above is not right. It's really the event sources that have to internally call `Provenance().add_input_file(path, reference_meta=reference_meta)` for each file they open.

I don't think we can have a unit test here in ctapipe to enforce this.

kosack:
>  I want to support event sources that have multiple input files, as I know those exist (and we'll have e.g. parallel zfits streams also for the ACADA data).

The reference metadata was designed specifically to be an _output_, not an _input_, so this may not be fully necessary to go that deep here.   All we need to get is the info that we will eventually need to write _out_, so e.g. the full list of input product_ids is not necessary. The back-links to that info is what would be done in a provenance database derived from the provenance logs.  

So I think the issue with multiple inputs in the *first* element in the processing chain is: _how to assign a `product_id` to the **set** of products_?  For current data from e.g. LST et al, this is not solved, however for CTAO since we will in any case have to define a "observation dataset package" for a single observation (containing all event, monitoring, service data). That package will have a single `product_id` as well as `obs_id` that can be traced to its contents externally.  For raw DL0 data that should never change, it could be that we define `obs_id`==dataset `product_id` for that oservation.

maxnoe:
I am surprised to hear this... in the original issue https://github.com/cta-observatory/ctapipe/issues/2571#issuecomment-2199980543 and the implementation https://github.com/cta-observatory/ctapipe/pull/2598  you supported the idea of getting the product ids for inputs

kosack:
> I am surprised to hear this... in the original issue https://github.com/cta-observatory/ctapipe/issues/2571#issuecomment-2199980543 and the implementation https://github.com/cta-observatory/ctapipe/pull/2598 you supported the idea of getting the product ids for inputs

I think I didn't explain well: I do support this (`product_id`s are better than just a list of input filename, as they provide a real link with no possibility of collisions).  This is a special case however: we are talking about the _first_ thing in the processing chain (a "source") where we don't have control over how the inputs were generated.  So my suggestion is we could support the case of inputs where we know there cannot be multiple reprocessed copies by just using something like the obs_id in that case.  It's not perfect, but solves the issue of not having a UUID generated by things outside of our control.  It only works for DL0 data.

The other issue is the difference between "Local Provenance" (inputs and outputs of an _Activity_) and"`ReferenceMetedata` (which relates only to a _single_  _Data Product_).    The former is for full provenance linking, and  the latter only stores the forward-part of the provenance, and doesn't  include the list of input files.  The use cases are different: provenance is for looking at the full graph, while the ReferenceMetadata is for retrieving a file by common attributes.   So I think we need the input product_ids in the local provanance, but not in the Reference metadata

maxnoe:
> So I think we need the input product_ids in the local provanance, but not in the Reference metadata.

That's how this works here. It is attaching the reference metadata of input files to the provenance, not to the output reference meta.

src/ctapipe/io/eventsource.py
```diff
@@ -134,23 +134,6 @@ def __new__(cls, input_url=Undefined, config=None, parent=None, **kwargs):
         return super().__new__(subcls)
 
     def __init__(self, input_url=None, config=None, parent=None, **kwargs):
-        """
```
Tobychev, line 137: Why is the help string to `__init__` being removed?

src/ctapipe/core/provenance.py
```diff
@@ -85,18 +217,18 @@ def start_activity(self, activity_name=sys.executable):
         activity.start()
         self._activities.append(activity)
         log.debug(f"started activity: {activity_name}")
+        return activity
```
Tobychev, line 220: This new return is not mentioned in the class or method doc strings, so they need updating to let people know they can start an activity and immediately use it. 

Though it seems like there's not many places `start_activity` is used outside of the tests anyway.

src/ctapipe/io/eventsource.py
maxnoe: Because it had outdated information and `__init__` shouldn't be used at all for documentation in the numpy style guide. The class docstring should be used (and is):

https://numpydoc.readthedocs.io/en/latest/format.html#class-docstring

src/ctapipe/core/provenance.py
maxnoe: This is mostly an internal method to be called when a `Tool` starts up. It made my linter a bit happier to return here instead of calling this and then return `self.current_activity` in the method below.

src/ctapipe/io/eventsource.py
Tobychev: The class docstring is missing the constructor parameters mandated by the styleguide

> The constructor (`__init__`) **should** also be documented here, the [Parameters](https://numpydoc.readthedocs.io/en/latest/format.html#params) section of the docstring details the constructor’s parameters.
> An Attributes section, **located below** the [Parameters](https://numpydoc.readthedocs.io/en/latest/format.html#params) section,
(emphasis mine).

maxnoe: I'd rather say that the section is wrongly labeled Attributes instead of parameters...

maxnoe: fixed

src/ctapipe/io/eventsource.py
```diff
@@ -79,16 +79,18 @@ class EventSource(Component):
...
+    input_url : str | Path
         Path to the input event file.
     max_events : int
         Maximum number of events to loop through in generator
     allowed_tels: Set or None
         Ids of the telescopes to be included in the data.
         If given, only this subset of telescopes will be present in the
         generated events. If None, all available telescopes are used.
+
+    Parameters
```
Tobychev, line 93: Added twice? You meant attributes I guess?

kosack (approved):

PR #2657, opened 2024-11-21
Use sphinx-changelog

maxnoe:
Replace custom solution with using sphinx-changelog, this also works on readthedocs and means we get the changelog previewed in the "latest" docs.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -206,9 +206,5 @@ jobs:
           git describe --tags
           python -c 'import ctapipe; print(ctapipe.__version__)'
 
-      - name: Produce Changelog
-        run: |
-          towncrier build --yes
-
       - name: Build docs
         run: make doc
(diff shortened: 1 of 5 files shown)


Tobychev:
Results look identical between the CI version and latest, so I'm confused about what you mean by this "we get the changelog previewed"?

I can spot one behaviour change between the [CI version](https://ctapipe--2657.org.readthedocs.build/en/2657/changelog.html#ctapipe-v0-23-0-2024-11-18) and [prod](https://ctapipe.readthedocs.io/en/stable/changelog.html): when you scroll all the way up on the CI the right hand side content menu doesn't collapse down to one line per release like it does on prod.

maxnoe:
> Results look identical between the CI version and latest, 

That's because we haven't merged anything yet since the last tag

kosack (approved):

maxnoe:
But this branch is not updated with respect to merge

Tobychev (approved):

maxnoe:
You can now see it here: https://ctapipe.readthedocs.io/en/latest/changelog.html @Tobychev

Issue #1110, opened 2019-07-17
Break up `utils` into subpackages with better names


MaxNoe:
Something called `utils` will mostly end up as a catch it all module of unrelated stuff.

kosack:
current contents (which you can see are already pretty unrelated):

* ``astro``: star catalog access
* ``datasets``: access to external (supporting) data like ctapipe-extra or files in the CTAPIPE_SVC_PATH
* ``CutFlow``: class to apply and track cuts applied to events
* ``fitshistogram``: class to create and manage N-D histograms, with read/write to FITS
* ``json2fits``: convert JSON docs to FITS headers (not used?) 
* ``linalg``: linear algebra helper functions
* ``dynamic_class``:  construct a class by name (I think). Doesn't seem to be used anymore...
* ``table_interpolator, table_interpolator, template_network_interpolator, unstructured_interpolator``: general ND interpolators used by ImPACT?
* ``rgbtohex``: color conversion for e.g. web guis using Bokeh

MaxNoe:
> I think it was for serializing config files to FITS)

God help us

MaxNoe:
Had a look at linalg:

* rotation_matrix_2d: should go into the coordinate trafos
* length: not used
* normalize and angle: verbatim copies of the same functions in  `HillasReconstructor` and not used anywhere else

I think this module should be removed and the function defined where they are used.

PR #2658, opened 2024-11-22
Fix uncertainty calculation in StereoMeanCombiner

LukasBeiske:
This fixes a problem with the uncertainties calculated by `StereoMeanCombiner` where the mapping of telescope events to array events did not work correctly. This resulted in many nan values in e.g. the `RandomForestRegressor_energy_uncert` column.

In doing so, I added a new module (`ctapipe.vectorization`) for the helper functions I implemented. This module also contains a faster (numba) alternative to `np.unique`, `ctapipe.vectorization.get_subarray_index`.

This is part of #2497, but I think this should be fixed regardless of the other content of that PR.


src/ctapipe/vectorization/aggregate.py
```diff
@@ -0,0 +1,145 @@
+"""Helper functions for vectorizing numpy operations."""
+
+import numpy as np
```
maxnoe, line 3: Do we really need a full new submodule for this? This looks like it could be in `ctapipe/reco/telescope_event_handling` or similar

LukasBeiske: I think I put it in `ctapipe.utils` initially, but that module is already pretty large. 
But that would be fine with me. Alternatively, what about `ctapipe.reco.utils`?

maxnoe: Sure, that also fits

kosack: Maybe just make a better-named, potentially private, sub-module under `reco` or `io`. Dumping random things into modules called "Utils" ends up being too generic (see #1110).

LukasBeiske: Good point. I went with Max' initial idea of `ctapipe.reco.telescope_event_handling`. Regarding the docs: Currently only the different reconstructors themselves are included in the docs, but none of the different helper/ preprocessing functions. Do we want to keep it this way?

src/ctapipe/reco/stereo_combination.py
```diff
@@ -328,29 +298,21 @@ def predict_table(self, mono_predictions: Table) -> Table:
             stereo_table[f"{self.prefix}_goodness_of_fit"] = np.nan
 
         elif self.property is ReconstructionProperty.ENERGY:
-            if len(valid_predictions) > 0:
-                mono_energies = valid_predictions[f"{prefix}_energy"].quantity.to_value(
+            if np.sum(valid) > 0:
```
maxnoe, line 301: count_nonzero

src/ctapipe/reco/stereo_combination.py
```diff
@@ -299,26 +268,27 @@ def predict_table(self, mono_predictions: Table) -> Table:
...
-        n_array_events = len(array_events)
-        weights = self._calculate_weights(valid_predictions)
+        weights = self._calculate_weights(mono_predictions[valid])
 
         if self.property is ReconstructionProperty.PARTICLE_TYPE:
-            if len(valid_predictions) > 0:
-                mono_predictions = valid_predictions[f"{prefix}_prediction"]
-                stereo_predictions = _weighted_mean_ufunc(
-                    mono_predictions, weights, n_array_events, indices[valid]
+            if np.sum(valid) > 0:
```
maxnoe, line 284: count_nonzero

src/ctapipe/reco/stereo_combination.py
```diff
@@ -369,22 +331,37 @@ def predict_table(self, mono_predictions: Table) -> Table:
             stereo_table[f"{self.prefix}_goodness_of_fit"] = np.nan
 
         elif self.property is ReconstructionProperty.GEOMETRY:
-            if len(valid_predictions) > 0:
+            if np.sum(valid) > 0:
```
maxnoe, line 334: count_nonzero

src/ctapipe/reco/telecope_event_handling.py
```diff
@@ -0,0 +1,145 @@
...
+    return combined_values
+
+
+def weighted_mean_std_ufunc(
+    tel_values,
+    valid_tel,
+    n_array_events,
+    indices,
+    multiplicity,
+    weights=np.array([1]),
```
maxnoe, line 91: Using 1 is the easiest code, but we should be able to gain performance and the difference in code should not be so large to special case the "no weights" case, right?

src/ctapipe/reco/telecope_event_handling.py
```diff
@@ -0,0 +1,145 @@
...
+    """
+    combined_values = np.zeros(n_array_events)
+    np.add.at(combined_values, indices, tel_data)
+    return combined_values
+
+
+def weighted_mean_std_ufunc(
+    tel_values,
+    valid_tel,
+    n_array_events,
```
maxnoe, line 88: Is this argument needed? `n_array_elements == len(multiplicities)`, right?

src/ctapipe/reco/telecope_event_handling.py
```diff
@@ -0,0 +1,145 @@
...
+        idx,
+    )
+
+
+def get_subarray_index(tel_table):
+    """
+    Get the obs_ids and event_ids of all subarray events contained
+    in a table of telescope events, their multiplicity and an array
+    giving the index of the subarray event for each telescope event.
+    This requires the telescope events to be SORTED by their corresponding
```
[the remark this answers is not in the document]
LukasBeiske: That is what I meant by "Sorted by their corresponding subarray events", but I admit that this is not a good way of wording it.

src/ctapipe/reco/telecope_event_handling.py
LukasBeiske: No, it's not. Your right :+1:

src/ctapipe/reco/telecope_event_handling.py
LukasBeiske: I am struggling to get the multiplicity including only the valid telescope events (as in what I need to divide `sum_prediction` by) without calculating something like a `sum_of_weights`. Obviously, it would be possible to get them in `StereoMeanCombiner` via `np.count_nonzero(events[<prefix-of-energy-reco>_telescopes])` and parse them into `weighted_mean_std_ufunc`, but I think that would be really ugly and confusing, since this would only be needed for `weights='none'` and I would prefer any other solution. 
I will have a look again on Monday. I'm probably missing something really obvious...

maxnoe (approved):

maxnoe:
The coverage here is too low because we cannot get a valid coverage reported for the numba jitted  code, the actual coverage is pretty close to 100 %.

maxnoe (approved):

docs/changes/2658.bugfix.rst
```diff
@@ -0,0 +1,2 @@
+Fix ``<reconstruction_property>_uncert`` calculations in ``ctapipe.reco.StereoMeanCombiner``.
+Add helper functions for vectorized numpy calculations as new ``ctapipe.vectorization`` module.
```
Hckjs, line 2: Should be adapted

maxnoe: done

maxnoe (approved):

PR #2666, opened 2024-12-04
Fixed unknown configs being silently ignored

kosack:
Fixes bug where if a configuration file with unknown file extension is passed to a tool, e.g. `--config myconf.conf` instead of `--config myconf.yaml`, it was silently ignored.   Config files must now have one of the following extensions to be recognized: yml, yaml, toml, json, py, and if not a `ToolConfigurationError` is raised.

### DIFF
--- a/src/ctapipe/core/tool.py
+++ b/src/ctapipe/core/tool.py
@@ -282,9 +282,16 @@ class Tool(Application):
             with open(path, "rb") as infile:
                 config = Config(toml.load(infile))
             self.update_config(config)
-        else:
-            # fall back to traitlets.config.Application's implementation
+        elif path.suffix in ["json", "py"]:
+            # fall back to traitlets.config.Application's implementation. Note
+            # that if we don't specify the file suffixes here, traitlets seems
+            # to silently ignore unknown ones.
             super().load_config_file(str(path))
+        else:
+            raise ToolConfigurationError(
+                f"The config file '{path}' is not in a known format. "
+                "The file should end in one: yml, yaml, toml, json, py"
+            )
 
         Provenance().add_input_file(path, role="Tool Configuration", add_meta=False)
 
--- a/src/ctapipe/core/tests/test_tool.py
+++ b/src/ctapipe/core/tests/test_tool.py
@@ -569,3 +569,27 @@ def test_exit_status_interrupted(tmp_path, provenance):
     provlog = activities[0]
     assert provlog["activity_name"] == InterruptTestTool.name
     assert provlog["status"] == "interrupted"
+
+
+def test_no_ignore_bad_config_type(tmp_path: Path):
+    """Check that if an unknown type of config file is given, an error is raised
+    instead of silently ignoring the file (which is the default for
+    traitlets.config)
+    """
+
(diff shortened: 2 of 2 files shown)


src/ctapipe/core/tool.py
```diff
@@ -282,9 +282,16 @@ def load_config_file(self, path: str | pathlib.Path) -> None:
             with open(path, "rb") as infile:
                 config = Config(toml.load(infile))
             self.update_config(config)
-        else:
```
maxnoe, line 285: Okay, so the issue is _traitlets_ silently ignores errors in `load_config_file`? That would be worth an upstream issue I guess.

The intended behavior here was that we pass any format not directly implemented by us to traitlets, so we get all traitlets formats without having to know what is supported.

maxnoe: Yes, good fix for now. Maybe we should follow up with an issue upstream

maxnoe (approved):

src/ctapipe/core/tool.py
maxnoe: Looking at traitlets, the situation is even worse.

Setting `--config=example.conf` will result in loading `example.json` and `example.py` if they are in the same directory. TRaitlets splits off the extension, never looks at it and then  loads all files that have the same stem and one of the supported extensions (json, py)

maxnoe: I opened an issue about this 1.5 years ago: https://github.com/ipython/traitlets/issues/850

kosack:
Even worse, I though maybe  could check `loaded_config_files` following the traitlets documentation, but even when it correctly loads, it seems that variable is not updated:

```
In [8]: class TestTool(Tool):
   ...:     a = traits.Int(0).tag(config=True)
   ...:

In [9]: tool = TestTool()

In [10]: tool.load_config_file("test.yaml")

In [11]: tool.loaded_config_files
Out[11]: []

In [12]: tool.a
Out[12]: 3
```

In general their API is a mix of internal and public methods, with no separation between them...

src/ctapipe/core/tool.py
```diff
@@ -282,9 +282,16 @@ def load_config_file(self, path: str | pathlib.Path) -> None:
...
-            # fall back to traitlets.config.Application's implementation
+        elif path.suffix in ["json", "py"]:
```
maxnoe, line 285: Suffix includes the ".", this is why tests are failing

maxnoe:
> FAILED src/ctapipe/io/tests/test_simteleventsource.py::test_prod6_issues - FileNotFoundError: Couldn't find resource: 'prod6_issues.simtel.zst', You might want to install ctapipe_resources
> FAILED src/ctapipe/tools/tests/test_process.py::test_prod6_issues - FileNotFoundError: Couldn't find resource: 'prod6_issues.simtel.zst', You might want to install ctapipe_resources

Damn... this is a cross issue with the switch to minio, we uploaded a new file after Gareth moved the data

Tobychev (approved):

maxnoe:
Gareth uploaded the file, CI is passing now

PR #2670, opened 2024-12-10
Muon parameters update

### LINKED ISSUE #2665: Implement Additional Muon Parameters in ctapipe for LST Analysis
Currently, we are lacking some useful muon parameters that are used in processing LST-1 data and can be beneficial for subsequent analysis. I think it would be good to have them in `ctapipe` because we will need to use them in `calibpipe` for the derivation of calibration coefficients. For example, such muon metrics as:

- `ring_size` and `size_outside` - the sum of the photoelectrons in the ring area, defined as a region `dist_mask` = (R-0.25R; R+0.25R), and outside of it. 
- `mean_pixel_charge_around_ring` - average pixel charge in pixels surrounding the outer part of the ring
- `radial_stdev` - standard deviation of (cleaned) light distribution along ring radius
- `radial_skewness` - skewness of (cleaned) light distribution along ring radius
- `radial_excess_kurtosis` - excess kurtosis of (cleaned) light distribution along ring radius
- `num_pixels_in_ring` - number of pixels in the integration area `dist_mask`

Taking the above into account, I would like to:

- Add methods for calculating these parameters to the module `ctapipe.image.muon.features` (following already implemented functionality in the `lstchain`).
- Add appropriate fields to the `MuonParameterContainers` to store the new parameters.
- Implement calculation of these parameters inside `MuonProcessor._calculate_muon_parameters()`

Please comment in case something is missing or we should follow other direction. Otherwise I can start to create corresponding PRs and implement these features.

jstvdk:
I added new functionality to the muon analysis as described in the Issue #2665 , specifically

- new fields to the `MuonParametersContainer`
- new methods to the `ctapipe/image/muon/features.py`
- new tests for above mentioned methods in `ctapipe/image/muon/tests/test_muon_features.py`
- additional code block in `ctapipe/image/muon/processor.py` to process and save new muon parameters

Added changes mainly follow existing code in `lstchain/image/muon/muon_analysis.py`, with one exception:

- I redefined the variable `num_pixels_in_ring` - now it takes into account only pixels that survived the cleaning, contrary to what currently implemented in `lstchain`, where this variable contains all pixels inside the geometrical area of the ring.

Closes #2665


src/ctapipe/containers.py
```diff
@@ -1189,6 +1189,18 @@ class MuonParametersContainer(Container):
         nan * u.deg**2,
         "MSE of the deviation of all pixels after cleaning from the ring fit",
     )
+    ring_size = Field(nan, "Number of pixels in the ring")
```
maxnoe, line 1192: Is this really number of pixels? The code below computes some of the pixel intensities.

If this is actually the sum of pixel charges, the name should be `ring_intensity` and the help string match what is actually computed.

src/ctapipe/containers.py
```diff
@@ -1189,6 +1189,18 @@ class MuonParametersContainer(Container):
...
+    size_outside = Field(nan, "Number of pixels outside the ring")
```
maxnoe, line 1193: same here: probably this should be `intensity_outside_ring`

src/ctapipe/containers.py
```diff
@@ -1189,6 +1189,18 @@ class MuonParametersContainer(Container):
...
+    num_pixels_in_ring = Field(nan, "Number of pixels in the ring")
```
maxnoe, line 1194: We use the `n_` prefix for variables that are numbers, not `num_`

src/ctapipe/containers.py
```diff
@@ -1189,6 +1189,18 @@ class MuonParametersContainer(Container):
...
+    mean_pixel_outside_ring = Field(nan, "Mean pixel charge outside the ring")
```
maxnoe, line 1195: there is an `intensity` or similar missing in the field name. 

I'd call it `mean_intensity_outside_ring`

src/ctapipe/containers.py
```diff
@@ -1189,6 +1189,18 @@ class MuonParametersContainer(Container):
...
+    standard_dev = Field(
+        nan * u.deg,
+        "Standard deviation of the light distribution along the ring radius.",
```
maxnoe, line 1198: "along the ring radius" sounds like this is somehow using spatial information.

But I think this is just "of pixels on the ring", right?

src/ctapipe/image/muon/tests/test_muon_features.py
```diff
@@ -50,3 +55,70 @@ def test_ring_completeness():
...
+        center_y,
+        pixel_x,
+        pixel_y,
+        ring_integration_width,
+        outer_ring_width,
+        image,
+        image_mask,
+    )
+
+    assert ring_size > 0
```
maxnoe, line 88: I think this test is not definitive enough. Please consider writing a test actually being sensitive on changes other than the value being <= 0.

I.e. use the ring gaussian toymodel to create a ring with known properties:
https://github.com/cta-observatory/ctapipe/blob/7858ddc3575681d59a266ecc1f20d1510f2f3555/src/ctapipe/image/toymodel.py#L375

maxnoe (changes requested):
Hi @jstvdk 

Thanks a lot for preparing this. This looks like a good starting point. I have a couple of first comments mainly about the names of the fields and their descriptions.

I will do second round on more technical issues of the code later.

src/ctapipe/image/muon/features.py
```diff
@@ -158,3 +158,113 @@ def ring_containment(radius, center_x, center_y, camera_radius):
...
+    """
+
+    if np.sum(image) == 0:
+        return np.nan * u.deg, np.nan, np.nan
+
+    x0 = center_x.to_value(u.deg)
+    y0 = center_y.to_value(u.deg)
+    pix_x = pixel_x.to_value(u.deg)
+    pix_y = pixel_y.to_value(u.deg)
+    pixel_r = np.sqrt((pix_x - x0) ** 2 + (pix_y - y0) ** 2)
```
kosack, line 262:
```suggestion
    pixel_r = np.hypot(pix_x - x0, pix_y - y0)
```

src/ctapipe/image/muon/features.py
```diff
@@ -158,3 +158,113 @@ def ring_containment(radius, center_x, center_y, camera_radius):
...
+        Excess kurtosis of the light distribution along the ring radius.
+    """
+
+    if np.sum(image) == 0:
+        return np.nan * u.deg, np.nan, np.nan
+
+    x0 = center_x.to_value(u.deg)
+    y0 = center_y.to_value(u.deg)
+    pix_x = pixel_x.to_value(u.deg)
+    pix_y = pixel_y.to_value(u.deg)
```
kosack, line 261: I don't think you need to convert and take the value here. Numpy works fine with unit quantities, and then the function will work no matter what unit the inputs are in.

kosack (dismissed):

src/ctapipe/image/muon/features.py
```diff
@@ -200,11 +200,13 @@ def ring_size_parameters(
     ring_intensity: float
         Sum of the p.e. inside the integration area of the ring
     intensity_outside_ring: float
-        Sum of the photons outside the ring integration area that passed the cleaning
+        Sum of the p.e. outside the ring integration area that passed the cleaning
     n_pixels_in_ring: int
         Number of pixels inside the ring integration area that passed the cleaning
     mean_intensity_outside_ring: float
-        Mean intensity of the pixels outside the ring, but still close to it
+        Mean intensity of the pixels outside the integration area of the ring,
```
mexanick, line 207: Is this the same as l1201 of `ctapipe.containers`? If so, why the docstrings are different?

jstvdk: They are the same, but I didn't want to duplicate the exact same text. 
Or should I just duplicate the doc strings from `ctapipe.containers` into the doc strings in the methods from `ctapipe.image.muon.features`?

mexanick: Perhaps this function simply should return `MuonParametersContainer`?
Also, the number of input parameters is very high, consider to wrap them into a structure (dict, container, ...)

src/ctapipe/image/muon/features.py
```diff
@@ -158,3 +158,125 @@ def ring_containment(radius, center_x, center_y, camera_radius):
...
+    ----------
+    center_x : float
+        x coordinate of the ring center.
+    center_y : float
+        y coordinate of the ring center.
+    pixel_x : array-like
+        x coordinates of the camera pixels.
+    pixel_y : array-like
+        y coordinates of the camera pixels.
+        Amplitude of image pixels.
```
mexanick, line 250: "image" is missing?

src/ctapipe/image/muon/features.py
jstvdk: > Perhaps this function simply should return `MuonParametersContainer`? 

All this module in fact is returning just one `MuonParametersContainer` and I adopted already existing style where each method returning some logically/technically connected part of `MuonParametersContainer`.  Should I reconsider this and return calculated parts of `MuonParametersContainer` from every method?

> Also, the number of input parameters is very high, consider to wrap them into a structure (dict, container, ...)

Good point

src/ctapipe/image/muon/features.py
```diff
@@ -158,3 +158,125 @@ def ring_containment(radius, center_x, center_y, camera_radius):
...
+
+    return (
+        ring_intensity,
+        intensity_outside_ring,
+        n_pixels_in_ring,
+        mean_intensity_outside_ring,
+    )
+
+
+def radial_light_distribution(center_x, center_y, pixel_x, pixel_y, image):
```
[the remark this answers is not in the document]
jstvdk: Sorry, let me repeat if I understood correctly. You are proposing to merge all the methods into one, which calculates and returns `MuonParametersContainer`?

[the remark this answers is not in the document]
mexanick: yes

mexanick:
docs are failing due to an issue with the type hint resolution with a few similar warnings:
```
/home/runner/work/ctapipe/ctapipe/src/ctapipe/image/muon/features.py:docstring of ctapipe.image.muon.features.intensity_ratio_inside_ring:1: WARNING: py:class reference target not found: ctapipe.containers.MuonRingContainer [ref.class]
```
@maxnoe any idea how to fix it (apart from adding more ctapipe-related stuff in `nitpick_ignore_regex`?

maxnoe:
Is the `MuonRingContainer` present in `__all__` in containers.py?

src/ctapipe/image/muon/features.py
```diff
@@ -73,88 +107,243 @@ def intensity_ratio_inside_ring(
...
 
-    a = (radius**2 - camera_radius**2 + d**2) / (2 * d)
-    return np.arccos(a / radius) / np.pi
+    a = (ring.radius**2 - camera_radius**2 + ring.center_distance**2) / (
+        2 * ring.center_distance
+    )
+    return np.arccos((a / ring.radius).to_value(u.dimensionless_unscaled)) / np.pi
+
+
+def ring_size_parameters(
```
maxnoe, line 215: `ring_intensity_parameters`?

mexanick: I was thinking about this change, but this function also computes the number of pixels in the ring, so I left it as is. If you have a strong feeling that ring_intensity_parameters will be better, let me know and I will update.

maxnoe: Yes, I think the `size` here came from when everything related to `intensity` was called `size`.

 So I'd say we rename

maxnoe (approved):

docs/changes/2670.feature.rst
```diff
@@ -0,0 +1,10 @@
+Add new fields to the MuonParametersContainer and refactor the code using the existing containers.
```
maxnoe, line 1: This reads like you only modified the container, but you also actually implemented the computation of these new features, right?

Maybe emphasize that a bit?

maxnoe (approved):

maxnoe:
ping @kosack

kosack (approved):

PR #2673, opened 2025-01-09
Fix SubarrayDescription.info for non-unique telescope description strings

### LINKED ISSUE #2672: `SubarrayDescrption` missing telescopes in case of non-unique telescope description string
E.g. with the LSTs having different specs in Prod6:

```
subarray.info()
Subarray : LaPalma-prod6
Num Tels : 32
Footprint: 0.65 km2
Height   : 2156.00 m
Lon/Lat  : -17.892030200000022 deg, 28.762166100000005 deg 

       Type       Count Tel IDs
----------------- ----- -------
   LST_LST_LSTcam     3 2-4    
LST_MAGIC1_MAGIC1     1 31     
LST_MAGIC2_MAGIC2     1 32     
MST_MST_NectarCam    26 5-30  
```

The LST-1 is missing, because LST 2 to 4 have the same Type string.

maxnoe:
Fixes #2672

### DIFF
--- a/docs/changes/2673.bugfix.rst
+++ b/docs/changes/2673.bugfix.rst
@@ -1,0 +1,2 @@
+Fix ``SubarrayDescription.info()`` in cases
+of different ``TelescopeDescription`` with the same string representation.
--- a/src/ctapipe/instrument/subarray.py
+++ b/src/ctapipe/instrument/subarray.py
@@ -2,6 +2,7 @@
 Description of Arrays or Subarrays of telescopes
 """
 import warnings
+from collections import defaultdict
 from collections.abc import Iterable
 from contextlib import ExitStack
 from copy import copy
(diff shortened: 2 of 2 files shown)


kosack (approved):
This also makes me wonder why we have the same string rep for different types in the first place, but good to fix.    Really, the telescope type string was supposed to have three parts: 
- a high-level type name: LST/MST/SST
- a optics name
- a camera name
When anything changes in the description, the optics or camera name should also be changed, i,e by adding a version, but we don't enforce that.  The optics name should *not* in general be just "LST" (which is redundant), but rather something version-able or that describes better the current optics design.  I would expect in the future we could have e.g.  `1: LST_v1_LSTCam`, `2,3,4: LST_v2_LSTCam`, but that needs to be done at the simualtion production level when setting the metadata.

morcuended (approved):

PR #2688, opened 2025-01-21
Make sure config is fully serialized to provenance json

maxnoe:


maxnoe:
With the changes made here, we are no down to the quantity.

I can test if storing the string works now.

Concerning the enums, it might be better to not rely on the implicit `IntEnum` or `StrEnum` but add an explicit case that stores `<Enum>.name` into the json file.

kosack:
> can test if storing the string works now.

What is allowed if these are specified on the command-line or in a YAML config? I guess it's just the string?  There would in any case only be two options: a single string `"option"="12 TeV"` or separating it into value and unit, e.g. `"option"={value="12",unit="TeV"}`

kosack (approved):

src/ctapipe/core/tests/test_tool.py
```diff
@@ -477,15 +479,33 @@ def start(self):
...
+
+    tool = MyTool()
+    config = {
+        "MyTool": {
+            "value": 10.0,
+            "path": str(tmp_path / "foo.txt"),
+            "time": "2025-01-01T00:00:00",
+            "Comp": {
+                "quantity": {"value": 5.0, "unit": "m"},
+            },
```
mexanick, line 504: let's add also a TelescopeParameter to test, to make sure that we can correctly serialize UserList of the AstroQuantities

maxnoe: done

mexanick (dismissed):

mexanick (approved):

kosack (approved):

PR #2696, opened 2025-02-12
Store subarray description in the output file of the stats tool

TjarkMiener:
We noticed that we would like to run an additional custom `OutlierDetectors` based on the deviation from the expected standard deviation of the number of photoelectrons in the `calibpipe ffactor tool` (see e.g. [calibpipe-MR](https://gitlab.cta-observatory.org/cta-computing/dpps/calibrationpipeline/calibpipe/-/blob/cb6bebf1f42f4c0253c64e0253ff90bb87a25367/src/calibpipe/tools/ffactor_calibrator.py#L28)). Therefore we need the `SubarrayDescription` in the output file of the `stats tool` since the `OutlierDetectors` are `TelescopeComponents`.

### DIFF
--- a/src/ctapipe/tools/calculate_pixel_stats.py
+++ b/src/ctapipe/tools/calculate_pixel_stats.py
@@ -99,14 +99,14 @@ class PixelStatisticsCalculatorTool(Tool):
                 )
         self.input_data.dl1_images = True
         # Load the subarray description from the input file
-        subarray = SubarrayDescription.from_hdf(self.input_data.input_url)
+        self.subarray = SubarrayDescription.from_hdf(self.input_data.input_url)
         # Get the telescope ids from the input data or use the allowed_tels configuration
         self.tel_ids = (
-            subarray.tel_ids if self.allowed_tels is None else self.allowed_tels
+            self.subarray.tel_ids if self.allowed_tels is None else self.allowed_tels
         )
         # Initialization of the statistics calculator
         self.stats_calculator = PixelStatisticsCalculator(
-            parent=self, subarray=subarray
+            parent=self, subarray=self.subarray
         )
 
     def start(self):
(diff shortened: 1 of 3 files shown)


src/ctapipe/tools/calculate_pixel_stats.py
```diff
@@ -100,18 +100,20 @@ def setup(self):
...
         # Load the subarray description from the input file
         subarray = SubarrayDescription.from_hdf(self.input_data.input_url)
-        # Get the telescope ids from the input data or use the allowed_tels configuration
-        self.tel_ids = (
-            subarray.tel_ids if self.allowed_tels is None else self.allowed_tels
+        # Select a new subarray if the allowed_tels configuration is used
+        self.subarray = (
+            subarray
+            if self.allowed_tels is None
+            else subarray.select_subarray(self.allowed_tels, "Subarray")
```
maxnoe, line 107: I'd leave the default name here, it will create a somewhat nice description of which telescopes were selected from which subarray.

src/ctapipe/tools/calculate_pixel_stats.py
```diff
@@ -171,13 +173,21 @@ def start(self):
...
+
+    def finish(self):
+        # Store the subarray description in the output file
+        self.subarray.to_hdf(self.output_path, overwrite=self.overwrite)
+        self.log.info(
+            "Subarray description was stored in '%s'",
+            self.output_path,
+        )
+        # Close the file in the TableLoader
+        self.input_data.close()
```
maxnoe, line 190: Better use `self.input_data = self.enter_context(TableLoader(...))` when opening.

The tool as an `ExitStack` to manage contexts.

maxnoe (approved):

maxnoe (changes requested):
Please add a test (or a line to the existing one) that reads the sub array back

maxnoe (dismissed):

maxnoe (approved):

PR #2707, opened 2025-03-10
Store event times using CTAO high resolution timestamps

### LINKED ISSUE #2428: Support writing high-precision timestamps to hdf5

Currently, we write out all timestamps as TAI MJD.
That is one of the possible formats for the CTA "low precision" timestamp.
However, some things might still require the high precision timestamps, so we should have an option on the datawriter whether to use high or low precision timestamps for event times and similar things that are high precision in the input.

Add option to store timestamps that are high precision keeping this high precision.

maxnoe:
Fixes #2428

### DIFF
--- a/src/ctapipe/tests/test_time.py
+++ b/src/ctapipe/tests/test_time.py
@@ -13,7 +13,7 @@ roundtrip_times = [
 
 @pytest.mark.parametrize("timestamp", roundtrip_times)
 def test_cta_high_round_trip(timestamp):
-    from ctapipe_io_zfits.time import cta_high_res_to_time, time_to_cta_high_res
+    from ctapipe.time import cta_high_res_to_time, time_to_cta_high_res
 
     # note: precision=9 only affects text conversion, not actual precision
     time = Time(timestamp, scale="tai", precision=9)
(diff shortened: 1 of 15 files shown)


src/ctapipe/tests/test_time.py
```diff
@@ -0,0 +1,44 @@
...
+    roundtrip_error = (time - time_back).to_value(u.ns)
+    np.testing.assert_almost_equal(roundtrip_error, 0.0)
+
+
+test_data = [
+    (Time(0, 12.25e-9, format="unix_tai"), 0, 49),
+    (Time(12345, 12.25e-9, format="unix_tai"), 12345, 49),
+    (Time(65123, 123456.25e-9, format="unix_tai"), 65123, 493825),
+    (Time("2200-01-02T00:00:00"), np.iinfo(np.uint32).max, 0),
+    (NAN_TIME, 0, 0),
```
Tobychev, line 34: I think using an explicit hardcoded time for the underflow case is more clear and would save an import. 

Also, it will be more future proof for that that in 2036 when astropy actually adds a nan-time! ;)

maxnoe: This is the time we use as default value in containers.py, this is why it is checked here

kosack (approved):

mdpunch (approved):

PR #2709, opened 2025-03-11
Add a method to read scheduling information with TableLoader

mdebony:
In the class TableLoader, a method called read_scheduling_information was added to read scheduling block information from files.

### DIFF
--- a/src/ctapipe/io/tableloader.py
+++ b/src/ctapipe/io/tableloader.py
@@ -29,6 +29,7 @@ TRUE_IMPACT_GROUP = "/simulation/event/telescope/impact"
 SIMULATION_CONFIG_TABLE = "/configuration/simulation/run"
 SHOWER_DISTRIBUTION_TABLE = "/simulation/service/shower_distribution"
 OBSERVATION_TABLE = "/configuration/observation/observation_block"
+SCHEDULING_TABLE = "/configuration/observation/scheduling_block"
 FIXED_POINTING_GROUP = "/configuration/telescope/pointing"
 POINTING_GROUP = "/dl0/monitoring/telescope/pointing"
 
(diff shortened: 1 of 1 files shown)


maxnoe:
Thanks, changes on its own look good, but:

- Please add a test
- For the other tables, we offer to join this info to the events. I think it is not particularly useful for the information in the scheduling blocks though, so maybe we skip it

src/ctapipe/io/tableloader.py
```diff
@@ -363,6 +364,12 @@ def read_observation_information(self):
         """
         return read_table(self.h5file, OBSERVATION_TABLE)
 
+    def read_scheduling_information(self):
```
maxnoe, line 367: I know it's inconsistent with the `observation_info`, but I I really don't like `scheduling_information`, let's be specific and call it `read_scheduling_blocks`

src/ctapipe/io/tests/test_table_loader.py
```diff
@@ -369,6 +369,19 @@ def test_read_shower_distributions(dl2_merged_file):
         assert np.all(histograms["histogram"].sum(axis=(1, 2)) == [2000, 1000])
 
 
+def test_read_scheduling_information(dl2_merged_file):
+    from ctapipe.io import TableLoader
+
+    with TableLoader(dl2_merged_file) as table_loader:
+        scheduling = table_loader.read_scheduling_information()
+        assert len(scheduling) == 2
+        assert np.all(scheduling["sb_id"] == [4, 1])
```
maxnoe, line 378: For better error messages, use `np.testing.assert_array_equal(scheduling["sb_id"], [4, 1])`

src/ctapipe/io/tableloader.py
mdebony: Done

src/ctapipe/io/tests/test_table_loader.py
mdebony: Done

maxnoe (approved):
Thanks a lot

Tobychev (approved):

PR #2710, opened 2025-03-11
Fix sonar for PRs from forks

maxnoe:
Was failing as the PR information is not included in the github data, but we already included it before in the artifact, so I could just re-arrange the order a bit and use the env variable from the artifact.


maxnoe:
This is tested using the project here: https://github.com/cta-observatory/sonar-test/pull/5

It will only take effect once the PR is merged and the new work flow definition is on main

Tobychev (approved):

PR #2713, opened 2025-03-12
Fix checkout cleaning up coverage report in sonar ci

maxnoe:
The checkout action by default removes all untracked files, which in this case removed the coverage report


Tobychev (dismissed):

maxnoe:
Still not working, I am going back to testing in the test repo

maxnoe:
Ok, now it's working

PR #2715, opened 2025-03-17
Processing gain-selected dl1 data with the pixel stats tool

TjarkMiener:
We noticed that we can not process gain-selected data. This PR is a bug fix for processing gain-selected dl1 data with the pixel stats tool and its components. I extended the unit tests accordingly. Before the unit tests were passing since we did not include the detection of outliers in the pixel tool testing.

### DIFF
--- a/src/ctapipe/tools/calculate_pixel_stats.py
+++ b/src/ctapipe/tools/calculate_pixel_stats.py
@@ -82,6 +82,8 @@ class PixelStatisticsCalculatorTool(Tool):
         TableLoader,
     ] + classes_with_traits(PixelStatisticsCalculator)
 
+    DL1_COLUMN_NAMES = ["image", "peak_time"]
+
     def setup(self):
         # Read the input data with the 'TableLoader'
         self.input_data = self.enter_context(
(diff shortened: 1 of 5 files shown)


src/ctapipe/tools/calculate_pixel_stats.py
```diff
@@ -137,6 +139,13 @@ def start(self):
                     f"Column '{self.input_column_name}' not found "
                     f"in the input data for telescope 'tel_id={tel_id}'."
                 )
+            # Check if the dl1 data is gain selected and add an extra dimension for n_channels
+            for col_name in self.DL1_COLUMN_NAMES:
+                if (
+                    col_name in dl1_table.colnames
+                    and len(dl1_table[col_name].shape) == 2
```
maxnoe, line 146:
```suggestion
                    and dl1_table[col_name].ndim == 2
```

src/ctapipe/tools/tests/test_calculate_pixel_stats.py
```diff
@@ -51,13 +58,13 @@ def test_calculate_pixel_stats_tool(tmp_path, dl1_image_file):
...
-            monitoring_file,
-            path=f"/dl1/monitoring/telescope/statistics/tel_{tel_id:03d}",
-        )["mean"]
-        is not None
-    )
+    mean_column = read_table(
+        monitoring_file,
+        path=f"/dl1/monitoring/telescope/statistics/tel_{tel_id:03d}",
+    )["mean"]
+    assert mean_column is not None
```
maxnoe, line 65: this will never happen.

If `mean` doesn't exist, the line above raises an error that the column doesn't exist.

mexanick (approved):

maxnoe (approved):

PR #2717, opened 2025-03-18
Ignore `tables.NaturalNameWarning` in `write_table`s meta

Hckjs:
Get a lot of 
```
NaturalNameWarning: object name is not a valid Python identifier: 'OBSGEO-X'; it does not match the pattern ``^[a-zA-Z_][a-zA-Z0-9_]*$``; you will not be able to use natural naming to access this object; using ``getattr()`` will still work, though
  check_attribute_name(name)
```
Warnings flushing my logs from [astropy_helpers.write_table](https://github.com/cta-observatory/ctapipe/blob/e5999e2667ddc7253dd86703a46b497fc3ae7ae0/src/ctapipe/io/astropy_helpers.py#L116) while executing `ctapipe-apply-models`

### DIFF
--- a/src/ctapipe/io/astropy_helpers.py
+++ b/src/ctapipe/io/astropy_helpers.py
@@ -2,7 +2,9 @@
 """
 Functions to help adapt internal ctapipe data to astropy formats and conventions
 """
+
 import os
+import warnings
 from contextlib import ExitStack
 from uuid import uuid4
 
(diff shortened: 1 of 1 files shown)


maxnoe:
I think we shouldn't get this warning... where does it appear?

The hdf5 files should have these keywords here:
https://github.com/cta-observatory/ctapipe/blob/e5999e2667ddc7253dd86703a46b497fc3ae7ae0/src/ctapipe/instrument/subarray.py#L757-L762

`OBSGEO-{X,Y,Z}` is for FITS.

Hckjs:
It appears during `ctapipe-apply-models` when writing the tables here with the `astropy_helpers.write_table` function: 
https://github.com/cta-observatory/ctapipe/blob/e5999e2667ddc7253dd86703a46b497fc3ae7ae0/src/ctapipe/tools/apply_models.py#L209-L214

I think the "OBSGEO-..." metadata is added to the tables in the `TableLoader` with the `instrument=True` option:
https://github.com/cta-observatory/ctapipe/blob/e5999e2667ddc7253dd86703a46b497fc3ae7ae0/src/ctapipe/io/tableloader.py#L598-L601

maxnoe:
> It appears during ctapipe-apply-models when writing the tables here with the astropy_helpers.write_table function:

Aha! We should probably exclude metadata when joining the tables...

PR #2722, opened 2025-03-20
Adding `meta_convention` option to `SubarrayDescription.to_table()` method

### LINKED ISSUE #2721: Handle metadata when joining tables
This comes up in #2717 from some `tables.NaturalNameWarning` during `ctapipe-apply-models` when writing the tables with the `astropy_helpers.write_table` function: 
https://github.com/cta-observatory/ctapipe/blob/e5999e2667ddc7253dd86703a46b497fc3ae7ae0/src/ctapipe/tools/apply_models.py#L209-L214
and the "OBSGEO-..." metadata is added to the tables in the `TableLoader` with the `instrument=True` option:
https://github.com/cta-observatory/ctapipe/blob/e5999e2667ddc7253dd86703a46b497fc3ae7ae0/src/ctapipe/io/tableloader.py#L598-L601

Hckjs:
As Discussed in #2717  and #2721.

closes #2721

### DIFF
--- a/src/ctapipe/instrument/subarray.py
+++ b/src/ctapipe/instrument/subarray.py
@@ -266,7 +266,7 @@ class SubarrayDescription:
             which table to generate (subarray or optics)
         """
 
-        if meta_convention not in ["hdf", "fits"]:
+        if meta_convention not in {"hdf", "fits"}:
             raise ValueError(
                 f"meta_convention must be 'hdf' or 'fits', not {meta_convention}"
             )
--- a/src/ctapipe/instrument/subarray.py
+++ b/src/ctapipe/instrument/subarray.py
@@ -1,6 +1,7 @@
 """
 Description of Arrays or Subarrays of telescopes
 """
+
 import warnings
 from collections import defaultdict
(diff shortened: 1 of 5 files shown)


src/ctapipe/instrument/subarray.py
```diff
@@ -265,6 +266,11 @@ def to_table(self, kind="subarray"):
             which table to generate (subarray or optics)
         """
 
+        if meta_convention not in ["hdf", "fits"]:
```
maxnoe, line 269:
```suggestion
        if meta_convention not in {"hdf", "fits"}:
```

maxnoe (changes requested):
Minor comment, but looks good in general.

- Test should be added for the new option.
- I think the dump-instrument tool now should pass the FITS option in case of FITS output
- I still think we should not include metadata from the instrument table at all when joining info, so I'd still add the `instrument_table.meta.clear()` in the `TableLoader`.

maxnoe (approved):
Thanks!

kosack (approved):
Looks good to me.

PR #2724, opened 2025-03-24
Make StereoCombiner configurable in ApplyModels

### LINKED ISSUE #2720: Config values not applied to loaded Reconstructor models in `ApplyModels`
**Describe the bug**
When using the `ApplyModels` tool, configuration values set via the command line (e.g., `--EnergyRegressor.stereo_combiner.weights=intensity`) or given as input in a config file are not applied to the loaded Reconstructor instances.

I think this issue is caused by the way the `Reconstructor` models are loaded using 
[`Reconstructor.read(path, parent=self, subarray=self.loader.subarray)`](https://github.com/cta-observatory/ctapipe/blob/e5999e2667ddc7253dd86703a46b497fc3ae7ae0/src/ctapipe/tools/apply_models.py#L152).
Since these instances are deserialized from a joblib file, their attributes are not automatically updated with values from the `ApplyModels` configuration.

**To Reproduce**
```
pytest -s src/ctapipe/tools/tests/test_apply_models.py::test_apply_energy_regressor
```
which includes https://github.com/cta-observatory/ctapipe/blob/e5999e2667ddc7253dd86703a46b497fc3ae7ae0/src/ctapipe/tools/tests/test_apply_models.py#L34
while checking `combiner.weights` here:https://github.com/cta-observatory/ctapipe/blob/e5999e2667ddc7253dd86703a46b497fc3ae7ae0/src/ctapipe/tools/apply_models.py#L238
(weights are set to default value = "none" )

**Expected behavior**
Configuration values specified via CLI or config files should be properly applied to all components of a loaded `Reconstructor`.

Hckjs:
Since the loaded joblib-pickled ``Reconstructor``s in ``ApplyModels`` https://github.com/cta-observatory/ctapipe/blob/e5999e2667ddc7253dd86703a46b497fc3ae7ae0/src/ctapipe/tools/apply_models.py#L152 are already instantiated, one cannot change their traits via the config system, which leads to a not configurable ``StereoCombiner``.

Instantiating a new ``Reconstructor.from_name(parent=self)`` and handing over its ``StereoCombiner`` would solve this Problem.

Fixes #2720

### DIFF
--- a/src/ctapipe/tools/apply_models.py
+++ b/src/ctapipe/tools/apply_models.py
@@ -149,7 +149,20 @@ class ApplyModels(Tool):
 
         self._reconstructors = []
         for path in self.reconstructor_paths:
-            r = Reconstructor.read(path, parent=self, subarray=self.loader.subarray)
+            r = Reconstructor.read(
+                path,
+                parent=self,
+                subarray=self.loader.subarray,
+            )
+            # DispReconstructor hat norm_cls und sign_cls - keine model_cls
+            r.stereo_combiner = Reconstructor.from_name(
+                r.__class__.__name__,
+                subarray=self.loader.subarray,
+                parent=self,
+                prefix=r.prefix,
+                model_cls=r.model_cls,
+            ).stereo_combiner
+
             if self.n_jobs:
                 r.n_jobs = self.n_jobs
             self._reconstructors.append(r)
--- a/src/ctapipe/tools/apply_models.py
+++ b/src/ctapipe/tools/apply_models.py
@@ -154,13 +154,18 @@ class ApplyModels(Tool):
                 parent=self,
                 subarray=self.loader.subarray,
             )
-            # DispReconstructor hat norm_cls und sign_cls - keine model_cls
+
+            # Init new Reconstructor with config parameters and overwrite the StereoCombiner
+            model_keys = ["model_cls", "norm_cls", "sign_cls"]
+            model_kwargs = {
+                key: getattr(r, key) for key in model_keys if hasattr(r, key)
+            }
             r.stereo_combiner = Reconstructor.from_name(
                 r.__class__.__name__,
                 subarray=self.loader.subarray,
                 parent=self,
                 prefix=r.prefix,
-                model_cls=r.model_cls,
+                **model_kwargs,
             ).stereo_combiner
 
             if self.n_jobs:
--- a/src/ctapipe/tools/tests/test_apply_models.py
+++ b/src/ctapipe/tools/tests/test_apply_models.py
@@ -1,5 +1,6 @@
 import numpy as np
 import pytest
+from numpy.testing import assert_allclose
 
 from ctapipe.containers import (
     EventIndexContainer,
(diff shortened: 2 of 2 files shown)


maxnoe:
I think we should rather re-evaluate how we store the models on disk / read them back.

Pickling the whole `Reconstructor` instances is somewhat nasty. We could also think about excluding the `StereoCombiner` explicitly from the pickle or add an explicit configuration parameter for overriding the `StereoCombiner` per `Reconstructor` in `apply-models`.

Hckjs:
> I think we should rather re-evaluate how we store the models on disk / read them back.
> 
> Pickling the whole `Reconstructor` instances is somewhat nasty. We could also think about excluding the `StereoCombiner` explicitly from the pickle or add an explicit configuration parameter for overriding the `StereoCombiner` per `Reconstructor` in `apply-models`.

One could maybe `allow_none` for the `stereo_combiner_cls` traits in the Reconstructor subclasses and add a 
```
    def _init_stereo_combiner(self, parent=None, overwrite=False):
        if self.stereo_combiner is None or overwrite:
            self.stereo_combiner = StereoCombiner.from_name(
                self.stereo_combiner_cls,
                prefix=self.prefix,
                property=self.property,
                parent=parent,
            )
```
which is called on its `__init__`, making it possible to not store the `StereoCombiner` but instead init the Combiner with
```
r._init_stereo_combiner(parent=self, overwrite=self.overwrite_stereo_combiner)
```
in `ApplyModels`. This would also allow to overwrite it.

Can't really think of a better solution yet...

kosack:
I never liked that the whole reconstructor was pickled in the first place - it's easy, but also makes it very hard to customize how we store the models in the future, i.e. if we want to use some standard format.   Decoupling how we store the models from the Reconstructor would be the best way, but that also requires a bit of thought and a lot of refactoring.  E.g. how to know which reconstructor to construct before loading the model? How to store the parameter list?

One way would be to change what is actually serialized in `Reconstructor.write()` to be just the model and class name rather than the whole thing, and then in `load()` constructing a new class from the name and setting the model.  Perhaps that could be done by using `Reconstructor.__get_state__()`, but I'm not sure.

LukasBeiske:
> I never liked that the whole reconstructor was pickled in the first place - it's easy, but also makes it very hard to customize how we store the models in the future, i.e. if we want to use some standard format. Decoupling how we store the models from the Reconstructor would be the best way, but that also requires a bit of thought and a lot of refactoring. E.g. how to know which reconstructor to construct before loading the model? How to store the parameter list?
> 
> One way would be to change what is actually serialized in `Reconstructor.write()` to be just the model and class name rather than the whole thing, and then in `load()` constructing a new class from the name and setting the model. Perhaps that could be done by using `Reconstructor.__get_state__(), Reconstructr.__set_state__()` (which is the protocol for setting what gets pickled in a class), but I'm not sure.

How about, instead of pickling the whole `Reconstructor`, we pickle a dictionary containing the class name, the model(s), and all the other necessary configuration options.
Since some of these config options (e.g. `log_target` for regressors) should not be overwritten when applying the model, we would have to add checks for this when "loading" the model, but otherwise constructing a new class with the models and config options from that dictionary should not be a problem, right?

I don't know if this could be fleshed out and be made into some standard format, but it would be a start and solve this issue for now.

maxnoe:
The dict is good I think, like that we can also attach a `meta` filed where we put in the reference metadata.

LukasBeiske:
> The dict is good I think, like that we can also attach a `meta` filed where we put in the reference metadata.

Ok, I'll get on this and open a second PR, since this is a bit wider in scope than this PR.

maxnoe:
Maybe it's also time to give https://onnx.ai/sklearn-onnx/ a shot again and not rely on pickle.

PR #2731, opened 2025-04-02
Add a StereoDispCombiner

Hckjs:
This is basically the StereoCombiner already implemented [here](https://github.com/cta-observatory/magic-cta-pipe/blob/bb838453bf8333a20d1b5e1217b9ee345a11b45b/magicctapipe/scripts/lst1_magic/lst1_magic_dl1_stereo_to_dl2.py#L79)
in `magic-cta-pipe`.

It combines the mono DISP reconstruction by
 1) calculating combinations of 2 telescopes for every array event
 2) calculating the minimum distance for the 4 possible SIGN pairs for each combination
 3) calculating the weighted average for the minimum distance SIGN pair (fov lon/lat) per combination
 4) calculating the weighted average of all combinations for an array event with the summed weights from 3).

**Update**:
The `StereoDispCombiner` has essentially been generalized so that it can be configured to reproduce both the `magic-cta-pipe` and the `EventDisplay` behavior.

 - The number N of telescope combinations was introduced as a traitlet.
 - In addition, it is now possible to choose how many of the best telescopes, ranked by their weights, should participate in the averaging.
 - Furthermore, a check analogous to the one in EventDisplay was introduced to reject subarray events with two participating telescopes whose shower axes are nearly parallel.

For each telescope combination of size `n_tel_combinations`, all possible DISP sign assignments are then evaluated, and the sign combination that minimizes the Sum of Squared Errors (SSE) between the participating telescopes is selected. The resulting per-combination FoV positions are then combined using the selected telescope weights, and a weighted mean FoV direction is computed for each subarray event.

If a subarray event has an valid telescope multiplicity smaller than `n_tel_combinations` (after applying `n_best_tels` and any angular-difference cuts), but at least two telescopes remain, the reconstruction is performed using all available telescopes of that subarray event. In this case, only one combination is formed, with a size equal to the event multiplicity, and the optimal DISP sign assignment is determined accordingly. Single-telescope events are handled separately using just the mono reconstruction.

Some hints for reviewing are [here](https://github.com/cta-observatory/ctapipe/pull/2731#issuecomment-4183083703)


src/ctapipe/image/statistics.py
```diff
@@ -139,3 +140,15 @@ def arg_n_largest(n, array):
     dummy = np.zeros(n, dtype=np.int64)
     idx = arg_n_largest_gu(dummy, array)
     return idx
+
+
+@njit
+def argmin(array):
```
maxnoe, line 146: Is this way faster then using `np.argmin` in a numba jit function?

maxnoe: https://numba.pydata.org/numba-doc/dev/reference/numpysupported.html#calculation

argmin is listed as one of the supported functions

maxnoe: In my simple test, relying on the numpy func is faster:

```
In [13]: @njit()
    ...: def by_hand(array):
    ...:     min_index = 0
    ...:     min_value = array[0]
    ...:     for i in range(1, len(array)):
    ...:         if array[i] < min_value:
    ...:             min_value = array[i]
    ...:             min_index = i
    ...:     return min_index
    ...: 

In [14]: @njit()
    ...: def numpy_func(array):
    ...:     return np.argmin(array)
    ...: 

In [15]: numpy_func(vals)
Out[15]: 477

In [16]: vals = rng.normal(size=5000)

In [18]: %timeit by_hand(vals)
2.87 μs ± 113 ns per loop (mean ± std. dev. of 7 runs, 100,000 loops each)

In [19]: by_hand(vals)
Out[19]: 4178

In [20]: numpy_func(vals)
Out[20]: 4178

In [21]: %timeit by_hand(vals)
2.86 μs ± 119 ns per loop (mean ± std. dev. of 7 runs, 100,000 loops each)

In [22]: %timeit numpy_func(vals)
2.3 μs ± 85.4 ns per loop (mean ± std. dev. of 7 runs, 100,000 loops each)
```

src/ctapipe/reco/stereo_combination.py
```diff
@@ -405,3 +415,251 @@ def predict_table(self, mono_predictions: Table) -> Table:
         stereo_table[f"{self.prefix}_telescopes"] = tel_ids
         add_defaults_and_meta(stereo_table, _containers[self.property], self.prefix)
         return stereo_table
+
+
+class StereoDispCombiner(StereoCombiner):
```
maxnoe, line 420: How does this algorithm differ from the implementation in EventDisplay by @GernotMaier ?

src/ctapipe/reco/stereo_combination.py
```diff
@@ -405,3 +415,251 @@ def predict_table(self, mono_predictions: Table) -> Table:
...
+            raise NotImplementedError(
+                f"Combination of {self.property} not implemented in {self.__class__.__name__}"
+            )
+
+    def _calculate_weights(self, data):
+        if isinstance(data, Container):
+            if self.weights == "intensity":
+                return data.hillas.intensity
+
+            if self.weights == "konrad":
```
kosack, line 445: I think I've made this comment before: can we pick a better name here, since calling it "konrad" is not very helpful?  This weighting is  _intensity_  x _aspect-ratio_, so we could call it "intensity-aspect-ratio" or "aspect-weighted-intensity" or maybe there is a shorter word...

kosack (commented):
We have a predicted `disp.sign_score` that is not used in this method, which seems odd.  Is that the reason why the pair-wise average including both disps is made rather than a simple weighted average over predictions?  Is there citation for the MARS  method, by the way? i'd be interested to see if simpler methods were tested.

src/ctapipe/reco/stereo_combination.py
Hckjs: > How does this algorithm differ from the implementation in EventDisplay by @GernotMaier ?

In general theses algorithms are very similar - ED instead calculates the minimum distance for all permutations of the 5 best (weighted) telescope events. E.g. for an Event with 8 valid telescopes:
 - Taking the 5 brightest telescope events (or any other weighting)
 - Generate all permutations of both signs for these 5 telescope events
 - Calculate distance for each permuation
 - Average permutation with minimum distance

src/ctapipe/reco/stereo_combination.py
```diff
@@ -405,3 +415,251 @@ def predict_table(self, mono_predictions: Table) -> Table:
...
+        fov_lon_values = []
+        fov_lat_values = []
+        weights = []
+        signs = np.array([-1, 1])
+        n_tel_combinations = 2
+
+        for tel_id, dl2 in event.dl2.tel.items():
+            if dl2.geometry[self.prefix].is_valid:
+                dl1 = event.dl1.tel[tel_id].parameters
+                hillas_fov_lon = dl1.hillas.fov_lon.to_value(u.deg)
```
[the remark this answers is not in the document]
Hckjs: I have to get rid of the units either way because of the njit'ed function later on. And since i had to either add or remove the unit when calculating the fov lon/lat (because of the "+") 
`hillas_fov_lon + signs * disp * np.cos(hillas_psi)`
I thought it might be easiest to remove it right at the beginning and add it back in the end

src/ctapipe/reco/stereo_combination.py
maxnoe: Could we enable using the EventDisplay definition with come configuration options? E.g. a `max_telescopes` and `telescope_priority_criterion` or similar? This would be great for performance studies to directly compare to EventDisplay

Hckjs:
> > We have a predicted disp.sign_score that is not used in this method, which seems odd.
> 
> Not really, this is more of a book-keeping for doing lower-level performance plots. It _could_ be used for a quality query or similar, but AFAIK, this was not yet done anywhere and would be a completely new development that needs to be tested. Would be a nice study though.

During the quantitative analysis of the `StereoDispCombiner` in its current form, i noticed that particularly high-energy events with a high scatter radius have an almost parallel reconstructed shower axis (often events with only 2 valid telescopes). This worsens performance compared to the `StereoMeanCombiner` at high energies because it isnt just averaging but choosing the complete wrong side. However, since this occurs mostly at high energies, the sign scores are also significantly higher, which I tested as a weighting factor in the selection of the minimum distance and was able to achieve significantly better performance for high-energy events. So far, I have only tested this on small data sets, but I will upload a few comparison plots soon.

src/ctapipe/reco/stereo_combination.py
Hckjs: In general yes, but this would need some severe refactoring of this PR at least for the table-wise processing. We have a Masters Student now (@AkramAki) who is working on that and would implement it (including maybe also a KMeans and DBScan based algorithm) the next months/next year. I would still suggest to merge the Combiner in the current form and do the refactoring and additional algorithms in separate PRs later on.

Hckjs:
**Algorithm:**
The phrase of "weighting" the distances was maybe a bit misleading. The idea was, that the algorithm forces to choose the direction predicted by the sign_classifier if the sign_score is above a certain limit (this limit usually should be very high / near 1). 
The calculated distance is just used to choose 1 of the 4 possible combinations you can get for combining two telescopes with each of them having + and - SIGN. What the current implementation does:
 - if you set a sign_score_limit, calculate a "weight" for each SIGN per telescope:
     - if sign_score > sign_score_limit: `(1 / (1 + sign_score))` for the choosen SIGN of the sign_classifier
     - else 1
- If none of the telescope events have a sign_score above the limit - nothing will happen
- if it comes now to calculate the 4 distances per telescope pair **and one of the have a sign_score above the limit**:
    - combine the weights for each of the 4 combinations by multiplying them together
    - each of the 4 distances has now a combined weight - either 1 for SIGN combinations **not** choosen bei a sign classifier -- or something < 1
    - it forces now to just consider the distances of combinations, where at least one of the signs were predicted by the sign classifier by setting the distance of a combination with combined weight of 1 to inf


Anything else stays the same - its just picking the sign predicted by the sign classifier with a very good sign_score. The weighting is not used as additional weight when averaging lon/lat.

**Dataset:**
- PROD6 MC 20/0 deg zen/az northern alpha array fully processed with ctapipe up to IRFs
- I used mainly ctapipe defaults configs and cuts - except lstchains image cleaning (without the nsb-part obviously)
- pointlike IRFs at 0.7deg offset

The following plots are generated on pointlike gammas with a `sign_score_limit = 0.99` and `multiplicity >= 2`:

 - Distribution of all Array-Events:
    number_events
<br>

 - Percentage of Array-Events in each energy-bin which have at least one valid telescope with a sign_score above the limit:
    events_above_ss_limit
<br>

 - How much (percentage) of these events (at least 1 tel above sign_score_limit) got a better/worse/same reconstruction (measured as $\theta^{2}$) with the additional sign_score option:
    theta2_comp_ss_limit_0 99
<br>

- Now just have a look at all the events, where $\theta^{2}$ have changed (better & worse)
    lon_lat_all
<br>

- Where it only gets better with the used sign_score_limit:
    lon_lat_better
<br>

- Where it only gets worse with the used sign_score_limit:
    lon_lat_worse
<br>

In Absolute Values:
 - Number of Events: 1534777
 - Number of Events with at least one telescope above sign_score_limit 0.99: 531576
 - Number of Events that got better reconstruction: 12856
 - Number of Events that got worse reconstruction: 7524

 - Mean $\theta^{2}$ improvement per event: 0.236584
 - Mean $\theta^{2}$ worsening per event: 0.017842
 - `mean_imp/mean_worse = 13.25981`


- 95% Ang Res Comparison:
    [ang_res_95_limit_0.99.pdf](https://github.com/user-attachments/files/23881904/ang_res_95_limit_0.99.pdf)
<br>

**Notes:**
- The used limit of 0.99 looks quite high - but the overall performance (if comparing `mean_imp/mean_worse`) is dropping fast below a limit of 0.98.
- It mostly affects events around a few hundred GeVs, but in general just a small percentage (~ 1%). 
- The overall gain per event (`mean_imp/mean_worse`) can get quite high by using a limit near 1.
- At least for the event-wise processing using the sign_score is about 1% slower (depends on the limit).

Although the use of sign scores only has an impact on very few events, the improvement per event is quite significant and requires only slightly more computing time. Since it can be optionally switched on and off, I think it's worth leaving it in.

src/ctapipe/reco/stereo_combination.py
```diff
@@ -74,23 +76,18 @@ class StereoMeanCombiner(StereoCombiner):
...
-        self.supported = {
-            ReconstructionProperty.ENERGY,
-            ReconstructionProperty.GEOMETRY,
-            ReconstructionProperty.PARTICLE_TYPE,
-        }
-        if self.property not in self.supported:
-            raise NotImplementedError(
-                f"Combination of {self.property} not implemented in {self.__class__.__name__}"
-            )
+    property = UseEnum(
```
maxnoe: Does it make sense to make this configurable? Shouldn't it just be `GEOMETRY`?

src/ctapipe/reco/stereo_combination.py
```diff
@@ -404,3 +478,313 @@ def predict_table(self, mono_predictions: Table) -> Table:
...
+        ids = []
+        fov_lon_values = []
+        fov_lat_values = []
+        weights = []
+        dist_weights = []
+
+        signs = np.array([-1, 1])
+
+        for tel_id, dl2 in event.dl2.tel.items():
+            if dl2.geometry[self.prefix].is_valid:
```
maxnoe: Shouldn't this check `disp` and not `geometry`? The combiner is the thing that fills geometry, right?

src/ctapipe/reco/stereo_combination.py
```diff
@@ -404,3 +478,313 @@ def predict_table(self, mono_predictions: Table) -> Table:
...
+        Perform DISP-based stereo direction reconstruction for a single event.
+
+        This is the entry point used for event-wise processing.
+        Calls :meth:`_combine_altaz` and stores the resulting stereo geometry
+        inside ``event.dl2.stereo`` under the configured prefix.
+        """
+
+        self._combine_altaz(event)
+
+    def predict_table(self, mono_predictions: Table) -> Table:
```
maxnoe: How much slower, if at all, is using `predict_table` for a single subarray event?

I'm asking because it would be nice if we could just use `predict_table` also for the `__call__` to reduce the code duplication and chance for behavorial differences between the container and the table case.

src/ctapipe/reco/stereo_combination.py
```diff
@@ -404,3 +478,313 @@ def predict_table(self, mono_predictions: Table) -> Table:
...
+    def _combine_altaz(self, event):
+        ids = []
+        fov_lon_values = []
+        fov_lat_values = []
+        weights = []
+        dist_weights = []
+
+        signs = np.array([-1, 1])
+
+        for tel_id, dl2 in event.dl2.tel.items():
```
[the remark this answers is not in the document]
Hckjs: Yes, ill add it. Would be also very nice to have it as a TelescopeComponent here #2158 (and additionally a table quality query for the table-wise processing)

src/ctapipe/reco/stereo_combination.py
Hckjs: In principle, yes - it should just be `GEOMETRY`, but the `StereoCombiner` is mainly instantiated via `.from_name()` in the `Reconstructor`-subclasses and checked by the `property`-trait to ensure that a valid `StereoCombiner` is selected for each `Reconstructor`, right? In that case, I definitely think it makes sense to prevent users from selecting the wrong `StereoCombiner` in the config and raising a proper error.
Moreover, even if `GEOMETRY` is the only reconstructable property, I don't think it's such a bad idea to require users to explicitly specify the property they want to reconstruct - even if they init the `StereoCombiner` directly, independently of a `Reconstructor`. It prevents misconfigurations early on.

Hckjs:
### Update:

- The algorithm has been generalized by `n_tel_combinations`, allowing the EventDisplay behavior to be reproduced.
    - A configurable `min_ang_diff` cut was added to reject events with nearly parallel shower axes (for multiplicity = 2).
    - A configurable `n_best_tels` trait was added to limit the stereo combination to the N telescopes with the highest weights

### Open points for discussion

- I set the `property` trait of the `StereoDispCombiner` to `config=False`, since it is not meant to be user-configurable (as discussed). At the same time, `.from_name()` in the Reconstructor subclasses still works and raises a proper error if, for example, a user assigns the `StereoDispCombiner` to the `EnergyReconstructor` via the config-system. I am not fully sure whether this is in line with the intended architecture, as we currently do not have other non-configurable traits.
- For event-wise processing, I integrated the `StereoQualityQuery`. I am unsure whether it makes sense to define an additional QualityQuery for table-wise processing as well; if so, we may need to revisit the naming.
- Since the `is_valid` field was removed from the DISP container, I added an explicit double check:

  ```python
  if not dl2.geometry[self.prefix].is_valid:
      continue
  if self.prefix not in dl2.disp:
      raise RuntimeError(
          f"No valid DISP reconstruction parameter found for "
          f"prefix='{self.prefix}'. "
          f"Make sure to apply the DispReconstructor before using the "
          f"StereoDispCombiner or adapt the prefix accordingly."
      )

Hckjs:
The `StereoDispCombiner` table-wise implementation in `predict_table()` became somewhat more complex through this generalization, so I wanted to add a few notes on the reasoning behind the implementation to make the review a bit easier. The event-wise processing was comparatively straightforward, so I leave that aside here.

With the introduction of the `n_tel_combinations` trait, one difficulty in the broadcasting-based table-wise implementation was that the dimension of the combinations becomes variable for subarray events with multiplicities smaller than `n_tel_combinations`. I considered several approaches for handling this:

- **Variable-length arrays:** in principle possible, but since we do not have a direct dependency on packages that support simple broadcasting and vectorized operations on variable-length arrays, I wanted to avoid this route.
- **Masked arrays + padding:** also possible in principle. I did look into this approach and thought about what the final implementation would need to look like, but I came to the conclusion that it would make the already fairly complex code significantly harder to maintain.
- **A small for-loop over multiplicities < `n_tel_combinations`:** this is the approach I ultimately chose. It allows me to reuse parts of the event-wise processing, keeps the overall code more maintainable, and since `n_tel_combinations` is not expected to be chosen very large anyway, the loop remains small. In practice, values larger than about 5 are not really advisable because the computation time increases rapidly. For example, with `n_tel_combinations = 5`, the loop only has to cover multiplicities 4, 3, and 2, so the additional overhead remains manageable.

A few additional notes on `_compute_altaz_for_valid()` for easier understanding:

- The main idea was to construct an array that allows me to compute the minimum SSE for all combinations via broadcasting. For this, I use the `index_tel_combs` array, which essentially contains all telescope combinations in sequence, with telescope indices of the input table as elements.
- To make this work, I needed a few helper functions for the computation itself, as well as mappings to assign the results back to the corresponding subarray events, most notably `combs_to_array_indices`.
- Beforehand, I generate a `combs_array`, which precomputes all `k`-combination patterns up to `max_multiplicity`. This is then later used with the actual `tel_indices` to construct the final `index_tel_combs` array.
- As mentioned above, this is only done for multiplicities `>= n_tel_combinations`. Afterwards, I loop over all smaller multiplicities, since in those cases each event has only a single combination with size equal to its multiplicity.
- Finally, single-telescope events are simply filled with the mono predictions.

Hope that helps a bit

maxnoe:
> Variable-length arrays: in principle possible, but since we do not have a direct dependency on packages that support simple broadcasting and vectorized operations on variable-length arrays, I wanted to avoid this route.

I'm not sure that this is the right approach here, but we already do something like this in the number of islands code, where we use the internal representation of a scipy csr sparse matrix, which essentially is an array of variable length arrays.

maxnoe:
You could write a numba function that takes a single 1d array of all disp values and an array of multiplicities per stereo event. You can then index into the 1d array of disp values.

> is the standard CSR representation where the column indices for row i are stored in indices[indptr[i]:indptr[i+1]] and their corresponding values are stored in data[indptr[i]:indptr[i+1]]. If the shape parameter is not supplied, the array dimensions are inferred from the index arrays.

https://docs.scipy.org/doc/scipy/reference/generated/scipy.sparse.csr_array.html

src/ctapipe/reco/stereo_combination.py
```diff
@@ -404,3 +519,525 @@ def predict_table(self, mono_predictions: Table) -> Table:
...
+        if multiplicity == 0:
+            return np.nan, np.nan, False
+
+        if multiplicity == 1:
+            stereo_fov_lon = fov_lon_values[0][1]
+            stereo_fov_lat = fov_lat_values[0][1]
+            return stereo_fov_lon, stereo_fov_lat, True
+
+        if (
+            multiplicity == 2
```
maxnoe, line 755: Why is two a special case here? Don't we need to check `min_ang_diff` for each pair and check if we got any pairs left after that check?

Hckjs:
> Why is two a special case here? Don't we need to check `min_ang_diff` for each pair and check if we got any pairs left after that check?

In a first quick check, it looked like the largest performance losses related to near-parallel main shower axes occur mainly for multiplicity n=2. At the same time, I had found an old comment in the EventDisplay code suggesting that this was also only checked for multiplicity 2, although I realized now that this comment is outdated.

After looking through the EventDisplay implementation more carefully, I think you are probably right that it makes more sense to apply this to all events. In EventDisplay, the code computes the mean angle over all pairwise telescope-axis combinations and then applies a cut on this averaged angle. So in that implementation the full event is either kept or rejected; it does not just remove the individual 2-telescope combinations with small angular separation.

I also noticed that EventDisplay uses a more L1-like minimum-distance criterion, i.e. a weighted mean pairwise distance, rather than the SSE that is currently implemented in this PR. The SSE penalizes outliers more strongly, which can be beneficial when the correct sign assignment leads to a tight cluster of telescope-wise solutions. On the other hand, the Euclidean distance criterion is probably somewhat more robust against poorly reconstructed main shower axes.

In addition, for multiplicities m>5, EventDisplay uses the Hillas-intersection result as a starting point and then chooses for each telescope the DISP sign whose solution lies closest to that Hillas-intersection estimate. This also seems broadly consistent with the results from Lukas Nickel’s Master’s thesis, where the geometric/Hillas reconstruction performs better at higher multiplicities.

It also looks like EventDisplay does not apply any explicit `n_best_telescopes` selection, at least not in the DISP combination logic I checked.

To reproduce the EventDisplay behavior more exactly for benchmarking studies, I would otherwise like to implement these three points:
 - Averaged angle cut
 - weighted mean pairwise distance instead of SSE (make it optional maybe for performance testing)
 - Make use of HillesReconstructor estimate on m > 5
 
 I am just not sure how quickly I will manage that over the next days or weeks, since I am currently finishing my thesis.
 
 

> You could write a numba function that takes a single 1d array of all disp values and an array of multiplicities per stereo event. You can then index into the 1d array of disp values.
> 
> > is the standard CSR representation where the column indices for row i are stored in indices[indptr[i]:indptr[i+1]] and their corresponding values are stored in data[indptr[i]:indptr[i+1]]. If the shape parameter is not supplied, the array dimensions are inferred from the index arrays.
> 
> https://docs.scipy.org/doc/scipy/reference/generated/scipy.sparse.csr_array.html

That looks like a good alternative. I am just not sure yet whether, and by how much, it would actually be faster in practice. I would need to test that first.

Thanks for the comments!

src/ctapipe/reco/stereo_combination.py
```diff
@@ -404,3 +519,525 @@ def predict_table(self, mono_predictions: Table) -> Table:
...
+
+    If an subarray event has an effective telescope multiplicity smaller than
+    ``n_tel_combinations`` (after applying ``n_best_tels`` and any angular
+    difference cuts), but at least two telescopes remain, the reconstruction
+    is performed using all available telescopes of that subevent. In this case,
+    just one combination is formed with a size equal to the event multiplicity
+    and the optimal DISP sign assignment is determined accordingly. Single-
+    telescope events are handled separately using the mono reconstruction.
+
+    **Traits**
```
maxnoe, line 601: This is not needed, we include traitlet help strings automatically.

src/ctapipe/reco/stereo_combination.py
```diff
@@ -404,3 +519,525 @@ def predict_table(self, mono_predictions: Table) -> Table:
...
+        default_value=None,
+        min=2,
+        allow_none=True,
+        help=(
+            "Select the best n_best_tels telescopes by weight for the combination. "
+            "Set to None to use all valid telescopes."
+        ),
+    ).tag(config=True)
+
+    min_ang_diff = Float(
```
maxnoe, line 644: This could be a Quantity traitlet

src/ctapipe/reco/stereo_combination.py
```diff
@@ -404,3 +519,525 @@ def predict_table(self, mono_predictions: Table) -> Table:
...
+            "Minimum angular separation (in degrees) between the reconstructed main "
+            "shower axes of two telescopes. This cut is applied only to subarray events "
+            "with exactly two participating telescopes (multiplicity = 2). Events for "
+            "which the angular difference between the two shower axes is smaller than "
+            "this value are rejected, as nearly parallel axes lead to problems solving "
+            "the head-tail ambiguity. Set to None to disable this cut."
+        ),
+    ).tag(config=True)
+
+    property = UseEnum(
```
maxnoe, line 658: This shouldn't be configurable, just a class variable: `property = ReconstructionProperty.GEOMETRY`

src/ctapipe/reco/stereo_combination.py
```diff
@@ -169,11 +268,17 @@ def _combine_classification(self, event):
...
+            dl1_params = event.dl1.tel[tel_id].parameters
+            quality_checks = self.quality_query(parameters=dl1_params)
+            if not mono.is_valid:
+                continue
+            if not all(quality_checks):
+                continue
+
+            values.append(mono.prediction)
+            dl1 = event.dl1.tel[tel_id].parameters
+            weights.append(self._calculate_weights(dl1) if dl1 else 1)
```
maxnoe, line 305: If I read this correctly, we here silently ignore the case where a user asked the disp computation to be weighted but the data does not contain the necessary parameters.

I think this should not be silent.

src/ctapipe/reco/stereo_combination.py
```diff
@@ -195,12 +300,18 @@ def _combine_altaz(self, event):
...
             mono = dl2.geometry[self.prefix]
-            if mono.is_valid:
-                alt_values.append(mono.alt)
-                az_values.append(mono.az)
-                dl1 = event.dl1.tel[tel_id].parameters
-                weights.append(self._calculate_weights(dl1) if dl1 else 1)
-                ids.append(tel_id)
+            dl1_params = event.dl1.tel[tel_id].parameters
+            quality_checks = self.quality_query(parameters=dl1_params)
+            if not mono.is_valid:
```
maxnoe, line 330: This check could be done before evaluating the quality query for performance

src/ctapipe/reco/stereo_combination.py
```diff
@@ -404,3 +519,525 @@ def predict_table(self, mono_predictions: Table) -> Table:
...
+        resulting stereo geometry inside ``event.dl2.stereo`` under the
+        configured prefix.
+        """
+
+        ids, hillas_psis, fov_lon_values, fov_lat_values, weights = (
+            self._collect_valid_tel_data(event)
+        )
+
+        multiplicity = len(ids)
+        alt = az = u.Quantity(np.nan, u.deg, copy=COPY_IF_NEEDED)
```
maxnoe, line 695: `np.nan` is an immutable float, it will always be copied, so no need to add `COPY_IF_NEEDED`.

maxnoe: you can see this also when passing `copy=False`, it will always raise an error that the copy could not be avoided.

maxnoe: I'd also move this line to an `else` below, so that it is only called when needed.

src/ctapipe/reco/stereo_combination.py
```diff
@@ -404,3 +519,525 @@ def predict_table(self, mono_predictions: Table) -> Table:
...
+        stereo_fov_lon, stereo_fov_lat, valid = self._compute_stereo_fov(
+            multiplicity=multiplicity,
+            hillas_psis=hillas_psis,
+            fov_lon_values=fov_lon_values,
+            fov_lat_values=fov_lat_values,
+            weights=weights,
+        )
+
+        if valid:
+            alt, az = telescope_to_horizontal(
```
maxnoe, line 706: This could be a separate issue, but I'm not sure that the code here would be correct for divergent pointing.

I think there, we would need to transform using individual `TelescopeFrame` per telescope with the telescope pointing to a common `NominalFrame` with the array pointing, combine, and then transform to altaz.

Or transform individually to AltAz and then combine in AltAz space.

src/ctapipe/reco/stereo_combination.py
```diff
@@ -404,3 +519,525 @@ def predict_table(self, mono_predictions: Table) -> Table:
...
+            dl1 = event.dl1.tel[tel_id].parameters
+            if not all(self.quality_query(parameters=dl1)):
+                continue
+
+            hillas_fov_lon = dl1.hillas.fov_lon.to_value(u.deg)
+            hillas_fov_lat = dl1.hillas.fov_lat.to_value(u.deg)
+            hillas_psi = dl1.hillas.psi
+            disp = disp_reco.parameter.to_value(u.deg)
+
+            fov_lons = hillas_fov_lon + signs * disp * np.cos(hillas_psi)
```
maxnoe, line 756: for performance, it would probably be better to create arrays of length `n_telescopes` before the loop, fill in the values and then vectorize the computations after the loop?

maxnoe: Similarly how it is done here for the `HillasReconstructor`: https://github.com/cta-observatory/ctapipe/blob/4f49f2c339ba7213408ba2f842fbee1978250243/src/ctapipe/reco/hillas_reconstructor.py#L219

src/ctapipe/reco/stereo_combination.py
```diff
@@ -404,3 +519,525 @@ def predict_table(self, mono_predictions: Table) -> Table:
...
+            return valid
+
+        weights = self._calculate_weights(mono_predictions[valid])
+        valid_idx = np.flatnonzero(valid)
+        valid_tel_to_array_indices = tel_to_array_indices[valid]
+
+        if valid_tel_to_array_indices.size == 0:
+            return valid
+
+        order = np.lexsort((-np.asarray(weights), valid_tel_to_array_indices))
```
maxnoe, line 982: I think these parts deserve some comments. It is not obvious to me what e.g. lexsort is achieving here

src/ctapipe/reco/stereo_combination.py
LukasBeiske: I've seen some sensitivity curves for divergent pointing using this code here and there indeed seems to be a problem. 
If we don't care about the `angular_distance_uncert` calculated in `StereoMeanCombiner` atm (which will eventually be replaced by the reconstruction quality event type anyway), transforming into AltAz before combining would probably be the easiest solution (using e.g. `scipy.stats.circmean` for azimuth).

LukasBeiske: @AkramAki fixes this in #3076

PR #2750, opened 2025-05-12
Remove one level of indirection from DVR classes

maxnoe:


kosack (approved):

kosack:
Looks good, just need to include the tel_id and selected_gain_channel in the tests and other places where it is used.

Hckjs (dismissed):

maxnoe:
> One could separate the actual reduction algorithm in the __call__-function to its own function and call it in the Reducers __call__-function to be more coherent with the ImageCleaners. That would make it also possible to call that algorithms by its own.

For the image cleaners, the signature between `__call__` and the lower-level functions is different. Here it is the same.

PR #2736, opened 2025-04-09
Muon fit improvement

burmist-git:
The muon ring fitter needs to be improved.
The existing toy tests for muons are overly simplified.
Realistic simulations reveal significant discrepancies between the true ring and the fitted result (see the attached PNG). The fitted parameters are shown in green.

Screenshot from 2025-04-09 16-55-44
  
The second major component will be fitting the muon impact point while accounting for the muon’s inclination.

I stated with :
Enhance the toy model for muon ring simulation by incorporating user-defined asymmetry in both the x and y directions.
Update the muon ring fitter test to include this ring asymmetry.

PR #2752, opened 2025-05-13
Use telescope frame for muon test instead of manual transform

maxnoe:
@burmist-git Another PR targeting yours to use the TelescopeFrame instead of the manual transformation you added.


maxnoe:
@burmist-git Merging this PR will fix the CI also in #2736

burmist-git:
All is good !!! Thanks learned something new !!

PR #2753, opened 2025-05-14
Add option to subtract pedestal baseline for MCs

TjarkMiener:
For MC calib events we want to skip this pedestal subtraction.

### DIFF
--- a/src/ctapipe/io/simteleventsource.py
+++ b/src/ctapipe/io/simteleventsource.py
@@ -546,6 +546,14 @@ class SimTelEventSource(EventSource):
         ),
     ).tag(config=True)
 
+    subtract_pedestal_baseline = Bool(
+        True,
+        help=(
+            "Subtract the pedestal baseline from the waveform samples."
+            " This is done by default, but can be disabled for calibration events."
+        ),
+    ).tag(config=True)
+
     atmosphere_profile_choice = UseEnum(
         AtmosphereProfileKind,
         default_value=AtmosphereProfileKind.AUTO,
@@ -997,7 +1005,10 @@ class SimTelEventSource(EventSource):
                 data.r0.tel[tel_id] = R0CameraContainer(waveform=adc_samples)
 
                 cam_mon = array_event["camera_monitorings"][tel_id]
-                pedestal = cam_mon["pedestal"] / cam_mon["n_ped_slices"]
+                if self.subtract_pedestal_baseline:
+                    pedestal = cam_mon["pedestal"] / cam_mon["n_ped_slices"]
+                else:
+                    pedestal = np.zeros_like(cam_mon["pedestal"])
                 dc_to_pe = array_event["laser_calibrations"][tel_id]["calib"]
 
                 # fill dc_to_pe and pedestal_per_sample info into monitoring


maxnoe:
The changes here look fine, however could you explain why you want to skip this pedestal subtraction for MC calibration events?

For observed CTAO data, we expect that the CAT-A calibration is also applied to interleaved events, so they would also come with a subtracted baseline.

So isn't the current situation (some initial baseline is subtracted) also what we expect in the real case?

TjarkMiener:
We wanted this subtraction to be disabled for testing and validation purposes of the camera calibration routines. We can also compare to Cat-A calibration.

src/ctapipe/io/simteleventsource.py
```diff
@@ -546,6 +546,14 @@ class SimTelEventSource(EventSource):
         ),
     ).tag(config=True)
 
+    subtract_pedestal_baseline = Bool(
+        True,
+        help=(
+            "Subtract the pedestal baseline from the waveform samples."
+            " This is done by default, but can be disabled for calibration events."
```
kosack, line 553: This implementation seems to turn It off globally (for all events), not just for calibration events. Would be be better to really apply It only to calibration event types?  See the list of event types: https://github.com/cta-observatory/ctapipe/blob/7e40bd23019a06793b4ea2d5c29cbf1403093467/src/ctapipe/containers.py#L151-L172

maxnoe: I think you need to turn it off globally: if you compute a new pedestal based on non-subtracted pedestal events, you also do not want to apply the initial pedestal subtraction of simtel to the physics events, right? You want to apply the newly computed pedestal, which will be absolute, not relative to the initial one.

maxnoe:
A test is missing for the `subtract_pedestal_baseline=False` case.

Also: shouldn't we also add an option for the pe conversion then? So we can test the computation of absolute calibration coefficients?

TjarkMiener:
@maxnoe Yes, exactly! I added only one option that skips the cat-A calib for simtel files. I'm testing if the R0 and R1 waveforms are identical.

src/ctapipe/io/simteleventsource.py
```diff
@@ -546,6 +555,14 @@ class SimTelEventSource(EventSource):
         ),
     ).tag(config=True)
 
+    skip_r1_calibration = Bool(
+        default_value=False,
+        help=(
+            "Skip the simtel category-A calibration. The R1 waveforms "
+            "(in ADC counts) will the same as the R0 waveforms."
```
maxnoe, line 562:
```suggestion
            "(in ADC counts) will be the same as the R0 waveforms."
```

src/ctapipe/io/tests/test_simteleventsource.py
```diff
@@ -236,6 +236,17 @@ def test_calibration_events():
...
 
+def test_skip_r1_calibration():
+    with SimTelEventSource(
+        input_url=calib_events_path,
+        skip_calibration_events=False,
+        skip_r1_calibration=True,
+        focal_length_choice="EQUIVALENT",
+    ) as reader:
+        for event in reader:
+            assert np.array_equal(event.r0.tel[1].waveform, event.r1.tel[1].waveform)
```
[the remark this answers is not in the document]
maxnoe: Also, this test does not fail in case no event is produced by the reader, so please add something like:

```
n_processed = 0
for event in ...
    n_processed += 1
    ...
assert n_processed == <n_expected>

```

src/ctapipe/io/tests/test_simteleventsource.py
```diff
@@ -236,6 +236,24 @@ def test_calibration_events():
...
+        skip_calibration_events=False,
+        skip_r1_calibration=True,
+        focal_length_choice="EQUIVALENT",
+    ) as reader:
+        n_processed = 0
+        for event in reader:
+            n_processed += 1
+            assert np.issubdtype(event.r0.tel[1].waveform.dtype, np.uint16)
+            assert np.issubdtype(event.r1.tel[1].waveform.dtype, np.float32)
+            assert np.allclose(event.r0.tel[1].waveform, event.r1.tel[1].waveform)
```
maxnoe, line 253: float32 can represent all values of uint16 exactly, so this should check actual equality not just `isclose`.

Also, for better failure messages, use the functions in `numpy.testing` instead of `assert ...`

mexanick (approved):
LGTM. @maxnoe, could you include it in the next release? We need this for DPPS 0.2.0

PR #2763, opened 2025-05-30
Rename 'classification' field from ReconstructedContainer to 'particle_type'

### LINKED ISSUE #2759: Naming missmatch between `classification` and `particle_type` group in HDF5EventSource
This came up using the `HDF5EventSource` to stream in some DL2 data with applied particle classifier where i got:
```
Unknown DL2 stereo group particle_type
Unknown DL2 telescope group particle_type
```

It tries to read in the `classification` group defined here:
https://github.com/cta-observatory/ctapipe/blob/e6eeae9747e96d4e021f661da36f80f4c5027694/src/ctapipe/io/hdf5eventsource.py#L66-L72

but the group is named "particle_type" via the `ReconstructionProperty`-Class
https://github.com/cta-observatory/ctapipe/blob/e6eeae9747e96d4e021f661da36f80f4c5027694/src/ctapipe/reco/reconstructor.py#L25-L45

I think "particle_type" would be the better naming here (more specific, fits better to geometry, energy). 
To be more coherent, we could also adapt the `ReconstructedContainer`-Field here:
https://github.com/cta-observatory/ctapipe/blob/e6eeae9747e96d4e021f661da36f80f4c5027694/src/ctapipe/containers.py#L1087-L1090

from `classification` to `particle_type` or just `particle`.

Hckjs:
- Update DataModel Version to v7.2.0 (DataWriter writes dl2 'particle_type' group instead of 'classification')
 - HDF5EventSource can read both (morphs 'classification' to 'particle_type')
 - TableLoader should work as it is (just loops over all nodes children)
 - particle_type field still reachable via 'classification' but raises DeprecationWarning (can delete it with the next breaking change)

fixes #2759 

 - Reduce cognitive complexity of `HDF5EventSource._generator` to calm SonarQube down

### DIFF
--- a/docs/changes/2763.bugfix.rst
+++ b/docs/changes/2763.bugfix.rst
@@ -2,4 +2,3 @@ Fix naming mismatch between ``classification`` and ``particle_type`` group
 in ``HDF5EventSource``.
 Rename the ``classification`` field from ``ReconstructedContainer`` to
 ``particle_type``.
-Update DataModel Version to v7.2.0


maxnoe:
Changelog missing and some sonarqube issues.

maxnoe (dismissed):

maxnoe (approved):

kosack (approved):

PR #2773, opened 2025-06-12
Rework ml model serialization

### LINKED ISSUE #2720: Config values not applied to loaded Reconstructor models in `ApplyModels`

LukasBeiske:
This is a first rework of how we save our ml models. To move away from pickle completely some more changes are still needed, but this fixes #2720 and supersedes #2724.

TODO:
- [ ] Tests
- [ ] What kind of metadata do we want to save in the output?

TODO to move away from pickle -> Maybe as a follow up PR:
- [ ]  Save and apply models using ONNX (https://onnx.ai/sklearn-onnx/)
- [ ] Properly serialize the SubarrayDescription

### DIFF
--- a/src/ctapipe/reco/sklearn.py
+++ b/src/ctapipe/reco/sklearn.py
@@ -300,7 +300,7 @@ class SKLearnRegressionReconstructor(SKLearnReconstructor):
             return np.log(y)
         return y
 
-    def write(self, path, meta={}, overwrite=False):
+    def write(self, path, meta=None, overwrite=False):
         """
         Save a dictionary using joblib-pickle, which contains all
         information/settings about an instance of a
@@ -928,7 +928,7 @@ class DispReconstructor(Reconstructor):
                 disp.n_jobs = n_jobs.new
                 sign.n_jobs = n_jobs.new
 
-    def write(self, path, meta={}, overwrite=False):
+    def write(self, path, meta=None, overwrite=False):
         """
         Save a dictionary using joblib-pickle, which contains all
         information/settings about an instance of a ``DispReconstructor`` .


src/ctapipe/reco/sklearn.py
```diff
@@ -386,6 +421,53 @@ def _predict_score(self, key, table):
     def _get_positive_index(self, key):
         return np.nonzero(self._models[key].classes_ == self.positive_class)[0][0]
 
+    def write(self, path, meta={}, overwrite=False):
```
maxnoe, line 424:
```suggestion
    def write(self, path, meta=None, overwrite=False):
```

No mutable default arguments.

src/ctapipe/reco/sklearn.py
```diff
@@ -386,6 +421,53 @@ def _predict_score(self, key, table):
...
+        """
+        Save a dictionary using joblib-pickle, which contains all
```
maxnoe, line 426: This looks like you are implementing your own way of

https://docs.python.org/3/library/pickle.html#object.__getstate__
and
https://docs.python.org/3/library/pickle.html#object.__setstate__

LukasBeiske:
> > but this fixes #2720 which is needed for #2731
> 
> Why does adding a new combiner depend on the possibility to change the combiner between training and application?

It does not really, my bad, but the changes from #2724 are included in #2731. I' adjusted the comment above.

src/ctapipe/reco/sklearn.py
LukasBeiske: I started this with the intention of completely removing pickle. I can also convert this to draft until I can finish this.

PR #2776, opened 2025-06-13
add HDF5Merger to the pix stats tool

TjarkMiener:
Unittests are updated accordingly.

### DIFF
--- a/docs/changes/2776.feature.rst
+++ b/docs/changes/2776.feature.rst
@@ -1,1 +1,2 @@
-Add the observation block and scheduling block info to the DL1 camcalib montoring file.
+Use the HDF5Merger to add the observation block and scheduling block info
+to the DL1 files produced by ctapipe-calculate-pixel-statistics tool.
--- a/docs/changes/2776.feature.rst
+++ b/docs/changes/2776.feature.rst
@@ -1,0 +1,1 @@
+Add HDF5Merger to the pix stats tool.
--- a/docs/changes/2776.feature.rst
+++ b/docs/changes/2776.feature.rst
@@ -1,1 +1,1 @@
-Add HDF5Merger to the pix stats tool.
+Add the observation block and scheduling block info to the DL1 camcalib montoring file.
--- a/src/ctapipe/tools/calculate_pixel_stats.py
+++ b/src/ctapipe/tools/calculate_pixel_stats.py
@@ -91,15 +91,9 @@ class PixelStatisticsCalculatorTool(Tool):
         self.input_data = self.enter_context(
             TableLoader(
                 parent=self,
+                dl1_images=True,  # Ensure that dl1 images are read
             )
         )
-        # Check that the DL1 images are selected in the TableLoader
-        if "dl1_images" in self.input_data.config.TableLoader:
-            if not self.input_data.dl1_images:
-                raise ToolConfigurationError(
-                    "The TableLoader must read dl1 images. Set 'dl1_images' to True."
-                )
-        self.input_data.dl1_images = True
         # Copy selected tables from the input file to the output file
         self.log.info(
             "Copying selected data and metadata to output destination using the HDF5Merger component."
@@ -133,7 +127,7 @@ class PixelStatisticsCalculatorTool(Tool):
         # Iterate over the telescope ids and calculate the statistics
         for tel_id in self.subarray.tel_ids:
             # Read the whole dl1 images for one particular telescope
-            dl1_table = self.input_data.read_telescope_events_by_id(
+            dl1_table = self.input_data.read_telescope_events(
                 telescopes=[
                     tel_id,
                 ],
@@ -145,8 +139,6 @@ class PixelStatisticsCalculatorTool(Tool):
                     tel_id,
                 )
                 continue
-            # Get the table for the telescope id
-            dl1_table = dl1_table[tel_id]
             # Check if the chunk size does not exceed the table length of the input data
             if self.stats_calculator.stats_aggregators[
                 self.stats_calculator.stats_aggregator_type.tel[tel_id]
--- a/src/ctapipe/tools/calculate_pixel_stats.py
+++ b/src/ctapipe/tools/calculate_pixel_stats.py
@@ -99,6 +99,7 @@ class PixelStatisticsCalculatorTool(Tool):
                 raise ToolConfigurationError(
                     "The TableLoader must read dl1 images. Set 'dl1_images' to True."
                 )
+        self.input_data.dl1_images = True
         # Copy selected tables from the input file to the output file
         self.log.info(
             "Copying selected data and metadata to output destination using the HDF5Merger component."
--- a/src/ctapipe/tools/calculate_pixel_stats.py
+++ b/src/ctapipe/tools/calculate_pixel_stats.py
@@ -110,6 +110,7 @@ class PixelStatisticsCalculatorTool(Tool):
         self.config["HDF5Merger"]["r0_waveforms"] = False
         self.config["HDF5Merger"]["r1_waveforms"] = False
         self.config["HDF5Merger"]["dl1_images"] = False
+        self.config["HDF5Merger"]["true_images"] = False
         with HDF5Merger(self.output_path, parent=self) as merger:
             merger(self.input_data.input_url)
         # Select a new subarray if the allowed_tels configuration is used
--- a/src/ctapipe/tools/calculate_pixel_stats.py
+++ b/src/ctapipe/tools/calculate_pixel_stats.py
@@ -10,8 +10,9 @@ from astropy.table import vstack
 from ctapipe.core import Tool
 from ctapipe.core.tool import ToolConfigurationError
 from ctapipe.core.traits import (
-    Bool,
+    CInt,
     Path,
+    Set,
     Unicode,
     classes_with_traits,
     flag,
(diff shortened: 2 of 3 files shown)


src/ctapipe/tools/calculate_pixel_stats.py
```diff
@@ -40,16 +38,6 @@ class PixelStatisticsCalculatorTool(Tool):
 
     """
 
-    allowed_tels = Set(
```
maxnoe, line 43: Why remove the `allowed_tels` option here?

TjarkMiener: I will add this option back based on internal discussion

src/ctapipe/tools/tests/test_calculate_pixel_stats.py
```diff
@@ -50,6 +49,9 @@ def test_calculate_pixel_stats_tool(tmp_path, dl1_image_file):
         argv=[
             f"--input_url={dl1_image_file}",
             f"--output_path={monitoring_file}",
+            "--r0-waveforms",
```
mexanick, line 52: is this carrying over R0 waveforms? why do we need it?Following the discussion on slack, by default we should strip everything apart from metadata and only write the results.

src/ctapipe/tools/calculate_pixel_stats.py
```diff
@@ -68,16 +56,40 @@ class PixelStatisticsCalculatorTool(Tool):
...
+            "Exclude r0 waveforms",
+        ),
+        **flag(
+            "r1-waveforms",
+            "HDF5Merger.r1_waveforms",
+            "Include r1 waveforms",
+            "Exclude r1 waveforms",
+        ),
+        **flag(
+            "dl1-images",
```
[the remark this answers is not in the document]
maxnoe: I think you can override the defaults of sub-components by adding a method:

```
@default("HDF5Merger.dl1_images")
def dl1_images_default(self):
    return False
```

src/ctapipe/tools/calculate_pixel_stats.py
```diff
@@ -68,16 +56,40 @@ class PixelStatisticsCalculatorTool(Tool):
 
     overwrite = Bool(help="Overwrite output file if it exists").tag(config=True)
 
+    append = Bool(help="Append to output file if it exists").tag(config=True)
```
maxnoe, line 59: This should be an alias to `HDF5Merger.append`, not it's own option

src/ctapipe/tools/calculate_pixel_stats.py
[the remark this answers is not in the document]
maxnoe: No... this doesn't work unfortunately.

I thought there was option for this, let me do some digging.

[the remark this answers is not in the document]
maxnoe: This seems to work:

```
from ctapipe.core import Tool, Component, traits


class MyComponent(Component):
    option = traits.Bool(default_value=True).tag(config=True)

    def do_something(self):
        self.log.warning("Option: %s", self.option)

class MyTool(Tool):

    flags = {
        **traits.flag("option", "MyComponent.option", "Switch option on", "Switch option off")
    }

    def setup(self):
        self.component = MyComponent(option=False, parent=self)
        self.component.update_config(self.config)

    def start(self):
        self.component.do_something()



if __name__ == "__main__":
    MyTool().run()

```

```
❯ python test_override_default.py --no-option
2025-06-13 15:01:59,140 WARNING [__main__.application.MyComponent] (test_override_default.do_something): Option: False
❯ python test_override_default.py
2025-06-13 15:02:05,710 WARNING [__main__.application.MyComponent] (test_override_default.do_something): Option: False
❯ python test_override_default.py --option
2025-06-13 15:02:09,660 WARNING [__main__.application.MyComponent] (test_override_default.do_something): Option: True
```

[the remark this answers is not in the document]
mexanick: Do we really need exposure of these flags? I would set defaults in the code, and they can be overridden in the config file if needed...

docs/changes/2776.feature.rst
```diff
@@ -0,0 +1 @@
+Add HDF5Merger to the pix stats tool.
```
maxnoe, line 1: I would say this is an implementation detail. What does this give the user?

src/ctapipe/tools/calculate_pixel_stats.py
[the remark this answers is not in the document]
maxnoe: As discussed, you can't really override the defaults with the current config system unfortunately.

src/ctapipe/tools/tests/test_calculate_pixel_stats.py
maxnoe: I think you want to at least copy:

- subarray
- observation block and scheduling block info

and maybe the simulation config in case of simulations

maxnoe: from the actual payload data in addition to the metadata.

maxnoe: Yes, for now I would hardcode what you need in your usecase

maxnoe: I thought having a context manager setting and resetting defaults would work... but it doesn't and I don't understand why yet. All looks good until the instance is created and it gets the wrong default for some reason

src/ctapipe/tools/calculate_pixel_stats.py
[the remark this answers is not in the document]
maxnoe: The best solution I could find was to override `Tool.initialize` with setting a default config:

```
import pytest
from traitlets.config import Config
from ctapipe.core import Tool, Component, traits


class MyComponent(Component):
    option = traits.Bool(default_value=True).tag(config=True)


class MyTool(Tool):

    flags = {
        **traits.flag("option", "MyComponent.option")
    }

    def initialize(self, argv=None):
        # override default values of sub-components
        config = Config()
        config.MyComponent.option = False

        self.update_config(config)
        return super().initialize(argv)

    def setup(self):
        self.component = MyComponent(parent=self)

    def start(self):
        pass



@pytest.mark.parametrize(
    ("argv", "expected"),
    [
        (None, False),
        ([], False),
        (["--MyComponent.option=True"], True),
        (["--MyComponent.option=False"], False),
        (["--option"], True),
        (["--no-option"], False),
    ]
)
def test_override(argv, expected):
    tool = MyTool()
    tool.initialize(argv)
    tool.setup()
    assert tool.component.option is expected
```

docs/changes/2776.feature.rst
TjarkMiener: Please check if rephrasing suits better.

src/ctapipe/tools/calculate_pixel_stats.py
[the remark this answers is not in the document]
TjarkMiener: Just seeing it now. I forced overwritting, so not letting the user write waveforms and images in the DL1 camcalib monitoring file, Please let me know which solution you think is more favorable.

src/ctapipe/tools/calculate_pixel_stats.py
TjarkMiener: I had some problems with the append flag in the calibpipe CI. Simply adding "`--append`" to the pix stats tool was somehow not giving the desired results.

mexanick:
@kosack could you please review? I shouldn't as I contributed to the change.

src/ctapipe/tools/calculate_pixel_stats.py
```diff
@@ -93,24 +93,35 @@ def setup(self):
                 parent=self,
             )
         )
-        # Check that the input and output files are not the same
-        if self.input_data.input_url == self.output_path:
-            raise ToolConfigurationError(
-                "Input and output files are same. Fix your configuration / cli arguments."
-            )
+        # Check that the DL1 images are selected in the TableLoader
         if "dl1_images" in self.input_data.config.TableLoader:
```
kosack, line 97: I'm not sure this logic is correct - the config only includes keys that the user has specified on the command-line or config file, and the default value for `dl1_images` is False.  So this would only work if the user ran with the explicit option `--TableLoader.dl1_images=False`

You don't really need to check the config, you can just check if `if not self.input_data.dl1_images`, since that will be set after the TableLoader is constructed to the default or to the value the user specified.

mexanick: this is a way to override the default in case it is not explicitly set

maxnoe: But now you force it anyway after. Please don't "butcher" the config system in this way. We know it's not really great to work with, but we should minimize these workarounds as much as possible. They tend to break.

kosack: Not really, you can still issue a warning by comparing `self.input_data.config.TableLoader.dl1_images` and self.input_data.dl1_images` if you want. But it might not always work, and certainly will break when we change the config system

kosack: By the way, it's not in the standard repr, but the HTML in a notebook one shows you explicitly which parameters have been changed from the defaults.  Not really useful for a command-line-tool, but worth mentioning.  You can also see what was overridden and used in the provenance outptu, which contains both the input config and the final config that was used after construction. So technically the user can see what is happening to their options, even if a bit convoluted

PR #2779, opened 2025-06-17
Fix conversion to cartesian images for VERITAS camera

maxnoe:
All existing tests are passing.

The new test is passing with the fix.

Example script:

```python
import astropy.units as u
import numpy as np
from astropy.table import QTable
from ctapipe.coordinates import CameraFrame, EngineeringCameraFrame

from ctapipe.instrument import CameraGeometry, PixelShape



table = QTable.read("veritas.camera.ecsv")


pixel_area =  2 * np.sqrt(3) * table["r"]**2

camera_geometry = CameraGeometry(
    name="VERITAS",
    pix_id=table["id"].copy(),
    pix_x=-table["y"].copy(),
    pix_y=-table["x"].copy(),
    pix_area=pixel_area,
    pix_rotation=30*u.deg,
    pix_type=PixelShape.HEXAGON,
    frame=CameraFrame(),
)

# camera_geometry = CameraGeometry.from_name('VERITAS')

def create_mock_image(geom, psi=25 * u.deg):
    """
    creates a mock image, which parameters are adapted to the camera size
    """

    camera_r = camera_geometry.guess_radius()
    model = Gaussian(
        x=0.3 * camera_r,
        y=0 * u.m,
        width=0.03 * camera_r,
        length=0.10 * camera_r,
        psi=psi,
    )

    _, image, _ = model.generate_image(
        geom, intensity=0.5 * geom.n_pixels, nsb_level_pe=3
    )
    return image

images = np.array(
    [
        create_mock_image(camera_geometry, psi=i * 30 * u.deg)
        for i in range(4)
    ]
)

fig, axs = plt.subplots(1, 3, layout="constrained", figsize=(15, 5))
disp1 = CameraDisplay(camera_geometry, image=images[0], ax=axs[0])
disp2 = CameraDisplay(camera_geometry.transform_to(EngineeringCameraFrame()), image=images[0], ax=axs[1])

images_2d = camera_geometry.image_to_cartesian_representation(images)

axs[0].set_title("CameraFrame")
axs[1].set_title("EngineeringCameraFrame")
axs[2].imshow(images_2d[0])
axs[2].set_title("To Cartesian")

plt.show()
```

Result:

veritas_cam_fixed

### DIFF
--- a/src/ctapipe/instrument/camera/tests/test_image_conversion.py
+++ b/src/ctapipe/instrument/camera/tests/test_image_conversion.py
@@ -36,9 +36,9 @@ def test_single_image(camera_geometry, image_conversion_path):
     Test if we can transform toy images for different geometries
     and get the same images after transforming back
     """
-    import matplotlib.pyplot as plt
-
     from ctapipe.visualization import CameraDisplay
+
+    plt = pytest.importorskip("matplotlib.pyplot")
 
     image = create_mock_image(camera_geometry)
     image_2d = camera_geometry.image_to_cartesian_representation(image)


maxnoe:
Yes, exactly. Thanks for confirming

maxnoe:
@TjarkMiener @nietootein @vuillaut 

I am not sure what disabling this "scale_aspect" option does for the hex to 2d cartesian conversions and would be happy for a cross check on what changing this option means for the ML people

maxnoe:
I checked all other cameras and with the exception of the Whipple camera that uses two different pixel sizes, the results look fine.
ASTRICam
CHEC
DigiCam
FACT
FlashCam
HESS-I
HESS-II
LSTCam
MAGICCam
NectarCam
SCTCam
VERITAS
Whipple490

maxnoe:
Ok, I figured this out... this code is a mess and much more complicated than it needs to be.

The real reason why it failed for the veritas cam was that `get_orthogonal_pixel_edges` modifiies `pix_x` _in place_ by x-scale.

This wasn't an issue for all bundled geometries, because these are defined in unit `m` and the code calls `to_value(u.m)`, however @deivdribeira provided units in `mm`, for which `to_value(u.m)` returns a copy and then the edges where in m, but the rot_x still in mm and not scaled.

I now made the code more correct by

* not modifying the input array in place
* properly aligning the edges
* replacing the complex calls to histogramdd by simple calls to `np.digitize`.

With this, we can leave the scaled option as before and the VERITAS geometry also works.

Proper edges:

edges

kosack (dismissed):

kosack (dismissed):

maxnoe:
I'll merge here since it is blocking other PRs and fixes #2784.

The only thing changed since @kosack approved was to make the CI pass in the job without matplotlib by skipping the test that now produces plots.

PR #2787, opened 2025-06-25
Modify camera monitoring data format

TjarkMiener:
This PR implements the camera monitoring data format, which we will be proposing next week at the DPPS workshop. HDF5merger is updated accordingly, so that we can use the `ctapipe-merge` for our camcalib data files.

### DIFF
--- a/src/ctapipe/tools/calculate_pixel_stats.py
+++ b/src/ctapipe/tools/calculate_pixel_stats.py
@@ -184,7 +184,7 @@ class PixelStatisticsCalculatorTool(Tool):
                     )
             # Cnstruct the output table name based on the event type and the selected column name
             output_table_name = (
-                f"{EventType(dl1_table["event_type"][0]).name}_{self.input_column_name}"
+                f"{EventType(dl1_table['event_type'][0]).name}_{self.input_column_name}"
             )
             self.log.info(f"{output_table_name}")
             # Write the aggregated statistics and their outlier mask to the output file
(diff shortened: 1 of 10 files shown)


mexanick (approved):

src/ctapipe/tools/calculate_pixel_stats.py
```diff
@@ -186,20 +182,19 @@ def start(self):
                         "No faulty chunks found for telescope 'tel_id=%d'. Skipping second pass.",
                         tel_id,
                     )
-            # Add metadata to the aggregated statistics
-            aggregated_stats.meta["event_type"] = dl1_table["event_type"][0]
-            aggregated_stats.meta["input_column_name"] = self.input_column_name
+            # Cnstruct the output table name based on the event type and the selected column name
+            output_table_name = f"{EventType(dl1_table['event_type'][0]).name.lower()}_{self.input_column_name}"
```
maxnoe, line 186: Any reason why we here use a composite table name instead of a group / table hierarchy?

mexanick: this is not really a table name, this is a part of the group/table hierarchy... 
the actual path to table will be e.g.

```
/dl1/monitoring/telescope/calibration/camera/pixel_statistics/flatfield_image/tel_001
```

maxnoe: It's fine.

Just noting that you can also iterate easily if it's in groups by doing `f.root[path].f_walk_nodes("Table")` for example.

maxnoe (approved):

PR #2789, opened 2025-06-26
Add full cut optimization as introduced in pyirf 0.13

### LINKED ISSUE #2771: Add optimize based on full cut optimization in pyrif.cuts.optimize_cuts

The new pyirf release includes a full grid-search cut optimization on multiplicty, theta and gammaness.

We should add a wrapper for this new function here in `ctapipe-optimize-event-selection`.

Add a new cut optimizer based on the new function in pyirf.

**Additional context**

LukasBeiske:
This changes the `PointSourceSensitivityOptimizer` to use the full cut optimization introduced in pyirf 0.13. The previous EventDisplay-like optimization can now be used via the `PointSourceSensitivityGhOptimizer`. 

I did a quick comparison of the three optimizers using prod6 files (multiplicity >= 2 for gh opt and percentile cuts, as the HillasReconstructer is used):
gh_cuts
theta_cuts
mult_cuts
sensitivity

I am not sure, why the EventDisplay-like optimization results in a better sensitivity at high energies.

Fixes #2771


maxnoe:
How fine did you make the scanning of the cuts? In principle, the full cut opt should always be at least as good as the one that is restricted,.if it is allowed to find the same cuts.

LukasBeiske:
> How fine did you make the scanning of the cuts? In principle, the full cut opt should always be at least as good as the one that is restricted,.if it is allowed to find the same cuts.

I kept everything at the default values, as we have them in here right now. So, if I'm not mistaken, the only difference would be that the EventDisplay-like optimization has a theta cut with 68% efficiency, while the full optimization only tries 60% and 70%.
But I doubt that this makes such a difference. I'll test that and take a closer look again tomorrow.

src/ctapipe/irf/optimize.py
```diff
@@ -129,7 +136,16 @@ def read(cls, file_name):
             gh_cuts = QTable.read(hdul[2])
             valid_energy = QTable.read(hdul[3])
             valid_offset = QTable.read(hdul[4])
-            spatial_selection_table = QTable.read(hdul[5]) if len(hdul) > 5 else None
+            spatial_selection_table = None
+            multiplicity_cuts = None
+            if len(hdul) == 6:
+                if hdul[5].header["EXTNAME"] == "RAD_MAX":
+                    spatial_selection_table = QTable.read(hdul[5])
```
maxnoe, line 143: you can just index by extname `hdul["RAD_MAX"]` and also check `"RAD_MAX" in hdul`

maxnoe: do not rely on indicies

LukasBeiske:
There was still an error with the application of the cuts in the irf tool. It should be correct now and this improved the sensitivity situation a lot, but there still some bins where the EventDisplay-like optimization outperforms the full optimization, even though the first uses a theta cut with 70% efficiency for this plot, which also gets tested in the full optimization.
sensitivity

kosack (changes requested):
>   gh_cut_efficiency_step=0.02
>   theta_cut_efficiency_step=0.02

If a smaller step is important to get a good sensitivity, please also update the _ctapipe-quickstart_ configurations to provide the best values for users, and also the default values in the tool itself.

maxnoe:
> There was still an error with the application of the cuts in the irf tool. 

Is this an error introduced here? Or does it affect main? If it affects also main, could you open a PR just with the bugfix please?

Hckjs:
> > > gh_cut_efficiency_step=0.02
> > > theta_cut_efficiency_step=0.02
> > 
> > 
> > If a smaller step is important to get a good sensitivity, please also update the _ctapipe-quickstart_ configurations to provide the best values for users, and also the default values in the tool itself.
> 
> I just noticed this (similar for the full optimization): sensitivity
> 
> The config is exactly the same (default values) besides the stepsize. This doesn't make sense. I'm re-running everything now and, if this behavior is still there, I'll convert this PR to draft until I figure out whats going on here.

Did you use the same dataset here to calculate the sensitivities as for the cut optimization?

maxnoe:
@Hckjs This is correct yes. One could probably get around this by optimizing again at least once or until it converges. But I think we should just use the global optimization scheme instead.

maxnoe:
Checking with @Hckjs, I think we understand what is happening.

The cut optimization is done in a single FoV range, by default from 0° to 5° degrees offset.

If you optimize the cuts in this full range but then compute sensitivity in different, smaller offset ranges, the assumption smaller step = better or equal sensitivity does not hold.

LukasBeiske:
That fixed it (See plots below)! Thanks for looking into this and sorry that it took me this long to respond. 

irf_sensitvity_comparison
cut_comparison

LukasBeiske:
Would it make sense to already refactor the irf code here to use Karls EventPreprocessor added in #2928? Or would it be better to wait for #2927 and then refactor the irf code to use both in a separate PR?

maxnoe:
Let's do independent things in independent PRs, this is already a great new feature

src/ctapipe/resources/optimize_cuts.yaml
```diff
@@ -29,10 +28,25 @@ DefaultRecoEnergyBins:
   reco_energy_max: 150 TeV
   reco_energy_n_bins_per_decade: 5
 
+PointSourceSensitivityOptimizer:
+  min_background_fov_offset: 0.0 deg
+  max_background_fov_offset: 5.0 deg
+  max_gh_cut_efficiency: 0.8
+  gh_cut_efficiency_step: 0.1
+  alpha: 0.2
+  theta_min_angle: None
```
Hckjs:
```suggestion
  theta_min_angle: null
```
Yaml converts the None to a string -> "None"

src/ctapipe/resources/optimize_cuts.yaml
```diff
@@ -29,10 +28,25 @@ DefaultRecoEnergyBins:
...
+  multiplicity_cuts: [2, 3, 4, 5, 6, 7, 8, 9, 10]
+
+PointSourceSensitivityGhOptimizer:
+  initial_gh_cut_efficency: 0.4
+
 ThetaPercentileCutCalculator:
-  theta_min_angle: -1 deg
-  theta_max_angle: 0.32 deg
-  theta_fill_value: 0.32 deg
+  theta_min_angle: None
```
Hckjs:
```suggestion
  theta_min_angle: null
```

src/ctapipe/core/traits.py
```diff
@@ -97,27 +97,34 @@ class AstroQuantity(TraitType):
 
     def __init__(self, physical_type=None, **kwargs):
         super().__init__(**kwargs)
-        if physical_type is not None:
```
maxnoe: These changes seem not directly related to the cut optimization.

What is the change here and why was it needed? If there is an actual behavior change here, could you maybe open a separate pull request just with the changes for the quantity trait?

src/ctapipe/io/dl2_tables_preprocessing.py
```diff
@@ -436,6 +440,29 @@ def make_derived_columns(self, events: QTable) -> QTable:
...
+                np.count_nonzero(
+                    events[f"{self.epp.energy_reconstructor}_telescopes"], axis=1
+                ),
+                np.count_nonzero(
+                    events[f"{self.epp.geometry_reconstructor}_telescopes"], axis=1
+                ),
+                np.count_nonzero(
+                    events[f"{self.epp.gammaness_classifier}_telescopes"], axis=1
+                ),
+            ],
```
maxnoe, line 453: I think this could be more readable using a loop:


```python
        reconstructors = (
            self.epp.energy_reconstructor,
            self.epp.geometry_reconstructor,
            self.epp.gammaness_classifier,
        )
        multiplicity = np.min([
            np.count_nonzero(events[f"{reconstructor}_telescopes"], axis=1)
            for reconstructor in  reconstructors
        ], axis=0)
```

src/ctapipe/io/dl2_tables_preprocessing.py
```diff
@@ -469,6 +496,11 @@ def make_event_weights(
         ValueError
             If ``fov_offset_bins`` is required but not provided.
         """
+        # Re-initialize weights = 0 to exclude events outside fov bins
+        # in the cut optimization.
+        # This will become unnecessary once fov bins are used during cut optimization.
+        events["weight"] = 0.0
```
maxnoe, line 496: This is a bugfix that I think should maybe also be done first in a separate pull request.

pyproject.toml
```diff
@@ -58,7 +58,7 @@ all = [
     "ctapipe[eventio]",
     "iminuit >=2",
     "matplotlib ~=3.0",
-    "pyirf ~=0.13.0"
+    "pyirf >=0.13.0",
```
maxnoe: Why this change? 0.13 is still the latest release.

src/ctapipe/core/traits.py
LukasBeiske: Sonarqube complained about the code complexity here. The only functional change was obtaining the `physical_type` from the `default_value`, if no `physical_type` is provided. 

Should this still be its own PR?

pyproject.toml
LukasBeiske: I don't remember and it also makes no sense to me right now. I'll revert it.

src/ctapipe/core/traits.py
maxnoe: sonarqube should only complain about complexity of the code you touched. 

Determining `physical_type` from the default sounds reasonable, but is a good isolated change. So yes, I would prefer that as a separate PR, not mixed in with the other more complex changes here.

LukasBeiske: https://github.com/cta-observatory/ctapipe/pull/3012

src/ctapipe/io/dl2_tables_preprocessing.py
LukasBeiske: #3013

kosack (dismissed):

src/ctapipe/irf/optimize.py
```diff
@@ -386,6 +383,30 @@ class PointSourceSensitivityOptimizer(CutOptimizerBase):
...
+                opt_sensitivity["reco_energy_low"][keep_mask][0],
+                opt_sensitivity["reco_energy_high"][keep_mask][-1],
+            ]
+        else:
+            raise ValueError("Optimal significance curve has internal NaN bins")
+
+
+class PointSourceSensitivityGhOptimizer(PointSourceSensitivityOptimizerBase):
+    """
+    EventDisplay-like procedure of optimizing a G/H cut for maximum point source
```
maxnoe: First line of the docstring should be a short title

src/ctapipe/irf/optimize.py
```diff
@@ -466,14 +487,86 @@ def __call__(
...
-        if all(np.diff(count[keep_mask]) == 1):
-            return [
-                opt_sensitivity["reco_energy_low"][keep_mask][0],
-                opt_sensitivity["reco_energy_high"][keep_mask][-1],
-            ]
-        else:
-            raise ValueError("Optimal significance curve has internal NaN bins")
+class PointSourceSensitivityOptimizer(PointSourceSensitivityOptimizerBase):
+    """
+    Finds the combination of G/H cut, theta cut, and cut on the event-multiplicity
```
maxnoe: First line should be a title.

I'd include a link to the pyirf docs here: https://pyirf.readthedocs.io/en/latest/api/pyirf.cut_optimization.optimize_cuts.html#pyirf.cut_optimization.optimize_cuts

src/ctapipe/irf/optimize.py
maxnoe: Also include a link to the pyirf docs please: https://pyirf.readthedocs.io/en/latest/api/pyirf.cut_optimization.optimize_gh_cut.html#pyirf.cut_optimization.optimize_gh_cut

maxnoe (dismissed):
Two minor comments on docstrings, otherwise looks good, thanks a lot!

maxnoe:
I resolved those myself

maxnoe (approved):

kosack (approved):

PR #2791, opened 2025-06-30
Generalise table preprocessing

### LINKED ISSUE #2790: Generalise DL2 tables loading and filtering
Telescope cross-calibration.

Currently, for telescope cross calibration we have implemented our own table filtering. However similar functionality has been implemented in IRF tools. Our idea is to reuse the existing functionality from IRF tools for cross calibration. The added benefits is increased maintainability. Moreover, cross calibration memory efficiency will be improved by reading data by chunks. Finally, this can potentially benefit other usecases that load DL2 tables. 

In order to do that, we will propose some modifications in irf/preprocessing. However we should ensure that the default behaviour stays the same.

Voutsi:
Closes #2790 

Refactoring the DL2 processing for producing IRFs. The motivation is to use the same code for more usecases that process DL2 data, e.g. the telescope cross calibration.

The refactored module is the irf/preprocessing.py

The default behaviour of the preprocessing remains the same as before.

### DIFF
--- a/docs/changes/2851.bugfix.rst
+++ b/docs/changes/2851.bugfix.rst
@@ -1,0 +1,3 @@
+Fixed bug in `ctapipe.core.Tool.get_current_config` where ``Tools`` containing lists of
+``Components`` did not get registered, and thus their configs did not end up in the
+provenance log.
--- a/src/ctapipe/core/component.py
+++ b/src/ctapipe/core/component.py
@@ -1,4 +1,5 @@
-""" Class to handle configuration for algorithms """
+"""Class to handle configuration for algorithms"""
+
 import html
 import warnings
 import weakref
@@ -226,12 +227,22 @@ class Component(Configurable, metaclass=AbstractConfigurableMeta):
         """return the current configuration as a dict (e.g. the values
         of all traits, even if they were not set during configuration)
         """
+        from .tool import Tool  # to avoid circular import
+
         name = self.__class__.__name__
         config = {name: {k: v.get(self) for k, v in self.traits(config=True).items()}}
 
         for val in self.__dict__.values():
-            if isinstance(val, Component):
+            if isinstance(val, (Component, Tool)):
                 config[name].update(val.get_current_config())
+            if (
+                isinstance(val, (list, tuple))
+                and val
+                and isinstance(val[0], (Component, Tool))
+            ):
+                for element in val:
+                    if isinstance(element, (Component, Tool)):
+                        config[name].update(element.get_current_config())
 
         return config
 
--- a/src/ctapipe/core/tests/test_tool.py
+++ b/src/ctapipe/core/tests/test_tool.py
@@ -199,6 +199,46 @@ def test_tool_current_config_subcomponents():
     assert current_config["MyTool"]["userparam"] == 2.0
 
 
+def test_tool_current_config_subcomponents_list():
+    """Check that we can get the full instance configuration for tools that
+    contain lists of subcomponents (which can be tools)"""
+    from ctapipe.core.component import Component
+
+    class SubComponent(Component):
+        param = Int(default_value=3).tag(config=True)
+
+    class SubComponent2(Component):
+        param = Int(default_value=3).tag(config=True)
+
+    class MyComponent(Component):
+        val = Int(default_value=42).tag(config=True)
+
+        def __init__(self, config=None, parent=None):
+            super().__init__(config=config, parent=parent)
+            self.subs = [SubComponent(parent=self), SubComponent2(parent=self)]
+
+    class MyTool(Tool):
+        description = "test"
+        userparam = Float(5.0, help="parameter").tag(config=True)
+
+        def setup(self):
+            self.my_comp = MyComponent(parent=self)
+
+    config = Config()
+    config.MyTool.userparam = 2.0
+    config.MyTool.MyComponent.val = 10
+    config.MyTool.MyComponent.SubComponent.param = -1
+
+    tool = MyTool(config=config)
+    tool.setup()
+
+    current_config = tool.get_current_config()
+    assert current_config["MyTool"]["MyComponent"]["val"] == 10
+    assert current_config["MyTool"]["MyComponent"]["SubComponent"]["param"] == -1
+    assert current_config["MyTool"]["MyComponent"]["SubComponent2"]["param"] == 3
+    assert current_config["MyTool"]["userparam"] == 2.0
+
+
 def test_tool_exit_code():
     """Check that we can get the full instance configuration"""
 
@@ -665,3 +705,45 @@ def test_no_ignore_bad_config_type(tmp_path: Path):
     # test correct case:
     tool.load_config_file(good_conf_path)
     assert tool.float_option > 1
+
+
+def test_tool_in_tool():
+    """Check that a Tool that calls other Tools has the right config in the
+    provenance log."""
+
+    class InnerTool(Tool):
+        param1 = traits.Integer(12).tag(config=True)
+
+        def start(self):
+            print(f"started inner: {self.get_current_config()}")
+
+    class CompoundTool(Tool):
+        param1 = traits.Integer(12).tag(config=True)
+        filename = traits.Unicode().tag(config=True)
+
+        classes = [
+            InnerTool,
+        ]
+
+        def setup(self):
+            self._inner = InnerTool(parent=self)
+
+        def start(self):
+            print(f"started calib: {self.get_current_config()}")
+
+            run_tool(self._inner)
+
+    conf = Config()
+    conf.InnerTool.param1 = 6
+    conf.CompoundTool.filename = "test.txt"
+    conf.CompoundTool.param1 = 100
+
+    tool = CompoundTool(config=conf)
+    run_tool(tool, raises=False)  # have to run it for setup()
+
+    assert "InnerTool" in tool.config
+
+    current_config = tool.get_current_config()
+    assert "CompoundTool" in current_config
+    assert "InnerTool" in current_config["CompoundTool"]
+    assert current_config["CompoundTool"]["InnerTool"]["param1"] == 6
--- a/src/ctapipe/core/tool.py
+++ b/src/ctapipe/core/tool.py
@@ -505,17 +505,7 @@ class Tool(Application):
         """return the current configuration as a dict (e.g. the values
         of all traits, even if they were not set during configuration)
         """
-        conf = {
-            self.__class__.__name__: {
-                k: v.get(self) for k, v in self.traits(config=True).items()
-            }
-        }
-
-        for val in self.__dict__.values():
-            if isinstance(val, Component):
-                conf[self.__class__.__name__].update(val.get_current_config())
-
-        return conf
+        return Component.get_current_config(self)
 
     def _repr_html_(self):
         """nice HTML rep, with blue for non-default values"""


src/ctapipe/irf/preprocessing.py
```diff
@@ -235,13 +277,34 @@ def load_preselected_events(
             return table, n_raw_events, meta
 
     def get_simulation_information(
-        self, loader: TableLoader, obs_time: u.Quantity
```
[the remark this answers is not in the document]
Voutsi: No it wasn't intentional. Thanks for spotting it.

src/ctapipe/irf/preprocessing.py
```diff
@@ -189,44 +133,158 @@ def make_empty_table(self) -> QTable:
...
+
+        default = {
+            f"{self.energy_reconstructor}_energy": "reco_energy",
+            f"{self.geometry_reconstructor}_az": "reco_az",
+            f"{self.geometry_reconstructor}_alt": "reco_alt",
+            f"{self.gammaness_classifier}_prediction": "gh_score",
+            "subarray_pointing_lat": "pointing_alt",
+            "subarray_pointing_lon": "pointing_az",
+        }
+        return {**default, **self.columns_to_rename_override}
```
LukasBeiske, line 159: Regarding my above thread, this only extends the default renaming and does not overwrite it, no?
Do you want to do something like this:
```python
if self.columns_to_rename_override is None:
    return default
else:
    return self.columns_to_rename_override
```

src/ctapipe/irf/preprocessing.py
```diff
@@ -67,110 +67,54 @@ class EventPreprocessor(Component):
...
-                    f"Required column {c} is missing."
-                )
+    apply_check_pointing = Bool(
+        default_value=True, help="Accept only pointing in altaz and not divergent."
+    ).tag(config=True)
 
-        events = QTable(events[keep_columns], copy=COPY_IF_NEEDED)
-        events.rename_columns(rename_from, rename_to)
-        return events
+    fixed_columns = List(
```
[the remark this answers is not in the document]
Voutsi: These are the columns we want to keep from the input table without any renaming. I modified a bit the help.

docs/changes/2791.optimization.rst
```diff
@@ -1,3 +1 @@
-Currently, for telescope cross calibration we have implemented our own table filtering. However similar functionality has been implemented in IRF tools. Our idea is to reuse the existing functionality from IRF tools for cross calibration. The added benefits is increased maintainability. Moreover, cross calibration memory efficiency will be improved by reading data by chunks. Finally, this can potentially benefit other usecases that load DL2 tables.
-
-In order to do that, we will propose some modifications in irf/preprocessing. However we should ensure that the default behaviour stays the same.
+Generalise the DL2 table processing implemented in IRF tools in order to be used by other usecases that ingest, filter and merge DL2 tables.
```
maxnoe, line 1: This is not really an optimization. More a new feature or an API change.

src/ctapipe/irf/preprocessing.py
```diff
@@ -85,15 +79,16 @@ class EventPreprocessor(Component):
...
+        help="Columns to keep from the input table without renaming.",
     ).tag(config=True)
 
-    # Optional user override
-    columns_to_rename_override = Dict(
-        key_trait=Unicode(),
-        value_trait=Unicode(),
-        default_value={},
-        help="Override of columns to rename. If empty, they are generated dynamically.",
+    columns_to_rename_override = Any(
```
maxnoe, line 85: Why did you change to Any from the correct `Dict[str, str]`?

maxnoe: If the problem was the `None`, the correct way is to specify `allow_none=True`.

maxnoe: I also do not understand why we need the None case.

Can't we keep the default but allow to set to an empty dict to disable?

For what do we need the `None`?

maxnoe: Ah, I see, the default depends on other traits.

Traitlets has a mechanism for this. Define a function and decorate it with `@default("columns_to_rename")`.

src/ctapipe/irf/preprocessing.py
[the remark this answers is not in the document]
mexanick: Following the internal discussion, we will factor out this option. The full list of columns to read, or the schema of the final table to produce, should be indicated only once. Right now this information is duplicated in the configuration, which is not good.

maxnoe:
>  instead of staying under ctapipe.irf as it will be generic now and will be used beyond IRF production. @maxnoe what do you think?

I'm wondering now if this really needs to be its own class or maybe this would be appropriate as a couple of options or additional methods to the `TableLoader`?

maxnoe:
Regardless of that comment, yes, I do think it makes more sense now in `ctapipe.io`.

src/ctapipe/io/dl2_tables_preprocessing.py
```diff
@@ -17,13 +17,14 @@
 from tables import NoSuchNodeError
 from traitlets import default
 
+from ctapipe.irf.spectra import SPECTRA, Spectra
```
maxnoe, line 20: Please keep using relative imports, it will work with two dots:

```
from ..irf.spectra import ...
```

However, please consider to not do this at the top-level but rather only where it is needed to not make https://github.com/cta-observatory/ctapipe/issues/2476 worse again

src/ctapipe/tools/compute_irf.py
```diff
@@ -17,6 +17,8 @@
 from pyirf.cuts import evaluate_binned_cut
 from pyirf.io import create_rad_max_hdu
 
+from ctapipe.io.dl2_tables_preprocessing import EventQualityQuery
```
maxnoe, line 20: Please use relative imports, `from ..io.dl2_tables_preprocessing import ..`

src/ctapipe/io/dl2_tables_preprocessing.py
```diff
@@ -0,0 +1,457 @@
...
+            ),
+            ("valid classifier", "RandomForestClassifier_is_valid"),
+            ("valid geom reco", "HillasReconstructor_is_valid"),
+            ("valid energy reco", "RandomForestRegressor_is_valid"),
+        ],
+        help=QualityQuery.quality_criteria.help,
+    ).tag(config=True)
+
+
+class EventPreprocessor(Component):
```
maxnoe, line 52: Add `DL2`?
```suggestion
class DL2EventPreprocessor(Component):
```
?

mexanick:
@maxnoe , @LukasBeiske , ready for review.

maxnoe (approved):

src/ctapipe/io/dl2_tables_preprocessing.py
```diff
@@ -0,0 +1,474 @@
...
+                        unit=u.deg,
+                        description="Simulated angular offset from pointing direction",
+                    ),
+                    Column(
+                        name="reco_source_fov_offset",
+                        unit=u.deg,
+                        description="Reconstructed angular offset from pointing direction",
+                    ),
+                    Column(
+                        name="gh_score",
```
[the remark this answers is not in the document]
mexanick: It is used in `make_derived_columns` to attach the unit (dimensionless_unscaled), so it is not just simple renaming.
I tried to change it, but then I get the zero division error, which I believe is related to the (absence of) units of the classifier in the input file. To fix this, we may need to update the code that computes/stores it. Or, alternatively, strip the units from gh_score. I suggest to address this in a separate MR, as here the `make_derived_columns` was directly migrated from the irf preprocessing module.

src/ctapipe/io/dl2_tables_preprocessing.py
```diff
@@ -0,0 +1,474 @@
...
+        help="Schema definition for output event QTable",
+    ).tag(config=True)
+
+    def __init__(self, config=None, parent=None, **kwargs):
+        super().__init__(config=config, parent=parent, **kwargs)
+        self.quality_query = DL2EventQualityQuery(parent=self)
+
+    @default("columns_to_rename")
+    def _default_columns_to_rename(self):
+        return {
```
kosack, line 130: A comment I had in the original PR for this as well: this renaming is really just a stop-gap hack until we have a proper Component to do the DL2→DL3 algorithm choice selection.  It was always in the design that we would be able to select per-event, not just per method, which columns to use for the DL3 output, e.g. which reconstructor to make the "final" one.  As currently implemented, this only lets you chose a single reconstruction.   Maybe we should at least add a comment that this is a temporary solution?

src/ctapipe/io/dl2_tables_preprocessing.py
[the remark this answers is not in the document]
LukasBeiske: I remember this problem, although I'm not sure what caused it exactly. 
Besides that though, I think `gh_score` should be included in the default table output schema since it is part of the default column renaming. Otherwise the default output of the `DL2EventPreprocessor` does not conform to the default value of the `output_table_schema`.

src/ctapipe/io/dl2_tables_preprocessing.py
```diff
@@ -0,0 +1,474 @@
...
+            ("valid classifier", "RandomForestClassifier_is_valid"),
+            ("valid geom reco", "HillasReconstructor_is_valid"),
+            ("valid energy reco", "RandomForestRegressor_is_valid"),
+        ],
+        help=QualityQuery.quality_criteria.help,
+    ).tag(config=True)
+
+
+class DL2EventPreprocessor(Component):
+    """Defines pre-selection cuts and the necessary renaming of columns."""
```
kosack, line 63: Not a problem, just a comment, this is not just generalizing the simulation IRF pre-processing and inter-telescope calibration:   this is really a common component to transform DL2.Event data into DL3.Event data! Therefore it should also be used in   #2727.   In principle in fact, what it looks like is that you really want to do cross-calibration on DL3 data, not DL2, so you use this component to produce that in-memory.  

Perhaps that also implies we should name this better?  Is it really pre-processing DL2 events, or post-processing them into DL3? I would say the latter.

src/ctapipe/io/dl2_tables_preprocessing.py
```diff
@@ -0,0 +1,474 @@
...
+                    f"Input files must conform to the ctapipe DL2 data model. "
+                    f"Required column {col} is missing."
+                )
+
+        events = QTable(events[columns_to_read], copy=COPY_IF_NEEDED)
+        if rename_from and rename_to:
+            events.rename_columns(rename_from, rename_to)
+        return events
+
+    def make_empty_table(self) -> QTable:
```
kosack, line 192: The only place where I could find this method being used is in a test fixture (`irf_events_table()`), and even there it's not really used directly, but the result is modified.  Maybe we should just remove this method and put what is needed in the fixture? Or does it have some other use?

src/ctapipe/io/dl2_tables_preprocessing.py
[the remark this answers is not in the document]
kosack: Yes, please fix that elsewhere. As is, running this fails:
```py
proc = DL2EventPreprocessor()
proc.make_empty_table()
```

```
astropy/units/core.py:1189
ZeroDivisionError:  float division by zero
```

src/ctapipe/io/dl2_tables_preprocessing.py
```diff
@@ -0,0 +1,474 @@
...
+
+        opts = dict(dl2=True, simulated=True, observation_info=True)
+
+        with TableLoader(self.file, parent=self, **opts) as load:
+            header = self.epp.make_empty_table()
+            sim_info, spectrum = self.get_simulation_information(load, obs_time)
+            meta = {"sim_info": sim_info, "spectrum": spectrum}
+            bits = [header]
+            n_raw_events = 0
+            reader_func = getattr(load, self.event_reader_function)
```
kosack, line 306: not for this PR, but we should probably open an issue to change  or add to the API of TableLoader to avoid this kind of hack, i.e. to provide a single read function that can work for chunked or non-chunked, depending on a config variable

src/ctapipe/io/dl2_tables_preprocessing.py
```diff
@@ -0,0 +1,474 @@
...
+            Metadata dictionary with simulation info and input spectrum.
+        """
+
+        opts = dict(dl2=True, simulated=True, observation_info=True)
+
+        with TableLoader(self.file, parent=self, **opts) as load:
+            header = self.epp.make_empty_table()
+            sim_info, spectrum = self.get_simulation_information(load, obs_time)
+            meta = {"sim_info": sim_info, "spectrum": spectrum}
+            bits = [header]
```
kosack, line 304: Does it need to be [header] if later you append header again? There must be a nicer way...

src/ctapipe/io/dl2_tables_preprocessing.py
```diff
@@ -0,0 +1,474 @@
...
+            Number of events before selection.
+        meta : dict
+            Metadata dictionary with simulation info and input spectrum.
+        """
+
+        opts = dict(dl2=True, simulated=True, observation_info=True)
+
+        with TableLoader(self.file, parent=self, **opts) as load:
+            header = self.epp.make_empty_table()
+            sim_info, spectrum = self.get_simulation_information(load, obs_time)
```
kosack, line 302: I think this is a carry-over from the original code, but please rename some of these variables to use the same names we use elsewhere or to clarify their meaning:
- header → ???   this is an empty  QTable, so the name "header" doesn't make sense to me
- load → loader 
- bits → ??? (what is this supposed to be? maybe something like `tables_to_stack`?)

src/ctapipe/io/dl2_tables_preprocessing.py
```diff
@@ -0,0 +1,474 @@
...
+        from ..exceptions import OptionalDependencyMissing
+
+        if SimulatedEventsInfo is None:
+            raise OptionalDependencyMissing("pyirf")
+
+        sim = loader.read_simulation_configuration()
+        try:
+            show = loader.read_shower_distribution()
+        except NoSuchNodeError:
+            show = Table([sim["n_showers"]], names=["n_entries"], dtype=[np.int64])
```
kosack, line 354: I think we should log a warning here, as this can effect results. 

"simulation distributions were not found in the input files, falling back to estimating the number of showers from the simulation configuration"

Though we should also add a "fixme" or issue here to use the distributions properly and not assume a power-law spectrum, since that is not guaranteed, particularly if we merge simulations done with different energy ranges (which will happen for efficiency purposes).   So the correct thing is to return an interpolation function of the spectrum generated from the histogram, not use `PowerLaw_from_simulation()`, but that can be a future PR as long as we don't forget about it!

Let's not forget that this is a very critical bit of code for IRF production, and if we don't do it in a robust way, we will get wrong science and might not notice until too late.

mdebony:
@Voutsi following Karl suggestion, I've taken a look at your PR.

In the PR #2727 I've modified quite heavily the EventPreprocessor class. A significant part is too have different treatment of the input file depending if the output is a dl3 or an irf as a lot of columns will be exclusive to one, but even some cuts like the direction cuts are specific to IRF (for the point like IRF case) and are not applied to real data.

How much do you thinks those approach are compatible ? Is it also compatible with you intended final use ? For sure some of the work I have done is generalization, but some of it is also specialization.

src/ctapipe/io/dl2_tables_preprocessing.py
Voutsi: > In principle in fact, what it looks like is that you really want to do cross-calibration on DL3 data

If I am not mistaken, in DL3 we don't have information per telescope right? So we want to apply the cross calibration in a subset of DL2 data, selected based on gammaness. The point of this MR is to reuse existing ctapipe functionality in order to merge/filter DL2 tables.

Voutsi:
Hi @kosack , while I agree with your points, I believe that they are beyond the scope of this pull request. I think it is better to be addressed in a future PR that focus on the improvement of DL3 production.

Voutsi:
> @Voutsi following Karl suggestion, I've taken a look at your PR.
> 
> In the PR #2727 I've modified quite heavily the EventPreprocessor class. A significant part is too have different treatment of the input file depending if the output is a dl3 or an irf as a lot of columns will be exclusive to one, but even some cuts like the direction cuts are specific to IRF (for the point like IRF case) and are not applied to real data.
> 
> How much do you thinks those approach are compatible ? Is it also compatible with you intended final use ? For sure some of the work I have done is generalization, but some of it is also specialization.

Hi @mdebony , thanks for looking into it. I think our usecases are different. What we intent to achieve with this MR is reading/filtering/merging DL2 tables and not process the tables to a next data level.

Voutsi:
> > Hi @mdebony , thanks for looking into it. I think our usecases are different. What we intent to achieve with this MR is generalising the reading/filtering/merging of DL2 tables and not process the tables to a next data level.
> 
> Those class (EventLoader, EventPreprocessor) are also in charge of producing all the derived columns and rename some. To my understanding it is still the case in your PR. This part is heavily dependent on what is the purpose. Also the filtering process depend if you intent to use it for IRF or DL3.
> 
> If we don't want to have just two class EventLoader and two class EventPreprocessor, we need to see how to take that into account, otherwise it will be against the original purpose of your PR.

Exactly. So the purpose here is to add flexibility in the output table schema, renaming (if needed) and in merging/filtering. So in principle it should help your task as well, at least this is the spirit of this PR. Do you see something that might be an obstacle for the production of DL3?

mdebony:
>  Do you see something that might be an obstacle for the production of DL3?

My point remain, if you look at the PR #2727, i add to implement quite a few switch with flags to control behavior depending of the output is for IRF of DL3. I could replicate those change in this new structure, but it could break things for your need as one of the most significant part is in the make_derived_columns, and there are not generalization on this point in this PR.
That's why I'm asking what do you need exactly to see how things could be converged.

src/ctapipe/io/dl2_tables_preprocessing.py
```diff
@@ -0,0 +1,474 @@
...
+    gammaness_classifier = Unicode(
+        default_value="RandomForestClassifier",
+        help="Prefix of the classifier `_prediction` column",
+    ).tag(config=True)
+
+    apply_derived_columns = Bool(
+        default_value=True, help="Whether to compute derived columns"
+    ).tag(config=True)
+
+    apply_check_pointing = Bool(
```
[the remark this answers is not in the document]
kosack: Ah, I understand now. This  is used to disable the check on the pointing coordinate frame.  I would suggest inverting it then: the default should always that an error is raised if the pointing frame is not supported, and you should then rename the option to be `allow_unsupported_pointing_frames=True`  for the case where you want to disable this check.  Then it's much more clear that this is really something that should be turned on in special circumstances.

kosack (changes requested):

Voutsi:
> > Do you see something that might be an obstacle for the production of DL3?
> 
> My point remain, if you look at the PR #2727, i add to implement quite a few switch with flags to control behavior depending of the output is for IRF of DL3. I could replicate those change in this new structure, but it could break things for your need as one of the most significant part is in the make_derived_columns, and there are not generalization on this point in this PR. That's why I'm asking what do you need exactly to see how things could be converged.

Hi, indeed there is no generalisation of the make_derived_columns, but that's deliberate. The scope pf this PR, as it's already mentioned, is to generalise the reading, filtering and merging of DL2 tables. Processing DL2 tables is specific for each use case, that's the reason it is not addressed here apart of making make_derived_columns optional. For example for my use case, telescope cross calibration, we don't need this functionality. Any further processing of DL2 data is taking place in CalibPipe.

So as long as make_derived_columns is optional and as you wrote, controlled depending on the use case, I think there is no danger to break things up.

mexanick:
> > > Do you see something that might be an obstacle for the production of DL3?
> > 
> > 
> > My point remain, if you look at the PR #2727, i add to implement quite a few switch with flags to control behavior depending of the output is for IRF of DL3. I could replicate those change in this new structure, but it could break things for your need as one of the most significant part is in the make_derived_columns, and there are not generalization on this point in this PR. That's why I'm asking what do you need exactly to see how things could be converged.
> 
> Hi, indeed there is no generalisation of the make_derived_columns, but that's deliberate. The scope pf this PR, as it's already mentioned, is to generalise the reading, filtering and merging of DL2 tables. Processing DL2 tables is specific for each use case, that's the reason it is not addressed here apart of making make_derived_columns optional. For example for my use case, telescope cross calibration, we don't need this functionality. Any further processing of DL2 data is taking place in CalibPipe.
> 
> So as long as make_derived_columns is optional and as you wrote, controlled depending on the use case, I think there is no danger to break things up.

@mdebony does this answer your concern?

/cc @maxnoe

maxnoe:
I think since this is the smaller pull request, we should include this first and then the DL3 PR should be rebases on top of this new component, which might need further generalization to also support the DL3 usecase.

@mdebony Is that ok? Or are there any simple changes that could be made here that would improve the situation for you in the DL3 PR?

mexanick:
@mdebony @maxnoe is there anything else we should do to proceed?

src/ctapipe/io/dl2_tables_preprocessing.py
```diff
@@ -0,0 +1,476 @@
...
+try:
+    from pyirf.simulations import SimulatedEventsInfo
+    from pyirf.spectral import (
+        DIFFUSE_FLUX_UNIT,
+        POINT_SOURCE_FLUX_UNIT,
+        PowerLaw,
+        calculate_event_weights,
+    )
+    from pyirf.utils import calculate_source_fov_offset, calculate_theta
+except ModuleNotFoundError:
```
maxnoe, line 19: I think with this many imports, it's simpler to do something like this:

```
try:
    from ...
    has_pyirf = True
except ModuleNotFounderror:
    has_pyirf = False
```

and then raise on `has_pyirf = False`.


```

maxnoe:
> @mdebony @maxnoe is there anything else we should do to proceed?

Most of the review comments by @kosack were not addressed yet, so the ball is in your park

Voutsi:
> > @mdebony @maxnoe is there anything else we should do to proceed?
> 
> Most of the review comments by @kosack were not addressed yet, so the ball is in your park

Hi @maxnoe , @kosack , I had commented above, after Karl's comments, that some of these comments are out of the scope of what this PR aims. I didn't get a reply, if you agree or not. So what do you think?

maxnoe:
At least the request for issuing a warning and for proper variable names should be addressed.

src/ctapipe/io/dl2_tables_preprocessing.py
mexanick: warning added as requested

mexanick (dismissed):
It looks to me the urgent requests has been addressed. I'd prefer to implement the rest in the separate PRs to have them easier to digest and review.

maxnoe:
@kosack please have another look here

mexanick:
> @kosack please have another look here

also, shouldn't it be bumped in priority in [DataPipe Followup](https://github.com/orgs/cta-observatory/projects/16/views/1) to be merged before [DL3 production](https://github.com/cta-observatory/ctapipe/pull/2727)?

Issue #2760, opened 2025-05-27
Enhancement of the muon analysis pipeline.


burmist-git:
**Please describe the use case that requires this feature.**
The muon analysis pipeline has recently been updated to incorporate the weighted Teubin fit, with initial parameters now estimated using the Chaudhuri and Kundu method.

However, the muon inclination is not included in the fit of the muon impact point or in the PDF calculation.

The PDF calculation for the expected number of photoelectrons has a weak dependence on the muon inclination, however, the inclination has a non-negligible effect on the determination of the muon impact point.
The muon impact point is one of the parameters used in the quality cuts. 

Using a poorly reconstructed impact point reduces efficiency and introduces additional systematic uncertainties, which in turn affect optical throughput and PSF measurements

**Describe the solution you'd like**
Muon inclination should be incorporated into the fit of the impact point and into the PDF.
A revision of the muon fit sequence is necessary, along with the implementation of relevant validation tests.

**Describe alternatives you've considered**
An alternative solution is to exclude the muon impact point quality cuts and having cuts applied to the fitted ring only. It is important to note that the muon impact point is derived from fitting the azimuthal distribution of Cherenkov photons—an additional step that follows the ring fitting process.

**Additional context**
The implementation of relevant validation tests need to be added.

burmist-git:
1)  The chord calculation has a bug. 
It does not account for non-physical solutions in the chord calculation for points outside the mirror. This is important for fitting with the central hole in the mirror, or when accounting for camera shadowing.
2) The default value of the LST mirror hole is incorrect, and the 'usual' YAML file used for analysis also contains the wrong value. It is very small.

burmist-git:
<img width="559" height="432" alt="Image" src="https://github.com/user-attachments/assets/0aa34028-32f0-45f0-9367-3a718ba1c34c" />

The code to reproduce this error is here: https://github.com/burmist-git/ctapipe_run_dir/blob/master/muons_chord_D.ipynb

The values from the side are non-physical.

burmist-git:
LST mirror hole is incorrect:
YAML:
<img width="495" height="150" alt="Image" src="https://github.com/user-attachments/assets/4edbb892-20f0-4921-8c3f-6e2d450d44e2" />

https://github.com/cta-observatory/ctapipe/blob/ec9b0fa988103f532368c42f051a7657fd60107d/src/ctapipe/image/muon/intensity_fitter.py#L463C1-L463C38

[...]

PR #2803, opened 2025-07-29
Muon fit chord length fix

burmist-git:
Modified the chord_length function to fix a bug. It now filters out non-physical solutions for chodr vs phi.
Updated the test for chord_length calculation in the muon intensity fitter. It now checks that no non-physical solutions are returned (more details in here : issue #2760).

### DIFF
--- a/docs/api-reference/image/muon.rst
+++ b/docs/api-reference/image/muon.rst
@@ -11,7 +11,7 @@ Algorithms for detecting and characterizing muon images
   :glob:
   :hidden:
 
-  chord_length
+  muon_chord_length
 
 Reference/API
 =============


docs/changes/2803.bugfix.rst
```diff
@@ -0,0 +1 @@
+Filters out non-physical solutions for chodr vs phi.
```
mexanick, line 1:
```suggestion
Filters out non-physical solutions for chord vs phi.
```

mexanick: @maxnoe any idea why this wasn't caught by the codespell pre-commit hook?

src/ctapipe/image/muon/tests/test_intensity_fit.py
```diff
@@ -18,6 +18,11 @@ def test_chord_length():
     length = chord_length(radius, rho, phi)
     assert np.isclose(length, 0, atol=1e-15)
 
+    rho = 3
+    phi = np.deg2rad(180)
+    length = chord_length(radius, rho, phi)
+    assert np.isclose(length, 0, atol=1e-15)
```
mexanick, line 24: according to sonar, these two lines are not covered by tests:

```python
        if np.abs(phi) < np.arcsin(1.0 / rho):
            chord = 2 * radius * np.sqrt(chord)
```

Please check and ensure this is covered by proper tests.

src/ctapipe/image/muon/intensity_fitter.py
```diff
@@ -69,7 +69,11 @@ def chord_length(radius, rho, phi):
         chord = radius * (np.sqrt(chord) + rho * np.cos(phi))
     else:
         # muon did not hit the mirror
-        chord = 2 * radius * np.sqrt(chord)
+        # Filtering out non-physical solutions for phi
```
mexanick, line 72: I think this function would benefit from better documentation. First of all, I find confusing calling normalized discriminant as "chord" on line 61. I suggest to rename it to `discriminant_norm` or similar. Secondly, the chord length in geometrical understanding is only computed if the muon didn't hit the ring. Otherwise, it is only part of the chord until the point where a muon crosses the plain of the circle. I suggest to call it "visible_chord_length".

mexanick (changes requested):

src/ctapipe/image/muon/tests/test_intensity_fit.py
mexanick: your test verifies the part when condition `np.abs(phi) < np.arcsin(1.0 / rho)` is **not met**, but looks like it doesn't check when **it is met**.

mexanick: You have four cases in the function: no chord, a muon outside of the circle, and two cases for muon inside the circle. All of them have to be tested. Also, border cases (e.g. a hit straight in the center) and at the tangentially with φ=π/2 could be interesting to test. as well.

src/ctapipe/image/muon/intensity_fitter.py
burmist-git: The first point is addressed (discriminant_norm). However, I would refer to the second one as chord, consistent with the publications. This chord is calculated for round mirrors and round cameras. In the case of the camera, it should be called non_visible_chord_length.

src/ctapipe/image/muon/tests/test_intensity_fit.py
burmist-git: Yes and now we have 4 tests : 
image

Two inside and two outside

src/ctapipe/image/muon/intensity_fitter.py
mexanick: I still find it confusing, as it is not a chord in the geometrical meaning. You can rename it to "effective_chord" to be consistent with the paper. Also, add the reference to the paper and equation in it to the docstring.

src/ctapipe/image/muon/intensity_fitter.py
```diff
@@ -42,7 +42,12 @@
 @vectorize([double(double, double, double)], cache=True)
 def chord_length(radius, rho, phi):
     """
-    Function for integrating the length of a chord across a circle (effective chord length)
+    Function for integrating the length of a chord across a circle (effective chord length).
```
mexanick, line 45: Add a blank line between one-line summary and the body of the docstring. Add the paper to [references](https://github.com/cta-observatory/ctapipe/blob/main/docs/references.bib) if it is not yet there, and cite it similar to this: https://github.com/cta-observatory/ctapipe/blob/26ded3fc2e411f93fa5d1ab6f38ec3ddb86652f9/src/ctapipe/image/cleaning.py#L378C5-L380C73

mexanick:
I'm not sure why the sonar coverage is ill-computed... @kosack @maxnoe any idea? Could it be a consequence of a transfer to the new sonar instance?

@burmist-git I compiled the documentation and discovered that no sphinx doc is produced for the the module-level functions in the `ctapipe.image.muon.intensity_fitter` module. Please fix this by:
- add in `__all__` all the functions that aren't private 
- Update module-level docstring (I believe it is outdated)
- perhaps something else I missed - please verify that the doc assembles correctly

Also I noticed the a couple of warnings like the following one, all in the `ctapipe.image.muon.ring_fitter` module:

```
UserWarning: potentially wrong underline length... 
Returns 
-------- in 
Perform a circle fit to ``image`` with the chosen ``fit_method``.
... in the docstring of __call__ in /ctapipe/src/ctapipe/image/muon/ring_fitter.py.
```

Please fix them too.

burmist-git:
This is a separate issue that was not addressed previously. 
The goal of this merge request is clearly defined: to fix the muon fit chord length. As we agreed some time ago, we generally avoid making such requests, ok for for this MR - we have some time now.

burmist-git:
The docs are ok now ...

kosack:
Looks good. Just did a quick test to see the effect, and I see no issues!

Before fix, the left lobe are unphysical:

muon_cord_length_before

After fix:
muon_chord_length_after

mexanick (dismissed):
@kosack I just fixed the appearance of the chord_length function in the docs (it is a decorated function, required some tricks to properly appear in the toc tree), please re-approve.

maxnoe:
> I'm not sure why the sonar coverage is ill-computed... @kosack @maxnoe any idea? Could it be a consequence of a transfer to the new sonar instance?

We didn't yet switch to the new server here.

Why do you think this is ill-computed?

docs/api-reference/image/chord_length.rst
```diff
@@ -0,0 +1,10 @@
+.. _chord_length:
+
+
+.. currentmodule:: ctapipe.image.muon.intensity_fitter
+
+************
+chord_length
+************
+
+.. autofunction:: ctapipe.image.muon.intensity_fitter.chord_length
```
maxnoe, line 10: Please remove this change. The function does not appear in the docs because it is not included in `__all__` in the module. This is intentional, it's treated as an internal function of the `MuonIntensityFitter`, which is the only public API class in this module.

maxnoe (changes requested):
See inline comments.

docs/api-reference/image/chord_length.rst
maxnoe: THe file is also in the wrong directory.

mexanick: the function doesn't appear in the docs because it is decorated, not because it is not included in the `__all__`.
This can be removed if you insist, but I don't see much of good reasons to exclude it from the API docs.

maxnoe: E.g. this function is also numba jitted and appears just fine in the docs:
https://ctapipe.readthedocs.io/en/latest/_modules/ctapipe/image/extractor.html#neighbor_average_maximum

mexanick: By trial-and-error, and then confirmed by web search.

maxnoe:
Unfortunately this is a long-standing issue: https://github.com/numba/numba/issues/4268

mexanick (dismissed):

docs/api-reference/image/chord_length.rst
burmist-git: @maxnoe I’ve removed it; please let me know.

maxnoe:
@burmist-git sorry, I discussed with @mexanick offline and we agreed that the docs should stay and it's the correct workaround for an issue with numba: https://github.com/numba/numba/issues/5755

Could you revert the last two commits? This will also fix the ci

burmist-git:
@maxnoe Sure, I can do that. The only remaining question is the directory of chord_length.rst. It was here: docs/api-reference/image/chord_length.rst, but it probably should be here: docs/api-reference/image/muon/chord_length.rst.

maxnoe:
@burmist-git Why did you revert and then unrevert the changes?

docs/api-reference/image/muon.rst
```diff
@@ -6,6 +6,12 @@ Muon Image Analysis
 
 Algorithms for detecting and characterizing muon images
 
+.. toctree::
+  :maxdepth: 1
+  :glob:
+  :hidden:
+
+  chord_length
```
maxnoe, line 14:
```suggestion
  muon_chord_length
```

burmist-git:
@maxnoe The only valid option for me was to review changes, could not apply them.

maxnoe (approved):

mexanick (approved):

Issue #2392, opened 2023-09-04
Use correct random number gen, and make seeds reproducible.


kosack:
**Describe the bug**

Anywhere where random numbers are used should both use Numpy's new Generator API and also we should probably allow a **configurable seed** for reproducibility (and _write the used seed_ to the provenance output as well).

We could define some core "services" of Tools/Components that have a shared config, with a common random number generator being one of them. 

**Supporting information**

```
> ruff --select NPY check . --exclude=tests | grep random

ctapipe/fitting.py:79:14: NPY002 Replace legacy `np.random.choice` call with `np.random.Generator`
ctapipe/fitting.py:149:5: NPY002 Replace legacy `np.random.seed` call with `np.random.Generator`
ctapipe/image/modifications.py:73:9: NPY002 Replace legacy `np.random.seed` call with `np.random.Generator`
ctapipe/image/modifications.py:80:20: NPY002 Replace legacy `np.random.poisson` call with `np.random.Generator`
ctapipe/image/modifications.py:94:28: NPY002 Replace legacy `np.random.multinomial` call with `np.random.Generator`
```
**Additional context**

kosack:
I.e. there should be one random number generator with a configurable seed that can be used by all components.

maxnoe:
Unfortunately, it's not that easy.

The occurences you found are in numba jitted code, which just supports one global rng via those "legacy" functions.

kosack:
It does say this in the manual:
> Numba supports numpy.random.Generator() objects. As of version 0.56, users can pass individual NumPy Generator objects into Numba functions and use their methods inside the functions. The same algorithms are used as NumPy for random number generation hence maintaining parity between the random number generated using NumPy and Numba under identical arguments (also the same documentation notes as NumPy Generator methods apply). The current Numba support for Generator is not thread-safe, hence we do not recommend using Generator methods in methods with parallel execution logic.

So as long as we don't use `parallel=True`, it should be able to use a centrally configured Generator.  Haven't tested it though

maxnoe:
Ah, great, that's pretty new.

PR #2819, opened 2025-08-25
Use modern random interface where possible

maxnoe:
Performance of the lts fit is about the same:

```
❯ git switch main
Switched to branch 'main'
170 μs ± 235 ns per loop (mean ± std. dev. of 7 runs, 10,000 loops each)
❯ git switch numba-rng
Switched to branch 'numba-rng'
❯ ipython time_fit.py
169 μs ± 401 ns per loop (mean ± std. dev. of 7 runs, 10,000 loops each)
```

This is a partial fix for #2392, the image modification still uses the old interface as there is no support for multinomial yet in numba for `numpy.random.Generator`.

### DIFF
--- a/src/ctapipe/fitting.py
+++ b/src/ctapipe/fitting.py
@@ -2,6 +2,7 @@ import numpy as np
 from numba import njit
 
 EPS = 2 * np.finfo(np.float64).eps
+FIT_RNG = np.random.default_rng(0)
 
 
 @njit(cache=True)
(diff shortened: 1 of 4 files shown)


maxnoe:
Note: coverage is underreported to the changes mainly being in numba jitted code for which we do not get coverage information.

src/ctapipe/fitting.py
```diff
@@ -2,6 +2,7 @@
 from numba import njit
 
 EPS = 2 * np.finfo(np.float64).eps
+FIT_RNG = np.random.default_rng(0)
```
mexanick, line 5: This is not an example, right? Why the initialization is done with a fixed seed?

maxnoe: To make it reproducible by default.

src/ctapipe/fitting.py
```diff
@@ -72,9 +73,25 @@ def residuals(X, y, beta):
...
+    if values[1] >= values[0]:
+        values[1] += 1
+    return values
+
+
+@njit(cache=True)
+def _lts_single_sample(X, y, sample_size, max_iterations, rng, eps=1e-12):
     # randomly draw 2 points for the initial fit
-    sample = np.random.choice(len(y), 2, replace=False)
+    sample = choose_two_without_replacement(rng, len(y))
```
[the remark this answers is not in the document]
maxnoe: Yes, rng.choice is not yet supported by numba

src/ctapipe/fitting.py
```diff
@@ -102,9 +119,14 @@ def _lts_single_sample(X, y, sample_size, max_iterations, eps=1e-12):
...
-@njit()
 def lts_linear_regression(
-    x, y, samples=20, relative_sample_size=0.85, max_iterations=20, eps=1e-12
+    x,
+    y,
+    samples=20,
+    relative_sample_size=0.85,
+    max_iterations=20,
+    eps=1e-12,
+    rng=FIT_RNG,
```
kosack, line 129: this means that where this is used (e.g. in impact.py or image/timing.py) the RNG will always be initialized with a fixed `seed=0`, not `seed=None`.    It means the random sequence used will always be the same if re-run.   Is that an issue in real use?  It's nice for tests, but maybe not for reality?

maxnoe: The seed is the same for each start of ctapipe. So we also process a real data file re-producibly if the same config is used. I think this is the good default and doing otherwise could lead to very surprising behavior and hard to debug issues.

I don't think this is an issue, we are not creating simulations. We are sampling which points to use for a linear regression fit for the image parameters.

It's also not a behavior change (it just uses it's own RNG (already better!) than mutating a global numba-rng instance.

PR #2824, opened 2025-08-26
Implement reading and writing column metadata for tables. Fix reading of dl2 ml tel output in HDF5EventSource

### LINKED ISSUE #2761: Missing dl2 tables metadata
Reading in data from the `DL2_TELESCOPE_GROUP` with the `HDF5EventSource` leads to default filled/empty dl2 telescope containers for DL2 files written with the `ApplyModels` Tool. 
~I think the `HDF5TableReader.read()`-Method is missing a `prefixes=(algorithm + "_tel",)` argument in the `HDF5EventSource` generator for the DL2_TELESCOPE_GROUP just like its written in the DL2_SUBARRAY_GROUP here:~
https://github.com/cta-observatory/ctapipe/blob/e6eeae9747e96d4e021f661da36f80f4c5027694/src/ctapipe/io/hdf5eventsource.py#L600-L606

because for prefix=None the TableReader tries to map the containers field_names to the tables column_names via its column_attribute "NAME" which doesn't exist for the DL2 Tables not written with the DataWriter. So we should  add the missing NAME attribute ~or add the above mentioned prefix~.

/edit
Turns out that the "NAME" is not the only attribute missing in the tables metadata. The "TRANSFORM" attribute is missing too which  leads to a not writeable dl2/stereo table by the `DataWriter` after reading the data in with the `HDF5EventSource` because of its 'telescopes' field. But even with the "TRANSFORM" attribute i think the [`get_column_transforms`](https://github.com/cta-observatory/ctapipe/blob/02088f1c16e8ebc1ccb521d99a3de61f11ae8923/src/ctapipe/io/hdf5tableio.py#L135) method called in the `HDF5TableReader` is still missing the `TelListToMaskTransform`.  I think this also relates to #2711

maxnoe:
Fixes #2761

### DIFF
--- a/docs/changes/2824.feature.rst
+++ b/docs/changes/2824.feature.rst
@@ -1,1 +1,1 @@
-``astropy.io.read_table`` and ``astropy.io.write_table`` now support custom column metadata.
+``ctapipe.io.read_table`` and ``ctapipe.io.write_table`` now support custom column metadata.
--- a/docs/changes/2824.feature.rst
+++ b/docs/changes/2824.feature.rst
@@ -1,0 +1,1 @@
+``astropy.io.read_table`` and ``astropy.io.write_table`` now support custom column metadata.
--- a/src/ctapipe/io/astropy_helpers.py
+++ b/src/ctapipe/io/astropy_helpers.py
@@ -267,6 +267,8 @@ def _parse_hdf5_attrs(table):
         "POSINF_VALUE",
         "NEGINF_VALUE",
         "MAXLEN",
+        "POS",
+        "PREFIX",  # is computed from NAME, not needed
     }
     column_meta = {}
     for col_name, col_attrs in column_attrs.items():
--- a/src/ctapipe/io/astropy_helpers.py
+++ b/src/ctapipe/io/astropy_helpers.py
@@ -84,7 +84,7 @@ def read_table(
             raise OSError(
                 f"Node {path} is a {table.__class__.__name__}, must be a Table"
             )
-        transforms, descriptions, meta = _parse_hdf5_attrs(table)
+        transforms, descriptions, meta, column_meta = _parse_hdf5_attrs(table)
 
         if condition is None:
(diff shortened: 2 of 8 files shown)


docs/changes/2824.feature.rst
```diff
@@ -0,0 +1 @@
+``astropy.io.read_table`` and ``astropy.io.write_table`` now support custom column metadata.
```
Hckjs, line 1: io.astropy_helpers.read_table / write_table

Hckjs, line 1:
```suggestion
``io.astropy_helpers.read_table`` and ``io.astropy_helpers.write_table`` now support custom column metadata.
```

src/ctapipe/io/astropy_helpers.py
```diff
@@ -228,7 +251,29 @@ def _parse_hdf5_attrs(table):
...
+        "UNIT",
+        "TIME_FORMAT",
+        "TIME_SCALE",
+        "TRANSFORM",
+        "TRANSFORM_SCALE",
+        "TRANSFORM_OFFSET",
+        "NAN_VALUE",
+        "POSINF_VALUE",
+        "NEGINF_VALUE",
+        "MAXLEN",
```
Hckjs, line 269: meta

Additionally maybe "POS"?

Hckjs:
I just noticed, that  the "NAME" attribute for the columns "obs_id" and "event_id" is missing in the dl2 stereo tables, which could be fixed if we add the meta here:

https://github.com/cta-observatory/ctapipe/blob/7abcd4f2f3f31c6d51a33c05a39864890d1c0950/src/ctapipe/reco/stereo_combination.py#L278-L279

like
```
stereo_table[colname].meta = mono_predictions[colname].meta
```

maxnoe:
Columns that are stored without a prefix do not need the additional `NAME`

kosack (approved):

PR #2841, opened 2025-09-19
Return actual statistics in sigma clipping aggregator

### LINKED ISSUE #2834: SigmaClippingAggregator returns wrong sample size (n_events) value
`SigmaClippingAggregator` returns `n_events` corresponding to the sample it was provided with, but not the "clipped" one. Therefore, this value is not quite useful (can't be used e.g. to calculate standard error of the mean).

mexanick:
Fixes #2834

### DIFF
--- a/src/ctapipe/monitoring/aggregator.py
+++ b/src/ctapipe/monitoring/aggregator.py
@@ -18,7 +18,6 @@ from abc import abstractmethod
 from collections import defaultdict
 
 import numpy as np
-from astropy.stats import sigma_clipped_stats
 from astropy.table import Table
 
 from ctapipe.containers import StatisticsContainer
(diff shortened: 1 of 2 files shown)


src/ctapipe/monitoring/aggregator.py
```diff
@@ -167,16 +169,27 @@ def compute_stats(self, images, masked_pixels_of_sample) -> StatisticsContainer:
         masked_images = np.ma.array(images, mask=masked_pixels_of_sample)
 
         # Compute the mean, median, and std over the chunk per pixel
-        pixel_mean, pixel_median, pixel_std = sigma_clipped_stats(
+        # Use sigma_clip to get the clipped data, then compute stats from it
+        from astropy.stats import sigma_clip
```
maxnoe, line 173: Why move the import?

mexanick: we only use it here, but not really important in this context. Will move up.

TjarkMiener (approved):
Thanks!

Issue #2844, opened 2025-09-25
Update the chord_length function to use the phi0, add add the PixelShape for image_prediction


burmist-git:
MR https://github.com/cta-observatory/ctapipe/pull/2845#issue-3453363534

Update the chord_length function to use the phi0 parameter instead of including it inside the fit. And add the PixelShape for image_prediction, we fixed the normalization factor in this way. Add relevant tests. This helps with reusing the function.

Issue #2815, opened 2025-08-20
New method to reconstruct muon impact point


burmist-git:
We need to implement a new method to reconstruct the impact point.
With the current method, we achieve a 1.7 m resolution on the impact point (from simulation),
whereas ARTEMIS achieved 0.75 m. More details in MR #2814

maxnoe:
Could you please give a reference for the ARTEMIS algorithm / results and why you think their resolution is transferable to ctapipe / LST?

maxnoe:
You say sigma is currently about 1.7 m for LST, that's roughly equal to 2.3 * 0.75 m.

burmist-git:
It was decided to split one relatively big issue into separate ones.
- Update the chord_length function to use the phi0 parameter instead of including it inside the fit. And add the PixelShape for image_prediction, we fixed the normalization factor in this way. Add relevant tests. This helps with reusing the function.
- Add a second method and split into separate classes inheriting from a single base class.

PR #2845, opened 2025-09-25
Fix the normalization for muon analysis to account for different PixelShapes.

burmist-git:
issue : https://github.com/cta-observatory/ctapipe/issues/2844#issue-3453360123
Parent issue : https://github.com/cta-observatory/ctapipe/issues/2815

The goal is to improve the existing intensity fitter and prepare it for adding another method to measure the impact point and optical efficiency.

This pull request contains two main tasks:
- Include phi0 as a parameter in chord_length(radius, rho, phi) in image.muon.intensity_fitter:
- Fix the normalization to account for different PixelShapes. Add a function to calculate the trivial solution for the number of photons (from a muon) incident on the telescope mirror, and include it in the tests.

### DIFF
--- a/src/ctapipe/image/muon/intensity_fitter.py
+++ b/src/ctapipe/image/muon/intensity_fitter.py
@@ -39,7 +39,7 @@ CIRCLE_HEXAGON_AREA_RATIO = np.pi / 2 / np.sqrt(3)
 SQRT2 = np.sqrt(2)
 
 
-def chord_length(radius, impact_point, phi, phi0=0):
+def chord_length(radius, impact_distance, phi, phi0=0):
     """
     Function for integrating the length of a chord across a circle (effective chord length).
 
@@ -49,8 +49,8 @@ def chord_length(radius, impact_point, phi, phi0=0):
     ----------
     radius: float or ndarray
         radius of circle
-    impact_point: float or ndarray
-        distance of impact point from circle center
+    impact_distance: float or ndarray
+        distance of impact distance from circle center
     phi: float or ndarray in radians
         rotation angles to calculate length
     phi0: float or ndarray in radians
@@ -71,13 +71,13 @@ def chord_length(radius, impact_point, phi, phi0=0):
 
     """
 
-    return _chord_length(radius, impact_point, phi, phi0)
+    return _chord_length(radius, impact_distance, phi, phi0)
 
 
 @vectorize(
     [double(double, double, double, double)], cache=not CTAPIPE_DISABLE_NUMBA_CACHE
 )
-def _chord_length(radius, impact_point, phi, phi0):
+def _chord_length(radius, impact_distance, phi, phi0):
     if radius <= 0:
         return 0
 
@@ -85,7 +85,7 @@ def _chord_length(radius, impact_point, phi, phi0):
 
     phi_modulo = (phi + np.pi) % (2 * np.pi) - np.pi
 
-    rho = np.abs(impact_point) / radius
+    rho = np.abs(impact_distance) / radius
     discriminant_norm = 1 - (rho**2 * np.sin(phi_modulo) ** 2)
     if discriminant_norm < 0:
         return 0


src/ctapipe/image/muon/intensity_fitter.py
```diff
@@ -31,12 +32,17 @@
...
 # Sqrt of 2, as it is needed multiple times
 SQRT2 = np.sqrt(2)
 
 
-@vectorize([double(double, double, double)], cache=not CTAPIPE_DISABLE_NUMBA_CACHE)
-def chord_length(radius, rho, phi):
+@vectorize(
+    [double(double, double, double, double)], cache=not CTAPIPE_DISABLE_NUMBA_CACHE
+)
+def chord_length(radius, rho, phi0, phi):
```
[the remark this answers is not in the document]
maxnoe: Not just the order, also the meaning.

Rho is now absolute instead of a fraction of the radius. Why? That seems like completely needless breaking change, just moving one division or multiplication around.

src/ctapipe/image/muon/intensity_fitter.py
```diff
@@ -64,29 +70,42 @@ def chord_length(radius, rho, phi):
...
-    valid = discriminant_norm >= 0
 
-    if not valid:
+    if radius <= 0:
+        return 0
+
+    phi = phi - phi0
+
+    phi_modulo = (phi + np.pi) % (2 * np.pi) - np.pi
+    if phi < 0:
```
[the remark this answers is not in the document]
burmist-git: If you are commenting about this line:  
```phi_modulo *= -1```
 then it is correct. Since modulo returns sign less value and we need to return the sign minus back.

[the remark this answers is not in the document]
Comment (maxnoe) [reply src/ctapipe/image/muon/intensity_fitter.py:82]: Isn't reflecting at the origin just wrong here? Shouldn't you add 2π instead?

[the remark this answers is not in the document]
Comment (burmist-git) [reply src/ctapipe/image/muon/intensity_fitter.py]: This is unnecessary
```phi_modulo *= -1```

Here is the test without it

image

File: src/ctapipe/image/muon/intensity_fitter.py [hunk 3]
```diff
@@ -64,29 +70,47 @@ def chord_length(radius, rho, phi):
...
-            effective_chord_length = 2 * radius * np.sqrt(discriminant_norm)
-        else:
+        if np.abs(phi_modulo) > np.arcsin(1.0 / rho_r):
             return 0
 
     return effective_chord_length
 
 
-def intersect_circle(mirror_radius, r, angle, hole_radius=0):
+def intersect_circle(mirror_radius, r, phi0, angle, hole_radius=0):
```
mexanick, line 113: phi0 is not used in the function, so there's no need (at least at the moment) to add it.
Also, when adding, considering adding it as a last-order keyword argument to preserve the legacy calls.

burmist-git: this is corrected

burmist-git:
@maxnoe Any feed back on this MR ?

src/ctapipe/image/muon/intensity_fitter.py
```diff
@@ -64,23 +70,41 @@ def chord_length(radius, rho, phi):
...
+def _chord_length(radius, rho, phi, phi0):
+    if radius <= 0:
         return 0
 
-    if rho <= 1.0:
+    phi = phi - phi0
+
+    phi_modulo = (phi + np.pi) % (2 * np.pi) - np.pi
+
+    rho_r = np.abs(rho) / radius
```
maxnoe, line 88: What is the motivation for this redefinition? It seems like an arbitrary choice, so why change it now?

Please revert.

src/ctapipe/image/muon/intensity_fitter.py
```diff
@@ -317,7 +344,12 @@ def image_prediction_no_units(
     # diameter. In any case, since in the end we do a data-MC comparison of the muon
     # ring analysis outputs, it is not critical that this value is exact.
```
maxnoe, line 345: The comment here provided by @moralejo is now longer correct, since you introduce the correction for the pixel shape.

Please update it.

Please also have a look here @moralejo.

src/ctapipe/image/muon/intensity_fitter.py
burmist-git: @maxnoe  concerning this : https://github.com/cta-observatory/ctapipe/pull/2845#discussion_r2405390011
here is the reason why :
image

src/ctapipe/image/muon/intensity_fitter.py
```diff
@@ -49,7 +49,7 @@ def chord_length(radius, rho, phi, phi0=0):
     ----------
     radius: float or ndarray
         radius of circle
-    rho: float or ndarray
+    impact_point: float or ndarray
```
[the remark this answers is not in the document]
burmist-git: Dear @maxnoe, this is a friendly reminder about this MR — let’s merge it :-)

src/ctapipe/image/muon/intensity_fitter.py
```diff
@@ -152,7 +176,7 @@ def create_profile(
     circumference = 2 * np.pi * radius
     pixels_on_circle = int(circumference / pixel_diameter)
 
-    ang = phi + linspace_two_pi(pixels_on_circle * oversampling)
+    ang = linspace_two_pi(pixels_on_circle * oversampling) - phi
```
maxnoe: Please explain this change. Why do you change the order of this array?

src/ctapipe/image/muon/intensity_fitter.py
```diff
@@ -267,7 +294,7 @@ def image_prediction_no_units(
     dy = pixel_y_rad - center_y_rad
     ang = np.arctan2(dy, dx)
     # Add muon rotation angle
-    ang += phi_rad
+    ang -= phi_rad
```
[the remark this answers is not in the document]
burmist-git: The answer is here : https://github.com/cta-observatory/ctapipe/pull/2845/files/397903ad39453ed0227fb8928d124e4b95c26e92#r2444298627

src/ctapipe/image/muon/intensity_fitter.py
maxnoe: And: if you are changing the coordinate system of the muon fit, don't you think this is a major change that needs to be mentioned both in the motivation for the pull request and in the changelog?

Please, explain what you are doing so that people can actually review and understand.

burmist-git: Please note that phi0 is set to 0. As we agree do not change the current fitter.

burmist-git: Well reuse this one for Fix the normalization for muon analysis to account for different PixelShapes.
I will rename the MR.

PR #2847, opened 2025-09-26
Waveform noise adder

moralejo:
First version of the component to add noise in MC waveforms, for adapting it to different observation conditions.
It adds at random N instances of waveforms from an NSB-only file to the waveforms of a showers file. ~~It seems to reproduce well the noise distributions produced directly by sim_telarray:~~

Update: there was a mistake in the plots below (only one telescope was used, multiple times). After correcting this, it turns out that the noise padding produces slightly larger noise than what is produced by sim_telarray itself:

image

image

~~Test function still not ready,~~ I open the draft PR to start discussion.

### DIFF
--- a/src/ctapipe/image/modifications.py
+++ b/src/ctapipe/image/modifications.py
@@ -125,7 +125,7 @@ class NoiseEventTypeFilter(EventTypeFilter):
 
 @njit(cache=not CTAPIPE_DISABLE_NUMBA_CACHE)
 def build_wf_noise_pixelwise(
-    waveforms, n_noise_realizations, nsb_level, rng, shuffle_full_cameras
+    waveforms, n_noise_realizations, nsb_level, rng, sample_pixels_independently
 ):
     """
     Combine "elemental noise waveforms" into total noise waveforms by
@@ -145,10 +145,12 @@ def build_wf_noise_pixelwise(
 
     rng: random number generator
 
-    shuffle_full_cameras: bool
-    if True, the waveform for each pixel in any given noise realization comes
-    from the same combination of the input elemental noise events. If False,
-    each pixel uses a different combination of the events
+    sample_pixels_independently: bool
+    if True, each pixel will use a different random combination of the elemental
+    noise events to build the noise to be added to its waveform.
+    if False, the waveform for each pixel in any given noise realization comes
+    from the same random combination of the input elemental noise events.
+
 
     Returns
     -------
@@ -161,11 +163,7 @@ def build_wf_noise_pixelwise(
     )
 
     for i in range(n_noise_realizations):
-        if shuffle_full_cameras:
-            chosen = rng.permutation(n_events)[:nsb_level]
-            for event in chosen:
-                noise[i] += waveforms[event]
-        else:
+        if sample_pixels_independently:
             for pixel in range(n_pixels):
                 chosen = rng.permutation(n_events)[:nsb_level]
                 # The line above is slower (especially for n_events much
@@ -174,6 +172,11 @@ def build_wf_noise_pixelwise(
 
                 for event in chosen:
                     noise[i, :, pixel] += waveforms[event, :, pixel]
+        else:
+            chosen = rng.permutation(n_events)[:nsb_level]
+            for event in chosen:
+                noise[i] += waveforms[event]
+
     return noise
 
 
@@ -198,8 +201,8 @@ class WaveformModifier(TelescopeComponent):
         whereas the nsb file is a real data DL0 file from which only the
         interleaved pedestals are used (all gain channels must be present
         for all pixels). In that case, nsb_level must be =1 (to
-        match the MC to the data) and shuffle_full_cameras=True (we do not
-        want e.g. to duplicate stars in the FoV).
+        match the MC to the data) and sample_pixels_independently=False
+        (we do not want e.g. to duplicate stars in the FoV).
 
 
     In case (1), the number of available noise events per telescope in the NSB
@@ -234,14 +237,15 @@ class WaveformModifier(TelescopeComponent):
         ),
     ).tag(config=True)
 
-    shuffle_full_cameras = Bool(
-        default_value=False,
+    sample_pixels_independently = Bool(
+        default_value=True,
         help=(
-            "If True, full cameras are "
-            "combined to generate noise "
-            "waveforms for all pixels. Else, "
-            "each pixel uses a different random "
-            "combination of the input noise events"
+            "If True, each pixel uses a different "
+            "random combination of the input noise "
+            "events"
+            "If False, all pixels use the same random "
+            "combination of noise events. That is, noise "
+            "events will be combined as full cameras"
         ),
     ).tag(config=True)
 
@@ -305,7 +309,7 @@ class WaveformModifier(TelescopeComponent):
                 self.n_noise_realizations,
                 self.nsb_level,
                 self.rng,
-                self.shuffle_full_cameras,
+                self.sample_pixels_independently,
             )
 
     def read_nsb_database(self):


moralejo:
The failures in the tests seems to be unrelated to this. Any idea of what is going on?

moralejo:
> The connection timeout is unrelated and should go away be re-running, but this error is in code you added:
> 
> ```
> src/ctapipe/tests/test_traitlets_configurable.py::test_all_traitlets_configurable - AssertionError: Class ctapipe.image.modifications.WaveformModifier is missing .tag(config=True) for traitlets: {'total_noise'}
> ```

Thanks, I had not noticed, because test ran well locally. Anyway it is not something that should be configured, it is something internally, so I just made a normal dict()

src/ctapipe/image/modifications.py
```diff
@@ -100,6 +106,179 @@ def _smear_psf_randomly(
...
+        self.rng = np.random.default_rng(self.rng_seed)
+
+        # Read in the waveforms in the NSB-only file. Store in a dictionary
+        # with one key per telescope, containing an array [n_events, n_gains,
+        # n_pixels, n_samples]
+        source = EventSource(input_url=self.nsb_file, skip_calibration_events=False)
+        nsb_database = dict()  # [nevents, ngains, npixels, nsamples]
+        for event in source:
+            for tel_id in event.trigger.tels_with_trigger:
+                if tel_id in nsb_database:
```
maxnoe, line 187: much cheaper and simpler to just accumulate into a list and then stack once in the end.

Also doesn't need the if.

src/ctapipe/image/modifications.py
```diff
@@ -100,6 +108,177 @@ def _smear_psf_randomly(
     return new_image
 
 
+class WaveformModifier(TelescopeComponent):
+    """
+    Component to add NSB noise to R1 waveforms (intended in principle to be
+    applied on MC simulations, to make them closer to real data)
+    The noise waveforms are read from a dedicated sim_telarray file, which
```
mexanick, line 115: After a brief look over the code, it seems we are not bound to simulations as a noise source.
I would change `from a dedicated sim_telarray file` to `from a dedicated NSB noise file (simulated or pedestal calibration)` and add event type filtering to select pedestal events (we are not 100% sure whether pedestal and flasher will come in the same or in different files).

src/ctapipe/image/modifications.py
```diff
@@ -100,6 +109,182 @@ def _smear_psf_randomly(
...
+        super().__init__(subarray=subarray, config=config, parent=parent, **kwargs)
+        self.rng = np.random.default_rng(self.rng_seed)
+
+        # Read in the waveforms in the NSB-only file. Store in a dictionary
+        # with one key per telescope, containing an array [n_events, n_gains,
+        # n_pixels, n_samples]
+        source = EventSource(input_url=self.nsb_file, skip_calibration_events=False)
+        nsb_database = defaultdict(list)  # [nevents, ngains, npixels, nsamples]
+        for event in source:
+            if event.trigger.event_type != EventType.SKY_PEDESTAL:
```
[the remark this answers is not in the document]
mexanick: yes, this is what I meant, but I would either pass `EventType.<Type>` as an argument of the `__call__`, or add it to the configurable traits, instead of passing as an additional parameter to the constructor.

src/ctapipe/image/modifications.py
```diff
@@ -100,6 +111,193 @@ def _smear_psf_randomly(
...
+        parent: ctapipe.core.Component or ctapipe.core.Tool
+            Parent of this component in the configuration hierarchy,
+            this is mutually exclusive with passing ``config``
+
+        """
+
+        super().__init__(subarray=subarray, config=config, parent=parent, **kwargs)
+        self.rng = np.random.default_rng(self.rng_seed)
+
+        event_type_filter = EventTypeFilter(allowed_types=self.nsb_event_types)
```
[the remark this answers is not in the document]
maxnoe: The line here should be:
```
event_type_filter = EventTypeFilter(parent=self)
```
but what we want is a different default, right? That's unfortunately not easy... so maybe what you have here is best, actually duplicating the config.

You should still pass `parent=self` though, as that also sets up the logging correctly.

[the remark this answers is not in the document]
moralejo: > The line here should be:
> 
> ```
> event_type_filter = EventTypeFilter(parent=self)
> ```
> 
> but what we want is a different default, right? That's unfortunately not easy... 

Exactly, that is what I failed to do. I want to have the default, because it actually should very rarely (if at all) be changed.

> 
> You should still pass `parent=self` though, as that also sets up the logging correctly.

Ok!

src/ctapipe/image/modifications.py
```diff
@@ -100,6 +111,195 @@ def _smear_psf_randomly(
     return new_image
 
 
+class WaveformModifier(TelescopeComponent):
+    """
+    Component to add NSB noise to R1 waveforms (intended in principle to be
```
maxnoe, line 116: First line of the docstring should be a title line.

Something like:

```
Component to add NSB noise to R1 waveforms.

This component is intended to ...
```

src/ctapipe/image/tests/test_modifications.py
```diff
@@ -1,7 +1,63 @@
...
+                original_wfs[tel_id].append(event.r1.tel[tel_id].waveform)
+            else:
+                original_wfs[tel_id] = [event.r1.tel[tel_id].waveform]
+
+            if tel_id in modified_wfs:
+                modified_wfs[tel_id].append(
+                    wf_mod(tel_id, event.r1.tel[tel_id].waveform)
+                )
+            else:
+                modified_wfs[tel_id] = [wf_mod(tel_id, event.r1.tel[tel_id].waveform)]
```
kosack, line 37: Should use a with-block here, otherwise the `EventSource` is never closed. You only need it around the part where you read the events:

```suggestion
    with EventSource(nsb_only_file, skip_calibration_events=False) as source:
        wf_mod = WaveformModifier(
            subarray=source.subarray, nsb_file=nsb_only_file, nsb_level=3
        )
        original_wfs = dict()
        modified_wfs = dict()

        # We test adding the noise on the same file from which the noise waveforms
        # are taken:
        for event in source:
            for tel_id in event.trigger.tels_with_trigger:
                if tel_id in original_wfs:
                    original_wfs[tel_id].append(event.r1.tel[tel_id].waveform)
                else:
                    original_wfs[tel_id] = [event.r1.tel[tel_id].waveform]

                if tel_id in modified_wfs:
                    modified_wfs[tel_id].append(
                        wf_mod(tel_id, event.r1.tel[tel_id].waveform)
                    )
                else:
                    modified_wfs[tel_id] = [wf_mod(tel_id, event.r1.tel[tel_id].waveform)]

```

src/ctapipe/image/modifications.py
```diff
@@ -100,6 +111,195 @@ def _smear_psf_randomly(
...
+        self.rng = np.random.default_rng(self.rng_seed)
+
+        event_type_filter = EventTypeFilter(
+            allowed_types=self.nsb_event_types, parent=self
+        )
+
+        # Read in the waveforms in the NSB-only file. Store in a dictionary
+        # with one key per telescope, containing an array [n_events, n_gains,
+        # n_pixels, n_samples]
+        source = EventSource(input_url=self.nsb_file, skip_calibration_events=False)
```
kosack, line 204: Please use the EventSource as a context manager here, otherwise it remains open indefinitely:
e.g.:
```py3
with EventSource(input_url=self.nsb_file, skip_calibration_events=False) as source:
   ...
```

src/ctapipe/image/modifications.py
```diff
@@ -100,6 +111,195 @@ def _smear_psf_randomly(
...
+            "Number of different realizations of the total NSB waveform to "
+            "be created per pixel"
+        ),
+    ).tag(config=True)
+
+    rng_seed = Int(default_value=1, help="Seed for the random number generator").tag(
+        config=True
+    )
+
+    nsb_event_types = List(
```
kosack, line 160: EventTypeFilter already has a config option called `allowed_types`, so this is not needed here if the filter is constructed inside this class.  See comment below about the construction of the filter:

src/ctapipe/image/modifications.py
```diff
@@ -100,6 +111,195 @@ def _smear_psf_randomly(
...
+            Parent of this component in the configuration hierarchy,
+            this is mutually exclusive with passing ``config``
+
+        """
+
+        super().__init__(subarray=subarray, config=config, parent=parent, **kwargs)
+        self.rng = np.random.default_rng(self.rng_seed)
+
+        event_type_filter = EventTypeFilter(
+            allowed_types=self.nsb_event_types, parent=self
```
kosack, line 198: This shadows the `WaveformModifier.EventTypeFilter.allowed_types` setting in the config. Probably best to just use that and not its own setting? Otherwise, might cause confusion.  Generally there is no reason to pass in settings explicitly inside a Component.

src/ctapipe/image/modifications.py
```diff
@@ -100,6 +111,195 @@ def _smear_psf_randomly(
...
+            # [n_noise_realizations, ngains, npixels, nsamples]
+
+            nevents = nsb_database[tel_id].shape[0]
+            for jj in range(self.n_noise_realizations):
+                # We take a random combination of a number self.nsb_level of
+                # instances of the noise events. The combination of events is
+                # common to the whole camera (making it different for each
+                # pixel might be better, but it is very slow - and I failed
+                # to do it using numba due to some numba limitations)
+                shuffled_event_indices = self.rng.permutation(nevents)[: self.nsb_level]
```
kosack, line 260: What about something like:
```python
suffled_event_indices = rng.choice(n_events, size=(self.n_noise_realizations, self.nsb_level))
```
That gives you a numpy array of the indices for each permutation. That is much faster and could be used to get a different shuffling per pixel if you wanted.   Since this is not run frequently, however, speed is not a giant concern.

src/ctapipe/image/modifications.py
moralejo: > This shadows the `WaveformModifier.EventTypeFilter.allowed_types` setting in the config. Probably best to just use that and not its own setting? Otherwise, might cause confusion. Generally there is no reason to pass in settings explicitly inside a Component.

The problem is I want to have a default value for allowed_types, specific to this class. See comment above by Max: https://github.com/cta-observatory/ctapipe/pull/2847#discussion_r2394359469, apparently that is not simple

moralejo: Alternatively, I could go back to not using EventTypeFilter, and keep the list of allowed types specific to this class.

kosack: Ah, yes, that's an issue since we can't change defaults.  What I did for other classes like QualityQuery is to create a blank subclass with the new defaults, e.g.:

```
class NoiseEventTypeFilter(EventTypeFilter):
    allowed_types =  allowed_types = Set(
        trait=UseEnum(EventType),
        default_value=[EventType.SKY_PEDESTAL],
        allow_none=True,
        help=(
            "The allowed types. Set to None to allow all types."
            f"Possible values: {_values}."
        ),
    ).tag(config=True)
```

Then use that class.  I think the config sustem should still then accept either `WaveformModifier.EventTypeFilter.allowed_types` or `WaveformModifier.NoiseEventTypeFilter.allowed_types` in your config file.

maxnoe: I think the simpler solution here would be using a default function to override the default instead of redefining the tratilet:

```
class NoiseEventTypeFilter(EventTypeFilter):
    @default
     def allowed_types_default(self):
        return {EventType.SKY_PEDESTAL, EventType.PEDESTAL}
```

src/ctapipe/image/modifications.py
maxnoe: What is `n_noise_realizations`? You pre-compute the full noise waveforms before-hand?

src/ctapipe/image/modifications.py
```diff
@@ -100,6 +111,201 @@ def _smear_psf_randomly(
...
+        event_type_filter = EventTypeFilter(
+            allowed_types=self.nsb_event_types, parent=self
+        )
+
+        # Read in the waveforms in the NSB-only file. Store in a dictionary
+        # with one key per telescope, containing an array [n_events, n_gains,
+        # n_pixels, n_samples]
+        nsb_database = defaultdict(list)  # [nevents, ngains, npixels, nsamples]
+        with EventSource(
+            input_url=self.nsb_file, skip_calibration_events=False
```
maxnoe, line 208: Do we need to switch off gain selection here to build the noise database?

src/ctapipe/image/modifications.py
maxnoe: For 100 noise events out of 10000 events, this takes 2 seconds. Vectorizing the sum takes it down to 1.4:

```
for i in range(n_noise):
    for pixel in range(n_pixels):
        chosen = rng.choice(n_events, size=nsb_level, replace=False)
        noise[i, :, pixel] += waveforms[chosen, :, pixel].sum(axis=0)
```

src/ctapipe/image/modifications.py
moralejo: I adopted this solution, I think now it can be set from config, i.e. I tested that this works:
```
config=Config({"WaveformModifier":{"NoiseEventTypeFilter":{"allowed_types":[EventType.FLATFIELD]}}})
wfm = WaveformModifier(subarray=source.subarray, nsb_file=nsb_only_file, nsb_level=3, config=config)
```

src/ctapipe/image/modifications.py
moralejo: Ah, ok, indeed it is the same, I thought you were producing as many noise realizations as the number of "elemental noise events".
For the record: n_events is the number of "elemental" noise events we have, n_noise_realizations is the number of combinations of them we want to make, nsb_level the number of elemental waveforms we combine in one of the total_noise waveforms.

moralejo: Will try it.

src/ctapipe/image/modifications.py
```diff
@@ -100,6 +111,205 @@ def _smear_psf_randomly(
...
+            raise ValueError("Please use an input nsb_file with more events!")
+
+        # Now shift the waveforms so that they have mean=0 and do not introduce
+        # any bias (just fluctuations). For each telescope and gain we average
+        # all pixels
+        for tel_id in nsb_database:
+            hg_mean = np.mean(nsb_database[tel_id][:, 0, :, :])  # HG
+            lg_mean = np.mean(nsb_database[tel_id][:, 1, :, :])  # LG
+            nsb_database[tel_id][:, 0, :, :] -= hg_mean
+            nsb_database[tel_id][:, 1, :, :] -= lg_mean
```
maxnoe, line 251:
```suggestion
            for channel in nsb_database[tel_id].shape[0]:
                mean = np.mean(nsb_database[tel_id][:, channel, :, :])
                nsb_database[tel_id][:, channel, :, :] -= mean
```
otherwise it won't work for single-gain cameras.

maxnoe:
This looks good in general. However, at the moment, it's not usable via `ctapipe-process`.

I think there are two options:
- Adding it in `ProcessorTool` directly as its own step
- Adding it to it to the `CameraCalibrator` class in the r1 to dl0 step.

I think this does not really belong into the `CameraCalibrator`, so I would add it directly to the process tool.

Similar to how the `ImageProcessor` is using the `ImageModifier` here:

https://github.com/cta-observatory/ctapipe/blob/04a1ec7b268c1c186571febac1455aa8ef57f23f/src/ctapipe/image/image_processor.py#L217-L219

src/ctapipe/image/modifications.py
```diff
@@ -100,6 +112,256 @@ def _smear_psf_randomly(
...
+    ).tag(config=True)
+
+    shuffle_full_cameras = Bool(
+        default_value=False,
+        help=(
+            "If True, full cameras are "
+            "combined to generate noise "
+            "waveforms for all pixels. Else, "
+            "each pixel uses a different random "
+            "combination of the inmput noise events"
```
maxnoe, line 231:
```suggestion
            "combination of the input noise events"
```

src/ctapipe/image/modifications.py
```diff
@@ -100,6 +112,256 @@ def _smear_psf_randomly(
...
+        ),
+    ).tag(config=True)
+
+    rng_seed = Int(default_value=1, help="Seed for the random number generator").tag(
+        config=True
+    )
```
maxnoe, line 237:
```suggestion
    rng_seed = Int(
        default_value=1,
        help="Seed for the random number generator",
    ).tag(config=True)
```

src/ctapipe/image/modifications.py
```diff
@@ -100,6 +112,256 @@ def _smear_psf_randomly(
...
+
+    def __init__(
```
maxnoe, line 239: I think this `__init__` would benefit by being broken down into several smaller functions, e.g.:

```python

nsb_database = self.load_noise_events()
self.check_noise_statistics(nsb_database)
self.normalize_nsb_database(nsb_database)
self.nsb_database = self.stack_noise_waveforms(nsb_database)
```

Also, after the `total_noise` is computed, there is no need to keep the `nsb_database` around right?

src/ctapipe/image/modifications.py
```diff
@@ -100,6 +112,256 @@ def _smear_psf_randomly(
...
+        # make sense to add the noise after DVR was applied)
+        if selected_gain_channel is None:
+            return (
+                waveforms
+                + self.total_noise[tel_id][self.rng.integers(self.n_noise_realizations)]
+            )
+        else:
+            return (
+                waveforms
+                + self.total_noise[tel_id][
```
maxnoe, line 357: This if else is a bit hard to read and see the differences.

Maybe it would be simpler like this:

```python
if selected_gain_channel is None:
    noise = ...
else:
   noise = ...

return waveforms + noise
```

src/ctapipe/image/modifications.py
```diff
@@ -100,6 +112,313 @@ def _smear_psf_randomly(
...
+    Event filter to select noise events for MC tuning
+    By default it selects SKY_PEDESTAL events
+    """
+
+    @default("allowed_types")
+    def allowed_types_default(self):
+        return {EventType.SKY_PEDESTAL}
+
+
+@njit
```
maxnoe:
```suggestion
@njit(cache=NOT CTAPIPE_DISABLE_NUMBA_CACHE)
```

PR #2848, opened 2025-09-29
Improve statistics aggregators

mexanick:
- Relax requirements on the data shape (aggregate arbitrary arrays on axis 0)
- Add option to aggregate statistics by time


src/ctapipe/monitoring/aggregator.py
```diff
@@ -1,159 +1,390 @@
...
+        ),
+    ).tag(config=True)
+
+    chunking_mode = CaselessStrEnum(
+        ["events", "time"],
+        default_value="events",
+        help=(
+            "Chunking strategy: 'events' for fixed event count, 'time' for time intervals. "
+            "When 'time', chunk_size and chunk_shift are interpreted as seconds."
+        ),
```
TjarkMiener, line 60: I'd slightly prefer to have two config parameters, something like:
```
chunk_size = Int(
        default_value=None,
        allow_none=True,
        help=(
            "Event-based size of the chunk (number of events per chunk) "
            "used for the computation of aggregated statistic values. "
        ),
    ).tag(config=True) 

chunk_duration = AstroQuantity(
        default_value=None,
        allow_none=True,
        physical_type=astropy.units.physical.time,
        help=(
            "Time-based size of the chunk (duration in 'astropy.units.physical.time' per chunk)
            "used for the computation of aggregated statistic values. "
        ), 
   ).tag(config=True)
```
And then follow the logic: If both `None`, use entire table as one chunk. If both set, break. And if either set take the one set.

src/ctapipe/monitoring/aggregator.py
```diff
@@ -1,159 +1,390 @@
...
         table : astropy.table.Table
-            table with images of shape (n_images, n_channels, n_pixels), event IDs and
-            timestamps of shape (n_images, )
-        masked_pixels_of_sample : ndarray, optional
-            boolean array of masked pixels of shape (n_channels, n_pixels) that are not available for processing
+            table with event-wise data of shape (n_events, \*data_dimensions), event IDs and
+            timestamps of shape (n_events, )
+        masked_elements_of_sample : ndarray, optional
+            boolean array of masked elements of shape (\*data_dimensions) that are not available for processing
         chunk_shift : int, optional
```
TjarkMiener, line 89: this could be `int` or `astropy.units.Quantity`, optional

src/ctapipe/monitoring/aggregator.py
```diff
@@ -1,159 +1,390 @@
...
-
-        # Check if the statistics of the table is sufficient to compute at least one complete chunk.
-        if len(table) < self.chunk_size:
-            raise ValueError(
-                f"The length of the provided table ({len(table)}) is insufficient to meet the required statistics for a single chunk of size ({self.chunk_size})."
+        # Set chunk_size to table length if None (use entire table as one chunk)
+        if self.chunk_size is None:
+            effective_chunk_size = len(table)
+            self.chunking_mode = (
+                "events"  # Force event-based chunking when using entire table
```
TjarkMiener, line 106: this force would be redundant with suggested config change above

src/ctapipe/monitoring/aggregator.py
```diff
@@ -1,159 +1,390 @@
...
+        total_duration = (end_time - start_time).to_value("s")
+
+        # Calculate time step: Use chunk_shift if provided, otherwise use chunk_size
+        time_step = chunk_shift if chunk_shift is not None else effective_chunk_size
+
+        current_time = start_time
+        while True:
+            chunk_end = current_time + effective_chunk_size * u.s
+            # Check if chunk end exceeds our data range (with 0.1s tolerance for floating point)
+            if chunk_end > end_time + 0.1 * u.s:
```
TjarkMiener, line 274: I'd assume that this tolerance depends on your use case. I'd pass it as an argument in the `__call__()`

src/ctapipe/monitoring/aggregator.py
```diff
@@ -1,159 +1,390 @@
...
+        ),
     ).tag(config=True)
 
     def __call__(
         self,
         table,
-        masked_pixels_of_sample=None,
+        masked_elements_of_sample=None,
         chunk_shift=None,
         col_name="image",
```
TjarkMiener, line 68: see below: I'd add `time_tolerance = 0.1 * u.s` to customise based on the use case.

src/ctapipe/monitoring/aggregator.py
```diff
@@ -1,159 +1,390 @@
...
+        if self.chunk_size is None:
+            yield table
+            return
+
+        if self.chunking_mode == "events":
+            yield from self._get_event_chunks(table, chunk_shift, effective_chunk_size)
+        elif self.chunking_mode == "time":
+            yield from self._get_time_chunks(table, chunk_shift, effective_chunk_size)
+
+    def _check_table_length(self, table, effective_chunk_size):
```
TjarkMiener, line 227: I'd consider moving this check into `_get_event_chunks()` and `_get_time_chunks()` and remove this extra function.

src/ctapipe/monitoring/aggregator.py
```diff
@@ -1,159 +1,390 @@
...
+        effective_chunk_size : int
+            Size of each chunk in events or seconds
+
+        Yields
+        ------
+        astropy.table.Table
+            Chunks of the input table
+        """
+        # If using entire table as one chunk, just yield the whole table
+        if self.chunk_size is None:
```
TjarkMiener, line 218: this would then be `if self.chunk_size is None and if self.chunk_duration is None:`

TjarkMiener (commented):
Looks already good to me, @mexanick! I'm attaching some comments and suggestions.

src/ctapipe/monitoring/aggregator.py
```diff
@@ -1,159 +1,390 @@
...
+        # Preserve units if present
+        if hasattr(table[col_name], "unit") and table[col_name].unit is not None:
+            self._set_result_units(result_table, table[col_name].unit)
+
+        return result_table
 
     @abstractmethod
-    def compute_stats(self, images, masked_pixels_of_sample) -> StatisticsContainer:
+    def _add_result_columns(self, data, masked_elements_of_sample, results_dict):
+        """
```
matteobalbo, line 147:
```suggestion
        r"""
```

src/ctapipe/monitoring/aggregator.py
```diff
@@ -1,163 +1,402 @@
...
+            raise ValueError(f"Table must have columns: {missing}")
+
+    @abstractmethod
+    def _generate_chunks(self, table) -> Generator[Table, None, None]:
+        """Generate chunks from table. Implemented by subclasses."""
+        pass
+
+
+class SizeChunking(BaseChunking):
+    """Divides tables into chunks based on number of events."""
```
maxnoe: Do we only plan to use this with events? Or might it also be used for a monitoring table of something other than events?

```suggestion
    """Divides tables into chunks based on number of samples."""
```

src/ctapipe/monitoring/aggregator.py
```diff
@@ -1,163 +1,402 @@
...
+    def _generate_chunks(self, table) -> Generator[Table, None, None]:
+        """Generate event-count based chunks."""
+        # Handle case where chunk_size is None (entire table)
+        if self.chunk_size is None:
+            yield table
+            return
+
+        # Validate chunk_size vs table length
+        if len(table) < self.chunk_size:
+            raise ValueError(
```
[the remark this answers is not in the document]
maxnoe: I feel like we should add some configuration options here on the behavior for the case of

a) table smaller than chunk size (raise vs. just yield what we have, maybe with a `min_chunk_size` option)
b) last chunk (overlapping full size chunk vs. just yielding what you have)

src/ctapipe/monitoring/aggregator.py
```diff
@@ -1,163 +1,402 @@
...
+        start_time = times[0]
+        end_time = times[-1]
+        total_duration = end_time - start_time
+
+        # Validate inputs
+        if total_duration < self.chunk_duration:
+            raise ValueError(
+                f"Total duration ({total_duration}) is less than chunk_duration ({self.chunk_duration})"
+            )
+        if self.chunk_duration == 0 * u.s:
```
maxnoe:
```suggestion
        if self.chunk_duration <= 0 * u.s:
```

mexanick: Is it needed? Does astropy allow negative time? It is AstroQuantity with the physical type of time.

src/ctapipe/monitoring/aggregator.py
maxnoe: Why does it need the event_id column? It just splits by len, right?

src/ctapipe/monitoring/aggregator.py
```diff
@@ -1,163 +1,474 @@
...
+        chunk_starts = start_time + np.arange(n_chunks) * time_step
+
+        # Filter chunk starts that would create main chunks (not extending beyond end)
+        main_chunk_starts = chunk_starts[chunk_starts + self.chunk_duration <= end_time]
+        last_chunk_start = main_chunk_starts[-1] + time_step
+
+        # Generate chunks for each main chunk start time
+        for chunk_start in main_chunk_starts:
+            chunk_end = chunk_start + self.chunk_duration
+            mask = (times >= chunk_start) & (times < chunk_end)
```
maxnoe: This will still recompute the chunks in each iteration looking at all timestamps.

Why does the proposed solution of using the remainder-less division not work here?

```python
from astropy.time import Time
import astropy.units as u
import numpy as np


def build_dummy_table():
    
    t0 = Time.now()
    # only for display
    t0.format = 'isot'
    
    timestamps = t0 + rng.uniform(0, 100, 1000) * u.s
    timestamps = timestamps.sort()
    
    return Table({"timestamp": timestamps, "value": rng.normal(size=len(timestamps))})


def get_chunk_borders(chunk_index):
    """For an array of chunk indices, generate the border indices for slicing
    
    E.g. for [0, 0, 0, 1, 1, 2, 2] (3 chunks), the following will allow slicing
    for each chunk: [0, 3, 5, 7].
    """
    n_chunks = chunk_index[-1] + 1
    chunk_borders = np.zeros(n_chunks + 1, dtype=int)
    chunk_borders[1:-1] = np.nonzero(np.diff(chunk_index))[0]
    chunk_borders[-1] = len(chunk_index)

    return chunk_borders

def generate_chunks(table, chunk_duration):
    chunk_duration = chunk_duration.to_value(u.s)
    relative_time = (table["timestamp"] - table["timestamp"][0]).to_value(u.s)
    
    chunk_index = (relative_time // chunk_duration).astype(int)
    chunk_borders = get_chunk_borders(chunk_index)

    for a, b in zip(chunk_borders[:-1], chunk_borders[1:]):
        yield a, b


table = build_dummy_table()

for a, b in generate_chunks(table, chunk_duration=6.5 * u.s):
    chunk = table[a:b]
    print(f"{a:5d} {b:5d}", chunk["timestamp"][0], chunk["timestamp"][-1], f"{np.mean(chunk["value"]):5.2f}")
```

maxnoe: Ah, right, I forgot about the overlapping chunk business.

src/ctapipe/monitoring/calculator.py
```diff
@@ -165,13 +154,22 @@ def first_pass(
...
         aggregator = self.stats_aggregators[self.stats_aggregator_type.tel[tel_id]]
-        # Pass through the whole provided dl1 table
-        aggregated_stats = aggregator(
-            table=table,
-            masked_pixels_of_sample=masked_pixels_of_sample,
-            col_name=col_name,
-            chunk_shift=None,
-        )
+
+        # Temporarily disable chunk_shift for first pass to ensure non-overlapping chunks
```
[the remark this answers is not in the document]
mexanick: It is a description of the internal logic operation: you have chunk_shift configured, but you do not use it for the first pass, you "re-enable" it for the second pass. I you have a suggestion how to rephrase this comment, I would be happy to implement it.

src/ctapipe/monitoring/tests/test_aggregator.py
```diff
@@ -176,3 +213,715 @@ def test_with_outliers(example_subarray):
...
+    # Should create 2 chunks (10 seconds / 5 seconds per chunk)
+    assert len(result_time_5s) == 2
+
+
+def test_time_based_chunking_with_shift():
+    """test time-based chunking with overlapping windows (rolling statistics)"""
+
+    # Create dummy data spanning exactly 5 seconds to get cleaner chunks
+    start_time = 60117.911
+    duration_days = 5 / 86400  # 5 seconds in days
```
maxnoe: See above comments about more redable time ranges

src/ctapipe/monitoring/aggregator.py
```diff
@@ -1,163 +1,465 @@
...
-    "StatisticsAggregator",
-    "PlainAggregator",
-    "SigmaClippingAggregator",
-]
+from ..core import Component
+from ..core.traits import AstroQuantity, Bool, ComponentName, Enum, Int
 
 
-class StatisticsAggregator(TelescopeComponent):
+class BaseChunking(Component, ABC):
```
maxnoe: I'm not exactly sure about the implications, but usually we use:

```suggestion
class BaseChunking(Component, metaclass=ABCMeta):
```

src/ctapipe/monitoring/aggregator.py
```diff
@@ -1,163 +1,402 @@
...
+        Generate chunks from the input table.
+
+        Parameters
+        ----------
+        table : astropy.table.Table
+            Input table with 'time' and 'event_id' columns
+
+        Yields
+        ------
+        astropy.table.Table
```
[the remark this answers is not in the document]
maxnoe: What's the conclusion here? We return a view?

maxnoe (changes requested):
Two minor comments and a question, otherwise looks good!

src/ctapipe/monitoring/aggregator.py
mexanick: Oldschool :) class `ABC` was introduced in 3.4 for convenience, there's no implication: 
https://github.com/python/cpython/blob/19ba59369ef55de3ebea721b9c9926c53532d840/Lib/abc.py#L184

I will do as in other places, for uniformity.

mexanick:
@maxnoe, @TjarkMiener suggestions implemented, docstring regarding copy vs reference return updated (we return references)

maxnoe (approved):

PR #2852, opened 2025-10-07
Add support for list of components in get_current_config

### LINKED ISSUE #2850: Tool.get_current_config() misses Components that are in lists
**Describe the bug**

e.g. in #2802 , there is a property of the Component/Tool that is a  list of Components. These do not show up in the provenance log since the logic inside `get_current_config()` only looks for Component or Tool *instances*, not *lists of instances* when it recurses.

**Expected behavior**
Config from lists of instances is found,.

maxnoe:
Fixes #2850


maxnoe:
Ah, sorry duplicate work with #2851

PR #2868, opened 2025-10-21
Visualizing schematic representation of Hillas parameters 

vuillaut:
This is an addition to the documentation to display as a schema what are the Hillas parameters:

annotated_hillas_parameters


sfegan:
The way you have drawn the length and width is confusing since it shows them as a "diameter" rather than a "radius". For example, you note that the length is 0.26m but the blue line that you indicate as the length is 0.52m in length. I would suggest you stop the blue and green lines at the center of the ellipse to avoid confusion.

PR #2869, opened 2025-10-21
Phi0 should be subtracted from phi not added.

burmist-git:
Phi0 gives us the azimuth of the impact point, whereas before it was giving us the complementary angle of the impact point.
https://github.com/cta-observatory/ctapipe/pull/2845#discussion_r2444359969
related to :
https://github.com/cta-observatory/ctapipe/pull/2845

### DIFF
--- a/src/ctapipe/image/muon/intensity_fitter.py
+++ b/src/ctapipe/image/muon/intensity_fitter.py
@@ -150,8 +150,8 @@ def create_profile(
     radius: float
         ring radius
     phi: float
-        Phase (phi0) defines the azimuth of the most probable
-        value in the phi distribution.
+        Phase corresponding to the azimuth of maximum
+        intensity of the ring.
     pixel_diameter: float
         pixel FoV
     oversampling: int


maxnoe (dismissed):

src/ctapipe/image/muon/intensity_fitter.py
```diff
@@ -141,12 +141,23 @@ def create_profile(
...
     impact_parameter: float
         Impact distance from mirror center
-    ang: ndarray
-        Angles over which to integrate
+    radius: float
+        ring radius
     phi: float
-        Rotation angle of muon image
+        phase (phi0) defines the azimuth of the most probable
+        value in the phi distribution. The rotation angle or muon image
```
maxnoe: Opposite direction to what?

burmist-git: It is to phi0, that’s why we subtract instead of add.

maxnoe: I'm not sure I understand. As it is now, this explanation is confusing to me and I expect for other users / developers of this function to. As written, it seems to imply I need to ensure some definition or direction. However, it seems that this is just taken care of by the function itself?

Is this really a helpful doc string for the phi parameter? Or more of an explanation for how the `ang` array is created inside of the function?

mexanick: or azimuth of maximum intensity of the ring?

maxnoe (approved):

PR #2873, opened 2025-10-24
Fix periodicity behavior of chord_length

maxnoe:
```
In [1]: from ctapipe.image.muon.intensity_fitter import chord_length
In [2]: import numpy as np
In [3]: phi = np.linspace(0, 2 * np.pi, 500)
In [4]: %timeit chord_length(10.0, 1.2, phi)
```

This branch:
```
5.6 μs ± 12.5 ns per loop (mean ± std. dev. of 7 runs, 100,000 loops each)
```

On branch of #2872:
```
7.6 μs ± 28.2 ns per loop (mean ± std. dev. of 7 runs, 100,000 loops each)
```

```python
from ctapipe.image.muon.intensity_fitter import chord_length
import numpy as np


phi = np.linspace(-4 * np.pi, 4 * np.pi, 1000)

fig, ax = plt.subplots(1, 1, layout="constrained", sharex=True)

for rho in (0.5, 1.2):
    ax.plot(phi, chord_length(10, rho, phi), label=f"{rho = }")

ax.legend()
fig.savefig("chord_length.png")
plt.show()
```

Before:
chord_length

Now:
chord_length

### DIFF
--- a/docs/changes/2873.bugfix.rst
+++ b/docs/changes/2873.bugfix.rst
@@ -1,0 +1,2 @@
+Fix a bug in the ``chord_length`` function used by the  ``MuonIntensityFitter`` likelihood function
+that resulted in incorrect periodicity. The function returned 0 outside of the domain phi [-π, π].
--- a/src/ctapipe/image/muon/intensity_fitter.py
+++ b/src/ctapipe/image/muon/intensity_fitter.py
@@ -68,7 +68,10 @@ def chord_length(radius, rho, phi):
 
 
     """
-    discriminant_norm = 1 - (rho**2 * np.sin(phi) ** 2)
+    sin_phi = np.sin(phi)
+    cos_phi = np.cos(phi)
+
+    discriminant_norm = 1 - (rho**2 * sin_phi**2)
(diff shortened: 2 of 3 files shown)


burmist-git:
So the modulo involves really expensive calculations… But are you sure that this replacement of the modulo works for all cases?

maxnoe:
```python

from ctapipe.image.muon.intensity_fitter import chord_length
import numpy as np


phi = np.linspace(-4 * np.pi, 4 * np.pi, 1000)
rho = 1.2

phi_mod = (phi + np.pi) % (2 * np.pi) - np.pi
mod_cond = np.abs(phi_mod) < np.arcsin(1 / rho)
new_cond = (np.abs(np.sin(phi)) < 1 / rho) & (np.cos(phi) > 0)

fig, ax = plt.subplots(1, 1, layout="constrained", sharex=True)

ax.plot(phi, mod_cond, label='mod condition')
ax.plot(phi, new_cond, ls=':', label='cos condition')
ax.legend()

fig.savefig("chord_conditions.png")
```
chord_length

PR #2889, opened 2025-11-12
Relax a bit time column requirements for chunking

mexanick:
This is needed to allow re-aggregation of already aggregated values, i.e. if we want to produce an OBS_ID-average pedestal or FF image (this is the actual use case leading to this change).

### DIFF
--- a/docs/changes/2889.feature.rst
+++ b/docs/changes/2889.feature.rst
@@ -1,0 +1,1 @@
+Allow subsequent aggregation of pre-aggregated time ranges (time_start, time_end) instead of just time column.


mexanick:
@maxnoe this is a minor change that's needed for the ACADA-LST1 test, and it would be great if can be added to 0.28, but if not - not a problem.

TjarkMiener (approved):
Fine for me! Just keep in mind that OBS_ID-average can be tricky in case we have a second pass in the `PixelStatisticsCalculator`.

maxnoe (approved):

PR #2903, opened 2025-12-10
Trusted publishing

maxnoe:
I configured pypi and this repository for trusted publishing, which avoids needing a personal access token.


maxnoe:
Numba dropped support for intel macs. @kosack Should we drop the CI for it or should I add a rule to install older numba on macos intel?

PR #2913, opened 2025-12-17
fix #2893, wrong n_showers when no sim dist in file

### LINKED ISSUE #2893: Number of showers fallback does not take reuse into account in dl2_tables_preprocessing
When using the `DL2EventLoader` e.g. with the `IRFTool` to load the simulation information, it first tries to read that from the showers_distribution Table. If this isn't available it estimates the number of showers from the simulation configuration. This estimate doesn't take into account the `shower_reuse`.

https://github.com/cta-observatory/ctapipe/blob/0882370af1da320134f9cd5e4f301ded1b78f976/src/ctapipe/io/dl2_tables_preprocessing.py#L352-L359

**Possible solution:**
substitute line 359 with something like:
```python
show = Table([sim["n_showers"] * sim["shower_reuse"]], names=["n_entries"], dtype=[np.int64])
```

Underestimating the number of showers could lead e.g. to an overestimation of the effective area.

kosack:
If no simulation distribution was present in the file, the fallback was to use the simulation config, but applying the shower reuse was missing. (see #2893)

Also adds some more consistency checks to ensure we don't make a mistake in the effective area.

### DIFF
--- a/src/ctapipe/io/dl2_tables_preprocessing.py
+++ b/src/ctapipe/io/dl2_tables_preprocessing.py
@@ -381,7 +381,7 @@ class DL2EventLoader(Component):
                     f"Unsupported: '{itm}' differs across simulation runs"
                 )
 
-        n_showers_config = sim_config["n_showers"].sum() * sim_config["shower_reuse"][0]
+        n_showers_config = (sim_config["n_showers"] * sim_config["shower_reuse"]).sum()
         if n_showers == 0:
             n_showers = n_showers_config
 


src/ctapipe/io/dl2_tables_preprocessing.py
```diff
@@ -349,29 +351,52 @@ def get_simulation_information(
...
+            "max_scatter_range",
+            "max_viewcone_radius",
+            "min_viewcone_radius",
+        ]:
+            if len(np.unique(sim_config[itm])) > 1:
                 raise NotImplementedError(
                     f"Unsupported: '{itm}' differs across simulation runs"
                 )
 
+        n_showers_config = sim_config["n_showers"].sum() * sim_config["shower_reuse"][0]
```
maxnoe:
```suggestion
        n_showers_config = (sim_config["n_showers"] * sim_config["shower_reuse"]).sum()
```

otherwise it will be wrong for individual re-use. Or check reuse above, but here it's easy to do the right thing.

PR #2916, opened 2026-01-13
Merge telescope-wise data from the same OB

### LINKED ISSUE #1625: Horizontal merging of DL1 files
To process observed data, e.g. MAGIC + LST, using stage1, we need a tool to merge DL1 files horizontally (merging subarrays instead of appending events).

TjarkMiener:
This PR adds a new merge strategy for combining telescope-wise data from the same OB. It also added a static method to the `SubarrayDescription` to merge subarrays into one. 

---
closes #1625

### DIFF
--- a/src/ctapipe/io/hdf5merger.py
+++ b/src/ctapipe/io/hdf5merger.py
@@ -223,11 +223,11 @@ class HDF5Merger(Component):
             )
         self.combine_telescope_data = self.merge_strategy == "combine-telescope-data"
         if self.combine_telescope_data and (
-            not self.telescope_events or self.dl2_subarray
+            not self.telescope_events or self.dl2_telescope or self.dl2_subarray
         ):
             raise traits.TraitError(
                 "Merge strategy 'combine-telescope-data' requires telescope_events=True "
-                "and dl2_subarray=False"
+                "and dl2_telescope=False and dl2_subarray=False"
             )
 
         output_exists = self.output_path.exists()
--- a/src/ctapipe/io/hdf5dataformat.py
+++ b/src/ctapipe/io/hdf5dataformat.py
@@ -43,6 +43,7 @@ __all__ = [
     "DL2_TEL_GEOMETRY_GROUP",
     "DL2_TEL_ENERGY_GROUP",
     "DL2_TEL_PARTICLETYPE_GROUP",
+    "DL2_TEL_IMPACT_GROUP",
     "DL0_TEL_POINTING_GROUP",
     "DL1_SUBARRAY_POINTING_GROUP",
     "DL1_TEL_POINTING_GROUP",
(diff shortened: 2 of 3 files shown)


src/ctapipe/io/hdf5merger.py
```diff
@@ -576,22 +693,40 @@ def _append_subarray(self, other):
...
+                self.subarray.to_hdf(self.h5file)
+
+        if not self.combine_telescope_data:
+            # Relax subarray matching requirements for attaching
+            # monitoring data of the same observation block.
+            if not self.single_ob or not self.attach_monitoring:
+                if self.subarray != subarray:
+                    raise CannotMerge(
+                        f"Subarrays do not match for file: {other.filename}"
+                    )
```
TjarkMiener: @Voutsi the default merge strategy will check if the subarrays match between different observation blocks. For your use case you would need to disable that check? Should we add another merge strategy for this then?

TjarkMiener: > Thinking about it, you just cannot simply combine subarrays, there are multiple arrays in the data that are arrays of length `n_telescopes`. You need to change these values to be consistent over all events, e.g.the `tels_with_trigger` in the trigger subarray trigger table.

Yes, we need to run the `_flush()` also for this usecase. I added a new merge strategy 'events-multiple-obs-different-subarrays' for it. I also exclude DL2 subarray from the merging since it is not really needed and it will contain arrays of length `n_telescopes`. It is also not a simple merge since it will involve some logic of how to combine those predictions.

TjarkMiener:
@maxnoe Sorry for the long silence, I misinterpreted your comment about the multiple files slightly. First, I thought with multiple files you mean per `tel_id` and then the `cross-calibration` tool should be able to digest those telescope-wise files (but then the crucial stereo impact reco is missing). Now, I think you mean that the cross calibration should just be able to digest per `obs_id` stereo DL2 files and keep track of the `SubarrayDescription` per `obs_id` since they can differ. Therefore indeed, we do not need the merge strategy `events-multiple-obs-different-subarrays`, so the last two commits should be removed. The overall pseudo-workflow would look like this:

1. `DL0` per `tel_id` will be processed via `ctapipe-process` to `DL1b` for each OB.
2. `DL1b` per `tel_id` will be merged in each OB via `ctapipe-merge` with `combine-telescope-events`.
3. Merged `DL1b` files for each OB will be processed via `ctapipe-process` enabling `--write-showers` (or this recompute flag of DL2).
4. `ctapipe-apply-models` tool to get the DL2 telescope and subarray predictions from ML reco. This would lead to a DL2 file for each `obs_id` following the same structure as @Voutsi's test simulation data.
5. `cross-calibration` tool reads multiple files per `obs_id` with (maybe) different subarrays and do the merging in memory  

Besides excluding DL2 subarray from merging with strategy `combine-telescope-events` as done in [2a3f0b8](https://github.com/cta-observatory/ctapipe/pull/2916/commits/2a3f0b8dad56dab8b5e62e86a56f0c77a1b44d40), we also need to exclude DL2 telescope infos such as the impact. Then, `combine-telescope-events` only produced files that are not violating the assumptions of the data model.

Can you please confirm so we can go ahead with the implementation, @maxnoe @kosack?

PR #2920, opened 2026-01-20
Minimal initial commit of FreePACT code

ParsonsRD:
This is a commit of a simple implementation of the FreePACT code. Updates to the ImPACT code are included from #2705


ParsonsRD:
Ugh sorry yes that should have been obvious. I blame on being very tired today :-).

src/ctapipe/reco/freepact.py
```diff
@@ -0,0 +1,214 @@
...
+    Parameters
+    ----------
+    subarray : ctapipe.instrument.SubarrayDescription
+        The telescope subarray to use for reconstruction
+    atmosphere_profile : ctapipe.atmosphere.AtmosphereDensityProfile
+        Density vs. altitude profile of the local atmosphere
+
+    References
+    ----------
+    .. [schwefer24] Schwefer, Parsons, & Hinton,  Astroparticle Physics 163 (2024), 103008
```
maxnoe: We use the bibtex extension for sphinx, please add the reference to the `docs/references.bib` file and use: 

```
:cite:p:`<bibtex key>`
```

maxnoe:
Please remove the `impact` extra from the CI job for `minimal-dependencies`, the task of this job is to test ctapipe without any of the optional dependencies installed.

ParsonsRD:
OK great, thanks for the help in working through the CI issues. I will work through the sonarqube warnings and fix where appropriate. I would also like to have @gschwefer check through this and make sure it is fully compatible with the training files he has created.

The other major missing work is to distribute a training code, which is currently a set of notebooks. This can come in a separate PR I think. Or we save it for a larger commit where we spin impact and freepact off into a plugin, which I think should probably be the ultimate goal (assuming Georg feels the same way).

ParsonsRD:
After speaking with @kosack and @maxnoe I created an ImPACT and FreePACT plugin here:
https://github.com/ParsonsRD/FreePACT-Reconstructor-plugin

If this seems like a better approach, we could close this request.

maxnoe:
@ParsonsRD looks mostly good. Didn't have a chance yet to try it out though.

Some observations:

- Claiming compatibility with `ctapipe >= 0.17` seems a bit... optimistic, did you test with versions that old?
- In the README, you tell people to instantiate via "Reconstructor.from_name`, however, I think this is not really meant for user code but for use inside of ctapipe to discover the plugin classes. Users should probably just just `ctapipe-process` configuring the `FreePACTReconstructor`, or use the `ShowerProcessor` or instatiate the class directly from the plugin.

PR #2923, opened 2026-01-21
Fix sphinx warnings

kosack:
The documentation build is broken on all PRs right now, since sphinx 9.x was released and it causes some problems.

Until we fix them, I have pinned it to 8.x here, and also made a few minor cleanups in the docs.  I also added a `html-noplot` target in the docs makefile, which skips the building of the example notebooks (useful when debugging the docs, as they take  long time to run)

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -87,10 +87,10 @@ docs = [
     "pandas",
     "pydata_sphinx_theme",
     "pypandoc",
-    "sphinx ~= 8.2.0",
+    "sphinx~=8.2.0",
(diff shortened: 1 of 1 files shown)


kosack:
Pinning to 8.2 here doesn't seem to have worked. I'm not sure why. On my laptop, 8.2 worked fine, but 9.0 gave these errors.  So perhaps it's not sphinx, but one of the plugins that needs to be pinned? Maybe it's automodapi? It's very hard to see where the errors are actually coming from, as it gives no details

PR #2930, opened 2026-01-28
set is_valid to true in VarianceExtractor

TjarkMiener:
only set false if all values are NaNs

currently all is_valid flags are set to the default container field which is false.


src/ctapipe/image/extractor.py
```diff
@@ -1329,8 +1329,10 @@ class VarianceExtractor(ImageExtractor):
     def __call__(
         self, waveforms, tel_id, selected_gain_channel, broken_pixels
     ) -> DL1CameraContainer:
+        image = np.nanvar(waveforms, dtype="float32", axis=2)
         container = DL1CameraContainer(
-            image=np.nanvar(waveforms, dtype="float32", axis=2),
+            image=image,
+            is_valid=not np.all(np.isnan(image)),
```
maxnoe:
```suggestion
            is_valid=np.any(np.isfinite(image)),
```

PR #2935, opened 2026-01-30
Retry vizier queries

maxnoe:
We observed since some time that CI pipelines fail from time to time due to vizier being unavailable.

I'm not sure if it's extended in time or not, let's try with a backoff of 10 s for now in case it is really a short-term transient issue.


maxnoe:
The issue is not vizier being unavailable, but actually a rate limit. So retrying would be even worse.

PR #2942, opened 2026-02-06
Implement service data loading

mexanick:
Add `from_service_data()` classmethod to construct SubarrayDescription from service data files using subarray id.
Implements #2946


src/ctapipe/instrument/subarray.py
```diff
@@ -808,3 +825,342 @@ def check_matching_subarrays(subarray_list: list) -> bool:
...
+        """
+        subarray_data = get_structured_dataset(
+            "subarray-ids", role="dl0.sub.svc.subarray"
+        )
+
+        if subarray_id is not None:
+            for subarray in subarray_data["subarrays"]:
+                if subarray["id"] == subarray_id:
+                    return subarray
+            raise ValueError(f"Subarray ID {subarray_id} not found")
```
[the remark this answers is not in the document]
maxnoe: or a `class UnknownSubarray(KeyError)`

src/ctapipe/instrument/subarray.py
```diff
@@ -808,3 +826,342 @@ def check_matching_subarrays(subarray_list: list) -> bool:
...
+        dict[int, str]
+            Dictionary mapping array element IDs to camera names
+        """
+        camera_names = {}
+        for ae_id in tel_positions.keys():
+            tel_name = ae_id_to_name.get(ae_id, "")
+            if "LST" in tel_name:
+                camera_names[ae_id] = "LSTCam"
+            elif "MST" in tel_name:
+                if site.lower() == "ctao_n":
```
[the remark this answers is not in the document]
mexanick: This is used only when the camera type needs to be inferred instead of being properly specified. Do you know if this is something to be updated in the data model?

kosack (changes requested):

src/ctapipe/instrument/subarray.py
```diff
@@ -808,3 +826,342 @@ def check_matching_subarrays(subarray_list: list) -> bool:
...
+        dict
+            Dictionary with 'metadata' and 'array_elements' keys.
+            'array_elements' is a list of dicts with 'id' and 'name' keys.
+        """
+        return get_structured_dataset(
+            "array-element-ids", role="dl0.sub.svc.array_elements"
+        )
+
+    @classmethod
+    def load_array_element_positions(cls, site="CTAO-North"):
```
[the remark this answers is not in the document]
maxnoe: the alternative would be to use the official table and add `pyproj` as an optional dependency to perform these conversions on the fly.

src/ctapipe/instrument/subarray.py
```diff
@@ -824,3 +826,224 @@ def check_matching_subarrays(subarray_list: list) -> bool:
...
+            ├── positions/
+            │   ├── CTAO-North_ArrayElementPositions.ecsv
+            │   └── CTAO-South_ArrayElementPositions.ecsv
+            └── subarray-ids.json
+
+        Files:
+
+        - array-element-ids.json: mapping of telescope IDs to names
+        - subarray-ids.json: subarray definitions
+        - positions/\*.ecsv: ECSV files with telescope positions for each site
```
maxnoe:
```suggestion
        - positions/<site name>_ArrayElementPositions.ecsv: ECSV files with telescope positions for each site
```
to be more explicit?

src/ctapipe/instrument/subarray.py
```diff
@@ -824,3 +826,224 @@ def check_matching_subarrays(subarray_list: list) -> bool:
...
+        tel_descriptions = {}
+        for ae_id in tel_positions.keys():
+            tel_name = ae_id_to_name.get(ae_id, f"Unknown-{ae_id}")
+            try:
+                tel_type = cls._resolve_telescope_type(ae_id)
+                tel_descriptions[ae_id] = cls._load_telescope_description(
+                    ae_id, tel_name, tel_type
+                )
+            except (FileNotFoundError, ValueError, KeyError) as e:
+                warnings.warn(
```
maxnoe: How is not finding a telescope description for a given subarray only a warning?

This needs to be an error I think.

src/ctapipe/instrument/conftest.py
```diff
@@ -34,11 +37,229 @@ def instrument_dir(tmp_path_factory):
...
+    lst_readout_path = get_dataset_path("LSTcam.camreadout.fits.gz")
+    shutil.copy(lst_readout_path, lst_dir / "LSTN.camreadout.fits.gz")
+
+    # LSTN optics
+    lst_optics = QTable(
+        {
+            "optics_name": ["LSTN"],
+            "size_type": [SizeType.LST.value],
+            "reflector_shape": [ReflectorShape.PARABOLIC.value],
+            "n_mirrors": [1],
```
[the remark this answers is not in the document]
maxnoe: But you don't need to modify the optics description code, right?
There is no `OpticsDescription.from_table`, that's on the subarray. So I'd either add that, or just adh-hoch instantiate an `OpticsDescription` from the info found in the metadata.

Or, if we do not change anything for now, I'd keep the format of the optics table and put one optics table at the root for all telescopes.

maxnoe:
Final comment on general aspects: ctapipe-dump-instrument should support this format as output and the tests (or at least a test) could check that we can load the output of it for a simulation file and it round-trips.

mexanick:
> Final comment on general aspects: ctapipe-dump-instrument should support this format as output and the tests (or at least a test) could check that we can load the output of it for a simulation file and it round-trips.

done!

mexanick:
@maxnoe I added reference metadata to subarray-level files and TAB_VER to optics. I would not add more to optics right now as we plan to extend it, so not sure whether it would be good now. Considering the telescope names being identical: the code uses `TelescopeDescription.name` attribute. I believe it should be fixed at source (i.e. in the simtel file) rather with some workarounds.

kosack (dismissed):

src/ctapipe/instrument/subarray.py
```diff
@@ -69,6 +71,10 @@ class SubarrayDescription:
     CURRENT_TAB_VERSION = "2.0"
     #: Version numbers supported by `SubarrayDescription.from_hdf`
     COMPATIBLE_VERSIONS = {"2.0"}
+    #: Current version of the service data format expected by `SubarrayDescription.from_service_data`
+    CURRENT_SERVICE_DATA_VERSION = "2.0"
```
maxnoe: Why do we start at 2.0?

src/ctapipe/instrument/subarray.py
```diff
@@ -824,3 +836,254 @@ def check_matching_subarrays(subarray_list: list) -> bool:
...
+
+        Returns
+        -------
+        str
+            Telescope type name (e.g., 'LSTN', 'MSTN')
+        """
+        import os
+        from pathlib import Path
+
+        searchpath = os.getenv("CTAPIPE_SVC_PATH")
```
maxnoe, line 865: This assumes that `CTAPIPE_SVC_PATH` directly has the instrument data structure. Do we really want this?

Or do we want something like:

```
for search_dir in searchpath.split(os.pathsep):
    candidate = Path(search_dir) / "instrument/array-elements" / ae_id_str
```

i.e. add one level "instrument" below the base service path.

src/ctapipe/instrument/subarray.py
```diff
@@ -824,3 +836,254 @@ def check_matching_subarrays(subarray_list: list) -> bool:
...
+    @classmethod
+    def from_service_data(cls, subarray_id):
+        """
+        Create a SubarrayDescription from CTAO service data files.
+
+        This method loads subarray definitions from the service data.
+
+        Expected directory structure::
+
+            instrument/
```
maxnoe, line 1003: Here you have the `instrument` level. So I would assume to have this as written above

src/ctapipe/instrument/subarray.py
kosack: Yes, that would be nicer - I didn't actually try loading from the service path.  Otherwise we would have to change the path for each instrument.

src/ctapipe/instrument/subarray.py
mexanick: Inherited from this:
>    CURRENT_TAB_VERSION = "2.0"

should I decouple?

maxnoe: That's different versions, right? One is the table format of the layout table, one is the directory structure.

These could change independently.

mexanick: Yes, they are different and could change independently, I just didn't feel comfortable introducing one more different version in our zoo of versions... I will decouple and reset it.

maxnoe:
Looks mostly good (already wanted to approve). Only there is a missing import now in the conftest

mexanick:
@maxnoe @kosack I've addressed the comments and rebased on the current main, should be ready for a final review pass.

maxnoe (approved):

mexanick:
@kosack could you check/re-approve? There're only minor tests-related changes since your last review

kosack (approved):
This looks good. The only worry I would have is that you have to add all sub-directories to the `CTAPIPE_SVC_PATH`, which while flexible (it doesn't actually rely on the directory structure), could perhaps lead to user mistakes.  But for now, without changing how we search for datasets, it's not easy to implement using a single base-path with structured sub-dirs, so we can maybe leave that to a future re-structuring of dataset reading.

PR #2945, opened 2026-02-10
Polar PSFModel

TjarkMiener:


src/ctapipe/instrument/optics.py
```diff
@@ -376,15 +375,16 @@ class ComaPSFModel(PSFModel):
...
         help=(
             "Describes the dependency of the polar scale on the "
-            "distance to the center of the camera. Used to calculate "
+            "distance to the optical axis. Used to calculate "
             r"the width :math:`S_\phi` of the polar Laplacian in the PSF "
             "as a function of the distance r to the optical axis"
         )
     ).tag(config=True)
 
-    pixel_width = Float(
```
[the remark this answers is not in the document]
maxnoe: PSFModel is a `TelescopeComponent`, so it always has access to the subarray info.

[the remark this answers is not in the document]
mexanick: also looks like it is only used in `pdf` production, that should probably just take a camera geometry as a parameter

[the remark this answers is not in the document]
maxnoe: (in the abstract interface)

src/ctapipe/instrument/optics.py
```diff
@@ -376,17 +383,12 @@ class ComaPSFModel(PSFModel):
     phi_scale_params = List(
         help=(
             "Describes the dependency of the polar scale on the "
-            "distance to the center of the camera. Used to calculate "
+            "distance to the optical axis. Used to calculate "
             r"the width :math:`S_\phi` of the polar Laplacian in the PSF "
             "as a function of the distance r to the optical axis"
         )
     ).tag(config=True)
```
TjarkMiener: Should we make those guys telescope-specific? Currently we need to write configs for each tel type.

mexanick: yes, all these parameters are telescope-specific and should be set as `TelescopeParameter`s

src/ctapipe/instrument/optics.py
```diff
@@ -355,89 +361,113 @@ class ComaPSFModel(PSFModel):
...
     """
 
-    asymmetry_params = List(
-        help=(
-            "Describes the dependency of the PSF on the distance "
-            "to the center of the camera. Used to calculate a PDF "
-            "asymmetry parameter :math:`K` of the asymmetric radial Laplacian "
-            "of the PSF as a function of the distance r to the optical axis"
-        )
+    asymmetry_max = FloatTelescopeParameter(
```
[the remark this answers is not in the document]
mexanick: I'd say too many details.. ) but if you find it better like that, I'm fine with it too.

mexanick (dismissed):

kosack (changes requested):
I think it would be good to give defaults for all the parameters, since running this now without configuring each one  causes an error that is not obvious (it's caused by no lookup being created since no value is set).  

```python
psf = PSFModel.from_name("ComaPSFModel", subarray=sub)
psf.subarray.info()
```
```
Subarray : LaPalma-prod6_1-11,14,19
Num Tels : 13
Footprint: 0.39 km2
Height   : 2156.00 m
Lon/Lat  : -17.892030200000022 deg, 28.762166100000005 deg 

       Type       Count  Tel IDs  
----------------- ----- ----------
   LST_LST_LSTcam     4 1-4       
MST_MST_NectarCam     9 5-11,14,19
```

```python
psf.pdf(1, 0.0*u.deg, 0.0*u.deg, 0.0*u.deg, 0.0*u.deg)
```
```
KeyError: "TelescopeParameterLookup: no parameter value was set for telescope with tel_id=1. Please set it explicitly, or by telescope type or '*'."
```

Another option is to not allow None as a config value, so raise an exception if the parameters are not configured. But due to how the config system works, just adding defaults is easier.  Right now the defaults are all `traitlets.Undefined`

kosack:
> @kosack we only know defaults for LST, but not for other telescopes. With the transition to lan/lot the parameterization should become more universal, so perhaps we can just put these defaults, but I'm hesitating a bit providing default values that could be meaningless.

I don't mean physical defaults for all use cases, but nominal numbers that make a generic PSF that works as expected.  That's how all other components work: we don't expect the defaults are always what you want for science, but the class should "work" without specifying all parameters.  E.g for this class, maybe the defaults that mimic a simple gaussian PSF with no coma distortion would be good defaults (so most set to 0.0, or even **all** set to 0.0).  The alternate is to forbid default values and raise a better error message if one is not configured, but to get people to use the class, having defaults is nice.   

Maybe a better way to say it: in all ctapipe components used now, config parameters are always optional. Parameters that are not optional are explicit in `__init__`.

src/ctapipe/instrument/optics.py
```diff
@@ -355,89 +352,161 @@ class ComaPSFModel(PSFModel):
...
+        return (
+            1
+            - self.asymmetry_max.tel[tel_id]
+            * np.tanh(self.asymmetry_decay_rate.tel[tel_id] * r)
+            - self.asymmetry_linear_term.tel[tel_id] * r
+        )
+
+    def _s_r(self, tel_id, r):
+        return (
+            self.radial_scale_center.tel[tel_id]
```
maxnoe: Why do we now manually compute a polynomial, when before we were using `polyval`, which should be numerically much better behaved due to use of Horner's scheme.

TjarkMiener:
Thanks @maxnoe! I implemented the code suggestions. I'd need to have a closer look on what is causing the  discrepancies!

maxnoe:
I had a mistake when scaling the radial parameters for the plate scale, with the correct scaling:

```
psf_model = ComaPSFModel(
    subarray=lst1,
    asymmetry_max=0.5,
    asymmetry_decay_rate=10 / lst_plate_scale_rad,
    asymmetry_linear_term=0.15 / lst_plate_scale_rad,
    radial_scale_offset=0.015 * lst_plate_scale_rad,
    radial_scale_linear=-0.1,
    radial_scale_quadratic=0.06 / lst_plate_scale_rad**2,
    radial_scale_cubic=0.03 / lst_plate_scale_rad**3,
    polar_scale_amplitude=0.25,
    polar_scale_decay=7.5 / lst_plate_scale_rad,
    polar_scale_offset=0.02 * lst_plate_scale_rad,
)
```

The center PSF now looks the same as on main:
psf_polar

maxnoe:
Actually no, we are not using any trig functions on the values.

kosack:
> Dividing by r works, but needs special treatment at r=0, which I didn't figure out yet.

It think computing the integral analytically of the function over $\int_0^\infty r dr \int_0^{2\pi} d\phi \ PSF(r,\phi)$ and dividing by it is the only way to do it properly.

maxnoe:
Yes. That computation is what I didn't invest time into

kosack:
I guess Tjark will do it, but it's pretty easy with SymPy. Here I did it with the definition from the paper, but as I said the implementation in scipy is normalized differently, so this is wrong if you use the `laplace` function from there. Would have to look up their exact definition of laplace and laplace_asymmetric.

```py
r, phi, s_phi = smp.symbols("r phi s_phi", real=True)
f_phi = (1.0 / 2 * s_phi) * smp.exp(-smp.Abs(phi / s_phi))
f_phi_norm = smp.integrate(f_phi, (phi,0,2*smp.pi))
f_phi_norm
```
$0.5 s_{\\phi} \\left|{s_{\\phi}}\\right| - 0.5 s_{\\phi} e^{- \\frac{2 \\pi}{\\left|{s_{\\phi}}\\right|}} \\left|{s_{\\phi}}\\right|$

The `f_r` part is a bit more complex, as it's piecewise, but also should be easy to do (and r and phi are separable here)

mexanick:
`r==0` is not a problem, we have a lower limit for `r`, the whole model is not valid at low `r`, and there's a condition on that.

maxnoe:
> the whole model is not valid at low r, and there's a condition on that.

Also the other model that takes action on low r is not properly normalized, the issue is the same. And to be properly normalized, you do need to _integrate_ from 0.

kosack:
>> the whole model is not valid at low r, and there's a condition on that.

> Also the other model that takes action on low r is not properly normalized, the issue is the same. And to be properly normalized, you do need to integrate from 0.

I missed that, yes indeed that makes normalization difficult. The Laplace distribution is not infinite at r=0 though. They mention orthorgonality being a problem in the code, but of course r and phi are not orthogonal in general, so I wonder if that's just due to not taking into account the jacobian?

src/ctapipe/instrument/tests/test_psf_model.py
```diff
@@ -6,57 +6,129 @@
 import numpy as np
 import pytest
 
-from ctapipe.core.traits import TraitError
 from ctapipe.instrument.optics import PSFModel
 
+# Handle numpy version compatibility for trapezoid/trapz
+if tuple(int(x) for x in np.__version__.split(".")[:2]) >= (2, 0):
+    trapz_func = np.trapezoid
```
maxnoe: Please add this here: https://github.com/cta-observatory/ctapipe/blob/main/src/ctapipe/compat.py

so we can remove in one go once dropping support for numpy < 2.0

src/ctapipe/instrument/optics.py
```diff
@@ -350,94 +348,191 @@ class ComaPSFModel(PSFModel):
...
-        # Phi is not defined at the center
-        at_center = np.isclose(r0, 0, atol=self.pixel_width)
-        polar_pdf = np.where(at_center, 1 / (2 * s_phi), polar_pdf)
         # Polar PDF is valid under approximation that the polar axis is orthogonal to the radial axis
         # Thus, we limit the PDF to a chord of 6 pixels or covering ~30deg around the radial axis, whichever is smaller
-        chord_length = min(6 * self.pixel_width, 0.5 * r0)
-        if r0 != 0:
+        chord_length = min(6 * self.pixel_width[tel_id], 0.5 * r0)
+
+        if not np.isclose(r0, 0, atol=self.pixel_width[tel_id]):
```
maxnoe: This seems a very complicated way to write:


```suggestion
        if r0 > self.pixel_width[tel_id]:
```

src/ctapipe/instrument/optics.py
```diff
@@ -350,94 +348,191 @@ class ComaPSFModel(PSFModel):
...
-        return proposal["value"]
-
-    @validate("phi_scale_params")
-    def _check_phi_scale_params(self, proposal):
-        if len(proposal["value"]) != 3:
-            raise TraitError("phi_scale_params needs to have length 3")
-        return proposal["value"]
+            polar_pdf = np.where(np.abs(delta_phi) <= dphi, polar_pdf, 0.0)
+
+        pixel = self.pixel_width[tel_id]
```
maxnoe: There is a factor of 0.5 missing here I think since the `pixel_width` is the diameter (circle) / in-circle diameter (hexagon) / side length (square).

I plotted the LST pixel size into the psf picture here.

Currently:

Image

With factor 0.5:

Image

I suggest to change this line to  (and update the ones below for the new variable name)
```suggestion
        pixel_radius = 0.5 self.pixel_width[tel_id]
```

src/ctapipe/instrument/optics.py
```diff
@@ -517,18 +517,20 @@ def pdf(self, tel_id, lon, lat, lon0, lat0):
         # Thus, we limit the PDF to a chord of 6 pixels or covering ~30deg around the radial axis, whichever is smaller
         chord_length = min(6 * self.pixel_width[tel_id], 0.5 * r0)
 
-        if not np.isclose(r0, 0, atol=self.pixel_width[tel_id]):
+        if (
+            r0 > self.pixel_width[tel_id]
+        ):  # only apply the chord limit for sources outside the central pixel
```
maxnoe, line 524: this should also check against pixel_radius, right? To be consistent?

mexanick: in general, this is different approximation condition, but for consistency I made it pixel_radius limit here as well.

maxnoe (approved):

PR #2963, opened 2026-03-05
Fix TypeError in HDF5EventSource

yaochengchen:
Replace dict_values indexing with next(iter(...)) in _is_hillas_in_camera_frame to avoid a Python 3 TypeError.

### DIFF
--- a/docs/changes/2963.bugfix.rst
+++ b/docs/changes/2963.bugfix.rst
@@ -1,0 +1,1 @@
+Fix crash in HDF5EventSource caused by indexing dict_values in _is_hillas_in_camera_frame.


yaochengchen:
> @yaochengchen Could you please share an example where this is an issue in a real world setting actually using the `HDF5EventSource`?
> 
> I am very certain that if there was an issue, it would be triggered by existing tests in this repository, as the HDF5EventSource is used extensively.

Hi @maxnoe,

You’re right that I haven’t encountered this through a real pipeline run myself. I noticed it while reading the implementation: in Python 3, `dict.values()` returns a `dict_values` view, which is iterable but not subscriptable. The current code does `values()[0]`, which raises a `TypeError` if that path is executed.

This seems to be a leftover from Python 2 behavior where `dict.values()` returned a list.

PR #2964, opened 2026-03-05
Implement CTAO CR56: Add field pixel_time_shift to R1 and DL0

moralejo:
New field which contains event-and-pixel-wise time corrections in ns (to be subtracted from reconstructed pixel times)

### DIFF
--- a/src/ctapipe/io/tests/test_hdf5eventsource.py
+++ b/src/ctapipe/io/tests/test_hdf5eventsource.py
@@ -31,7 +31,7 @@ def test_is_compatible(compatible_file, request):
 def test_metadata(dl1_file):
     with HDF5EventSource(input_url=dl1_file) as source:
         assert source.is_simulation
-        assert source.datamodel_version == (7, 4, 0)
+        assert source.datamodel_version == (7, 5, 0)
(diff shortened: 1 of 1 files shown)


src/ctapipe/containers.py
```diff
@@ -649,6 +649,20 @@ class R1CameraContainer(Container):
...
 
+    pixel_time_shift = Field(
+        None,
+        (
+            "Numpy array containing a time shift value in nanoseconds"
+            "for each pixel and each gain, originating from the DRS4 sampling"
+            "time interval inhomogeneity. This value must be subtracted from"
+            "the reconstructed time to achieve uniform time response for all"
+            "pixels"
+            "Shape: (num_channels, num_pixels)"
```
kosack: Missing spaces in concatinated strings:
```suggestion
            "Numpy array containing a time shift value in nanoseconds"
            " for each pixel and each gain, originating from the DRS4 sampling"
            " time interval inhomogeneity. This value must be subtracted from"
            " the reconstructed time to achieve uniform time response for all"
            " pixels."
            " Shape: (num_channels, num_pixels)"
```

PR #2969, opened 2026-03-24
add polygon_chord

burmist-git:
Polygonal description of the chord.

[096_arbitrary_shaped_mirror_and_polygon_chord.pdf](https://github.com/user-attachments/files/26209538/096_arbitrary_shaped_mirror_and_polygon_chord.pdf)

Analytical solution :

Input:
    Ray origin, ray direction, polygon edges

1. Compute the ray direction vector.
2. Compute the intersection parameters (t, s) for all polygon edges.
3. Keep only intersections satisfying:
       0 ≤ t < 1 and s ≥ 0.
4. Compute the intersection coordinates.

5. If no intersections:
       return 0

6. If one intersection:
       return distance from the ray origin to the intersection

7. If two intersections:
       return distance between the two intersections

8. Otherwise:
       Sort the distances from the ray origin to all intersections.
       Return the alternating sum of the sorted distances.

image

Approximation with Fourier transformation:

gr_reco_lst
gr_reco_box
gr_reco_hexagon


src/ctapipe/image/muon/intensity_fitter.py
```diff
@@ -40,12 +40,67 @@
...
+        X-coordinate of the ray origin, which is the muon's impact point (x).
+    mu_y : float
+        Y-coordinate of the ray origin, which is the muon's impact point (y).
+    phi : float
+        Angle defining the direction of the ray.
+    ri_x : ndarray of shape (N,)
+        X-coordinates of the starting points of the polygon edges.
+    ri_y : ndarray of shape (N,)
+        Y-coordinates of the starting points of the polygon edges.
+    vi_x : ndarray of shape (N,)
```
mexanick, line 63: I think we should deduce these (if needed) from the vertices. You can write a helper for this.

src/ctapipe/image/muon/intensity_fitter.py
```diff
@@ -39,6 +39,98 @@
...
+    Compute the chord length of a ray intersecting a polygon.
+
+    This function calculates the length of the segment formed by the intersection
+    of a ray (defined by an origin and direction) with a polygon defined by its
+    edges. The polygon is represented parametrically using starting points and
+    direction vectors for each edge.
+
+    Parameters
+    ----------
+    mu_x : float
```
mexanick, line 53: These should be quantities, as we discussed offline, the best would be to use telescope coordinate system lat/lon.

src/ctapipe/image/muon/intensity_fitter.py
```diff
@@ -39,6 +39,98 @@
...
+    ri_y : ndarray of shape (N,)
+        Y-coordinates of the starting points of the polygon edges.
+    vi_x : ndarray of shape (N,)
+        X-components of the edge direction vectors.
+    vi_y : ndarray of shape (N,)
+        Y-components of the edge direction vectors.
+
+    Returns
+    -------
+    float
```
mexanick, line 70: Should be also a unit of length (or angular, equivalent in the telescope frame)

src/ctapipe/image/muon/intensity_fitter.py
mexanick: Also, perhaps tuples of (x,y) or arrays of them would be more straightforward to define points (arrays of points)

src/ctapipe/image/muon/intensity_fitter.py
```diff
@@ -39,6 +39,98 @@
...
+        - accumulated segment length for multiple intersections (handles complex polygons).
+
+    Notes
+    -----
+    - The ray is defined parametrically as:
+          (x, y) = (mu_x, mu_y) + s * (cos(phi), sin(phi)), with s >= 0
+    - Each polygon edge is defined as:
+          (x, y) = (ri_x, ri_y) + t * (vi_x, vi_y), with 0 <= t < 1
+    - The function computes intersections by solving a 2D linear system.
+    - A small epsilon_d (`1e-20`) is added to the denominator to avoid division by zero.
```
mexanick, line 84: why not using
```
try:
  ...
except ZeroDivisionError:
  # determinant is zero, the cord goes on the edge, handle this
```

src/ctapipe/image/muon/intensity_fitter.py
maxnoe: Yes, I'd also prefer the vertices as input. But maybe the function as it is here is fine and that transformation is made once outside of this function.

src/ctapipe/image/muon/intensity_fitter.py
```diff
@@ -39,6 +39,98 @@
 SQRT2 = np.sqrt(2)
 
 
+def polygon_chord(mu_x, mu_y, phi, ri_x, ri_y, vi_x, vi_y):
```
maxnoe, line 42: This function should be `njit`, as it will be called. If possible, it would also be good if it could be vectorized over multiple tiles.

src/ctapipe/image/muon/intensity_fitter.py
kosack: Just remember that catching an exception is quite a bit slower than adding an epsilon, so if you need this to be fast, the epsilon method might be better. Also if you want to use numba to speed things up

maxnoe: And doesn't vectorize

src/ctapipe/image/muon/intensity_fitter.py
burmist-git: Yes agree with @maxnoe . This is a low-level function. Let's keep it as it is, since all the necessary computations should be handled by the higher-level function that calls it.

src/ctapipe/image/muon/intensity_fitter.py
burmist-git: Yes, I see. Let me finish implementing it with multiple tiles.

src/ctapipe/image/muon/intensity_fitter.py
Voutsi: Every branch returns (length, x_int, y_int) and not only a float. This should be reflected in the docstring.

src/ctapipe/image/muon/intensity_fitter.py
```diff
@@ -39,6 +39,98 @@
...
+    t = ( c1 * vmu_y - c2 * vmu_x ) / determinant
+    s = ( vi_y * c1 - vi_x * c2 ) / determinant
+
+    status = np.column_stack((vi_x, ri_x, vi_y, ri_y, t, s))
+    mask = (status[:,4] >= 0) & (status[:,4] < 1) & (status[:,5] >= 0)
+
+    x_int = status[mask][:,0]*status[mask][:,4] + status[mask][:,1]
+    y_int = status[mask][:,2]*status[mask][:,4] + status[mask][:,3]
+
+    if x_int.shape[0] == 0 :
```
Voutsi: I suggest to return the same type (an array) for every if loop branch. I don't say that the code is wrong, but returning the same type will make it cleaner.

src/ctapipe/image/muon/intensity_fitter.py
burmist-git: It is indeed only a float. The function is not vectorized.

src/ctapipe/image/muon/intensity_fitter.py
burmist-git: It is indeed only a float. The function is not vectorized. All of them return only one number.

burmist-git:
@Voutsi @kosack @maxnoe The code, tests, and documentation are ready. All tests are green and have passed. There is currently an issue with the SonarQube check related to code complexity (there are several nested if conditions). I know how to fix it, but before making the change, I want to make sure that the current architecture is appropriate and will be accepted for merging. I will fix it as soon as I receive feedback from the reviewers. For the tests with the full simulation chain (not the unit tests), I have requested the simulation from SimPipe.https://gitlab.cta-observatory.org/groups/cta-computing/dpps/-/work_items?sort=created_date&state=opened&search=muons&first_page_size=20&show=eyJpaWQiOiIyOTUiLCJmdWxsX3BhdGgiOiJjdGEtY29tcHV0aW5nL2RwcHMvY2FsaWJyYXRpb25waXBlbGluZS9jYWxpYnBpcGUiLCJpZCI6MTg4OTJ9

maxnoe:
> The code, tests, and documentation are ready.

The polygon_chord function is a low-level function. I'd expect it to be (maybe optionally) used in the muon fitter.

At the moment, this function is not used anywhere. I don't really see a benefit in adding this function without it being possible to actually use it in the muon fit.

maxnoe:
I also thought that we had concluded at the meeting in Geneva that we would need to handle individual mirror tiles, so what is the motivation for still adding this polygon approximation?

src/ctapipe/image/muon/intensity_fitter.py
```diff
@@ -39,6 +39,172 @@
...
+    are summed for every projection angle.
+    """
+
+    ri_x = []
+    ri_y = []
+    vi_x = []
+    vi_y = []
+
+    for ver_i in vertices_list:
+        ver_f = np.roll(ver_i, 1, axis=0)
```
maxnoe: In my experience, `np.roll` is extremely expensive, especially when called in a loop here, as it makes full copies of its input data.

In general I am much worried about the (runtime) performance of the code here, as it contains a lot of raw python loops without numba compilation for functions that are meant to be called in a likelihood optimization.

src/ctapipe/image/muon/intensity_fitter.py
```diff
@@ -39,6 +39,172 @@
...
+            return 0.0, np.nan, np.nan
+        return 0.0
+    elif x_int.shape[0] == 1:
+        if return_intersections:
+            return (
+                np.squeeze(np.sqrt((x_int - mu_x) ** 2 + (y_int - mu_y) ** 2)),
+                np.squeeze(x_int),
+                np.squeeze(y_int),
+            )
+        return np.squeeze(np.sqrt((x_int - mu_x) ** 2 + (y_int - mu_y) ** 2))
```
maxnoe: please do not duplicate computation of the same things in different branches. 

compute first, then return, i.e.

```
chord_length = ...
if return intersections:
    return chord_length, ...
return chord_length
```

burmist-git:
> I also thought that we had concluded at the meeting in Geneva that we would need to handle individual mirror tiles, so what is the motivation for still adding this polygon approximation?

The hexagonal shape of the individual tiles can also be used or any other shape. However, that was not really the conclusion we reached regarding the use of the tiles individually. We considered using the individual tiles as an option, and this is now possible.

burmist-git:
> > The code, tests, and documentation are ready.
> 
> The polygon_chord function is a low-level function. I'd expect it to be (maybe optionally) used in the muon fitter.
> 
> At the moment, this function is not used anywhere. I don't really see a benefit in adding this function without it being possible to actually use it in the muon fit.

Sure, this has to be done. I would consider adding it in the next PR, but if needed, I can add it now.

maxnoe:
As I am very worried about the performance of the code being added here (I think it will be prohibitively slow as is), I'd like to see the performance of it when used as part of the fitter before merging.

src/ctapipe/image/muon/intensity_fitter.py
burmist-git: For the optimization, let’s leave it for the next iteration. The Numba optimization can be applied to cases with zero, one, or two intersection points. However, for more complex shapes, such as the one we want to have for a the LST mirror (not the single facet), this requires a sorting function that cannot be used with Numba.

burmist-git:
> As I am very worried about the performance of the code being added here (I think it will be prohibitively slow as is), I'd like to see the performance of it when used as part of the fitter before merging.

These are the requirements - we need to fulfill them. As we agreed during the meeting, performance comes later as a requirement. So, this option should be available in ctapipe.

maxnoe:
I disagree, if we see here that the performance is as bad as I expect it to be from looking at the code here, this will not be a question of CalibPipe performance requirements, but just that I won't merge this in ctapipe.

burmist-git:
> I disagree, if we see here that the performance is as bad as I expect it to be from looking at the code here, this will not be a question of CalibPipe performance requirements, but just that I won't merge this in ctapipe.

Please add this comment here, under the Level C requirements for muons : https://gitlab.cta-observatory.org/cta-computing/dpps/calibrationpipeline/documentation/calibpipe-level-c-requirements/-/merge_requests/10

| C-CALP-0308 | Arbitrary description of the primary mirror shape. | CalibPipe shall account for the non-circularity of the reflectors when reconstructing the muon impact point. | T: Dedicated simulations of muon events with a fixed impact point shall be processed using CalibPipe. Qualification shall be conducted by comparing the reconstructed impact point against the ground truth, taking into account the corresponding uncertainties. | CTAO-B_DPPS-494, CTAO-B_DPPS-439, CTAO-B_DPPS-408 | UC-120.2.2, UC-120.2.2.1, UC-120.2.2.2, UC-120.2.2.3, UC-120.2.10, UC-120.2.10.1, UC-120.2.10.2, UC-120.2.10.3, UC-120.2.13, UC-120.2.14|

Now we are in a position to keep track of whether the requirements have been fulfilled or not, and why.

burmist-git:
@maxnoe please look at the functional requirements regarding C-CALP-0308 : https://gitlab.cta-observatory.org/cta-computing/dpps/calibrationpipeline/documentation/calibpipe-level-c-requirements/-/merge_requests/10#note_422640

burmist-git:
I understand the overall feedback and will work on optimizing the code and incorporating it into the intensity fitter.

Voutsi:
> I understand the overall feedback and will work on optimizing the code and incorporating it into the intensity fitter.

Thanks @burmist-git 

> I disagree, if we see here that the performance is as bad as I expect it to be from looking at the code here

@maxnoe , will be the optimisation discussed above enough in order for this PR to be acceptable? I am speaking only performance-wise, of course all other threads should be as well resolved.

burmist-git:
Just in case, @maxnoe, this is the optimization I’m proposing :

https://github.com/cta-observatory/ctapipe/pull/2969#discussion_r3751536383

src/ctapipe/image/muon/intensity_fitter.py
```diff
@@ -39,6 +39,172 @@
...
+    epsilon_d = 1.0e-20
+
+    c1 = mu_x - ri_x
+    c2 = mu_y - ri_y
+    determinant = vi_x * vmu_y - vi_y * vmu_x + epsilon_d
+
+    t = (c1 * vmu_y - c2 * vmu_x) / determinant
+    s = (vi_y * c1 - vi_x * c2) / determinant
+
+    status = np.column_stack((vi_x, ri_x, vi_y, ri_y, t, s))
```
maxnoe: Why do you `column_stack` this (making a copy) just to then index into the individual comments making it basically unreadable what you are accessing?

burmist-git: Yes, thank you. It can be removed. Initially, I did it to have everything in the same 2D array, but you’re right, it can be removed.

Voutsi:
Hi @burmist-git , a main concern for this PR is that it accepts a list of vertices which is not part of the telescope description. We should first clarify how this information will be provided to the fitter. The suggestion is to open a PR in ctapipe/instrument (could be a side PR to this one) which will make the geometry available here.

This PR should also discuss the point raised by @maxnoe , whether we need the individual facets, or the boundary polygon, or both.

burmist-git:
@Voutsi I fixed the vectorization issue. The individual facets are now fully vectorized, taking into account that the mirrors have the same shape and are convex polygons. I am finishing the tests, including performance tests and integration with the intensity fitter. Yes, we need both, and they are ready.

For the mirror shape and position description in the instrument, in my opinion, these should be two separate PRs.

burmist-git:
@Voutsi @maxnoe Please provide your feedback.

PR #2979, opened 2026-04-02
Add DL3 writer

mdebony:
Add a DL3 writer for the GADF format to catapipe. 

This PR is sub element of #2727 to simplify the review.


src/ctapipe/io/dl3.py
```diff
@@ -0,0 +1,1002 @@
...
+        np.ndarray
+            Interpolated angular values in degrees, in [0, 360).
+        """
+        fp_rad = np.deg2rad(np.asarray(fp_deg, dtype=float))
+        fp_unwrapped = np.unwrap(fp_rad)
+        result_rad = np.interp(x, xp, fp_unwrapped)
+        return np.rad2deg(result_rad) % 360
+
+    @staticmethod
+    def _circular_mean(angles_deg):
```
kosack: I'd suggest just using quantities here: 

```suggestion
    def _circular_mean(angles: u.Quantity[deg]) → u.Quantity[deg]:
```

And that simplifies the code, since numpy will use the correct angle unit automatically:

```
In [1]: a = [0,45,90] *u.deg

In [2]: np.sin(a)
Out[2]: <Quantity [0.        , 0.70710678, 1.        ]>

In [3]: np.sin(a.to_value("rad"))
Out[3]: array([0.        , 0.70710678, 1.        ])
```

Then below, you can drop using `np.rad2deg`, and it should all work.

src/ctapipe/io/tests/test_dl3.py
```diff
@@ -0,0 +1,685 @@
...
+        meta["pointing"] = {
+            "pointing_mode": PointingMode(row["pointing_mode"]).name,
+            "pointing_list": [(start_time, pointing), (stop_time, pointing)],
+        }
+        return meta
+
+
+@pytest.fixture(scope="session")
+def dl2_events_for_dl3(single_obs_gamma_diffuse_full_reco_file, dl2_meta_for_dl3):
+    preprocessor = DL2EventPreprocessor(
```
kosack, line 84: If you like, I can open a PR to this branch to convert this to use the new `EventPreprocessor`, which could simplify a lot of the code.

src/ctapipe/io/dl3.py
kosack: Or better yet, why not just use [astropy's implementation](https://docs.astropy.org/en/stable/api/astropy.stats.circstats.circmean.html)

kosack (changes requested):
I'm still working through this, so this is not yet a complete review, but here are a few initial comments

src/ctapipe/io/tests/test_dl3.py
kosack: See  #3004 where I started to see what changes I need to EventPreprocessor to support the DL2 to DL3 transition

src/ctapipe/io/dl3.py
```diff
@@ -0,0 +1,1002 @@
...
+
+from ..compat import COPY_IF_NEEDED
+from ..core import Component
+from ..core.traits import AstroTime, Bool
+from ..version import version as ctapipe_version
+
+__all__ = ["DL3EventsWriter", "DL3GADFEventsWriter"]
+
+
+class DL3EventsWriter(Component):
```
maxnoe: My main comment here is that I think that the DL3 writer should not store state of the observation. 

The inputs to the actual writing function should be the event list, the irfs and additionally needed data, but it should not keep that as state on the writer instance.

It should be possible to write multiple observation blocks using the same instance of the writer.

src/ctapipe/io/dl3.py
```diff
@@ -0,0 +1,1002 @@
...
+    ).tag(config=True)
+
+    reference_time = AstroTime(
+        default_value=Time("2018-01-01T00:00:00", scale="tai"),
+        help="The reference time that will be used in the FITS file",
+    ).tag(config=True)
+
+    def __init__(self, **kwargs):
+        super().__init__(**kwargs)
+        self._obs_id = None
```
maxnoe: It looks more like this is a data class storing all information that should go into a DL3 file, not a writer.

src/ctapipe/io/dl3.py
mdebony: Done

src/ctapipe/io/dl3.py
mdebony: I've added the a data class as suggested and it should address your comments. Let me know.

src/ctapipe/io/dl3.py
mdebony: Done

src/ctapipe/io/tests/test_dl3.py
mdebony: Please feel free to do it.

maxnoe:
@mdebony please rebase against the main branch to get the fixes for the CI

PR #2994, opened 2026-04-15
Add proper typing support for Container fields

maxnoe:
With e.g. pyright installed in an IDE:

Before:

Screenshot_20260415_125405

Now:

Screenshot_20260415_125335

And e.g. in `CameraCalibrator`:

Before:
Screenshot_20260415_131022

Now:

Screenshot_20260415_131045

It cannot deduce the type of the `Map`, but that can be explicitly type hinted as well later.


kosack (approved):

mexanick (approved):
thanks, great QoL improvement!

PR #2996, opened 2026-04-20
added HistogramsAggregator

### LINKED ISSUE #577: Pixel-wise histogramming
Hi all,

I am wondering if there is any plan to have a class/function to create an histogram along a given axis (e.g. pixel-wise).
#### Example on 2 waveforms: 
`data_cube = np.array([[1, 2, 2, 2, 5] ,[6, 7, 7, 8, 10]])
bins = np.arange(0, 11, 1)
hist = utils.histogram.1DHistrogram(bins=bins)
hist.fill(data)
hist_data = hist.data
print(hist_data)
`
[[0, 1, 3, 0, 0, 1, 0, 0, 0, 0, 0], [0, 0, 0, 0, 0, 0, 1, 2, 1, 0, 1]]

Right now the workaround I have for that in numpy is : 

`
hist_data = np.zeros((data.shape[0], bins.shape[0]))
for pixel in range(data.shape[0]):
     hist_data[pixel] +=  np.histogram(data[pixel], bins=bins)[0]
`
Of course this is taking much more time than if there was in numpy anything like : 
`
hist_data = np.histogram(data, bins=bins, axis=0)
`
I believe this could be useful if we want to do per pixel ADC distributions (for instance single p.e. per pixel or to check if a pixel is noisy). This might be linked to the point rose by @kosack in #571.

The issue with the actual event-wise reading and pixel-wise histogram construction is that it could have a huge performance impact if we need e.g. 10k waveforms of ~50 samples form ~ 2k pixels to build pixel-wise distributions.

Do you have any suggestions ? Or better workaround I missed ?
Thanks, Cyril

tjarkmiener:
closes #577


src/ctapipe/containers.py
```diff
@@ -1236,6 +1238,31 @@ class CameraCalibrationContainer(Container):
...
+    """Store histogram data of a pixel-wise quantity for each channel"""
+
+    counts = Field(
+        None,
+        "counts of a pixel-wise quantity for each channel"
+        "Type: float; Shape: (n_bins, n_channels, n_pixel)",
+    )
+    bins = Field(
+        None,
+        "bins of a pixel-wise quantity for each channel"
```
maxnoe: Is this the histogrammed data? That shouldn't be called bins, as e.g. np.histogram use the `bins` name for the edges.

I'd call it `histogram` or just `data`.

src/ctapipe/monitoring/aggregator.py
```diff
@@ -381,6 +397,124 @@ def _set_result_units(self, table, unit):
         pass
 
 
+class HistogramsAggregator(BaseAggregator):
+    """
+    Base component to handle the computation of aggregated histograms from a table
+    containing any event-wise quantities (e.g., images, scalars, vectors, or other arrays).
+
+    Aggregation is performed along axis=0 (the event dimension) for any N-dimensional data.
+    """
```
maxnoe, line 400: There should be configuration options here for the histogram, i.e. number of bins or directly the edges.

You could model this after the scikit histogram classes:
https://hist.readthedocs.io/en/latest/user-guide/axes.html#axis-names

or even just use it.

maxnoe:
Note that this kind of histogram monitoring is for example also needed for the cleaning algorithm implemented in EventDisplay (#1857), that uses the pedestal distribution (not just mean or std) in the cleaning.

maxnoe:
I think for these histograms to be useful, we need to ensure that each chunk is using the same binning. I.e. make the bin edges (or n_bins / min / max) configuration options.

see also inline comment about using scikit hist.

src/ctapipe/containers.py
mexanick: what's a difference between these `bins` and `counts` above besides the field name?

mexanick: Also since this is specific to the camera, I would avoid calling it with generic name such as "HistogramContainer", we may have histograms of other quantities, not per-pixel ones.

TjarkMiener: Those are the bin edges. I'm proposing to rename `bins` -> `edges` and `HistogramContainer` -> `CameraHistogramContainer`.

maxnoe: then the "shape" docstring is wrong. The edges have length `n_bins + 1`.

TjarkMiener:
> I think for these histograms to be useful, we need to ensure that each chunk is using the same binning. I.e. make the bin edges (or n_bins / min / max) configuration options.

yes, correct! We should ensure the same binnng.

> see also inline comment about using scikit hist.

I was unaware of the scikit hist. Do you think we need anything else then `np.linspace(self.range[0], self.range[1], self.n_bins + 1)`?

src/ctapipe/monitoring/aggregator.py
```diff
@@ -381,6 +389,109 @@ def _set_result_units(self, table, unit):
...
+
+    n_bins = Int(
+        default_value=10,
+        help="Number of bins for the histogram.",
+    ).tag(config=True)
+
+    range = List(
```
maxnoe: I'd suggest to use `min` and `max` instead of a range list, as it is always two values, not an arbitrary length list.

src/ctapipe/monitoring/aggregator.py
```diff
@@ -381,6 +389,109 @@ def _set_result_units(self, table, unit):
...
+        trait=Float(),
+        default_value=[1.0, 2.0],
```
maxnoe: These seem very arbitrary. 

I would not provide such defaults. Since we at the moment anyway load the full table into memory, you could add `default_value=None`, `allow_none=True` and that would take the range of the data.

src/ctapipe/monitoring/aggregator.py
```diff
@@ -381,6 +389,109 @@ def _set_result_units(self, table, unit):
...
+        masked_elements_of_sample,
+        results_dict,
+    ):
+        histos = self.compute_histos(
+            data,
+            masked_elements_of_sample=masked_elements_of_sample,
+        )
+        results_dict["n_events"].append(histos.n_events)
+        results_dict["counts"].append(histos.counts)
+        results_dict["edges"].append(histos.edges)
```
maxnoe: Since the edges are forced to be the same for all rows, I would not add them to the table.

Add them instead to the metadata of the table, like here:

https://gitlab.cta-observatory.org/cta-array-elements/lst/analysis/lstcam_calib/-/blob/main/src/lstcam_calib/tools/aggregate_drs4_timelapse_data.py?ref_type=heads#L177-183

(this is for a 2d hist)

src/ctapipe/containers.py
kosack: Maybe follow what we do in `SimulatedShowerDistribution`, which stores a histogram.

src/ctapipe/monitoring/aggregator.py
maxnoe: I don't understand the latest changes here. You had configuration options here, but now they are gone again.

How are people suppost to configure the pixel stats tool if there are no configuration options?

tjarkmiener: @maxnoe sorry for the confusion! Indeed, the last changes mixing up different things. The hist.axis should be configurable, so the user can set the config of the `HistogramsAggregator` accordingly in any tool. I will add a list of dicts that set up any `hist.axis` class with their corresponding kwargs. I think we should leave the pixel stats tool or better the `PixelStatisticsCalculator` component untouched, i.e. only allowing the user to set a `StatisticsAggregator` there. For the `HistogramsAggregator`, we do not need the functionality of the first_pass() and second_pass() and then storing the results in the monitoring file. In my opinion, histograms should be computed on-the-fly and then we should run the secondary operation, e.g. SPE fitting, and only store the high-level monitoring results.

maxnoe: For the usecase mentioned below (the MAGIC/EventDisplay cleaning) the histograms are the monitoring quantity and would need to be stored.

maxnoe: Also for the SPE fitting, I think we should _be able_ to store the distributions to which the SPE curve is fit, e.g. to produce quality plots for the fit results.

src/ctapipe/monitoring/aggregator.py
```diff
@@ -646,6 +649,8 @@ def compute_stats(
         n_bins = hist_object.axes[0].size
         hist_counts = hist_object.values()  # shape: (bins, n_pixels)
         hist_counts = hist_counts.reshape((n_bins,) + spatial_shape)
+        hist_variances = hist_object.variances()  # shape: (bins, n_pixels)
```
maxnoe: Does this actually provide additional information?

I would assume it is actually just returning the counts assuming poisson statistics in each bin, since we are not using any weights.

src/ctapipe/io/datawriter.py
```diff
@@ -44,8 +44,9 @@ def _get_tel_index(event, tel_id):
 #   (meaning readers need to update scripts)
 # - increase the minor number if new columns or datasets are added
 # - increase the patch number if there is a small bugfix to the model.
-DATA_MODEL_VERSION = "v7.5.0"
+DATA_MODEL_VERSION = "v7.6.0"
 DATA_MODEL_CHANGE_HISTORY = """
+- v7.6.0: - Add new column 'histogram' to the chunk aggreagted statistical values.
```
tjarkmiener:
```suggestion
- v7.6.0: - Add new column 'histogram' to the chunk aggregated statistical values.
```

src/ctapipe/monitoring/aggregator.py
```diff
@@ -518,4 +546,153 @@ def compute_stats(
...
+
+
+class HistogramAggregator(StatisticsAggregator):
+    """
+    Compute aggregated statistic values and histograms from a chunk of event-wise data using Hist.
+
+    Works with any N-dimensional event-wise data by aggregating along axis=0 (event dimension).
+    """
+
+    hist_axis_dict = Dict(
```
maxnoe: please don't put the type in the name, e.g. just `axis_definition` or `binning`.

src/ctapipe/monitoring/aggregator.py
```diff
@@ -518,4 +546,153 @@ def compute_stats(
...
+        allow_none=False,
+        help=(
+            "Dictionary that contains ``axis_class_name`` and ``kwargs`` "
+            "to construct a ``hist.axis.<axis_class_name>(**kwargs)`` instance. "
+            "E.g. ``{'axis_class_name': 'Regular', 'kwargs': {'bins': 40, 'start': 20.0, 'stop': 80.0}}``."
```
maxnoe: Could we not have the "kwargs" level?

E.g. use something like this:

```
kwargs = self.hist_axis.copy()
cls = kwargs.pop("axis_class_name")
axis = getattr(hist, cls)(**kwargs)
```

That would I think make the config file nicer to read, i.e. you'd have:

```
hist_axis:
   class_name: Regular
   bins: 40
   start: 20.0
   stop: 80.0
```

examples/tutorials/histogram_aggregation.py
```diff
@@ -0,0 +1,280 @@
...
+
+print(f"Number of chunks: {len(result)}")
+print(f"histogram shape per chunk: {result[0]['histogram'].shape}")
+print(f"bin edges shape per chunk: {result[0].meta['bin_edges'].shape}")
+print(f"bin centers shape per chunk: {result[0].meta['bin_centers'].shape}")
+print(f"n_events shape per chunk: {result[0]['n_events'].shape}")
+
+
+# -------------------------------------------------------------------
+# Inspect one pixel histogram in both chunks and both gain channels
```
mexanick: please add a better comment describing that you have two chunks by design, otherwise the line 
```
for chunk_index, ax in enumerate(axes):
```
looks very weird

src/ctapipe/monitoring/aggregator.py
```diff
@@ -360,7 +370,13 @@ def __call__(
         return result_table
 
     @abstractmethod
-    def _add_result_columns(self, data, masked_elements_of_sample, results_dict):
+    def _add_result_columns(
```
mexanick, line 385: please do not introduce a breaking change here. Instead of passing metadata as a mandatory argument here, you can e.g. add it as an object attribute.

src/ctapipe/monitoring/aggregator.py
```diff
@@ -360,7 +370,13 @@ def __call__(
...
+        self,
+        data,
+        masked_elements_of_sample,
+        results_dict,
+        metadata,
```
mexanick:
```suggestion
```

src/ctapipe/monitoring/aggregator.py
```diff
@@ -371,7 +387,9 @@ def _add_result_columns(self, data, masked_elements_of_sample, results_dict):
         masked_elements_of_sample : ndarray, optional
             Boolean mask of shape (\*data_dimensions) for elements to exclude
         results_dict : dict
-            Dictionary to which statistic columns should be added.
+            Dictionary to which statistic or histogram columns should be added.
+        metadata : dict
```
mexanick:
```suggestion
```

src/ctapipe/monitoring/aggregator.py
```diff
@@ -371,7 +387,9 @@ def _add_result_columns(self, data, masked_elements_of_sample, results_dict):
...
+            Shared metadata container that can be mutated by subclasses.
```
mexanick:
```suggestion
```

src/ctapipe/monitoring/aggregator.py
```diff
@@ -389,21 +407,30 @@ class StatisticsAggregator(BaseAggregator):
     Aggregation is performed along axis=0 (the event dimension) for any N-dimensional data.
     """
 
-    def _add_result_columns(self, data, masked_elements_of_sample, results_dict):
+    def _add_result_columns(
+        self,
+        data,
+        masked_elements_of_sample,
+        results_dict,
+        metadata,
```
mexanick:
```suggestion
```

src/ctapipe/monitoring/aggregator.py
```diff
@@ -389,21 +407,30 @@ class StatisticsAggregator(BaseAggregator):
...
+        results_dict["histogram"].append(stats.histogram)
+        if "meta" not in metadata and stats.meta:
+            metadata["meta"] = stats.meta
 
     def _set_result_units(self, table, unit):
         """
         Set units for statistics columns that inherit from the input data.
 
-        For StatisticsAggregator, the mean, median, and std columns
+        For StatisticsAggregator, the mean, median, std, and histogram columns
```
mexanick, line 657: why do you always add histogram, why don't we only compute it when needed?

src/ctapipe/containers.py
```diff
@@ -1254,6 +1254,11 @@ class StatisticsContainer(Container):
         "standard deviation of a pixel-wise quantity for each channel"
         "Type: float; Shape: (n_channels, n_pixel)",
     )
+    histogram = Field(
```
mexanick: I don't like mixing it like this. Moreover, certain statistics you can calculate from the histogram itself, so to me it is counter-intuitive. Please add a new container for the histograms and use where appropriate instead, if it is needed (is it?)

mexanick (changes requested):

src/ctapipe/containers.py
tjarkmiener: But then `HistogramAggregator` can not inherit from `StatisticsAggregator` which returns a `ChunkStatisticsContainer` for the `compute_stats()` function or am I missing something?

src/ctapipe/monitoring/aggregator.py
tjarkmiener: okay sounds good

src/ctapipe/containers.py
mexanick: Why don't you inherit from `BaseAggregator`?

tjarkmiener: Mainly because then it did not require any changes to the pixel calculator component nor pixel stats tool. Only a new configuration. Also the `HistogramAggregator` do compute the complete statistical description (mean, median, std) as other `StatisticsAggregator`s  (plain or sigma clipping).

src/ctapipe/tools/calculate_pixel_stats.py
```diff
@@ -205,17 +210,22 @@ def start(self):
                     )
             # Construct the output table name based on the event type and the selected column name
             output_table_name = f"{EventType(dl1_table['event_type'][0]).name.lower()}_{self.input_column_name}"
+            # Add a suffix to the output table name if the aggregated statistic values contain histograms
+            if isinstance(aggregator, HistogramAggregator):
+                table_grp = DL1_PIXEL_HISTOGRAMS_GROUP
```
maxnoe: should we make this a class attribute of the aggregator?

tjarkmiener: yes would be nice

src/ctapipe/tools/tests/test_calculate_pixel_stats.py
```diff
@@ -130,7 +134,7 @@ def test_calculate_pixel_stats_tool_with_histogram_aggregator(tmp_path, dl1_imag
 
     stats = read_table(
         output_file,
-        path=f"{DL1_PIXEL_STATISTICS_GROUP}/subarray_image/tel_{tel_id:03d}",
+        path=f"{DL1_PIXEL_HISTOGRAMS_GROUP}/subarray_image/tel_{tel_id:03d}",
```
maxnoe, line 139: Why does this test compute a histogram without changes to its config?

maxnoe: Sorry, nevermind, I was looking at a single commit's diff, not the PR diff.

src/ctapipe/tools/tests/test_calculate_pixel_stats.py
```diff
@@ -85,6 +89,62 @@ def test_calculate_pixel_stats_tool(tmp_path, dl1_image_file):
...
+    output_file = tmp_path / "subarray_image_hist_monitoring.dl1.h5"
+    config = Config(
+        {
+            "PixelStatisticsCalculatorTool": {
+                "allowed_tels": [3],
+                "input_column_name": "image",
+            },
+            "PixelStatisticsCalculator": {
+                "stats_aggregator_type": [
+                    ("type", "*", "HistogramAggregator"),
```
maxnoe, line 108: Do we support multiple aggregators? I.e. can I compute stats and histograms or multiple histograms in one run of the process tool?

src/ctapipe/tools/tests/test_calculate_pixel_stats.py
tjarkmiener: All tests here are also a little cringe with one event...

src/ctapipe/tools/tests/test_calculate_pixel_stats.py
tjarkmiener: no, we only support one at a time and merge them later with the merger tool. It is more motivated that every aggregation needs distinct configuration (see e.g. confgis for the pixel stats [here](https://gitlab.cta-observatory.org/cta-computing/dpps/calibrationpipeline/calibpipe/-/tree/main/docs/source/user_guide/telescope/camera/configuration?ref_type=heads))

maxnoe:
I tried running on images of pedestals using this config:

```
PixelStatisticsCalculatorTool:
  input_column_name: "image"

PixelStatisticsCalculator:
  stats_aggregator_type: [
    ["type", "*", "HistogramAggregator"],
  ]

HistogramAggregator:
  chunking_type: SizeChunking
  axis_definition:
    class_name: Regular
    bins: 100
    start: -25.0
    stop: 25.0
    name: "pixel_intensity"

SizeChunking:
  chunk_size: 100
```

and got this error:

```
❯ ctapipe-calculate-pixel-statistics -i lst_pedestals.h5 -o test.h5 -c hist.yaml --overwrite
2026-04-24 17:37:35,088 ERROR [ctapipe.ctapipe-calculate-pixel-statistics] (tool.run): Caught unexpected exception: The axis name value could not be found
  File "/home/maxnoe/CTAO/ctapipe/src/ctapipe/core/tool.py", line 437, in run
    self.start()
    ~~~~~~~~~~^^
  File "/home/maxnoe/CTAO/ctapipe/src/ctapipe/tools/calculate_pixel_stats.py", line 177, in start
    aggregated_stats = self.stats_calculator.first_pass(
        table=dl1_table,
        tel_id=tel_id,
        col_name=self.input_column_name,
    )
  File "/home/maxnoe/CTAO/ctapipe/src/ctapipe/monitoring/calculator.py", line 160, in first_pass
    aggregated_stats = aggregator(
        table=table,
        masked_elements_of_sample=masked_pixels_of_sample,
        col_name=col_name,
    )
  File "/home/maxnoe/CTAO/ctapipe/src/ctapipe/monitoring/aggregator.py", line 353, in __call__
    self._add_result_columns(
    ~~~~~~~~~~~~~~~~~~~~~~~~^
        chunk[col_name].data,
        ^^^^^^^^^^^^^^^^^^^^^
        masked_elements_of_sample,
        ^^^^^^^^^^^^^^^^^^^^^^^^^^
        results,
        ^^^^^^^^
    )
    ^
  File "/home/maxnoe/CTAO/ctapipe/src/ctapipe/monitoring/aggregator.py", line 442, in _add_result_columns
    histograms = self.compute_histograms(data, masked_elements_of_sample)
  File "/home/maxnoe/CTAO/ctapipe/src/ctapipe/monitoring/aggregator.py", line 496, in compute_histograms
    hist_object.fill(value=values, pixel=pix)
    ~~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/maxnoe/.local/conda/envs/cta-dev/lib/python3.13/site-packages/hist/basehist.py", line 319, in fill
    self._name_to_index(k) if isinstance(k, str) else k: v  # type: ignore[redundant-expr]
    ~~~~~~~~~~~~~~~~~~~^^^
  File "/home/maxnoe/.local/conda/envs/cta-dev/lib/python3.13/site-packages/hist/basehist.py", line 246, in _name_to_index
    raise ValueError(f"The axis name {name} could not be found")
ValueError: The axis name value could not be found
```

any idea?

The input file was produced from:

```
$  curl -LO https://minio-cta.zeuthen.desy.de/dpps-testdata-public/data/calibpipe-test-data/camera-calibration-test-data/pedestals_LST_dark.simtel.gz
$ ctapipe-process -i pedestals_LST_dark.simtel.gz -o lst_pedestals.h5 --write-images --SimTelEventSource.skip_calibration_events=False --overwrite --no-write-parameters --progress
```

maxnoe:
I needed to set the axis name to "value". 

If this is a hard requirement and it doesn't work with arbitrary user provided values, it should not be needed in the configuration.


Results look good though:

pedestal_hists

maxnoe:
I notice that I get min/mean/std etc. values in the table containing the histogram data. But I didn't specify a statisticsaggregator.

tjarkmiener:
> I notive that I get min/mean/std etc. values in the table containing the histogram data. But I didn't specify a statisticsaggregator.

The values are calculated based on the histograms by weighting the bin counts. Median is then the bin where the cumulative distribution is passing 50%. Let me find the code snippet and ping you there.

src/ctapipe/monitoring/aggregator.py
```diff
@@ -381,6 +404,161 @@ def _set_result_units(self, table, unit):
...
+                where=cdf_denominator[np.newaxis, ...] != 0,
+            )
+        median_idx = np.argmax(cdf >= 0.5, axis=0)
+        median = centers[median_idx]
+
+        # Mark elements with no valid entries as NaN.
+        invalid = counts_sum == 0
+        mean = np.where(invalid, np.nan, mean)
+        std = np.where(invalid, np.nan, std)
+        median = np.where(invalid, np.nan, median)
```
tjarkmiener: @maxnoe this is the code snippet to calc the values from the histo data

maxnoe: I am opposed to adding this adh-hoc, histogram based computed statistics when we have access to the actual values which could be used directly.

What is the benefit of computing values based after the loss of resolution due to the histogramming?

Why can users not just use the statistics aggregator and the histogram aggregator at the same time to get stats and histograms?

Why implement these stats manually here when scikit-hist already has this functionality?

https://hist.readthedocs.io/en/latest/user-guide/accumulators.html#weightedmean

tjarkmiener: sure, we can remove this caluclation, but then need to make sure that the stats pix tool only runs the first_pass.

src/ctapipe/monitoring/aggregator.py
```diff
@@ -381,6 +404,116 @@ def _set_result_units(self, table, unit):
...
+        results_dict["n_events"].append(histograms.n_events)
+        results_dict["histogram"].append(histograms.histogram)
+        if "meta" not in results_dict and histograms.meta:
+            results_dict["meta"] = histograms.meta
+
+    def _set_result_units(self, table, unit):
+        """
+        Set units for histogram columns that inherit from the input data.
+
+        For HistogramAggregator, the histogram columns should have the same units as the input data.
```
maxnoe, line 470: I think this is wrong. The unit of the histogram should be dimensionless, in the unweighted case.

The edges share the units with the input quantity.

src/ctapipe/monitoring/calculator.py
```diff
@@ -170,13 +166,14 @@ def first_pass(
...
             aggregator.chunking.chunk_shift = original_chunk_shift
 
-        # Detect faulty pixels with multiple instances of ``OutlierDetector``
-        # and append the outlier masks to the aggregated statistics
-        self._find_and_append_outliers(aggregated_stats)
-        # Get valid chunks and add them to the aggregated statistics
-        aggregated_stats["is_valid"] = self._get_valid_chunks(
-            aggregated_stats["outlier_mask"]
-        )
+        if not isinstance(aggregator, HistogramAggregator):
```
maxnoe: This is a violation of our design principles here. We want an abstract API for aggregators. If you add custom code dependent on the class of the aggregator, it means the abstraction is leaking and we should cover the case in the abstract API.

maxnoe: Why don't we run outlier detectors on the histograms? I could imagine specific outlier detection algorithms for histograms.

src/ctapipe/monitoring/aggregator.py
tjarkmiener: True

src/ctapipe/monitoring/calculator.py
tjarkmiener: Indeed, we can use the outlier detectors (ODs) on the histograms as well and maybe add tailored ones in the fuuture to the OD API. So we can use the component and tool without breaking changes.

src/ctapipe/monitoring/aggregator.py
```diff
@@ -381,6 +404,116 @@ def _set_result_units(self, table, unit):
...
 
 
+class HistogramAggregator(BaseAggregator):
+    """
+    Compute aggregated statistic values and histograms from a chunk of event-wise data using Hist.
+
+    Works with any N-dimensional event-wise data by aggregating along axis=0 (event dimension).
+    """
+
+    TABLE_GROUP = DL1_PIXEL_HISTOGRAMS_GROUP
```
mexanick: This looks like a dead code. What's the intended use? Also, why it is mentioning "pixel" and hardcoding "DL1"? Same question for the other Aggregator classes.

src/ctapipe/monitoring/aggregator.py
```diff
@@ -381,6 +404,116 @@ def _set_result_units(self, table, unit):
...
+
+    def _set_result_units(self, table, unit):
+        """
+        Set units for histogram columns that inherit from the input data.
+
+        For HistogramAggregator, the histogram columns should have the same units as the input data.
+        """
+        table["histogram"].unit = unit
+
+    def compute_histograms(
```
mexanick: Why is this a public method? Is it used outside of this class?

src/ctapipe/containers.py
```diff
@@ -1279,6 +1280,23 @@ class ChunkStatisticsContainer(StatisticsContainer):
...
+    time_start = Field(NAN_TIME, "high resolution start time of the chunk")
+    time_end = Field(NAN_TIME, "high resolution end time of the chunk")
+    event_id_start = Field(None, "event id of the first event of the chunk")
+    event_id_end = Field(None, "event id of the last event of the chunk")
+
+    n_events = Field(-1, "number of events used for the histogram calculation")
+
+    histogram = Field(
+        None,
+        "histogram of a pixel-wise quantity for each channel"
```
mexanick: why a pixel-wise quantity?

src/ctapipe/containers.py
```diff
@@ -1267,6 +1268,12 @@ class ChunkStatisticsContainer(ChunkContainer):
     std = Field(None, "standard deviation of the chunk distribution")
 
 
+class ChunkHistogramsContainer(ChunkContainer):
```
kosack: I'd name this singular `ChunkHistogramContainer` to match how we name others. It only stores one "histogram" (even if it's one with categories per pixel and channel)

src/ctapipe/containers.py
```diff
@@ -1267,6 +1268,12 @@ class ChunkStatisticsContainer(ChunkContainer):
...
+    """Container for histograms of the chunk distribution"""
+
+    histogram = Field(None, "histogram of the chunk distribution")
+
```
kosack, line 1286: I would be nice to have a helper function to convert this Container back to a Hist object.    Here is an example:


```python
def hist_from_container(
    cont: ChunkHistogramsContainer, axis_names=["pedestal", "channel", "pixel"]
) -> Hist:
    """Returns a Hist constructed from a stored ChunkHistogramsContainer."""

    bin_edges = cont.meta["bin_edges"]
    axes = [hist.axis.Variable(edges=bin_edges, name=axis_names[0])]

    # the rest of the dimensions
    for name, n_bins in zip(axis_names[1:], cont.histogram.shape[1:]):
        if n_bins == 2:
            axes.append(hist.axis.IntCategory(categories=np.arange(2), name=name))
        else:
            axes.append(
                hist.axis.Regular(bins=n_bins, start=0, stop=n_bins - 1, name=name)
            )

    h = Hist(*axes)
    h[...] = cont.histogram[...]
    return h
```

Then I can do things like:
```python
with HDF5TableReader("stats.h5") as reader:
    for i, container in enumerate(
        reader.read(
            table_name="/dl1/monitoring/telescope/calibration/camera/pixel_histograms/sky_pedestal_image/tel_001",
            containers=ChunkHistogramsContainer,
            prefixes=[""],
        )
    ):
        h = hist_from_container(container)

        fig, ax = plt.subplots(1, 3, figsize=(10, 3), layout="constrained")
        fig.suptitle(f"Chunk {i}")
        h[:, 0, :].plot(ax=ax[0], norm="log")
        h[:, 1, :].plot(ax=ax[1], norm="log")
        h.integrate("pixel").stack("channel").plot(ax=ax[2], legend=True)

        ax[0].set_title(f"Channel 0")
        ax[1].set_title(f"Channel 1")
        ax[2].set_title("Integral oval all pixels")
```

image


Ideally, you could also store the axis names  (`["pedestal", "channel_id", "pixel_id"]`)  in the metadata, that way making one of these plots is trivial from any file.

examples/tutorials/histogram_aggregation.py
```diff
@@ -0,0 +1,248 @@
...
+
+# -------------------------------------------------------------------
+# Initialize hist, fill it and plot via Hist functionality
+# -------------------------------------------------------------------
+
+# Create a Hist object with the same binning as the aggregator
+bin_edges = result[0].meta["bin_edges"]
+h = Hist(
+    hist.axis.Regular(len(bin_edges) - 1, bin_edges[0], bin_edges[-1], name="value")
+)
```
kosack, line 234: See prev comment: a helper function to do this would be nice.

src/ctapipe/monitoring/aggregator.py
```diff
@@ -381,6 +399,111 @@ def _set_result_units(self, table, unit):
...
+        hist_counts = hist_counts.reshape((n_bins,) + spatial_shape)
+        # Count valid entries per element (excludes masked and invalid values)
+        n_events_valid = np.sum(~flat_mask, axis=0).reshape(spatial_shape)
+        # Build and return the ChunkHistogramsContainer
+        return ChunkHistogramsContainer(
+            n_events=n_events_valid,
+            histogram=hist_counts,
+            meta={
+                "bin_edges": hist_object.axes[0].edges,
+                "bin_centers": hist_object.axes[0].centers,
```
kosack: would be nice to add some sort of title here as well, so we know what was being aggregated.  When making a plot, I would want to know the axes names as well (e.g. which one is pixel_id, channel_id). this could be stored.

kosack (changes requested):

src/ctapipe/containers.py
kosack: What we did in datapipe-testbench is to serialize all the necessary Hist info in the metadata: the axis names, the axis types (e.g. Regular, Variable, Category), units, etc.  That would work well here as well (and in the future I could use this class directly).

kosack:
Since you are using `Hist`, consider keeping track of **overflow** and **underflow** values: it lets you store them as extra bins in the histogram, for example, or you can write them in the metadata.   That would be very useful for cases where values fall outside the range.

src/ctapipe/containers.py
kosack: For column-wise access, you could also do a similar `hist_from_table()` that would then give back a 4d histogram, with an added time dimension (maybe use the mean-time of the chunk) . Then you can easily do many plots like "plot pixel 3's pedestal over time"

kosack: One more nice demo of using this if it's a Hist:
```python
fig, ax = plt.subplots(1,2, figsize=(10,3))
disp1 = CameraDisplay(geom, image=h[:,0,:].profile("pedestal").values(), ax=ax[0])
disp2 = CameraDisplay(geom, image=h[:,0,:].profile("pedestal").variances(), ax=ax[1])

ax[0].set_title(ax[0].get_title() +  " Pedestal")
ax[1].set_title(ax[0].get_title() +  " Pedestal Variance")

disp1.add_colorbar()
disp2.add_colorbar()
```

image

There, you don't even need the stats, however, as @maxnoe pointed out above, computing the mean and variance from a histogram isn't as precise as doing it from the events.

tjarkmiener:
Thanks for the useful comments @kosack. I took them into consideration, so the current version properly deal with under/overflow and keep track of the hist object configuration in the metadata so it can be reconstructed later using the helper function `hist_from_container()` as suggested.

src/ctapipe/monitoring/aggregator.py
tjarkmiener: This is currently a public method for consistency with the analogue in `StatisticsAggregator`, i.e. `compute_stats()`. We can consider making them private, but at the moment some tests outside the class are using this function (which should be minor to fix), and I’m also not sure whether any pipeline is calling it directly instead of going through `__call__()`. Please let me know if you prefer to make them private functions @mexanick.

tjarkmiener:
SonarQube is currently failing due to complexity issues in pre-existing code, although this change does add one additional complexity point. Please let me know if this is acceptable as-is, or if you’d prefer that I refactor the function to address the SonarQube complexity warning as part of this PR @maxnoe

src/ctapipe/monitoring/aggregator.py
mexanick: Base class doesn't define such interface, therefore I prefer to keep it private. Also I'm not sure how this even passed the CI, as all public methods should have documentation and this one doesn't.

src/ctapipe/monitoring/aggregator.py
```diff
@@ -381,6 +399,165 @@ def _set_result_units(self, table, unit):
...
+        self, data, masked_elements_of_sample
+    ) -> ChunkHistogramContainer:
+        # Build the histograms over the event dimension (axis=0) for each element of the data dimensions
+        spatial_shape = data.shape[1:]
+        # Broadcast mask to full shape
+        if masked_elements_of_sample is not None:
+            mask = np.broadcast_to(masked_elements_of_sample, data.shape)
+        else:
+            mask = np.zeros_like(data, dtype=bool)
+        # Mask invalid values (NaN, inf)
```
mexanick: I would not mask all invalid values like this. They are outliers of a different origin and should be treated differently. Null-values are underflow, inf values are overflow and NaNs are separate category. BTW, do we actually have NaNs here?

src/ctapipe/monitoring/aggregator.py
tjarkmiener: Sounds good to me @mexanick! It is implemented in [6a40c6a](https://github.com/cta-observatory/ctapipe/pull/2996/commits/6a40c6acaff9552f909dbeab4569664cdc70abf3). I thought it might be useful for the user  to also have a small wrapper for reconstructing a hist object from a tablerow by delegating the actual reconstruction to `hist_from_container()`. Let me know if you think it can be useful to add.

```
    @staticmethod
    def hist_from_tablerow(
        row: Row,
        axis_names: list[str] | None = None,
    ):
        """Build a ``hist.Hist`` from an Astropy table row produced by the
        ``HistogramAggregator``.

        This is a thin wrapper that constructs a ``ChunkHistogramContainer``
        from the provided ``row`` (reading ``n_events``, ``histogram`` and
        ``row.meta``) and delegates the actual reconstruction to
        :meth:`hist_from_container`.

        Note
        ----
        ``HistogramAggregator.__call__`` produces an ``astropy.table.Table``
        with one row per chunk. To rebuild the histogram for chunk ``i`` you
        can do::

            result_table = aggregator(event_table)
            hist = HistogramAggregator.hist_from_tablerow(result_table[i])

        This helper handles extracting the stored ndarray and metadata from
        the row and returns a complete ``hist.Hist`` object.

        Parameters
        ----------
        row : astropy.table.Row
            Table row created by ``HistogramAggregator.__call__``. Expected to
            contain a ``histogram`` ndarray with shape ``(n_bins, *data_dims)``,
            number of valid events in ``n_events``, and  metadata in ``row.meta``
            describing the original axis.
        axis_names : list[str] | None
            Optional axis names for non-value axes in the rebuilt ``Hist``.

        Returns
        -------
        hist.Hist
            Reconstructed histogram object with axes and counts populated.
        """

        cont = ChunkHistogramContainer(**dict(zip(row.colnames, row])))
        cont.meta = row.meta
        return HistogramAggregator.hist_from_container(cont, axis_names)

```

src/ctapipe/monitoring/aggregator.py
tjarkmiener: Yes good catch, it is kind of an artifact from where we did not had the under/overflow implemented. I will only mask NaNs here. Implemented in [30550f1](https://github.com/cta-observatory/ctapipe/pull/2996/commits/30550f12baa5d0284e66c207dcdf00cd189a6a79)

> BTW, do we actually have NaNs here?

For camera data, I'd not think so. But to keep the component generic I think we have to mask those NaN values.

mexanick (dismissed):

tjarkmiener:
@maxnoe can this PR please be marked for the `v0.31.0` milestone? For when is `v0.31.0` planned and do you think it is feasible to include the histogram aggregation? Thanks in advance!

maxnoe:
Looking at the hdf5 output shape, I'm a bit confused that the number of bins seems to be the first axis?

```
      histogram
int64[100,2,1855]
```

not a big deal, but I was wondering if there is a reason behind?

I would have expected [2, 1855, 100], i.e. `gains, pixels, bins`.

maxnoe (dismissed):

maxnoe:
Minor thing: the `HistogramAggregator` is the only aggregator not exposed at the `ctapipe.monitoring` level: 
https://github.com/cta-observatory/ctapipe/blob/1233e325bb2d33422c7270b48ec68e75535baddf/src/ctapipe/monitoring/__init__.py#L5

src/ctapipe/monitoring/aggregator.py
```diff
@@ -381,6 +399,224 @@ def _set_result_units(self, table, unit):
...
+        cont: ChunkHistogramContainer,
+        axis_names: list[str] | None = None,
+    ):
+        """Construct a ``hist.Hist`` from a ``ChunkHistogramContainer``.
+
+        Parameters
+        ----------
+        cont : ChunkHistogramContainer
+            Stored histogram container.
+        axis_names : list[str] | None
```
kosack: I think you could drop this `axis_names` parameter, and just store the axis names in the metadata like you do with the `axis_class_name`.  That way the user doesn't have to remember the axis names and order - it's stored in the output as metadata. That reduces human error.

maxnoe (approved):

mexanick (dismissed):

tjarkmiener:
> Looking at the hdf5 output shape, I'm a bit confused that the number of bins seems to be the first axis?
> 
> ```
>       histogram
> int64[100,2,1855]
> ```
> 
> not a big deal, but I was wondering if there is a reason behind?
> 
> I would have expected [2, 1855, 100], i.e. `gains, pixels, bins`.


I also noticed that this shape convention is not particularly convenient when browsing the HDF5 file with ViTables.

There is no strong reason behind the current ordering. It was mainly chosen because it makes filling along the `hist_axis` a bit more straightforward. It is also consistent with the expected input/output shapes used elsewhere: the input data has shape `(n_events, *data_dimensions)`, and the outlier detection code expects the aggregated dimension to come first:

```python
Parameters
----------
column : astropy.table.Column
    Column with chunk-wise aggregated statistic values (mean, median, or std)
    of shape (n_entries, n_channels, n_pixels), or histograms of shape
    (n_bins, n_channels, n_pixels).

Returns
-------
boolean mask
    Mask of outliers of shape (n_entries or n_bins, n_channels, n_pixels)
```

maxnoe (dismissed):

kosack (changes requested):
Looks good overall, just one change request below to store the axis names rather than require the user to know what they are.

src/ctapipe/monitoring/aggregator.py
tjarkmiener: `axis_names` actually depends on the input shape of `__call__()` so the `axis_names` can not be a class attribute. So we would need to decide if we want to add it in the `BaseAggregator.__call__()` or here in the user helper function.

maxnoe: I think @kosack means that you should _store_ the information in the data, not at the class level

maxnoe: So that it can be loaded from the `container.meta` here.

Please also call the input `container` or `hist_container`, not `cont`.

kosack: Yes, the idea was to just write the axis names to the output file via the Container metadata, just as you do with the axis types. Then the user doesn't need to specify them when turning it back into a Hist object. That prevents them from accidentally mis-labeling them or getting the order wrong.   In short: you serialize everything but the axis names, so please just also serialize those.

src/ctapipe/monitoring/aggregator.py
```diff
@@ -381,6 +407,220 @@ def _set_result_units(self, table, unit):
...
+            Parent of this component in the configuration hierarchy
+        """
+        super().__init__(config=config, parent=parent, **kwargs)
+        self.axis_kwargs = self.axis_definition.copy()
+        if "class_name" not in self.axis_kwargs.keys():
+            raise TraitError(
+                "The ``axis_definition`` trait is missing required key 'class_name'."
+            )
+        self.axis_class_name = self.axis_kwargs.pop("class_name")
+        self.axis_kwargs["name"] = "value"
```
maxnoe: I think this should be configurable or be determined from the input data.

But like this, all histograms will just be called "value".

src/ctapipe/monitoring/aggregator.py
tjarkmiener: Great, sounds good! Please check [d1ba714](https://github.com/cta-observatory/ctapipe/pull/2996/commits/d1ba714fea2073b1bbfcdced9c13ec397d4eff90) if the implementation of this comment is okay. Many thanks in advance! @maxnoe @kosack

src/ctapipe/monitoring/aggregator.py
tjarkmiener: This was set because https://github.com/cta-observatory/ctapipe/pull/2996#issuecomment-4314438381. As I understood it was not working with arbitrary names?

maxnoe: It wasn't working because there was a configuration for the name but other places assumed "value".

I'd prefer if it was possible to actually override the name.

maxnoe: I think the issue was that I set the name to something else (i.e. pedvar) but the container loaded it back as `value`, probably precisely because the names weren't serialized. BUt we do that now, so it should be possible?

tjarkmiener: I think it was the filling of the histogram [here](https://github.com/cta-observatory/ctapipe/blob/histo_agg/src/ctapipe/monitoring/aggregator.py#L620) which takes the name as argument.

tjarkmiener: We could do something like:

`hist_object.fill(**{self.hist_axis.name: valid_values})`

tjarkmiener: Yes worked in [3091a4b](https://github.com/cta-observatory/ctapipe/pull/2996/commits/3091a4bb4f464319b5479b2c429dd9ddfd15e635). Thanks for the comment! @maxnoe

kosack (changes requested):
I think you still have a bug:

1. the axis names when I run with the config below are "axis_1", "axis_2", and "value". I would expect "channel", "pixel", "pixel_intensity" 
2. I get a crash when reading the file back (see below):

```yaml
PixelStatisticsCalculatorTool:
  input_column_name: "image"

PixelStatisticsCalculator:
  stats_aggregator_type: [
    ["type", "*", "HistogramAggregator"],
  ]

HistogramAggregator:
  chunking_type: SizeChunking
  axis_definition:
    class_name: Regular
    bins: 50
    start: -25.0
    stop: 25.0
    name: "pixel_intensity"

SizeChunking:
  chunk_size: 40
```

```shell
ctapipe-calculate-pixel-statistics -i lst_pedestals.h5 --config stat_config.yaml  -o stats2.h5 --overwrite
```
Crash:
```python
import numpy as np
from ctapipe.containers import ChunkHistogramContainer
from ctapipe.io import HDF5TableReader, read_table
from ctapipe.monitoring import HistogramAggregator

with HDF5TableReader("stats2.h5") as reader:
    for i, container in enumerate(
        reader.read(
            table_name="/dl1/monitoring/telescope/calibration/camera/pixel_histograms/sky_pedestal_image/tel_001",
            containers=ChunkHistogramContainer,
            prefixes=[""],
        )
    ):
        h = HistogramAggregator.hist_from_container(container)

```
```
--> 601 axis_names = axis_kwargs_meta.pop("axis_names")
    603 # Reconstruct the original axis class so metadata round-trips exactly,
    604 # including Integer and transformed[/circular](http://localhost:8891/circular) Regular axes.
    605 axes = [axis_class(**axis_kwargs_meta)]

KeyError: 'axis_names'
```
Looking at the `container.meta`, it seems to have lost some of the axis_kwargs that we see in the file. I'm still looking to see why that could happen... it's a bit odd that the rest of the metadata looks ok, so maybe a bug in HDF5TableReader. It's like the "axis_names" field of the `axis_kwargs` field is just gone when I load it back.
```
{'axis_class_name': 'Regular',
 'axis_kwargs': {'bins': 50, 'start': -25.0, 'stop': 25.0, 'name': 'value'},
 'bin_centers': array([-24.5, -23.5, -22.5, -21.5, -20.5, -19.5, -18.5, -17.5, -16.5,
        -15.5, -14.5, -13.5, -12.5, -11.5, -10.5,  -9.5,  -8.5,  -7.5,
         -6.5,  -5.5,  -4.5,  -3.5,  -2.5,  -1.5,  -0.5,   0.5,   1.5,
          2.5,   3.5,   4.5,   5.5,   6.5,   7.5,   8.5,   9.5,  10.5,
         11.5,  12.5,  13.5,  14.5,  15.5,  16.5,  17.5,  18.5,  19.5,
         20.5,  21.5,  22.5,  23.5,  24.5]),
 'bin_edges': array([-25., -24., -23., -22., -21., -20., -19., -18., -17., -16., -15.,
        -14., -13., -12., -11., -10.,  -9.,  -8.,  -7.,  -6.,  -5.,  -4.,
         -3.,  -2.,  -1.,   0.,   1.,   2.,   3.,   4.,   5.,   6.,   7.,
          8.,   9.,  10.,  11.,  12.,  13.,  14.,  15.,  16.,  17.,  18.,
         19.,  20.,  21.,  22.,  23.,  24.,  25.])}
```


image

src/ctapipe/monitoring/aggregator.py
```diff
@@ -381,6 +407,219 @@ def _set_result_units(self, table, unit):
...
+        Parameters
+        ----------
+        hist_container : ChunkHistogramContainer
+            Stored histogram container.
+        """
+
+        # Extract axis information from container metadata
+        axis_kwargs_meta = hist_container.meta.get("axis_kwargs", {})
+        axis_class = getattr(hist.axis, hist_container.meta["axis_class_name"])
+        axis_names = axis_kwargs_meta.pop("axis_names")
```
kosack: Found the bug: do not use pop here, as it modifies the metadata for all rows!


```suggestion
        axis_names = axis_kwargs_meta.get("axis_names", None)
```

kosack (changes requested):

src/ctapipe/monitoring/aggregator.py
kosack: Wait, no that might not be the issue. I'll report back shortly sorry.

kosack: I think the way you are the axis_names is more complex than necessary: it could just be a class config option in `HistogramAggregator`, not part of `__call__`, right? they do not change between calls, and they are not needed for the base class, only for the HistogramAggregator.  

I.e. you can just have a `axis_names` config option,  or expand what is allowed in the `axis_definition` (but that so far only allows a single axis to be defined, whereas the internal axes are somehow implicit. 

At least, this would be the simplest way:
```yaml
PixelStatisticsCalculatorTool:
  input_column_name: "image"

HistogramAggregator:
  chunking_type: SizeChunking
  axis_definition: # this defines the value axis
    class_name: Regular
    bins: 50
    start: -25.0
    stop: 25.0
  axis_names: # the other axes are defined by the shape of the input column:
    - channel
    - pixel
    - value
```

or perhaps better, it could be like this, which makes a bit more sense, but is more complex to implement. Here the internal axes of the histogram are specified where you specifify the colum, which is really what defines them .I you use a column like "hillas_intensity",  you have no internal axes, so need no names. 

```yaml

PixelStatisticsCalculatorTool:
  input_column_name: "image"
  input_column_axis_names:  # for multi-dimensional input columns, define the internal axes
    - channel_id
    - pixel_id

HistogramAggregator:
  chunking_type: SizeChunking
  axis_definition: # this defines the value axis, which is the user-configurable one.
    class_name: Regular
    bins: 50
    start: -25.0
    stop: 25.0
    name: value  # or "intensity" or whatever. 
```

tjarkmiener: [9e7bb50](https://github.com/cta-observatory/ctapipe/pull/2996/commits/9e7bb50d78f82d4715a24543b03b7a0dcd75c281) set the axis names for the `PixelStatisticsCalculatorTool` to `["channel", "pixel"]`. Let me know if the user should be able to set there custom axis names through the configuration

tjarkmiener: I think the `pop()` is needed because of the init of the Hist() below:
 `axes = [axis_class(**axis_kwargs_meta)]`

maxnoe: But you mutate the container metadata in that operation. You need to make sure this doesn't happen, either by not using pop but get and creating the appropriate dicts fresh or by copying the meta of the container before calling `.pop`.

kosack: Also: The bug does indeed only happen on the *second* chunk loaded in the example I had in the last review. So it does seem to have to do with a pop somewhere with a missing `copy()` on the metadata. The first chunk loads correctly and finds the metadata.

src/ctapipe/monitoring/aggregator.py
```diff
@@ -381,6 +407,219 @@ def _set_result_units(self, table, unit):
...
+        """Construct a ``hist.Hist`` from a ``ChunkHistogramContainer``.
+
+        Parameters
+        ----------
+        hist_container : ChunkHistogramContainer
+            Stored histogram container.
+        """
+
+        # Extract axis information from container metadata
+        axis_kwargs_meta = hist_container.meta.get("axis_kwargs", {})
```
kosack: This is the real bug you need to copy it, not modify it in place:
```python
axis_kwargs_meta = hist_container.meta.get("axis_kwargs", {}).copy()
```

That at least fixes the crash.  Need to add a test as well.

src/ctapipe/monitoring/aggregator.py
kosack: > [9e7bb50](https://github.com/cta-observatory/ctapipe/pull/2996/commits/9e7bb50d78f82d4715a24543b03b7a0dcd75c281) set the axis names for the PixelStatisticsCalculatorTool to ["channel", "pixel"]. Let me know if the user should be able to set there custom axis names through the configuration

Ok will take a look now. I think hard-coding them for the PixelStatisticsCalculatorTool is fine, since it's about PixelStatistics.

tjarkmiener: hold on for one sec please @kosack

tjarkmiener: I implmented the comment to make the axis names as class traits. It is much simpler and now you can set just in the config of the PixelStatsTool your axis_names in the HistogramAggregator config section. See [85d8dcf](https://github.com/cta-observatory/ctapipe/pull/2996/commits/85d8dcf6ca287678bbb6a088ad7ce83f2f420a4c). Not sure if this solves the metadata issues on the second chunk.

src/ctapipe/monitoring/aggregator.py
kosack: You still need to fix this - it breaks looping through chunks

kosack (changes requested):

kosack:
It's fine to hard-code the axes to pixel/channel in PixelStatisticsCalculator. THey do not need to be user-configurable there, since it only supports the "image" column, right? If we had a more general tool that can histogram any column, then being user-configurable would help.   So maybe just keep the user-configurability but make the defaults be "channel", "pixel" for that tool.   Or just revert that last commit

src/ctapipe/monitoring/aggregator.py
maxnoe: but the bug is still here. You `.get()` from the container meta, which returns the instance and later pop from it, mutating the meta of the container.

maxnoe: In 598 you `.get()`, in 600, you `.pop()`, mutating the container.

maxnoe (approved):

maxnoe (approved):

kosack (approved):
Looks good now. 

The only thing missing is storing the outliers (which is supported by Hist).  That will be necessary if you want it to be useful, but we can open an issue for that and add it as a feature in a future PR if you want to get this in quickly.

maxnoe:
@TjarkMiener Let's merge this and do improvements in new contributions.

PR #2999, opened 2026-04-27
reorganized stats/chunk containers

tjarkmiener:
We noticed in #2996 that containers dealing with chunks and stats needed some maintenance and docstring polishing.

### DIFF
--- a/src/ctapipe/containers.py
+++ b/src/ctapipe/containers.py
@@ -61,6 +61,7 @@ __all__ = [
     "TelescopePointingContainer",
     "ArrayPointingContainer",
     "StatisticsContainer",
+    "ChunkContainer",
     "ChunkStatisticsContainer",
     "ImageStatisticsContainer",
     "IntensityStatisticsContainer",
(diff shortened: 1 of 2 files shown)


kosack (dismissed):
Looks good, just need a changelog entry

mexanick (dismissed):

src/ctapipe/containers.py
```diff
@@ -1237,46 +1238,33 @@ class CameraCalibrationContainer(Container):
 
 
 class StatisticsContainer(Container):
```
maxnoe, line 1240: This container is used in the `HDF5EventSource`, but now has much fewer fields. Yet the code in the source is the same.

Why?

tjarkmiener: It is used only in the `HDF5MonitoringSource`, where we only fill the mean, median, and std, i.e. the "interpolated statistical values".

src/ctapipe/containers.py
```diff
@@ -1237,46 +1238,33 @@ class CameraCalibrationContainer(Container):
...
+    outlier_mask = Field(None, "boolean mask indicating outliers in the chunk")
+    is_valid = Field(False, "true if chunk values are valid")
+
+
+class ChunkStatisticsContainer(ChunkContainer):
+    """Container for descriptive statistics of the chunk distribution"""
+
+    mean = Field(None, "mean value of the chunk distribution")
+    median = Field(None, "median value of the chunk distribution")
+    std = Field(None, "standard deviation of the chunk distribution")
```
tjarkmiener, line 1267: I was thinking of replacing this duplication by
```
class ChunkStatisticsContainer(ChunkContainer):
    """Container for descriptive statistics of the chunk distribution"""
    stats = Field(
        default_factory=StatisticsContainer,
        description="Statistical description of the chunk distribution",
    )
```
which would be nice but would also lead to breaking changes here and elsewhere.
@mexanick @maxnoe @kosack

src/ctapipe/containers.py
maxnoe: ah ok , sorry, misread the filename.

src/ctapipe/containers.py
maxnoe: What we do elsewhere is have separate containers for the index and the data and write multiple containers to the same table.

I.e. you could have the `ChunkContainer`, `StatsContainer` and a `HistogramContainer` and write like this:

```
writer.write(table, (chunk_container, stats_container))
```

like this, you also don't need separate containers for the interpolated result and the chunk storage.

maxnoe: See e.g.: 

https://github.com/cta-observatory/ctapipe/blob/87056062f170bf12133da98304abdce9929ce966/src/ctapipe/io/datawriter.py#L309-L312

tjarkmiener: I think `n_events` should be moved to the `StatisticsContainer` and then `compute_stats()` from the  `PlainAggregator` and `SigmaClippingAggregator` should return `StatisticsContainer` and the `compute_histo() `from `HistorgramAggregator` from #2996 should return a `HistogramContainer`

maxnoe: mmh, that seems a bit weird. the `n_events` is a property of the chunk and should be the same for all chunk aggregations, regardless of whether you compute a histogram or stats or something else.

`n_events` should be really the number of events inside the chunk, which is very interesting in case of time-based chunks.

It should also be independent from which values are actually used due to e.g. outlier detection, sigma clipping or under/overflow.

tjarkmiener: right, the number of events are related to the chunks and the number of entries are related to the number of values used in the aggregation.

tjarkmiener: @maxnoe Coming back to this now; sorry for the delay! I’d slightly prefer to keep the changes as they are for now and avoid introducing breaking changes, since the aggregation and chunking components are already used across various pipelines. We could tackle a major reorganization of the containers after this PR and the dependent histogramaggregation PR #2996.

tjarkmiener:
After the docs fix #3007 has been merged the CI is not running here and in #2996. Can you please have a look and help out, @maxnoe? Thanks a lot in advance!

EDIT: After the latest push, all CI jobs are running! Sorry for the noise!

maxnoe (approved):

mexanick (approved):

PR #3009, opened 2026-05-27
refactor to reduce cognitive complexity

mexanick:


src/ctapipe/tools/calculate_pixel_stats.py
```diff
@@ -138,95 +138,112 @@ def setup(self):
         )
 
     def start(self):
-        # Iterate over the telescope ids and calculate the statistics
+        """Iterate over all telescopes, process their statistics, and save the results."""
+        # Track the last used table group and name for the summary log statement
+        output_table_name = ""
+        table_group = ""
```
tjarkmiener, line 144: I'd slightly prefer to set to `None` and then check for `not None`. But no strong opinion about it.

tjarkmiener (approved):
Thanks for taking the initiative to solve the SonarQube issue. Minor comment which can be ignored (if so please merge into the `histo_agg` branch).

src/ctapipe/tools/calculate_pixel_stats.py
mexanick: In our scenario empty string is a valid part of scenario (do not log in this case), therefore I prefer to have it as a string and not as None.

PR #3011, opened 2026-05-29
Adding a single documentation page with image parameters (features) 

vuillaut:
This PR introduces a new entry in the user doc "analysis steps" (could be completed later) including a page with all features definitions.
Newcomers find it rather confusing to not have a list of features definitions in the doc.
I also tried to clarify some definitions.

### DIFF
--- a/src/ctapipe/containers.py
+++ b/src/ctapipe/containers.py
@@ -304,9 +304,11 @@ class BaseHillasParametersContainer(Container):
     be assigned to an ImageParametersContainer as well.
     """
 
-    intensity = Field(nan, "total intensity (size)")
-    skewness = Field(nan, "measure of the asymmetry")
-    kurtosis = Field(nan, "measure of the tailedness")
+    intensity = Field(nan, "Total integrated charge (often called 'size')")
+    skewness = Field(nan, "Longitudinal asymmetry of the image")
+    kurtosis = Field(
+        nan, "Pearson's definition of kurtosis (Normal=3.0), measure of the tailedness"
+    )
 
 
 class CameraHillasParametersContainer(BaseHillasParametersContainer):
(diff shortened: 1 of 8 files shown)


src/ctapipe/containers.py
```diff
@@ -304,9 +304,11 @@ class BaseHillasParametersContainer(Container):
     be assigned to an ImageParametersContainer as well.
     """
 
-    intensity = Field(nan, "total intensity (size)")
-    skewness = Field(nan, "measure of the asymmetry")
-    kurtosis = Field(nan, "measure of the tailedness")
+    intensity = Field(nan, "Total integrated charge (often called 'size')")
+    skewness = Field(nan, "Longitudinal asymmetry of the image")
+    kurtosis = Field(
+        nan, "Pearson's definition of kurtosis (Normal=3.0), measure of the tailedness"
```
maxnoe: Should also mention "Longitudinal"

src/ctapipe/containers.py
```diff
@@ -344,22 +346,28 @@ class HillasParametersContainer(BaseHillasParametersContainer):
...
     fov_lat = Field(
         nan * u.deg,
-        "latitude angle in a spherical system centered on the pointing position",
+        "Centroid latitude angle in a spherical system centered on the pointing position",
         unit=u.deg,
     )
-    r = Field(nan * u.deg, "radial coordinate of centroid", unit=u.deg)
-    phi = Field(nan * u.deg, "polar coordinate of centroid", unit=u.deg)
+    r = Field(
+        nan * u.deg, "Radial coordinate of the centroid center of gravity", unit=u.deg
```
maxnoe: I think it should be "centroid" or "center of gravity", not both

maxnoe:
Thanks a lot, better docs is always appreciated! Two minor comments inline

src/ctapipe/containers.py
```diff
@@ -344,22 +346,28 @@ class HillasParametersContainer(BaseHillasParametersContainer):
...
+    )
+    phi = Field(
+        nan * u.deg, "Polar coordinate of the centroid center of gravity", unit=u.deg
```
kosack: I think in general I prefer "centroid" to "center of gravity" (since this is not mass).  As @maxnoe said above, centroid == center-of-gravity, so using both is redundant.

vuillaut:
Thanks for the feedback that I took into consideration

maxnoe (approved):

kosack (approved):

PR #3013, opened 2026-06-01
Do not use events outside fov bins for cut optimization

LukasBeiske:
Since we do not use fov bins in the cut optimization yet, events outside the fov bins used for the irfs
were used in the cut optimization, which results in cuts which do not actually maximize the sensitivity
for the relevant fov.
This fixes that issue until we implement fov bins for the cut optimization.

This was found in #2789

### DIFF
--- a/docs/changes/3013.bugfix.rst
+++ b/docs/changes/3013.bugfix.rst
@@ -1,4 +1,3 @@
-Since we do not use fov bins in the cut optimization yet, events outside the fov bins used for the irfs
-were used in the cut optimization, which results in cuts which do not actually maximize the sensitivity
-for the relevant fov.
-This fixes that issue until we implement fov bins for the cut optimization.
+Fix a bug in the event weighting that resulted in events outside of the selected FoV range for the sensitivity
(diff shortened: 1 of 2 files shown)


docs/changes/3013.bugfix.rst
```diff
@@ -0,0 +1,4 @@
+Since we do not use fov bins in the cut optimization yet, events outside the fov bins used for the irfs
+were used in the cut optimization, which results in cuts which do not actually maximize the sensitivity
+for the relevant fov.
+This fixes that issue until we implement fov bins for the cut optimization.
```
maxnoe: I think this changelog entry is a bit hard to understand in the context of reading the changes introduced in a release, users do not see the diff there.

Maybe formulate it a bit more along the lines of:

> Fix a bug in the event weighting that resulted in events outside of the selected FoV range for the sensitivity computation being used in the cut optimization leading to results where the selected cuts did not necessarily provide the best sensitivity in the selected FoV range.

maxnoe (approved):

kosack (approved):

PR #3014, opened 2026-06-03
Make function to compute true disp public in ctapipe.reco

### LINKED ISSUE #2950: Calculate true_disp when running ctapipe-process
Currently (as far as I can tell), ctapipe-process does not store the true_disp per-telescope at any point when processing data from dl0->dl3. This is despite code to calculate this quantity being present in the train-disp-reconstructor tool. It would be useful if it did for the purposes of debugging angular reconstruction, and would save it needing to be calculated every time train-disp-reconstructor is run.

At the dl0-dl1 stage, ctapipe-process adds true_disp_norm and true_disp_sign fields to the root.simulation.event.telescope.parameters.tel_x[:] containers.

maxnoe:
```python
from ctapipe.io import TableLoader
from ctapipe.reco.disp import compute_true_disp

with TableLoader(path) as loader:
    tel_events = loader.read_telescope_events(telescopes=[1], observation_info=True)

tel_events["true_disp"] = compute_true_disp(tel_events)
```

Some plots:

true_disp_vs_impact
true_disp_vs_energy

### DIFF
--- a/src/ctapipe/reco/disp.py
+++ b/src/ctapipe/reco/disp.py
@@ -2,17 +2,40 @@
 Common functions related to the disp reconstruction.
 """
 
+import warnings
 from typing import Annotated
 
 import astropy.units as u
 import numpy as np
 from astropy.table import Table
 
+from ..containers import CoordinateFrameType
 from .preprocessing import horizontal_to_telescope
 
 __all__ = [
     "compute_true_disp",
 ]
+
+
+def _get_tel_pointing(table):
+    prefix = "telescope_pointing_"
+    tel_alt = f"{prefix}_altitude"
+    tel_az = f"{prefix}_azimuth"
+
+    if {tel_alt, tel_az}.issubset(table.colnames):
+        return table[tel_alt].quantity, table[tel_az].quantity
+
+    prefix = "subarray_pointing"
+    if not np.all(table[f"{prefix}_frame"] == CoordinateFrameType.ALTAZ.value):
+        raise ValueError(
+            "Subarray pointing information for disp computation"
+            " has to be provided in horizontal coordinates"
+        )
+
+    warnings.warn(
+        "Input table does not contain telescope pointings, falling back to array pointing"
+    )
+    return table[f"{prefix}_lat"].quantity, table[f"{prefix}_lon"].quantity
 
 
 def compute_true_disp(
@@ -36,11 +59,13 @@ def compute_true_disp(
     disp:
         Disp as a signed quantity
     """
+    pointing_alt, pointing_az = _get_tel_pointing(table)
+
     fov_lon, fov_lat = horizontal_to_telescope(
         alt=table["true_alt"],
         az=table["true_az"],
-        pointing_alt=table["subarray_pointing_lat"],
-        pointing_az=table["subarray_pointing_lon"],
+        pointing_alt=pointing_alt,
+        pointing_az=pointing_az,
     )
 
     # numpy's trigonometric functions need radians
--- a/src/ctapipe/reco/disp.py
+++ b/src/ctapipe/reco/disp.py
@@ -17,7 +17,26 @@ __all__ = [
 ]
 
 
-def _get_tel_pointing(table):
+def get_tel_pointing(table):
+    """
+    Get telescope pointing from telescope events table.
+
+    Prefers columns telescope_pointing_{altitude,azimuth} but will fallback
+    to subarray_pointing_{lat,lon,frame} for backwards compatibility with older
+    datasets.
+
+    Parameters
+    ----------
+    table : Table
+        table of telescope events.
+
+    Returns
+    -------
+    alt : u.Quantity[angle]
+        pointing altitude
+    az : u.Quantity[angle]
+        pointing azimuth
+    """
     prefix = "telescope_pointing"
     tel_alt = f"{prefix}_altitude"
     tel_az = f"{prefix}_azimuth"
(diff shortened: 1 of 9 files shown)


maxnoe:
I think doing this at the table level is preferable to implementing it in the event loop in ctapipe-process, so I'll connect it to closing #2950

STSpencer (dismissed):
Hi @maxnoe,

Thanks for working on this PR! Everything looks good to me.

All the best,

Sam

src/ctapipe/reco/disp.py
```diff
@@ -0,0 +1,62 @@
...
+    delta_lon = fov_lon - cog_lon
+    delta_lat = fov_lat - cog_lat
+
+    true_disp = np.cos(psi) * delta_lon + np.sin(psi) * delta_lat
+    true_sign = np.sign(true_disp)
+
+    if project_disp:
+        true_norm = np.abs(true_disp)
+    else:
+        true_norm = np.sqrt((fov_lon - cog_lon) ** 2 + (fov_lat - cog_lat) ** 2)
```
kosack, line 104: wouldn't it be better to return this earlier if `project_disp` is False? Saves some unused computation

src/ctapipe/reco/disp.py
```diff
@@ -0,0 +1,62 @@
...
+        Disp as a signed quantity
+    """
+    fov_lon, fov_lat = horizontal_to_telescope(
+        alt=table["true_alt"],
+        az=table["true_az"],
+        pointing_alt=table["subarray_pointing_lat"],
+        pointing_az=table["subarray_pointing_lon"],
+    )
+
+    # numpy's trigonometric functions need radians
```
kosack, line 90: they actually support quantities, no need to convert.

```python
np.cos(np.pi/2*u.rad), np.cos(90*u.deg)
```
```
(<Quantity 6.123234e-17>, <Quantity 6.123234e-17>)
```

src/ctapipe/reco/disp.py
kosack: Also, it may be clearer to use something like `method="projected"` or `method="euclidean"` here.

src/ctapipe/reco/disp.py
```diff
@@ -0,0 +1,62 @@
...
+    psi = table["hillas_psi"].quantity.to_value(u.rad)
+    cog_lon = table["hillas_fov_lon"].quantity
+    cog_lat = table["hillas_fov_lat"].quantity
```
kosack, line 93: Is this really the "true" disp if you mix true pointing position and reconstructed shower parameters (which include noise)? Shouldn't you use the true (noise-free) hillas parameters? I guess it depends on what this is intended to be used for.

kosack (changes requested):

src/ctapipe/reco/disp.py
maxnoe: > Shouldn't you use the true (noise-free) hillas parameters? 

No, because we train a machine learning model to predict the distance from the reconstructed cog, with given cleaning method / threshold etc.

true_disp is a bit special here, as it depends on both the Hillas parameters computed on the image and the true source position.

This is the main reason why I don't want it to be part of the `event.simulation` tree.

src/ctapipe/reco/disp.py
maxnoe: There is no early exit here

maxnoe: The other computatations are necessary for `disp_sign`

src/ctapipe/reco/disp.py
maxnoe: For performance reasons we convert to numpy arrays. Those numpy arrays need to be in radians as when you pass arrays and not quantities, numpy expects radians

maxnoe: Like this, the conversion happens only once, not twice for sin and cos

maxnoe:
@kosack I moved the existing function to a new place, I'd rather not make any actual changes to the code here.

src/ctapipe/reco/disp.py
kosack: Yeah, that's why the name "true_disp" is a bit confusing, as all the "true_x" parameters are in the simulation tree. So I would suggest renaming this. Maybe "expected_disp"?

maxnoe: This PR was about moving functionality to be better accessible, do we really need to change terminology here? It is "true" in the sense that it is the "true" input for the machine learning training.

maxnoe:
@codex review

maxnoe:
The required changes here to implement this properly became a bit larger than planned.

The main point is that in the current data format, we do have good access to individual telescope pointings in the simulations, which should be preferred over the array pointing, obviously for divergent pointing.

However, the old test files we use for the tests of the machine learning code do not have this, thus I added (with help of codex) unit tests for now for the `compute_true_disp` function.

We should however urgently address #3031 to have datasets suitable for training tests that are using the current data format.

maxnoe:
@codex review

src/ctapipe/tools/utils.py
```diff
@@ -99,25 +99,23 @@ def read_training_events(
         true_parameters=False,
         instrument=True,
         observation_info=True,
+        pointing=True,
```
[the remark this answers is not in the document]
maxnoe: Pointing will be used also by energy / particle classifier in the future for the declination lines. Leaving it in.

src/ctapipe/reco/disp.py
kosack: Ok, it's fine

kosack (approved):

STSpencer (approved):

PR #3015, opened 2026-06-04
Improve CI caching

maxnoe:


maxnoe:
It seems the cache action only supports unpacking into the ci dir, so we move all the caching to `.cache`.

mexanick:
I think we need cache restoration for docs too

mexanick:
also tests seems to not locate cache correctly... One of the failed test job got http timeout trying to access the minio, which it should not have attempt if I understand the logic correctly.

maxnoe:
> One of the failed test job got http timeout trying to access the minio, which it should not have attempt if I understand the logic correctly.

I think this happened because the cached data was uploaded by a test not actually running all tests that need input data. I now ensured that the cache is populated by the "codecov" job, which should be the test downloading all neded data files.

mexanick (approved):

kosack (approved):

PR #3017, opened 2026-06-05
Debug caching

maxnoe:


maxnoe:
The cache never successfully uploaded on the main, which means it's not available to PRs.

The issue seems to be that upload keys have to be unique, globally, but keys for restoration are prefix-matched.

Make the upload key unique, restore key is the base prefix.

PR #3023, opened 2026-06-08
Make sure ctapipe-dump-instrument -f service supports different types with same name

maxnoe:


src/ctapipe/tools/dump_instrument.py
```diff
@@ -339,57 +339,39 @@ def write_service_data(self, subarray_id=1, site=None):
...
 
         # Write shared files once per unique telescope type
-        for type_key, first_desc in type_to_first_desc.items():
+        for index, tel_desc in enumerate(sub.telescope_types):
+            type_key = f"TEL-TYPE-{index:02d}"
             type_dir = array_elements_dir / type_key
             type_dir.mkdir(exist_ok=True, parents=True)
             self.log.debug("Writing shared type directory %s", type_key)
 
-            optics_table = QTable()
```
mexanick, line 356: I see you've removed a number of fields from the table meta and put them back in the table itself. Is it something we want to have? We've changed it to be in the meta because this is something that normaly only change with the change of design, not with the evolution of the telescope.

maxnoe: That wasn't intentional, I just wanted to avoid the code duplication and re-use the existing `subarray.to_table(kind=optics)`.

What actually changed in the output?

mexanick: Before your change:
```
# %ECSV 1.0
# ---
# datatype: []
# meta: !!omap
# - {TAB_VER: '4.0'}
# - {SOURCE: /Users/mdalchen/.cache/ctapipe/minio-cta.zeuthen.desy.de/dpps-testdata-public/data/ctapipe-test-data/v1.1.0/gamma_prod5.simtel.zst}
# - {optics_name: LST}
# - {size_type: LST}
# - {reflector_shape: PARABOLIC}
# - {n_mirrors: 1}
# - {equivalent_focal_length: 28.0 m}
# - {effective_focal_length: 29.30565071105957 m}
# - {mirror_area: 386.7332458496094 m2}
# - {n_mirror_tiles: 198}
# schema: astropy-2.0
```

after your change:
```
# %ECSV 1.0
# ---
# datatype:
# - {name: optics_name, datatype: string}
# - {name: size_type, datatype: string}
# - {name: reflector_shape, datatype: string}
# - {name: mirror_area, unit: m2, datatype: float64}
# - {name: n_mirrors, datatype: int64}
# - {name: n_mirror_tiles, datatype: int64}
# - {name: equivalent_focal_length, unit: m, datatype: float64}
# - {name: effective_focal_length, unit: m, datatype: float64}
# meta: !!omap
# - {TAB_VER: '4.0'}
# - {ORIGIN: ctapipe.instrument.SubarrayDescription}
# - {SUBARRAY: MonteCarloArray}
# - {SOFT_VER: 0.30.1.dev123+gc1b3700eb}
# - {TAB_TYPE: optics}
# schema: astropy-2.0
optics_name size_type reflector_shape mirror_area n_mirrors n_mirror_tiles equivalent_focal_length effective_focal_length
ASTRI SST SCHWARZSCHILD_COUDER 14.126235008239746 2 2 2.1500000953674316 2.1519100666046143
```
The idea behind the change was that we could describe the individual mirrors in the table, while keeping the structural things as meta.

shall we update `sub.to_table(kind=optics)`?

maxnoe: mmh... this means you created a new table format, where the current optics table is a per-subarray table, this new one is per telescope. The change is fine, but it should be a different table kind, probably exposed as `OpticsDescription.{to,from}_table()`?

Maybe we call it `tel_optics` to differentiate and start at tab version 1.0?

maxnoe: I'll revert this change here, but I think this needs to be addressed before making the 0.31 release.

maxnoe: change reverted, the diff is very minimal now.

I'll open another one after this one to handle this new table format.

PR #3022, opened 2026-06-08
Fix service data file structure and symlinks

mexanick:
```
instrument tree
.
├── array-element-ids.json
├── array-elements
│   ├── 001
│   │   ├── 001.camgeom.fits.gz -> ../LST_LST_LSTCam/LST_LST_LSTCam.camgeom.fits.gz
│   │   ├── 001.camreadout.fits.gz -> ../LST_LST_LSTCam/LST_LST_LSTCam.camreadout.fits.gz
│   │   └── 001.optics.ecsv -> ../LST_LST_LSTCam/LST_LST_LSTCam.optics.ecsv
│   ...
│   ├── 005
│   │   ├── 005.camgeom.fits.gz -> ../MST_MST_FlashCam/MST_MST_FlashCam.camgeom.fits.gz
│   │   ├── 005.camreadout.fits.gz -> ../MST_MST_FlashCam/MST_MST_FlashCam.camreadout.fits.gz
│   │   └── 005.optics.ecsv -> ../MST_MST_FlashCam/MST_MST_FlashCam.optics.ecsv
│   ...
│   ├── 128
│   │   ├── 128.camgeom.fits.gz -> ../MST_MST_NectarCam/MST_MST_NectarCam.camgeom.fits.gz
│   │   ├── 128.camreadout.fits.gz -> ../MST_MST_NectarCam/MST_MST_NectarCam.camreadout.fits.gz
│   │   └── 128.optics.ecsv -> ../MST_MST_NectarCam/MST_MST_NectarCam.optics.ecsv
│   ...
│   ├── 131
│   │   ├── 131.camgeom.fits.gz -> ../SST_ASTRI_CHEC/SST_ASTRI_CHEC.camgeom.fits.gz
│   │   ├── 131.camreadout.fits.gz -> ../SST_ASTRI_CHEC/SST_ASTRI_CHEC.camreadout.fits.gz
│   │   └── 131.optics.ecsv -> ../SST_ASTRI_CHEC/SST_ASTRI_CHEC.optics.ecsv
│   ...
│   ├── LST_LST_LSTCam
│   │   ├── LST_LST_LSTCam.camgeom.fits.gz
│   │   ├── LST_LST_LSTCam.camreadout.fits.gz
│   │   └── LST_LST_LSTCam.optics.ecsv
│   ├── MST_MST_FlashCam
│   │   ├── MST_MST_FlashCam.camgeom.fits.gz
│   │   ├── MST_MST_FlashCam.camreadout.fits.gz
[...]

PR #3028, opened 2026-06-10
Use local dummy server to test behavior of nonexistent file

maxnoe:
Further mitigation of the issues we have with the minio server at DESY.

This avoids requesting a non-existent file by making the test with the local dummy server we already user for other tests, not the real test data server.


maxnoe:
Merged as part of #3022

PR #3033, opened 2026-06-12
bugfix HistoAgg with existing OutlierDetector

tjarkmiener:
make __call__() of ODs work with 3D and 4D

### DIFF
--- a/src/ctapipe/monitoring/aggregator.py
+++ b/src/ctapipe/monitoring/aggregator.py
@@ -586,8 +586,13 @@ class HistogramAggregator(BaseAggregator):
         hist.Hist
             Reconstructed histogram object with axes and counts populated.
         """
-
-        hist_container = ChunkHistogramContainer(**dict(zip(row.colnames, row)))
+        # Remove 'outlier_mask_detector_X' column names from the row
+        # to avoid issues when creating the ChunkHistogramContainer
+        row_dict = dict(zip(row.colnames, row)).copy()
+        for col_name in list(row_dict.keys()):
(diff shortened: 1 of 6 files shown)


tjarkmiener:
this should be included v0.31.0 @maxnoe . f you run the PixelStatsTool with HistogramAggregator and ODs in the configuration, the code is breaking in the OD __cal__ because we required a fix shape.

src/ctapipe/monitoring/aggregator.py
```diff
@@ -586,8 +586,13 @@ def hist_from_tablerow(
...
             Reconstructed histogram object with axes and counts populated.
         """
-
-        hist_container = ChunkHistogramContainer(**dict(zip(row.colnames, row)))
+        # Remove 'outlier_mask_detector_X' column names from the row
+        # to avoid issues when creating the ChunkHistogramContainer
+        row_dict = dict(zip(row.colnames, row)).copy()
+        for col_name in list(row_dict.keys()):
+            if col_name.startswith("outlier_mask_detector_"):
+                del row_dict[col_name]
```
maxnoe:
```suggestion
    row_dict = {
        col: row[col]
        for col in row.colnames
        if not col_name.startswith("outlier_mask_detector_")
    }
```

No need to create a new dict, then copy, as it is already a new dict. Also, just filtering the keys instead of removing them after.

maxnoe:
@TjarkMiener Sorry, the indentation was off in the comment. Need to fix locally and also run pre-commit

maxnoe (approved):

kosack (approved):

PR #3035, opened 2026-06-12
Add full cut opt

Hckjs:


Hckjs:
sorry, that was unintentional :sweat_smile:

PR #3044, opened 2026-06-22
Use newer dl2 file for model training tests

### LINKED ISSUE #3031: Update DL1 test dataset for machine learning tests
We should update the dl1 test dataset used for the machine learning unit / integration tests.

The current datasets are from ctapipe 0.17 and while we actively still support reading the format of that, we need to also test with the current format.

One example is pointing data, in ctapipe 0.17, we did not have a good way to store the fixed-per-run telescope pointing in simulations, which is needed for disp training for divergent pointing for example.

Related to #3014

maxnoe:
Fixes #3031

I'm not totally happy with how these tests are organized, this would very much benefit from a cleanup and defining fixtures for datasets depending on versions that we can then use to parametrize tests.

### DIFF
--- a/src/ctapipe/conftest.py
+++ b/src/ctapipe/conftest.py
@@ -186,6 +186,13 @@ def prod5_gamma_lapalma_simtel_path():
 
 
 @pytest.fixture(scope="session")
+def prod6_gamma_lapalma_simtel_path():
+    return get_dataset_path(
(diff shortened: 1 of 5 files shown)


maxnoe:
This also required me to generate a small simtel prod6 LaPalma file for the application tests, which is now also on the test data server.

kosack (approved):

PR #3053, opened 2026-07-02
Adding calibration dataformat constants

### LINKED ISSUE #3052: Add calibration-related HDF5 data format constants
There are a few HDF5 Data format Constants currently missing in ctapipe.io.hdf5dataformat, which are require by DPPS CalibPipe. These are:

- `"dl2/event/telescope/impact"`, which should be implemented as `DL2_TEL_IMPACT_GROUP`
- `"/dl1/monitoring/atmosphere"`, which should be implemented as `DL1_MONITORING_ATMOSPHERE`
- `"/dl1/monitoring/atmosphere/clear_night"`, which should be implemented as `DL1_MONITORING_CLEAR_NIGHT`

These data format constants are related to telescope intercalibration and atmospheric calibration.

juanpala1997:
This PR closes #3052 

It adds three new H5DF data format constants to ctapipe.io.hdf5dataformat

### DIFF
--- a/src/ctapipe/io/hdf5dataformat.py
+++ b/src/ctapipe/io/hdf5dataformat.py
@@ -46,6 +46,7 @@ __all__ = [
     "DL2_TEL_GEOMETRY_GROUP",
     "DL2_TEL_ENERGY_GROUP",
     "DL2_TEL_PARTICLETYPE_GROUP",
+    "DL2_TEL_IMPACT_GROUP",
     "DL0_TEL_POINTING_GROUP",
     "DL1_SUBARRAY_POINTING_GROUP",
     "DL1_TEL_POINTING_GROUP",
@@ -61,6 +62,8 @@ __all__ = [
     "DL1_SKY_PEDESTAL_IMAGE_GROUP",
     "DL1_FLATFIELD_IMAGE_GROUP",
     "DL1_FLATFIELD_PEAK_TIME_GROUP",
+    "DL1_MONITORING_ATMOSPHERE",
+    "DL1_MONITORING_CLEAR_NIGHT",
     "DL2_SUBARRAY_MONITORING_GROUP",
     "DL2_SUBARRAY_INTER_CALIBRATION_GROUP",
     "DL2_SUBARRAY_CROSS_CALIBRATION_GROUP",
@@ -102,6 +105,7 @@ DL2_TEL_GROUP = "/dl2/event/telescope"
 DL2_TEL_GEOMETRY_GROUP = "/dl2/event/telescope/geometry"
 DL2_TEL_ENERGY_GROUP = "/dl2/event/telescope/energy"
 DL2_TEL_PARTICLETYPE_GROUP = "/dl2/event/telescope/particle_type"
+DL2_TEL_IMPACT_GROUP = "/dl2/event/telescope/impact"
 
 DL2_GROUP = "/dl2"
 DL2_SUBARRAY_GROUP = "/dl2/event/subarray"
@@ -152,6 +156,8 @@ DL1_FLATFIELD_PEAK_TIME_GROUP = (
 DL1_PIXEL_HISTOGRAMS_GROUP = (
     "/dl1/monitoring/telescope/calibration/camera/pixel_histograms"
 )
+DL1_MONITORING_ATMOSPHERE = "/dl1/monitoring/atmosphere"
+DL1_MONITORING_CLEAR_NIGHT = "/dl1/monitoring/atmosphere/clear_night"
 DL2_SUBARRAY_MONITORING_GROUP = "/dl2/monitoring/subarray"
 DL2_SUBARRAY_INTER_CALIBRATION_GROUP = "/dl2/monitoring/subarray/inter_calibration"
 DL2_SUBARRAY_CROSS_CALIBRATION_GROUP = "/dl2/monitoring/subarray/cross_calibration"
--- a/src/ctapipe/io/hdf5dataformat.py
+++ b/src/ctapipe/io/hdf5dataformat.py
@@ -63,7 +63,6 @@ __all__ = [
     "DL1_FLATFIELD_IMAGE_GROUP",
     "DL1_FLATFIELD_PEAK_TIME_GROUP",
     "DL1_MONITORING_ATMOSPHERE",
-    "DL1_MONITORING_CLEAR_NIGHT",
     "DL2_SUBARRAY_MONITORING_GROUP",
     "DL2_SUBARRAY_INTER_CALIBRATION_GROUP",
     "DL2_SUBARRAY_CROSS_CALIBRATION_GROUP",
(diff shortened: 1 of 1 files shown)


src/ctapipe/io/hdf5dataformat.py
```diff
@@ -46,6 +46,7 @@
     "DL2_TEL_GEOMETRY_GROUP",
     "DL2_TEL_ENERGY_GROUP",
     "DL2_TEL_PARTICLETYPE_GROUP",
+    "DL2_TEL_IMPACT_GROUP",
```
mexanick, line 49: Why do you need it? Impact parameters between each shower and telescopes are calculated and stored, see [here](https://github.com/cta-observatory/ctapipe/blob/eb54d3d8875a0ab9012a2f0fd0bd208f57789a1a/src/ctapipe/reco/reconstructor.py#L207)

src/ctapipe/io/hdf5dataformat.py
```diff
@@ -61,6 +62,8 @@
     "DL1_SKY_PEDESTAL_IMAGE_GROUP",
     "DL1_FLATFIELD_IMAGE_GROUP",
     "DL1_FLATFIELD_PEAK_TIME_GROUP",
+    "DL1_MONITORING_ATMOSPHERE",
+    "DL1_MONITORING_CLEAR_NIGHT",
```
mexanick: What kind of data would be stored there? Is it format unique for clear nights?

src/ctapipe/io/hdf5dataformat.py
```diff
@@ -61,6 +62,8 @@
     "DL1_SKY_PEDESTAL_IMAGE_GROUP",
     "DL1_FLATFIELD_IMAGE_GROUP",
     "DL1_FLATFIELD_PEAK_TIME_GROUP",
+    "DL1_MONITORING_ATMOSPHERE",
```
mexanick, line 65: What's the plan of interfacing of this with [existing atmospheric module](https://ctapipe.readthedocs.io/en/latest/api-reference/atmosphere/index.html)?

src/ctapipe/io/hdf5dataformat.py
juanpala1997: We need it to access the value of the impact parameter; check [here](https://gitlab.cta-observatory.org/cta-computing/dpps/calibrationpipeline/calibpipe/-/blob/v0.5.0/src/calibpipe/tools/telescope_energy_asymmetry_calculator.py?ref_type=tags#L285)

maxnoe:
I don't understand why a group would be named `clear_night`.

I would expect an assessment of the atmosphere to be stored in some metrics inside of a dataset, not in a group name.

src/ctapipe/io/hdf5dataformat.py
Voutsi: No, not unique. I agree that the naming (clear night) should be changed. The data to be stored there will be the assessment of the night. In the table structure, a column will be the night assessment, i.e. clear, night, Calima, single Cirrus layer etc.

Voutsi:
> I don't understand why a group would be named `clear_night`.
> 
> I would expect an assessment of the atmosphere to be stored in some metrics inside of a dataset, not in a group name.

I see your point. As I wrote above, the naming I chose for the group is unfortunate. I would suggest to remove then the DL1_MONITORING_CLEAR_NIGHT group and write the assessment of the night inside a table under DL1_MONITORING_ATMOSPHERE

src/ctapipe/io/hdf5dataformat.py
Voutsi: I don't understand your comment. Do you mean the connection with ATMOSPHERE_DENSITY_PROFILE_TABLE?
Those are very different things. The point of adding the DL1_MONITORING_ATMOSPHERE group is to write down information monitoring the quality of the atmosphere. I think this is information is useful in order to monitor long term trends.

mexanick: I got an impression you were planning also to store the contemporary VAOD profiles, and I would naively assume that have to be eventually combined with the density profiles in order to compute the total attenuation of the Cherenkov light.

src/ctapipe/io/hdf5dataformat.py
mexanick: could you provide a provisional schema for the table?

src/ctapipe/io/hdf5dataformat.py
Voutsi: I see. Let me clarify. I do plan to store the VAOD. However the VAOD is not a profile, it is a single value expressing the optical depth at zenith from the ground to the top of the atmosphere. 

> I would naively assume that have to be eventually combined with the density profiles in order to compute the total attenuation of the Cherenkov light.

Indeed one has to combine the Rayleigh scattering profile (not density) with molecular absorption and aerosol profiles. But this is outside of the scope of this PR.

src/ctapipe/io/hdf5dataformat.py
Voutsi: Table schema:   
name,                                                  type,         units
night_assessment                               String          None       
VAOD                                                  float64      None       
fractional_rms_variability_amplitude  float64      None

mexanick:
In case a group (like `DL1_MONITORING_ATMOSPHERE`) will have multiple tables that aren't separated by the telescopes (or other array element ID), the table paths should be also prescribed in separate constants, like e.g. 
```
ATMOSPHERE_DENSITY_PROFILE_TABLE = "/simulation/service/atmosphere_density_profile"
```

src/ctapipe/io/hdf5dataformat.py
kosack: the atmosphere density profile is *Service* data, and this is *Monitoring*, so it think it's clear that they are different.

kosack (approved):

src/ctapipe/io/hdf5dataformat.py
Voutsi: Hi @kosack , I am wondering if the atmosphere density profile should be service data. CalibPipe will produce density profiles to be used for tailored simulations. That should be then monitoring data, right?

kosack: Monitoring data is something that changes in time during an observation, i.e. it is a table with a TIMESTAMP column, and values that are interpolated to the event time.

An atmosphere profile is at least static for a given day. or obs_id, I guess, right? You would not expect to have to interpolate it to an event time.  If so, it should still be Service data.   But if you plan to write out a time-series of these for a given night or something like that, then it could be Monitoring.  What is the plan?

kosack (changes requested):

src/ctapipe/io/hdf5dataformat.py
Voutsi: Hi @kosack , an atmospheric density model will be static for a day or OB (I am not sure if that's somewhere documented but I don't think it makes sense to produce density models in time intervals shorter than an OB, the density changes slowly). 

On the other hand, we might produce more than one extinction profile within one OB, when estimating good time intervals.

> You would not expect to have to interpolate it to an event time. If so, it should still be Service data.

That makes me think that the inter - cross calibration coefficients should go to service data as well...

PR #3054, opened 2026-07-09
fix(HillasIntersection): use geometry frame focal length for CameraFrame reconstruction

STSpencer:
image
## Summary

`HillasIntersection._predict` builds a `CameraFrame` for coordinate transforms using the telescope's `equivalent_focal_length` when processing `CameraHillasParametersContainer` input. However, the camera geometry — and therefore the Hillas parametrisation itself — is defined with the focal length chosen at read time (`SimTelEventSource.focal_length_choice`, which defaults to `FocalLengthKind.EFFECTIVE`). The effective and equivalent focal lengths differ by ~4–5% for typical CTA telescopes (e.g. LST: ~29.3 m effective vs 28 m equivalent).

Because the reconstructed angular position is proportional to `camera_position / focal_length`, using the wrong value introduces a plate-scale error. This error grows with the distance of the image from the camera centre. High-energy showers produce large, off-axis images far from the centre, so they are disproportionately affected — degrading angular resolution exactly at high energies.

The sibling reconstructor `HillasReconstructor` already handles this correctly by reading `camera.geometry.frame.focal_length`.

## Change

```python
# Before
focal_length = self.subarray.tel[tel_id].optics.equivalent_focal_length

# After
focal_length = self.subarray.tel[tel_id].camera.geometry.frame.focal_length
```

This reads back the focal length that was actually used when building the camera geometry frame, regardless of which `focal_length_choice` was configured — making `HillasIntersection` consistent with `HillasReconstructor` and with the parametrisation frame.

## Notes

- Only affects reconstruction from `CameraHillasParametersContainer` (i.e. `ImageProcessor.use_telescope_frame=False`). The default telescope-frame path is unaffected as it operates directly in angular coordinates.
- All 62 reco tests pass.
- `ruff format --check` and `ruff check` pass clean on the changed file.
- SonarQube analysis reports 0 security issues on the changed file; no new code smells are introduced.


kosack:
That is definitely a bug! I suspect that is one of the reasons why `HillasIntersection` never gave as good results as `HillasReconstructor`.  Maybe we should consider just dropping `HillasIntersection` altogether? Is it worth maintaining two versions of the same algorithms?

docs/changes/3054.bugfix.rst
```diff
@@ -0,0 +1,8 @@
+Fix ``HillasIntersection`` using the ``equivalent_focal_length`` instead of the
```
kosack, line 1: You have the wrong filename here: it should be `3054.bugfix.rst` (the number must be the PR, which is why the check is failing)

kosack (changes requested):

maxnoe (approved):

kosack (approved):
Looks ready to merge - the docs now fail due to a Sphinx problem resolving some of the Astropy classes. That is not due to this PR, so will need to fix it separately.

kosack (changes requested):
Actually, the doc problem is already fixed in main. Can you rebase this on main?

maxnoe (approved):

src/ctapipe/reco/hillas_intersection.py
```diff
@@ -217,7 +217,9 @@ def _predict(self, hillas_dict, array_pointing, telescopes_pointings=None):
 
         for tel_id, hillas in hillas_dict.items():
             if isinstance(hillas, CameraHillasParametersContainer):
-                focal_length = self.subarray.tel[tel_id].optics.equivalent_focal_length
+                focal_length = self.subarray.tel[
+                    tel_id
+                ].camera.geometry.frame.focal_length
```
maxnoe, line 222: this can be simplified a bit using the frame directly:

```
# add pointing information to frame
camera_frame = self.subarray.tel[tel_id].camera.geometry.frame
camera_frame = camera_frame.replicate_without_data(telescope_pointing=telescope_pointings[tel_id])
```

maxnoe:
Merging, since @kosack approved pending the docs build fix

PR #3056, opened 2026-07-17
Add PSF model based on Zernike coefficients

mexanick:
## Summary

Adds `ZernikePSFModel`, to `ctapipe.instrument.optics` alongside the existing `ComaPSFModel`.

Unlike `ComaPSFModel`'s parametric radial/polar scaling approach, this model reconstructs the optical wavefront directly from Zernike coefficients (using Noll indexing) and propagates it through Fraunhofer diffraction (FFT of the aberrated pupil).

## Implementation details

- Builds a circular pupil grid (`pupil_size`, `pupil_diameter_fraction`) with a soft-edged aperture (`pupil_edge_softness`) to avoid FFT ringing.
- Reconstructs the wavefront from per-telescope Zernike coefficients `z2`–`z11` (tilt, defocus, astigmatism, coma, trefoil, spherical), using the [`zernike`](https://pypi.org/project/zernike/) package's `RZern` for polynomial evaluation.
- Adds off-axis aberration growth: coma grows linearly with field angle θ (`z7_theta`, `z8_theta`), astigmatism grows quadratically with θ² (`z5_theta2`, `z6_theta2`).
- Computes the focal-plane intensity via FFT of the complex pupil function, averaged over `wavelength_samples` wavelengths between `wavelength_min` and `wavelength_max`, weighted by a `cherenkov_spectrum_index` power law (`dN/dλ ∝ λ^-index`).
- Applies Gaussian smoothing (`focal_plane_smoothing_sigma_pix`) to the resulting intensity to approximate pixel/light-guide response.
- Exposes the standard `pdf(tel_id, lon, lat, lon0, lat0)` interface shared with `ComaPSFModel`, interpolating the precomputed PSF grid via `scipy.ndimage.map_coordinates` and normalizing to unit integral.

image


kosack (changes requested):
Looks good, but a couple of small requests:

* Can you add a citation to the docs to where the "Zernike" coefficients are defined?
* Just a question: you make all the parameters generic names like "z2-z6", but they could have nicer names, e.g. z2 could be `tilt_x`, etc.  Maybe that is because of how they are defined in the reference paper? If so, then adding the citation is fine, but otherwise, why not use human-readable names?

mexanick:
> * Can you add a citation to the docs to where the "Zernike" coefficients are defined?

I've added a reference to Wikipedia. The original Zernike article is in German, and Noll's article is behind a paywall, but Wikipedia provides all the necessary information.

> * Just a question: you make all the parameters generic names like "z2-z6", but they could have nicer names, e.g. z2 could be `tilt_x`, etc.  Maybe that is because of how they are defined in the reference paper? If so, then adding the citation is fine, but otherwise, why not use human-readable names?

I would like to keep them as is (Zi), as this is the industry standard. For example, Ansys Zemax uses this notation, which makes it easy to copy or cross-check coefficient values directly from ray-tracing software. The docstrings for each parameter already contain the human-readable descriptions, so I think it's a good compromise between readability and standardization.

kosack (dismissed):

src/ctapipe/instrument/optics.py
```diff
@@ -627,3 +638,350 @@ def pdf(self, tel_id, lon, lat, lon0, lat0):
...
+        coma_y = coma_radial * uy + coma_tangential * ux
+        noll_coeffs = [
+            0.0,
+            self.z2.tel[tel_id].to_value(u.m),
+            self.z3.tel[tel_id].to_value(u.m),
+            self.z4.tel[tel_id].to_value(u.m),
+            self.z5.tel[tel_id].to_value(u.m) + z5_theta2_m_per_deg2 * theta2,
+            self.z6.tel[tel_id].to_value(u.m) + z6_theta2_m_per_deg2 * theta2,
+            self.z7.tel[tel_id].to_value(u.m) + coma_x,
+            self.z8.tel[tel_id].to_value(u.m) + coma_y,
```
Voutsi: According to wikipedia link, the Noll coefficient z7 is the vertical and z8 the horizontal. I think then you need to swap the comas, I mean add coma_x to z8 and coma_y to z7

mexanick: This was an artifact of some experimenting with the internal to the model coordinate transforms, I've simplified it.

src/ctapipe/instrument/optics.py
```diff
@@ -627,3 +638,344 @@ def pdf(self, tel_id, lon, lat, lon0, lat0):
...
+        weights /= np.sum(weights)
+        intensity = np.zeros_like(wavefront, dtype=float)
+
+        for wavelength, weight in zip(wavelengths, weights):
+            phase = 2 * np.pi * wavefront / wavelength
+            phase_safe = np.zeros_like(phase)
+            phase_safe[mask] = phase[mask]
+            pupil = aperture * np.exp(1j * phase_safe)
+            field = fft2(pupil)
+            intensity += weight * fftshift(np.abs(field) ** 2)
```
Voutsi: The angular coordinates depend on the wavelength. I think here we ignore this dependence, it should be accounted for before the final summation.

Voutsi:
Hi @mexanick , thanks a lot for this. I asked for one more thing to be clarified. Other than that looks good to me.

mexanick:
> Hi @mexanick , thanks a lot for this. I asked for one more thing to be clarified. Other than that looks good to me.

Thanks, it is fixed now. I also found and fixed an error in the astigmatism application. An updated PSF map:
image

src/ctapipe/instrument/optics.py
```diff
@@ -627,3 +638,344 @@ def pdf(self, tel_id, lon, lat, lon0, lat0):
...
+      model.
+
+    In the current parameterization, field-dependent aberrations are
+    represented by linear coma terms and quadratic astigmatism terms,
+    providing a compact phenomenological description of off-axis optical
+    degradation while retaining a physically motivated wavefront model.
+    """
+
+    # Universal model performance parameters
+    pupil_size = Int(
```
maxnoe: Is this ok as a global parameter? Maybe the required resolution could be telescope dependent?

src/ctapipe/instrument/optics.py
```diff
@@ -627,3 +638,344 @@ def pdf(self, tel_id, lon, lat, lon0, lat0):
...
+    ).tag(config=True)
+
+    wavelength_max = AstroQuantity(
+        default_value=550e-9 * u.m,
+        physical_type=u.physical.length,
+        help="Maximum wavelength for polychromatic averaging",
+    ).tag(config=True)
+
+    # Per-telescope optical parameters
+    psf_reference = TelescopeParameter(
```
maxnoe: Maybe better `psf_extent`?

src/ctapipe/instrument/optics.py
```diff
@@ -627,3 +638,344 @@ def pdf(self, tel_id, lon, lat, lon0, lat0):
...
+    # physical types, so `physical_type` is derived from the unit itself
+    # rather than a named constant like u.physical.length/angle.
+    coma_radial_growth = TelescopeParameter(
+        trait=AstroQuantity(physical_type=(u.m / u.deg).physical_type),
+        default_value=1.919e-07 * u.m / u.deg,
+        help="Radial coma growth",
+    ).tag(config=True)
+
+    z5_theta2 = TelescopeParameter(
+        trait=AstroQuantity(physical_type=(u.m / u.deg**2).physical_type),
```
maxnoe, line 810: why not just specify `unit=`?

pyproject.toml
```diff
@@ -44,6 +44,7 @@ dependencies = [
     "tables ~=3.4",
     "tqdm >=4.32",
     "traitlets ~=5.6",
+    "zernike",
```
maxnoe: Please only add `zernike` as an optional dependency here, it's a very specialized package for a very specific part of ctapipe.

maxnoe (changes requested):
Thanks, this looks very good. 

Aside from the comment by @Voutsi, I'd just request that we add the new package only as an optional dependency.

src/ctapipe/instrument/optics.py
mexanick: converted to telescope parameter alongside with a couple of others

src/ctapipe/instrument/optics.py
mexanick: because only `physical_type` is passed directly via `AstroQuantity`'s constructor, and the `unit` is forwarded to traitlets, causing deprecation warning and requirement to use via `*.tag(attr=value)` that I find even less readable than the above.

pyproject.toml
mexanick: good catch, I forgot about the deps splitting, should be fixed now.

docs/conf.py
```diff
@@ -154,9 +154,9 @@ def add_reference_type(prefix, objs):
         "DTypeLike",
         # astropy, new errors in 8.0, see https://github.com/astropy/astropy/issues/19933
         "Attribute",
-        "r.BaseDifferential",
-        "r.BaseRepresentation",
-        "r.BaseRepresentationOrDifferential",
+        "BaseDifferential",
```
maxnoe, line 157: With the rebase with main, there are now duplicate entries here

maxnoe (approved):

src/ctapipe/instrument/optics.py
```diff
@@ -627,3 +644,393 @@ def pdf(self, tel_id, lon, lat, lon0, lat0):
...
+            raise ValueError("wavelength_samples must be >= 1")
+        if lam_min <= 0 or lam_max <= 0 or lam_max < lam_min:
+            raise ValueError("Invalid wavelength_min/wavelength_max")
+
+        wavelengths = np.linspace(lam_min, lam_max, self.wavelength_samples)
+        weights = wavelengths ** (-self.cherenkov_spectrum_index)
+        weights /= np.sum(weights)
+
+        # Reference wavelength (mean of the range)
+        lambda_ref = np.mean(wavelengths)
```
Voutsi, line 935: Is it desirable that the reference wavelength (used later on scaling) depends on the configuration parameters min and max wavelenght?

mexanick: I don't think it is critical how the reference wavelength is selected as long as it is within the [min, max] that's driven by telescope's sensors.

Voutsi (approved):

PR #3060, opened 2026-07-30
fix(DispReconstructor): robust sign probability and valid-only cross-validation metrics

STSpencer:
## Summary

Fixes two bugs in the `DispReconstructor` that can affect reconstruction accuracy and the reported cross-validation performance.

### 1. Robust sign-probability lookup in `DispReconstructor._predict`

Previously the sign probability was read with a hard-coded column index:

```python
sign_proba = self._models[key][1].predict_proba(X)[:, 1]
```

This assumes the classifier always has exactly two classes ordered `[-1, +1]`. That assumption breaks in two ways:

- If a training fold / telescope type only ever saw a single sign, `predict_proba` returns a `(n, 1)` array and `[:, 1]` raises an `IndexError`, crashing training.
- If `np.sign(true_disp)` ever yields `0` (source projected exactly onto the COG), a third class `0` is introduced. With sorted `classes_ = [-1, 0, 1]`, `[:, 1]` then returns `P(class=0)` instead of `P(+1)`, silently flipping the reconstructed disp sign for every telescope.

The probability is now looked up by class *value* (`classes_ == 1`), with a safe fallback to zero probability when the positive class was never seen during training.

### 2. Cross-validation metrics no longer include invalid predictions

`_cross_validate_disp` discarded the validity mask returned by `_predict`:

```python
disp, sign_score, _ = models._predict(telescope_type, test)
r2 = r2_score(np.abs(truth), np.abs(disp))
accuracy = accuracy_score(np.sign(truth), np.sign(disp))
```

Rows with invalid features get `disp = NaN`. `np.sign(NaN) = NaN` becomes a spurious class in `accuracy_score` and NaNs corrupt `r2_score`, so the reported disp CV metrics could be silently wrong. The metrics are now computed only over the valid events.

## Tests

- `pytest src/ctapipe/reco/tests/test_sklearn.py src/ctapipe/reco/tests/test_disp.py` passes locally.

Opened as a draft for review.

### DIFF
--- a/docs/changes/3060.bugfix.rst
+++ b/docs/changes/3060.bugfix.rst
@@ -1,4 +1,4 @@
-Fix two bugs in the ``DispReconstructor``:
+Fix several bugs affecting the ``DispReconstructor``:
 
 - Look up the sign probability in ``DispReconstructor._predict`` by class value
   instead of a fixed column index, so it no longer crashes or returns the wrong
@@ -6,3 +6,9 @@ Fix two bugs in the ``DispReconstructor``:
   class) during training.
 - Compute the disp cross-validation metrics only over valid events, so invalid
   (``NaN``) predictions no longer corrupt the reported ``R^2`` and accuracy.
+- Treat non-finite (``inf``) feature values as invalid in ``check_valid_rows``,
+  so events with infinite features are dropped instead of being silently clipped
+  to a huge finite value and fed into the models.
+- Exclude events with an undefined disp sign (``norm == 0``) from training,
+  avoiding a spurious ``0`` class for the sign classifier and ``log(0) = -inf``
+  targets for the norm regressor when ``log_target`` is enabled.
--- a/docs/changes/3060.bugfix.rst
+++ b/docs/changes/3060.bugfix.rst
@@ -1,0 +1,8 @@
+Fix two bugs in the ``DispReconstructor``:
+
+- Look up the sign probability in ``DispReconstructor._predict`` by class value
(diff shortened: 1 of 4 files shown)


STSpencer:
Pushed two additional fixes uncovered while reviewing the monoscopic disp path:

- **Non-finite features:** `check_valid_rows` only masked `NaN`, so rows with `±inf` features passed validation and were then clipped to ~`3.4e38` by `table_to_float` and fed into both models. It now treats `inf` as invalid (consistent with `NaN`).
- **Undefined disp sign:** when `true_disp == 0`, `np.sign` yields `0`, which added a spurious third class to the sign classifier (also the root cause of the fixed `predict_proba[:, 1]` indexing) and produced `log(0) = -inf` targets when `log_target=True`. Such events are now excluded from training.

Note: an exact-zero disp can still, in principle, make `CrossValidator`'s `StratifiedKFold` split on `np.sign(target)` see a size-1 class. It's measure-zero in float data so I left it, but happy to collapse the `0` label in the stratification if reviewers prefer.

maxnoe:
@STSpencer the issues you mention here were also reported to me when I asked a coding model to review the disp code, however I concluded that these two edge cases (true_disp == 0.0 and evaluating the performance on invalid rows) cannot happen when running the tools on proper input data.

That these cases raise exceptions is good, the only thing to do would be to move these checks upwards and provide better error messages, e.g.:

"Input data contains events with `true_disp == 0.0`, this should never happen"

and

"Input data only contains a single disp sign, this should never happen"

PR #3062, opened 2026-08-03
Add containment-weighted-intensity stereo weighting

STSpencer:
## What

Adds a new `weights` option, `containment-weighted-intensity`, to `StereoMeanCombiner._calculate_weights` (shared by energy, geometry and particle-type combination).

The weight reproduces EventDisplay's geometric disp weight:

```
w = (intensity * (1 - width / length))**2 * (1 - leakage_intensity_width_2)**4
```

i.e. it combines image brightness, elongation, and a strong **containment** term `(1 - leakage)**4` that down-weights truncated (leaky) images.

## Why

High-energy angular resolution in the disp reconstruction is dominated by **image truncation (leakage)**: bright, leaky images have large along-axis `|disp|` scatter but are currently *over*-weighted (intensity correlates positively with leakage). None of the existing weighting schemes (`none`, `intensity`, `aspect-weighted-intensity`) account for containment.

In a benchmark on stereo test events (68% containment angular resolution, degrees), the containment weight was the single most robust option across all energies and matched or beat EventDisplay's disp-error weighting at the highest energies:

## Notes

- The new option is purely additive; the default (`none`) and existing options are unchanged.
- Uses only parameters already available at combination time (`hillas.intensity/width/length`, `leakage.intensity_width_2`).
- Complementary to PR #2731 (`StereoDispCombiner`), which improves head/tail **sign** handling but reuses the same three pre-existing weight options — this PR fills the containment-weighting gap. Because the weight lives in the shared `_calculate_weights`, #2731 would inherit it automatically.

## Tests

Extended `test_stereo_combination.py` to exercise the new scheme in both the table- and container-based paths.

Draft: opening for discussion / benchmarking on the full IRF pipeline.

### DIFF
--- a/src/ctapipe/reco/tests/test_stereo_combination.py
+++ b/src/ctapipe/reco/tests/test_stereo_combination.py
@@ -260,10 +260,23 @@ def test_mean_prediction_single_event(weights):
         assert u.isclose(event.dl2.stereo.energy["dummy"].energy, 30 * u.GeV)
         assert u.isclose(event.dl2.stereo.geometry["dummy"].alt, 60.9748605 * u.deg)
         assert u.isclose(event.dl2.stereo.geometry["dummy"].az, 316.0365515 * u.deg)
+    # For none/intensity/aspect the weighted mean of [1.0, 0.0, 0.8] coincidentally
+    # equals 0.6 because the test intensities (100, 200, 400) and equal aspect ratios
+    # cancel out.  containment-weighted-intensity breaks that coincidence because
+    # (1 - leakage)**4 differs per telescope (leakage = 0.0, 0.1, 0.2), so the
+    # containment weight is a general image-quality weight that affects all prediction
+    # types, not just geometry.
     if weights == "containment-weighted-intensity":
-        # weight = (intensity * (1 - width / length))**2 * (1 - leakage)**4
+        w = np.array(
+            [
+                (100 * (1 - 0.1 / 0.3)) ** 2 * (1 - 0.0) ** 4,  # tel 25
+                (200 * (1 - 0.1 / 0.3)) ** 2 * (1 - 0.1) ** 4,  # tel 125
+                (400 * (1 - 0.1 / 0.3)) ** 2 * (1 - 0.2) ** 4,  # tel 130
+            ]
+        )
+        expected = np.average([1.0, 0.0, 0.8], weights=w)
         assert event.dl2.stereo.particle_type["dummy"].prediction == pytest.approx(
-            0.6133700137551582
+            expected
         )
     else:
         assert event.dl2.stereo.particle_type["dummy"].prediction == pytest.approx(0.6)
--- a/src/ctapipe/reco/tests/test_stereo_combination.py
+++ b/src/ctapipe/reco/tests/test_stereo_combination.py
@@ -8,6 +8,7 @@ from ctapipe.containers import (
     ArrayEventContainer,
     HillasParametersContainer,
     ImageParametersContainer,
+    LeakageContainer,
     ParticleClassificationContainer,
     ReconstructedContainer,
     ReconstructedEnergyContainer,
@@ -31,6 +32,7 @@ def mono_table():
             "hillas_intensity": [1, 2, 0, 1, 5, 9],
             "hillas_width": [0.1, 0.2, 0.1, 0.1, 0.2, 0.1] * u.deg,
             "hillas_length": 3 * ([0.1, 0.2, 0.1, 0.1, 0.2, 0.1] * u.deg),
+            "leakage_intensity_width_2": [0.0, 0.1, 0.0, 0.2, 0.05, 0.3],
             "dummy_tel_energy": [1, 10, 4, 0.5, 0.7, 1] * u.TeV,
             "dummy_tel_is_valid": [
                 True,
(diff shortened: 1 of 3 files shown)


maxnoe:
Thanks a lot, this looks good.

Could you show the benchmark you ran?

src/ctapipe/reco/tests/test_stereo_combination.py
```diff
@@ -239,7 +260,13 @@ def test_mean_prediction_single_event(weights):
         assert u.isclose(event.dl2.stereo.energy["dummy"].energy, 30 * u.GeV)
         assert u.isclose(event.dl2.stereo.geometry["dummy"].alt, 60.9748605 * u.deg)
         assert u.isclose(event.dl2.stereo.geometry["dummy"].az, 316.0365515 * u.deg)
-    assert event.dl2.stereo.particle_type["dummy"].prediction == pytest.approx(0.6)
+    if weights == "containment-weighted-intensity":
```
maxnoe: I don't understand why the prediction for the particle type would depend on the disp weighting method.

STSpencer:
>Thanks a lot, this looks good.
Thanks!

> Could you show the benchmark you ran?

Using using the dataset gamma_diffuse_dl2_train_small.dl2.h5 split into training and test samples (as a small test) I get this for the angular resolution:

| Energy bin | `none` | `intensity` | `aspect-wtd` | **`containment-wtd`** | `exp(−kσ)` |
|---|---|---|---|---|---|
| 0.1–1 TeV | 0.829° | 0.669° | 0.727° | **0.690°** | 0.786° |
| 1–3 TeV | 1.140° | 0.628° | 0.655° | **0.485°** | 1.062° |
| 3–10 TeV | 1.631° | 0.770° | 0.768° | **0.558°** | 1.642° |
| 10–30 TeV | 2.540° | 1.609° | 1.393° | **0.994°** | 2.508° |
| 30–100 TeV | 4.339° | 2.993° | 2.671° | **2.328°** | 5.968° |

My/copilot's implementation of Gernot's `disp-error-exp(−kσ)` method is a bit dodgy as it really needs a separately trained, calibrated disp-error predictor (as EventDisplay provides via a BDT) to work well (rather than using extratrees variance as I did here). But `containment-weighted-intensity` still works well without needing an additional model.

src/ctapipe/reco/tests/test_stereo_combination.py
maxnoe: ah, sorry, I see this test is using the weight for all reconstructions, not just disp.

In this case, I think it would be better to parametrize the test also along the expected results, e.g.:



```
@pytest.mark.parametrize(
    ("weights", "expected_particle_prediction"),
    [
        ("aspect-weighted-intensity", 0.6)
        ("containment-weighted-intensity", 0.6133),
        ("intensity", 0.6),
        ("none", 0.6),
    ],
)
test_mean_prediction_single_event(weights, expected_particle_prediction)
```

it would also be better to chose dummy values that would actually be different between all tested weighting schemes, but that's not really related to your changes here, so I'd leave it for now

maxnoe:
> is a bit dodgy as it really needs a separately trained, calibrated disp-error predictor (as EventDisplay provides via a BDT) 

This is something we could add. We have something similar planned for event types, predicting the disp error is not really different from predicting the overall mono reconstruction error for the direction. Using the tree variance is not good, the trees are still highly correlated.

src/ctapipe/reco/tests/test_stereo_combination.py
STSpencer: Added in latest commit

STSpencer:
> This is something we could add. We have something similar planned for event types, predicting the disp error is not really different from predicting the overall mono reconstruction error for the direction. Using the tree variance is not good, the trees are still highly correlated.

I agree, but I think that should probably be a separate PR.

maxnoe:
> I agree, but I think that should probably be a separate PR.

Yes, definitely. There is an old draft PR already open:  https://github.com/cta-observatory/ctapipe/pull/2503

maxnoe:
To fix the docs build, please rebase with upstream main here.

LukasBeiske:
Should these weights also be available for HillasReconstructor? Currently the aspect-weighted-intensity weights are hard-coded there: https://github.com/cta-observatory/ctapipe/blob/8159403e8bb8ebb15a0a593fdb53446cde3adb83/src/ctapipe/reco/hillas_reconstructor.py#L282

STSpencer:
> Should these weights also be available for HillasReconstructor? Currently the aspect-weighted-intensity weights are hard-coded there:
> 
> https://github.com/cta-observatory/ctapipe/blob/8159403e8bb8ebb15a0a593fdb53446cde3adb83/src/ctapipe/reco/hillas_reconstructor.py#L282

Thanks for the idea! It's not implemented in eventdisplay as far as I can tell for it's versions of the HillasReconstructor, I'm not certain if containment weighting would make a difference or not in the simple case of a Hillas reconstruction, I'd have to try it out.

STSpencer:
I've tested it and the difference for the HillasReconstructor is marginal, I'd leave things the way they are for simplicity's sake.

maxnoe (approved):

LukasBeiske (approved):

Hckjs (approved):

LukasBeiske:
@maxnoe I think this can be merged now

maxnoe:
I was missing the sonarqube report and restarted the CI for to trigger that, this picked up a new issue though, fixed in #3073

maxnoe:
The sonar CI has stopped working for forks because github changes security settings and made them stricter.

I will merge here without sonar and try to fix it properly.

Issue #3055, opened 2026-07-10
Generalize Pointing and optics corrections so work with non-HDF5 inputs


kosack:
**Please describe the use case that requires this feature.**

Right now the only way to load a `PointingInterpolator` and have it apply  interpolated pointing  is in the `HDF55EventSource`, so it only works with ctapipe HDF5 files (so existing DL1 files for example).  That makes it _not usable for general data processing_, particularly for raw data in ZFITS format, where we want to process to DL1-parameters without requiring writing a separate DL1-image file. 

The current chain of processing is:

* `EventSource` (only if HDF5, use PointingInterpolator) → `CameraCalibrator` → `ImageProcessor` → `ShowerProcessor`

**Describe the solution you'd like**

Make the loading and application of interpolated pointing to an event  a step that happens *after* the `EventSource` and `CameraCalibrator`, so that it is independent of input file format.  It should do moret han just interpolating the pointing and filling the `PointingContainer`, but and actually applying it to the `TelescopeFrame` parameters in the `SubarrayDescription` per event and telescope.

For that reason, I think creating a new class `OpticsCalibrator` that runs after `CameraCalibrator` that does the pointing application(and any deformation corrections if available) and makes sure the coordinate `Frames` for each telescope are correct.

The chain in `ctapipe-process` would then:

* `EventSource` → `CameraCalibrator` → **`OpticsCalibrator`** → `ImageProcessor` → `ShowerProcessor`

That way we can apply pointing corrections and deformations to *any* input, not specifically HDF5 files. If a `EventSource` like `LSTEventSource` already applies corrections, it can just be disabled.

**Describe alternatives you've considered**

 The only fall-back solution right now is to run `ctapipe-process` twice:

* DL0 (ZFITS) →  DL1a (HDF5)
* DL1a (HDF5) + pointing → DL1b (HDF5)

That is inefficient, and also does not apply the corrections to the `CameraFrame` to `TelescopeFrame` (leaving each `Reconstructor` to have to do it itself, which is error prone)

**Additional context**

This is blocking the full implementation of processing observed data, e.g. from LST1.

maxnoe:
This has already happened, it's the `MonitoringSource` interface.

mexanick:
[...]

PR #3063, opened 2026-08-04
Add a few containers for monitoring data

mexanick:
Containers implementation for #3055

### DIFF
--- a/src/ctapipe/containers.py
+++ b/src/ctapipe/containers.py
@@ -1238,16 +1238,20 @@ class CameraDisplacementContainer(Container):
     """
 
     default_prefix = "camera_displacement"
-    delta_x = Field(
+    translation_x = Field(
         nan * u.m, "Camera displacement along X axis (horizontal)", unit=u.m
     )
-    delta_y = Field(nan * u.m, "Camera displacement along Y axis (vertical)", unit=u.m)
-    delta_z = Field(
+    translation_y = Field(
+        nan * u.m, "Camera displacement along Y axis (vertical)", unit=u.m
+    )
+    translation_z = Field(
         nan * u.m, "Camera displacement along optical Z axis (focal offset)", unit=u.m
     )
-    tilt_x = Field(nan * u.rad, "Camera tilt angle around X axis (pitch)", unit=u.rad)
-    tilt_y = Field(nan * u.rad, "Camera tilt angle around Y axis (yaw)", unit=u.rad)
-    tilt_z = Field(
+    rotation_x = Field(
+        nan * u.rad, "Camera tilt angle around X axis (pitch)", unit=u.rad
+    )
+    rotation_y = Field(nan * u.rad, "Camera tilt angle around Y axis (yaw)", unit=u.rad)
+    rotation_z = Field(
         nan * u.rad, "Camera rotation angle around optical Z axis (roll)", unit=u.rad
     )
 
--- a/src/ctapipe/containers.py
+++ b/src/ctapipe/containers.py
@@ -1468,26 +1468,13 @@ class TelescopeMonitoringContainer(Container):
         default_factory=CameraMonitoringContainer,
         description="Container for monitoring data for camera",
     )
-    structure_pointing = Field(
-        default=None,
-        type=TelescopeStructurePointingContainer,
-        description="Pointing of the telescope structure in local coordinates (RAW).",
-    )
-    structure_displacement = Field(
-        default=None,
-        type=TelescopeStructureDisplacementContainer,
-        description="Difference between ideal and real direction of the telescope axis in local coordinates.",
-    )
+
     camera_displacement = Field(
         default=None,
         type=CameraDisplacementContainer,
         description="Displacement and tilt of the camera w.r.t. its nominal position in the telescope structure.",
     )
-    aux_pointing_correction = Field(
-        default=None,
-        type=AuxiliaryPointingCorrectionContainer,
-        description="Pointing correction in celestial coordinates derived using auxiliary hardware.",
-    )
+
     pointing = Field(
         default=None,
         type=TelescopePointingContainer,
@@ -1508,11 +1495,6 @@ class MonitoringContainer(Container):
     pointing = Field(
         default_factory=ArrayPointingContainer,
         description="Array pointing positions",
-    )
-    weather = Field(
-        default=None,
-        type=WeatherMonitoringContainer,
-        description="Weather monitoring data",
     )
 
 
--- a/src/ctapipe/containers.py
+++ b/src/ctapipe/containers.py
@@ -1279,10 +1279,10 @@ class TelescopePointingCorrectionContainer(Container):
     delta_azimuth = Field(nan * u.rad, "Residual Azimuth offset correction", unit=u.rad)
 
     # Optional uncertainty tracking
-    sigma_altitude = Field(
+    uncertainty_altitude = Field(
         nan * u.rad, "Uncertainty on Altitude offset correction", unit=u.rad
     )
-    sigma_azimuth = Field(
+    uncertainty_azimuth = Field(
         nan * u.rad, "Uncertainty on Azimuth offset correction", unit=u.rad
     )
 
--- a/src/ctapipe/containers.py
+++ b/src/ctapipe/containers.py
@@ -62,6 +62,7 @@ __all__ = [
     "TelescopeStructurePointingContainer",
     "TelescopeStructureDisplacementContainer",
     "CameraDisplacementContainer",
+    "AuxiliaryPointingCorrectionContainer",
     "TelescopePointingCorrectionContainer",
     "ArrayPointingContainer",
     "StatisticsContainer",
--- a/src/ctapipe/containers.py
+++ b/src/ctapipe/containers.py
@@ -1279,10 +1279,10 @@ class TelescopePointingCorrectionContainer(Container):
     delta_azimuth = Field(nan * u.rad, "Residual Azimuth offset correction", unit=u.rad)
 
     # Optional uncertainty tracking
-    uncertainty_altitude = Field(
+    altitude_uncertainty = Field(
         nan * u.rad, "Uncertainty on Altitude offset correction", unit=u.rad
     )
-    uncertainty_azimuth = Field(
+    azimuth_uncertainty = Field(
         nan * u.rad, "Uncertainty on Azimuth offset correction", unit=u.rad
     )
 
--- a/src/ctapipe/containers.py
+++ b/src/ctapipe/containers.py
@@ -1265,20 +1265,26 @@ class AuxiliaryPointingCorrectionContainer(Container):
 
 class TelescopePointingCorrectionContainer(Container):
     """
-    Residual sky-pointing corrections in celestial coordinates (RA/Dec/Rotation).
+    Residual sky-pointing corrections in local coordinates (Alt/Az).
 
-    This container holds abstract celestial offsets regardless of how they were
+    This container holds pointing offsets in local coordinates regardless of how they were
     derived (e.g. analytical StarTracker, auxiliary hardware devices,
     or a combination/fusion model).
     """
 
     default_prefix = "pointing_correction"
-    delta_alt = Field(nan * u.rad, "Residual Altitude offset correction", unit=u.rad)
-    delta_az = Field(nan * u.rad, "Residual Azimuth offset correction", unit=u.rad)
+    delta_altitude = Field(
+        nan * u.rad, "Residual Altitude offset correction", unit=u.rad
+    )
+    delta_azimuth = Field(nan * u.rad, "Residual Azimuth offset correction", unit=u.rad)
 
     # Optional uncertainty tracking
-    sigma_alt = Field(nan * u.rad, "Uncertainty on Alt offset correction", unit=u.rad)
-    sigma_az = Field(nan * u.rad, "Uncertainty on Az offset correction", unit=u.rad)
+    sigma_altitude = Field(
+        nan * u.rad, "Uncertainty on Altitude offset correction", unit=u.rad
+    )
+    sigma_azimuth = Field(
+        nan * u.rad, "Uncertainty on Azimuth offset correction", unit=u.rad
+    )
 
 
 class TelescopePointingContainer(Container):
--- a/src/ctapipe/containers.py
+++ b/src/ctapipe/containers.py
@@ -1324,7 +1324,8 @@ class WeatherMonitoringContainer(Container):
         nan * u.hPa, "Atmospheric pressure at observatory site", unit=u.hPa
     )
     relative_humidity = Field(
-        nan, "Relative humidity fraction (0.0 to 1.0)", unit=u.dimensionless_unscaled
+        nan,
+        "Relative humidity fraction (0.0 to 1.0)",
     )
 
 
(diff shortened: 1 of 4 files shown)


src/ctapipe/containers.py
```diff
@@ -1187,8 +1193,108 @@ class DL2Container(Container):
...
+
+
+class TelescopeSkyPointingContainer(Container):
+    """
+    Fully corrected pointing mapping the camera center to
+    celestial coordinates (ICRS).
+    """
+
+    default_prefix = "sky_pointing"
+    ra = Field(nan * u.rad, "Right Ascension of camera center (ICRS)", unit=u.rad)
```
maxnoe: Why do we express the corrected pointing in ICRS and not AltAz?

The frame transformations assume the pointing is in AltAz and the reconstruction also happens in AltAz.

maxnoe: Also, this container is not added to the event, which keeps the alt az variant.

mexanick: I would consider exposing both, and hide the coord transform. In order to do the transform correctly, we need to modify the telescope geometry at per-event basis to accodomate cdm data, therefore I would provide end users with correctly calculated pointings in both coordinate systems and hide the conversion.

maxnoe: Shouldn't we convert the cdm correction to AltAz and just provide a correct alt/az pointing?

maxnoe: I think we have to, because this is a per telescope correct that needs to be applied before the stereo reconstruction.

mexanick: Perhaps that's the way to go, However, in this PR I only want to add the monitoring containers (we also need all this data for monitoring purposes). I also didn't compute any kind of "telescope pointing", whether it is in AltAz or RA/Dec, I'm just gathering the input ingredients at the moment.

maxnoe: That's clear, the reason was to ask why/if the container with ra/dec is needed.

maxnoe:
I'm not sure that having the containers with nan defaults in the event structure is the right choice in this case. Maybe better use `default=None, type=TelescopeStructurePointingContainer` for the fields in the event?

Like this, it's clear that the monitoring was not filled at all, not that we tried to fill it and didn't succeed.

src/ctapipe/containers.py
mexanick: Ah, perhaps I should remove it right now, as it is anyhow not used yet. Eventually we may need to apply the final corrections from auxiliary devices like StarGuider, that provide the corrections in Ra/Dec, but perhaps we should rather transform to local coordinate frame...

mexanick: SkyPointing is removed, final corrections would be foreseen in Alt/Az. We can change this in future if needed.

src/ctapipe/containers.py
```diff
@@ -1187,8 +1192,87 @@ class DL2Container(Container):
...
+
+# Pointing-related containers
+class TelescopeStructurePointingContainer(Container):
+    """
+    Pointing of the telescope structure in local coordinates (Alt/Az)
+    as measured by the telescope hardware (e.g. drive encoders).
+    """
+
+    default_prefix = "structure_pointing"
+    azimuth = Field(None, "Raw azimuth measured by structure encoders", unit=u.rad)
```
maxnoe: I think you misunderstood my comment. I wasn't asking to change the field default of the containers themselves, here I think nan deg is ok. I meant the field for the whole container added to the MonitoringContainer.

src/ctapipe/containers.py
```diff
@@ -1347,6 +1450,22 @@ class TelescopeMonitoringContainer(Container):
         default_factory=CameraMonitoringContainer,
         description="Container for monitoring data for camera",
     )
+    structure_pointing = Field(
+        default_factory=TelescopeStructurePointingContainer,
```
maxnoe: I meant changing here `default=None, type=TelescopeStructurePointingContainer` instead of changing the actual value fields.

mexanick: I was following the doc, prescribing to use `default` only to immutable fields and `default_factory` for mutable ones. I wouldn't call `Container` immutable... Its structure is immutable, but the payload is not.

src/ctapipe/containers.py
mexanick: Yeah, I see your point. Please see my answer to your related comment below.

src/ctapipe/containers.py
maxnoe: `None` is immutable. It applies to the default value. The question is whether the value that is assigned to `default` can be safely used as is (assigned to new containers) or not (call a factory to produce a fresh value for a container).

maxnoe: > This should only be used for immutable values. For mutable values, use default_factory instead.

This refers to the value passed to `default`.

maxnoe (dismissed):

src/ctapipe/containers.py
```diff
@@ -1187,8 +1192,95 @@ class DL2Container(Container):
...
+    default_prefix = "aux_pointing_correction"
+    delta_ra = Field(
+        nan * u.rad, "Residual Right Ascension offset correction", unit=u.rad
+    )
+    delta_dec = Field(nan * u.rad, "Residual Declination offset correction", unit=u.rad)
+
+
+class TelescopePointingCorrectionContainer(Container):
+    """
+    Residual sky-pointing corrections in celestial coordinates (RA/Dec/Rotation).
```
Voutsi: Please update the docstring in order to reflect what the container actually holds (alt/az instead of RA/Dec/Rotation)

src/ctapipe/containers.py
```diff
@@ -1187,8 +1192,95 @@ class DL2Container(Container):
...
+
+    This container holds abstract celestial offsets regardless of how they were
+    derived (e.g. analytical StarTracker, auxiliary hardware devices,
+    or a combination/fusion model).
+    """
+
+    default_prefix = "pointing_correction"
+    delta_alt = Field(nan * u.rad, "Residual Altitude offset correction", unit=u.rad)
```
Voutsi: In TelescopeStructureDisplacementContainer this parameter is delta_altitude. I suggest to keep a consistent naming, i.e. renaming the parameters to delta_altitude and delta_azimuth. Similarly I suggest to rename the uncertainties to alt_uncert/az_uncert

src/ctapipe/containers.py
```diff
@@ -1214,6 +1306,29 @@ class ArrayPointingContainer(Container):
...
+    default_prefix = "weather"
+    temperature = Field(
+        nan * u.K, "Ambient air temperature at observatory site", unit=u.K
+    )
+    pressure = Field(
+        nan * u.hPa, "Atmospheric pressure at observatory site", unit=u.hPa
+    )
+    relative_humidity = Field(
+        nan,
+        "Relative humidity fraction (0.0 to 1.0)",
```
Voutsi, line 1339: What if the weather station reports the relative humidity in percentage? It is not uncommon.

src/ctapipe/containers.py
```diff
@@ -1214,6 +1306,29 @@ class ArrayPointingContainer(Container):
...
+    Environmental parameters at the observatory site.
+
+    Note: Astronomical refraction derived from weather measurements SHALL ONLY
+    be used when fitting star positions during offline optical pointing calibration.
+    It SHALL NOT be applied during Cherenkov atmospheric shower direction reconstruction.
+    """
+
+    default_prefix = "weather"
+    temperature = Field(
+        nan * u.K, "Ambient air temperature at observatory site", unit=u.K
```
Voutsi, line 1332: What if the weather station reports the temperature in Celsius?

src/ctapipe/containers.py
maxnoe: > renaming the parameters to delta_altitude and delta_azimuth. Similarly I suggest to rename the uncertainties to alt_uncert/az_uncert

Please stay consistent whether you abbreviate alt/az or use the full term altitude/azimuth, at least within the same container.

src/ctapipe/containers.py
mexanick: It will be converted in the monitoring source. The containers here represent the data structures used internally in ctapipe.

src/ctapipe/containers.py
mexanick: same as above

src/ctapipe/containers.py
mexanick: good point, the field names and docstrings have been updated.

src/ctapipe/containers.py
```diff
@@ -1187,8 +1192,101 @@ class DL2Container(Container):
...
+    """
+
+    default_prefix = "pointing_correction"
+    delta_altitude = Field(
+        nan * u.rad, "Residual Altitude offset correction", unit=u.rad
+    )
+    delta_azimuth = Field(nan * u.rad, "Residual Azimuth offset correction", unit=u.rad)
+
+    # Optional uncertainty tracking
+    sigma_altitude = Field(
```
maxnoe: We don't use `sigma_` for uncertainties, all other containers use `<field>_uncertainty` or `<field>_uncert`. For new containers, I'd take the long form.

src/ctapipe/containers.py
```diff
@@ -1187,8 +1192,101 @@ class DL2Container(Container):
...
+    """
+
+    default_prefix = "pointing_correction"
+    delta_altitude = Field(
+        nan * u.rad, "Residual Altitude offset correction", unit=u.rad
+    )
+    delta_azimuth = Field(nan * u.rad, "Residual Azimuth offset correction", unit=u.rad)
+
+    # Optional uncertainty tracking
+    uncertainty_altitude = Field(
```
maxnoe:
```suggestion
    altitude_uncertainty = Field(
```

src/ctapipe/containers.py
```diff
@@ -1187,8 +1192,101 @@ class DL2Container(Container):
...
+        nan * u.rad, "Uncertainty on Altitude offset correction", unit=u.rad
+    )
+    uncertainty_azimuth = Field(
```
maxnoe:
```suggestion
    azimuth_uncertainty = Field(
```

src/ctapipe/containers.py
```diff
@@ -1347,8 +1468,29 @@ class TelescopeMonitoringContainer(Container):
         default_factory=CameraMonitoringContainer,
         description="Container for monitoring data for camera",
     )
+    structure_pointing = Field(
```
maxnoe: Following our discussion on slack on when / how lower-level monitoring is combined, I don't really see the need to add these containers to the event structure.

Is there a use case remaining?

I think we agreed that the more efficient approach is to compute the final correction on the monitoring time scales based on tabular data and interpolate that to the event.

In that scenario, we would not need event-wise data structures for these lower-level parts.

maxnoe (changes requested):

src/ctapipe/containers.py
mexanick: I think we still want them for the star tracking. However, in this case I should leave them as standalone containers, but remove from being linked to the event, so it is only filled on request and not for every event processed with ctapipe-process tool, right?

In this case, I will leave TelescopePointing and CameraDisplacement containers here (I believe there's no easy way around for Camera Displacement, as I believe it would be only used to modify image, so we just take it as is and interpolate to the event time).

maxnoe: You mean as output containers from the star tracking which would work on bunches of events? Or as intermediate input per event?

mexanick: as intermediate event-wise input to star tracker. The output of star tracker would be a grid of corrections (common grid with the structure monitoring).

maxnoe: Why would the star-tracker not use the pointing filled from Structure + BendingModel into the telescope poiinting container?

mexanick: It will use the pointing from Structure + BendingModel as the seed, but it will also use CDM for a given event and weather values for the given event.

maxnoe: Sure, weather is needed and should be added here.

But for the pointing, I would think that you configure the DL0MonitoringSource to read Structure, BendingModel and CDM files and use that to fill the already existing pointing container.

mexanick: Yes, it will be done like that. However, I still need to define the TelescopeStructurePointingContainer, TelescopeStructureDisplacementContainer, etc for conversion, to abstract telescope-specific implementations. They, however, would not be added to the event monitoring structure.

maxnoe: I don't understand for what you need the containers then

src/ctapipe/containers.py
```diff
@@ -1187,8 +1193,101 @@ class DL2Container(Container):
...
+        nan * u.rad,
+        "Altitude offset due to structure bending/deformation",
+        unit=u.rad,
+    )
+
+
+class CameraDisplacementContainer(Container):
+    """
+    Displacement and tilt of the camera w.r.t. its nominal position in the
+    telescope structure.
```
kosack, line 1237: Somewhere in the docs you should be specific about the order of applying these transforms: _rotate then translate_ is not the same as _translate then rotate_. 

 Also maybe name them like that? E.g. `translation_x`, `rotation_z` to be more standard? (tilt could mean a rotation or a shear, so it's not as clear)

kosack: Which brings up one more question: do we need to worry about camera or optics shear? Or is it negligible?

src/ctapipe/containers.py
```diff
@@ -1347,8 +1469,16 @@ class TelescopeMonitoringContainer(Container):
         default_factory=CameraMonitoringContainer,
         description="Container for monitoring data for camera",
     )
+
+    camera_displacement = Field(
```
kosack, line 1477: How are dual-mirror telescopes supported by this? They need camera, and primary mirror corrections, I guess?

mexanick: As far as I know, the SST-2M hardware doesn't foresee the camera displacement monitoring hardware. They have Pointing Monitoring Camera (PMC) that's mounted at the tip of the structure and used for online pointing if I understand correctly. Then, the offline corrections are done with the star tracking method using the camera monitoring data (pixel voltages SC readout).

src/ctapipe/containers.py
mexanick: I don't think we could have any sizeable camera shear... For the optics, this is possible since we have faceted design, but I think this would just affect the PSF and as soon as that is under control, we should not worry about it.

src/ctapipe/containers.py
mexanick: I will need to convert the raw monitoring tables to the schema defined by these containers and perhaps they would be needed and filled at event-basis for calibration (pedestal) events for star tracking algo. So the main use: define the centralized schema of pointing monitoring data.

src/ctapipe/containers.py
```diff
@@ -1214,6 +1317,29 @@ class ArrayPointingContainer(Container):
     array_dec = Field(nan * u.rad, "Array pointing declination", unit=u.rad)
 
 
+# Environmental Monitoring containers
+class WeatherMonitoringContainer(Container):
```
maxnoe, line 1321: Why is this container not added to the event?

mexanick: What would be the reason to add? For the pointing determination, it is only need for star tracking, that only analyzes pedestal events, and to convert auxiliary camera corrections (e.g. StarGuider), but this will be done prior to physics events processing and the correction would be already in local AltAz, interpolated to the event time.

PR #3066, opened 2026-08-05
Add simple pointing calibrator

mexanick:
Adding a simple pointing calibrator that combines raw structure pointing with structure displacement correction.
Partially addresses #3055

### DIFF
--- a/docs/changes/3063.datamodel.rst
+++ b/docs/changes/3063.datamodel.rst
@@ -1,0 +1,1 @@
+Add monitoring containers for pointing-related data
--- a/docs/changes/3066.feature.rst
+++ b/docs/changes/3066.feature.rst
@@ -1,0 +1,1 @@
+Add basic pointing calibrator that combines the raw structure pointing with structure displacement corrections.
--- a/src/ctapipe/calib/__init__.py
+++ b/src/ctapipe/calib/__init__.py
@@ -4,8 +4,10 @@ Module for calibration code
 """
 
 from .camera import CameraCalibrator, GainSelector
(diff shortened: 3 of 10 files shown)


kosack:
Looks like a good start, but I'm trying to understand how this will be used and evolve in the future: the output I would expect for a pointing calibrator is a  `TelescopeFrame` instance, or at least all the parameters I need to construct a `TelescopeFrame` instance.  Right now, this just interpolates and applies alt/az corrections to generate the corrected `TelescopePointingContainer` for the event, which is a first step.   In principle, we don't even need to apply the corrections to the pointing container, but rather set the`TelescopeFrame` for each telescope so that the  ImageProcessor correctly transforms it when computing parameters.  

How would you forsee in the future including other deformations?  Would we add fields to the `TelescopePointingContainer`, and then have a helper function `.to_frame()`?    Or would we just add the corrections to the Frame, and keep the pointing as is?

maxnoe:
> pointing calibrator is a TelescopeFrame

A telescope frame is a local field of view frame, that's not what I would expect for the pointing calibrator to output.

This would be missing the corrections needed to go from pixels to field of view (camera rotation / displacement), right?

So I would expect either a fully specified `CameraFrame`, so that pixel coordinates can be transformed to sky coordinates (AltAz, Nominal, ICRS, ...) or the necessary frame attributes to construct that.

kosack:
Yes, that was a mistake: it should be CameraFrame, not TelescopeFrame (we want to convert from CameraFrame → TelescopeFrame via the pointing transformations)

PR #3068, opened 2026-08-10
Muon charge extractor or waveform-based cleaning algorithm

burmist-git:
The LST hardware stereo trigger preferentially selects muons whose parent air shower, or part of it, falls within the same camera field of view (these events are classified as stereo muon events). This is because the density of the muon light pool decreases with the square of the distance from the muon's impact point, consequently, an isolated muon (i.e., a muon significantly deflected from its parent shower) cannot, in general, trigger two telescopes simultaneously. Therefore, a waveform-based cleaning algorithm needs to be used for LST muon reconstruction. The cleaning is applied at the waveform level, as the Cherenkov light from the muon arrives within a very short time window (about hundreds of picoseconds), slightly ahead of time from the rest of the shower light, and geometrically concentrated in a ring.

### DIFF
--- a/src/ctapipe/conftest.py
+++ b/src/ctapipe/conftest.py
@@ -213,8 +213,11 @@ def calibpipe_camcalib_obslike_different_chunks():
 
 
 @pytest.fixture(scope="session")
-def ctapipe_ref_pules()
-    return get_dataset_path("calibpipe_camcalib_single_chunk_i0.1.0.dl1.h5")
+def ctapipe_ref_pules():
+    return read_table(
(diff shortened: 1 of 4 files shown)


burmist-git:
[099_muon_wf_cleaning.pdf](https://github.com/user-attachments/files/30897295/099_muon_wf_cleaning.pdf)

https://indico.cern.ch/event/1713776/contributions/7208820/attachments/3320971/5946068/sipm_cam_all_clean_1k.pdf

burmist-git:
[short_cam_all_clean.pdf](https://github.com/user-attachments/files/30903612/short_cam_all_clean.pdf)

PR #3073, opened 2026-08-13
Ignore numpy deprecation warning from joblib

maxnoe:
Yesterday, numba 0.67 was release with support for numpy 2.5, which is why we are getting it now in CI.

This triggered a DeprecationWarning in joblib (fixed upstream but not yet released) failing CI.


LukasBeiske (approved):

STSpencer (approved):
Thanks @maxnoe, this looks good to me

PR #3076, opened 2026-08-14
StereoDispCombiner: Add Transformation to Nominalframe so divergent pointing is supported

AkramAki:
This introduces support for divergent pointing for the StereoDispCombiner. Since the StereoDispCombiner PR is not finished yet this PR targets the corresponding branch instead of main.

### DIFF
--- a/src/ctapipe/reco/stereo_combination.py
+++ b/src/ctapipe/reco/stereo_combination.py
@@ -878,16 +878,17 @@ class StereoDispCombiner(StereoCombiner):
         valid = mono_predictions[f"{prefix_tel}_is_valid"].copy()
         self._require_disp_column(mono_predictions, prefix_tel)
 
-        fov_lon_values, fov_lat_values = calc_fov_lon_lat(
-            mono_predictions, prefix_tel
-        )
+        fov_lon_values, fov_lat_values = calc_fov_lon_lat(mono_predictions, prefix_tel)
         fov_lon_values, fov_lat_values = self._transform_to_nominal(
             mono_predictions, fov_lon_values, fov_lat_values
         )
-        hillas_psis = np.arctan2(
-            np.diff(fov_lat_values, axis=1),
-            np.diff(fov_lon_values, axis=1),
-        )[:, 0] * u.rad
+        hillas_psis = (
+            np.arctan2(
+                np.diff(fov_lat_values, axis=1),
+                np.diff(fov_lon_values, axis=1),
+            )[:, 0]
+            * u.rad
+        )
 
         obs_ids, event_ids, _, tel_to_array_indices = get_subarray_index(
             mono_predictions
--- a/src/ctapipe/reco/stereo_combination.py
+++ b/src/ctapipe/reco/stereo_combination.py
@@ -2,11 +2,17 @@ from abc import abstractmethod
 
 import astropy.units as u
 import numpy as np
-from astropy.coordinates import AltAz, CartesianRepresentation, SphericalRepresentation
+from astropy.coordinates import (
+    AltAz,
+    CartesianRepresentation,
+    SkyCoord,
+    SphericalRepresentation,
+)
 from astropy.table import Table
 from traitlets import UseEnum
 
 from ctapipe.containers import ImageParametersContainer
+from ctapipe.coordinates import NominalFrame, TelescopeFrame
 from ctapipe.core import Component, Container
 from ctapipe.core.traits import (
     Bool,
@@ -756,6 +762,13 @@ class StereoDispCombiner(StereoCombiner):
         weights = []
 
         signs = np.array([-1, 1])
+        # Use the array pointing as the common frame for all telescopes.
+        array_pointing = SkyCoord(
+            alt=event.monitoring.pointing.array_altitude,
+            az=event.monitoring.pointing.array_azimuth,
+            frame=AltAz(),
+        )
+        nominal_frame = NominalFrame(origin=array_pointing)
 
         for tel_id, dl2 in event.dl2.tel.items():
             if not dl2.geometry[self.prefix].is_valid:
@@ -781,6 +794,21 @@ class StereoDispCombiner(StereoCombiner):
 
             fov_lons = hillas_fov_lon + signs * disp * np.cos(hillas_psi)
             fov_lats = hillas_fov_lat + signs * disp * np.sin(hillas_psi)
+
+            # Transform both DISP candidates and their axis to the common frame.
+            telescope_pointing = SkyCoord(
+                alt=event.monitoring.tel[tel_id].pointing.altitude,
+                az=event.monitoring.tel[tel_id].pointing.azimuth,
+                frame=AltAz(),
+            )
+            candidates = SkyCoord(
+                fov_lon=fov_lons * u.deg,
+                fov_lat=fov_lats * u.deg,
+                frame=TelescopeFrame(telescope_pointing=telescope_pointing),
+            ).transform_to(nominal_frame)
+            fov_lons = candidates.fov_lon.to_value(u.deg)
+            fov_lats = candidates.fov_lat.to_value(u.deg)
+            hillas_psi = np.arctan2(np.diff(fov_lats), np.diff(fov_lons))[0] * u.rad
 
             fov_lon_values.append(fov_lons)
             fov_lat_values.append(fov_lats)
@@ -850,15 +878,26 @@ class StereoDispCombiner(StereoCombiner):
         valid = mono_predictions[f"{prefix_tel}_is_valid"].copy()
         self._require_disp_column(mono_predictions, prefix_tel)
 
+        fov_lon_values, fov_lat_values = calc_fov_lon_lat(
+            mono_predictions, prefix_tel
+        )
+        fov_lon_values, fov_lat_values = self._transform_to_nominal(
+            mono_predictions, fov_lon_values, fov_lat_values
+        )
+        hillas_psis = np.arctan2(
+            np.diff(fov_lat_values, axis=1),
+            np.diff(fov_lon_values, axis=1),
+        )[:, 0] * u.rad
+
         obs_ids, event_ids, _, tel_to_array_indices = get_subarray_index(
             mono_predictions
         )
         n_array_events = len(obs_ids)
 
         valid = self._apply_min_ang_diff_cut(
-            mono_predictions=mono_predictions,
             valid=valid,
             tel_to_array_indices=tel_to_array_indices,
+            hillas_psis=hillas_psis,
         )
         valid = self._apply_n_best_tels_cut(
             mono_predictions=mono_predictions,
@@ -873,6 +912,8 @@ class StereoDispCombiner(StereoCombiner):
             tel_to_array_indices=tel_to_array_indices,
             n_array_events=n_array_events,
             prefix_tel=prefix_tel,
+            fov_lon_values=fov_lon_values[valid],
+            fov_lat_values=fov_lat_values[valid],
         )
 
         stereo_table[f"{self.prefix}_alt"] = alt
@@ -907,9 +948,9 @@ class StereoDispCombiner(StereoCombiner):
 
     def _apply_min_ang_diff_cut(
         self,
-        mono_predictions: Table,
         valid: np.ndarray,
         tel_to_array_indices: np.ndarray,
+        hillas_psis: u.Quantity,
     ) -> np.ndarray:
         if self.min_ang_diff is None:
             return valid
@@ -921,7 +962,7 @@ class StereoDispCombiner(StereoCombiner):
 
         valid_idx = np.flatnonzero(valid)
         pairs_in_valid = np.flatnonzero(mask_multi2_tels).reshape(-1, 2)
-        valid_psis = mono_predictions["hillas_psi"][valid]
+        valid_psis = hillas_psis[valid]
 
         keep_pairs = check_ang_diff(
             self.min_ang_diff,
@@ -984,6 +1025,8 @@ class StereoDispCombiner(StereoCombiner):
         tel_to_array_indices: np.ndarray,
         n_array_events: int,
         prefix_tel: str,
+        fov_lon_values: np.ndarray,
+        fov_lat_values: np.ndarray,
     ):
         if np.count_nonzero(valid) == 0:
             nan = u.Quantity(
@@ -993,10 +1036,6 @@ class StereoDispCombiner(StereoCombiner):
 
         weights = self._calculate_weights(mono_predictions[valid])
         _, _, valid_multiplicity, _ = get_subarray_index(mono_predictions[valid])
-
-        fov_lon_values, fov_lat_values = calc_fov_lon_lat(
-            mono_predictions[valid], prefix_tel
-        )
 
         combs_array, combs_to_multi_indices = create_combs_array(
             valid_multiplicity.max(), self.n_tel_combinations
@@ -1082,6 +1121,29 @@ class StereoDispCombiner(StereoCombiner):
             ]
 
         return alt, az
+
+    @staticmethod
+    def _transform_to_nominal(mono_predictions, fov_lons, fov_lats):
+        """Transform telescope-frame DISP candidates to the shared nominal frame."""
+        telescope_pointing = SkyCoord(
+            alt=mono_predictions["telescope_pointing_altitude"].quantity[:, None],
+            az=mono_predictions["telescope_pointing_azimuth"].quantity[:, None],
+            frame=AltAz(),
+        )
+        array_pointing = SkyCoord(
+            alt=mono_predictions["subarray_pointing_lat"].quantity[:, None],
+            az=mono_predictions["subarray_pointing_lon"].quantity[:, None],
+            frame=AltAz(),
+        )
+        candidates = SkyCoord(
+            fov_lon=fov_lons * u.deg,
+            fov_lat=fov_lats * u.deg,
+            frame=TelescopeFrame(telescope_pointing=telescope_pointing),
+        ).transform_to(NominalFrame(origin=array_pointing))
+        return (
+            candidates.fov_lon.to_value(u.deg),
+            candidates.fov_lat.to_value(u.deg),
+        )
 
     def _collect_telescopes_per_event(
         self,
--- a/src/ctapipe/reco/stereo_combination.py
+++ b/src/ctapipe/reco/stereo_combination.py
@@ -1146,13 +1146,13 @@ class StereoDispCombiner(StereoCombiner):
 
     @staticmethod
     def _transform_to_nominal(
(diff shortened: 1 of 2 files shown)


src/ctapipe/reco/stereo_combination.py
```diff
@@ -727,6 +740,21 @@ def _collect_valid_tel_data(self, event: ArrayEventContainer):
...
+                frame=AltAz(),
+            )
+            candidates = SkyCoord(
+                fov_lon=fov_lons * u.deg,
+                fov_lat=fov_lats * u.deg,
+                frame=TelescopeFrame(telescope_pointing=telescope_pointing),
+            ).transform_to(nominal_frame)
+            fov_lons = candidates.fov_lon.to_value(u.deg)
+            fov_lats = candidates.fov_lat.to_value(u.deg)
+            hillas_psi = np.arctan2(np.diff(fov_lats), np.diff(fov_lons))[0] * u.rad
```
maxnoe: `* u.rad` makes a copy, better is `u.Quantity(array, u.rad, copy=False)`, which just wraps the existing array in a quantity of the given unit

src/ctapipe/reco/stereo_combination.py
```diff
@@ -727,6 +740,21 @@ def _collect_valid_tel_data(self, event: ArrayEventContainer):
             fov_lons = hillas_fov_lon + signs * disp * np.cos(hillas_psi)
             fov_lats = hillas_fov_lat + signs * disp * np.sin(hillas_psi)
 
+            # Transform both DISP candidates and their axis to the common frame.
+            telescope_pointing = SkyCoord(
```
maxnoe: Why is one branch making the transform manually and one using the helper method?

src/ctapipe/reco/stereo_combination.py
```diff
@@ -795,15 +823,27 @@ def predict_table(self, mono_predictions: Table) -> Table:
...
+        fov_lon_values, fov_lat_values = calc_fov_lon_lat(mono_predictions, prefix_tel)
+        fov_lon_values, fov_lat_values = self._transform_to_nominal(
+            mono_predictions, fov_lon_values, fov_lat_values
+        )
+        hillas_psis = (
+            np.arctan2(
+                np.diff(fov_lat_values, axis=1),
+                np.diff(fov_lon_values, axis=1),
+            )[:, 0]
+            * u.rad
```
maxnoe: see above

src/ctapipe/reco/stereo_combination.py
```diff
@@ -1028,6 +1068,29 @@ def _compute_altaz_for_valid(
...
+            az=mono_predictions["telescope_pointing_azimuth"].quantity[:, None],
+            frame=AltAz(),
+        )
+        array_pointing = SkyCoord(
+            alt=mono_predictions["subarray_pointing_lat"].quantity[:, None],
+            az=mono_predictions["subarray_pointing_lon"].quantity[:, None],
+            frame=AltAz(),
+        )
+        candidates = SkyCoord(
+            fov_lon=fov_lons * u.deg,
```
maxnoe: see above. Why is it necessary to add the unit here? Shouldn't they already have a unit?

src/ctapipe/reco/stereo_combination.py
LukasBeiske: Are you referring to `_transform_to_nominal` and why its used for table-wise prediction and not in the event-loop? If yes, that was an oversight, we'll fix it 👍

src/ctapipe/reco/stereo_combination.py
AkramAki: The existing code striped the units and I tried to adhere to the base code as closely as possible. E.g. line 722

LukasBeiske: I don't see a reason atm why keeping the units would be a problem though. So changing that might be good, but I'd prefer keeping the diff in this sub-PR as small as possible.

LukasBeiske:
@maxnoe two things: 
- Does this need a changelog, since it's not getting merged into main?
- FYI: I had to approve the starting of the CI manually. I guess this is related to the new github security rules... Just in case that is not the intended behaviour .

maxnoe:
> Does this need a changelog, since it's not getting merged into main?

no

> FYI: I had to approve the starting of the CI manually. 

This is not new, any workflow of a first time outside contributor has to be approved first.

LukasBeiske:
@maxnoe And just to be sure: We can merge this into the stereo_combiner branch at our own discretion once the tests pass, correct?

maxnoe:
Sure, that's your PR branch

Hckjs:
https://github.com/cta-observatory/ctapipe/pull/3076#discussion_r3783719241
> see above. Why is it necessary to add the unit here? Shouldn't they already have a unit?

A couple of comments to @maxnoe comment:
I guess you have a point here, that the current unit handling is a bit messy, because in an earlier version of the stereo_combiner branch, the idea was to get rid of the units at the beginning and add them back later in the end because of some njit'ed functions later on an
`hillas_fov_lon + signs * disp * np.cos(hillas_psi)` in the beginning
See some old discussion here: https://github.com/cta-observatory/ctapipe/pull/2731#discussion_r2486574635

After some generalization i merged the calculation of the minium distances to one function used by the table-wise and event-wise processing (`calc_combs_min_distances`), which can handle astropy units. I did unfortunatly not simplify the unit handling accordingly...

The only function, which currently cannot handle astropy units is `weighted_mean_std_ufunc`. For the rest i think we can save some overhead by reducing the conversions/strip of units to a minimum. i'll make some suggestions in the following

/edit see https://github.com/cta-observatory/ctapipe/pull/3076#issuecomment-5299584492

src/ctapipe/reco/stereo_combination.py
```diff
@@ -776,12 +786,34 @@ def _collect_valid_tel_data(self, event: ArrayEventContainer):
 
             hillas_fov_lon = dl1.hillas.fov_lon.to_value(u.deg)
             hillas_fov_lat = dl1.hillas.fov_lat.to_value(u.deg)
-            hillas_psi = dl1.hillas.psi
+            hillas_psi = dl1.hillas.psi.to_value(u.rad)
             disp = disp_reco.parameter.to_value(u.deg)
```
Hckjs, line 790: Since all the functions used in the event-wise processing can handle units, we can keep them here, so just:
```
            hillas_fov_lon = dl1.hillas.fov_lon
            hillas_fov_lat = dl1.hillas.fov_lat
            hillas_psi = dl1.hillas.psi
            disp = disp_reco.parameter
```

`hillas_fov_lon + signs * disp * np.cos(hillas_psi)` will work then, since both sides of the "add" are convertible.

src/ctapipe/reco/stereo_combination.py
```diff
@@ -776,12 +786,34 @@ def _collect_valid_tel_data(self, event: ArrayEventContainer):
...
 
             fov_lons = hillas_fov_lon + signs * disp * np.cos(hillas_psi)
             fov_lats = hillas_fov_lat + signs * disp * np.sin(hillas_psi)
 
+            # Convert to quantity to ensure the helper functions can handle the inputs correctly
+            fov_lons = u.Quantity(fov_lons, u.deg, copy=COPY_IF_NEEDED)
+            fov_lats = u.Quantity(fov_lats, u.deg, copy=COPY_IF_NEEDED)
```
Hckjs: Not needed anymore, since we already kept the units

maxnoe:
It's fine to strip units for performance, but then one needs to be careful when and how they are added back.

E.g. this is bad:

```
SkyCoord(alt=alt * u.deg, az=az*u.deg, frame=AltAz())
```

because it makes unnecessary copies of potentially large arrays.

These can be avoided by not using the naive `* unit` syntax, e.g. using the `u.Quantity(array, unit, copy=False)` approach or even passing `unit=` to `SkyCoord`:

```
In [1]: import numpy as np

In [2]: import astropy.units as u

In [3]: from astropy.coordinates import SkyCoord, AltAz

In [4]: frame = AltAz()

In [5]: az = np.random.normal(90, 5, size=10000)

In [6]: alt = np.random.normal(70, 2, size=10000)

In [7]: %timeit SkyCoord(alt=alt * u.deg, az=az * u.deg, frame=frame)
78 μs ± 812 ns per loop (mean ± std. dev. of 7 runs, 10,000 loops each)

In [8]: %timeit SkyCoord(alt=alt, az=az, unit=u.deg, copy=False, frame=frame)
60.6 μs ± 935 ns per loop (mean ± std. dev. of 7 runs, 10,000 loops each)
```

src/ctapipe/reco/stereo_combination.py
```diff
@@ -776,12 +786,34 @@ def _collect_valid_tel_data(self, event: ArrayEventContainer):
...
+                subarray_pointing_alt,
+                subarray_pointing_az,
+                fov_lons,
+                fov_lats,
+            )
+            hillas_psi = u.Quantity(
+                np.arctan2(np.diff(fov_lats), np.diff(fov_lons))[0],
+                u.rad,
+                copy=COPY_IF_NEEDED,
+            )
```
Hckjs, line 811: Quantity wrapper not needed, since lon and lat kept the units, so just:
```
hillas_psi = np.arctan2(np.diff(fov_lats), np.diff(fov_lons))[0]
```

src/ctapipe/reco/stereo_combination.py
Hckjs, line 854: Numpy tries to generate a common datatype when wrapping the fov_lon/lat_values list here which will not work with astropy units, so you can just :
```
           u.Quantity(fov_lon_values)[best_tels_idx],
           u.Quantity(fov_lat_values)[best_tels_idx],
```

src/ctapipe/reco/stereo_combination.py
```diff
@@ -850,15 +882,51 @@ def predict_table(self, mono_predictions: Table) -> Table:
         valid = mono_predictions[f"{prefix_tel}_is_valid"].copy()
         self._require_disp_column(mono_predictions, prefix_tel)
 
+        # Returns values as to_value(u.deg)
```
Hckjs: This returns quantities when we adapt the `calc_fov_lon_lat`-function accordingly

src/ctapipe/reco/stereo_combination.py
```diff
@@ -850,15 +882,51 @@ def predict_table(self, mono_predictions: Table) -> Table:
...
+        ]
+
+        fov_lon_values, fov_lat_values = self._transform_to_nominal(
+            tel_pointing_alt,
+            tel_pointing_az,
+            subarray_pointing_alt,
+            subarray_pointing_az,
+            fov_lon_values,
+            fov_lat_values,
+        )
```
Hckjs, line 904: Keeping the quantities, we adapt the _transform_to_nominal, so just:

```
         fov_lon_values, fov_lat_values = self._transform_to_nominal(
             mono_predictions["telescope_pointing_altitude"][:, None],
             mono_predictions["telescope_pointing_azimuth"][:, None],
             mono_predictions["subarray_pointing_lat"][:, None],
             mono_predictions["subarray_pointing_lon"][:, None],
             fov_lon_values,
             fov_lat_values,
         )
```

The upper part below
```
# Convert to radians for the angular difference calculation
```
can be deleted

src/ctapipe/reco/stereo_combination.py
```diff
@@ -850,15 +882,51 @@ def predict_table(self, mono_predictions: Table) -> Table:
...
+        hillas_psis = u.Quantity(
+            np.arctan2(
+                np.diff(fov_lat_values, axis=1),
+                np.diff(fov_lon_values, axis=1),
+            )[:, 0],
+            u.rad,
+            copy=COPY_IF_NEEDED,
+        )
```
Hckjs, line 912: As in the event-wise processing, since we have kept the units:
```
        hillas_psis = np.arctan2(
            np.diff(fov_lat_values, axis=1),
            np.diff(fov_lon_values, axis=1),
        )[:, 0]

```

src/ctapipe/reco/stereo_combination.py
Hckjs, line 1082: Here we need to get rid of the unit, because this function cant handle it:
```
comb_fov_lons.to_value(u.deg)
```

Hckjs, line 1089: same as above

src/ctapipe/reco/stereo_combination.py
```diff
@@ -1083,6 +1151,54 @@ def _compute_altaz_for_valid(
...
+        candidates = SkyCoord(
+            fov_lon=fov_lons,
+            fov_lat=fov_lats,
+            frame=TelescopeFrame(telescope_pointing=telescope_pointing),
+        ).transform_to(NominalFrame(origin=array_pointing))
+        return (
+            candidates.fov_lon.to_value(u.deg),
+            candidates.fov_lat.to_value(u.deg),
+        )
```
Hckjs, line 1196: With the changes above we dont have to strip here, just:
```
            candidates.fov_lon,
            candidates.fov_lat,
```

Hckjs:
~in `telescope_event_handling.py` you also have to keep the quantities in `calc_fov_lon_lat`, so just:~
```
    hillas_fov_lon = tel_table["hillas_fov_lon"].quantity
    hillas_fov_lat = tel_table["hillas_fov_lat"].quantity
    hillas_psi = tel_table["hillas_psi"].quantity
    disp = tel_table[f"{prefix}_parameter"].quantity
```

Hckjs:
I think it would be also nice to have a unit test for an event with at least two participating telescopes with different pointings.

src/ctapipe/reco/stereo_combination.py
Hckjs: i checked the performance of `calc_combs_min_distances` and its way faster (up to 6 times) to use it without units for single events in the event loop. For the table-wise processing with ~100k events the difference is not mentionable. so i would strip the unit here:

```
u.Quantity(fov_lon_values).to_value(u.deg)[best_tels_idx]
```

and then again a `u.Quantity(...)` wrapper afterwards. Maybe add a small inline comment

LukasBeiske:
> I think it would be also nice to have a unit test for an event with at least two participating telescopes with different pointings.

If we just change the telescope pointings in the dummy table to not be identical for all telescope events, this should be covered, no?

Hckjs:
I just checked the performances for single array events in the event-loop and it looks like its much faster calculating the fov_lon/lats and the minimum distances without units (also adding the strip and wrap back of the units...):
```python
import numpy as np
import astropy.units as u

signs = np.array([-1, 1])

hillas_fov_lon = 0.5 * u.deg
disp = 0.8 * u.deg
hillas_psi = 40 * u.deg

%timeit hillas_fov_lon + signs * disp * np.cos(hillas_psi)
%timeit (u.Quantity(hillas_fov_lon.to_value(u.deg) + signs * disp.to_value(u.deg) * np.cos(hillas_psi.to_value(u.rad)), u.deg, copy=False,))
```
```
28.8 μs ± 344 ns per loop (mean ± std. dev. of 7 runs, 10,000 loops each)
7.99 μs ± 40.9 ns per loop (mean ± std. dev. of 7 runs, 100,000 loops each)
```

```
# Without units
%timeit calc_combs_min_distances(index_tel_combs, fov_lon, fov_lat, weights)
68 μs ± 194 ns per loop (mean ± std. dev. of 7 runs, 10,000 loops each)

# With units
%timeit calc_combs_min_distances(index_tel_combs, fov_lon_q, fov_lat_q, weights)
386 μs ± 1.31 μs per loop (mean ± std. dev. of 7 runs, 1,000 loops each)
```

Im sorry, didn't expect to be that much of a difference. I would just leave it as it is.

Hckjs:
> > I think it would be also nice to have a unit test for an event with at least two participating telescopes with different pointings.
> 
> If we just change the telescope pointings in the dummy table to not be identical for all telescope events, this should be covered, no?

If you write a new unit test, adapting the dummy table just inside the test, i think yes. Otherwise i guess you also have to adapt all the expected reconstructed parameter aswell in the other tests

src/ctapipe/reco/stereo_combination.py
```diff
@@ -850,15 +882,51 @@ def predict_table(self, mono_predictions: Table) -> Table:
         valid = mono_predictions[f"{prefix_tel}_is_valid"].copy()
         self._require_disp_column(mono_predictions, prefix_tel)
 
+        # Returns values as to_value(u.deg)
+        fov_lon_values, fov_lat_values = calc_fov_lon_lat(mono_predictions, prefix_tel)
+
+        # Convert to radians for the angular difference calculation
```
LukasBeiske: This comment seems misplaced/wrong

src/ctapipe/reco/stereo_combination.py
```diff
@@ -776,12 +786,34 @@ def _collect_valid_tel_data(self, event: ArrayEventContainer):
...
             hillas_fov_lat = dl1.hillas.fov_lat.to_value(u.deg)
-            hillas_psi = dl1.hillas.psi
+            hillas_psi = dl1.hillas.psi.to_value(u.rad)
             disp = disp_reco.parameter.to_value(u.deg)
 
             fov_lons = hillas_fov_lon + signs * disp * np.cos(hillas_psi)
             fov_lats = hillas_fov_lat + signs * disp * np.sin(hillas_psi)
 
+            # Convert to quantity to ensure the helper functions can handle the inputs correctly
+            fov_lons = u.Quantity(fov_lons, u.deg, copy=COPY_IF_NEEDED)
```
LukasBeiske: Maybe move this conversion into the `_transform_to_nominal` call itself, since right now its a bit confusing, that you transform them here into quantities, but then they get overwritten by `_transform_to_nominal` which returns plain numbers.

src/ctapipe/reco/stereo_combination.py
```diff
@@ -850,15 +882,51 @@ def predict_table(self, mono_predictions: Table) -> Table:
         valid = mono_predictions[f"{prefix_tel}_is_valid"].copy()
         self._require_disp_column(mono_predictions, prefix_tel)
 
+        # Returns values as to_value(u.deg)
+        fov_lon_values, fov_lat_values = calc_fov_lon_lat(mono_predictions, prefix_tel)
+
+        # Convert to radians for the angular difference calculation
+        fov_lon_values = u.Quantity(fov_lon_values, u.deg, copy=COPY_IF_NEEDED)
```
LukasBeiske: See above

src/ctapipe/reco/stereo_combination.py
```diff
@@ -1083,6 +1151,54 @@ def _compute_altaz_for_valid(
 
         return alt, az
 
+    @staticmethod
+    def _transform_to_nominal(
```
LukasBeiske, line 1148: Adding either the `@quantity_input` decorator or type hints for the parameters would be good

Hckjs (approved):

PR #3078, opened 2026-08-19
Fixed bug in get_cache_path

kosack:
The `env_override` variable was not used for the actual environment name, and instead it used the hard-coded "CTAPIPE_CACHE" even if the user specified a different variable.  This did not affect anything inside ctapipe, only external packages that use this method.


maxnoe (approved):

maxnoe:
Please rebase with main to fix the CI failure

PR #3079, opened 2026-08-19
Ignore numpy DeprecationWarnings from astropy

maxnoe:


maxnoe:
Unfortunately, I could not filter by module here, as the warning is captured by pytest and then re-emitted in some cases, which makes it a local warning apparently.

mexanick (approved):

PR #3082, opened 2026-08-25
Raise HillasParameterizationError on images containing negative values

### LINKED ISSUE #3081: Hillas parameters failing if image contains negative values
**Describe the bug**
The `ctapipe.image.hillas.hillas_parameters()` function fails if at least one element of the image np.array is negative.

Usually this is not a problem, because the standard tail-cut cleaning does not retain pixel with values lower than the secondary threshold (which is never set to negative). However this might not be always the case if an other cleaning algorithm is used , if no cleaning is used or dilate is called several times on the cleaned mask.

**To Reproduce**
1. Create and image with a negative value in it
2. Call hillas_parameters on this image

(See below test created to reproduce this error)

**Expected behavior**

I expected either the code to not fail if any element of the image is negative or that it raises a `HillasParameterizationError`. 

**Supporting information**
```bash
============================================================================================================================ FAILURES =============================================================================================================================
_________________________________________________________________________________________________________________ test_hillas_on_negative_images __________________________________________________________________________________________________________________

    def test_hillas_on_negative_images():
        s = SubarrayDescription.read("dataset://gamma_prod5.simtel.zst")
        geometry = s.tel[1].camera.geometry
        image, mask = create_sample_image(geometry=geometry)
    
        assert np.min(image) < 0
    
>       hillas_parameters(geometry, image)

src/ctapipe/image/tests/test_hillas.py:404: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _
src/ctapipe/image/hillas.py:119: in hillas_parameters
    cov = np.cov(delta_x, delta_y, aweights=image, ddof=0)
          ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _

m = array([-0.18162264, -0.21941934, -0.22886811, ...,  0.48925769,
       -0.63518747, -0.68243288], shape=(1855,)), y = array([[-0.27776759, -0.31050002, -0.26140093, ..., -1.24338641,
        -1.35794305, -1.34157634]], shape=(1, 1855)), rowvar = True
bias = False, ddof = 0, fweights = None, aweights = array([-1.01671159, -0.01671159, -0.01671159, ...,  1.98328841,
        0.98328841, -0.01671159], shape=(1855,)), dtype = dtype('float64')

    @array_function_dispatch(_cov_dispatcher)
    def cov(m, y=None, rowvar=True, bias=False, ddof=None, fweights=None,
            aweights=None, *, dtype=None):
        """
        Estimate a covariance matrix, given data and weights.
    
        Covariance indicates the level to which two variables vary together.
        If we examine N-dimensional samples, :math:`X = [x_1, x_2, ..., x_N]^T`,
        then the covariance matrix element :math:`C_{ij}` is the covariance of
        :math:`x_i` and :math:`x_j`. The element :math:`C_{ii}` is the variance
        of :math:`x_i`.
    
        See the notes for an outline of the algorithm.
    
        Parameters
        ----------
        m : array_like
            A 1-D or 2-D array containing multiple variables and observations.
            Each row of `m` represents a variable, and each column a single
            observation of all those variables. Also see `rowvar` below.
        y : array_like, optional
            An additional set of variables and observations. `y` has the same form
            as that of `m`.
        rowvar : bool, optional
            If `rowvar` is True (default), then each row represents a
            variable, with observations in the columns. Otherwise, the relationship
            is transposed: each column represents a variable, while the rows
            contain observations.
        bias : bool, optional
            Default normalization (False) is by ``(N - 1)``, where ``N`` is the
            number of observations given (unbiased estimate). If `bias` is True,
            then normalization is by ``N``. These values can be overridden by using
            the keyword ``ddof`` in numpy versions >= 1.5.
        ddof : int, optional
            If not ``None`` the default value implied by `bias` is overridden.
            Note that ``ddof=1`` will return the unbiased estimate, even if both
            `fweights` and `aweights` are specified, and ``ddof=0`` will return
            the simple average. See the notes for the details. The default value
            is ``None``.
        fweights : array_like, int, optional
            1-D array of integer frequency weights; the number of times each
            observation vector should be repeated.
        aweights : array_like, optional
            1-D array of observation vector weights. These relative weights are
            typically large for observations considered "important" and smaller for
            observations considered less "important". If ``ddof=0`` the array of
            weights can be used to assign probabilities to observation vectors.
        dtype : data-type, optional
            Data-type of the result. By default, the return data-type will have
            at least `numpy.float64` precision.
    
            .. versionadded:: 1.20
    
        Returns
        -------
        out : ndarray
            The covariance matrix of the variables.
    
        See Also
        --------
        corrcoef : Normalized covariance matrix
    
        Notes
        -----
        Assume that the observations are in the columns of the observation
        array `m` and let ``f = fweights`` and ``a = aweights`` for brevity. The
        steps to compute the weighted covariance are as follows::
    
            >>> m = np.arange(10, dtype=np.float64)
            >>> f = np.arange(10) * 2
            >>> a = np.arange(10) ** 2.
            >>> ddof = 1
            >>> w = f * a
            >>> v1 = np.sum(w)
            >>> v2 = np.sum(w * a)
            >>> m -= np.sum(m * w, axis=None, keepdims=True) / v1
            >>> cov = np.dot(m * w, m.T) * v1 / (v1**2 - ddof * v2)
    
        Note that when ``a == 1``, the normalization factor
        ``v1 / (v1**2 - ddof * v2)`` goes over to ``1 / (np.sum(f) - ddof)``
        as it should.
    
        Examples
        --------
        >>> import numpy as np
    
        Consider two variables, :math:`x_0` and :math:`x_1`, which
        correlate perfectly, but in opposite directions:
    
        >>> x = np.array([[0, 2], [1, 1], [2, 0]]).T
        >>> x
        array([[0, 1, 2],
               [2, 1, 0]])
    
        Note how :math:`x_0` increases while :math:`x_1` decreases. The covariance
        matrix shows this clearly:
    
        >>> np.cov(x)
        array([[ 1., -1.],
               [-1.,  1.]])
    
        Note that element :math:`C_{0,1}`, which shows the correlation between
        :math:`x_0` and :math:`x_1`, is negative.
    
        Further, note how `x` and `y` are combined:
    
        >>> x = [-2.1, -1,  4.3]
        >>> y = [3,  1.1,  0.12]
        >>> X = np.stack((x, y), axis=0)
        >>> np.cov(X)
        array([[11.71      , -4.286     ], # may vary
               [-4.286     ,  2.144133]])
        >>> np.cov(x, y)
        array([[11.71      , -4.286     ], # may vary
               [-4.286     ,  2.144133]])
        >>> np.cov(x)
        array(11.71)
    
        """
        # Check inputs
        if ddof is not None and ddof != int(ddof):
            raise ValueError(
                "ddof must be integer")
    
        # Handles complex arrays too
        m = np.asarray(m)
        if m.ndim > 2:
            raise ValueError("m has more than 2 dimensions")
    
        if y is not None:
            y = np.asarray(y)
            if y.ndim > 2:
                raise ValueError("y has more than 2 dimensions")
    
        if dtype is None:
            if y is None:
                dtype = np.result_type(m, np.float64)
            else:
                dtype = np.result_type(m, y, np.float64)
    
        X = array(m, ndmin=2, dtype=dtype)
        if not rowvar and m.ndim != 1:
            X = X.T
        if X.shape[0] == 0:
            return np.array([]).reshape(0, 0)
        if y is not None:
            y = array(y, copy=None, ndmin=2, dtype=dtype)
            if not rowvar and y.shape[0] != 1:
                y = y.T
            X = np.concatenate((X, y), axis=0)
    
        if ddof is None:
            if bias == 0:
                ddof = 1
            else:
                ddof = 0
    
        # Get the product of frequencies and weights
        w = None
        if fweights is not None:
            fweights = np.asarray(fweights, dtype=float)
            if not np.all(fweights == np.around(fweights)):
                raise TypeError(
                    "fweights must be integer")
            if fweights.ndim > 1:
                raise RuntimeError(
                    "cannot handle multidimensional fweights")
            if fweights.shape[0] != X.shape[1]:
                raise RuntimeError(
                    "incompatible numbers of samples and fweights")
            if any(fweights < 0):
                raise ValueError(
                    "fweights cannot be negative")
            w = fweights
        if aweights is not None:
            aweights = np.asarray(aweights, dtype=float)
            if aweights.ndim > 1:
                raise RuntimeError(
                    "cannot handle multidimensional aweights")
            if aweights.shape[0] != X.shape[1]:
                raise RuntimeError(
                    "incompatible numbers of samples and aweights")
            if any(aweights < 0):
>               raise ValueError(
                    "aweights cannot be negative")
E               ValueError: aweights cannot be negative

../../../../.micromamba/envs/cta-dev/lib/python3.13/site-packages/numpy/lib/_function_base_impl.py:2870: ValueError
```

**Additional context**

calispac:
Solves #3081 according to discussions


src/ctapipe/image/hillas.py
```diff
@@ -101,6 +110,11 @@ def hillas_parameters(geom, image):
     if size == 0.0:
         raise HillasParameterizationError("size=0, cannot calculate HillasParameters")
 
+    if (image < 0).any():
```
maxnoe: I'm not sure this is a good idea. This will perform an expensive operation (comparing each entry of `image`) just to change an error message.

This penalizes performance on the "happy path"

calispac: you prefer to try and catch the np.cov() ?

maxnoe: yes

calispac: done

calispac: This now makes the ctao-sonarqube unhappy "[Refactor this function to reduce its Cognitive Complexity from 17 to the 15 allowed.]"

maxnoe: This is acceptable, but I think you could also just wrap the error message:

```
raise HillasParametrizationError("Failed to compute covariance") from e
```

calispac: Ok I changed it. Hopefully it passes now.

maxnoe (approved):

kosack (approved):

PR #3083, opened 2026-08-25
Bump sphinx to 9.1 and fix doc issues

kosack:
Docs are now build with Sphinx 9.x. The documentation pages was also revised to
avoid documenting the same things in both top-level modules and submodules,
which lead to duplicate reference errors.


maxnoe:
I think we can fix those warnings with adding `:imported-members:` to the modules where we import classes defined lower in the module hierarchy

kosack:
There's also some places where we document the classes by module, and some by submodule, which I think also causes this duplicate entry. E.g. we have rst files with 

```rst
.. automodapi:: ctapipe.instrument.optics
    :no-inheritance-diagram:
```

Which is so we get the docstring from the submodule, but then we get also the members. So adding `:noindex:` to those and just documenting them at the parent module level, should help.

ideally, It would be nice to have any submodules that are not part of the public API be renamed to `_submodule.py` so it's clear which should be imported, but that's a larger issue.

kosack:
I'll use this opportunity to do some small doc restructuring to unify when we document modules vs submodules... Should fix the duplication warnings.   So far it looks reasonably easy, but sphinx always gives me headaches.

docs/api-reference/core/traits.rst
```diff
@@ -19,3 +19,4 @@ Reference/API
 .. automodapi:: ctapipe.core.traits
     :include-all-objects:
     :no-inheritance-diagram:
+    :skip: Int
```
maxnoe: We have a lot of aliases from traitlets in that module:

https://github.com/cta-observatory/ctapipe/blob/fd6f178de04ecbedd8d0b9a502bb219f702ca8aa/src/ctapipe/core/traits.py#L74-L92

Why is `Int` special?

kosack: Not sure, actually! Somewhere is triggering it to be documented, but not the others. I couldn't find where - Sphinx doesn't give very good feedback about where errors come from.  It's also not clear why it's documentation causes problems.

kosack: Even stranger is that it should be excluded by the nitpick_ignore, or the hack in conf.py that fixes the aliases.  But skipping it here does indeed get rid of the warnings

maxnoe: I think the problem with nitpick ignore was that you only had `...Int`, but it complaiend about methods like `...Int.some_method`.

you can specify `nitpick_ignore_regex`, e.g. `Int\..*`

kosack: I'll try that. Still not sure why only Int has this issue - the others have methods as well, and don't get warnings

kosack: Strangely, traitlets.Integer (which is an alias for traitlets.Int) is documented correctly. I think it is that we mix both that is causing the problem.

kosack: yes, the issue is the hack in `conf.py`: it rewrites the name and module for both Int and Integer, but then one overwrites the other.

maxnoe: aha! yes, if that is an alias already in traitlets, it might very well confuse sphinx

kosack:
So the issue was that we had both Int and Integer in various places, but those point to the same class implementation in traitlets. I see no reason to support both here, so I changed all to use `Int`, and then our alias re-writing code in conf.py doesn't try to overwrite the same class twice.   That fixes a lot of the doc errors.

kosack:
But just in case this change could break some external code that uses `ctapipe.traitlets.Integer`, I added that back in, but ensure that it is not documented.  Therefore there is no API change here, just some internal refactoring and fixed docs.

maxnoe (approved):

PR #3085, opened 2026-08-26
Add mirror facet info optics

burmist-git:


src/ctapipe/instrument/tests/test_MirrorFacetsDescription.py
```diff
@@ -0,0 +1,40 @@
...
+    """Create a mirror facet description and convert its shapes to enums."""
+    description = MirrorFacetsDescription(
+        id=np.array([1, 2, 3]),
+        x=np.array([0.0, 1.0, 2.0]) * u.m,
+        y=np.array([0.0, 0.5, 1.0]) * u.m,
+        z=np.array([10.0, 10.1, 10.2]) * u.m,
+        nx=np.array([0.0, 0.0, 0.0]),
+        ny=np.array([0.0, 0.0, 0.0]),
+        nz=np.array([1.0, 1.0, 1.0]),
+        diameter=np.array([1.2, 1.2, 1.2]) * u.m,
```
kosack: In `CameraGeometry`, we store the pixel area because the diameter is not well defined if you have different shapes.  Similar would work here: instead of diameter, store the area, and then you know depending on the shape how to draw it (or use it in a ray-tracing environment).

src/ctapipe/instrument/optics.py
```diff
@@ -97,6 +97,51 @@ class ReflectorShape(Enum):
...
+        mirror_shape,
+    ):
+        self.id = id
+        self.x = x
+        self.y = y
+        self.z = z
+        self.nx = nx
+        self.ny = ny
+        self.nz = nz
+        self.diameter = diameter
```
kosack: diameter → area

src/ctapipe/instrument/tests/test_MirrorFacetsDescription.py
kosack: See the functions like _create_hex_patches() in [CameraDisplay](https://ctapipe.readthedocs.io/en/latest/_modules/ctapipe/visualization/mpl_camera.html#CameraDisplay) . Those convert a given pixel _shape_ and _area_ into a patch (polygon).  A similar method could be used for muon studies or visualization of the mirror

```python
    @staticmethod
    def create_patches(shape, pix_x, pix_y, pix_width, pix_rotation=0 * u.deg):
        if shape == PixelShape.HEXAGON:
            return CameraDisplay._create_hex_patches(
                pix_x, pix_y, pix_width, pix_rotation
            )

        if shape == PixelShape.CIRCLE:
            return CameraDisplay._create_circle_patches(pix_x, pix_y, pix_width)

        if shape == PixelShape.SQUARE:
            return CameraDisplay._create_square_patches(
                pix_x, pix_y, pix_width, pix_rotation
            )

        raise ValueError(f"Unsupported pixel shape {shape}")

```

Edit: Actually, it's a bit more convoluted than needs to be there, as we convert between area→ width (CameraGeometry.pixel_width(), per type) and then width→ shape in the visualization.  But here, you could just go from area to shape directly.  The point is that with the area + shape, you know how to build a regular polygon or circle.

burmist-git: Great suggestion! With the mirror shape, one can also retrieve the flat-to-flat, side-to-side, or indeed the diameter.

src/ctapipe/instrument/optics.py
burmist-git: yes, thank you.

burmist-git:
@kosack Just in case, I am preparing a constructor called from_table_csv, where these parameters need to be stored. Is it allowed to have them in a CSV, or are only Astropy Tables suitable for this?

kosack:
> @kosack Just in case, I am preparing a constructor called from_table_csv, where these parameters need to be stored. Is it allowed to have them in a CSV, or are only Astropy Tables suitable for this?

You should use astropy.table, which supports CSV.  In fact, CSV is not allowed, as it has no notion of metadata. So even when using astropy.table, you should use ECSV files in your tests. See how this is done in all the other instrument classes: we have metadata for e.g. the version and other common attributes, and use astropy.table for all storage. Then you get the power behind that: you can support any format that has metadata storage. 

Note that you should be able to run `ctapipe-dump-instrument` and get your output as FITS or ECSV files if everything is done correctly.

The interface is not well specified (it should be), but we have for each class a `to_table()` and `from_table()` method. Those define the table format.  You can just look at e.g. CameraGeometry, CameraReadout, OpticsDescription and copy what they do.

burmist-git:
@kosack Who is going to provide the table in the real world? We need to have it for every telescope type, and probably for every individual telescope, since some mirrors are going to be aged (removed), and this can differ even within the same telescope type. I feel that this particular code development is more about simpipe?

maxnoe:
>  I feel that this particular code development is more about simpipe?

No, this is instrument information that is needed as _input_ for simpipe in the same way it is needed as input here. It needs to be provided by the telescope teams to us, e.g. via the instrument configuration in SOSS.

burmist-git:
> > I feel that this particular code development is more about simpipe?
> 
> No, this is instrument information that is needed as _input_ for simpipe in the same way it is needed as input here. It needs to be provided by the telescope teams to us, e.g. via the instrument configuration in SOSS.

Thank you for your answer. I’m continuing with the implementation.

kosack:
> I feel that this particular code development is more about simpipe?

SimPipe already has this information in fact, inside the telescope model. So if we need some table for each optics type to use for testing, we can read that from the telecope model and dump it in the table format you define here. That is exactly what we did for the other instrument components: we have test data extracted from simulations that we use in unit tests.  So the main thing here  is just to define a sufficient common data format using astropy.table and an API to access it (the MirrorFacetDescription).  Then step 2 is to collect the test tables from the simulation model so you can write unit tests that use this info, e.g. in the muon analysis.   

Note if you have those tables already, send them to @maxnoe or me and we can add them to the ctapipe test data server.

Finally, there is now an interface to read a full InstrumentDescription from a directory full of instrument tables. The final step would be to add this information to that structure.  That directory will in the future be populated from the SOSS database for CTAO, but that is not required: we just have  way to specify the instrument.   Then you will get the mirror info automatically in the SubarrayDescription  that is provided by the EventSource (which is loaded from `SubarrayDescription.from_service_data()`) 

This makes it transparent to the user: the muon code just looks in the subarray description for the info. Where it came from doesn't matter

kosack:
For the last part, the current structure has files like the following for each tel_id:
```
    │   ├── 002/
    │   │   ├── 002.optics.ecsv -> ../LSTN/LSTN.optics.ecsv
    │   │   ├── 002.camgeom.fits.gz -> ../LSTN/LSTN.camgeom.fits.gz
    │   │   └── 002.camreadout.fits.gz -> ../LSTN/LSTN.camreadout.fits.gz
```

So this would just add a new file, like `002.mirrorfacets.fits.gz` that can either be a symlink to a common one per telescope type,  like in the example, or be different for each telescope if that is needed.

kosack:
One more question I have here is: how do you propose to support dual-mirror telescopes? Do we need a higher-level class like `MirrorDescription` that can contain one or more `MirrorFacetDescriptions`? Or you could add a column for the `mirror_id`, which would be e.g. 0=primary, 1=secondary, or something like that?

src/ctapipe/instrument/optics.py
```diff
@@ -97,6 +97,51 @@ class ReflectorShape(Enum):
...
+        self.mirror_shape = np.array(
+            [MirrorFacetShape(shape) for shape in mirror_shape]
+        )
+
```
kosack, line 142: I also just realized that you will also need a rotation angle, since for non-circular mirror facets, the "up" direction is important. Also careful to define it in a way that is clear which axis it starts from, and we should document for each shape how the shapes are defined (e.g. for a hexagon, is the top flat or pointy when the rotation angle is 0.0°).

maxnoe:
> One more question I have here is: how do you propose to support dual-mirror telescopes? 

I'm not against making it more explicit, but shouldn't this already be covered by the normal vectors?

kosack:
> I'm not against making it more explicit, but shouldn't this already be covered by the normal vectors?

Yes, I guess we could make a helper function like "get_mirror_facets(mirror=Mirror.Primary)" that would select based on the normal direction.  Though just making it explicitly linked to a mirror is probably much safer and clearer.    

The relationship is: 

```mermaid

classDiagram
  Telecope *-- "1" Optics
   Optics "1" *-- "1..*"  Mirror
   Mirror "1" *-- "1..*" MirrorFacet 
```

So having a mirror_id and mirror_facet_id  makes sense.

maxnoe:
I think we need a clear definition of the coordinate system that is used and it should preferably be consistent with the "Reflector Coordinate Reference System" from this document: https://plm-extern.desy.de/#/com.siemens.splm.clientfx.tcui.xrt.showObject?uid=ywgZn0K0KPLEjD :

Screenshot_20260828_104310

maxnoe:
This can come in a separete PR, but a `.peek()` plotting the mirrors would also be helpful to check that we do it right.

burmist-git:
> > I feel that this particular code development is more about simpipe?
> 
> **SimPipe already has this information in fact,** inside the telescope model. So if we need some table for each optics type to use for testing, we can read that from the telecope model and dump it in the table format you define here. That is exactly what we did for the other instrument components: we have test data extracted from simulations that we use in unit tests. So the main thing here is just to define a sufficient common data format using astropy.table and an API to access it (the MirrorFacetDescription). Then step 2 is to collect the test tables from the simulation model so you can write unit tests that use this info, e.g. in the muon analysis.

Could you please point me to the code? I’ll probably learn something new there.

burmist-git:
> One more question I have here is: how do you propose to support dual-mirror telescopes? Do we need a higher-level class like `MirrorDescription` that can contain one or more `MirrorFacetDescriptions`? Or you could add a column for the `mirror_id`, which would be e.g. 0=primary, 1=secondary, or something like that?

In principle, one could add an additional column, which is not a bad idea. However, the z-position and normal vector direction already provide the answer to this question. That said, I’m in favor of the option with the additional column, as it would require less code to retrieve information about the primary mirror in case of SSTs.

src/ctapipe/instrument/optics.py
burmist-git: Yes, let’s add this rotation angle. However, in principle, it can also be retrieved from the shape of the individual tile and center positions. This rotation angle will also be provided as a 2D projection onto the XY plane, assuming that the telescope’s optical axis points in the Z direction.

burmist-git:
> I think we need a clear definition of the coordinate system that is used and it should preferably be consistent with the "Reflector Coordinate Reference System" from this document: https://plm-extern.desy.de/#/com.siemens.splm.clientfx.tcui.xrt.showObject?uid=ywgZn0K0KPLEjD :

Somehow, I can’t log in to [https://plm-extern.desy.de](https://plm-extern.desy.de/?utm_source=chatgpt.com). Is this normal? Could I have access to it as well?

src/ctapipe/instrument/optics.py
maxnoe: I think the rotation angle is needed, at least for squares and hexagons, they are not required to be ax-parallel by design, right? And at least for hexagons, you would have to include two shape versions (flat top, pointy top) in case you would require ax-parallel tiles.

burmist-git:
> This can come in a separete PR, but a `.peek()` plotting the mirrors would also be helpful to check that we do it right.

By .peek(), do you mean the visualization that Karl suggested?

burmist-git:
> For the last part, the current structure has files like the following for each tel_id:
> 
> ```
>     │   ├── 002/
>     │   │   ├── 002.optics.ecsv -> ../LSTN/LSTN.optics.ecsv
>     │   │   ├── 002.camgeom.fits.gz -> ../LSTN/LSTN.camgeom.fits.gz
>     │   │   └── 002.camreadout.fits.gz -> ../LSTN/LSTN.camreadout.fits.gz
> ```
> 
> So this would just add a new file, like `002.mirrorfacets.fits.gz` that can either be a symlink to a common one per telescope type, like in the example, or be different for each telescope if that is needed.

So, where is this kind of information stored?

maxnoe:
> By .peek(), do you mean the visualization that Karl suggested?

Yes, a new function similar to the ones we have on `SubarrayDescription` that produces a plot of the mirror.

> So, where is this kind of information stored?

That should not matter for the implementation here, what is important here is to implement writing and loading of that file into/from that directory structure.

In DPPS production, the jobs would get this info on CVMFS, eventually exported from SOSS and filled by hand before.

kosack:
Was this closed on purpose? It seems to be moving in a good direction

burmist-git:
> Yes, a new function similar to the ones we have on `SubarrayDescription` that produces a plot of the mirror.

image

but more like this : 
image

or better like this:
image

maxnoe:
> or better like this:

Looks good! I think this is from simtools?

https://github.com/gammasim/simtools/blob/87e5b9cb42ad43de7f6151206fc5749583bbe137/src/simtools/visualization/plot_mirrors.py#L203

burmist-git:
@Voutsi, somehow I could not trigger the upload of the FITS/ECSV files from MinIO.

src/ctapipe/instrument/optics.py
```diff
@@ -97,6 +97,274 @@ class ReflectorShape(Enum):
...
+        mirror_shape,
+    ):
+        self.id = id
+        self.x = x
+        self.y = y
+        self.z = z
+        self.nx = nx
+        self.ny = ny
+        self.nz = nz
+        self.surface = surface
```
kosack: What is surface? if it is the area, please call it `area` or `surface_area`. Should also add docstrings to describe these.

src/ctapipe/instrument/optics.py
```diff
@@ -97,6 +97,274 @@ class ReflectorShape(Enum):
...
+    UNKNOWN = "UNKNOWN"
+    #: Individual mirror facets of circular shape.
+    CIRCLE = "CIRCLE"
+    #: Individual mirror facets of square shape.
+    SQUARE = "SQUARE"
+    #: Individual mirror facets of hexagonal shape.
+    HEXAGON = "HEXAGON"
+
+
+class MirrorFacetsDescription:
```
kosack: Can just call this `MirrorDescription`, like we do with other classes.  E.g. we have `CameraGeometry` that contains descriptions of the pixels, not `CameraPixelsGeometry`.

kosack (changes requested):

Voutsi:
> @Voutsi, somehow I could not trigger the upload of the FITS/ECSV files from MinIO.

We uploaded the files in minio last week. But maybe here you mean download from MinIO? Is that the reason all tests are skipped?

Voutsi:
> > @Voutsi, somehow I could not trigger the upload of the FITS/ECSV files from MinIO.
> 
> We uploaded the files in minio last week. But maybe here you mean download from MinIO? Is that the reason all tests are skipped?

There are two tests, test_get_facet_size_per_shape and test_create_patches_per_shape, that are based on synthetic data and shouldn't be skipped in case the test data are not avaialble

src/ctapipe/instrument/tests/test_MirrorFacetsDescription.py
```diff
@@ -0,0 +1,177 @@
...
+import pytest
+
+from ctapipe.instrument.optics import MirrorFacetsDescription, MirrorFacetShape
+
+# Real mirror facet tables shipped alongside the repository for local testing.
+DATA_DIR = Path(__file__).resolve().parents[4] / "tmp_files"
+ECSV_PATH = DATA_DIR / "mirror_CTA-N-LST1_v2019-03-31.ecsv"
+FITS_PATH = DATA_DIR / "mirror_CTA-N-LST1_v2019-03-31.fits"
+
+pytestmark = pytest.mark.skipif(
```
Voutsi:
```suggestion
requires_test_data = pytest.mark.skipif(
```

src/ctapipe/instrument/tests/test_MirrorFacetsDescription.py
```diff
@@ -0,0 +1,177 @@
...
+    assert description.z[0].to_value(u.cm) == pytest.approx(120.53307587693142)
+    assert description.surface[0].to_value(u.cm**2) == pytest.approx(19746.245231688983)
+
+    assert description.nx[0] == pytest.approx(-0.08077963744975176)
+    assert description.ny[0] == pytest.approx(0.18640162657079892)
+    assert description.nz[0] == pytest.approx(0.9791471206030518)
+
+    assert np.all(description.shape == MirrorFacetShape.HEXAGON)
+
+
```
Voutsi:
```suggestion
@requires_test_data
```

src/ctapipe/instrument/tests/test_MirrorFacetsDescription.py
```diff
@@ -0,0 +1,177 @@
...
+def test_mirror_facets_description_from_ecsv():
+    description = MirrorFacetsDescription.from_table(ECSV_PATH)
+    _check_description(description)
+
+
```
Voutsi:
```suggestion
@requires_test_data
```

src/ctapipe/instrument/tests/test_MirrorFacetsDescription.py
```diff
@@ -0,0 +1,177 @@
...
+    )
+
+    size = description.get_facet_size()
+
+    assert size[0].to_value(u.m) == pytest.approx(1.0)  # radius
+    assert size[1].to_value(u.m) == pytest.approx(1.0)  # side
+    assert size[2].to_value(u.m) == pytest.approx(np.sqrt(3))  # flat-to-flat
+    assert np.isnan(size[3].to_value(u.m))
+
+
```
Voutsi:
```suggestion
@requires_test_data
```

src/ctapipe/instrument/tests/test_MirrorFacetsDescription.py
```diff
@@ -0,0 +1,177 @@
...
+def test_mirror_facets_description_from_ecsv():
+    description = MirrorFacetsDescription.from_table(ECSV_PATH)
+    _check_description(description)
+
+
+def test_mirror_facets_description_from_fits():
+    description = MirrorFacetsDescription.from_table(FITS_PATH)
+    _check_description(description)
+
+
```
Voutsi:
```suggestion
@requires_test_data
```

src/ctapipe/instrument/tests/test_MirrorFacetsDescription.py
```diff
@@ -0,0 +1,177 @@
...
+    )
+    assert isinstance(hexagon, RegularPolygon)
+    assert hexagon.radius == pytest.approx(1.0 / np.sqrt(3))
+
+    with pytest.raises(ValueError, match="Unsupported mirror facet shape"):
+        MirrorFacetsDescription.create_patches(
+            MirrorFacetShape.UNKNOWN, [0.0], [0.0], [1.0]
+        )
+
+
```
Voutsi:
```suggestion
@requires_test_data
```

src/ctapipe/instrument/tests/test_MirrorFacetsDescription.py
```diff
@@ -0,0 +1,177 @@
...
+        "surface",
+        "shape",
+    ]
+    assert table.meta["EXTNAME"] == "MIRRORS"
+
+    roundtripped = MirrorFacetsDescription.from_table(table)
+    _check_description(roundtripped)
+    np.testing.assert_array_equal(roundtripped.id, description.id)
+
+
```
Voutsi:
```suggestion
@requires_test_data
```

src/ctapipe/instrument/optics.py
```diff
@@ -97,6 +97,274 @@ class ReflectorShape(Enum):
...
+            table = Table.read(table, **kwargs)
+
+        table = QTable(table)
+
+        # FITS tables are read back with byte strings by default, while
+        # ECSV tables are read back as unicode strings.
+        mirror_shape = np.asarray(table["shape"])
+        if mirror_shape.dtype.kind == "S":
+            mirror_shape = mirror_shape.astype(str)
+
```
Voutsi, line 206: I suggest to enforce units here.

src/ctapipe/instrument/tests/test_MirrorFacetsDescription.py
maxnoe: This whole mechanism is not ok and needs to be removed.

We use synthetic test data or the test data server via `get_dataset_path`.

Voutsi:
For using the test files from MinIO I suggest you to have a look how it is done in e.g. test_subarray, using the get_dataset_path.

So you can use a fixture like:

```
MIRROR_LIST = "mirror_CTA-N-LST1_v2019-03-31.ecsv"


@pytest.fixture(scope="session")
def lst1_mirror_path():
    """Path to the real LST-1 mirror facet list from the test data server."""
    return get_dataset_path(MIRROR_LIST)


@pytest.fixture(scope="session")
def lst1_facets(lst1_mirror_path):
    return MirrorFacetsDescription.from_table(lst1_mirror_path)
```

A more minor comment, I would also rename the test file such as: test_mirror_facets_description.py to be more ctapipe-like.

src/ctapipe/instrument/tests/test_MirrorFacetsDescription.py
Voutsi: On second thought I agree. We should remove the pytest.mark.skipif and ignore my suggestions (I will remove them) otherwise we might have silently failing tests.

> We use synthetic test data or the test data server via get_dataset_path.

Indeed. @burmist-git please see the comment below https://github.com/cta-observatory/ctapipe/pull/3085#issuecomment-5618387319

src/ctapipe/instrument/optics.py
burmist-git: done

src/ctapipe/instrument/optics.py
burmist-git: done

src/ctapipe/instrument/optics.py
```diff
@@ -97,6 +97,274 @@ class ReflectorShape(Enum):
...
+        return cls(
```
Voutsi, line 207: Please check if the input file has units. In case it doesn't have, I suggest to raise ValueError

Voutsi:
Please add MirrorDescription and MirrorFacetShape to instrument/__init__.py and to __all__ in optics.py.

src/ctapipe/instrument/tests/test_MirrorDescription.py
```diff
@@ -0,0 +1,160 @@
...
+    assert description.z[0].to_value(u.cm) == pytest.approx(120.53307587693142)
+    assert description.surface_area[0].to_value(u.cm**2) == pytest.approx(
+        19746.245231688983
+    )
+
+    assert description.nx[0] == pytest.approx(-0.08077963744975176)
+    assert description.ny[0] == pytest.approx(0.18640162657079892)
+    assert description.nz[0] == pytest.approx(0.9791471206030518)
+
+    assert np.all(description.shape == MirrorFacetShape.HEXAGON)
```
Voutsi: I would rather change 'shape' to something like facet_shape. The reason is to avoid confusion with numpy attribute shape.

PR #3089, opened 2026-09-07
Reduce log level to DEBUG for skipping columns in hdf5 reader/writer

maxnoe:
We now have a lot of optional columns in several stages that we intentionally skip writing/reading if they are not present.

Thus, the log level of WARNING is not appropriate, these are entirely expected messages in most cases.


kosack:
Sounds reasonable to me

kosack (approved):

Issue #3088, opened 2026-09-07
Reading ctapipe-format R1 files broken for simtel input


kosack:
**Describe the bug**

Processing a file to R1, and then processing that file to DL1 fails due to missing calibration monitoring information.

**To Reproduce**

```bash
pixi init
pixi add ctapipe
make
```

Makefile:
```makefile
SIMULATION_FILE=gamma_20deg_0deg_run000999___cta-prod6-2156m-LaPalma-dark+magic_cone10.simtel.zst
PROCESSING_CONFIG=dl0_to_dl1.yml
MAX_EVENTS=100

all: events-dl1-compression.h5

events-r1.h5: $(SIMULATION_FILE) 
	pixi run ctapipe-process --config $(PROCESSING_CONFIG) \
        --input $(SIMULATION_FILE) \
        --output "events-dl0-compression.h5" \
        --DataWriter.write_r1_waveforms=True \
        --max-events $(MAX_EVENTS) \
        --overwrite --progress


events-dl1.h5: events-r1.h5
	pixi run ctapipe-process --config $(PROCESSING_CONFIG) \
		--input events-r1.h5 \
		--output events-dl1.h5 \
		--write-images --write-parameters \
		--max-events $(MAX_EVENTS) \
		--overwrite --progress
```

**Expected behavior**

We can process files written at any data level allowed as output by ctapipe-process.

**Supporting information**

The error is:
```
2026-09-07 17:11:02,192 ERROR [ctapipe.ctapipe-process] (tool.run): Caught unexpected exception: 'NoneType' object is not subscriptable
Traceback (most recent call last):
  File "/Users/kkosack/Projects/CTA/Working/datapipe-verification-playground/DL0CompressionEffect/.pixi/envs/default/lib/python3.14/site-packages/ctapipe/core/tool.py", line 437, in run
    self.start()
    ~~~~~~~~~~^^
  File "/Users/kkosack/Projects/CTA/Working/datapipe-verification-playground/DL0CompressionEffect/.pixi/envs/default/lib/python3.14/site-packages/ctapipe/tools/process.py", line 372, in start
    self.calibrate(event)
    ~~~~~~~~~~~~~~^^^^^^^
  File "/Users/kkosack/Projects/CTA/Working/datapipe-verification-playground/DL0CompressionEffect/.pixi/envs/default/lib/python3.14/site-packages/ctapipe/calib/camera/calibrator.py", line 309, in __call__
[...]

PR #3090, opened 2026-09-07
Fix ImageExtractors for case where broken_pixels is None

maxnoe:
While this is only a sympton, I think `ImageExtractor` should handle the case of `broken_pixels=None` as "No broken pixels".


kosack (approved):
This sounds reasonable.  It does make #3088 work, though of course silently incorrect, but that can be fixed by added a test in a separate PR to ensure monitoring gets propegated to R1/DL0

maxnoe:
Better solved with #3095

PR #3092, opened 2026-09-08
Make time_tolerance of HDF5MonitoringSource configurable, set default to 1s

maxnoe:
Fixes case where `HDF5MonitoringSource` rejects the expected case of a physics event arriving before the first calibration event

PR #3094, opened 2026-09-08
Validity time range for camera calibration based on pointing timestamps

Voutsi:
An alternative approach to https://github.com/cta-observatory/ctapipe/pull/3092/changes that is based on the assumption that telescope pointing monitoring data should cover the whole OB.


maxnoe:
I really don't want to conflate camera calibrations with pointing data.

This is not ok, these are completely independent.

At the very least, this logic should not be in the code _reading_ the calibration coefficients but the file should be written with correct validity range.

Using the timestamp of the first calibration event is certainly not correct.

PR #3095, opened 2026-09-09
Fill broken pixels from pixel_status in HDF5EventSource

maxnoe:
Test demonstrating #3088

### DIFF
--- a/src/ctapipe/io/hdf5eventsource.py
+++ b/src/ctapipe/io/hdf5eventsource.py
@@ -11,6 +11,7 @@ from astropy.utils.decorators import lazyproperty
 from ..atmosphere import AtmosphereDensityProfile
 from ..containers import (
     ArrayEventContainer,
+    CameraCalibrationContainer,
     CameraHillasParametersContainer,
     CameraTimingParametersContainer,
     ConcentrationContainer,
@@ -30,6 +31,7 @@ from ..containers import (
     ObservationBlockContainer,
     ParticleClassificationContainer,
     PeakTimeStatisticsContainer,
+    PixelStatus,
     R1CameraContainer,
     ReconstructedEnergyContainer,
     ReconstructedGeometryContainer,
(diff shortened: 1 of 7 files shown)


src/ctapipe/io/hdf5eventsource.py
```diff
@@ -938,15 +940,21 @@ def _fill_r1_for_tel(self, data, tel_id, key, waveform_readers):
...
             )
-            data.r1.tel[tel_id].waveform = r1_waveform[np.newaxis, ...]
+            r1.waveform = r1.waveform[np.newaxis, ...]
+
+        data.r1.tel[tel_id] = r1
+        # fill calibration outlier_mask initially from pixel status
+        disabled_pixels = PixelStatus.get_channel_info(r1.pixel_status) == 0
+        data.monitoring.tel[tel_id].camera.coefficients = CameraCalibrationContainer(
+            outlier_mask=disabled_pixels,
+        )
```
Hckjs, line 968: From the Field docstring of `outlier_mask` it should have shape (n_channels, n_pixels), but its now just one-dimensional (n_pixels,). Maybe also check shapes in the unit test

maxnoe: good catch, thanks. Wonder why it works then with the test...

Hckjs: the `n_dim` is not set for the `outlier_mask`-Field, its also missing from `time_shift`. Maybe we should add that? But i think the test would still fail, because the container is not validated/written afterwards?

maxnoe: Test just iterated through the source and did not call `CameraCalibrator`, updating the test now.

maxnoe:
> the n_dim is not set for the outlier_mask-

Added ndim and dtype for these

src/ctapipe/io/hdf5eventsource.py
```diff
@@ -938,15 +940,34 @@ def _fill_r1_for_tel(self, data, tel_id, key, waveform_readers):
...
+        data.r1.tel[tel_id] = r1
+        # fill calibration outlier_mask initially from pixel status
+
+        n_channels = self.subarray.tel[tel_id].camera.readout.n_channels
+        disabled_pixels = np.zeros((n_channels, len(r1.pixel_status)), dtype=bool)
+
+        if n_channels == 2:
+            disabled_pixels[0] = (
+                r1.pixel_status & PixelStatus.HIGH_GAIN_STORED
+            ).astype(bool)
```
Hckjs: == 0, instead of .astype(bool)?  same below. Otherwise its always true except when the pixel is disabled (==0)

Hckjs: Or do we need to check both gain bits? the mask should just be true (for each channel) if the pixel is neither as High nor as low gain stored, right?

maxnoe: I think in the case where only one channel is broken, we expect this to be reflected here. I.e. in this case `outlier_mask[0] != outlier_mask[1]`

Hckjs: But then we still need to invert 
```
(r1.pixel_status & PixelStatus.HIGH_GAIN_STORED).astype(bool)
```
because this way its true, when the pixel is stored in highgain, dont we?

maxnoe: yes, correcto

maxnoe: fixed

kosack (dismissed):

Hckjs (dismissed):

maxnoe:
I needed to adjust one test that was abusing the container definition by putting multiple calibration coefficient sets for multiple times into the same container.

kosack (approved):

PR #3096, opened 2026-09-09
Update abstract MonitoringSource interface

### LINKED ISSUE #3093: Remove **kwargs from MonitoringSource abstract interface
**Describe the bug**

"Implementation-specific kwargs" cannot be known by the caller side, so these cannot be part of an abstract interface.

@tjarkmiener @mexanick @Voutsi 

Please advise what these kwargs are / were used for and propose how to change the interface to make this use cases possible without allowing or requiring  `**kwargs` in the abstract interface.

**To Reproduce**

See:

https://github.com/cta-observatory/ctapipe/blob/ca3ee634c6628afaea77acc6eeeb206b500feeac/src/ctapipe/io/monitoringsource.py#L69-L87

and 

https://github.com/cta-observatory/ctapipe/blob/ca3ee634c6628afaea77acc6eeeb206b500feeac/src/ctapipe/io/monitoringsource.py#L102-L124

In most cases, I would expect that these should be transformed into configuration traitlets of the concrete sources instead of method arguments, otherwise they cannot be influenced when the classes are used in ctapipe tools.

See #3092

maxnoe:
I noticed several issues that I try to improve here by updating slightly the abstract interface of the `MonitoringSource`:

1. Separate methods for array-wide and telescope-specific monitoring
   
This is because I see in the implementations a lot of checks about if tel_id is given for a given monitoring type or not. Separating methods for array-wide and telescope-wise monitoring allows making `tel_id` a required arg for the telescope data and not present at all for the array-wide data.   

2. Remove **kwargs, cannot be used by a generic caller side, add subtype as argument

`**kwargs` need to be removed, they cannot be known by a generic caller side only knowing the abstract interface. We cannot have implementation specific arguments, as these will never be able to be passed by e.g. `ctapipe-process`, which can only know the abstract interface.

3.  Let available data getters return all available (type, subtype) combinations
  
Same reasoning as above: if subtypes are needed (which I gather they are because enumerating all possible event types / peak time vs. image pixel statistics as main types would not be great), they need to be discoverable via the abstract interface.

Fixes #3093 

Please have a look here at @kosack @Voutsi @mexanick. If we agree with the appraoch, then I'll update the implementation in the `HDF5MonitoringSource` and also start working on making the `SimTelEventSource` and `HDF5EventSource` monitoring sources, so we can solve #3088 by including the simtel monitoring data in the output.


maxnoe:
@codex review

Please ignore test failures and missing updates on the concrete implementations for now, only review the changes to the abstract class.

src/ctapipe/io/monitoringsource.py
```diff
@@ -94,38 +121,68 @@ def get_table(
...
     @abstractmethod
     def get_values(
         self,
-        monitoring_type: MonitoringType,
         time: astropy.time.Time,
-        tel_id: int = None,
-        **kwargs,
-    ):
+        monitoring_type: MonitoringType,
+        subtype: str | None = None,
```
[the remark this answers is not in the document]
maxnoe: The existing code only worked for telescope monitoring types anyway, so no need to keep this backwards compatible

src/ctapipe/io/monitoringsource.py
```diff
@@ -40,51 +41,77 @@ class MonitoringSource(TelescopeComponent):
...
+        available_data : tuple[tuple[ctapipe.io.MonitoringType, str | None]]
+            A tuple of (type, subtype) pairs of the available monitoring data.
+        """
 
     @property
     @abstractmethod
-    def monitoring_types(self) -> tuple[MonitoringType]:
+    def available_telescope_data(
+        self,
+    ) -> dict[int, tuple[TelescopeMonitoringType, str | None]]:
```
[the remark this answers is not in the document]
maxnoe: done

mexanick:
LGTM, thank you!

src/ctapipe/io/monitoringsource.py
```diff
@@ -40,51 +41,77 @@ class MonitoringSource(TelescopeComponent):
 
     plugin_entry_point = "ctapipe_monitoring"
 
-    def __init__(self, subarray=None, config=None, parent=None, **kwargs):
-        super().__init__(subarray=subarray, config=config, parent=parent, **kwargs)
-        self.metadata = {"is_simulation": False}
+    @property
+    @abstractmethod
+    def available_data(self) -> tuple[tuple[MonitoringType, str | None], ...]:
```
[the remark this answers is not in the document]
maxnoe: Solved via allowing `subarray=None` in general

src/ctapipe/tools/process.py
```diff
@@ -259,6 +240,64 @@ def setup(self):
...
+            }
+            if not available_types.intersection(COMPATIBLE_MONITORINGTYPES):
+                msg = (
+                    f"'{mon_source_name}' does not provide any "
+                    f"of the known monitoring types: {COMPATIBLE_MONITORINGTYPES}, "
+                    f"{mon_source_name} provides only '{mon_source.available_telescope_data}'. "
+                    f"Please make sure the '{mon_source_name}' and its input "
+                    f"are suitable for calibrating the data you are processing."
+                )
+                self.log.warning(msg)
```
Voutsi, line 297: Why the behaviour changed from raising to warning?

maxnoe: Good point, I did not add this to the changelog, but I don't really think it's a fundamental issue that stop all processing when there is a monitoring source configured that does not have data.

In that case `fill_event` should just be a no-op and not crash data processing.

src/ctapipe/io/monitoringsource.py
```diff
@@ -40,51 +44,75 @@ class MonitoringSource(TelescopeComponent):
 
     plugin_entry_point = "ctapipe_monitoring"
 
-    def __init__(self, subarray=None, config=None, parent=None, **kwargs):
```
Voutsi, line 43: So now TelescopeComponent can accept subarray=None, however by removing the __init__ here means that the subarray=None should be moved to the subclasses right? I don't say that's wrong, it looks ok imo, however I would like to understand the motivation for this change.

maxnoe: There is no reason why the `MonitoringSource` should be special in this regard. After enabling `subarray=None` for the parent class `TelescopeComponent`, the `__init__` became superflous, because the `subarray` handling was all it did.

maxnoe: Ah, I think I understand your comment now.

I intentionally did not change the default of `TelescopeComponent` for subarray, it is still a positional argument. You can however now _explicitly_ pass `subarray=None`.

I think there are very limit usecases for not passing a subarray to a `TelescopeComponent` (including `MonitoringSource`), so the default should not just be `subarray=None`.

kosack (dismissed):

Voutsi (dismissed):

maxnoe:
@codex review

src/ctapipe/core/telescope_component.py
```diff
@@ -41,6 +41,9 @@ def subarray(self):
     @subarray.setter
     def subarray(self, subarray):
         self._subarray = subarray
+        if subarray is None:
+            return
```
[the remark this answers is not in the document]
maxnoe: This is the desired behavior, only explicit `subarray=None` should be allowed.

src/ctapipe/io/hdf5monitoringsource.py
```diff
@@ -252,81 +256,76 @@ def _process_single_file(self, file):
...
-                for col in ["mean", "median", "std"]:
-                    self._pixel_statistics[tel_id][name][col][
-                        self._pixel_statistics[tel_id][name]["outlier_mask"].data
-                    ] = np.nan
+                if name not in self._pixel_stats:
+                    try:
+                        interpolator = _interpolators[name](parent=self)
+                    except KeyError:
+                        msg = f"Unsupported pixel statistics subtype '{name}'"
+                        raise ValueError(msg) from None
```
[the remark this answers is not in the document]
maxnoe: Skipping and logging skipped groups instead of raising an exception.

src/ctapipe/tools/process.py
```diff
@@ -259,6 +240,64 @@ def setup(self):
...
+            )
+
+            self.log.info(
+                "%s provides the following subarray monitoring data:", mon_source_name
+            )
+            for mon_type, subtype in mon_source.available_data:
+                self.log.info(
+                    "  - %s%s",
+                    mon_type.name,
+                    f", {subtype}" if subtype is not None else "",
```
mexanick:
```suggestion
                    ", %s" % subtype if subtype is not None else "",
```
How did this slipped through ruff?

src/ctapipe/tools/process.py
```diff
@@ -259,6 +240,64 @@ def setup(self):
...
+                tel_id,
+                available_tel_data,
+            ) in mon_source.available_telescope_data.items():
+                self.log.log(level, "  tel_id = %3d", tel_id)
+                for mon_type, subtype in available_tel_data:
+                    self.log.log(
+                        level,
+                        "    - %s%s",
+                        mon_type.name,
+                        f", {subtype}" if subtype is not None else "",
```
mexanick:
```suggestion
                        ", %s" % subtype if subtype is not None else "",
```

src/ctapipe/io/hdf5monitoringsource.py
```diff
@@ -252,81 +256,77 @@ def _process_single_file(self, file):
...
 
-            for name, interpolator in self.pixel_stats_dict.items():
-                # Read the tables from the monitoring file
-                self._pixel_statistics[tel_id][name] = read_table(
-                    file,
-                    f"{DL1_PIXEL_STATISTICS_GROUP}/{name}/tel_{tel_id:03d}",
-                )
+                if name not in _interpolators:
+                    self.log.info(
+                        f"Skipping unsupported pixel statistics subtype '{name}'"
```
mexanick:
```suggestion
                        "Skipping unsupported pixel statistics subtype '%s'" % name
```

mexanick (dismissed):
Looks very good, just a couple of nitpicks considering logging with f-strings.

src/ctapipe/tools/process.py
maxnoe: Because it's not in the main log string and the proposed change doesn't work

src/ctapipe/io/hdf5monitoringsource.py
maxnoe: Fixed

src/ctapipe/tools/process.py
maxnoe: Updated to improve readability.

src/ctapipe/tools/process.py
maxnoe: Updated to improve readability

mexanick (approved):

maxnoe:
@kosack @Voutsi Please have a look again

Voutsi (approved):

PR #3099, opened 2026-09-17
Add angular-error regressor and associated disp combination weighting scheme

STSpencer:
As discussed in #3062, it would be helpful if ctapipe supported the ability to estimate the error on a per-telescope angular reconstruction. This is needed for a few reasons:

1. To be able to perform event-class analysis following the prescription of https://arxiv.org/abs/2607.09286. This PR would be a step towards that in mono, but not stereo (which I think would need to be another PR).

2. To replicate one of the options in eventdisplay for stereo disp combination, which uses this per-telescope angular error estimate in an exponential weighting term. According to @GernotMaier this is what has actually been used for the eventdisplay prod6 analysis and not the simpler truncation fourth power weighting scheme described in #3062. 

3. For divergent pointing analysis; as the odds of an image being truncated in divergent pointing mode are a function of the telescope position in the array, using a direct truncation penalty in the stereo disp combination apparently results in inferior performance. For this reason I added the telescope position as a feature in the regressor.

This PR is my attempt to implement this using copilot and claude opus 4.8. Daniele Ambrosino is going to test it out. I've tried to ensure future compatibility with #2731.

Here's the summary from Claude of what this PR changes:

**Description**
This PR adds an optional angular-error regressor to the DispReconstructor, replicating the EventDisplay approach of predicting each telescope's expected angular reconstruction error and using it as an exponential penalty weight when combining per-telescope disp directions into the array-level direction.

**What's added**
Angular-error regressor (ctapipe.reco.sklearn.DispReconstructor): when predict_angular_error=True, ctapipe-train-disp-reconstructor trains an additional per-telescope-type regressor after the norm/sign models. Its target is the angular distance between the disp-reconstructed and true source position (compute_true_angular_error in reco/disp.py). The prediction is stored as the per-telescope ang_distance_uncert, both in the event loop and in predict_table. Backwards compatible with previously trained models.
Telescope position as a feature: angular_error_features can include pos_x, pos_y, pos_z so the regressor accounts for divergent pointing. The training tool loads these features and retains the columns needed to compute the target.
Exponential-penalty stereo weighting: new weights = "angular-error" option and angular_error_scale trait on the base StereoCombiner, applying exp(-ang_distance_uncert / angular_error_scale). The default angular_error_scale = 0.2 deg reproduces the EventDisplay default exp(-5 · ang_distance_uncert) (DispError_BDTWeight = 5). Placing these on the base class keeps the weighting available to any combiner.
Example config via ctapipe-quickstart: train_disp_reconstructor_angular_error.yaml is shipped as a resource and written out by ctapipe-quickstart.
Docs (dl2_guide.rst) and a changelog entry.

**Behavior / compatibility**
Default behavior is unchanged: the regressor is only trained when predict_angular_error: True, and only used for weighting when weights: "angular-error".
The weighting logic lives on the base StereoCombiner, so it remains forward-compatible with additional disp stereo combiners.

**Tests**
New unit tests for the angular-error stereo weighting (table and event-wise paths, plus fallback branches).
New training-tool test exercising the shipped example config end-to-end (including the pos_x/pos_y/pos_z divergent-pointing features).
Existing disp/stereo/apply/process tests pass.


src/ctapipe/reco/disp.py
```diff
@@ -104,3 +105,42 @@ def compute_true_disp(
...
+    reco_lat = table["hillas_fov_lat"].quantity + reco_disp * np.sin(psi)
+
+    true_lon, true_lat = horizontal_to_telescope(
+        alt=table["true_alt"],
+        az=table["true_az"],
+        pointing_alt=pointing_alt,
+        pointing_az=pointing_az,
+    )
+
+    return np.sqrt((reco_lon - true_lon) ** 2 + (reco_lat - true_lat) ** 2)
```
maxnoe: This is a small angle approximation, please use `angular_separation`

maxnoe:
@STSpencer I think it would be better to discuss such features in issues before starting an implementation.

There is also related work, using a separate regressor, not one tied to the disp regressor here: https://github.com/cta-observatory/ctapipe/pull/2503

I think this should not be tied to the disp regressor, as it can work for all reconstruction algorithms if it directly worked on the telescope-wise reconstructed alt/az values instead of disp values.

src/ctapipe/reco/tests/test_disp.py
```diff
@@ -59,3 +59,30 @@ def test_compute_true_disp(subarray_pointing):
...
+    true_az = [0, 10, 180] * u.deg
+
+    error = compute_true_angular_error(reco_alt, reco_az, true_alt, true_az)
+
+    # matches astropy's great-circle angular separation
+    expected = angular_separation(reco_az, reco_alt, true_az, true_alt)
+    np.testing.assert_allclose(error.to_value(u.deg), expected.to_value(u.deg))
+
+    # zero error when reconstructed == true direction
+    assert error[0].to_value(u.deg) == pytest.approx(0.0, abs=1e-9)
```
maxnoe: Use `u.isclose` or `from astropy.tests.helper import assert_quantity_allclose`

